PR #1181, opened 2021-04-21
ConfirmationDialog

### LINKED ISSUE #1065: Dialog
## Description
A dialog is a type of overlay that can be used for confirming actions, asking for disambiguation, and presenting small forms. They generally allow the user to focus on a quick task without having to navigate to a different page.

**Dialogs appear in the page after a direct user interaction**. Don't show dialogs on page load or as system alerts.

**Dialogs appear centered in the page**, with a visible backdrop that dims the rest of the window for focus.

**All dialogs should have a title and a close button**. Use an overlay header to compose these required elements.

**Dialogs are modal**. Dialogs can be dismissed by clicking on the close button, or by interacting with another button in the overlay. To avoid losing information and missing important messages, clicking outside of the dialog will not close it.

**Make sure larger dialogs remain mobile-friendly**. Even large dialogs need to be responsive. A dialog's width and height will be readjusted depending on the screen size and should never exceed the available space. Use responsive solutions to adjust the UI so they behave well on smaller screens.

**Simple and small dialogs can remain as-is on mobile**. As more elements are added to it, mobile-specific style variations such as **Bottom sheet** and **Full-screen** should be considered.

## Resources
Figma:
Prototype:

## Screenshot
dialog

#### Behaviors

- No body scrolling when dialog is active.

- Display a browser confirmation dialog when leaving tab or updating history state when dialog contains unsaved data (e.g. changed form fields).

#### Props

name: `width`
type: `"auto" | 296 | 320 | 480 | 640` (px)
default: `"auto"`
description:

-----

name: `height`
type: `"auto" | 480 | 640`
default: `"auto"`
description:

T-Hugs:
This PR adds a `ConfirmationDialog` component and a helper hook, `useConfirm`. A `ConfirmationDialog` is an opinionated dialog that is used specifically to confirm a user's choice. See the [rendered documentation](https://primer-components-7no4mqitb-primer.vercel.app/components/Dialog2) for more details.

Closes #1065 

### Screenshots
image

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


docs/content/Dialog2.mdx
```diff
@@ -105,8 +105,107 @@ The `DialogHeaderProps` interface extends `DialogProps` and adds these additiona
...
+An alternate way to use `ConfirmationDialog` is with the `useConfirm()` hook. It takes no parameters and returns an `async` function, `confirm`. When `confirm` is called (see ConfirmOptions below for its argument), it shows the confirmation dialog. When the dialog is dismissed, a promise is resolved with `true` or `false` depending on whether or not the confirm button was used.
+
+### Example (with `useConfirm` hook)
+
+```javascript live noinline
+function ShorthandExample() {
+  const confirm = useConfirm()
+  const buttonClick = React.useCallback(
+    e => {
+      // @todo The docs site live code examples don't work with async/await, so using the raw promise instead!
```
smockle, line 129: Out of curiosity: Do you know why the docs site live code examples don’t work with `async`/`await`? And is that tracked in an issue?

T-Hugs: My guess is it's an old babel dependency or configuration. I don't believe it's tracked in an issue. I assume this would be a doctocat bug? I'm not exactly sure how the live code samples get compiled.

colebemis: Looks like there's an open issue for it in `react-live` (the library we use for live code examples) https://github.com/FormidableLabs/react-live/issues/206

T-Hugs: Updated the sample to use `async function()` since apparently that works 🤷

docs/content/Dialog2.mdx
```diff
@@ -105,8 +105,80 @@ The `DialogHeaderProps` interface extends `DialogProps` and adds these additiona
...
+An alternate way to use `ConfirmationDialog` is with the `useConfirm()` hook. It takes no parameters and returns an `async` function, `confirm`. When `confirm` is called (see ConfirmOptions below for its argument), it shows the confirmation dialog. When the dialog is dismissed, a promise is resolved with `true` or `false` depending on whether or not the confirm button was used.
+
+### Example (with `useConfirm` hook)
+
+```javascript live noinline
+function ShorthandExample2() {
+  const confirm = useConfirm()
+  const buttonClick = React.useCallback(
+    async function (e) {
+      // @todo The docs site live code examples don't work with async/await, so using the raw promise instead!
```
VanAnderson, line 129: It looks like you found a way to use async await instead of working with the raw promise, is this comment still valid?

docs/content/Dialog2.mdx
```diff
@@ -105,8 +105,80 @@ The `DialogHeaderProps` interface extends `DialogProps` and adds these additiona
...
+    )
+  }}
+</State>
+```
+
+### ConfirmationDialogProps
+
+| Prop name          | Type                                                                  | Default    | Description                                                                                                                   |
+| :----------------- | :-------------------------------------------------------------------- | :--------- | :---------------------------------------------------------------------------------------------------------------------------- |
+| title              | `string`                                                              |            | Required. Sets the title of the dialog, which by default is also used as the `aria-labelledby` attribute.                     |
```
VanAnderson, line 170: has the behavior of assigning the title to `aria-labelledby` been implemented? Can't see where that is happening in either Dialog or ConfirmationDialog

VanAnderson (approved):
Just a couple of small things I saw that I had some questions on, probably nothing blocking. Really great job on this!

src/Dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,190 @@
...
+    margin: 0;
+    border-right: 0;
+    &:last-child {
+    }
+    &:first-child {
+      border-left: 0;
+    }
+  }
+`
+const ConfirmationFooter: React.FC<DialogProps> = ({footerButtons}) => {
```
dgreif, line 89: Do we want to create a focus zone in here?

dgreif (changes requested):
Really cool to see these components starting to stack!  Great implementation, this is going to be really fun to use ✨

docs/content/Dialog2.mdx
T-Hugs: Whoops, forgot to remove the comment. Thanks!

docs/content/Dialog2.mdx
T-Hugs: It happens here: https://github.com/primer/components/blob/main/src/Dialog/Dialog.tsx#L286

Were you expecting something else?

src/Dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const onCancelButtonClick = useCallback(() => {
+    onClose('cancel')
+  }, [onClose])
+  const onConfirmButtonClick = useCallback(() => {
+    onClose('confirm')
+  }, [onClose])
+  const cancelButton: DialogButtonProps = {
+    text: cancelButtonLabel,
+    onClick: onCancelButtonClick,
+    autoFocus: true,
```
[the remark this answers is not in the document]
T-Hugs: Precisely. In theory, you are only going to show a confirmation dialog for an action that is difficult or impossible to reverse, so auto-focusing the confirm button seems a bit dangerous.

src/Dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,190 @@
...
+    </Dialog>
+  )
+}
+
+export type ConfirmOptions = Omit<ConfirmationDialogProps, 'onClose'> & {description: string}
+async function confirm(themeProps: ThemeProviderProps, options: ConfirmOptions): Promise<boolean> {
+  const {description, ...confirmationDialogProps} = options
+  return new Promise(resolve => {
+    const hostElement = document.createElement('div')
+    document.body.append(hostElement)
```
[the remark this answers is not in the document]
T-Hugs: I am not sure React has a way to run render without a host element (even though it goes completely untouched in this case). I could render directly into `document.body` but React generates a warning when you try that.

src/Dialog/ConfirmationDialog.tsx
T-Hugs: Yes! Accidentally left that out.

src/Dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * The text to use for the cancel button. Default: "Cancel".
+   */
+  cancelButtonLabel?: string
+
+  /**
+   * The text to use for the confirm button. Default: "OK".
+   */
+  confirmButtonLabel?: string
```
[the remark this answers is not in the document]
T-Hugs: Yeah, I considered this, just didn't get around to it. I will make the change.

src/Dialog/ConfirmationDialog.tsx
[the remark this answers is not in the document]
T-Hugs: I tried calling `ReactDOM.render` with `null` as the second argument, and I get an error.

src/Dialog/ConfirmationDialog.tsx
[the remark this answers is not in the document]
T-Hugs: I made similar changes throughout Dialog. Let me know if you think it's too overboard.

src/Dialog/ConfirmationDialog.tsx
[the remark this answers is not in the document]
dgreif: Makes sense.  In that case, we may want to tweak the focus styles.  Right now it's not rendering well in chrome with the outline, and only shows the outline if you opened the dialog with the keyboard

image

src/Dialog/ConfirmationDialog.tsx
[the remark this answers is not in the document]
dgreif: You can still _create_ the div, but just not attach it to the DOM.  I tried it yesterday and it seemed to work fine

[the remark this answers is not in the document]
T-Hugs: Oh, interesting. Fixed!

src/Dialog/ConfirmationDialog.tsx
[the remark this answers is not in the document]
T-Hugs: Yeah, I'm aware of this.

> only shows the outline if you opened the dialog with the keyboard
This is intended.

dgreif (approved):

src/Dialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,192 @@
...
+  return (
+    <StyledConfirmationHeader>
+      <StyledTitle id={dialogLabelId}>{title}</StyledTitle>
+      <Dialog.CloseButton onClose={onCloseClick} />
+    </StyledConfirmationHeader>
+  )
+}
+const StyledConfirmationBody = styled(Box)`
+  font-size: ${get('fontSizes.1')};
+  padding: 0 ${get('space.3')} 22px ${get('space.3')};
```
vdepizzol, line 67:
```suggestion
  padding: 0 ${get('space.3')} ${get('space.3')} ${get('space.3')};
```

The Figma mockup had `22px` as a total calculation considering this would use an `OverlayHeader` to generate the Title + description (8px padding + original 32px total size minus the text line-height). Since this is a one-off, let's keep it at `space.3`. ✌️

docs/content/Dialog2.mdx
VanAnderson: whoops, must have missed that!

PR #1185, opened 2021-04-28
fix(useFocusZone): handle add/remove nodes that have focusable children

dgreif:
Within a focus zone, we watch for child nodes to be added/removed.  Currently, we only do something with those changes if the node that was changed `isFocusable`, or has been registered as a focus target.  _If_ that condition is met, we search all children of the changed node to find focus targets.  This setup does not account for scenarios when the added/removed node is not _itself_ focusable, but it does have children that are focusable.  This change removes the top-level check and simply always searches for focusable children.  This is a noticeable issue in `SelectPanel`, where the list items are added/removed frequently with changes to the filter.

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


T-Hugs (approved):
Good fix! I think I had the `isFocusable` check in there before adding support for iterating the focusable children and didn't think to remove it.

PR #1186, opened 2021-04-28
ActionList and ActionMenu API updates

VanAnderson:
This PR:

- makes icons respect the variant styling
- adds a disabled flag that styles an item as disabled and also takes it out of the tab order
- adds item level `onAction` callback

Closes https://github.com/github/primer/issues/125
Closes https://github.com/github/primer/issues/126
Closes https://github.com/github/primer/issues/127

### Screenshots
A bit of a contrived example here, but this shows both disabled and variant styling. Wondering if the `trailingText` should also be styled by the variant?

Screen Shot 2021-04-28 at 10 16 32 AM

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -69,10 +69,7 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
   /**
    * Callback that will trigger both on click selection and keyboard selection.
    */
-  onAction?: (
-    item: Partial<ItemProps>,
-    event?: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>
-  ) => void
+  onAction?: (item: ItemProps, event?: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>) => void
 }
 
 const getItemVariant = (variant = 'default', disabled?: boolean) => {
@@ -106,7 +103,7 @@ const getItemVariant = (variant = 'default', disabled?: boolean) => {
   }
 }
 
-const StyledItem = styled.div<{variant: ItemProps['variant']; item?: ItemProps} & SxProp>`
+const StyledItem = styled.div<{variant: ItemProps['variant']; item?: ItemInput} & SxProp>`
   /* 6px vertical padding + 20px line height = 32px total height
    *
    * TODO: When rem-based spacing on a 4px scale lands, replace
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -4,6 +4,7 @@ import {get} from '../constants'
 import sx, {SxProp} from '../sx'
 import {ItemInput} from './List'
 import styled from 'styled-components'
+import {themeGet} from '..'
 
 /**
  * Contract for props passed to the `Item` component.
@@ -59,9 +60,50 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
    * Designates the group that an item belongs to.
    */
   groupId?: string
+
+  /**
+   * Items that are disabled can not be clicked, selected, or navigated through.
+   */
+  disabled?: boolean
+
+  /**
+   * Callback that will trigger both on click selection and keyboard selection.
+   */
+  onAction?: (item: Partial<ItemProps>, event?: MouseEvent | KeyboardEvent) => void
 }
 
-const StyledItem = styled.div<{variant: ItemProps['variant']} & SxProp>`
+const getItemVariant = (variant = 'default', disabled?: boolean) => {
+  if (disabled) {
+    return {
+      color: themeGet('colors.text.disabled'),
+      iconColor: themeGet('colors.text.disabled'),
+      annotationColor: themeGet('colors.text.disabled'),
+      hoverBackground: 'inherit',
+      hoverCursor: 'default'
+    }
+  }
+
+  switch (variant) {
+    case 'danger':
+      return {
+        color: themeGet('colors.text.danger'),
+        iconColor: themeGet('colors.icon.danger'),
+        annotationColor: themeGet('colors.text.disabled'),
+        hoverBackground: themeGet('colors.bg.danger'),
+        hoverCursor: 'pointer'
+      }
+    default:
+      return {
+        color: 'inherit',
+        iconColor: themeGet('colors.text.disabled'),
+        annotationColor: themeGet('colors.text.disabled'),
+        hoverBackground: themeGet('colors.selectMenu.tapHighlight'),
+        hoverCursor: 'pointer'
+      }
+  }
+}
+
+const StyledItem = styled.div<{variant: ItemProps['variant']; item?: ItemProps} & SxProp>`
   /* 6px vertical padding + 20px line height = 32px total height
    *
    * TODO: When rem-based spacing on a 4px scale lands, replace
@@ -70,13 +112,12 @@ const StyledItem = styled.div<{variant: ItemProps['variant']} & SxProp>`
   padding: 6px ${get('space.2')};
   display: flex;
   border-radius: ${get('radii.2')};
-  color: ${({variant}) => (variant === 'danger' ? get('colors.text.danger') : 'inherit')};
+  color: ${({variant, item}) => getItemVariant(variant, item?.disabled).color};
 
   @media (hover: hover) and (pointer: fine) {
     :hover {
-      background: ${props =>
-        props.variant === 'danger' ? get('colors.bg.danger') : get('colors.selectMenu.tapHighlight')};
-      cursor: pointer;
+      background: ${({variant, item}) => getItemVariant(variant, item?.disabled).hoverBackground};
+      cursor: ${({variant, item}) => getItemVariant(variant, item?.disabled).hoverCursor};
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/ActionList/Item.tsx
```diff
@@ -59,9 +60,50 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
...
+
+  /**
+   * Items that are disabled can not be clicked, selected, or navigated through.
+   */
+  disabled?: boolean
+
+  /**
+   * Callback that will trigger both on click selection and keyboard selection.
+   */
+  onAction?: (item: Partial<ItemProps>, event?: MouseEvent | KeyboardEvent) => void
```
VanAnderson, line 72: CI is failing right now because it's not liking the event type here. Any thoughts on what `event` needs to be?

src/ActionList/Item.tsx
```diff
@@ -59,9 +60,50 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
...
+      hoverBackground: 'inherit',
+      hoverCursor: 'default'
+    }
+  }
+
+  switch (variant) {
+    case 'danger':
+      return {
+        color: themeGet('colors.text.danger'),
+        iconColor: themeGet('colors.icon.danger'),
```
smockle, line 90: Non-blocking: I wonder if we can/should have `iconColor` follow `color` in all cases, rather than setting these two color variations explicitly.

The use case I have in mind would be someone passing a `color` system-prop to an `Item`, but I’m not sure whether we want to support that. cc @colebemis, since this is somewhat related to https://github.com/github/primer/issues/115

src/ActionList/Item.tsx
```diff
@@ -70,13 +112,12 @@ const StyledItem = styled.div<{variant: ItemProps['variant']} & SxProp>`
...
-  color: ${({variant}) => (variant === 'danger' ? get('colors.text.danger') : 'inherit')};
+  color: ${({variant, item}) => getItemVariant(variant, item?.disabled).color};
 
   @media (hover: hover) and (pointer: fine) {
     :hover {
-      background: ${props =>
-        props.variant === 'danger' ? get('colors.bg.danger') : get('colors.selectMenu.tapHighlight')};
-      cursor: pointer;
+      background: ${({variant, item}) => getItemVariant(variant, item?.disabled).hoverBackground};
+      cursor: ${({variant, item}) => getItemVariant(variant, item?.disabled).hoverCursor};
```
smockle, line 120: Good catch with `cursor`! I was only thinking about the color changes. 🎊

smockle: Related: Should we add [`pointer-events: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) to `disabled` `Item`s?

src/ActionList/Item.tsx
```diff
@@ -125,22 +166,50 @@ const DescriptionContainer = styled.span`
...
+    ...props
+  } = itemProps
   return (
-    <StyledItem tabIndex={-1} variant={variant} aria-selected={selected} {...props}>
+    <StyledItem
+      tabIndex={disabled ? undefined : -1}
+      variant={variant}
+      aria-selected={selected}
+      {...props}
+      onKeyPress={event => {
```
smockle, line 191: Should this anonymous function (and the `onClick` value below) become replaced with a `React.useCallback`?

src/ActionMenu.tsx
```diff
@@ -44,11 +44,27 @@ const ActionMenuBase = ({
         ...itemProps,
         role: 'menuitem',
         onKeyPress: _event => {
-          onAction?.(itemProps as ItemProps)
+          if (itemProps.disabled) {
+            return
+          }
+
+          if (itemProps.onAction) {
```
smockle, line 51: Would this branching logic (and the analogous `onClick`) be more readable if we followed the style of [`renderGroup`](https://github.com/primer/components/blob/a795d6ea386c39be48e3ebe715d2cb61f73435fc/src/ActionList/List.tsx#L129-L130) in `ActionList`?

Something like this:
```
const actionCallback = itemProps.onAction ?? onAction
actionCallback?.(itemProps as ItemProps)
```

src/stories/ActionMenu.stories.tsx
```diff
@@ -96,11 +96,15 @@ export function SimpleListStory(): JSX.Element {
...
-            {text: 'New file', trailingText: '⌘O'},
+            {text: 'New file', trailingText: '⌘O', disabled: true, leadingVisual: ProjectIcon},
             ActionList.Divider,
             {text: 'Copy link', trailingText: 'ctrl+C'},
             {text: 'Edit file', trailingText: '⌘E'},
-            {text: 'Delete file', variant: 'danger', trailingText: '⌘D'}
+            {
+              text: 'Delete file',
+              variant: 'danger',
+              trailingText: '⌘D'
```
smockle, line 106: Love seeing the Storybook updated! Nice work demoing variations! 🚀

src/ActionList/Item.tsx
T-Hugs: Try `React.MouseEvent | React.KeyboardEvent`. These are the react "synthetic" versions of these event objects, which are generated when you listen with one of the `on{action}` event handlers on a JSX element.

Another fix, which might be preferable, is to keep the `MouseEvent | KeyboardEvent`, but when you call `onAction` you pass `event.nativeEvent` instead of just `event`.

src/ActionList/Item.tsx
```diff
@@ -59,9 +60,50 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
...
+        iconColor: themeGet('colors.text.disabled'),
+        annotationColor: themeGet('colors.text.disabled'),
+        hoverBackground: themeGet('colors.selectMenu.tapHighlight'),
+        hoverCursor: 'pointer'
+      }
+  }
 }
 
-const StyledItem = styled.div<{variant: ItemProps['variant']} & SxProp>`
+const StyledItem = styled.div<{variant: ItemProps['variant']; item?: ItemProps} & SxProp>`
```
T-Hugs, line 106: Can we remove `variant` here and just use `item`?

src/ActionList/Item.tsx
T-Hugs: >Related: Should we add pointer-events: none to disabled Items?

I don't think so. A tooltip explaining why something is disabled is a pretty common pattern, and `pointer-events: none` would prevent that from being possible.

src/ActionList/Item.tsx
T-Hugs: I think it's pretty common to have icon colors differ from the text color. Found an example on this page, in fact:
image

src/ActionList/Item.tsx
VanAnderson: I don't think we can here... I'm not 100% sure why, but `item` does not include `variant` here, `variant` is only a sibling key to `item`.

src/ActionList/Item.tsx
VanAnderson: done!

src/ActionList/Item.tsx
smockle: Ah, great point and great example—thanks @T-Hugs!

Given “it’s pretty common to have icon colors differ from the text color,” I wonder if we should find a more ergonomic API than what we have now, as demonstrated in Storybook: https://github.com/primer/components/blob/fb77284291c9a0028bd23970f033f60011a3905f/src/stories/ActionList.stories.tsx#L119-L123

`Item` is a container for several children, including the leading visual (e.g. an `Icon`). I’m generally wary of mirroring child props on the parent (e.g. adding an `iconColor` prop to `Icon` that is passed through to as `color` to `Icon`), because:

- It can be ambiguous which child will receive the prop (e.g. would `iconColor` apply to the leading visual, to the trailing visual, or both?)
- It makes a container/parent’s props contract huge/overwhelming (consider the set of all props in `ItemProps` plus all props in `IconProps`)
- Every scenario mirrored-children-props supports can already be addressed with render-props (as in the example above)

That said, maybe my general wariness should be overridden by ergonomic concerns in this specific instance? I’m open to considering any approach.

src/ActionList/Item.tsx
smockle: Fixed in https://github.com/primer/components/pull/1186/commits/052c8dddc1b21c9e2fb94e5be06b7f0c27ffd0b4

src/ActionMenu.tsx
smockle: Fixed in https://github.com/primer/components/pull/1186/commits/052c8dddc1b21c9e2fb94e5be06b7f0c27ffd0b4

src/ActionList/Item.tsx
T-Hugs: Hm, I'm a bit confused by this since it looks like you're literally extracting the type for `variant` from `ItemProps`. Don't want to block progress, so approving, but I would like to understand this better.

T-Hugs (approved):
One lingering [comment](https://github.com/primer/components/pull/1186/files/17f913b60f71e4bb47133a2a14c4c326753dd378#r623173864) but approving to avoid blocking progress. Thanks!

src/ActionMenu.tsx
```diff
@@ -43,12 +43,22 @@ const ActionMenuBase = ({
...
-          onAction?.(itemProps as ItemProps)
+          if (itemProps.disabled) {
+            return
+          }
+
+          const actionCallback = itemProps.onAction ?? onAction
+          actionCallback?.(itemProps as ItemProps, event)
           onClick?.(event)
           onClose()
         }
```
dgreif, line 64: Instead of adding secondary key/click listeners here, can we just pass in an `onAction` and have it call `itemProps.onAction` if one exists, along with `onAction` from the `ActionMenu` props?

PR #1187, opened 2021-04-28
fix(AnchoredOverlay): use list focus if open on initial render

dgreif:
Currently, if an `AnchoredOverlay` is rendered with `open={true}` on the _initial_ render, the internal `focusType` is `null` and _cannot_ change to `anchor` or `list` no matter how the user interacts with the page.  This change makes the assumption that rendering the overlay initially open indicates the user will be interacting with it immediately, so `focusType` should be set to `list` immediately.

### Screenshots
With open set on initial render:
image

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


T-Hugs (approved):
Brilliant ✅

PR #1188, opened 2021-04-28
Update `restoreFocusOnCleanUp` to accept a JIT function

T-Hugs:
This PR updates the `restoreFocusOnCleanUp` option of `useFocusTrap` to accept an `HTMLElement` or a function that returns an `HTMLElement` to JIT decide the proper element to focus. It also updates the related story.


src/hooks/useFocusTrap.ts
```diff
@@ -22,10 +22,12 @@ interface FocusTrapHookSettings {
...
   /**
-   * If true, when this focus trap is cleaned up, restore focus to the element that had
-   * focus immediately before the focus trap was enabled. (Default: false)
+   * Return focus to the specified element when the focus trap is cleaned up or disabled.
+   * If true, return focus to the element that was focused immediately before the focus
+   * trap was enabled. If a function is passed, it will be called when the focus trap is
+   * cleaned up to determine the element to return focus to. (Default: false)
    */
-  restoreFocusOnCleanUp?: boolean
+  restoreFocusOnCleanUp?: HTMLElement | boolean | (() => HTMLElement | boolean)
```
dgreif, line 30: The new `useAwaitableState` seems to be all that is needed to fix the issue that brought on this PR.  @colebemis @smockle, do you think we should keep the changes for restoring focus as well, or revert back to the simpler setup and only use `useAwaitableState` to solve the immediate issue.

dgreif (changes requested):

T-Hugs:
I have confirmed the `useAwaitableState` fix resolves the issue at hand in Memex 🚀

PR #1191, opened 2021-04-29
Pass onAction directly to Item and check for selection keys

VanAnderson:
Quick follow up I wanted to put in to address some comments @dgreif made in https://github.com/primer/components/pull/1186#discussion_r623215001 and https://github.com/primer/components/pull/1186#pullrequestreview-648382448

This cleans up the code a bit by delegating more click/keypress handling to Item, and it also makes sure we are checking for selection keys (space/enter) as a condition to firing onAction.

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -202,7 +202,9 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
         return
       }
       onClick?.(event)
-      onAction?.(itemProps as ItemProps, event)
+      if (!event.defaultPrevented) {
+        onAction?.(itemProps as ItemProps, event)
+      }
     },
     [onAction, disabled, itemProps, onClick]
   )
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -188,10 +188,10 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
       if (disabled) {
         return
       }
-      if (onAction) {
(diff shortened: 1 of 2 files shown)


src/ActionList/Item.tsx
```diff
@@ -201,10 +201,8 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
       if (disabled) {
         return
       }
-      if (onAction) {
-        onAction(itemProps as ItemProps, event)
-      }
       onClick?.(event)
+      onAction?.(itemProps as ItemProps, event)
```
dgreif, line 205: I like the re-ordering here.  We may want to check for event.defaultPrevented on the click as well, just to give the consumer an easy way to override the behavior

src/ActionMenu.tsx
```diff
@@ -39,30 +39,21 @@ const ActionMenuBase = ({
...
-    ({onClick, ...itemProps}) =>
-      renderItem({
+    ({onAction: itemOnAction, ...itemProps}) => {
+      const onActionWithClose = (
+        props: ItemProps,
+        event?: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>
+      ) => {
+        const actionCallback = itemOnAction ?? onAction
+        actionCallback?.(props as ItemProps, event)
+        onClose()
```
dgreif, line 49: So much cleaner!  This will need to be rebased in #1188, but it should be pretty easy on @T-Hugs side.

dgreif (changes requested):

dgreif (approved):
Super clean, thank you! ✨

PR #1194, opened 2021-04-29
Handle onAction for DropdownMenu Items

dgreif:
Refactoring `DropdownMenu` to use the new `onAction` callback from `ActionList.Item`, rather than handling click/key press manually.  Also allowing `item.onAction` to call `event.preventDefault()`, which will now prevent `DropdownMenu` from closing and triggering `onChange`

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


VanAnderson (approved):
Looks great! Basically parallel to the other work and I think the places where you just inlined the callback make the code cleaner that what I had 👍

PR #1196, opened 2021-04-29
Allow custom `children` in `ActionItem`

dgreif:
Some use cases for `ActionList` have revealed that the built-in rendering is not always providing enough flexibility, such as if you want to render the text inside a `<Label>`.  Although we can override rendering with `renderItem`, this fully replaces the _entire_ item and removes helpful tools like `onAction` and the `selected` checkmark. Additionally, the requirement of `text` on `ItemProps` forces consumers to convert existing data structures into a secondary data structure in order to provide `text`.  This change makes all fields optional (which opens it up for use by any item interface), and adds a `children` prop that will render a custom `ReactNode` before the `text` section of the item.  

### Screenshots
image

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -21,11 +21,6 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
   description?: string
 
   /**
-   * Custom children content to show if text/description is not flexible enough.  If supplied, leading/trailing elements can still be used.
-   */
-  children?: React.ReactNode
-
-  /**
    * Secondary text style variations. Usage is discretionary.
    *
    * - `"inline"` - Secondary text is positioned beside primary text.
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -13,12 +13,17 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
   /**
    * Primary text which names an `Item`.
    */
-  text: string
+  text?: string
 
   /**
    * Secondary text which provides additional information about an `Item`.
    */
   description?: string
+
+  /**
+   * Custom children content to show if text/description is not flexible enough.  If supplied, leading/traling elements can still be used.
+   */
+  children?: React.ReactNode
 
   /**
    * Secondary text style variations. Usage is discretionary.
@@ -179,6 +184,7 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
     disabled,
     onAction,
     onKeyPress,
+    children,
     onClick,
     ...props
   } = itemProps
@@ -224,10 +230,13 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
           <LeadingVisual />
         </LeadingVisualContainer>
       )}
-      <StyledTextContainer descriptionVariant={descriptionVariant}>
-        <div>{text}</div>
-        {description && <DescriptionContainer>{description}</DescriptionContainer>}
-      </StyledTextContainer>
+      {children}
+      {(text || description) && (
+        <StyledTextContainer descriptionVariant={descriptionVariant}>
+          {text && <div>{text}</div>}
+          {description && <DescriptionContainer>{description}</DescriptionContainer>}
+        </StyledTextContainer>
+      )}
       {(TrailingIcon || trailingText) && (
         <TrailingVisualContainer variant={variant} disabled={disabled}>
           {trailingText && <div>{trailingText}</div>}
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -21,7 +21,7 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
   description?: string
 
   /**
-   * Custom children content to show if text/description is not flexible enough.  If supplied, leading/traling elements can still be used.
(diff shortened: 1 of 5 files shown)


VanAnderson (approved):
Looks pretty straight forward - non blocking but maybe we could update the docs to at least list the new option to use children?

dgreif:
@VanAnderson I couldn't find any docs that explicitly define `ItemProps`.  Happy to update them if I just overlooked them.  For now, I think it's just examples within the `ActionList` docs.  This is an area that we will need to revisit when we have solidified the API I think.

src/ActionList/Item.tsx
```diff
@@ -13,13 +13,18 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
...
-  text: string
+  text?: string
 
   /**
    * Secondary text which provides additional information about an `Item`.
    */
   description?: string
 
+  /**
+   * Custom children content to show if text/description is not flexible enough.  If supplied, leading/traling elements can still be used.
```
smockle, line 24:
```suggestion
   * Custom children content to show if text/description is not flexible enough.  If supplied, leading/trailing elements can still be used.
```

src/ActionList/Item.tsx
```diff
@@ -13,13 +13,18 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
...
+   */
+  children?: React.ReactNode
```
smockle, line 26: Is `children` implicitly part of every component’s props contract already?

src/ActionList/Item.tsx
```diff
@@ -224,10 +230,13 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
           <LeadingVisual />
         </LeadingVisualContainer>
       )}
-      <StyledTextContainer descriptionVariant={descriptionVariant}>
-        <div>{text}</div>
-        {description && <DescriptionContainer>{description}</DescriptionContainer>}
-      </StyledTextContainer>
+      {children}
```
smockle, line 233: If we do retain `text` (https://github.com/primer/components/pull/1196/files#r624194497), should it be mutually-exclusive with `children`?

src/stories/ActionList.stories.tsx
```diff
@@ -238,3 +240,27 @@ export function HeaderStory(): JSX.Element {
...
+              ),
+              trailingIcon: ArrowLeftIcon
+            }
+          ]}
+        />
+      </ErsatzOverlay>
+    </>
+  )
+}
+SimpleListStory.storyName = 'Simple List'
```
smockle, line 266:
```suggestion
CustomItemChildren.storyName = 'Custom Item Children'
```

src/stories/ActionMenu.stories.tsx
```diff
@@ -48,7 +48,7 @@ const ErsatzOverlay = styled.div`
 export function ActionsStory(): JSX.Element {
   const [option, setOption] = useState('Select an option')
   const onAction = (itemProps: ItemProps) => {
-    setOption(itemProps.text)
+    setOption(itemProps.text || '')
```
smockle, line 51: If no `Item`s have `text` (i.e. if each `Item` receives custom `children`), would the selected option be distinguishable, or would `option` be `""` for any selected `Item`?

src/ActionList/Item.tsx
```diff
@@ -13,13 +13,18 @@ export interface ItemProps extends React.ComponentPropsWithoutRef<'div'>, SxProp
   /**
    * Primary text which names an `Item`.
    */
-  text: string
+  text?: string
```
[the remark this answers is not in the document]
dgreif: I think it makes sense to keep `text` because of the complex rendering possible if you include `text` and `description` on the same item.  Seems like a case that could be fairly common, and it's already supported out of the box.

src/stories/ActionMenu.stories.tsx
dgreif: In this particular example with ActionMenu, `text` is being used to determine which option was clicked.  In real-world use, it's more likely that that the options would specify either an `onAction` or a `key` of some sort that would make them identifiable if they used `children` to provide the content.

src/ActionList/Item.tsx
dgreif: Indeed it is!  Removed the explicit prop ✅

smockle (approved):
LGTM! 😎 Nice work on this!

src/ActionList/Item.tsx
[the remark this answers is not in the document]
T-Hugs: I've always had the opinion that rendering should be kept separate from the text _data_. You need the raw text for things like keyboard navigation and filtering, but the rendering can be very complex and custom. Shouldn't text be required? Otherwise how can you u implement the "typeahead" pattern?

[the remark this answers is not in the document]
T-Hugs: Implement a default renderer that throws the provided text in a div or whatever, and use that if they don't provide children.

src/stories/ActionMenu.stories.tsx
T-Hugs, line 51:
```suggestion
    setOption(itemProps.text ?? '')
```

dgreif: Thanks @T-Hugs! Go enjoy the mountains 😉 ⛰️

PR #1197, opened 2021-04-30
Publish TypeScript declaration files to lib-esm

dmarcey:
Currently we only publish our `.d.ts` declaration files to the `lib` directory during a `yarn dist`.

This means that consumers will get TS errors when trying to import from a `lib-esm` path like:
`import {ConfirmationDialog} from '@primer/components/lib-esm/Dialog/ConfirmationDialog'`

This PR adds an extra `yarn` script for running `tsc` to output the declaration files in the `lib-esm` directory, in addition to the `lib` directory.


package.json
```diff
@@ -12,7 +12,8 @@
     "dist:rollup": "cross-env NODE_ENV=production rollup -c",
     "dist:transpile:cjs": "cross-env BABEL_MODULE=commonjs babel src --out-dir lib --extensions '.js,.ts,.jsx,.tsx'",
     "dist:transpile:esm": "babel src --out-dir lib-esm --extensions '.js,.ts,.jsx,.tsx'",
-    "dist:types": "tsc",
+    "dist:types:cjs": "tsc",
+    "dist:types:esm": "tsc --outDir lib-esm",
```
T-Hugs, line 16: Instead of re-running `tsc` can't we just copy the `.d.ts` file(s)?

dmarcey: Also cross-platform copy possibly requires a new `npm` package? 😬

colebemis (approved):

VanAnderson:
Looks good to me 👍

PR #1199, opened 2021-04-30
Action Menu can have its open state be controlled externally

VanAnderson:
Just a small API change to optionally accept `open` and `setOpen` - if these props are present then the open/closed state of ActionMenu will be externally controlled.

Fixes https://github.com/github/primer/issues/135


src/stories/ActionMenu.stories.tsx
```diff
@@ -113,6 +113,41 @@ export function SimpleListStory(): JSX.Element {
...
+  const [option, setOption] = useState('Select an option')
+  const [open, setOpen] = useState(false)
+  const onAction = (itemProps: ItemProps) => {
+    setOption(itemProps.text)
+  }
+  return (
+    <>
+      <h1>Simple List</h1>
+      <h2>Last option activated: {option}</h2>
+      <h2>External Open State: {open ? 'Open' : 'Closed'}</h2>
```
dgreif, line 126: Might be nice to add an "Open" button that will set the state from the outside, just to test that thoroughly

src/ActionMenu.tsx
```diff
@@ -21,11 +23,26 @@ const ActionMenuBase = ({
...
+      setOpen(true)
+    }
+  }, [setOpen])
+
+  const onClose = useCallback(() => {
+    setInternalOpen(false)
+    if (setOpen) {
+      setOpen(false)
+    }
+  }, [setOpen])
```
dgreif, line 45: This seems like a pattern that we will want to replicate into DropdownMenu and SelectPanel.  What are your thoughts on pulling out a hook that could be used like `[open, setOpen] = useOptionallyExternalState(externalOpen, setExternalOpen, false)` (definitely open to different naming).  Then we would only have one state and not maintain both internal and external.

dgreif (changes requested):
Nice pattern, I think we will want more of these 😄

src/ActionMenu.tsx
VanAnderson: @dgreif yeah, I think that would clean things up quite a bit! To be clear, I think we would still want to be maintaining the internal state within the `useOptionallyExternalState` hook since we could not conditionally implement hooks - we just wouldn't expose that conditional internal state outside of the hook. Is that basically what you are suggesting?

dgreif: @VanAnderson something similar to https://github.com/primer/components/blob/main/src/hooks/useProvidedRefOrCreate.ts.  It would create the internal state, but not use it if an external state was provided.  `useProvidedStateOrCreate` may be more in-line now that I see this other name 😄

src/hooks/useProvidedStateOrCreate.ts
```diff
@@ -0,0 +1,26 @@
+import {useCallback, useState} from 'react'
+
+/**
+ * There are some situations where we only want to create a new ref if one is not provided to a component
+ * or hook as a prop. However, due to the `rules-of-hooks`, we cannot conditionally make a call to `React.useRef`
+ * only in the situations where the ref is not provided as a prop.
+ * This hook aims to encapsulate that logic, so the consumer doesn't need to be concerned with violating `rules-of-hooks`.
+ * @param providedRef The ref to use - if undefined, will use the ref from a call to React.useRef
+ * @type TRef The type of the RefObject which should be created.
+ */
```
VanAnderson, line 10: whoops, copy pasta - I'll create a proper doc comment...

src/hooks/useProvidedStateOrCreate.ts
```diff
@@ -0,0 +1,27 @@
+import {useCallback, useState} from 'react'
+
+/**
+ * There are some situations where we want to give users the option to control state externally with their own state handlers
+ * or default to using internal state handlers. Because of the 'rules-of-hooks', we cannot conditionally make a call to `React.useState`
+ * only in the situations where the state is not provided as a prop.
+ * This hook aims to encapsulate that logic, so the consumer doesn't need to be concerned with violating `rules-of-hooks`.
+ * @param externalState The state to use - if undefined, will use the state from a call to React.useState
+ * @param setExternalState The setState to use - if undefined, will use the state from a call to React.useState
```
dgreif, line 9:
```suggestion
 * @param setExternalState The setState to use - if undefined, will use the setState from a call to React.useState
```

src/ActionMenu.tsx
```diff
@@ -64,7 +69,7 @@ const ActionMenuBase = ({
   }, [open])
 
   return (
-    <AnchoredOverlay renderAnchor={renderMenuAnchor} open={open} onOpen={onOpen} onClose={onClose}>
+    <AnchoredOverlay renderAnchor={renderMenuAnchor} open={!!combinedOpenState} onOpen={onOpen} onClose={onClose}>
```
dgreif, line 72: Is the `!!` necessary here?  The types for `useProvidedStateOrCreate` look like they should always have `boolean` for `combinedOpenState`

VanAnderson: @dgreif I believe that `combinedOpenState` can also be `undefined` as a result of the generics used in the hooks. Maybe there was a better way to implement that?

dgreif (approved):

PR #1200, opened 2021-04-30
Perform ActionMenu actions after overlay has closed

dgreif:
This is an alternate approach to solve the problems addressed by #1188.  When an action is selected from the menu, we want to _wait_ until the menu has closed the overlay before performing the action.  This is important in situations where the action moves focus to a different element on the page.  Without this change, the focus change from the action would be detected by the overlay focus trap, preventing focus from actually shifting to the intended target.  While #1188 did present two valid ways of solving this problem, @T-Hugs and @smockle agreed that the approach in this PR is a simpler solution for the immediate issue. 

Closes #1188 

### Screenshots
CleanShot 2021-04-30 at 11 48 09

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionMenu.tsx
+++ b/src/ActionMenu.tsx
@@ -2,7 +2,7 @@ import {GroupedListProps, List, ListPropsBase} from './ActionList/List'
 import {Item, ItemProps} from './ActionList/Item'
 import {Divider} from './ActionList/Divider'
 import Button, {ButtonProps} from './Button'
-import React, {useCallback, useState} from 'react'
+import React, {useCallback, useEffect, useRef, useState} from 'react'
 import {AnchoredOverlay} from './AnchoredOverlay'
 
 export interface ActionMenuProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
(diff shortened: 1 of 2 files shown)


src/ActionMenu.tsx
```diff
@@ -45,14 +46,23 @@ const ActionMenuBase = ({
...
   )
 
+  useEffect(() => {
+    // Wait until menu has re-rendered in a closed state before triggering action.
+    // This is needed in scenarios where the action will move focus, which would otherwise be captured by focus trap
+    if (!open && pendingActionRef.current) {
+      pendingActionRef.current()
+      pendingActionRef.current = undefined
+    }
+  }, [open])
```
smockle, line 64: To clarify: `open` is sufficient as a dependency? (i.e. we don’t need both `open` and `pendingActionRef.current`)?

smockle (approved):
Thanks for updating this! 🎊

src/ActionMenu.tsx
dgreif: `open` should be enough here because we only want to run this effect when `open` goes from `true` to `false`, and  `pendingActionRef.current` won't ever trigger a re-render when set

VanAnderson (approved):
looks good, will be glad to have this in 👍

PR #1201, opened 2021-04-30
Fix overlay position

T-Hugs:
Fixes overlay position. Since we were passing a number to the "top" and "left" props, they were getting converted to "space.x" where x is one of the primitives (so 4 would become 24px).

Closes #[internal issue 138]

### Merge checklist
- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/AnchoredOverlay/AnchoredOverlay.tsx
+++ b/src/AnchoredOverlay/AnchoredOverlay.tsx
@@ -90,6 +90,7 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({renderAnchor, c
     },
     [overlayRef.current]
   )
+  const overlayPosition = position && {top: `${position.top}px`, left: `${position.left}px`}
 
   useFocusZone({containerRef: overlayRef, disabled: !open || focusType !== 'list' || !position})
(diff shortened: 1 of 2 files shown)


dgreif (approved):
Nice! I have been wondering why these were thrown off over the last week.  Bad refactor on my part I guess 😅 .  Really appreciate you tracking this down so quickly

src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -90,6 +90,7 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({renderAnchor, c
     },
     [overlayRef.current]
   )
+  const overlayPosition = position && {top: `${position.top}px`, left: `${position.left}px`}
```
dmarcey, line 93: Worth a `React.useMemo`? Looks like most of the other props to `Overlay` are memoized.

dgreif: `useMemo` pushed @dmarcey.  LGTY?

PR #1209, opened 2021-05-06
AnchoredOverlay and ActionList fixes for SelectPanel

dgreif:
These changes address a number of issues/missing features that were found during development of the `SelectPanel` component (currently being built in a separate project)

* `Overlay` height/width should be able to be set _through_ the `AnchorOverlay` component. This allows you to have a fixed height/width on the overlay if list items are loaded asynchronously after the overlay opens. 
* It is common for `Item` data to include an `id` field, which could be a `string` or `number`.  We should allow these to be passed in on `ItemProps` and use them as the `key` prop when present.  We also _don't_ want to use them on the actual `Item` element in the DOM as they will likely not be _globally_ unique
* `ActionList` will create a single group wrapper if it is passed a list of items without any group details.  This group was previously receiving a `uniqueId` for its `groupId` on _every_ render.  This caused react to remove the existing DOM elements and create all new ones for the entire list contents.  Not ideal for perf or for focus management.
* `AnchoredOverlay` was delaying the focus zone from activating until user interaction triggered `list` mode.  This is a problem if the user changes focus by _clicking_ into the list.  There is no reason not to activate the focus zone as soon as the overlay is opened.  Focus _trap_ should still be deferred, but there are remaining issues around when that activates with user clicks (#1210) 

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/List.tsx
+++ b/src/ActionList/List.tsx
@@ -137,14 +137,8 @@ export function List(props: ListProps): JSX.Element {
    */
   const renderItem = (itemProps: ItemInput, item: ItemInput) => {
     const ItemComponent = ('renderItem' in itemProps && itemProps.renderItem) || props.renderItem || Item
-    return (
-      <ItemComponent
-        {...itemProps}
-        key={itemProps.key || uniqueId()}
-        sx={{...itemStyle, ...itemProps.sx}}
-        item={item}
-      />
-    )
+    const key = itemProps.key ?? itemProps.id?.toString() ?? uniqueId()
+    return <ItemComponent {...itemProps} key={key} sx={{...itemStyle, ...itemProps.sx}} item={item} />
   }
 
   /**
--- a/src/ActionList/List.tsx
+++ b/src/ActionList/List.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useMemo} from 'react'
 import type {AriaRole} from '../utils/types'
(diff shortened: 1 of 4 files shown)


src/ActionList/List.tsx
```diff
@@ -153,10 +147,11 @@ export function List(props: ListProps): JSX.Element {
   let groups: (GroupProps | (Partial<GroupProps> & {renderItem?: typeof Item; renderGroup?: typeof Group}))[] = []
 
   // Collect rendered `Item`s into `Group`s, avoiding excess iteration over the lists of `items` and `groupMetadata`:
+  const singleGroupId = useMemo(uniqueId, [])
```
smockle, line 150: To clarify, this is here to ensure the same number of hooks is called on each render, i.e. to avoid conditionally calling `useMemo` as Group count changes?

smockle (approved):
🚀

src/ActionList/List.tsx
dgreif: Yep, that's exactly why.  Hooks shouldn't be called inside loops or conditionals to ensure consistent execution order across renders.

VanAnderson (approved):
👍  Looks good to me!

PR #1215, opened 2021-05-07
ActionList Fixes

dgreif:
Addressing a number of issues found when integrating `ActionList` into `SelectPanel`

- [x] Divider should be hidden above a group that has a filled header
- [x] `inline` description is not inline
- [x] Multi-select action lists should use checkboxes
  - [x] Checkbox instead of checkmark icon
  - [x] Properly align checkbox vertically
  - [x] ~Wrap item in a `<label>`?~ (Using `aria-label={text}` to avoid branching DOM and for succinct announcements)
  - [x] Prevent tabbing to checkbox if in focusZone (this one might be tough)
  - [x] ~Update docs with new `selectionVariant` option~ (Prop is commented in the `ItemProps` interface definition, but `ActionList.Item` does not have a corresponding `.mdx` file) 

### Screenshots
| Before | After |
|--------|--------|
| image | image |
| image | image |  
| image | image |

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/blue-items-repeat.md
+++ b/.changeset/blue-items-repeat.md
@@ -3,6 +3,3 @@
 ---
 
 Hide divider before `ActionList.Group`s with filled header
-Align `Item` description to when rendered in-line
-Add `selectionVariant: 'multiple'` for `Item`s.  These will use a checkbox input instead of a checkmark icon for selected state
-Allow `focusZoneSettings` to be passed into `AnchoredOverlay`
--- a/.changeset/blue-items-repeat.md
+++ b/.changeset/blue-items-repeat.md
@@ -1,0 +1,8 @@
+---
+"@primer/components": patch
+---
+
+Hide divider before `ActionList.Group`s with filled header
+Align `Item` description to when rendered in-line
+Add `selectionVariant: 'multiple'` for `Item`s.  These will use a checkbox input instead of a checkmark icon for selected state
+Allow `focusZoneSettings` to be passed into `AnchoredOverlay`
--- a/src/ActionList/List.tsx
+++ b/src/ActionList/List.tsx
@@ -45,6 +45,11 @@ export interface ListPropsBase {
    * - `"full"` - `List` children are flush (vertically and horizontally) with `List` edges
    */
   variant?: 'inset' | 'full'
+
+  /**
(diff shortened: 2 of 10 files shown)


VanAnderson (approved):
LGTM!

.changeset/blue-items-repeat.md
```diff
@@ -0,0 +1,8 @@
+---
+"@primer/components": patch
+---
+
+Hide divider before `ActionList.Group`s with filled header
+Align `Item` description to when rendered in-line
+Add `selectionVariant: 'multiple'` for `Item`s.  These will use a checkbox input instead of a checkmark icon for selected state
+Allow `focusZoneSettings` to be passed into `AnchoredOverlay`
```
colebemis, line 8: If it's not too much trouble, it might be nice to put all these distinct changes in their own changeset.

src/ActionList/List.tsx
```diff
@@ -82,6 +87,12 @@ export type ListProps = ListPropsBase | GroupedListProps
 
 const StyledList = styled.div`
   font-size: ${get('fontSizes.1')};
+  /* 14px font-size * 1.428571429 = 20px line height
+   *
+   * TODO: When rem-based spacing on a 4px scale lands, replace
+   * hardcoded '1.428571429'
+   */
+  line-height: 1.428571429;
```
colebemis, line 95: Could we just hard code 20px here instead?
```suggestion
  line-height: 20px;
```

colebemis (approved):
Looks great, @dgreif! Thanks for cleaning this up ❤️ 

Just left a few non-blocking comments.

src/ActionList/List.tsx
dgreif: @smockle thoughts on this?

smockle: I’m fine either way: The hardcoded value is going to be replaced with a variable after the [“new scaling and typography internal variables”](https://github.com/github/primer/issues/41#issuecomment-839264461) land. cc @vdepizzol for visibility

dgreif: Thanks @smockle.  I made the switch to `20px` since it's a little easier to read for now.

PR #1217, opened 2021-05-10
onMouseDown passthrough in ActionMenu and AnchoredOverlay

VanAnderson:
This PR provides a passthrough for onMouseDown events in `ActionMenu` to `AnchoredOverlay`

Also fills out some doc comments I realized were missing in `ActionMenu`

### DIFF
--- a/src/AnchoredOverlay/AnchoredOverlay.tsx
+++ b/src/AnchoredOverlay/AnchoredOverlay.tsx
@@ -5,10 +5,6 @@ import {useFocusZone} from '../hooks/useFocusZone'
 import {useAnchoredPosition, useRenderForcingRef} from '../hooks'
 import {uniqueId} from '../utils/uniqueId'
 
-function preventDefault(event: React.UIEvent) {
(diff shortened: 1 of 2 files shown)


src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -30,6 +26,11 @@ export interface AnchoredOverlayProps extends Pick<OverlayProps, 'height' | 'wid
    * A callback which is called whenever the overlay is currently open and a "close gesture" is detected.
    */
   onClose?: (gesture: 'click-outside' | 'escape') => unknown
+
+  /**
+   * A callback that occurs when a user presses a mouse button over an element.
+   */
+  onMouseDown?: (event: React.MouseEvent) => unknown
```
dgreif, line 33: I would match the `ActionMenu` and use `onOverlayMouseDown` here since we still technically have a split between anchor and `Overlay` in this component

dgreif (changes requested):

PR #1221, opened 2021-05-12
Improved performance for lists in layered components

dgreif:
Pulling in some performance tweaks that I found useful when working on `SelectPanel`

* Map `items` to new `ItemProps` objects, rather than overwriting `renderItem`.  This creates less functions, and prevents new DOM elements from being created if the list is re-rendered while open
* `useMemo` for the item conversion, which prevents the conversioning happening before the menu opens _and_ after it opens.  Now it just converts items once.

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionMenu.tsx
+++ b/src/ActionMenu.tsx
@@ -2,7 +2,7 @@ import {GroupedListProps, List, ListPropsBase} from './ActionList/List'
 import {Item, ItemProps} from './ActionList/Item'
 import {Divider} from './ActionList/Divider'
 import Button, {ButtonProps} from './Button'
-import React, {useCallback, useEffect, useRef} from 'react'
+import React, {useCallback, useEffect, useMemo, useRef} from 'react'
 import {AnchoredOverlay} from './AnchoredOverlay'
 import {useProvidedStateOrCreate} from './hooks/useProvidedStateOrCreate'
 import {OverlayProps} from './Overlay'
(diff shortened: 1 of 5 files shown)


src/ActionMenu.tsx
```diff
@@ -71,23 +71,26 @@ const ActionMenuBase = ({
     [anchorContent, renderAnchor]
   )
 
-  const renderMenuItem: typeof Item = useCallback(
-    ({onAction: itemOnAction, ...itemProps}) => {
-      return renderItem({
-        ...itemProps,
+  const itemsToRender = useMemo(() => {
+    return items.map(item => {
+      if ('renderItem' in item) {
```
smockle, line 76: Why are `ItemProps` with `renderItem` a special case? (i.e. Why is this early return here?) The `renderItem` function is called with other `ItemProps` as an argument, and it seems like we’d want to include e.g. `role: 'menuitem'` (merged in on line 82) to the `ItemProps` it receives.

dgreif: This is intended as a short-circuit for `ActionList.Divider` and any other `ItemInput` values that are not actually `ItemProps`.  Happy to change to another approach if you have a suggestion. As far as real `ItemProps` that supply a `renderItem`, I figured those already take full responsibility for how they are rendered, so should do the same with how actions are handled.

dgreif: After some discussion, we decided it is best to remove these short-circuits for now.  The `ActionList.Divider` does not actually spread any additional props, so there is no harm in passing them.  The danger in _not_ passing them is if we have a custom _item_ that expects things like `role: menuitem` to be passed in automatically.

VanAnderson (approved):
👍

PR #1222, opened 2021-05-12
Trap focus in AnchoredOverlay as soon as it opens

### LINKED ISSUE #1210: Activate focus trap when clicking inside AnchoredOverlay
**Describe the bug**
`AnchoredOverlay` currently only activates the internal focus trap if the user interacts with the anchor element in specific ways.  New uses of the `AnchoredOverlay` include behaviors where a user can click inside the overlay multiple times, which removes focus from the anchor _without_ activating the focus trap.  In these scenarios, the user is able to tab to other parts of the DOM without restriction.  We should handle this scenario and activate the focus trap if the anchor loses focus in some unexpected way.

**To Reproduce**
1. Go to https://primer.style/components/DropdownMenu
2. Open the menu in the example
3. Click on the _edge_ of the menu (small margin on left or right side, but not in an item)
4. Note that focus has left the anchor button.  You can now press `Tab` to focus on other parts of the page

**Expected behavior**
If focus leaves the anchor without the overlay closing, focus trap should activate immediately

dgreif:
In the current implementation of `AnchoredOverlay`, the behavior for when to activate focus trapping is fairly complicated.  If the user clicks or presses space/Enter to open the overlay, focus stays on the anchor.  They can then use up/down arrow, enter, or tab to move focus _into_ the overlay.  If, however, they press up/down on the anchor initially, focus immediately moves into the overlay.  In initial development, this felt like a good approach.  However, real use has uncovered some corner cases which indicate the complex initial state might not be worth it.  In particular, some new `ActionMenu`s that also include a text input, as well as the `SelectPanel`, need focus to be trapped as soon as the overlay is opened.  The easiest solution is to simply _always_ move focus into the overlay, rather than conditionally starting focus on the anchor. 

Closes #1210

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/silver-scissors-love.md
+++ b/.changeset/silver-scissors-love.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Trap focus in `AnchoredOverlay` as soon as it opens, regardless of the event that triggered it to open
--- a/src/AnchoredOverlay/AnchoredOverlay.tsx
+++ b/src/AnchoredOverlay/AnchoredOverlay.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react'
+import React, {useCallback, useMemo, useRef} from 'react'
 import Overlay, {OverlayProps} from '../Overlay'
 import {useFocusTrap} from '../hooks/useFocusTrap'
 import {useFocusZone} from '../hooks/useFocusZone'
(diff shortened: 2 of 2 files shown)


VanAnderson (approved):
👍

colebemis (approved):
🚀

PR #1223, opened 2021-05-13
feat: ActionList.Item dividers

smockle:
This PR adds a `showDivider` prop to `ActionList.Item` and corresponding `showItemDividers` props to `ActionList.Group` and `ActionList` to set `showDivider` for each `ActionList.Item` in an `ActionList.Group` or `ActionList`, respectively. 

Dividers are not displayed when an `ActionList.Item` immediately follows an `ActionList.Header` or `ActionList.Divider`, to avoid excessive horizontal rules.

### Screenshots
| `ActionList.Item`-level | `ActionList.Group`-level | `ActionList`-level |
|:---:|:---:|:---:|
| Screenshot showing the effect of passing the 'showDivider' prop to 'ActionList.Item's | Screenshot showing the effect of passing the 'showItemDividers' prop to 'ActionList.Group' | Screenshot showing the effect of passing the 'showItemDividers' prop to 'ActionList' |

### Merge checklist
- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -54,6 +54,11 @@ export interface ItemProps extends Omit<React.ComponentPropsWithoutRef<'div'>, '
   variant?: 'default' | 'danger'
 
   /**
+   * Whether to display a divider above the `Item` when it does not follow a `Header` or `Divider`.
+   */
+  showDivider?: boolean
+
+  /**
    * For `Item`s which can be selected, whether the `Item` is currently selected.
    */
   selected?: boolean
@@ -119,7 +124,9 @@ const StyledItemContent = styled.div`
   width: 100%;
 `
 
-const StyledItem = styled.div<{variant: ItemProps['variant']; item?: ItemInput} & SxProp>`
+const StyledItem = styled.div<
+  {variant: ItemProps['variant']; showDivider: ItemProps['showDivider']; item?: ItemInput} & SxProp
+>`
   /* 6px vertical padding + 20px line height = 32px total height
    *
    * TODO: When rem-based spacing on a 4px scale lands, replace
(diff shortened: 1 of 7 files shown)


src/ActionList/Item.tsx
```diff
@@ -122,6 +136,7 @@ const StyledItem = styled.div<{variant: ItemProps['variant']; item?: ItemInput}
   display: flex;
   border-radius: ${get('radii.2')};
   color: ${({variant, item}) => getItemVariant(variant, item?.disabled).color};
+  position: relative;
```
dgreif, line 139: I don't think this one is required since the `:before` moves itself `relative`.

src/ActionList/Item.tsx
```diff
@@ -130,6 +145,21 @@ const StyledItem = styled.div<{variant: ItemProps['variant']; item?: ItemInput}
...
+    margin-top: ${({showDivider}) => (showDivider ? `1px` : '0')};
+
+    ${StyledItemContent}::before {
+      content: ' ';
+      display: block;
+      position: relative;
+      top: -7px;
+      // NB: This 'get' won’t execute if it’s moved into the arrow function below.
+      border: 0 solid ${get('colors.selectMenu.borderSecondary')};
+      border-top-width: ${({showDivider}) => (showDivider ? `1px` : '0')};
```
dgreif, line 159: Can you elaborate on this?  I'm not familiar with the `NB:` acronym 😄

dgreif (changes requested):
I love the simplicity. Can't wait to get this added to SelectPanel!

src/ActionList/Item.tsx
smockle: To elaborate on “NB”—it’s short for [_“nota bene”_](https://en.wikipedia.org/wiki/Nota_bene), which is Latin for “note well.” The phrase is used to indicate extra attention or care should be paid to a remark.

To elaborate on the comment—I don’t know why interpolating the call to `get`, as in—
```js
border: ${({showDivider}) => showDivider ? `1px solid ${get('colors.selectMenu.borderSecondary')}` : '0')};
```
—doesn’t work, but it doesn’t: If used, no `border` style is applied to the pseudoelement (other styles—`content`, `display`, `position`, `top`—(without interpolated calls to `get`) are applied)

src/ActionList/Item.tsx
smockle: Yup, good catch, thanks! Removed in https://github.com/primer/components/pull/1223/commits/2940816381af41209ecca7fe5ab2ca8d1c920807

dgreif (approved):

Issue #1071, opened 2021-02-23
SelectPanel


smockle:
Requires https://github.com/primer/components/issues/1070

## Description
SelectPanels are used to select one or more values from a list presented in a Popover. Unlike [DropdownMenu](https://github.com/primer/components/issues/1070), SelectPanels can be used for infinite-scrolling lists and for lists containing arbitrarily-many user-provided list items.

SelectPanels must be positioned **beside** their Source.

**SelectPanels always contain a search field**. If a search field is not required, DropdownMenu may be a more-suitable component.

## Resources
Figma:
Prototype:

## Screenshots
![selectpanel-single-simple](https://user-images.githubusercontent.com/3104489/108860069-dc486c00-75bb-11eb-9f6f-603cd47c9f2a.png)
![selectpanel-single-complex](https://user-images.githubusercontent.com/3104489/108860080-de122f80-75bb-11eb-809b-821fab3adef5.png)
![selectpanel-multi](https://user-images.githubusercontent.com/3104489/108860088-dfdbf300-75bb-11eb-8d1c-0732ad5a1ebb.png)

#### Behaviors

A SelectPanel component renders an ActionList (`role="listbox"`) containing DropdownMenuListItems (`role="option"`) into a closable Popover, below a filter input (`role="combobox"`).

note: A set of attributes, e.g. `aria-expanded`, `aria-owns`, `aria-haspopup`, `aria-autocomplete`, `aria-controls`, `aria-activedescendant` should be applied (by the component internally, not by library consumers) to the appropriate components.

PR #1224, opened 2021-05-13
Add SelectPanel alpha component

dgreif:
This is the initial pass at `SelectPanel` (#1071).  There are still a number of enhancements that will need to be made, to be implemented in subsequent PRs.  Basic scaffolding for docs and tests have been added since this component is `Alpha`.

Also including the `FilteredActionList` component.  We _could_ incorporate this component directly into the `SelectPanel`, but I think it's a little easier to separate responsibilities by pulling it out.  If we intend to use it elsewhere without `SelectPanel`, we will need to add additional docs and tests.

I'm definitely open to discussing the API and implementation.  The current API makes this component _ultra_ controlled, which means flexibility, but also a _lot_ of setup on the consumer side to use it.  I think this is evident if you look at the story and tests.  There are ~10 props required to use it properly.

### Screenshots
image

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/SelectPanel.mdx
+++ b/docs/content/SelectPanel.mdx
@@ -1,0 +1,10 @@
+---
+title: SelectPanel
+status: Alpha
+---
+
+A `SelectPanel` provides an anchor that will open an overlay with a list of selectable items, and a text input to filter the selectable items
+
+## Example
+
+## Component props
--- a/src/FilteredActionList/FilteredActionList.tsx
+++ b/src/FilteredActionList/FilteredActionList.tsx
@@ -1,6 +1,6 @@
-import React from 'react'
+import React, {useCallback} from 'react'
 import {GroupedListProps, ListPropsBase} from '../ActionList/List'
-import TextInput from '../TextInput'
+import TextInput, {TextInputProps} from '../TextInput'
 import Box from '../Box'
 import SelectMenu from '../SelectMenu'
 import {ActionList} from '../ActionList'
@@ -8,31 +8,36 @@ import {ActionList} from '../ActionList'
 export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
   loading?: boolean
   placeholderText: string
-  filterValue: string
-  onChange: (e: React.ChangeEvent<HTMLInputElement>) => unknown
+  setFilter: (value: string, e: React.ChangeEvent<HTMLInputElement>) => void
+  textInputProps?: Partial<TextInputProps>
 }
 
 export function FilteredActionList({
   loading = false,
   placeholderText,
-  filterValue,
-  onChange,
+  setFilter,
   items,
+  textInputProps,
   ...listProps
 }: FilteredActionListProps): JSX.Element {
+  const onInputChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      const value = e.target.value
+      setFilter(value, e)
+    },
+    [setFilter]
+  )
+
   return (
     <>
       <TextInput
         block
         width="auto"
         color="text.primary"
-        defaultValue={filterValue}
-        onChange={onChange}
+        onChange={onInputChange}
         placeholder={placeholderText}
         aria-label={placeholderText}
-        mx={2}
-        my={2}
-        contrast
+        {...textInputProps}
       />
       <Box flexGrow={1} overflow="auto">
         {loading ? <SelectMenu.LoadingAnimation /> : <ActionList items={items} {...listProps} role="listbox" />}
--- a/src/FilteredActionList/FilteredActionList.tsx
+++ b/src/FilteredActionList/FilteredActionList.tsx
@@ -9,7 +9,7 @@ export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, 
   loading?: boolean
   placeholderText: string
   onFilterChange: (value: string, e: React.ChangeEvent<HTMLInputElement>) => void
-  textInputProps?: Partial<TextInputProps>
+  textInputProps?: Partial<Omit<TextInputProps, 'onChange'>>
 }
 
 export function FilteredActionList({
--- a/src/FilteredActionList/FilteredActionList.tsx
+++ b/src/FilteredActionList/FilteredActionList.tsx
@@ -8,14 +8,14 @@ import {ActionList} from '../ActionList'
 export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
   loading?: boolean
   placeholderText: string
-  setFilter: (value: string, e: React.ChangeEvent<HTMLInputElement>) => void
+  onFilterChange: (value: string, e: React.ChangeEvent<HTMLInputElement>) => void
   textInputProps?: Partial<TextInputProps>
 }
 
 export function FilteredActionList({
   loading = false,
   placeholderText,
-  setFilter,
+  onFilterChange,
   items,
   textInputProps,
   ...listProps
@@ -23,9 +23,9 @@ export function FilteredActionList({
   const onInputChange = useCallback(
     (e: React.ChangeEvent<HTMLInputElement>) => {
       const value = e.target.value
-      setFilter(value, e)
+      onFilterChange(value, e)
     },
-    [setFilter]
+    [onFilterChange]
   )
 
   return (
--- a/src/FilteredActionList/FilteredActionList.tsx
+++ b/src/FilteredActionList/FilteredActionList.tsx
@@ -1,0 +1,44 @@
+import React from 'react'
+import {GroupedListProps, ListPropsBase} from '../ActionList/List'
+import TextInput from '../TextInput'
+import Box from '../Box'
+import SelectMenu from '../SelectMenu'
+import {ActionList} from '../ActionList'
+
+export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
+  loading?: boolean
+  placeholderText: string
+  filterValue: string
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => unknown
+}
+
+export function FilteredActionList({
+  loading = false,
+  placeholderText,
+  filterValue,
+  onChange,
+  items,
+  ...listProps
+}: FilteredActionListProps): JSX.Element {
+  return (
+    <>
+      <TextInput
+        block
+        width="auto"
+        color="text.primary"
+        defaultValue={filterValue}
+        onChange={onChange}
+        placeholder={placeholderText}
+        aria-label={placeholderText}
+        mx={2}
+        my={2}
+        contrast
+      />
+      <Box flexGrow={1} overflow="auto">
+        {loading ? <SelectMenu.LoadingAnimation /> : <ActionList items={items} {...listProps} role="listbox" />}
+      </Box>
+    </>
+  )
+}
+
+FilteredActionList.displayName = 'FilteredActionList'
--- a/src/FilteredActionList/FilteredActionList.tsx
+++ b/src/FilteredActionList/FilteredActionList.tsx
@@ -2,8 +2,8 @@ import React, {useCallback} from 'react'
 import {GroupedListProps, ListPropsBase} from '../ActionList/List'
 import TextInput, {TextInputProps} from '../TextInput'
 import Box from '../Box'
-import SelectMenu from '../SelectMenu'
 import {ActionList} from '../ActionList'
+import Spinner from '../Spinner'
 
 export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
   loading?: boolean
@@ -40,7 +40,13 @@ export function FilteredActionList({
         {...textInputProps}
       />
       <Box flexGrow={1} overflow="auto">
-        {loading ? <SelectMenu.LoadingAnimation /> : <ActionList items={items} {...listProps} role="listbox" />}
+        {loading ? (
+          <Box width="100%" display="flex" flexDirection="row" justifyContent="center" pt={6} pb={7}>
+            <Spinner />
+          </Box>
+        ) : (
+          <ActionList items={items} {...listProps} role="listbox" />
+        )}
       </Box>
     </>
   )
--- a/src/SelectPanel/SelectPanel.tsx
+++ b/src/SelectPanel/SelectPanel.tsx
@@ -106,8 +106,7 @@ export function SelectPanel({
           if (isMultiSelectVariant(selected)) {
             // multi select
             const otherSelectedItems = selected.filter(selectedItem => selectedItem !== item)
-            const newSelectedItems =
-              !item || selected.includes(item) ? otherSelectedItems : [...otherSelectedItems, item]
+            const newSelectedItems = selected.includes(item) ? otherSelectedItems : [...otherSelectedItems, item]
 
             const multiSelectOnChange = onSelectedChange as SelectPanelMultiSelection['onSelectedChange']
             multiSelectOnChange(newSelectedItems)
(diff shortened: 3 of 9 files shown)


docs/content/SelectPanel.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: SelectPanel
+status: Alpha
+---
+
+A `SelectPanel` provides an anchor that will open an overlay with a list of selectable items, and a text input to filter the selectable items
+
+## Example
+
+## Component props
```
smockle, line 10: 1. Would you mind opening an issue tracking either a. “Promote `SelectPanel` to “beta” status` or b. “Provide examples in the `SelectPanel` docs”?
2. What do you think about incorporating some of the prop documentation from https://github.com/primer/components/issues/1071 and https://github.com/primer/components/issues/1070?

src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -0,0 +1,44 @@
...
+import TextInput from '../TextInput'
+import Box from '../Box'
+import SelectMenu from '../SelectMenu'
+import {ActionList} from '../ActionList'
+
+export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
+  loading?: boolean
+  placeholderText: string
+  filterValue: string
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => unknown
```
smockle, line 12:
```suggestion
  onChange: (filterValue: string, event: React.ChangeEvent<HTMLInputElement>) => unknown
```
Suggestion: Pass both the new filter value and the event.

src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -0,0 +1,44 @@
...
+  items,
+  ...listProps
+}: FilteredActionListProps): JSX.Element {
+  return (
+    <>
+      <TextInput
+        block
+        width="auto"
+        color="text.primary"
+        defaultValue={filterValue}
```
smockle, line 29: I know we’d discussed using `TextInput` as a controlled component, which would require `value`, not `defaultValue`—where did we land on that?

src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -0,0 +1,44 @@
...
+        onChange={onChange}
+        placeholder={placeholderText}
+        aria-label={placeholderText}
+        mx={2}
+        my={2}
+        contrast
```
smockle, line 35: What is the effect of passing this `contrast` prop?

src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <>
+      <TextInput
+        block
+        width="auto"
+        color="text.primary"
+        defaultValue={filterValue}
+        onChange={onChange}
+        placeholder={placeholderText}
+        aria-label={placeholderText}
+        mx={2}
```
smockle, line 33: 1. Should we shift these style props (`width`, `block`, `color`, `mx`, `my`) into a `StyledTextInput`?
2. Should we apply `mx` and `my` to `FilteredActionList`s outside `SelectPanel`?

src/SelectPanel/SelectPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {ItemInput} from '../ActionList/List'
+import {FocusZoneSettings} from '../behaviors/focusZone'
+import {DropdownButton, DropdownButtonProps} from '../DropdownMenu'
+import {ItemProps} from '../ActionList'
+import {AnchoredOverlay} from '../AnchoredOverlay'
+import Flex from '../Flex'
+
+export interface SelectPanelProps extends Omit<FilteredActionListProps, 'onChange'> {
+  open: boolean
+  onOpen: (gesture: 'anchor-click' | 'anchor-key-press') => unknown
```
smockle, line 14:
```suggestion
  onOpen: AnchoredOverlay['onOpen']
```
Suggestion: Since this is passed through to `AnchoredOverlay`, consider referencing its [`onOpen` type](https://github.com/primer/components/blob/84c7e771f8ac9bb707fbb977658e9e40620bc20c/src/AnchoredOverlay/AnchoredOverlay.tsx#L21-L24).

Alternative suggestion: In addition to `extends`’ing `Omit<FilteredActionListProps, 'onChange'>`, also `extends` from `Pick<AnchoredOverlay, 'onOpen' | /*…*/ >`

src/SelectPanel/SelectPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+  onClose: (gesture: 'click-outside' | 'escape') => unknown
+  renderAnchor?: <T extends React.HTMLAttributes<HTMLElement>>(props: T) => JSX.Element
+  placeholder?: string
+  selectedItems: ItemInput[]
+  selectionVariant?: 'single' | 'multiple'
+  onChange?: (selectedItems: ItemInput[]) => unknown
```
smockle, line 20: What do you think about typing event handler as void-returning functions? IIUC, that’s the typing `react` uses: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1fe08659b58a63d4864a57c60b2a25f76fb01867/types/react/index.d.ts#L1293

src/SelectPanel/SelectPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+  focusableElementFilter: element => {
+    return !(element instanceof HTMLInputElement) || element.type !== 'checkbox'
+  }
+}
+
+export function SelectPanel({
+  open,
+  onOpen,
+  onClose,
+  renderAnchor = <T extends DropdownButtonProps>(props: T) => <DropdownButton {...props} />,
```
smockle, line 36:
```suggestion
  renderAnchor = DropdownButton,
```
For my own learnings, is this line equivalent?

src/SelectPanel/SelectPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+import Flex from '../Flex'
+
+export interface SelectPanelProps extends Omit<FilteredActionListProps, 'onChange'> {
+  open: boolean
+  onOpen: (gesture: 'anchor-click' | 'anchor-key-press') => unknown
+  onClose: (gesture: 'click-outside' | 'escape') => unknown
+  renderAnchor?: <T extends React.HTMLAttributes<HTMLElement>>(props: T) => JSX.Element
+  placeholder?: string
+  selectedItems: ItemInput[]
+  selectionVariant?: 'single' | 'multiple'
```
smockle, line 19: What do you think about making this a [discriminated union type](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions)? That’d prevent a component consumer from passing multiple `selectedItems` to a `SelectPanel` with `selectionVariant` `'single'`.

```tsx
type SingleSelectPanelProps = {
  selectionVariant: 'single'
  selectedItems: ItemInput
  onChange?: (selectedItem: ItemInput) => void
}
type MultipleSelectPanelProps = {
  selectedItems: ItemInput[]
  selectionVariant: 'multiple'
  onChange?: (selectedItems: ItemInput[]) => void
}
```

src/SelectPanel/SelectPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+        role: 'option',
+        selected: 'selected' in item && item.selected === undefined ? undefined : selectedItems.includes(item),
+        onAction: (itemFromAction, event) => {
+          item.onAction?.(itemFromAction, event)
+
+          if (event.defaultPrevented) {
+            return
+          }
+
+          if (selectionVariant === 'single') {
```
smockle, line 69: Lines 69–76 could be refactored, depending on  https://github.com/primer/components/pull/1224/files#r632789982

src/stories/SelectPanel.stories.tsx
```diff
@@ -0,0 +1,81 @@
+import {Meta} from '@storybook/react'
+import React, {useCallback, useState} from 'react'
+import {theme, ThemeProvider} from '..'
```
smockle, line 3: If you run `yarn run dist` does it warn about a circular reference? If so, I’d suggest `import`’ing from a location other than `..`.

src/__tests__/SelectPanel.tsx
```diff
@@ -0,0 +1,65 @@
+import {cleanup, render as HTMLRender} from '@testing-library/react'
+import 'babel-polyfill'
+import {axe, toHaveNoViolations} from 'jest-axe'
+import React from 'react'
+import theme from '../theme'
+import {SelectPanel} from '../SelectPanel'
+import {COMMON} from '../constants'
+import {behavesAsComponent, checkExports} from '../utils/testing'
+import {BaseStyles, ThemeProvider} from '..'
```
smockle, line 9: If you run `yarn run dist` does it warn about a circular reference? If so, I’d suggest `import`’ing from a location other than `..`.

src/SelectPanel/SelectPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {FocusZoneSettings} from '../behaviors/focusZone'
+import {DropdownButton, DropdownButtonProps} from '../DropdownMenu'
+import {ItemProps} from '../ActionList'
+import {AnchoredOverlay} from '../AnchoredOverlay'
+import Flex from '../Flex'
+
+export interface SelectPanelProps extends Omit<FilteredActionListProps, 'onChange'> {
+  open: boolean
+  onOpen: (gesture: 'anchor-click' | 'anchor-key-press') => unknown
+  onClose: (gesture: 'click-outside' | 'escape') => unknown
```
smockle, line 15: `ActionMenu` [uses `open` and `setOpen`](https://github.com/primer/components/blob/84c7e771f8ac9bb707fbb977658e9e40620bc20c/src/ActionMenu.tsx#L28-L36)—should we be consistent (either changing `ActionMenu` to use `open`/`onOpen`/`onClose`, or changing `SelectPanel` to use `open`/`setOpen`)?

src/SelectPanel/SelectPanel.tsx
smockle: Alternatives:
- Get rid of `selectionVariant` and use `selectedItem` vs `selectedItems` (mutually-exclusive)
- Use `selection?: ItemInput | ItemInput[]` instead of `selectedItem`/`selectedItems`

src/SelectPanel/SelectPanel.tsx
```diff
@@ -0,0 +1,107 @@
...
+  renderAnchor?: <T extends React.HTMLAttributes<HTMLElement>>(props: T) => JSX.Element
+  placeholder?: string
+  selectedItems: ItemInput[]
```
smockle, line 18: Non-blocking: I wonder if we should merge `items` and `selectedItems`, i.e. add a `selected` boolean to selected items in `items`. Since `items` updates as filters are applied (possible to exclude `selected` items not matching the current filter), we’d need to refactor some logic/assumptions (e.g. perhaps always retaining the full list of `items`, and adding an additional `matchesFilter` boolean).

src/stories/SelectPanel.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+        )}
+        placeholderText="Filter Labels"
+        open={open}
+        onOpen={onOpen}
+        onClose={onClose}
+        items={filteredItems}
+        selectedItems={selectedItems}
+        onChange={setSelectedItems}
+        filterValue={filterValue}
+        onFilterChange={onFilterChange}
```
smockle, line 75: From @dgreif in https://github.com/primer/components/pull/1224#issue-643683963:
>  The current API makes this component ultra controlled, which means flexibility, but also a lot of setup on the consumer side to use it. I think this is evident if you look at the story and tests. There are ~10 props required to use it properly.

It’s interesting to consider how a [slotted children API](https://codesandbox.io/s/slot-api-forked-rx5ci?file=/src/App.tsx) could reduce the props of `SelectPanel`, e.g. if `SelectPanel` had a `SelectPanel.Filter` child, `placeholderText`, `filterValue`, and `onFilterChange` could be moved there. cc @colebemis

src/FilteredActionList/FilteredActionList.tsx
colebemis: `contrast` gives the text input a slightly darker background.

src/stories/SelectPanel.stories.tsx
colebemis: Yeah, I'm interested to see if we could simplify the props by using the content-as-children API.

src/stories/SelectPanel.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const filteredItems = items.filter(item => item.text?.toLowerCase().startsWith(filterValue.toLowerCase()))
+  const onFilterChange = useCallback((e: React.ChangeEvent<HTMLInputElement>) => setFilterValue(e.target.value), [])
+  const [open, setOpen] = useState(false)
+  const onOpen = useCallback(() => setOpen(true), [])
+  const onClose = useCallback(() => setOpen(false), [])
+
+  return (
+    <>
+      <h1>Label Select Panel</h1>
+      <div id="">Please select labels that describe your issue:</div>
```
colebemis, line 60: What's the `id=""` for?

colebemis (commented):
This looks great for an alpha component ✨ Thanks for all your hard work and attention to detail on this, @dgreif! ❤️ I'll give this a more thorough review after you've addressed @smockle's thoughtful comments.

Before we promote this to beta, let's explore a content-as-children API to simplify the props.

src/stories/SelectPanel.stories.tsx
dgreif: Nothing, will remove 😄

src/FilteredActionList/FilteredActionList.tsx
dgreif: I'm going to remove the `filterValue` prop for now and have it be a 1-way piece of info that gets sent to the parent, but cannot be passed in.

src/SelectPanel/SelectPanel.tsx
dgreif: Going with the `setOpen` approach, passing `setOpen(open: boolean, gesture: 'possible gestures here')`.  We will need to circle back and probably change the `DropdownMenu` and `AnchoredOverlay` to have the same API

src/SelectPanel/SelectPanel.tsx
dgreif: This could get really tricky with combined local/remote loaded items.  I think the current API provides the most flexibility for now, but I'm willing to reconsider as we work on solidifying the API for all the layered components

src/SelectPanel/SelectPanel.tsx
smockle: > Going with the setOpen approach, passing setOpen(open: boolean, gesture: 'possible gestures here').

👍 Sounds good, thanks!

> We will need to circle back and probably change the `DropdownMenu` and `AnchoredOverlay` to have the same API.

Great idea! Could you please open an issue to track API consistency updates?

src/stories/SelectPanel.stories.tsx
dgreif: Nope, doesn't seem to care since it's in the tests

src/SelectPanel/SelectPanel.tsx
dgreif: I've simplified it to `props => <DropdownButton {...props} />`, but I think that's the smallest we can go.  Because of the forwarded ref, simple `DropdownButton` by itself has an `| null` in the return type.

src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -0,0 +1,49 @@
...
+import TextInput, {TextInputProps} from '../TextInput'
+import Box from '../Box'
+import SelectMenu from '../SelectMenu'
+import {ActionList} from '../ActionList'
+
+export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
+  loading?: boolean
+  placeholderText: string
+  onFilterChange: (value: string, e: React.ChangeEvent<HTMLInputElement>) => void
+  textInputProps?: Partial<TextInputProps>
```
colebemis, line 12: Should we omit `onChange` from this type?

colebemis (approved):
Looks good for alpha 👍  Let's experiment with the API a bit more before promoting to beta.

Excellent work, @dgreif ✨

PR #1225, opened 2021-05-13
fix(focusZone): handle complex child reordering

dgreif:
Using `focusZone` in `SelectPanel`, I have noticed scenarios where filtering causes items to still be present, but reorder based on similarity to the filter.  In these scenarios, focusZone currently has a tendency to mix up the focus order.  This happens when two items are moved, but put back in opposite order.  To avoid this confusion, we can simply handle all removals _first_, then handle additions.

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/__tests__/behaviors/focusZone.tsx
+++ b/src/__tests__/behaviors/focusZone.tsx
@@ -7,6 +7,9 @@ async function nextTick() {
   return new Promise(resolve => setTimeout(resolve, 0))
 }
 
+const moveDown = () => userEvent.type(document.activeElement!, '{arrowdown}')
+const moveUp = () => userEvent.type(document.activeElement!, '{arrowup}')
+
 // Since we use strict `isTabbable` checks within focus trap, we need to mock these
 // properties that Jest does not populate.
 beforeAll(() => {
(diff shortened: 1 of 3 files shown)


src/__tests__/behaviors/focusZone.tsx
```diff
@@ -472,3 +475,49 @@ it('Should set aria-activedescendant correctly', () => {
...
+  expect(document.activeElement).toEqual(firstButton)
+
+  moveDown()
+  expect(document.activeElement).toEqual(secondButton)
+
+  moveUp()
+  expect(document.activeElement).toEqual(firstButton)
+
+  // move secondButton and thirdButton to the end of the zone, in reverse order
+  focusZoneContainer.appendChild(thirdButton)
```
smockle, line 505: TIL (from [`Node.appendChild()` on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild):
> If the given child is a reference to an existing node in the document, `appendChild()` moves it from its current position to the new position (there is no requirement to remove the node from its parent node before appending it to some other node).

src/__tests__/behaviors/focusZone.tsx
```diff
@@ -7,6 +7,9 @@ async function nextTick() {
   return new Promise(resolve => setTimeout(resolve, 0))
 }
 
+const moveDown = () => userEvent.type(document.activeElement!, '{arrowdown}')
```
smockle, line 10: Verified the magic string `'{arrowdown}'`: https://github.com/testing-library/user-event#special-characters 👍

smockle (approved):

PR #1228, opened 2021-05-14
Action Menu correctly fires onAction callback after close

VanAnderson:
A couple of small fixes here that seem to be possibly from a botched rebase!

The after close behavior wasn't working quite as it should have been, because the `useEffect` hook had the wrong dependency (open vs combinedOpenState for the work to make ActionMenu optionally controlled) - further, the onAction callback was firing twice in the code when the component was operating as a controlled component due to both the `pendingActionRef` being populated and the call on L82.

### DIFF
--- a/.changeset/bright-donkeys-exist.md
+++ b/.changeset/bright-donkeys-exist.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Action Menu correctly fires onAction callback after close.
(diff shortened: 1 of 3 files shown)


dgreif (approved):
Good catch!

.changeset/bright-donkeys-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Action Menu correctly fires onAction callback after close.
```
dgreif, line 5: Might want to update these release notes to reflect the fact that it will be called _before_ close now

dgreif (approved):

PR #1229, opened 2021-05-14
refactor(focusZone): remove `currentFocusedIndex`

dgreif:
The current implementation for `focusZone` relies on two split values for state: `currentFocusedElement` and `currentFocusedIndex`.  In theory, `currentFocusedIndex` should always be equivalent to `focusableElements.indexOf(currentFocusedElement)`.  While the split state is currently maintained well if the DOM is static, there are a few corner cases that are not handled for DOM elements being added/removed.  This became apparent when I started work on active descendant integration for `SelectPanel`.  My changes here remove `currentFocusedIndex` in favor of _only_ tracking `currentFocusedElement`, and then calculating `currentFocusedIndex` in the one place that it is used.  This should be more robust and handle DOM changes more gracefully.

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/behaviors/focusZone.ts
+++ b/src/behaviors/focusZone.ts
@@ -344,7 +344,7 @@ export function focusZone(container: HTMLElement, settings?: FocusZoneSettings):
   const activeDescendantCallback = settings?.onActiveDescendantChanged
 
   let activeDescendantSuspended = Boolean(activeDescendantControl)
-  let currentFocusedElement = activeDescendantControl ? undefined : focusableElements[0]
+  let currentFocusedElement: HTMLElement | undefined
 
   function updateFocusedElement(to?: HTMLElement) {
     const from = currentFocusedElement
@@ -352,7 +352,7 @@ export function focusZone(container: HTMLElement, settings?: FocusZoneSettings):
 
     if (!activeDescendantControl) {
       if (from && from !== to && savedTabIndex.has(from)) {
-        from?.setAttribute('tabindex', '-1')
+        from.setAttribute('tabindex', '-1')
       }
 
       to?.setAttribute('tabindex', '0')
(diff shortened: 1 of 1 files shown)


src/behaviors/focusZone.ts
```diff
@@ -343,9 +343,18 @@ export function focusZone(container: HTMLElement, settings?: FocusZoneSettings):
...
+  let currentFocusedElement = activeDescendantControl ? undefined : focusableElements[0]
+
+  function updateFocusedElement(to?: HTMLElement) {
+    const from = currentFocusedElement
+    currentFocusedElement = to
+
     if (!activeDescendantControl) {
-      from?.setAttribute('tabindex', '-1')
+      if (from && from !== to && savedTabIndex.has(from)) {
+        from?.setAttribute('tabindex', '-1')
```
smockle, line 355: Is optional chaining necessary here, given the `if (from)` check on the preceding line?

src/behaviors/focusZone.ts
```diff
@@ -343,9 +343,18 @@ export function focusZone(container: HTMLElement, settings?: FocusZoneSettings):
   const activeDescendantControl = settings?.activeDescendantControl
   const activeDescendantCallback = settings?.onActiveDescendantChanged
 
-  function updateTabIndex(from?: HTMLElement, to?: HTMLElement) {
+  let activeDescendantSuspended = Boolean(activeDescendantControl)
+  let currentFocusedElement = activeDescendantControl ? undefined : focusableElements[0]
```
smockle, line 347: Since `focusableElements` is set to `[]` on line 336, isn’t `focusableElements[0]` also `undefined`?

```suggestion
  let currentFocusedElement: HTMLElement | undefined
```

src/behaviors/focusZone.ts
dgreif: Good catch! We need a lint rule for that

smockle (approved):
Looks good! 🚢  I left a couple non-blocking suggestions.

Sidebar: I tested the `useFocusZone` stories, and discovered an issue in [Changing Subtree](http://localhost:6006/?path=/story/hooks-usefocuszone--changing-subtree) (but it repros in `main` too, so it’s not a regression): If you remove all three buttons, then re-add three buttons, they aren’t part of the focus order (screen recording attached).

Screen recording demonstrating newly-added buttons are not in the focus order

PR #1235, opened 2021-05-18
Dialog Focus Fix

VanAnderson:
Small tweak to how dialog initial focus works.


dgreif (approved):
Looks good, seems less hacky than the `setTimeout` for sure 😄

PR #1238, opened 2021-05-19
New sizes for Overlay and Dialog

dgreif:
Introducing new sizes for `Overlay` because the originals were designed with `Dialog` in mind.  Since `Dialog` has it's own sizing, `Overlay` now has the opportunity to support a wider range.  

Also switching from `sm` to `small` naming convention to match PVC and other places in PRC like `Label`.  Still up in the air on the `xLarge` values.  Should they be `xLarge`, `extraLarge` or `xl`? 

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


docs/content/Dialog2.mdx
```diff
@@ -89,8 +89,8 @@ By default, the Dialog component implements the design and interactions defined
...
-| width         | `"sm" │ "md" │ "lg" │ "xl"`                    | `"xl"`     | The fixed width of the dialog.                                                                                                                                                                                                                                          |
-| height        | `"sm" │ "lg" │ "auto"`                         | `"auto"`   | The fixed height of the dialog, or, auto to adjust the height based on its contents.                                                                                                                                                                                    |
+| width         | `"small" │ "medium" │ "large" │ "xLarge"`                    | `"xLarge"`     | The fixed width of the dialog.                                                                                                                                                                                                                                          |
```
colebemis, line 92: Chatted with @yaili about this and we prefer `xlarge` (with a lowercase `l`). What do you think?

```suggestion
| width         | `"small" │ "medium" │ "large" │ "xlarge"`                    | `"xlarge"`     | The fixed width of the dialog.                                                                                                                                                                                                                                          |
```

colebemis (approved):

PR #1240, opened 2021-05-20
Fix Check Spacing on selected items

VanAnderson:
This was @dmarcey's suggestion for fixing spacing issues with ActionList based components that have both a leading visual and a selected check mark.

Before (bug):

image

After:

image


dgreif (approved):
Thanks for getting this fixed!

PR #1243, opened 2021-05-21
feat(SelectPanel): active descendant focus

dgreif:
Within the `SelectPanel`, focus will now be controlled using `active-descendant` instead of true focus movement.  This allows the user to keep their cursor in the text input, while using <kbd>up</kbd>/<kbd>down</kbd>/<kbd>enter</kbd> keyboard navigation to select/deselect items.  

I had to make a few tweaks within `focusZone` to optimize behavior when items are filtered.  The original implementation didn't fully take changing DOM into account.

I also had to _disable_ the `focusZone` that is naturally created by `AnchoredOverlay` within `SelectPanel` because it was causing odd behavior due to the overlap in `focusZones`. It currently doesn't handle focus well if you _click_ on an item, because <kbd>up</kbd>/<kbd>down</kbd> arrows do nothing at that point, but hitting <kbd>tab</kbd> brings you right back to the text input.  We may need to revisit this down the road, but overall the new setup feels really good.

### Screenshots
CleanShot 2021-05-20 at 21 15 23

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/odd-melons-rush.md
+++ b/.changeset/odd-melons-rush.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Use `active-descendant` to control focus in `SelectPanel`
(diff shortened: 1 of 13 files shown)


VanAnderson (approved):
👍

PR #1244, opened 2021-05-21
fix: allow focus to move from overlay on mousedown

### LINKED ISSUE #1236: Cannot click on item outside of ActionMenu to focus it while menu is open
**Describe the bug**

As a minimal repro, I updated the `ExternalOpenState` story in `ActionMenu.stories.tsx` to have a `<TextInput />` right next to the `<Button />` to control the menu's state.

**To Reproduce**
1. Open an ActionMenu
2. Click on something outside of the menu that it's easy to tell if it has focus or not (like an `<input>`)

**Expected behavior**
The menu closes and the the`<input>` should receive focus.

**Screenshots**
https://user-images.githubusercontent.com/5487287/118857827-790e6300-b8a6-11eb-88a3-59137bd63204.mov

dgreif:
All components built on `AnchoredOverlay` currently prevent the user from clicking into an input _outside_ the overlay and keeping focus there.  As soon as you click, the focus trap pulls focus back into the overlay, and then the overlay closes and focus moves to the anchor ref.  The only change needed to make this happen was to use `mousedown` instead of `click` to detect intent to close the overlay.  `mousedown` happens _before_ the `focus` event, but `click` happens _after_.  This means we can detect the `mousedown` and close the overlay + disable the `focusTrap` before the browser tries to move focus to an input outside of the overlay. 

Some additional changes I made here:
* Use `focus` instead of `focusin` to detect focus movement in `focusTrap`.  This isn't _required_, but it should be more reliable.  One note here, `focus` does not bubble, so we have to add the event listener to the `capture` phase.
* Use `capture` for the `mousedown` event as well, which prevents some other callback from using `stopPropagation` and accidentally keeping the overlay open.

Closes #1236

#### Click Outside Moves Focus
CleanShot 2021-05-21 at 09 21 43

#### Close with Escape Moves To Anchor
CleanShot 2021-05-21 at 09 22 51

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/nine-boxes-change.md
+++ b/.changeset/nine-boxes-change.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Allow focus to move to an input outside an overlay on mousedown
--- a/src/__tests__/ActionMenu.tsx
+++ b/src/__tests__/ActionMenu.tsx
@@ -104,7 +104,7 @@ describe('ActionMenu', () => {
     expect(portalRoot).toBeTruthy()
     const somethingElse = (await menu.baseElement.querySelector('#something-else')) as HTMLElement
     act(() => {
-      fireEvent.click(somethingElse)
+      fireEvent.mouseDown(somethingElse)
     })
     expect(portalRoot?.textContent).toEqual('') // menu items are hidden
   })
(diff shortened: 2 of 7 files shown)


colebemis (approved):
Love the attention to detail 💖 Excellent work, @dgreif!

PR #1246, opened 2021-05-21
fix(ConfirmationDialog): use correct border radius on buttons

### LINKED ISSUE #1239: Style issues with confirmation dialog
**Describe the bug**
Some styling issues with confirmation dialog component.

**To Reproduce**
Look at this component.

1. Title text is too bold --
2. Border radius on bottom (buttons) is too small compared to top border radii

**Expected behavior**

1. Should use `font-weight: 600` for bold text as the rest of the bold text in Primer does
2. Border radii are consistent in all 4 corners

**Screenshots**
image

**Desktop (please complete the following information):**
All

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

dgreif:
Because the `Dialog` now allows overflow (From #1203), the smaller border radius from the `button`s is overflowing past the larger border radius of the overlay, causing a different appearance in the header vs the footer.  This change puts the same border radius on the buttons as what is used in the Dialog so that they match

Second, the `ConfirmationDialog` title had a hard-coded `font-weight` of `700`.  This likely came from the Overaly Prototype, but it is not the correct value.  Switching to `fontWeights.bold` which is `600`.

Closes #1239

#### Before
image

#### After
image

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/thirty-lies-try.md
+++ b/.changeset/thirty-lies-try.md
@@ -2,4 +2,4 @@
 "@primer/components": patch
 ---
 
-Fix border radius on ConfirmationDialog buttons
+Fix border radius on buttons and title `font-weight` in ConfirmationDialog
--- a/.changeset/thirty-lies-try.md
+++ b/.changeset/thirty-lies-try.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Fix border radius on ConfirmationDialog buttons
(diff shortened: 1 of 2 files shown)


VanAnderson (approved):
Looks good!

PR #1247, opened 2021-05-21
 fix(SelectPanel): auto scroll when list overflows

dgreif:
Currently, SelectPanel is allowing the list contents to overflow.  This change brings back `overflow: auto` behavior.  It also adds `scrollIntoView` that will trigger when active descendant changes, so that selection is always in view.  I also added an example to the docs so that behavior can be tested in Vercel.

Closes https://github.com/github/primer/issues/169

### Screenshots
CleanShot 2021-05-24 at 20 38 54

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/soft-schools-nail.md
+++ b/.changeset/soft-schools-nail.md
@@ -2,4 +2,4 @@
 "@primer/components": patch
 ---
 
-Handle overflow and active-decendant scrolling within `SelectPanel`
+Handle overflow and active-descendant scrolling within `SelectPanel`
--- a/.changeset/soft-schools-nail.md
+++ b/.changeset/soft-schools-nail.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Handle overflow and active-decendant scrolling within `SelectPanel`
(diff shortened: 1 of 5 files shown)


VanAnderson (approved):
Pulled it down to check it out, looks good to me 👍

smockle:
From @dgreif in https://github.com/primer/components/pull/1247#issue-650448015:
> It also adds `scrollIntoView` that will trigger when active descendant changes, so that selection is always in view.

I’d initially misinterpreted “selection” in “so that selection is always in view” as the selected item (rather than as the active descendent), and was surprised to discover single selection `SelectPanel` `Item`s are not brought into view (either by moving them to the top of the list, or by scrolling to their list position) (screen recording):



I think this is non-blocking/out-of-scope. If this affects single selection `SelectPanel`s outside of this one Storybook, is it something we’d want to track separately?

dgreif:
@smockle great discovery!  My initial implementation of scroll into view is focused on `active-descendant`, but we could definitely expand it to include scrolling to the first selected option in the list.  I do think that's outside the scope of this change, but definitely worth discussing as another enhancement.

smockle (approved):

PR #1249, opened 2021-05-25
feat(SelectPanel): allow `filterValue` to be provided

dgreif:
The current implementation of `SelectPanel` and `FilteredActionList` maintains the filter input value _internally_ in the `FilteredActionList`.  This creates a few problems:

1. Filter value cannot be maintained across close/re-open of a `SelectPanel`.  I had to add [extra work-arounds](https://github.com/primer/components/pull/1224/files#diff-684f564a94f8ff9be2098b19e8c6451b63a606aaf8ab2bba93cb9e1ecbd0db43R70-R77) to specifically handle this case in the original implementation since the `input` would be re-created on every open.
2. The solution to problem 1 introduced a scenario where `onFilterChange` in `SelectPanel` needed to be called _without_ and `event`.  Ideally we wouldn't need this extra prop type, and we could just inherit  `onFilterChange` from `FilteredActionList` with type differences.
3. An _initial_ filter value cannot be provided.  This is an issue where typing a character to open the `SelectPanel` is a desired behavior.  This has been requested on an internal GitHub project.

The solution here is to use the `useProvidedStateOrCreate` hook in both `SelectPanel` and `FilteredActionList`.  This solves problems 1 and 2 with _no changes_ from the consumer.  State is now maintained across re-opens by the state within `SelectPanel`.  Problem 3 is solved by the consumer adding a `filterValue` prop to their `SelectPanel` declaration, which they will need to keep up-to-date on `onFilterChange` calls (which they are presumable doing already for item filtering purposes).  Overall, this provides much more flexibility to consumers, while making the types better and functionality more intuitive.

### Screenshots
CleanShot 2021-05-24 at 21 14 11

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/FilteredActionList/FilteredActionList.tsx
+++ b/src/FilteredActionList/FilteredActionList.tsx
@@ -7,10 +7,12 @@ import Spinner from '../Spinner'
 import {useFocusZone} from '../hooks/useFocusZone'
 import {uniqueId} from '../utils/uniqueId'
 import {itemActiveDescendantClass} from '../ActionList/Item'
+import {useProvidedStateOrCreate} from '../hooks/useProvidedStateOrCreate'
 
 export interface FilteredActionListProps extends Partial<Omit<GroupedListProps, keyof ListPropsBase>>, ListPropsBase {
   loading?: boolean
   placeholderText: string
+  filterValue?: string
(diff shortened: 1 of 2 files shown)


colebemis (approved):
✨

PR #1251, opened 2021-05-25
Handle on outside clicks in reverse order for overlay

VanAnderson:
overlay on click-3

Transplant logic from onEscapePress to onOutsideClick - this allows us to have Overlays inside of overlays that will respect proper close behavior!


src/hooks/useOnOutsideClick.tsx
```diff
@@ -47,12 +65,19 @@ export const useOnOutsideClick = ({containerRef, ignoreClickRefs, onClickOutside
...
     [onClickOutside, containerRef, ignoreClickRefs]
   )
-
   useEffect(() => {
-    // use capture to ensure we get all events
-    document.addEventListener('mousedown', onOutsideClickInternal, {capture: true})
+    if (handlers.length === 0) {
+      document.addEventListener('click', handleClick)
+    }
+    setTimeout(() => handlers.push(onOutsideClickInternal))
```
VanAnderson, line 72: 😭 This is the only way I can get things to work - otherwise the outside click event for the DropdownModal fires at the same time that the dropdown opens.

src/hooks/useOnOutsideClick.tsx
```diff
@@ -41,18 +59,27 @@ const shouldCallClickHandler = ({ignoreClickRefs, containerRef, e}: ShouldCallCl
...
+      return wasClickOutside
     },
     [onClickOutside, containerRef, ignoreClickRefs]
   )
-
   useEffect(() => {
-    // use capture to ensure we get all events
-    document.addEventListener('mousedown', onOutsideClickInternal, {capture: true})
+    if (handlers.length === 0) {
+      document.addEventListener('mousedown', handleClick, {capture: true})
```
dgreif, line 72:
```suggestion
      // use capture to ensure we get all events
      document.addEventListener('mousedown', handleClick, {capture: true})
```

dgreif (approved):
This new implementation requires users to memoize their `useOnOutsideClick` inputs properly, but I think it's the best we can do for today.  Looks good to me, but it would be good to get another set of eyes on it since I helped with implementation

smockle:
@dgreif @VanAnderson Would y’all mind taking a look at the edits in https://github.com/primer/components/pull/1251/commits/2b2f5e8ee20b962698520eb0d41a31662067e2fc? I synchronously reviewed them with @dgreif already, but I’m a bit wary to approve and merge this since I contributed to that refactor.

dgreif:
@smockle you changes look great.  I'm comfortable merging this PR at this point.  We can always circle back with refactors if needed, but the current code both _solves_ the problem _and doesn't have side-effects_ like the first couple iterations that were dependent on memoized props.

PR #1254, opened 2021-05-26
feat: Check lib & docs deps with Dependabot

smockle:
Part of https://github.com/github/primer/issues/181

This PR configures Dependabot to check dependencies in the root (library codebase) directory in addition to the existing check in the `/docs` (documentation codebase) directory.

The Dependabot docs include an [example configuration for checking multiple directories](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#directory) that I followed in this PR.


dgreif (commented):
Wow, surprising we were only checking the docs directory.  Nice find!  I'm curious though, could we simplify this with npm 7 workspaces?  I assume it would check workspaces if it runs the check at the `/` level and `/docs` is a workspace.

Edit: after checking the lock file again, it looks like you stuck with npm 6 😅.  Disregard this for now, but something to look forward to!

colebemis (approved):

PR #1257, opened 2021-05-27
RefAnchoredOverlay composes AnchoredOverlay

VanAnderson:
RefAnchoredOverlay is a composable piece to AnchoredOverlay that can be used independently in situations where the anchor is not a simple inline component.


VanAnderson:
closed in favor of https://github.com/primer/components/pull/1266

PR #1258, opened 2021-05-27
fix(focusTrap): prevent blur if trap container is not in DOM

dgreif:
There are some `AnchoredOverlay` use cases where the consumer moves focus to another element upon closing the overlay.  Currently, it's common that they move focus after closing the overlay, but the `focusTrap` has not yet been disabled because of the execution order of the effects.  In these cases, the `focusTrap` `container` is not physically present in the DOM at the time of moving focus.  This solution has `focusTrap` ensure that it is still actually in the DOM before trying to trap focus. 

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/__tests__/behaviors/focusTrap.tsx
+++ b/src/__tests__/behaviors/focusTrap.tsx
@@ -170,6 +170,35 @@ it('Should should release the trap when the signal is aborted', async () => {
   expect(document.activeElement).toEqual(durianButton)
 })
 
+it('Should should release the trap when the container is removed from the DOM', async () => {
+  const {container} = render(
+    <div>
+      <div id="trapContainer">
+        <button tabIndex={0}>Apple</button>
+      </div>
+      <button id="durian" tabIndex={0}>
+        Durian
+      </button>
+    </div>
+  )
+
+  const trapContainer = container.querySelector<HTMLElement>('#trapContainer')!
+  const durianButton = container.querySelector<HTMLElement>('#durian')!
+  const firstButton = trapContainer.querySelector('button')!
+
+  focusTrap(trapContainer)
+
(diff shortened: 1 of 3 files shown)


src/behaviors/focusTrap.ts
```diff
@@ -71,7 +71,7 @@ export function focusTrap(
   // is not found, blur the recently-focused element so that focus doesn't leave the
   // trap zone.
   function ensureTrapZoneHasFocus(focusedElement: EventTarget | null) {
-    if (focusedElement instanceof HTMLElement) {
+    if (focusedElement instanceof HTMLElement && document.contains(container)) {
```
smockle, line 74: Is the condition below more robust (i.e. supports either Light DOM or Shadow DOM)?

```suggestion
    if (focusedElement instanceof HTMLElement && focusedElement.getRootNode().contains(container)) {
```

[Documentation for Node.getRootNode()](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode)

src/__tests__/behaviors/focusTrap.tsx
```diff
@@ -170,6 +170,35 @@ it('Should should release the trap when the signal is aborted', async () => {
   expect(document.activeElement).toEqual(durianButton)
 })
 
+it('Should should release the trap when the container is removed from the DOM', async () => {
```
smockle, line 173: Great test! Thanks for covering this case. ✨

smockle (approved):
A couple non-blocking comments; this is good to merge! 🚀

PR #1266, opened 2021-05-28
Anchored overlay can take an external AnchorRef

VanAnderson:
This PR allows for more flexible relationships between anchor and overlay by allowing a user to supply their own external AnchorRef.


src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -10,7 +10,12 @@ export interface AnchoredOverlayProps extends Pick<OverlayProps, 'height' | 'wid
    * A custom function component used to render the anchor element.
    * Will receive the selected text as `children` prop when an item is activated.
    */
-  renderAnchor: <T extends React.HTMLAttributes<HTMLElement>>(props: T) => JSX.Element
+  renderAnchor: <T extends React.HTMLAttributes<HTMLElement>>(props: T) => JSX.Element | null
```
dgreif, line 13: Should we allow a direct value of `null` instead of `() => null`?  That way it isn't seen as a prop change every time the consumer re-renders.  If so, we may want to get a little fancier with the types and _require_ an `anchorRef` when `renderAnchor` is `null`

src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -53,7 +59,8 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({
   overlayProps,
   focusZoneSettings
 }) => {
-  const anchorRef = useRef<HTMLElement>(null)
+  const internalAnchorRef = useRef<HTMLElement>(null)
+  const anchorRef = externalAnchorRef || internalAnchorRef
```
dgreif, line 63: `useProvidedRefOrCreate` here as well

src/stories/Overlay.stories.tsx
```diff
@@ -101,3 +102,94 @@ export const DialogOverlay = () => {
...
+        onMouseDown: (e: React.MouseEvent) => {
+          e.preventDefault()
+        }
+      }
+    ],
+    []
+  )
+  const [selectedItem, setSelectedItem] = React.useState<ItemInput | undefined>()
+  return (
+    <Position position="absolute" top={0} left={0} bottom={0} right={0} ref={anchorRef}>
```
dgreif, line 140: It doesn't look like this story actually passes in `anchorRef` to any of the overlays.  Is it still needed?  I feel like this was actually from the overlay click outside testing.

src/stories/RefAnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const RefAnchoredOverlay = () => {
+  const [isOpen, setIsOpen] = useState(false)
+  const buttonRef = useRef<HTMLButtonElement>(null)
+  const confirmButtonRef = useRef<HTMLButtonElement>(null)
+  const anchorRef = useRef<HTMLDivElement>(null)
+  const closeOverlay = () => setIsOpen(false)
+  return (
+    <Position position="absolute" top={0} left={0} bottom={0} right={0} ref={anchorRef}>
+      <Button ref={buttonRef} onClick={() => setIsOpen(!isOpen)}>
```
dgreif, line 31: Would it be better to have the button as the anchor?  Or some other separate element that really shows off the disconnection between the open button and the menu positioning.

src/stories/RefAnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+          <BaseStyles>
+            <Story />
+          </BaseStyles>
+        </ThemeProvider>
+      )
+    }
+  ]
+} as Meta
+
+export const RefAnchoredOverlay = () => {
```
dgreif, line 23: This is a better/more direct example.  Since we no longer have the `RefAnchoredOverlay`, it probably makes sense to move this into one of the other `.stories` files.  I'd suggest switching this to use `ActionMenu` and renaming it to `ActionMenuWithExternalAnchor` or something like that.

dgreif (changes requested):
Great changes!  I think there will be many use cases for external anchors, and I like this approach.

src/stories/Overlay.stories.tsx
VanAnderson: whoops, that must have snuck in here!

dgreif (approved):
Looks good!

smockle (approved):

PR #1270, opened 2021-05-28
fix: Don’t focus first 'Item' of 'DropdownMenu' and 'SelectMenu' on open

smockle:
Fixes https://github.com/github/primer/issues/188

### Screenshots

Note that the first `Item` is no longer focused for `DropdownMenu` or `ActionMenu` when the components open:

**`DropdownMenu`**
DropdownMenu

**`ActionMenu`**
ActionMenu

Note that the filter is still focused when a `SelectPanel` opens:

**`SelectPanel`**
SelectPanel

### Merge checklist
- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/behaviors/focusTrap.ts
+++ b/src/behaviors/focusTrap.ts
@@ -68,8 +68,7 @@ export function focusTrap(
   // Ensure focus remains in the trap zone by checking that a given recently-focused
   // element is inside the trap zone. If it isn't, redirect focus to a suitable
   // element within the trap zone. If need to redirect focus and a suitable element
-  // is not found, blur the recently-focused element so that focus doesn't leave the
-  // trap zone.
+  // is not found, focus the container.
   function ensureTrapZoneHasFocus(focusedElement: EventTarget | null) {
     if (focusedElement instanceof HTMLElement && document.contains(container)) {
       if (container.contains(focusedElement)) {
@@ -80,16 +79,14 @@ export function focusTrap(
         if (lastFocusedChild && isTabbable(lastFocusedChild) && container.contains(lastFocusedChild)) {
           lastFocusedChild.focus()
           return
+        } else if (initialFocus && container.contains(initialFocus)) {
+          initialFocus.focus()
+          return
         } else {
-          const toFocus = initialFocus && container.contains(initialFocus) ? initialFocus : getFocusableChild(container)
-          if (toFocus) {
-            toFocus.focus()
-            return
-          } else {
-            // no element focusable within trap, blur the external element instead
-            // eslint-disable-next-line github/no-blur
-            focusedElement.blur()
-          }
+          container.setAttribute('tabindex', '-1')
+          container.focus()
+          container.removeAttribute('tabindex')
+          return
         }
       }
     }
--- a/src/behaviors/focusTrap.ts
+++ b/src/behaviors/focusTrap.ts
@@ -83,9 +83,14 @@ export function focusTrap(
           initialFocus.focus()
           return
         } else {
-          container.setAttribute('tabindex', '-1')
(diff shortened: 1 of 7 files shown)


bscofield (approved):
👍

src/behaviors/focusTrap.ts
```diff
@@ -80,16 +79,14 @@ export function focusTrap(
...
-          if (toFocus) {
-            toFocus.focus()
-            return
-          } else {
-            // no element focusable within trap, blur the external element instead
-            // eslint-disable-next-line github/no-blur
-            focusedElement.blur()
-          }
+          container.setAttribute('tabindex', '-1')
+          container.focus()
```
dgreif, line 87: I'm hesitant to take this approach.  We are forcing the container to be focusable for _all_ focus traps, but we only need this behavior for Overlays.  Our overlays have an `outline: none`, but all consumers will need to add that as well if they don't actually want focus styling on their container.  Additionally, putting focus on the container means it could easily receive key presses or click events that are not intended as an action.  

Is there a specific reason we can't simply `blur`?  Maybe if `initialFocus === null`?

dgreif: @smockle and I discussed multiple approaches to this problem at length

## Switch back to `blur`
The plus here is that we don't ever focus an element that isn't intended to be focused.  The down-side is that focus zones will lose their key bindings because focus is on the body after a `blur`, and the zone only listens for key presses on the container.  We _could_ add key listeners on the body, but it would be difficult to accurately keep track of which trap blurred focus and which zones that impacts.

## Add a "sentinel" element for focus
This is an approach @smockle has used before.  Add a non-visible but focusable element to the beginning of the focus trap that can receive focus initially.  Once focus moves to other elements within the trap, the sentinel can be removed so that it doesn't get focused again later.  This is _similar_ to the container focus approach in that it keeps focus within the zone if a zone and trap are used on the same container, but using the container prevents us from needing a new element, and seems slightly better for accessibility.

## Bring back anchor/list state in AnchoredOverlay.
Previously, the AnchoredOverlay initially had focus on the _anchor_ and then moved focus into the Overlay on certain key presses.  This was changed in https://github.com/primer/components/pull/1222 to simplify the component and user interactions.  The plus side to the original setup was that it had the desired behavior for this PR, which is that focus didn't move to the first item in ActionMenu/Dropdown as soon as they are opened.

In the end, we decided it makes sense to keep this container focus approach for now.  We will continue to think through alternatives and can easily revert this change if needed down the road

smockle:
@dgreif I think I’ve addressed the two outstanding issues you identified:
- Consolidate index logic in `getCurrentFocusedIndex`: https://github.com/primer/components/pull/1270/commits/f7cb7d33092831d1c4dfd124112199f4ab20f7aa
- Preserve existing container `tabIndex`: https://github.com/primer/components/pull/1270/commits/8378b25b579d26b7c010e4268333ac55e1344639

I’d appreciate another review when you have a chance! 🦺

src/behaviors/focusZone.ts
```diff
@@ -587,6 +587,10 @@ export function focusZone(container: HTMLElement, settings?: FocusZoneSettings):
       return 0
     }
 
+    if (currentFocusedElement === container) {
+      return -1
+    }
```
dgreif, line 592: Does this work in a situation where `container` is already focusable?  Technically we would want to return `0` in that case, right?

smockle: Fixed in https://github.com/primer/components/pull/1270/commits/40fac046d9d09f4bf9b548f0f6b1c50167959052

dgreif (approved):
🚀

PR #1275, opened 2021-06-01
data-test-id can be used with OverlayProps

VanAnderson:
This is a simple fix to https://github.com/github/primer/issues/162 to enable `data-test-id` to be used in OverlayProps.

Curious if there might be some better solutions here, it seems like it would be better to have a way to add all `data-*` props to the OverlayProps.

### DIFF
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -65,6 +65,7 @@ export type OverlayProps = {
   onClickOutside: (e: TouchOrMouseEvent) => void
   onEscape: (e: KeyboardEvent) => void
   visibility?: 'visible' | 'hidden'
+  'data-test-id'?: string
 } & Omit<ComponentProps<typeof StyledOverlay>, 'visibility' | keyof SystemPositionProps>
 
 /**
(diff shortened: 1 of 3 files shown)


src/Overlay.tsx
```diff
@@ -65,6 +65,7 @@ export type OverlayProps = {
   onClickOutside: (e: TouchOrMouseEvent) => void
   onEscape: (e: KeyboardEvent) => void
   visibility?: 'visible' | 'hidden'
+  'data-test-id'?: string
```
dgreif, line 68: This does feel very specific to the use case we are trying to solve.  I did some research and it looks like template literals would be great here, like ``[key: `data-${string}`]: string``, but that isn't supported at this time (https://github.com/microsoft/TypeScript/pull/26797).  As a workaround, we could allow addition keys with something like `[additionalKey: string]: unknown`

VanAnderson: Great idea, thanks for weighing in.

dgreif (approved):

PR #1276, opened 2021-06-02
implement Scroll flash into SelectPanel

VanAnderson:
Closes https://github.com/github/primer/issues/183

`useScrollFlash` takes in a ref and implements https://adxlv.computer/projects/flash-scrollers/

The issue implies that this should be a global solution for overlay components, but I think we will want to implement it in the specific use cases of Overlay being used - this is because the scrollable container itself is implemented in the specific use cases of `Overlay` rather than globally on overlay itself.

Looking at the issue, it seems the primary place where this was needed was SelectPanel, but we can easily put it elsewhere just by using the hook!

### Screenshots

scroll-flash

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/hooks/useScrollFlash.ts
+++ b/src/hooks/useScrollFlash.ts
@@ -17,8 +17,5 @@ export default function useScrollFlash(scrollContainerRef: React.RefObject<HTMLE
 
     scrollContainer.scrollTop = altScroll
     scrollContainer.scrollTop = currentScroll
-
-    const flashScrollersEvent = new Event('flashScrollers')
-    scrollContainer.dispatchEvent(flashScrollersEvent)
   }, [scrollContainerRef])
 }
--- a/src/hooks/useScrollFlash.ts
+++ b/src/hooks/useScrollFlash.ts
@@ -1,0 +1,27 @@
+import React, {useEffect} from 'react'
+/**
+ * This hook will flash the scrollbars for a ref of a container that has scrollable overflow
+ * @param scrollContainerRef The ref of the scrollable content
+ */
+export default function useScrollFlash(scrollContainerRef: React.RefObject<HTMLElement>) {
+  // https://adxlv.computer/projects/flash-scrollers/
+  useEffect(() => {
+    const scrollContainer = scrollContainerRef.current
+    if (!scrollContainer) {
+      return
+    }
+    const currentScroll = scrollContainer.scrollTop
+    const maxScroll = scrollContainer.scrollHeight
+
+    const altScroll = (() => {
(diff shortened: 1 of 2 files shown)


labrose:
Nice! Tried it out in all those browsers and it's looking great 😄 The fade out is smooth but the bar is there long enough for you to notice.

---

Unrelated to the scope of your pull request but I did notice when testing in Firefox the checkboxes are much smaller compared to other browsers. Version 89.0

Checkboxes to select labels half size of color dot

src/hooks/useScrollFlash.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    const altScroll = (() => {
+      if (currentScroll < 1 || currentScroll < maxScroll) return currentScroll + 1
+      else return currentScroll - 1
+    })()
+
+    scrollContainer.scrollTop = altScroll
+    scrollContainer.scrollTop = currentScroll
+
+    const flashScrollersEvent = new Event('flashScrollers')
```
smockle, line 24: Are we listening for this event somewhere?

src/hooks/useScrollFlash.ts
```diff
@@ -0,0 +1,27 @@
...
+  // https://adxlv.computer/projects/flash-scrollers/
+  useEffect(() => {
+    const scrollContainer = scrollContainerRef.current
+    if (!scrollContainer) {
+      return
+    }
+    const currentScroll = scrollContainer.scrollTop
+    const maxScroll = scrollContainer.scrollHeight
+
+    const altScroll = (() => {
```
smockle, line 16: I do love [IIFEs](https://developer.mozilla.org/en-US/docs/Glossary/IIFE), but a ternary may be clearer in this case:

```suggestion
    const altScroll = currentScroll < Math.min(1, maxScroll) ? currentScroll + 1 : currentScroll - 1;
```

smockle (commented):
A couple comments re: code cleanup. I like the overall direction of this—nice idea, making it a hook! ✨

src/hooks/useScrollFlash.ts
smockle: Resolved in https://github.com/primer/components/pull/1276/commits/bc4b4b1ba4bdcbe5240de7216c9db172901ca5ab

smockle (approved):
Nice work! LGTM! ✨

vdepizzol:
This is awesome! 🎉

Issue #1278, opened 2021-06-07
Overlay component box shadows in dark and dimmed modes


pmarsceill:
**Describe the bug**
Using the overlay component in dark and dimmed mode doesn't adjust the box shadow to the proper values. On Dotcom, we use different box shadows for dark modes:

<img width="366" alt="image" src="https://user-images.githubusercontent.com/896475/121066651-1cf97900-c798-11eb-8a7e-e403872d4b91.png">

Currently PRC uses the same shadows for dark and light modes which makes the dropdown look like it's foggy and sad in dark modes:

<img width="217" alt="image" src="https://user-images.githubusercontent.com/896475/121067078-9ee9a200-c798-11eb-9926-0c7eeb136494.png">

**Note:** As this is a public repo, please be careful not to include details or screenshots from unreleased GitHub products or features. In most cases, a good bug report should be able to describe the new component without including business logic or feature details, but if you must discuss context relating to an unreleased feature, please open an issue in the private [Design Systems repo](https://github.com/github/design-systems/issues/new/choose) and link to it here.

**To Reproduce**
Use an overlay component in dark mode

**Expected behavior**
Same dropshadow for dark mode as we use on Dotcom 

**Desktop (please complete the following information):**
n/a

**Smartphone (please complete the following information):**
n/a

bscofield:
https://github.com/primer/components/pull/1279

PR #1279, opened 2021-06-07
Dark mode color tweaks

VanAnderson:
Using functional colors for shadow on the overlay. I looked at what we were doing in dotcom, and it seems to simply be using `shadow.large`.

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/Overlay.tsx
```diff
@@ -34,7 +34,7 @@ const widthMap = {
 /*TODO replace with shadow functional color variable when it's shipped to primer/primitives*/
 const StyledOverlay = styled.div<StyledOverlayProps & SystemCommonProps & SystemPositionProps & SxProp>`
   background-color: ${get('colors.bg.overlay')};
-  box-shadow: 0px 1px 3px rgba(0, 0, 0, 0.12), 0px 8px 24px rgba(149, 157, 165, 0.2);
+  box-shadow: ${get('shadows.shadow.large')};
```
dgreif, line 37: It looks like the old `Dialog` uses `large` (https://github.com/primer/components/blob/main/src/Dialog.tsx#L22), but the old `SelectMenu` uses `small` (https://github.com/primer/components/blob/main/src/SelectMenu/SelectMenuModal.tsx#L37).  Might be worth checking with @vdepizzol to see which would would prefer in the short term

vdepizzol:
Hello @VanAnderson @dgreif! I'd like to make sure the Overlay patterns use the new shadows as proposed here: https://github.com/github/primer/issues/43.

I'd wait for functional variables to have updated values before merging this.

c/c @auareyou

VanAnderson:
> Hello @VanAnderson @dgreif! I'd like to make sure the Overlay patterns use the new shadows as proposed here: [github/primer#43](https://github.com/github/primer/issues/43).
> 
> I'd wait for functional variables to have updated values before merging this.
> 
> c/c @auareyou

I think @pmarsceill mentioned this fix (https://github.com/primer/components/issues/1278) should be prioritized for the upcoming Memex release. Unless https://github.com/github/primer/issues/43 is right around the corner, do you think we could merge this and then follow up to update it with the new functional vars?

vdepizzol:
> I think @pmarsceill mentioned this fix (#1278) should be prioritized an upcoming release of the project they are working on. Unless github/primer#43 is right around the corner, do you think we could merge this and then follow up to update it with the new functional vars?

👍  Absolutely! 

@auareyou do you have a timeline of when we can fix the shadows in primer primitives?

auareyou:
I'm on it! We'll get this updated today

vdepizzol:
@VanAnderson for reference, @simurai is working on a Primer Primitives here: https://github.com/primer/primitives/pull/135

VanAnderson:
> I'm on it! We'll get this updated today

Awesome, thanks @vdepizzol! @auareyou if you could ping me in slack once you complete that work I will make sure to expedite getting this work in :)

colebemis (approved):

PR #1280, opened 2021-06-09
Export composable pieces of List/Item

VanAnderson:


src/ActionList/Item.tsx
```diff
@@ -229,7 +229,10 @@ const StyledItem = styled.div<
   ${sx}
 `
 
-const StyledTextContainer = styled.div<{descriptionVariant: ItemProps['descriptionVariant']}>`
+export const TextContainer = styled.div<{
+  dangerouslySetInnerHtml?: {__html: string}
```
smockle, line 233: I’ll investigate whether we can import [this type from React](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/78f0288bd82dba437b7e5a97ec4af61ac75b390c/types/react/index.d.ts#L1346-L1348) after my interview.

I’m thinking it’ll be something like this, but haven’t had a chance to test yet:
```suggestion
  dangerouslySetInnerHtml?: React.DOMAttributes<HTMLDivElement>["dangerouslySetInnerHTML"]
```

VanAnderson: this worked well for me locally! Thanks!

smockle (approved):
Nice work! Thanks for the quick turnaround on this! 🚀

PR #1286, opened 2021-06-10
Handle invalid color scheme names

colebemis:
## Problem

If `ThemeProvider` is a passed an invalid color scheme name (e.g. `<ThemeProvider dayScheme="foo">`), it could potentially break an entire application. We noticed this happening in an internal React application.

## Solution

This PR updates `ThemeProvider` to default to the first defined color scheme if it's passed an invalid color scheme name.

### DIFF
--- a/src/ThemeProvider.tsx
+++ b/src/ThemeProvider.tsx
@@ -154,7 +154,10 @@ function applyColorScheme(theme: Theme, colorScheme: string) {
   if (!theme.colorSchemes[colorScheme]) {
     // eslint-disable-next-line no-console
     console.error(`\`${colorScheme}\` scheme not defined in \`theme.colorSchemes\``)
-    return theme
+
+    // Apply the first defined color scheme
+    const defaultColorScheme = Object.keys(theme.colorSchemes)[0]
+    return deepmerge(theme, theme.colorSchemes[defaultColorScheme])
   }
 
   return deepmerge(theme, theme.colorSchemes[colorScheme])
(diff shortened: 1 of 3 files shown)


src/ThemeProvider.tsx
```diff
@@ -154,7 +154,10 @@ function applyColorScheme(theme: Theme, colorScheme: string) {
   if (!theme.colorSchemes[colorScheme]) {
     // eslint-disable-next-line no-console
     console.error(`\`${colorScheme}\` scheme not defined in \`theme.colorSchemes\``)
-    return theme
+
+    // Apply the first defined color scheme
+    const defaultColorScheme = Object.keys(theme.colorSchemes)[0]
```
smockle, line 159: I wonder if index `0` is [magic-numbery](https://en.wikipedia.org/wiki/Magic_number_(programming))? Though it may be our best option, since [the `Theme` type definition](https://github.com/primer/components/blob/f0f8edc7f55ac243d58c76e8346072886299cedb/src/ThemeProvider.tsx#L11) doesn’t require a default scheme to be designated (example below). 🤔 

```ts
export type Theme = {
  colorSchemes: {
    default: { [key: string]: any }
    [key: string]: any
  }
  [key: string]: any
}
```

smockle (approved):

dgreif (approved):
Great change.  Definitely makes sense to have a fallback here

src/ThemeProvider.tsx
colebemis: I agree that `[0]` feels a little bit wrong here. Unfortunetly, we can't make any assumptions about the format of the `theme` object without making a breaking change to the library. Let's move forward with this "incrementally correct" solution, and we can revisit TypeScript types for the `theme` object later.

PR #1287, opened 2021-06-10
feat: Add `height="initial"` to `Overlay`

smockle:
Partial-fix for https://github.com/github/primer/issues/161. Specifically, this fixes the case where `SelectPanel` items are available initially (i.e. not asynchronously fetched).

This PR adds an `"initial"` value to `Overlay`’s `height` prop, which cause the component height to fit contents initially (like `"auto"`), then be fixed at that initial height, so that as contents change (e.g. by filtering), the container remains at the initial height (unlike, and preferred to, `"auto"`).

Sidebar: The asynchronously-fetch case is partially-supported: A component consumer can initially (before items have been retrieved) pass `height="auto"` then (once items are retrieved) pass `height="initial"`. Because `heightKey` is a dependency of the `useEffect`, the height-after-fetch can be set using this approach.

### Screen recordings
**Overflowing items**
The following recording demonstrates that an `Overlay` with `height="initial"` initially opens shorter than its contents (which are scrollable, because they do not fit within the `Overlay`’s `max-height`), and remains at this height as contents are filtered. This is functionally identical to `height="XSmall"` when items overflow.

**Underflowing items**
The following recording demonstrates that an `Overlay` with `height="initial"` initially opens exactly as large as its contents (which are not scrollable, because they fit within the `Overlay`’s `max-height`), and remains at this height as contents are filtered. This is superior to both `height="XSmall"` (which, with this few items, would undesirably include empty space underneath them when the `Overlay` opens) and `height="auto"` (which would remove empty space after filtering, undesirably varying the `Overlay` height).

### Merge checklist
- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -10,7 +10,7 @@ import {useCombinedRefs} from './hooks/useCombinedRefs'
 type StyledOverlayProps = {
   width?: keyof typeof widthMap
   height?: keyof typeof heightMap
-  maxHeight?: keyof typeof heightMap
+  maxHeight?: keyof Omit<typeof heightMap, 'auto' | 'initial'>
   visibility?: 'visible' | 'hidden'
 }
 
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -19,7 +19,8 @@ const heightMap = {
   medium: '320px',
   large: '432px',
   xlarge: '600px',
-  auto: 'auto'
+  auto: 'auto',
+  initial: 'auto' // Passing 'initial' initially applies 'auto'
 }
 
 const widthMap = {
@@ -68,7 +69,7 @@ export type OverlayProps = {
   onClickOutside: (e: TouchOrMouseEvent) => void
   onEscape: (e: KeyboardEvent) => void
   visibility?: 'visible' | 'hidden'
-  height?: keyof typeof heightMap | 'initial'
+  height?: keyof typeof heightMap
   [additionalKey: string]: unknown
 } & Omit<ComponentProps<typeof StyledOverlay>, 'height' | 'visibility' | keyof SystemPositionProps>
 
@@ -103,7 +104,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OverlayProps>(
     const overlayRef = useRef<HTMLDivElement>(null)
     const combinedRef = useCombinedRefs(overlayRef, forwardedRef)
 
-    const overlayProps = useOverlay({
+    useOverlay({
       overlayRef,
       returnFocusRef,
       onEscape,
@@ -112,18 +113,16 @@ const Overlay = React.forwardRef<HTMLDivElement, OverlayProps>(
       initialFocusRef
     })
 
-    const initialHeight = useRef<string>('auto')
     useEffect(() => {
-      if (overlayRef.current?.clientHeight) {
-        initialHeight.current = `${overlayRef.current.clientHeight}px`
+      if (heightKey === 'initial' && combinedRef.current?.clientHeight) {
+        combinedRef.current.style.height = `${combinedRef.current.clientHeight}px`
       }
-    }, [overlayRef])
-    const height = heightKey === 'initial' ? initialHeight.current : heightMap[heightKey || 'auto']
+    }, [heightKey, combinedRef])
+    const height = heightMap[heightKey || 'auto']
 
     return (
       <Portal>
         <StyledOverlay
-          {...overlayProps}
           aria-modal="true"
           role={role}
           height={height}
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -9,7 +9,7 @@ import {useCombinedRefs} from './hooks/useCombinedRefs'
 
 type StyledOverlayProps = {
   width?: keyof typeof widthMap
-  height?: string
+  height?: keyof typeof heightMap
   visibility?: 'visible' | 'hidden'
 }
 
@@ -39,7 +39,7 @@ const StyledOverlay = styled.div<StyledOverlayProps & SystemCommonProps & System
   position: absolute;
   min-width: 192px;
   max-width: 640px;
-  height: ${props => props.height};
+  height: ${props => heightMap[props.height || 'auto']};
   width: ${props => widthMap[props.width || 'auto']};
   border-radius: 12px;
   overflow: hidden;
@@ -69,9 +69,8 @@ export type OverlayProps = {
   onClickOutside: (e: TouchOrMouseEvent) => void
   onEscape: (e: KeyboardEvent) => void
   visibility?: 'visible' | 'hidden'
-  height?: keyof typeof heightMap
   [additionalKey: string]: unknown
-} & Omit<ComponentProps<typeof StyledOverlay>, 'height' | 'visibility' | keyof SystemPositionProps>
+} & Omit<ComponentProps<typeof StyledOverlay>, 'visibility' | keyof SystemPositionProps>
 
 /**
  * An `Overlay` is a flexible floating surface, used to display transient content such as menus,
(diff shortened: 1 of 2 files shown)


src/Overlay.tsx
```diff
@@ -9,7 +9,7 @@ import {useCombinedRefs} from './hooks/useCombinedRefs'
 
 type StyledOverlayProps = {
   width?: keyof typeof widthMap
-  height?: keyof typeof heightMap
+  height?: string
```
dgreif, line 12: This could be switched back now that we have `initial` in the map, right?

src/Overlay.tsx
```diff
@@ -68,8 +69,9 @@ export type OverlayProps = {
   onClickOutside: (e: TouchOrMouseEvent) => void
   onEscape: (e: KeyboardEvent) => void
   visibility?: 'visible' | 'hidden'
+  height?: keyof typeof heightMap
   [additionalKey: string]: unknown
-} & Omit<ComponentProps<typeof StyledOverlay>, 'visibility' | keyof SystemPositionProps>
+} & Omit<ComponentProps<typeof StyledOverlay>, 'height' | 'visibility' | keyof SystemPositionProps>
```
dgreif, line 74: I think these changes can all be removed as well

src/stories/SelectPanel.stories.tsx
```diff
@@ -106,3 +106,66 @@ export function SingleSelectStory(): JSX.Element {
...
+        )}
+        placeholderText="Filter Labels"
+        open={open}
+        onOpenChange={setOpen}
+        items={filteredItems}
+        selected={selected}
+        onSelectedChange={setSelected}
+        onFilterChange={setFilter}
+        showItemDividers={true}
+        overlayProps={{width: 'small', height: 'initial', sx: {maxHeight: '192px'}}}
```
dgreif, line 134: I am a fan of the idea of lifting `maxHeight` to a full prop and using the keys from the map as values.  I can't remember if we came to a final decision on that approach.

dgreif (changes requested):
Thanks for taking this on @smockle!  I've been dreading this change and you did a fantastic job.  Super simple solution and it works great!

src/Overlay.tsx
smockle: Yup, that change is no longer necessary—we can realign the internal (`StyledOverlayProps`) and external (`OverlapProps`) contracts.

Incidentally, the reason isn’t because `initial` is included in `heightMap`, but rather because [imperatively setting `height`](https://github.com/primer/components/blob/c54c7e5b8e025983dd8190de9d50d356ab55e924/src/Overlay.tsx#L118) means we are no longer passing arbitrary `height` values to the `StyledOverlay` component, so the type of `height` no longer needs to be widened from `keyof typeof heightMap` to `string`.

smockle: Fixed in https://github.com/primer/components/pull/1287/commits/eb770c82b89c2d7659dd5f6fd06dfe50a80f368c

smockle: Sidebar: Do you think CSP would block the imperative approach? If so, we could restore the original approach (in https://github.com/primer/components/pull/1287/commits/eb298ad5d7cbacfb9b26d77cdf01d54ebf420f53).

src/Overlay.tsx
smockle: Fixed in https://github.com/primer/components/commit/eb770c82b89c2d7659dd5f6fd06dfe50a80f368c

src/Overlay.tsx
```diff
@@ -40,6 +41,7 @@ const StyledOverlay = styled.div<StyledOverlayProps & SystemCommonProps & System
   min-width: 192px;
   max-width: 640px;
   height: ${props => heightMap[props.height || 'auto']};
+  max-height: ${props => props.maxHeight && heightMap[props.maxHeight]};
```
smockle, line 44: The [`styled-components` docs](https://styled-components.com/docs/advanced#tagged-template-literals) explain:
> Speaking of which, during flattening, styled-components ignores interpolations that evaluate to undefined, null, false, or an empty string (""), which means you're free to use short-circuit evaluation to conditionally add CSS rules.

I verified this in Storybook: The styles for the `Overlay` in the “Single Select” story [exclude `max-height`](https://github.com/primer/components/blob/16d5568e9cddaa5257337df1aa30207d5db4473d/src/stories/SelectPanel.stories.tsx#L103); the styles for the `Overlay` in the “SelectPanel, Height: Initial, Overflowing Items” story [include `max-height`](https://github.com/primer/components/blob/16d5568e9cddaa5257337df1aa30207d5db4473d/src/stories/SelectPanel.stories.tsx#L134), as expected.

smockle: Incidentally, I’m intentionally diverging from the [general guidance to prefer ternaries](https://github.com/primer/components/pull/1135#discussion_r614358653) here:
- because `props.maxHeight` has a fixed set of non-falsey possible values
- because both surrounding lines use logical operators for fallbacks
- because the rule is omitted in the falsey case, rather than set to a default value

src/stories/SelectPanel.stories.tsx
smockle: Lifted in https://github.com/primer/components/pull/1287/commits/16d5568e9cddaa5257337df1aa30207d5db4473d. Additional comments re: implementation details here: https://github.com/primer/components/pull/1287#discussion_r649610297

src/Overlay.tsx
```diff
@@ -10,6 +10,7 @@ import {useCombinedRefs} from './hooks/useCombinedRefs'
 type StyledOverlayProps = {
   width?: keyof typeof widthMap
   height?: keyof typeof heightMap
+  maxHeight?: keyof typeof heightMap
```
dgreif, line 13: Should we will want to exclude `auto` and `initial` on this one?

src/Overlay.tsx
dgreif: I'm good with this implementation.  I think the short circuit gotchas were very react-specific, so I'm fine with taking the easier approach if styled-components is more robust in what it accepts.

src/Overlay.tsx
```diff
@@ -81,31 +84,48 @@ export type OverlayProps = {
...
  * @param width Sets the width of the `Overlay`, pick from our set list of widths, or pass `auto` to automatically set the width based on the content of the `Overlay`. `small` corresponds to `256px`, `medium` corresponds to `320px`, `large` corresponds to `480px`, `xlarge` corresponds to `640px`, `xxlarge` corresponds to `960px`.
- * @param height Sets the height of the `Overlay`, pick from our set list of heights, or pass `auto` to automatically set the height based on the content of the `Overlay`. `xsmall` corresponds to `192px`, `small` corresponds to `256px`, `medium` corresponds to `320px`, `large` corresponds to `432px`, `xlarge` corresponds to `600px`.
+ * @param height Sets the height of the `Overlay`, pick from our set list of heights, or pass `auto` to automatically set the height based on the content of the `Overlay`, or pass `initial` to set the height based on the initial content of the `Overlay` (i.e. ignoring content changes). `xsmall` corresponds to `192px`, `small` corresponds to `256px`, `medium` corresponds to `320px`, `large` corresponds to `432px`, `xlarge` corresponds to `600px`.
```
dgreif, line 87: Should probably add `maxHeight` to the docs

src/Overlay.tsx
smockle: Excluded in https://github.com/primer/components/pull/1287/commits/010c82ef6e38c63b2406e05079ac1b9ba3860cec

src/Overlay.tsx
smockle: Documented in https://github.com/primer/components/pull/1287/commits/286c756766fdeb236c1f41da0fd250bbd02e245b

dgreif (approved):
Looks great!

PR #1288, opened 2021-06-10
fix(useFocusZone): remove suspended active descendant state

dgreif:
The `active-descendant` variant of focus zone currently supports a "suspended" state.  The zone is suspended upon first creation, and any time the user presses `esc`.  After further integration of the `SelectPanel`, it's been determined that we should activate the first descendant _immediately_, rather than requiring the user to press the down arrow to exit suspended state.  To accomplish this goal, I have completely removed the concept of a suspended state. The zone will now remember the last active descendant and activate it immediately when focus moves into the `activeDescendantControl` element.  When focus _leaves_ that element, we will keep track of the selected descendant, but "clear" the active descendant state from the control element and inform the consumer that there is no longer an active descendant.  If focus returns to the control element, the same descendant will become active again.  Further, if one of the descendant elements is _focused_, the zone will immediately move focus to the control element and set the active descendant to the element that originally received focus.  This works great for clicking on items in the `SelectPanel`

As part of this work, I was able to remove the `subscribeToActiveElementChanges` code.  This was in place with some assumptions about overlapping zones and a desire to focus descendant elements if the control element didn't have focus.  This approach was very complicated and does not appear to have any benefit in our real-world use cases so far.  I also added a `NestedZones` story to verify that regular zone behavior works as expected.  

Finally, I updated the `FilteredActionList` refs to more closely match the setup from the active descendant story.  The big difference is that the control input is no longer _in_ the container that is passed to the zone.

Closes https://github.com/github/primer/issues/184

### Screenshots

https://user-images.githubusercontent.com/3026298/121610109-20943680-ca0a-11eb-93be-1780ef52b391.mp4

https://user-images.githubusercontent.com/3026298/121610112-22f69080-ca0a-11eb-84bb-b56ec01583ac.mp4

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.changeset/smart-files-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Removed "suspended" state from active-descendant focus zones.  Active descendant will now be active as soon as the `activeDescendantControl` element is focused, and cannot be deactivated by pressing `esc`.  This is a breaking change to `useFocusZone`, but this behavior is still considred to be in `alpha`.
```
colebemis, line 5: Is there a simpler way to explain this change? As someone not familiar with the intricacies of focus zones, this description is a little hard to follow

dgreif: Updated 😄

.changeset/smart-files-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Focus zones with an `activeDescendantControl` will now activate the first descendant as soon as the control element is focused, rather than waiting for an up/down arrow press.  Descendants stay active until the control element is blured.  This is a breaking change to `useFocusZone`, but this behavior is still considred to be in `alpha`.
```
colebemis, line 5:
```suggestion
Focus zones with an `activeDescendantControl` will now activate the first descendant as soon as the control element is focused, rather than waiting for an up/down arrow press.  Descendants stay active until the control element is blurred.  This is a breaking change to `useFocusZone`, but this behavior is still considered to be in `alpha`.
```

.changeset/smart-files-promise.md
colebemis: Much clearer. Thank you ❤️

.changeset/smart-files-promise.md
dgreif: Thank you!

colebemis (approved):
Excellent work on this ✨

PR #1292, opened 2021-06-11
chore: update dependencies

dgreif:
Bumping all the base level dependencies (not docs) in hopes of clearing a bunch of dependabot PRs

* Jest had a breaking change, which changed some defaults with timers and the default environment.  I'm setting `jsdom` as our default for now, but we _could_ use `node` as the default and only use `jsdom` for the files that actually need it.  Here's the release notes for those interested: https://jestjs.io/blog/2021/05/25/jest-27
* I updated TypeScript, but there is an issue with `react-docgen-typescript` that prevents us from going all the way to the latest ts version. Looks like it is fixed and should be released soon: https://github.com/styleguidist/react-docgen-typescript/issues/356
* I _did not_ update `styled-components` because it causes all of our snapshots to fail.  It looks like v5 has some major performance improvements, so we should definitely consider updating soon.
* @primer/primitives is currently on a canary build, so we should consider updating that soon as well.  Would love input from @colebemis and @VanAnderson on why we are currently using a canary build.


colebemis (approved):
Thanks for cleaning this up, @dgreif! Let's bump `@primer/primitives` to `latest` in this PR as well 👍

colebemis:
@VanAnderson is going to run a quick test of `@primer/primitives@latest` in Memex to double-check that it's safe to upgrade

colebemis:
This snapshot diff looks odd to me:

CleanShot 2021-06-11 at 12 56 16@2x

https://github.com/primer/components/runs/2806099805#step:7:2042

I think this is happening because `prState.open.border` changed from `"rgba(0,0,0,0)"` to `"transparent"`

PR #1293, opened 2021-06-11
Restore "fix: Don’t focus first 'Item' of 'DropdownMenu' and 'SelectMenu' on open"

smockle:
Reverts primer/components#1291, restoring primer/components#1270

When [these lines](https://github.com/primer/components/blob/9fe54d65863cdfe97df770321f77233324ca4e32/src/behaviors/focusTrap.ts#L91-L93) (from primer/components#1270) removed temporary `tabIndex` from the `container`, certain browsers (e.g. Chrome) (but not others, e.g. Safari) set `body` as the `activeElement`. Because `body` is outside `FocusTrap` (and `FocusZone`), keyboard navigation between the `container`’s children was no longer possible.

By deferring the removal of the temporary `tabIndex` until the `container` has already lost focus, we avoid focusing `body`, and so we avoid breaking keyboard navigation. I added code comments in this PR to explain this slightly-more-complicated codepath and the scenario it prevents.

ℹ️ [Reviewing this commit](https://github.com/primer/components/commit/b343d47c064ee304c0147411cc4b0888152993a9) may help clarify the difference between this PR’s approach and that in primer/components#1270: https://github.com/primer/components/commit/b343d47c064ee304c0147411cc4b0888152993a9

### Screen recordings

**Safari**
safari

**Chrome**
chrome

### Checklist

- [x] Tested in Safari
- [x] Tested in Chrome

### DIFF
--- a/.changeset/good-grapes-collect.md
+++ b/.changeset/good-grapes-collect.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Restore "fix: Don’t focus first 'Item' of 'DropdownMenu' and 'SelectMenu' on open" by deferring the removal of a temporary `tabIndex` until focus moves within the container.
--- a/src/AnchoredOverlay/AnchoredOverlay.tsx
+++ b/src/AnchoredOverlay/AnchoredOverlay.tsx
@@ -1,6 +1,6 @@
 import React, {useCallback, useMemo} from 'react'
 import Overlay, {OverlayProps} from '../Overlay'
-import {useFocusTrap} from '../hooks/useFocusTrap'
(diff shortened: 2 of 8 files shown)


smockle:
> I think we could just pass `once: true` to get this same behavior. Alternatively, we could keep the abort controller, but it would probably be good to have it abort when the upper level abort controller gets a signal as well.

Oh neat, that sounds good! TIL [`once: true`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters). It has [broad browser support](https://caniuse.com/once-event-listener) too!

dgreif (approved):
Thanks for tracking this down 🏅

PR #1297, opened 2021-06-11
Upgrade primer/primitives to 4.4.0

VanAnderson:
This upgrades primer/primitives and can make some crucial design fixes.


src/__tests__/__snapshots__/StateLabel.tsx.snap
```diff
@@ -23,7 +23,7 @@ exports[`StateLabel renders children 1`] = `
   border-style: solid;
   background-color: #28a745;
   color: #ffffff;
-  border-color: rgba(0,0,0,0);
+  border-color: prState.open.border;
```
colebemis, line 26: This seems odd.

colebemis: I think this is happening because `prState.open.border` changed from "rgba(0,0,0,0)" to "transparent".

We have have a couple options:
1. Update primer/primitives to use `"rgba(0,0,0,0)"` instead of `"transparent"`
2. Update primer/components to handle `"transparent"` values

Do you have a preference?

VanAnderson: I don't have a preference - not sure what #2 involves though.

VanAnderson:
image

Also a bit concerning

src/__tests__/__snapshots__/StateLabel.tsx.snap
smockle: None of the changed lines in this PR contain `linear-gradient`, but something to be aware of—`transparent` breaks `linear-gradient` in Safari: https://github.com/github/catalyst/pull/87

colebemis: I'll update primer/primitives to use `rgba(0,0,0,0)` 👍

colebemis: https://github.com/primer/primitives/pull/142

colebemis:
> image
> 
> Also a bit concerning

Looks like this is happening on `main` too. I believe it's a result of using portals to render the action menu. Not a concern for this PR 👍

colebemis (approved):

PR #1304, opened 2021-06-14
docs: Add a story demoing height-after-fetch

smockle:
Follow-up to https://github.com/primer/components/pull/1287. From that PR’s description:
> Sidebar: The asynchronously-fetch case is partially-supported [Ed: It is—in fact—fully-supported]: A component consumer can initially (before items have been retrieved) pass `height="auto"` then (once items are retrieved) pass `height="initial"`. Because `heightKey` is a dependency of the `useEffect`, the height-after-fetch can be set using this approach.

This PR adds a story demonstrating this pattern.

https://github.com/github/primer/issues/161 described two cases:
1. A `SelectPanel` whose `Item`s are known during the initial render
2. A `SelectPanel` whose `Item`s are asynchronously fetched

This story demonstrates that the `Overlay` component supports content-matched height in both cases (regardless of whether the content is initially available or fetched), so https://github.com/github/primer/issues/161 can considered addressed. Fixes https://github.com/github/primer/issues/161.

### Merge checklist
- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/stories/SelectPanel.stories.tsx
+++ b/src/stories/SelectPanel.stories.tsx
@@ -1,3 +1,4 @@
+import type {OverlayProps} from '../Overlay'
 import {Meta} from '@storybook/react'
 import React, {useState} from 'react'
 import {theme, ThemeProvider} from '..'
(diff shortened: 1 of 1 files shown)


bscofield (approved):
🎉

PR #1305, opened 2021-06-15
Revert "chore(deps-dev): bump typescript from 4.2.2 to 4.3.2 (#1302)"

dgreif:
This reverts commit 0b8dc48f

I mentioned this in #1292, but there is currently a bug with `react-docgen-typescript` (https://github.com/styleguidist/react-docgen-typescript/issues/356) if you update TypeScript beyond `4.2.2`.  This only bug only shows up when you run storybook, which is why CI passed in #1302.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25164,9 +25164,9 @@
       }
     },
     "typescript": {
-      "version": "4.3.2",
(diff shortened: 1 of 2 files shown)


colebemis (approved):
Oops. Good catch!

PR #1308, opened 2021-06-15
feat(useFocusZone): update active-descendant on mousemove

dgreif:
This introduces some additional refinements to `active-descendant` and its use in `ActionList`.
* Active descendant will now move when there is a `mousemove` event over any "focusable" item within the container.  Practically speaking, this means mousing over items in an `ActionList` will cause the active descendant to move with the mouse.  This is more in-line with MacOS system menus, which move focus with mouse and/or keyboard
* Active descendants will now directly receive an attribute which can be used for styling them (rather than needing to apply/remove a class in the update callback).  It will be in the form of `data-is-active-descendant="activated-directly"` or `data-is-active-descendant="activated-indirectly"`.  The presence of `data-is-active-descendant` tells you that the item _is_ active, and `activated-*` gives insight into _how_ it was activated.  The only way to activate "directly" right now is to use the up/down arrows or directly focus an item (by clicking on it).  All other interactions (`mouseover`, adding/removing elements, initial zone creation) are considered actions that moved active descendant "indirectly".  Using these additional details, we can now render `ActionList` items differently depending on how users interacted with the list.
* focusZone will now add an additional `data-has-active-descendant` to the _container_.  This is helpful because the `aria-activedescendant` attribute is put on the _control_ element.
* Combining the new `data-has-active-descendant` and `:focus-within`, we are able to detect at the list level if there is an active or focused item within the list.  If so, we want to ignore `:hover` styles to avoid confusion about which element will receive events if you press `Enter`.
* Updated `SelectPanel` to have a more proper active descendant setup.  The `ActionList` is now the `container`, and the input used for controlling the list is outside of the container.  This required the addition of `forwaredRef` for the `List` component.

Closes https://github.com/github/primer/issues/199

#### SelectPanel (Active Descendant)
CleanShot 2021-06-22 at 11 37 02

#### ActionMenu (Focus)
CleanShot 2021-06-22 at 11 39 12

### Merge checklist
- [x] Added/updated tests
- [] Added/updated documentation
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/ActionList/Item.tsx
```diff
@@ -178,6 +178,8 @@ const StyledItem = styled.div<
   display: flex;
   border-radius: ${get('radii.2')};
   color: ${({variant, item}) => getItemVariant(variant, item?.disabled).color};
+  // 2 frames on a 60hz monitor
+  transition: background 133.333ms linear;
```
vdepizzol, line 182: @dgreif that should be `33.333ms` ;)

src/ActionList/Item.tsx
```diff
@@ -176,10 +178,13 @@ const StyledItem = styled.div<
...
```
vdepizzol, line 182:
```suggestion
  transition: background 33.333ms linear;
```

src/ActionList/Item.tsx
```diff
@@ -205,27 +210,55 @@ const StyledItem = styled.div<
...
+      50% {
+        transform: scale(1);
+        box-shadow: inset 0 0 0 rgba(${get('colors.scale.black')}, 0.04);
+      }
+      100% {
+        transform: scale(0.97);
+        box-shadow: inset 0 3px 9px rgba(${get('colors.scale.black')}, 0.04);
+      }
+    }
+  }
```
vdepizzol, line 260: @dgreif Unfortunately there are some rendering quirks when rendering a `transform` on delay like that. Unless there's a quick fix to avoid this 1px movement, we can just ignore it for now.

https://user-images.githubusercontent.com/293280/122989231-6bda0d80-d357-11eb-894a-84f0b5ec0899.mov

```suggestion
  &:active {
    background: ${({focusBackground}) => focusBackground};
  }
```

dgreif: I think using the same background as the active state will make this more consistent for now as well, so I'm good with reverting back to simply the background color

src/ActionList/Item.tsx
dgreif: Great catch!  I increased it to verify I could see it locally, then forgot to remove the `1` 🤦

src/behaviors/focusZone.ts
```diff
@@ -484,6 +509,23 @@ export function focusZone(container: HTMLElement, settings?: FocusZoneSettings):
         updateFocusedElement(event.target)
       }
     })
+    container.addEventListener(
+      'mousemove',
+      ({target}) => {
+        if (!(target instanceof HTMLElement)) {
```
smockle, line 515: Are we using this to intentionally exclude `SVGElement`?

src/behaviors/focusZone.ts
```diff
@@ -484,6 +509,23 @@ export function focusZone(container: HTMLElement, settings?: FocusZoneSettings):
         updateFocusedElement(event.target)
       }
     })
+    container.addEventListener(
+      'mousemove',
+      ({target}) => {
```
smockle, line 514: Should we debounce this?

src/ActionList/List.tsx
```diff
@@ -242,4 +248,6 @@ export function List(props: ListProps): JSX.Element {
       })}
     </StyledList>
   )
-}
+})
+
+List.displayName = 'ActionList'
```
smockle, line 253: Related: Should we set `Item`, `Group`, and `Header` `displayName` to `ActionList.Item`, `ActionList.Group`, and `ActionList.Header`, respectively?

src/__tests__/__snapshots__/ActionList.tsx.snap
```diff
@@ -11,6 +11,12 @@ exports[`ActionList renders consistently 1`] = `
   line-height: 20px;
 }
 
+.c0[data-has-active-descendant],
+.c0:focus-within {
+  --item-hover-bg-override: none;
+  --item-hover-divider-border-color-override: hsla(214,13%,93%,1);
```
smockle, line 17: I was surprised this is `hsla`, but it seems correct: [`List.tsx#L109`](https://github.com/primer/components/blob/cdc79e633d42f8eaeeec793046423cd09764912d/src/ActionList/List.tsx#L109) sets this variable to `colors.selectMenu.borderSecondary`, which `@primer/primitives` defines in [`light.tsx#L535`](https://github.com/primer/primitives/blob/7a436d367e5c6316a165787e7b89b789245780ac/data/colors/light.ts#L535) to be `lighten(get('scale.gray.2'), 0.03)`. According to the [SCSS docs for `lighten`](https://sass-lang.com/documentation/modules/color#lighten), that function “Increases the HSL lightness of `$color`…”.

src/behaviors/focusZone.ts
```diff
@@ -311,6 +312,11 @@ function shouldIgnoreFocusHandling(keyboardEvent: KeyboardEvent, activeElement:
   return false
 }
 
+export const isActiveDescendantAttribute = 'data-is-active-descendant'
+export const activeDescendantActivatedDirectly = 'activated-directly'
```
smockle, line 316: Would this be a good place to include a comment describing what `activated-directly` means?

src/behaviors/focusZone.ts
dgreif: Hmm, hadn't considered that.  The `focusableElements` array is of type `HTMLElement[]`, so we need to narrow the type in order to do `focusableElements.includes(target)`.  The check in this particular callback can use `instanceof Node`, but some of the other cannot be changed unless we want to switch `Element` or `Node` throughout this file, and in all of the `iterateFocusableElement` functions.

src/behaviors/focusZone.ts
dgreif: I think `mousemove` already has some built in (tho very minor) throttling built into the browser. In order to have the best user experience, I think we need to respond to every event.  With that said, there is a short circuit [here](https://github.com/primer/components/pull/1308/files#diff-7db8dd22df056db4e87f4cae691dcded5aaec5a8d8392a541cdbc33bad844e66R376-R382) that prevents these events from triggering changes all the way out to active descendant callbacks if the active descendant _has not changed_

src/ActionList/List.tsx
dgreif: I think that's a great idea.  I _had_ to do it here because of the `forwardRef`, but I think adding those others would be a great separate PR

src/behaviors/focusZone.ts
dgreif: Yeah, that's a good idea.  Added some comments in https://github.com/primer/components/pull/1308/commits/0a6057c65a41d71d7f57f711803a7d2a6853c289

smockle (approved):

PR #1309, opened 2021-06-15
Include all shadow variables in theme object

colebemis:
## Problem

We were not including all shadow variables from [Primer Primitives](https://github.com/primer/primitives) in the Primer React [theme object](https://primer.style/components/theme-reference).

## Background

In Primer Primitives, colors and shadows are stored in one object. In Primer React, we need to store colors and shadows in two separate objects. We split up the Primer Primitives object using this `partitionColors` function:

```jsx
function partitionColors(colors) {
  return {
    colors: filterObject(colors, value => isColorValue(value)),
    shadows: filterObject(colors, value => isShadowValue(value))
  }
}
```

I figured out that we were omitting some shadow variables in the theme object because `isShadowValue()` was incorrectly return `false` when passed some valid shadow values.

## Solution

This PR updates the `isShadowValue()` function to correctly handle more shadow values and adds tests to avoid future regressions.

cc @dmarcey @mattcosta7

### DIFF
--- a/.changeset/forty-books-clean.md
+++ b/.changeset/forty-books-clean.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Include all shadow variables in theme object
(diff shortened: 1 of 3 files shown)


dgreif (approved):
Thanks for adding tests!!

PR #1311, opened 2021-06-16
Expose `textInputProps` from `SelectPanel`

lukewar:
### Description

Adding option to pass `textInputProps`. This allows customizing the input field.

### Screenshots

| Before | After |
|--------|--------|
| before | after | 

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/SelectPanel/SelectPanel.tsx
```diff
@@ -129,9 +124,19 @@ export function SelectPanel({
 
   const inputRef = React.useRef<HTMLInputElement>(null)
   const focusTrapSettings = {
-    initialFocusRef: inputRef
+    initialFocusRef: inputRef,
+    disabled: textInputProps?.disabled
```
dgreif, line 128: Transferring comment from https://github.com/lukewar/components/commit/b0cacf73b28717bdd246aaa7f45ec93a19ecf514#commitcomment-52246711

> Could you elaborate on why this was needed? I don't think it will be great practice to disable the focus trap here, because that would allow focus to move outside of the SelectPanel. If you have another Overlay or Dialog you want to show in you error case, it can add its own focusTrap that will temporarily override this one.

 Response 
> Good question @dgreif. Without disabling the focus trap, the focus will be still set to the input field, even when disabled.
> 
> For example, the screenshot below has the text field disabled but it is still in focus:
> CleanShot 2021-06-16 at 15 50 48@2x
> 
> Is there a better way to go around that?

src/SelectPanel/SelectPanel.tsx
```diff
@@ -129,9 +124,19 @@ export function SelectPanel({
...
   }
 
+  const extendedTextInputProps: Partial<TextInputProps> = useMemo(() => {
+    return {
+      mx: 2,
+      my: 2,
+      contrast: true,
+      ...textInputProps
+    }
+  }, [textInputProps])
```
dgreif, line 138: Transferring comment from https://github.com/lukewar/components/commit/b0cacf73b28717bdd246aaa7f45ec93a19ecf514#r52246102

Two suggestions here:
1. We should spread the `textInputProps` _after_ the defaults, so that all of them can be overridden if needed
2. We should memoize this value with `useMemo` to avoid churn on the `FilteredActionList` props

src/SelectPanel/SelectPanel.tsx
dgreif: Is there a problem with leaving focus on the disabled input?  What behavior would you rather see in this scenario?

src/SelectPanel/SelectPanel.tsx
lukewar: Thanks for the feedback. Addressed.

src/SelectPanel/SelectPanel.tsx
lukewar: That's a very good question. Now that I thought more about that, I think that I shouldn't do that. Leaving focus and highlight on the input provides to users the visual clue for navigation, even when the input is disabled.

dgreif (approved):
Looks great, thanks for the contribution!

PR #1312, opened 2021-06-17
feat: When clicking an 'AnchoredOverlay' trigger, don’t close—then immediately reopen—the 'AnchoredOverlay'

smockle:
Fixes https://github.com/github/primer/issues/218

### Screen recordings
| Before | After |
|:---:|:---:|
| Screen recording demonstrating that clicking the trigger closes then undesirably reopens the menu | Screen recording demonstrating that clicking the trigger either closes or opens the menu, as expected |

### Merge checklist
- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/AnchoredOverlay/AnchoredOverlay.tsx
+++ b/src/AnchoredOverlay/AnchoredOverlay.tsx
@@ -46,7 +46,7 @@ interface AnchoredOverlayBaseProps extends Pick<OverlayProps, 'height' | 'width'
   /**
    * A callback which is called whenever the overlay is currently open and a "close gesture" is detected.
    */
-  onClose?: (gesture: 'click-outside' | 'escape') => unknown
+  onClose?: (gesture: 'anchor-click' | 'click-outside' | 'escape') => unknown
 
   /**
    * Props to be spread on the internal `Overlay` component.
@@ -110,7 +110,7 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({
       if (!open) {
         onOpen?.('anchor-click')
       } else {
-        onClose?.('click-outside')
+        onClose?.('anchor-click')
       }
     },
     [open, onOpen, onClose]
--- a/src/AnchoredOverlay/AnchoredOverlay.tsx
+++ b/src/AnchoredOverlay/AnchoredOverlay.tsx
@@ -104,11 +104,16 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({
   )
   const onAnchorClick = useCallback(
     (event: React.MouseEvent<HTMLElement>) => {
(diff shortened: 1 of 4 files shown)


src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -104,11 +104,16 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({
   )
   const onAnchorClick = useCallback(
     (event: React.MouseEvent<HTMLElement>) => {
-      if (!event.defaultPrevented && event.button === 0 && !open) {
+      if (event.defaultPrevented || event.button !== 0) {
```
dgreif, line 107: @dmarcey had mentioned an issue with double clicking the anchor causing the menu to immediately open and close.  I think we could avoid this by adding a simple `|| event.detail === 2` here or `|| event.detail > 1`.  Playing around with it locally, I could see users getting frustrated by triple clicking not causing any change, so `=== 2` might be the safer bet.  Thoughts?

dgreif (commented):
One small enhancement recommendation, but otherwise this looks perfect!  Thanks for knocking out this annoying bug 🔨 🐛 💥

src/AnchoredOverlay/AnchoredOverlay.tsx
smockle: I’m not adamant about this, but adding special handling for double-clicks to the library feels like a code smell to me, since double-clicking is uncommon on the web. This PR makes it possible to customize double-click behavior in consumer code—I added [a story demonstrating this](https://github.com/primer/components/blob/5feb1ce19d47689be6f772719b280f2ea4e2d166/src/stories/ActionMenu.stories.tsx#L277-L322).

dgreif: I agree that this is the right solution.  Thanks for adding a story to show exactly how to implement that on the consumer side!

src/SelectPanel/SelectPanel.tsx
```diff
@@ -76,7 +76,7 @@ export function SelectPanel({
 
   const onOpen: AnchoredOverlayProps['onOpen'] = useCallback(gesture => onOpenChange(true, gesture), [onOpenChange])
   const onClose = useCallback(
-    (gesture: 'click-outside' | 'escape' | 'selection') => {
+    (gesture: Parameters<Exclude<AnchoredOverlayProps['onClose'], undefined>>[0] | 'selection') => {
```
dgreif, line 79: Show off 😜.  That's super cool

dgreif (approved):

PR #1314, opened 2021-06-18
fix: Correct `role`s and add missing attributes

smockle:
Partial fix for https://github.com/github/primer/issues/217

- `<button>` element (from `<DropdownButton>`) should have `type="button"`
- `<Overlay>` (from `<AnchoredOverlay>`) should have `role="dialog"` or `role="menu"`, not `role="listbox"`
- `<Item>`s should have separate labels and descriptions

### Screenshot

**No visual change for `<Item descriptionVariant={"inline"} />`**
| `main` branch | This PR |
|:---:|:---:|
| main-inline | PR-inline |

**No visual change for `<Item descriptionVariant={"block"} />`**
| `main` branch | This PR |
|:---:|:---:|
| main-block | PR-block |

**Improved `label` for `menuitem` node**
| `main` branch | This PR |
|:---:|:---:|
| main-AT | PR-AT |

### Merge checklist
- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

PR #1318, opened 2021-06-23
fix: Type 'DropdownButton' as 'button'

smockle:
Split from #1314

`<button>` elements in `<form>` contexts get `type="submit"` [by default](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), but clicking a `<DropdownButton>` should never submit a `<form>`, so this PR passes `type="button"`, which indicates “client-side scripts listen to the element's events”.

This PR has no visual changes.


dgreif (approved):
Great catch!

PR #1320, opened 2021-06-23
fix: Split `<Item>` labels

smockle:
Split from #1314

This PR splits `<Item>` labels (into a label + a description) to decrease verbosity and improve clarity.

### Screenshot

**No visual change for `<Item descriptionVariant={"inline"} />`**
| `main` branch | This PR |
|:---:|:---:|
| main-inline | PR-inline |

**No visual change for `<Item descriptionVariant={"block"} />`**
| `main` branch | This PR |
|:---:|:---:|
| main-block | PR-block |

**Improved `label` for `<Item>`**
| `main` branch | This PR |
|:---:|:---:|
| main-AT | PR-AT |

### Merge checklist
- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/ActionList/Item.tsx
```diff
@@ -151,9 +151,13 @@ const getItemVariant = (variant = 'default', disabled?: boolean) => {
...
 }
 
-const StyledItemContent = styled.div`
+const StyledItemContent = styled.div<{
+  descriptionVariant: ItemProps['descriptionVariant']
+}>`
+  align-items: baseline;
   display: flex;
   min-width: 0;
+  flex-direction: ${({descriptionVariant}) => (descriptionVariant === 'inline' ? 'row' : 'column')};
```
dgreif, line 160: Want to try changing this to a css variable?  I do think it will lead to much better performance if we do it consistently throughout the components, and this seems like a good place to start.  Let me know if you want to pair on it

src/ActionList/Item.tsx
```diff
@@ -310,10 +307,12 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
     onKeyPress,
     children,
     onClick,
-    id,
+    id: _id,
     ...props
   } = itemProps
 
+  const id = useMemo(() => _id ?? uniqueId(), [_id])
```
dgreif, line 314: I'm not sure we can rely on `id` being unique here.  For example, if I have a list of cats and a list of dogs, they could have overlapping _data_ ids that are not unique at a global level, but are unique within the list.  We should probably just do `uniqueId` in all cases.

src/ActionList/Item.tsx
```diff
@@ -393,35 +394,31 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
...
+      {(TrailingIcon || trailingText) && (
+        <TrailingVisualContainer variant={variant} disabled={disabled}>
+          {trailingText && <div>{trailingText}</div>}
+          {TrailingIcon && (
+            <div>
+              <TrailingIcon />
+            </div>
+          )}
+        </TrailingVisualContainer>
+      )}
```
dgreif, line 421: I think moving this outside of `StyledItemContent` will cause the item divider lines to end after the text/description.  Is that desired?

src/ActionList/Item.tsx
```diff
@@ -358,6 +357,8 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
       variant={variant}
       showDivider={showDivider}
       aria-selected={selected}
+      aria-labelledby={text ? `${id}-label` : undefined}
+      aria-describedby={description ? `${id}-description` : undefined}
```
dgreif, line 361: Would probably be good to throw `${id}-label` and `${id}-description` into variables to avoid accidental changes in only one of the two spot

src/ActionList/Item.tsx
```diff
@@ -358,6 +357,8 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
       variant={variant}
       showDivider={showDivider}
       aria-selected={selected}
+      aria-labelledby={text ? `${id}-label` : undefined}
```
dgreif, line 360: Does this override the default speech for the checkboxes in SelectPanel?  If not, we may want to consider removing the `aria-label` on the `MultiSelectInput` below, which I believe is what causes the doubling effect we noticed last week.

dgreif (changes requested):
Really appreciate you taking the time to dive into each of these components and find accessibility solutions!

src/ActionList/Item.tsx
smockle: Done: https://github.com/primer/components/pull/1320/commits/254a4a3a7092f00b1eafce450a508b6b758006d8

src/ActionList/Item.tsx
smockle: Good point! I mistakenly thought `id` was being applied as the `id` attribute, which is supposed to be globally-unique, as opposed to the `data-id`. Revised in https://github.com/primer/components/pull/1320/commits/254a4a3a7092f00b1eafce450a508b6b758006d8

src/ActionList/Item.tsx
smockle: Good catch! That is not desired. 👍

smockle: Fixed in https://github.com/primer/components/pull/1320/commits/408ff2473c40a867b41c8bbedc1cf9499fc677ae, but that commit may have introduced other issues. I haven’t found any regressions, but I’d appreciate more eyes on it.

src/ActionList/Item.tsx
smockle: Great question! I’d like to follow-up on this in a separate issue/PR. Earlier this week, @ichelsea, @JoseInTheArena, and I discussed using the [`menuitemradio`](https://www.w3.org/TR/wai-aria-1.2/#menuitemradio) and [`menuitemcheckbox`](https://www.w3.org/TR/wai-aria-1.2/#menuitemcheckbox) roles for single-select and multi-select `Item`s, respectively.

Demonstrations of this pattern in WAI-ARIA Authoring Practices 1.2 such as [“Editor Menubar Example”](https://w3c.github.io/aria-practices/examples/menubar/menubar-editor.html) do not include separate `<input type="checkbox">`, so we might want to remove `<MultiSelectInput>` from the Accessibility Tree (e.g. using `aria-hidden="true"`; ref [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute#deciding_between_aria-hiddentrue_rolepresentation_and_rolenone)). Similarly, [“Example 4”](https://www.w3.org/TR/wai-aria-1.2/#example-4) includes a purely decorative checkbox image with `role="presentation"` inside a `menuitemcheckbox`.

smockle: Opened https://github.com/github/primer/issues/228 to track this

src/ActionList/Item.tsx
smockle: 👍 Migrated in https://github.com/primer/components/pull/1320/commits/73eed06a2f81cd18e657f0a0c15f62cbde924555

src/ActionList/Item.tsx
dgreif: Sounds like a great plan, thanks for opening an issue to track it!

dgreif (approved):
Thanks for all the tweaks, this looks like a solid change and should hopefully make all of our layered components more accessible!

src/ActionList/Item.tsx
```diff
@@ -248,27 +259,16 @@ const StyledItem = styled.div<
   ${sx}
 `
 
-export const TextContainer = styled.div<{
+export const TextContainer = styled.span<{
   dangerouslySetInnerHtml?: React.DOMAttributes<HTMLDivElement>['dangerouslySetInnerHTML']
-  descriptionVariant: ItemProps['descriptionVariant']
```
VanAnderson, line 253: This will require an adjustment on the memex side.

PR #1322, opened 2021-06-23
feat(Overlay): slide away from anchor based on position

### LINKED ISSUE #1231: Overlay Animation Causes Overflow
**Describe the bug**
All components built with `Overlay` can cause their parent container to overflow and show scrollbars for a second while the `Overlay` performs a slide-up animation after it is opened.  This happens when the overlay is placed such that it can _fit_ between its anchor and the bottom of the view port, but the slide-up animation cause it to transition from a placement that is _outside_ the dimensions of the viewport.

**Proposed Solution**
We could account for the distance of the slide-up animation when calculating the placement of the overlay.  If it won't fit below the anchor if you consider overlay height + transition height, then it needs to be placed elsewhere. 

Alternatively, we could still place the overlay below the anchor, but shorten or skip the slide-up if there isn't room for it.

Open to other suggestions @vdepizzol 

**To Reproduce**
1. Go to the DropdownMenu story in storybook
2. Adjust the Actions panel height so that there is _just_ enough room for the DropdownMenu to open below the button
3. Open the DropdownMenu
4. Note the scroll bar flicker in and out on the right

**Expected behavior**
Overlay animation should not cause temporary overflow

**Screenshots**
CleanShot 2021-05-17 at 09 30 17

dgreif:
The main goal here is to prevent scrollbars from showing briefly when an `AnchoredOverlay` is opened.  This can happen if the `Overlay` portion ends up slightly outside the currently visible area upon creation, or because of the slide animation.  There are two main changes to correct this behavior

* Pass `anchorSide` from `useAnchoredPosition` all the way into `Overlay`.  With this information, we can infer which direction the `Overlay` should "slide" when it is opened.  We always want to to start slightly overlapping the anchor and then move into position.  Previous behavior had it _always_ sliding "up", which is only desired if the `Overlay` is positioned above its anchor.  If not passed an `anchorSide`, there will now be _no_ slide animation.
* Remove the `visibility` setting from `Overlay`.  This was a bandaid to handle scenarios where the `Overlay` contents would briefly flash before moving into the correct position.  While setting `visibility` removed the main visible flash, it did not prevent scrollbars from flashing if the unpositioned content happened to overflow the current frame.  The _more correct_ solution here is to replace a few `useEffect` with `useLayoutEffect`, which allows `AnchoredOverlay` to create the `Overlay`, calculate position, and re-position the `Overlay` before the browser paints.  

Closes #1231

### Screenshots
| Side | Animation |
|--------|--------|
| Bottom | CleanShot 2021-06-23 at 16 01 39 |
| Right | CleanShot 2021-06-23 at 16 03 03 |
| Top | CleanShot 2021-06-23 at 16 02 19 |
| Left | CleanShot 2021-06-23 at 16 00 02 | 
| No Side Provided | CleanShot 2021-06-23 at 16 07 52 |

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


smockle:
From @dgreif in https://github.com/primer/components/pull/1322#issue-676677163:
> The main goal here is to prevent scrollbars from showing briefly when an `AnchoredOverlay` is opened.

To clarify: This PR prevents the *page* from flashing scrollbars. The scrollbar flash added via https://github.com/primer/components/pull/1276 to `SelectPanel` is still desired and has been retained.

src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -154,7 +161,6 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({
           onEscape={onEscape}
           ref={updateOverlayRef}
           role="listbox"
-          visibility={position ? 'visible' : 'hidden'}
```
smockle, line 157: `visibility='hidden'` [removes an element from the Accessibility Tree](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns), so I was initially concerned that removing the `visibility` prop would cause a hidden `Overlay` to remain in the AT. 

However, IIUC, this PR dispenses with the concept of “a hidden `Overlay`” (instead [rendering `null`](https://github.com/primer/components/blob/8374dd5b5995a73beab9b953a54b8eb25b292322/src/AnchoredOverlay/AnchoredOverlay.tsx#L171)).

src/Overlay.tsx
```diff
@@ -33,6 +34,20 @@ const widthMap = {
   auto: 'auto'
 }
 
+function getSlideAnimationFactors(anchorSide?: AnchorSide): {x: number; y: number} {
```
smockle, line 37: Nit: Is “animation factors” a widely-used/understood phrase? If not, is there a more descriptive name for these values?

src/Overlay.tsx
```diff
@@ -128,7 +155,15 @@ const Overlay = React.forwardRef<HTMLDivElement, OverlayProps>(
           height={height}
           {...rest}
           ref={combinedRef}
-          visibility={visibility}
+          style={
```
smockle, line 158: Cool use of the [CSS variables + `styled-components` combo](https://www.joshwcomeau.com/css/styled-components/) we recently discussed! ❤️

src/Overlay.tsx
```diff
@@ -128,7 +155,15 @@ const Overlay = React.forwardRef<HTMLDivElement, OverlayProps>(
...
+            {
+              '--slide-animation-x-factor': slideAnimationFactors.x,
+              '--slide-animation-y-factor': slideAnimationFactors.y,
+              '--slide-animation-timing':
+                slideAnimationFactors.x === 0 && slideAnimationFactors.y === 0 ? '0' : undefined // 0 if we don't want to animate, fallback in css if we do
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            } as any
```
smockle, line 165:
```suggestion
            } as React.CSSProperties
```

dgreif: Nice!  It didn't like the unknown keys without a type cast, but it appears to handle it fine if I explicitly cast it as you suggested.  This is specifically called out in the react docs, but they didn't have a suggested approach like this one (tho I think they should): https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d0cd19269f13cffe248bc1439ad6ae959d70524c/types/react/index.d.ts#L1543-L1552 ->  https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors

src/Overlay.tsx
dgreif: Not at all common, naming is just hard 😄.  Happy to discuss alternatives

dgreif: Updated in https://github.com/primer/components/pull/1322/commits/db69e072593a12c20aece971e81a98840b0448f3

smockle (approved):

PR #1325, opened 2021-06-25
Remove system props

VanAnderson:
Closes https://github.com/github/primer/issues/225

Removing system props from most components, updating docs and creating a codemod.

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/Button/ButtonBase.tsx
```diff
@@ -1,11 +1,10 @@
 import styled from 'styled-components'
 import {compose, fontSize, FontSizeProps, variant} from 'styled-system'
-import {COMMON, LAYOUT, SystemCommonProps, SystemLayoutProps} from '../constants'
 import {ComponentProps} from '../utils/types'
 import buttonBaseStyles from './ButtonStyles'
 
-export const buttonSystemProps = compose(fontSize, COMMON, LAYOUT)
-export type ButtonSystemProps = FontSizeProps & SystemCommonProps & SystemLayoutProps
+export const buttonSystemProps = compose(fontSize)
```
colebemis, line 6: `fontSize` is also a system prop. Let's remove it too.

src/Dialog/Dialog.tsx
```diff
@@ -241,60 +245,62 @@ const DefaultFooter: React.FC<DialogProps> = ({footerButtons}) => {
...
+            </StyledDialog>
+          </Backdrop>
+        </Portal>
+      </>
+    )
+  }
+)
 _Dialog.displayName = 'Dialog'
 
 const Header = styled(Box).attrs({as: 'header'})`
```
colebemis, line 306: Let's avoid extend `Box` now that it accepts all system props. That could create components that accept many more props than we expect.

src/__tests__/Avatar.tsx
```diff
@@ -5,11 +5,10 @@ import {px, render, behavesAsComponent, checkExports} from '../utils/testing'
 import {render as HTMLRender, cleanup} from '@testing-library/react'
 import {axe, toHaveNoViolations} from 'jest-axe'
 import 'babel-polyfill'
-import {COMMON} from '../constants'
 expect.extend(toHaveNoViolations)
 
 describe('Avatar', () => {
-  behavesAsComponent({Component: Avatar, systemPropArray: [COMMON]})
+  behavesAsComponent({Component: Avatar, systemPropArray: []})
```
colebemis, line 11: Do we need to pass an array to `systemPropArray` if it's empty?
```suggestion
  behavesAsComponent({Component: Avatar})
```

src/__tests__/Dialog.tsx
```diff
@@ -10,7 +9,7 @@ expect.extend(toHaveNoViolations)
 const comp = (
   <Dialog isOpen onDismiss={() => null} aria-labelledby="header">
     <Dialog.Header id="header">Title</Dialog.Header>
-    <Box p={3}>
+    <Box sx={{p: 3}}>
```
colebemis, line 12: We don't need to make these changes to Box since it's the only component that still accepts system props.

src/__tests__/Avatar.tsx
VanAnderson: had to slightly tweak the typing of behavesAsComponent, but with an easy fix this was possible 👍

VanAnderson:
@colebemis rebased this to get this prepped and cleaned up - there are a couple of things that need looking into.

1) we need a pattern to replace default props here that were previously system props. I'm not sure if we can easily specify default sx values that will not get overridden when another sx prop is passed in. Things like labels relied on default values for system props.

2) still not quite sure what the interaction is between https://github.com/primer/doctocat/pull/240 and this PR - basing on the canary branch for https://github.com/primer/doctocat/pull/240 did not seem to fix the issues with the doc chroma.

VanAnderson:
> @colebemis rebased this to get this prepped and cleaned up - there are a couple of things that need looking into.
> 
> 1. we need a pattern to replace default props here that were previously system props. I'm not sure if we can easily specify default sx values that will not get overridden when another sx prop is passed in. Things like labels relied on default values for system props.
> 2. still not quite sure what the interaction is between [Deprecate system props doctocat#240](https://github.com/primer/doctocat/pull/240) and this PR - basing on the canary branch for [Deprecate system props doctocat#240](https://github.com/primer/doctocat/pull/240) did not seem to fix the issues with the doc chroma.

trying to accomplish 2 in https://github.com/primer/components/pull/1325/commits/8364acfa13526cea1ea846e8de04595b3fe31127 - not sure why this approach doesn't work. The test should pass when this is fixed!

colebemis:
Closing this because we're removing system props incrementally

PR #1330, opened 2021-07-14
[experiment] Pass a typed theme to styled-components and sx

mattcosta7:
fixes https://github.com/github/memex/issues/4145

WIP - open to suggestions for improving/modifying this to cover as many use cases as possible
- still have a bunch of type related work to do here - this might be easier to accomplish by removing `@types/styled-system__css` entirely and using a local version that extends it

This will probably necessitate a major version, since anyone incorrectly accessing types will go from building without errors to build-time errors

Some friction occurs when consuming Primer components and overriding styles with both the `sxProp` and extending styles via styled-components.

In both of these cases the 'Theme' is not properly typed, leading to potential issues of accessing properties incorrectly in consumers. 

The generated types for things like

```
<Flex sx={{ color: theme => ..... }} /> // theme is inferred as 'any'
```

```
// theme is inferred as 'any' - even with a local project override of sc's 'DefaultTheme' interface, since the 
// generated type for Flex is StyledComponent<"div", any, ...Props, never>; where the second Type Arg is the theme type
const StyledFlex = styled(Flex)`
   ${props => props.theme....} 
`
```

We've had a few issues related to this in a project where properties are incorrectly accessed on the theme leading to rendering issues, and/or styling not being applied properly

Describe your changes here.

Override sx prop types with a themed version of these.
provide a default theme to styled-components

### Screenshots

### Merge checklist
- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/ThemeProvider.tsx
```diff
@@ -146,17 +146,18 @@ function chooseColorScheme(colorMode: ColorMode, dayScheme: string, nightScheme:
   }
 }
 
-function applyColorScheme(theme: Theme, colorScheme: string) {
+function applyColorScheme(theme: ThemePrevalType, colorScheme: string): Theme {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
   if (!theme.colorSchemes) {
-    return theme
+    return theme as Theme
   }
```
mattcosta7, line 153: Figuring out the case here - is this a real scenario? (does theme-preval copy a colorscheme by default - still discovering)

src/ThemeProvider.tsx
```diff
@@ -2,13 +2,13 @@ import React from 'react'
...
 import deepmerge from 'deepmerge'
+import {ThemePrevalType} from './theme-preval'
 
 const defaultColorMode = 'day'
 const defaultDayScheme = 'light'
 const defaultNightScheme = 'dark'
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Theme = {[key: string]: any}
+export type Theme = ThemePrevalType & Pick<ThemePrevalType['colorSchemes']['light'], 'colors' | 'shadows'>
```
mattcosta7, line 11: A first pass at overriding this type

src/ThemeProvider.tsx
```diff
@@ -146,17 +146,18 @@ function chooseColorScheme(colorMode: ColorMode, dayScheme: string, nightScheme:
...
+    return theme as Theme
   }
 
   if (!theme.colorSchemes[colorScheme]) {
     // eslint-disable-next-line no-console
     console.error(`\`${colorScheme}\` scheme not defined in \`theme.colorSchemes\``)
 
     // Apply the first defined color scheme
-    const defaultColorScheme = Object.keys(theme.colorSchemes)[0]
+    const defaultColorScheme = Object.keys(theme.colorSchemes)[0] as keyof ThemePrevalType['colorSchemes']
```
mattcosta7, line 160: just to make the type access at theme.colorSchemes[defaultColorScheme] happy - but should improve this access

src/sx.ts
```diff
@@ -1,9 +1,61 @@
...
+  | SystemCssProperties
+  | CSSPseudoSelectorProps
+  | CSSSelectorObject
+  | VariantProperty
+  | UseThemeFunction
+  | EmotionLabel
+  | null
+  | undefined
+
+/* #endregion  */
```
mattcosta7, line 61: I'd prefer to be overriding this in `node_modules/@types/styled-system__css` but for some reason local overrides in `@types/....` don't seem to be picked up - will explore that more

src/theme-preval.js
```diff
@@ -87,6 +88,44 @@ const darkDimmedColors = omitScale(darkDimmed.colors)
...
+ * @property {{normal: string; mono: string}} fonts
+ * @property {ReadonlyArray<string>} fontSizes
+ * @property {Record<'light' | 'normal' | 'semibold' | 'bold', number>} fontWeights
+ * @property {Record<'condensedUltra' | 'condensed' | 'default', number>} lineHeights
+ * @property {ReadonlyArray<string>} radii
+ * @property {Record<'small' | 'medium' | 'large' | 'xlarge', string>} sizes
+ * @property {ReadonlyArray<string>} space
+ * @property {{
+ *   light: {
+ *      colors: typeof lightColors,
```
mattcosta7, line 108: are these correctly Partials of the color from primitives - maybe there's a more correct typing I can apply to colors and shadows here

src/utils/shadows-types.ts
```diff
@@ -0,0 +1,58 @@
+export interface Shadows {
```
mattcosta7, line 1: getting these from primitives would be preferred, but since we process them here that doesn't seem practical

src/utils/colors-types.ts
```diff
@@ -0,0 +1,694 @@
+export interface Colors {
```
mattcosta7, line 1: getting these from primitives would be preferred, but since we process them here that doesn't seem practical

src/__tests__/Heading.tsx
```diff
@@ -18,7 +19,7 @@ const theme = {
     mono: 'Consolas,monospace'
   },
   lineHeights: {
-    normal: 1.5,
+    default: 1.5,
```
mattcosta7, line 22: this test was using a lineheight that  isn't in the theme spec - typings catch this now :-D

mattcosta7:
A few tests were updated to cast a local partial theme to a full theme type - it might be better to have a complete theme and pass it between tests instead of making a new partial one for each test that passes a theme directly?

src/theme-preval.js
```diff
@@ -58,35 +59,73 @@ const darkDimmed = partitionColors(primitives.colors['dark_dimmed'])
...
- * @type Partial<typeof primitives.colors.dark_dimmed>
+ * @type {Omit<typeof darkDimmed['shadows'], 'scale'>}
  */
 const darkDimmedShadows = omitScale(darkDimmed.shadows)
 
+/**
+ * @todo are these types all correct - especially the partial colors from
+ * primitives?  if not - let's tighten that up!
+ *
+ * @typedef {Object} ThemePrevalType
```
mattcosta7, line 95: It would be awesome if we just had this - 

It might be valuable for me to just take the theme object from primer.style theme reference and generate types directly from that via a json -> type converter (like I did for shadows/colors paritionig), and just import('....') that type here instead of defining it in this way

PR #1332, opened 2021-07-17
Improve treeshaking by setting package.json sideEffects

### LINKED ISSUE #1331: Support treeshaking via sideEffects in package.json
**Describe the bug**

When consuming a single `@primer/components` component, the entire set of components is included in the build (even when using lib-esm), since the package.json does not declare the package as free of `sideEffects` via `sideEffects: false` or `sideEffects: ['file 1', 'file 2']`

Deep importing like

`import {Button} from '@primer/components/lib-esm/components/Button'` seems to work properly, however if any import to `import {} from '@primer/components'` is made, then eveything exported from `@primer/components` is included in the build

**To Reproduce**
1. Open a new project
2. Add `@primer/components` as a dependency
3. Import a single component like `import {Button} from '@primer/components'`
4. Generate and analyze a bundle, note that all components are included in bundle

**Expected behavior**

Only the used components are included in the build

**Screenshots**

My project, bundled with webpack 4 emits a bundle containing this:

Screen Shot 2021-07-15 at 10 43 21 AM

Which includes components like <Timeline/> <StateLabel> <Pagination /> etc which are not used in my project.

Updating the node_modules package.json for `@primer/components` to include `sideEffects: false` and rebuilding with the same command leads to a bundle containg this

Screen Shot 2021-07-15 at 10 34 54 AM

Which removes unused components from my project, leading to a large reduction in bundle size

**Desktop (please complete the following information):**
 - N/A

**Smartphone (please complete the following information):**
- N/A
-

mattcosta7:
Adds a `sideEffects` declaration in package.json to ensure that bundlers can effectively treeshake unused code.

This is especially useful for treeshaking files that simply re-export other files, such as the index.ts entries. Without this change, everything exported in this way is included in the final minified build, since bundlers cannot determine whether it's safe to treeshake those re-exports

2 files seems to contain side effects, related to initializing a polyfill, in focusTrap and focusZone. I added in line comments to this pr denoting those areas

Closes #1331 

### Merge checklist
- [ ] Added/updated tests (N/A)
- [ ] Added/updated documentation (N/A)
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

I tested this by installing in my project, making these changes locally and then generating a tar from this project and consuming that in my project, verifying that everything seems to work as intended.  I don't consume every component though, so additional testing with a canary build would be best to ensure compatibility across the board

Analyzed outputs from my consuming project using `source-map-explorer`

Before:

Screen Shot 2021-07-17 at 10 27 32 AM

After: 

Screen Shot 2021-07-17 at 10 26 29 AM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@primer/components",
-  "version": "28.2.3",
+  "version": "28.2.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "module": "lib-esm/index.js",
   "typings": "lib/index.d.ts",
+  "sideEffects": [],
   "scripts": {
     "start": "cd docs && npm run develop",
     "predist": "rm -rf dist & rm -rf lib & rm -rf lib-esm",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "module": "lib-esm/index.js",
   "typings": "lib/index.d.ts",
   "sideEffects": [
-    "(lib|lib-esm)/BaseStyles.js",
     "(lib|lib-esm)/behaviors/focusZone.js",
     "(lib|lib-esm)/behaviors/focusTrap.js"
   ],
--- a/package.json
+++ b/package.json
@@ -6,12 +6,9 @@
   "module": "lib-esm/index.js",
   "typings": "lib/index.d.ts",
   "sideEffects": [
-    "lib/BaseStyles.js",
-    "lib-esm/BaseStyles.js",
-    "lib/behaviors/focusZone.js",
-    "lib-esm/behaviors/focusZone.js",
-    "lib/behaviors/focusTrap.js",
-    "lib-esm/behaviors/focusTrap.js"
+    "(lib|lib-esm)/BaseStyles.js",
+    "(lib|lib-esm)/behaviors/focusZone.js",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -5,6 +5,10 @@
   "main": "lib/index.js",
   "module": "lib-esm/index.js",
   "typings": "lib/index.d.ts",
+  "sideEffects": [
+    "(lib|lib-esm)/behaviors/focusZone.js",
+    "(lib|lib-esm)/behaviors/focusTrap.js"
```
mattcosta7, line 10: Screen Shot 2021-07-17 at 10 31 19 AM

This polyfill call is the side effect

package.json
```diff
@@ -5,6 +5,10 @@
   "main": "lib/index.js",
   "module": "lib-esm/index.js",
   "typings": "lib/index.d.ts",
+  "sideEffects": [
+    "(lib|lib-esm)/behaviors/focusZone.js",
```
mattcosta7, line 9: Screen Shot 2021-07-17 at 10 31 13 AM

This polyfill call is the side effect

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@primer/components",
-  "version": "28.2.3",
+  "version": "28.2.5",
```
mattcosta7, line 3: this just pulled the version number forward to the package-lock, since it appears it wasn't updated when the last version numbers were bumped

colebemis (approved):
Thank you, @mattcosta7 ❤️

PR #1333, opened 2021-07-20
fix(AnchedOverlay): hide overlay while positioning

dgreif:
As noted in https://github.com/github/primer/issues/240, the updates from #1322 caused the `AnchoredOverlay` components to start positioning erratically.  After further investigation, I found that they were actually being position at the _bottom of the page_ (wherever the `Portal` is located) and rendered to determine their size before positioning.  Because #1322 removed the `visibility: hidden` style, they were both visible and _focusable_ before actually being positioned.  The focus trap would activate, causing the page to scroll to the bottom, and then the position would be calculated from the anchor.  This caused the menus in full page apps to be off significantly.

Here I'm bringing back the `visibility` property on `Overlay` to allow for temporarily hiding `Overlays` before positioning.  I've tested this in the new Projects Beta and it's working as expected.

I also found a small bug with `SelectPanel`s when they transitioned from a `loading: true` to `loading: false` state.  Because the `FilteredActionList` doesn't render the `List` element until loading is done, the `listContainerRef` was null when the `useFocusZone` call was made.  We need to ensure `useFocusZone` re-checks the ref when `loading` changes, so I added it to the `dependencies` param.

Closes https://github.com/github/primer/issues/240

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/Overlay.tsx
```diff
@@ -71,7 +72,7 @@ const StyledOverlay = styled.div<StyledOverlayProps & SystemCommonProps & System
       opacity: 1;
     }
   }
-
+  visibility: ${props => props.visibility || 'visible'};
```
smockle, line 75: Non-blocking: Would it be better to use a CSS variable for this (ref https://www.joshwcomeau.com/css/styled-components/#css-variables)?

smockle (approved):
Nice job tracking this down! Left a non-blocking comment above.

Non-blocking: Can we add a test for this? Either automatic, or a tall tale in Storybook for manual testing?

src/Overlay.tsx
dgreif: Great catch!  After a month with my head out of PRC, this didn't occur to me but is definitely a pattern I want to keep bringing in wherever possible. Fixed in 54d8783

dgreif:
Added a test with the anchored overlay open, but I'm not sure I can recreate the scroll issues in JSDOM.  Added a story as well for manual testing.

PR #1336, opened 2021-07-21
Deprecate system props, but don't remove system props

VanAnderson:
This PR deprecates system props in all components except `Box`, but does not actually remove them. A codemod with documentation is provided to enable users to easily make the transition.

### Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


docs/content/Breadcrumbs.md
```diff
@@ -27,6 +27,8 @@ This ensures that the NavLink gets `activeClassName='selected'`
 
 ## System props
 
+**Note: System props for components other than `Box` are deprecated and will be removed soon.**
```
VanAnderson, line 30: This will be easy to find and replace across the codebase if we want to teak the message a bit.

VanAnderson:
Much more tweaking on the codemod here, we should be to the place where running it on a production codebase like memex will only take about 15 minutes give or take of post codemod tweaking to get things right. 👍

colebemis (approved):
✨

PR #1337, opened 2021-07-22
Fix codespace setup

max:
- Use `node` not `root` as the user to avoid permission errors
- Install useful extensions for prettier and eslint


smockle (approved):
💖 Thanks for updating this!

PR #1347, opened 2021-07-27
Remove Playroom

colebemis:
We haven't been using [Playroom](https://github.com/seek-oss/playroom) because we couldn't figure out how to deploy it. Let's remove it for now. We can reinstall Playroom if there's a high demand in the future.


VanAnderson (approved):
👍

PR #1349, opened 2021-07-27
Deploy with GitHub actions instead of Vercel integration

colebemis:
Copied from https://github.com/primer/css/pull/1303:

> This PR disables the `vercel-bot` auto deployment and uses this Vercel action instead https://github.com/marketplace/actions/vercel-action
>
> The reasoning behind this is more control on when we deploy. Here's the new logic:
>
> * **Preview deploys** Preview deploys deploy to preview Vercel URLs whenever a push happens on any branch except `main`.
> * **Production deploys** Production deploys deploy to primer.style/components whenever a push happens only on the `main` branch. The workflow also checks the changesets to see if this is a new version release or if there's a release being worked on. When there's currently a release being worked on and hasn't been published yet, the script doesn't deploy.

### DIFF
--- a/.github/workflows/deploy_preview.yml
+++ b/.github/workflows/deploy_preview.yml
@@ -1,0 +1,46 @@
+name: Deploy
+on:
+  push:
+    branches-ignore:
+      - 'main'
+      - 'dependabot/**'
+jobs:
+  deploy-preview:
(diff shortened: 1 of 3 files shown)


jonrohan (approved):
Looks worth trying 👍🏻

PR #1355, opened 2021-07-27
Replace octoface icon with spinner in SelectMenu component

colebemis:
We've deprecated the use of the `octoface` icon in favor of the [Spinner](https://primer.style/components/spinner) component.

| Before | After |
| --- | --- |
| image | image |


ashygee (approved):
:shipit:

PR #1378, opened 2021-08-11
Single confirm dialog

stephenotalora:
Adds an alert dialog component to account for cases where 1 or no buttons in a dialog are needed

### Screenshots

**Before:**
light | dark
------------ | -------------
|Screen Shot 2021-08-12 at 5 54 46 PM|Screen Shot 2021-08-12 at 5 50 44 PM|

**After:**
light | dark
------------ | -------------
|Screen Shot 2021-08-11 at 7 16 41 PM|Screen Shot 2021-08-12 at 4 59 09 PM|

Please note ☝🏽   is the desired design.

CC/@cmwinters 

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/neat-coins-mix.md
+++ b/.changeset/neat-coins-mix.md
@@ -1,5 +1,5 @@
 ---
-'@primer/components': patch
+'@primer/components': minor
 ---
 
-ConfirmationDialog can now render with one button, by passing a falsy value to `cancelButtonContent`.
+adding `AlertDialog` to account for cases where only one button is needed.
--- a/.changeset/neat-coins-mix.md
+++ b/.changeset/neat-coins-mix.md
@@ -2,4 +2,4 @@
 '@primer/components': minor
 ---
 
-adding `AlertDialog` to account for cases where only one button is needed in a dialog.
+add `AlertDialog` component for cases where only one button is needed in a dialog.
--- a/.changeset/neat-coins-mix.md
+++ b/.changeset/neat-coins-mix.md
@@ -2,4 +2,4 @@
 '@primer/components': minor
 ---
 
-adding `AlertDialog` to account for cases where only one button is needed.
+adding `AlertDialog` to account for cases where only one button is needed in a dialog.
--- a/.changeset/neat-coins-mix.md
+++ b/.changeset/neat-coins-mix.md
@@ -1,0 +1,5 @@
+---
+'@primer/components': patch
+---
+
+ConfirmationDialog can now render with one button, by passing a falsy value to `cancelButtonContent`.
--- a/docs/content/Dialog2.mdx
+++ b/docs/content/Dialog2.mdx
@@ -158,6 +158,27 @@ render(<ShorthandExample2 />)
 </State>
 ```
 
+### Example (using `ConfirmationDialog` component with one confirm option)
+
+```jsx live
+<State default={false}>
+  {([isOpen, setIsOpen]) => {
+    const openDialog = React.useCallback(() => setIsOpen(true), [setIsOpen])
+    const closeDialog = React.useCallback(() => setIsOpen(false), [setIsOpen])
+    return (
+      <>
+        <Button onClick={openDialog}>Open</Button>
+        {isOpen && (
+          <ConfirmationDialog title="Confirm action?" onClose={closeDialog} cancelButtonContent={null}>
+            Are you sure you want to confirm this action?
+          </ConfirmationDialog>
+        )}
+      </>
+    )
+  }}
+</State>
+```
+
 ### ConfirmationDialogProps
 
 | Prop name            | Type                                                                  | Default    | Description                                                                                                                   |
--- a/docs/content/Dialog2.mdx
+++ b/docs/content/Dialog2.mdx
@@ -180,38 +180,39 @@ render(<ShorthandExample2 />)
 
 ## AlertDialog
 
-An `AlertDialog` is a special kind of dialog with more rigid behavior. It is used to alert or confirm a single user action. AlertDialogs can only have at most one button: no buttons or one to confirm the alert dialog. No custom rendering capabilities are provided for AlertDialog.
+AlertDialog is a special kind of dialog with an urgent interruption, and informs the user about a situation.
+They summarize a **single** decision that requires acknowledgement and can only be dismissed by the alert close button or `escape` key.
+Use a clear question or statement for the alert title with an explanation in the content area, such as "Erase Disk storage?".
+Avoid apologies, ambiguity, or questions, such as “Warning!” or “Are you sure?”
+No custom rendering capabilities are provided for AlertDialog.
 
 ### useAlert hook
 
-An alternate way to use `AlertDialog` is with the `useAlert()` hook. It takes no parameters and returns an `async` function, `alert`. When `alert` is called (see AlertOptions below for its argument), it shows the alert dialog. When the dialog is dismissed, a promise is resolved with `true` or `false` depending on whether or not the confirm button was used.
+An alternate way to use `AlertDialog` is with the `useAlert()` hook. It takes no parameters and returns an `async` function, `alert`. When `alert` is called (see AlertOptions below for its argument), it shows the alert dialog. When the dialog is dismissed, a promise is resolved with `true` or `false` depending on whether or not the acknowledgement button was used.
 
 ### Example (with `useAlert` hook)
 
 ```javascript live noinline
-function ShorthandExample2() {
+function ShorthandExample() {
   const alert = useAlert()
   const buttonClick = React.useCallback(
     async function (e) {
       if (
         await alert({
-          title: 'Are you sure?',
-          content: 'You must confirm this action to continue.',
-          confirmButtonContent: 'OK'
+          title: 'Erase USB Storage?',
+          content: 'By pressing "OK" you acknowledge All files will be deleted and cannot be recovered',
+          buttonContent: 'OK'
+          buttonType: 'danger'
         })
       ) {
-        e.target.textContent = 'Confirmed!'
+        e.target.textContent = 'Acknowledged!'
       }
     },
     [confirm]
   )
-  return (
-    <>
-      <Button onClick={buttonClick}>Confirmable action</Button>
-    </>
-  )
+  return <Button onClick={buttonClick}>Confirmable action</Button>
 }
-render(<ShorthandExample2 />)
+render(<ShorthandExample />)
 ```
 
 ### Example (using `AlertDialog` component)
@@ -225,8 +226,8 @@ render(<ShorthandExample2 />)
       <>
         <Button onClick={openDialog}>Open</Button>
         {isOpen && (
-          <AlertDialog title="Confirm action?" onClose={closeDialog}>
-            Are you sure you want to confirm this action?
+          <AlertDialog title="Invalid or Expired credit card" onClose={closeDialog}>
+            Please enter a valid form of payment.
           </AlertDialog>
         )}
       </>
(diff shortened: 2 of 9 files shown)


src/__tests__/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+    })
+
+    for (const expected of [cancel, confirm]) {
+      expect(screen.queryByText(expected)).toBeDefined()
+    }
+  })
+
+  it.each([
+    {prop: 'cancelButtonContent', visible: 'OK', notVisible: 'Cancel'},
+    {prop: 'confirmButtonContent', visible: 'Cancel', notVisible: 'OK'}
```
stephenotalora, line 69: This test will render the new design for confirmation dialog with one button

src/stories/ConfirmationDialog.stories.tsx
```diff
@@ -103,3 +103,33 @@ export const ShorthandHookFromActionMenu = () => {
...
+
+export const SingleConfirmFromActionMenu = () => {
+  const confirm = useConfirm()
+  const [text, setText] = useState('open me')
+  const onButtonClick = useCallback(async () => {
+    if (
+      await confirm({
+        title: 'Are you sure?',
+        content: 'Do you really want to do a trick?',
+        cancelButtonContent: null
```
stephenotalora, line 115: Added a story for testing purposes

src/Dialog/ConfirmationDialog.tsx
```diff
@@ -157,7 +172,7 @@ export const ConfirmationDialog: React.FC<ConfirmationDialogProps> = props => {
     variant: 'large',
     onClick: onConfirmButtonClick
   }
-  const footerButtons = [cancelButton, confirmButton]
+  const footerButtons = [cancelButton, confirmButton].filter(btn => !!btn.content)
```
stephenotalora, line 175: I could also add a `cancelButtonContentDisabled` prop with a default boolean value set to false to make the prop optional, rather than checking for existence. I didn't want to make a big change before consulting with the team, but definitely open to suggestions 🙂

colebemis (approved):
These changes look good to me 👍 Thanks for doing this, @stephenotalora!

cc @vdepizzol for design review

vdepizzol:
Hello @stephenotalora! I think we need a few more changes to support this. "Confirmation dialog" as the component is called, is a specific reference to dialogs that have both a confirmation action and a dismissive action. As written in the confirmation dialog docs:

> **ConfirmationDialog**
> A `ConfirmationDialog` is a special kind of dialog with more rigid behavior. It is used to confirm a user action. `ConfirmationDialog`s always have exactly two buttons: one to cancel the action and one to confirm it. No custom rendering capabilities are provided for `ConfirmationDialog`.
> https://primer.style/react/Dialog2#confirmationdialog

With the changes you're proposing, the naming no longer works. The design you're trying to incorporate is something I'd call an "alert dialog". [In some systems](https://developer.android.com/reference/androidx/appcompat/app/AlertDialog), "alert dialog" is a more generic name that can be used for one or more actions.

We've had conversations on simplifying confirmation dialog's button style (so they aren't flushed to the edges of the overlay and have buttons like the screenshot you have above), I wonder if by following that format it'd be easier to have a single "Alert dialog" component that can support 1) single-call alerts and 2) confirmation options (with confirmation and dismiss actions), or if we should have both `AlertDialog` and `ConfirmationDialog` as different components.

------

@cmwinters I think we need a better example to validate this since we don't have a lot of cases of alerts that pop up on the users' screen. Can you have a look at a final message this could have? `title` should probably be more specific... also, do you have other use cases of alerts across dotcom? Thanks!

> image

stephenotalora:
> I wonder if by following that format it'd be easier to have a single "Alert dialog" component that can support 1) single-call alerts and 2) confirmation options (with confirmation and dismiss actions), or if we should have both AlertDialog and ConfirmationDialog as different components

@vdepizzol my original intention with this PR was to minimize the number of changes to the API, I think having a separate more specialized component with a single call would make more sense. I'm happy to make those changes basing it off the `ConfirmDialog` API.

@cmwinters could you kindly please confirm if that's the direction we would want to go to?

cmwinters:
@stephenotalora +1 to moving forward with an `AlertDialog` based on the `ConfirmationDialog` API. 

@vdepizzol I'll work on the copy. The closest thing that I can think of across dotcom is the [Popover component](https://primer.style/css/components/popover)

Screen Shot 2021-08-16 at 11 51 04 AM

That's where I got the idea for the 'Got it!' button. We use these throughout dotcom to introduce new UI to users.

stephenotalora:
@colebemis @vdepizzol for  `AlertDialog` component I propose the following API which (again) is based off `ConfirmDialog` the confirm dialog interface

```ts
/**
 * Props to customize the AlertDialog.
 */
export interface AlertDialogProps {
  /**
   * Required. This callback is invoked when a gesture to close the dialog
   * is performed. The first argument indicates the gesture.
   */
  onClose: (gesture: 'confirm' | 'close-button' | 'cancel' | 'escape') => void

  /**
   * Required. The title of the ConfirmationDialog. This is usually a brief
   * question.
   */
  title: React.ReactNode

  /**
   * The text to use for the cancel button. Default: "Cancel".
   */
  cancelButtonContent?: React.ReactNode

  /**
   * The type of button to use for the confirm button. Default: Button.
   */
  confirmButtonType?: 'normal' | 'primary' | 'danger'
}

```

please note `AlertDialog`:
1. removes the `cancelButtonContent` prop
2. removes the `cencel` gesture argument value for the `onClose` callback

.changeset/neat-coins-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/components': minor
+---
+
+adding `AlertDialog` to account for cases where only one button is needed in a dialog.
```
colebemis, line 5:
```suggestion
Add `AlertDialog` component for cases where only one button is needed in a dialog.
```

docs/content/Dialog2.mdx
```diff
@@ -177,3 +177,78 @@ render(<ShorthandExample2 />)
 | cancelButtonContent  | `React.ReactNode`                 | `"Cancel"` | The content to use for the cancel button.                                                                 |
 | confirmButtonContent | `React.ReactNode`                 | `"OK"`     | The content to use for the confirm button.                                                                |
 | confirmButtonType    | `"normal" │ "primary" │ "danger"` | `Button`   | The type of button to use for the confirm button.                                                         |
+
+## AlertDialog
+
+An `AlertDialog` is a special kind of dialog with more rigid behavior. It is used to alert or confirm a single user action. AlertDialogs can only have at most one button: no buttons or one to confirm the alert dialog. No custom rendering capabilities are provided for AlertDialog.
```
colebemis, line 183: Can we be more explicit here about when to use a ConfirmationDialog vs an AlertDialog?

docs/content/Dialog2.mdx
```diff
@@ -177,3 +177,78 @@ render(<ShorthandExample2 />)
...
+    },
+    [confirm]
+  )
+  return (
+    <>
+      <Button onClick={buttonClick}>Confirmable action</Button>
+    </>
+  )
+}
+render(<ShorthandExample2 />)
```
colebemis, line 214:
```suggestion
function ShorthandExample() {
  const alert = useAlert()
  const buttonClick = React.useCallback(
    async function (e) {
      if (
        await alert({
          title: 'Are you sure?',
          content: 'You must confirm this action to continue.',
          confirmButtonContent: 'OK'
        })
      ) {
        e.target.textContent = 'Confirmed!'
      }
    },
    [confirm]
  )
  return <Button onClick={buttonClick}>Confirmable action</Button>
}
render(<ShorthandExample />)
```

src/Dialog/AlertDialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+import styled from 'styled-components'
+import {get} from '../constants'
+import Box from '../Box'
+import {useFocusZone} from '../hooks/useFocusZone'
+import {FocusKeys} from '../behaviors/focusZone'
+import ReactDOM from 'react-dom'
+import {ThemeProvider, useTheme, ThemeProviderProps} from '../ThemeProvider'
+
+/**
+ * Props to customize the ConfirmationDialog.
```
colebemis, line 12:
```suggestion
 * Props to customize the AlertDialog.
```

src/Dialog/AlertDialog.tsx
```diff
@@ -0,0 +1,171 @@
...
+ */
+export interface AlertDialogProps {
+  /**
+   * Required. This callback is invoked when a gesture to close the dialog
+   * is performed. The first argument indicates the gesture.
+   */
+  onClose(gesture: 'confirm' | 'close-button' | 'escape'): void
+
+  /**
+   * Required. The title of the ConfirmationDialog. This is usually a brief
```
colebemis, line 22:
```suggestion
   * Required. The title of the AlertDialog. This is usually a brief
```

src/Dialog/ConfirmationDialog.tsx
```diff
@@ -16,7 +16,7 @@ export interface ConfirmationDialogProps {
    * Required. This callback is invoked when a gesture to close the dialog
    * is performed. The first argument indicates the gesture.
    */
-  onClose: (gesture: 'confirm' | 'cancel' | 'close-button' | 'cancel' | 'escape') => void
```
colebemis, line 19: Nice catch ✨

src/__tests__/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import {fireEvent, screen} from '@testing-library/dom'
+import {ConfirmationDialog, ConfirmationDialogProps} from '../Dialog/ConfirmationDialog'
+import {render, RenderResult} from '@testing-library/react'
+
+describe('ConfirmationDialog', () => {
```
colebemis, line 6: Thanks for adding tests 💖

vdepizzol:
Sorry for having this discussion in a PR.

vdepizzol:
@cmwinters I wonder if a better flow would include a suggestion to convert to an issue right from the dialog, like the screenshot below. I'm worried we're introducing a new component here without clear guidelines or examples of when it should be used.

Screen Shot 2021-08-18 at 11 49 04

(Sorry for having this discussion in a PR 😭.)

.changeset/neat-coins-mix.md
stephenotalora: updated

docs/content/Dialog2.mdx
stephenotalora: updated with a more explicit description, please let me know if that is a good direction. 👍🏽

docs/content/Dialog2.mdx
stephenotalora: updated based on the description of the component

src/Dialog/AlertDialog.tsx
stephenotalora: rolled these up into one commit

src/Dialog/AlertDialog.tsx
stephenotalora: rolled these up into one commit

docs/content/Dialog2.mdx
```diff
@@ -180,38 +180,39 @@ render(<ShorthandExample2 />)
 
 ## AlertDialog
 
-An `AlertDialog` is a special kind of dialog with more rigid behavior. It is used to alert or confirm a single user action. AlertDialogs can only have at most one button: no buttons or one to confirm the alert dialog. No custom rendering capabilities are provided for AlertDialog.
+AlertDialog is a special kind of dialog with an urgent interruption, and informs the user about a situation.
+They summarize a **single** decision that requires acknowledgement and can only be dismissed by the alert close button or `escape` key.
+Use a clear question or statement for the alert title with an explanation in the content area, such as "Erase Disk storage?".
```
colebemis, line 185: This is a bit nitpicky but isn't `Erase Disk storage` a confirmation message that we would use a Confirmation Dialog for?

colebemis: I think we can avoid having to add `AlertDialog` right now by rewording the dialog message in Memex (like @vdepizzol suggested https://github.com/primer/react/pull/1378#issuecomment-901351218)

stephenotalora: @cmwinters thoughts?

PR #1384, opened 2021-08-19
Document ESLint plugin

colebemis:


docs/content/linting.md
```diff
@@ -0,0 +1,35 @@
+---
+title: Linting
+description: Primer React offers an ESLint plugin to enforce best practices and fix common problems. 
+---
+
+<Note variant="warning">
```
jfuchs, line 6: 🤔 do we treat all `.md` files as `.mdx`?

jfuchs (approved):

docs/content/linting.md
colebemis: Yup. Here's where that's configured: https://github.com/primer/doctocat/blob/master/theme/gatsby-config.js#L21

PR #1388, opened 2021-08-23
Overlay and its composing components take a portalContainerName prop

### LINKED ISSUE #1169: ActionMenu and DropdownMenu overlay do not reposition when window is resized.
**Describe the bug**

ActionMenu and DropdownMenu rely on Overlay for the expandable panel that pops out upon triggering the anchor. Overlay is positioned with absolute positioning and relies on the `useAnchoredPosition` hook to position the overlay at the anchors position. There seems to be an issue with this when the window is resized:

https://github.com/primer/components/pull/1152#issuecomment-820750194

**To Reproduce**
1. Go to `[primer-docs]/DropdownMenu`
2. Click on the Dropdown Anchor
3. Resize the width of the window
4. Observe how the overlay changes position

**Expected behavior**
The expanded overlay stays in place

**Screenshots**
Screen Shot 2021-04-15 at 6 47 45 PM

jfuchs:
This change adds a `portalContainerName` prop to `Overlay`, `AnchoredOverlay`, `DropdownMenu`, `SelectPanel`, and `ActionMenu`. This allows overlays with an anchor inside a scrolling container to track with their anchor, so long as the specified portal is also inside that scrolling container.

Closes https://github.com/github/primer/issues/229.

### Screenshots

https://user-images.githubusercontent.com/21195/130521483-10a9c965-5dfb-43c0-9692-ee6ae12de624.mov

### Merge checklist

- [ ] Added/updated tests
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


colebemis:
Could you explain why the `portalContainerName` prop is necessary?

colebemis:
@jfuchs That makes sense! Can you add that to the PR description and the changeset?

docs/content/ActionMenu.mdx
```diff
@@ -68,13 +68,14 @@ An `ActionMenu` is an ActionList-based component for creating a menu of actions
...
+| onAction            | (props: ItemProps) => void            |    `undefined`    | Optional. If defined, this function will be called when a menu item is activated either by a click or a keyboard press.                                                                            |
+| open                | boolean                               |    `undefined`    | Optional. If defined, ActionMenu will use this to control the open/closed state of the Overlay instead of controlling the state internally. Should be used in conjunction with the `setOpen` prop. |
+| setOpen             | (state: boolean) => void              |    `undefined`    | Optional. If defined, ActionMenu will use this to control the open/closed state of the Overlay instead of controlling the state internally. Should be used in conjunction with the `open` prop.    |
+| portalContainerName | boolean                               |    `undefined`    | Optional. See the `portalContainerName` prop of `Overlay`                                                                                                                                          |
```
dgreif, line 81: Looks like this type is wrong (in each of the docs). Should be `string`

src/stories/AnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+    }
+  ]
+} as Meta
+
+const HeaderAndLayout = ({children, includePortal}: {children: JSX.Element; includePortal: boolean}) => {
+  const scrollingElementRef = useRef<HTMLDivElement>(null)
+  useEffect(() => {
+    if (scrollingElementRef.current && includePortal) {
+      registerPortalRoot(scrollingElementRef.current)
+    }
```
dgreif, line 30: This story doesn't appear to use `portalContainerName`.  Instead, it looks like it just registers the default portal within the scrolling element

src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -168,6 +169,7 @@ export const AnchoredOverlay: React.FC<AnchoredOverlayProps> = ({
           top={position?.top || 0}
           left={position?.left || 0}
           anchorSide={position?.anchorSide}
+          {...(portalContainerName ? {portalContainerName} : {})}
           {...overlayProps}
```
dgreif, line 173: We already have an option to pass any `overlayProps` through the Anchored Overlays, which seems like an OK path for this option as well.  It should be rarely used, kind of like height/width which use this other approach.  Should significantly cut down on the surface area of this PR as well

dgreif (changes requested):
Looks like the right approach, but I think we can simplify this quite a bit by leveraging the existing props

src/Overlay.tsx
```diff
@@ -15,6 +15,7 @@ type StyledOverlayProps = {
   maxHeight?: keyof Omit<typeof heightMap, 'auto' | 'initial'>
   visibility?: 'visible' | 'hidden'
   anchorSide?: AnchorSide
+  portalContainerName?: string
```
dgreif, line 18: We should only need to define this prop in `OverlayProps`.  The values here are specifically for what gets passed in to the styled component.  You should be able to remove this line without issues

dgreif (approved):
Looks much cleaner, thanks for simplifying and updating the stories!  One small item then this should be good to go

dgreif (approved):

PR #1390, opened 2021-08-24
Update ConfirmationDialog buttons

colebemis:
Updates `ConfirmationDialog` buttons to match @vdepizzol's updated design:

| Before | After |
| --- | --- |
| CleanShot 2021-08-23 at 17 20 25@2x | CleanShot 2021-08-23 at 17 21 39@2x |

Closes https://github.com/github/primer/issues/262


vdepizzol (approved):
✨ ✨

Issue #1395, opened 2021-08-31
SelectPanel moves when page content changes


jclem:
**Describe the bug**

When page content changes while a SelectPanel is open, the SelectPanel tends to move around on the page.

**To Reproduce**

See [this CodeSandbox](https://codesandbox.io/s/lucid-dawn-7kgh4?file=/src/App.js) to reproduce.

**Expected behavior**

I would expect the position of SelectPanel to remain stationary while page content changes that doesn't affect the anchor's position.

**Screenshots**

https://user-images.githubusercontent.com/412282/131536200-94b2bba2-6deb-40a3-aba7-8001fadb8b04.mp4

This happens in any browser/OS I have tested it in (macOS Chrome/Safari) with @primer/components v28.4.0

colebemis:
Thank you for the reproduction, @jclem! I just added this to the "Up next" column of our project board.

jfuchs:
I believe this is a consequence of the positioning of our portal (in this case the default portal which is statically positioned div appended after the react root). I think it'd be reasonable to style that default portal with `position:absolute;top:0;left:0`, and if you don't want that you can always BYO portal. @jclem @colebemis does that sound reasonable?

jfuchs:
@jclem, in your case where you ran into this, are you specifying the portal root or letting primer create one? You should be able to specify a portal root and absolutely position it to get things working. If you're letting primer create one and you're not in a hurry for a fix, that should be fixed in https://github.com/primer/react/pull/1398.

jfuchs:
With https://github.com/primer/react/pull/1398 merged, I'm going to close this. Feel free to re-open if you still have issues!

jclem:
I think the absolute positioning will probably help, yeah. I think that calling out how this positioning works more explicitly in the Overlay documentation would help a lot (so that one could more quickly ascertain that the default positioning may be causing some issue), and linking from the docs for components like SelectPanel to the Overlay docs would also be helpful.

jfuchs:
> I think that calling out how this positioning works more explicitly in the Overlay documentation would help a lot (so that one could more quickly ascertain that the default positioning may be causing some issue), and linking from the docs for components like SelectPanel to the Overlay docs would also be helpful.

Good call. I'll take a pass at improving the docs!

PR #1398, opened 2021-08-31
Default portal containers created by primer are absolutely positioned at 0,0

jfuchs:
This PR adds three styling rules to the primer-created portal container root: `position: absolute, top: 0, left: 0`. This doesn't add any styling to either roots provided via `registerPortalRoot` or roots provided by placing an element with id `__primerPortalRoot__` in the DOM.

### Why?

Compared to statically positioning a portal root after body content, absolutely positioning a portal root relative to the body will lead to more stable overlay positioning across DOM changes.

See also https://github.com/primer/react/issues/1395

### Screenshots

N/A

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


colebemis:
Does this fix this issue? https://github.com/primer/react/issues/1395

dgreif (approved):
I'm a little nervous about this change, given the odd behavior we have been seeing in Safari _without_ it.  I'd give this some good testing, both in stories here and in Projects Beta, in all the three major browsers.  If all looks good, I'm ok with this solution

PR #1409, opened 2021-09-07
Use @react-aria/ssr for isomorphic ID generation.

jfuchs:
Just a draft for now, need to wait for https://github.com/adobe/react-spectrum/pull/2278 to land and then a version bump.

----

Describe your changes here.

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/ssr.mdx
+++ b/docs/content/ssr.mdx
@@ -4,7 +4,7 @@ title: Server-side rendering with Primer React
 
 ## SSR-safe ID generation
 
-Some Primer components generate their own DOM IDs. Those IDs must be unique across the DOM and isomorphic (so that server-side rendering and client-side rendering yield the same ID, avoiding hydration issues). We use [@react-aria/ssr](https://react-spectrum.adobe.com/react-aria/ssr.html) to generate those IDs. In client-only rendering, this doesn't require any additional work. In SSR contexts, you must wrap your application with at least one `SSRProvider`:
+Some Primer components generate their own DOM IDs. Those IDs must be isomorphic (so that server-side rendering and client-side rendering yield the same ID, avoiding hydration issues) and unique across the DOM. We use [@react-aria/ssr](https://react-spectrum.adobe.com/react-aria/ssr.html) to generate those IDs. In client-only rendering, this doesn't require any additional work. In SSR contexts, you must wrap your application with at least one `SSRProvider`:
 
 ```
 import {SSRProvider} from '@primer/components';
--- a/docs/content/ssr.mdx
+++ b/docs/content/ssr.mdx
@@ -4,7 +4,7 @@ title: Server-side rendering with Primer React
 
 ## SSR-safe ID generation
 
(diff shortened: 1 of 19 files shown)


jfuchs:
per https://github.com/adobe/react-spectrum/pull/2278#issuecomment-914531233, we might get a release early next week.

docs/content/ssr.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: Server-side rendering with Primer React
```
colebemis, line 2: Thank you for writing docs! 💖

docs/content/ssr.mdx
```diff
@@ -0,0 +1,37 @@
...
+---
+
+## SSR-safe ID generation
+
+Some primer components generate their own DOM IDs. Those IDs must be unique across the DOM and isomorphic (so that server-side rendering and client-side rendering yield the same ID, avoiding hydration issues). We use [@react-aria/ssr](https://react-spectrum.adobe.com/react-aria/ssr.html) to generate those IDs. In client-only rendering, this doesn't require any additional work. In SSR contexts, you must wrap your application with at least one `<SSRProvider>`:
```
colebemis, line 7:
```suggestion
Some Primer components generate their own DOM IDs. Those IDs must be unique across the DOM and isomorphic (so that server-side rendering and client-side rendering yield the same ID, avoiding hydration issues). We use [@react-aria/ssr](https://react-spectrum.adobe.com/react-aria/ssr.html) to generate those IDs. In client-only rendering, this doesn't require any additional work. In SSR contexts, you must wrap your application with at least one `SSRProvider`:
```

colebemis: For some reason, "unique across the DOM and isomorphic" sounds awkward to me. Maybe it would read better if we flipped the order: "isomorphic and unique across the DOM"?

docs/content/ssr.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+```
+import {SSRProvider} from '@primer/components';
+
+<SSRProvider>
+  <MyApplication />
+</SSRProvider>
```
colebemis, line 14: Having a floating JSX expression might be confusing to some less experienced readers:
```suggestion
function App() {
  return (
    <SSRProvider>
      <MyApplication />
    </SSRProvider>
  )
}
```

src/utils/uniqueId.ts
```diff
@@ -1,3 +1,5 @@
+// Note: uniqueId may be unsafe in SSR contexts if it is used create DOM IDs or otherwise cause a hydration warning. Use useSSRSafeId instead.
+
 let idSeed = 10000
 export function uniqueId(): string {
```
colebemis, line 4: Is `uniqueId` still useful if we have `useSSRSafeId`?

colebemis (approved):
Looks great! Just left a few minor comments.

src/utils/uniqueId.ts
jfuchs: There's one place where it's still used — the `useFocusZone` behavior. I believe it's safe because it's only called downstream of a user interaction. As far as I can tell, we'd need to do a decent refactoring or just require that IDs are already in place to get rid of it.

docs/content/ssr.mdx
colebemis: Can you add this doc to the side nav? https://github.com/primer/react/blob/main/docs/src/%40primer/gatsby-theme-doctocat/nav.yml

Issue #1377, opened 2021-08-11
Inaccurate type for TextInput's onChange property


jclem:
**Describe the bug**

I expected the `onChange` property of `TextInput` to have the same type as `JSX.InstrinsicElements['input']`, but it appears that `currentTarget` on the event is more than `HTMLInputElement`:

<img width="709" alt="CleanShot 2021-08-11 at 15 04 56@2x" src="https://user-images.githubusercontent.com/412282/129109702-77999eac-2698-4671-a9cc-e48c2ddf2a9d.png">

**To Reproduce**
1. Create a component with `TextInput` whose `value` is a piece of string state
2. Try to set the string state with `onChange={e => setState(e.currentTarget.value)}`

**Expected behavior**

No type error occurs.

**Screenshots**

See above with the current typing for `onChange`.

colebemis:
Added this to our planning project board. Thank you!

Issue #1163, opened 2021-04-14
v27.0.0 breaks TypeScript typings


hehex9:
**Describe the bug**
1. some components (Button*/Dialog/TextInput/FilterList/...) have incorrect type definitions, and ts keeps complaining:
```sh
Property 'css' is missing in type '{}' but required in type 'Pick<TextInputInternalProps, "form" | "p" | "slot" | "style" | "title" | "pattern" | "children" | "type" | "theme" | "key" | "defaultChecked" | "defaultValue" | ... 317 more ... | "icon">'.ts(2741)
```
2. callbacks like `onChange`/`onKeyPress` in `TextInput` are not giving us the right types, it will cause an error like:
```sh
Parameter 'e' implicitly has an 'any' type.ts(7006)
```

vscode shows the `onChange`'s type definition is:
```ts
onChange?: (((event: React.FormEvent<HTMLSpanElement>) => void) & ((event: React.ChangeEvent<HTMLInputElement>) => void)) | undefined
```
which is not accurate

**To Reproduce**
[codesanbox](https://codesandbox.io/s/primer-components-ts-mcsfk?file=/src/index.tsx)

**Expected behavior**
no error

**Screenshots**
<img width='420' alt='codesanbox' src='https://user-images.githubusercontent.com/9209882/114643583-8618bf00-9d08-11eb-9029-c54b89fff6fc.jpg' />

**Additional context**
v26 was fine

colebemis:
Thanks for the bug report and clear reproduction, @hehex9! We'll investigate this week.

hehex9:
> we're not seeing these TypeScript errors in any of the internal projects

That's weird, since it's pretty easy to reproduce in codesandbox i think :(

I'll look deeper into it when I have time, would you mind sharing your attempts here? Thanks

velocity23:
Having issues as well, having to `@ts-ignore` for `TextInput`

yareyaredesuyo:
Maybe related to using React.ComponentType?
I think there is no need to use it, just declare type and export it.

Refactor like this (using Box):

before

```tsx
const Box = styled.div<SystemCommonProps & SystemFlexProps & SystemLayoutProps & SxProp>`
  ${COMMON}
  ${FLEX}
  ${LAYOUT}
  ${sx};
`

export type BoxProps = ComponentProps<typeof Box>
```

after

```tsx
export type BoxProps = SystemCommonProps & SystemFlexProps & SystemLayoutProps & SxProp

[...]

PR #1414, opened 2021-09-08
Fix TextInput prop types & remove system props

jfuchs:
This PR should give us a more accurate type for the props of `TextInput`. Previously the props included:

```
type TextInputInternalProps = {
  // etc
} & ComponentProps<typeof Wrapper> &
  ComponentProps<typeof Input>
```

But that left some props as a messy intersection of types, like `onChange` (https://github.com/primer/react/issues/1377). We can resolve that by explicitly omitting any fields from Input's props that are explicitly destructured from TextInput's props.

Closes https://github.com/primer/react/issues/1377

See also https://github.com/primer/react/issues/1163

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/cuddly-rice-march.md
+++ b/.changeset/cuddly-rice-march.md
@@ -1,0 +1,5 @@
+---
+'@primer/components': major
+---
+
+Removed system props support from `<TextInput>` and fixed its type definition.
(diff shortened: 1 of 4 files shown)


jfuchs:
I've got some tests to fix, but wanted to put up the PR to get feedback on the approach

colebemis (approved):
👍

PR #1416, opened 2021-09-09
Avatar: add alt and src

### LINKED ISSUE #1405: Avatar: document `src` and `alt` props
## Problem 

The [props table](https://primer.style/react/Avatar#props) for the Avatar component does not include the `src` and `alt` props.

[CleanShot 2021-09-05 at 17 51 08](https://primer.style/react/Avatar)

## Context

Prop documentation for Avatar is generated by [react-docgen-typescript](https://www.npmjs.com/package/react-docgen-typescript). `src` and `alt` are missing for the docs because they're not included in the props type definition for Avatar: https://github.com/primer/react/blob/main/src/Avatar.tsx#L6-L12

## Do this

- [ ] Add `src` and `alt` to the props type definition for Avatar

## Open questions

- Should `src` and `alt` be required?

---
cc @joelhawksley @ashygee

siddharthkp:
Describe your changes here.

Closes #1405

### Screenshots

before:
image
after:
image

### Worth mentioning

I've made `alt` text optional based on [@alliethu's comment](https://github.com/primer/react/issues/1405#issuecomment-915608230). We set the default value to `""` (empty string) so that screen readers ignore it.

### Room for improvement

1. The default value for `alt` shows up as `-` even though the value is an empty string. This is because we set [`savePropValueAsString`](https://github.com/primer/react/blob/main/docs/gatsby-node.js#L47) which treats `undefined` and `""` the same. It would be interesting to show an explicit empty string as `null` and empty string have different meaning for `alt` ([reference](https://webaim.org/techniques/alttext/#basics) via @alliethu, thx! 👋 )
2. Right now, we don't indicate a prop is required, it would be really nice to have that.
3. The order of props in the documentation is not the same as the order they are defined. It would be nice to control the order

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Avatar.tsx
+++ b/src/Avatar.tsx
@@ -8,6 +8,10 @@ type StyledAvatarProps = {
   size?: number
   /** Sets the shape of the avatar to a square if true. If false, the avatar will be circular. */
   square?: boolean
+  /** URL of the avatar image. */
+  src: string
+  /** Provide an alt text in case the Avatar is used without the user's name next to it. */
+  alt?: string
 } & SystemCommonProps &
   SxProp
 
--- a/src/Avatar.tsx
+++ b/src/Avatar.tsx
@@ -10,7 +10,7 @@ type StyledAvatarProps = {
   square?: boolean
   /** URL of the avatar image. */
   src: string
-  /** Provide an alt text in case the Avatar is used without the user's name next to it. */
+  /** Provide alt text when the Avatar is used without the user's name next to it. */
   alt?: string
 } & SystemCommonProps &
   SxProp
--- a/src/__tests__/Avatar.tsx
+++ b/src/__tests__/Avatar.tsx
@@ -16,7 +16,7 @@ describe('Avatar', () => {
   })
 
   it('should have no axe violations', async () => {
-    const {container} = HTMLRender(<Avatar />)
+    const {container} = HTMLRender(<Avatar src="primer.png" />)
     const results = await axe(container)
     expect(results).toHaveNoViolations()
     cleanup()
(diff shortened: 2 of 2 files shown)


alliethu:
> The default value for alt shows up as - even though the value is an empty string.

@siddharthkp Is this `alt` update in a place I can access it and run through it with a ScreenReader? I'm a little concerned the output value being `-`, might not be accessible to SR users.

siddharthkp:
> @siddharthkp Is this alt update in a place I can access it and run through it with a ScreenReader? I'm a little concerned the output value being -, might not be accessible to SR users.

Hi @alliethu!

Whoops, bad choice of words on my side! The rendered alt text is still an empty string by default (didn't change that). However, we show the wrong value in the documentation part (`-` instead of empty string)

The preview is deployed here: [preview/Avatar](https://primer-components-70pv7kc23-primer.vercel.app/react/Avatar)

image


The rendered alt text is an empty string:

image

alliethu:
@siddharthkp Perfect! Thank you for clarifying!! 🙇🏼‍♀️

colebemis (approved):
Looks great. Thanks for doing this, @siddharthkp! ✨ 

I agree with all of the things you listed in `Room for improvement`. Do you want to open up PRs for those next?

src/Avatar.tsx
```diff
@@ -8,6 +8,10 @@ type StyledAvatarProps = {
   size?: number
   /** Sets the shape of the avatar to a square if true. If false, the avatar will be circular. */
   square?: boolean
+  /** URL of the avatar image. */
+  src: string
+  /** Provide an alt text in case the Avatar is used without the user's name next to it. */
```
colebemis, line 13: [Minor] I could see this being misunderstood as "always provide alt text just in case the avatar isn't used next to a username." I think we could reword it to make a little more clear:
```suggestion
  /** Provide alt text when the Avatar is used without the user's name next to it. */
```

siddharthkp:
> I agree with all of the things you listed in Room for improvement. Do you want to open up PRs for those next?

Sounds good 👍 

What's our merge process, do I just merge this when the build passes?

colebemis:
> What's our merge process, do I just merge this when the build passes?

Yeah, go ahead and merge it 🚢

PR #1436, opened 2021-09-17
add colormode and colorscheme toggles to storybook toolbar

### LINKED ISSUE #1433: Preview components in different themes
## Background

Primer React components can be rendered in many [different themes](https://primer.style/primitives/colors#themes). See the [Theming](https://primer.style/react/theming) for more details.

| Light | Dark |
| --- | --- |
| CleanShot 2021-09-16 at 16 27 32 | CleanShot 2021-09-16 at 16 28 17 |

In fact, one of the criteria for a component to be [alpha](https://primer.style/contribute/component-lifecycle#alpha) is that "the component is compatible with color modes and can adapt to different themes."

## Problem

We don't have an easy way to preview components in different themes in documentation examples and Storybook.

## Tasks

- [ ] Implement a way to preview components in different themes in documentation examples
- [x] Implement a way to preview components in different themes in Storybook https://github.com/primer/react/pull/1436

langermank:
Adds a [custom decorator](https://storybook.js.org/docs/react/essentials/toolbars-and-globals#create-a-decorator) that wraps all stories with `ThemeProvider`

The decorator controls `colorMode` and `colorScheme` for stories using a new dropdown menu in the toolbar

Handles the storybook implementation for #1433 

To test locally, checkout `storybook-exploration` and run `npm run start:storybook`

### Open questions

- Should we find a way to combine `dayScheme` and `nightScheme`? Without much experience with schemes, do they cancel each other out when both set?
- Are these menu icons clear enough? The [options](https://5a375b97f4b14f0020b0cda3-wbeulgbetj.chromatic.com/?path=/story/basics-icon--labels) are limited, unfortunately (they don't have a moon or sun) 🤷 

### Screenshots

Screen Shot 2021-09-17 at 10 53 38 AM

colormodes-storybook


.storybook/preview.js
```diff
@@ -1,4 +1,63 @@
...
+      showName: true
+    }
+  },
+  dayScheme: {
+    name: 'Day color scheme',
+    description: 'Day color scheme',
+    defaultValue: 'light',
+    toolbar: {
+      icon: 'circlehollow',
+      items: ['light', 'dark', 'dark_dimmed'],
```
colebemis, line 31: We can avoid hard-coding these names here if we do something like this:

```js
import {theme} from '../src'

export const globalTypes = {
  ...
  dayScheme: {
    toolbar: {
      items: Object.keys(theme.colorSchemes)
    }
  }
}
```

.storybook/preview.js
```diff
@@ -1,4 +1,63 @@
...
+    color: ${themeGet('colors.text.primary')}
+  }
+`
+export const globalTypes = {
+  colorMode: {
+    name: 'Color mode',
+    description: 'Color mode (day, night, auto)',
+    defaultValue: 'day',
+    toolbar: {
+      icon: 'paintbrush',
```
colebemis, line 19: Does storybook have a black and white circle icon something like this: 🌗?

.storybook/preview.js
```diff
@@ -1,4 +1,63 @@
...
+    defaultValue: 'dark',
+    toolbar: {
+      icon: 'circle',
+      items: ['light', 'dark', 'dark_dimmed'],
+      showName: true
+    }
+  }
+}
+
+// context.globals.X references items in globalTypes
```
colebemis, line 47: This is a helpful comment. Thanks for including this!

.storybook/preview.js
```diff
@@ -1,4 +1,63 @@
 import {addons} from '@storybook/addons'
+import {ThemeProvider} from '../src/ThemeProvider'
+import {themeGet} from '../src'
```
colebemis, line 3: ThemeProvider is also exported from `src/index.ts` so we can combine these imports:

```suggestion
import {ThemeProvider, themeGet} from '../src'
```

.storybook/preview.js
```diff
@@ -1,4 +1,63 @@
...
+import {createGlobalStyle} from 'styled-components'
+
+// set global theme styles for each story
+const GlobalStyle = createGlobalStyle`
+  body {
+    background-color: ${themeGet('colors.bg.primary')};
+    color: ${themeGet('colors.text.primary')}
```
colebemis, line 10:
```suggestion
    color: ${themeGet('colors.text.primary')};
```

.storybook/preview.js
colebemis: Looks like they don't 😞 I

 think a `paintbrush` should be fine 👍

colebemis:
This is amazing, @langermank! 💖 Thank you for doing this.

These controls will make it much easier for us to build themeable components!

colebemis:
> Should we find a way to combine dayScheme and nightScheme? Without much experience with schemes, do they cancel each other out when both set?

`dayScheme` is only active in `day` mode and `nightScheme` is only active in `night` mode. They don't conflict with each other when they're both set

.storybook/preview.js
langermank: I'm sure there's a way to pass custom icons to the toolbar, but I couldn't find much after a quick search. Someday!

.storybook/preview.js
```diff
@@ -1,4 +1,62 @@
...
+`
+export const globalTypes = {
+  colorMode: {
+    name: 'Color mode',
+    description: 'Color mode (day, night, auto)',
+    defaultValue: 'day',
+    toolbar: {
+      icon: 'paintbrush',
+      // array of colorMode items
+      items: Object.keys(theme.colorSchemes),
```
colebemis, line 20: This should stay hard-coded 😅 

"color modes" and "color schemes" are _different_ (but related) ideas in Primer React.

```suggestion
      items: ['day', 'night', 'auto'],
```

colebemis (approved):
Woooo! 🎉

PR #1441, opened 2021-09-20
Exclude storybook stories from TypeScript compilation for builds

jfuchs:
In the course of investigating TypeScript type questions with @colebemis, we noticed that building our typedefs with _any_ Storybook story included would lead to some components gaining a css prop in their typedef. E.g., lib/Button/Button.d.ts would include the line:

```
      css?: import("@emotion/core").InterpolationWithTheme<any>;
```

And excluding all stories would give us a build without any `css` props attached to components. This solution should give us clean type defs in our builds while still allowing typechecking of stories. Some caveats:

- There is a possibility that some errors will not be caught until running the build script, which isn't part of most people's workflow, from what I can tell. I suspect these will be rare.
- We still don't know _why_ the optional property became required as it did in the linked bug.
- This bug might only have surfaced with the intersection of multiple factors, and we're removing only one. There might be some other patterns we should still clean up (wild speculation: incorrect typing around PRC components spreading "rest" props on subcomponents)

Closes https://github.com/primer/react/issues/1163

### Screenshots

image
😅 

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/brave-rats-sneeze.md
+++ b/.changeset/brave-rats-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@primer/components': patch
+---
+
+Exclude storybook stories from TypeScript compilation for builds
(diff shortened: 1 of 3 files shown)


dgreif (approved):
Nice fix!

PR #1446, opened 2021-09-21
Move text input style overrides into sx prop

colebemis:
## Problem

We recently removed system props from the `TextInput` component (https://github.com/primer/react/pull/1414) but we still relied on TextInput system props in SelectPanel. This caused a visual regression in SelectPanel:

CleanShot 2021-09-21 at 11 30 29@2x

## Solution

To fix this issue, I moved the TextInput style overrides into the `sx` prop instead of system props.

CleanShot 2021-09-21 at 11 38 50@2x

### DIFF
--- a/.changeset/brave-rats-sneeze.md
+++ b/.changeset/brave-rats-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@primer/components': patch
+---
+
+Exclude storybook stories from TypeScript compilation for builds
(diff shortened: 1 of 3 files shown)


src/SelectPanel/SelectPanel.tsx
```diff
@@ -136,8 +136,7 @@ export function SelectPanel({
 
   const extendedTextInputProps: Partial<TextInputProps> = useMemo(() => {
     return {
-      mx: 2,
-      my: 2,
+      sx: { m: 2 },
```
colebemis, line 139:
```suggestion
      sx: {m: 2},
```

jfuchs (approved):
Good catch, thanks. I wonder why this typechecked without error...

PR #1448, opened 2021-09-21
Rename and deprecate Breadcrumb to Breadcrumbs

### LINKED ISSUE #1393: Rename `Breadcrumb` to `Breadcrumbs`
In [this Slack discussion](https://github.slack.com/archives/GACAW0NPM/p1629410372409700) we decided that the `Breadcrumb` component should be renamed to `Breadcrumbs`

## Action items
- [ ] Deprecate `Breadcrumb`
- [ ] Export `Breadcrumbs`

SferaDev:
Closes #1393

- Rename ``Breadcrum`` to ``Breadcrums`` and ``BreadcrumItem`` to ``BreadcrumsItem`` with their types.
- Keep and deprecate existing exports
- Update tests to the new name

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### Notes

Not sure if we should update the existing codemods to include this rename. Also as an outsider the documentation for the codemods is very limited (aside from the migrating.md file).


src/Breadcrumbs.tsx
```diff
@@ -79,9 +79,16 @@ const BreadcrumbItem = styled.a.attrs<StyledBreadcrumbItemProps>(props => ({
...
-export type BreadcrumbItemProps = ComponentProps<typeof BreadcrumbItem>
-export default Object.assign(Breadcrumb, {Item: BreadcrumbItem})
+export type BreadcrumbsItemProps = ComponentProps<typeof BreadcrumbsItem>
+export default Object.assign(Breadcrumbs, {Item: BreadcrumbsItem})
+
+/**
+ * @deprecated Use the Breadcrumbs component instead (i.e. <Breadcrumb> → <Breadcrumbs>)
+ */
+export const Breadcrumb = Object.assign(Breadcrumbs, {Item: BreadcrumbsItem})
+export type BreadcrumbProps = ComponentProps<typeof BreadcrumbsBase>
```
colebemis, line 93: Can we add deprecation warnings for these types as well?

src/Breadcrumbs.tsx
```diff
@@ -31,39 +31,39 @@ const Wrapper = styled.li`
...
   ${sx};
 `
 
-export type BreadcrumbProps = ComponentProps<typeof BreadcrumbBase>
+export type BreadcrumbsProps = ComponentProps<typeof BreadcrumbsBase>
 
-function Breadcrumb({className, children, theme, ...rest}: React.PropsWithChildren<BreadcrumbProps>) {
-  const classes = classnames(className, 'Breadcrumb')
+function Breadcrumbs({className, children, theme, ...rest}: React.PropsWithChildren<BreadcrumbsProps>) {
+  const classes = classnames(className, 'Breadcrumbs')
```
colebemis, line 45: I wonder why we need to add this classname 🤔

src/__tests__/Breadcrumbs.tsx
```diff
@@ -1,31 +1,32 @@
...
-describe('Breadcrumb', () => {
-  behavesAsComponent({Component: Breadcrumb, systemPropArray: [COMMON]})
+describe('Breadcrumbs', () => {
+  behavesAsComponent({Component: Breadcrumbs, systemPropArray: [COMMON]})
 
-  checkExports('Breadcrumb', {
-    default: Breadcrumb
+  checkExports('Breadcrumbs', {
+    default: Breadcrumbs,
+    Breadcrumb: Breadcrumb
```
colebemis, line 15:
```suggestion
    Breadcrumb
```

src/Breadcrumbs.tsx
SferaDev: Sure thing! I followed the same approach as when Position subcomponents where deprecated (component had depreciation warning but not the props types). In any case makes sense to see the message in the IDE if you're importing the outdated types.

colebemis (approved):
Looks good 👍 Thank you, @SferaDev!

src/Breadcrumbs.tsx
colebemis: If there are no references it anywhere else in the codebase, let's go ahead and remove it

SferaDev: There's a specific test case that validates a ``Breadcrum`` class exists, so it was done on purpose for what it seems. There're other components with readable classes injected to the components. But I don't think it's a convention as other components do not add them.

The removal or rename to ``Breadcrums`` will be breaking if someone relies on the class name for their tests (although it's not a good practice).

colebemis: Let's go ahead and remove it (and the test case) and add a changeset to indicate that it's a breaking change

src/Breadcrumbs.tsx
```diff
@@ -79,9 +78,24 @@ const BreadcrumbItem = styled.a.attrs<StyledBreadcrumbItemProps>(props => ({
...
-BreadcrumbItem.displayName = 'Breadcrumb.Item'
+BreadcrumbsItem.displayName = 'Breadcrumbs.Item'
 
-export type BreadcrumbItemProps = ComponentProps<typeof BreadcrumbItem>
-export default Object.assign(Breadcrumb, {Item: BreadcrumbItem})
+export type BreadcrumbsItemProps = ComponentProps<typeof BreadcrumbsItem>
+export default Object.assign(Breadcrumbs, {Item: BreadcrumbsItem})
+
+/**
+ * @deprecated Use the Breadcrumbs component instead (i.e. <Breadcrumb> → <Breadcrumbs>)
```
colebemis, line 89:
```suggestion
 * @deprecated Use the `Breadcrumbs` component instead (i.e. `<Breadcrumb>` → `<Breadcrumbs>`)
```

src/Breadcrumbs.tsx
```diff
@@ -79,9 +78,24 @@ const BreadcrumbItem = styled.a.attrs<StyledBreadcrumbItemProps>(props => ({
...
+ */
+export const Breadcrumb = Object.assign(Breadcrumbs, {Item: BreadcrumbsItem})
+
+/**
+ * @deprecated Use the BreadcrumbsProps type instead
```
colebemis, line 94:
```suggestion
 * @deprecated Use the `BreadcrumbsProps` type instead
```

src/Breadcrumbs.tsx
```diff
@@ -79,9 +78,24 @@ const BreadcrumbItem = styled.a.attrs<StyledBreadcrumbItemProps>(props => ({
...
+ */
+export type BreadcrumbProps = ComponentProps<typeof BreadcrumbsBase>
+
+/**
+ * @deprecated Use the BreadcrumbsItemProps type instead
```
colebemis, line 99:
```suggestion
 * @deprecated Use the `BreadcrumbsItemProps` type instead
```

Issue #1431, opened 2021-09-16
Enforce consistent component documentation page structure


colebemis:
## Problem

The structure of our component documentation pages is inconsistent. 

* Some pages are missing component descriptions, source links, statuses
* Headings are named inconsistently (e.g. `Component props` vs `Props` vs `AvatarStack Component props`)
* Headings are ordered inconsistently (e.g. sometimes props appear before examples and vice versa)

<details>
  <summary>Examples</summary>

| Box | AvatarStack | Details |
| --- | --- | --- |
| ![image](https://user-images.githubusercontent.com/4608155/133693011-1516cb8d-d461-4c8e-97d1-43553fbc1c93.png) | ![image](https://user-images.githubusercontent.com/4608155/133693199-b05fbb47-9973-4a84-b5b2-9ec0b842615d.png) | ![image](https://user-images.githubusercontent.com/4608155/133693305-f6353fa0-c528-4070-b6f2-cf47f4aeb908.png) |

</details>

This inconsistency stems from a lack of guidelines and tooling around component documentation pages.

## Why is this important?

A prerequisite to moving a component to [stable](https://primer.style/contribute/component-lifecycle#stable) is that the documentation for the component is "complete." However, we don't have any guidelines or tooling that document what "complete" documentation looks like for a Primer React component. We need those guidelines before we can consider components "stable."

## Tasks

- [ ] Decide on a structure for component documentation pages
  - See [Radix Primitives](https://www.radix-ui.com/docs/primitives/components/accordion), [Braid](https://seek-oss.github.io/braid-design-system/components/Accordion), [Spectrum](https://react-spectrum.adobe.com/react-spectrum/ActionButton.html), [Primer ViewComponents](https://primer.style/view-components/components/beta/autocomplete), etc. for inspiration
  - Chat with folks working with Primer React and Primer ViewComponents
- [ ] Document the structure for component documentation pages
- [ ] Enforce documentation structure (maybe using a linter?)

siddharthkp:
Should we bring back the template from https://github.com/primer/react/pull/1452?

colebemis:
This will be resolved in the new Primer React docs

PR #1452, opened 2021-09-22
Component document template

siddharthkp:
Context: https://github.com/primer/react/issues/1431

Added a new `Template` to add some structure to how we document components. We can fork the template and fill in copy + example while keeping the structure intact.

Used `ActionMenu` as the example for template so that we can talk about it with real examples.

[Start with the checklist](https://github.com/primer/react/pull/1452/files#diff-01f9c9f0a87db01320bb5442c78f038f2873533c9428ea4d7532eb2bab5ac4fbR280)

and then leep these two side by side: 
1. [Here's a deployed version](https://primer-components-ov4oy9dol-primer.vercel.app/react/Template)
2. [Accompanying notes in markdown](https://github.com/primer/react/pull/1452/files#diff-01f9c9f0a87db01320bb5442c78f038f2873533c9428ea4d7532eb2bab5ac4fb)

Note: For now, I've added the template to side nav, but it would be nice if we only see it in DEV mode.


siddharthkp:
Even though I think it's important to show realistic examples, they might run too long on the page. We can fold our examples and show one at a time with Tabs or Select - or a good old details/summary combo.

docs/content/Template.mdx
```diff
@@ -0,0 +1,313 @@
...
+import {Props} from '../src/props'
+import {ActionMenu} from '@primer/components'
+
+<!-- List root component and it's children -->
+
+<Props of={ActionMenu} />
+
+<br />
+
+## Best practices
```
colebemis, line 229: These best practices don't seem implementation-specific. I wonder if we should just link to the corresponding documentation in our [interface guidelines](https://primer.style/design/components/action-list) to avoid duplicating content.

docs/content/Template.mdx
```diff
@@ -0,0 +1,313 @@
+---
+title: Action Menu
+status: Alpha
+---
+
+<!-- Start with a Description about the `Component` and when would you use it -->
+
+An action menu is a vertical list of interactive actions. It’s composed of items presented in a consistent single-column format, with room for icons, descriptions, side information, and other rich visuals.
```
colebemis, line 8: We have a frontmatter variable for descriptions that we could use here:
```suggestion
status: Alpha
# Start with a Description about the `Component` and when would you use it -->
description: An action menu is a vertical list of interactive actions. It’s composed of items presented in a consistent single-column format, with room for icons, descriptions, side information, and other rich visuals.
```

docs/content/Template.mdx
```diff
@@ -0,0 +1,313 @@
+---
+title: Action Menu
```
colebemis, line 2: Should we use the component name here? `ActionMenu` vs `Action Menu`

docs/content/Template.mdx
```diff
@@ -0,0 +1,313 @@
...
+<!-- Start with a Description about the `Component` and when would you use it -->
+
+An action menu is a vertical list of interactive actions. It’s composed of items presented in a consistent single-column format, with room for icons, descriptions, side information, and other rich visuals.
+
+<br />
+
+<!-- Preview up top, so that folks can figure out if this is the component they want -->
+<!-- Rendered version is ideal, but an image is a good fallback -->
+
+![Action menu preview](https://user-images.githubusercontent.com/293280/123880964-95b3a700-d8f8-11eb-9775-cbaf165207ed.png)
```
colebemis, line 15: I would love for this to be a live code example. Maybe something similar to the Radix Primitives examples: https://www.radix-ui.com/docs/primitives/components/accordion

siddharthkp: Yeeep! 

I struggled to create a `defaultOpen` action menu that doesn't interfere with scroll! I think I have new hacks now, I can give another try!

docs/content/Template.mdx
siddharthkp: I'm split here. On one hand, yes absolutely! On the other, I don't know if folks will read the other page before copying examples :(

PR #1458, opened 2021-09-23
Add script to publish storybook in docs

siddharthkp:
Right now, there is not easy way to share our stories. This publishes storybook right in the docs. Adds about 45s to our deploy time

Here's a preview: https://primer-components-lwncezwdw-primer.vercel.app/react/storybook

Closes https://github.com/github/primer/issues/300


docs/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "clean": "gatsby clean",
     "develop": "gatsby develop",
-    "build": "pushd .. && npm install && popd && cross-env NODE_ENV=production gatsby build --prefix-paths"
+    "build": "pushd .. && npm install && popd && cross-env NODE_ENV=production gatsby build --prefix-paths && npm run add:storybook",
+    "add:storybook": "cd .. && ./script/deploy-storybook"
```
colebemis, line 9: Should we call this `build` instead of `add`/`deploy` since that's closer to what's happening in these scripts?
```suggestion
    "build:storybook": "cd .. && ./script/build-storybook"
```

colebemis (approved):
Thank you, @siddharthkp ❤️  This is great!

docs/package.json
siddharthkp: Yeah, I wasn't sure if i should call it build, because it's so focused on building it inside the docs and doesn't work standalone.

maybe we should call it `publish-storybook`?

colebemis: Ah, I understand your hesitation but `deploy` and `publish` both make me think the script might actually deploy or publish the site (e.g. make some calls to Vercel). 

`build` seems fine to me since we only ever need to build storybook inside our docs. Also having this script inside the `docs/package.json` implies that the build is specifically for docs.

siddharthkp: Makes sense! Changed to build and shipped :shipit:

Issue #1454, opened 2021-09-22
ActionList.Item should accept a ref


jclem:
I'm proposing using `forwardRef` in `ActionList.Item`.

There may be a simpler way of doing this, but when attempting to wrap an `ActionList.Item` in a Next.js link like so:

```tsx
import NextLink from 'next/link'
import {Item, ItemProps} from '@primer/components/lib/ActionList/Item'

export const MyComponent = (itemProps: ItemProps) => (
  <NextLink href="/foo" passHref>
    <Item as="a" {...itemProps} />
  </NextLink>
)
```

...a warning is emitted because `NextLink` tries to pass a ref to its child. I expect use cases like this to be somewhat common with `ActionList.Item`, so I'm wondering if it makes sense to use `forwardRef` in `ActionList.Item`.

My current workaround for this issue works okay, but it involves adding extraneous DOM elements (it also doesn't actually have the `ref` in the right place, but it avoids the warning for now):

```tsx
import {ItemProps} from '@primer/components/lib/ActionList'
import {Item} from '@primer/components/lib/ActionList/Item'
import {ComponentType, forwardRef} from 'react'

// Add "as" back to Item props
// SEE: https://github.com/primer/react/issues/1413
type FixedItemProps = ItemProps & {as?: string | React.ComponentType<unknown>}
type FixedItem = ComponentType<FixedItemProps>
export const FixedItem = Item as FixedItem

export const FixedActionListItem = forwardRef<HTMLDivElement, FixedItemProps>(
  (props, ref) => (
    <div ref={ref}>
      <FixedItem {...props} />
    </div>
  )
)

FixedActionListItem.displayName = 'FixedActionListItem'
```

Issue #1413, opened 2021-09-08
Allow ActionList.Item to be an anchor


jfuchs:
## Description

Allow ActionList.Item to be a link, so that an ActionList could be used as a navigation menu

### Component API

Possibilities:

```
<Item as={Link} href={...} />
```

```
<LinkItem href={...} />
```

mattcosta7:
As far as the component API - I think the case of

`<Item as={Link} href="" />` might be better generalized as `<Item as={Link} {...LinkProps} />` which would encapsulate the react-router scenario of `<Item as={ReactRouterLink} {...ReactRouterLinkProps} />`

jclem:
To generalize upon what @mattcosta7 was saying, `<Item as={T} {...P} />` where `T` is `string | React.ComponentType` and `P` is `React.ComponentProps<T> & ItemProps`.

smockle:
From @mattcosta7 in https://github.com/primer/react/issues/1413#issuecomment-915409164:
> As far as the component API - I think the case of
> 
> `<Item as={Link} href="" />` might be better generalized as `<Item as={Link} {...LinkProps} />` which would encapsulate the react-router scenario of `<Item as={ReactRouterLink} {...ReactRouterLinkProps} />` (which would use to={LinkDescriptor} instead of href directly)

From @jclem in https://github.com/primer/react/issues/1413#issuecomment-915413364:
> To generalize upon what @mattcosta7 was saying, `<Item as={T} {...P} />` where `T` is `string | React.ComponentType` and `P` is `React.ComponentProps<T> & ItemProps`.

We’ve [previously explored type-safe `as`](https://github.com/github/primer/discussions/130#discussioncomment-696081) using [`@radix-ui/react-polymorphic`](https://www.radix-ui.com/docs/primitives/utilities/polymorphic). Here’s a quick demo of how this could work: https://codesandbox.io/s/as-prop-demo-forked-ox2b1?file=/src/App.tsx

jfuchs:
Given these usage options...

```
        <ActionList
          items={[
            {text: 'A) Vanilla link', as: 'a', href: '/about'},
            {text: 'B) as React Router link', href: '/about', as: ReactRouterLikeLink},
            {text: 'C) React Router link via renderItem', href: '/about', renderItem: props => (
              <ReactRouterLikeLink component={ActionList.Item} {...props} />
            )},
            {text: 'D) NextJS style via renderItem', renderItem: props => (
              <NextJSLikeLink>
[...]

PR #1463, opened 2021-09-24
ActionList.Item accepts a polymorphic 'as' prop

jfuchs:
This PR adds an `as` prop to ActionList.Item so that Primer users can put HTML `Anchor`s, Next.js `Link`s, React Router `Link`s, or whatever else in their action lists.

I am adding a dependency (@radix-ui/react-polymorphic) for some type definitions. The library is:
- MIT licensed
- Pretty widely used
- Has no dependences
- Shouldn't add any bytes to bundles as it is just types

Closes https://github.com/primer/react/issues/1413
Closes https://github.com/primer/react/issues/1454

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/ActionList/List.tsx
```diff
@@ -162,15 +164,14 @@ export const List = React.forwardRef<HTMLDivElement, ListProps>((props, forwarde
...
-    const key = itemProps.key ?? itemProps.id?.toString() ?? itemIndex.toString()
+    const key = itemProps.id?.toString() ?? itemIndex.toString()
     return (
       <ItemComponent
         showDivider={props.showItemDividers}
         selectionVariant={props.selectionVariant}
         {...itemProps}
         key={key}
         sx={{...itemStyle, ...itemProps.sx}}
-        item={item}
```
jfuchs, line 173: @dgreif after my changes, `tsc` was complaining about this line, and I honestly couldn't figure out what the intent was here. Any ideas? Is it safe to remove?

dgreif: It's used by the callback handlers for onAction if I remember correctly.  This comes back to the object equivalence issue for selection.  We couldn't just pass an item "like" the one that was selected, we need the actual item that caused the `ItemComponent` to get rendered in the first place.  That whole system has made a lot of things more complicated then I would like them to be 😅

jfuchs: Ohh that makes sense. I was wondering about that callback too. Any idea why the item in the callback was necessary?

jfuchs: (that is, I can't picture the usage where this is helping simplify things...why wouldn't the user just keep their item in the closure of the callback?)

src/ActionList/List.tsx
```diff
@@ -162,7 +164,7 @@ export const List = React.forwardRef<HTMLDivElement, ListProps>((props, forwarde
   const renderItem = (itemProps: ItemInput, item: ItemInput, itemIndex: number) => {
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     const ItemComponent = ('renderItem' in itemProps && itemProps.renderItem) || props.renderItem || Item
-    const key = itemProps.key ?? itemProps.id?.toString() ?? itemIndex.toString()
+    const key = itemProps.id?.toString() ?? itemIndex.toString()
```
smockle, line 167: Non-blocking: Out of curiosity, why was this change to `key` necessary?

src/stories/ActionList.stories.tsx
```diff
@@ -362,3 +362,59 @@ export function SizeStressTestingStory(): JSX.Element {
...
+const NextJSLikeLink = forwardRef(
+  ({href, children}: {href: string; children: React.ReactNode}, ref): React.ReactElement => {
+    const child = React.Children.only(children)
+    const childProps = {
+      ref,
+      href
+    }
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    return React.cloneElement(child, childProps)
```
smockle, line 383: Does this accomplish the same result without a `@ts-ignore`?
```suggestion
    return (
      <>
        {React.isValidElement(child)
          ? React.cloneElement(child, childProps)
          : null}
      </>
    );
```

src/ActionList/List.tsx
jfuchs: Good catch — this was my mistake. TS caught that `key` wasn't on `ItemInput` so I removed the line intending to circle back to it. Will fix!

src/ActionList/List.tsx
dgreif: I think the item in the callback was for the same reason.  Higher components listening to the callback need the exact item to store which thing got selected

docs/content/ActionList.mdx
```diff
@@ -62,11 +65,36 @@ An `ActionList` is a list of items which can be activated or selected. `ActionLi
...
-| showItemDividers | `boolean`                           |      `false`      | Optional. If `true` dividers will be displayed above each `ActionList.Item` which does not follow an `ActionList.Header` or `ActionList.Divider`        |
+| Name             | Type                                |               Default               | Description                                                                                                                                             |
+| :--------------- | :---------------------------------- | :---------------------------------: | :------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------- |
+| items            | `Array<ItemProps                    | (props: ItemProps) => JSX.Element>` | `undefined`                                                                                                                                             | Required. A list of item objects conforming to the `ActionList.Item` props interface. |
```
colebemis, line 97: Looks like the `|` in the type is confusing Markdown: 
CleanShot 2021-09-29 at 09 05 53

src/ActionList/Item.tsx
```diff
@@ -124,6 +126,21 @@ export interface ItemProps extends Omit<React.ComponentPropsWithoutRef<'div'>, '
...
+
+  /**
+   * The ARIA role describing the function of `List` component. `option` is a common value.
+   */
+  role?: AriaRole
+
+  /**
+   * An item to pass back in the `onAction` callback, meant as
+   */
+  item?: ItemInput
```
colebemis, line 143: Can you explain why we needed to add these props?

src/ActionList/Item.tsx
```diff
@@ -457,4 +478,6 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
       </DividedContent>
     </StyledItem>
   )
-}
+}) as PolymorphicForwardRefComponent<'div', ItemProps>
+
+export {Item}
```
colebemis, line 483: Is there a reason you wrote the export like this instead of doing: `export const Item ...` on line 318

src/ActionList/List.tsx
```diff
@@ -8,7 +8,9 @@ import {get} from '../constants'
 import {SystemCssProperties} from '@styled-system/css'
 import {hasActiveDescendantAttribute} from '../behaviors/focusZone'
 
-export type ItemInput = ItemProps | (Partial<ItemProps> & {renderItem: typeof Item})
+type RenderItemFn = (props: ItemProps) => React.ReactElement
+
+export type ItemInput = ItemProps | ((Partial<ItemProps> & {renderItem: RenderItemFn}) & {key?: Key})
```
colebemis, line 13: Just curious, what's the reason for including `key` here?

docs/content/ActionList.mdx
```diff
@@ -2,6 +2,9 @@
 title: ActionList
 ---
 
+import {Props} from '../src/props'
+import {ActionList} from '@primer/components'
+
```
colebemis, line 7: Looks like we're not using these (😢 )
```suggestion
```

docs/content/ActionList.mdx
```diff
@@ -2,6 +2,9 @@
 title: ActionList
 ---
```
colebemis, line 3:
```suggestion
title: ActionList
status: Alpha
source: https://github.com/primer/react/tree/main/src/ActionList
---
```

src/stories/ActionList.stories.tsx
```diff
@@ -362,3 +362,59 @@ export function SizeStressTestingStory(): JSX.Element {
...
+            {
+              text: 'D. NextJS style',
+              renderItem: props => (
+                <NextJSLikeLink href="/about">
+                  <ActionList.Item as="a" {...props} />
+                </NextJSLikeLink>
+              )
+            }
+          ]}
+        />
```
colebemis, line 415: Could we add a Jest test that will fail CI if `ActionList.Item` doesn't render the element passed through the `as` prop correctly?

I want to make sure we don't accidentally break this if we refactor `ActionList.Item` (which will likely happen soon)

src/ActionList/Item.tsx
jfuchs: These three fields were because TSC told me to :-) ... but also, they all make sense and I wonder if they only worked before because of gaps in the type coverage

src/ActionList/Item.tsx
jfuchs: Nope, will fix.

src/ActionList/List.tsx
jfuchs: Same thing as the new props for item — added because tsc caught issues that it wasn't catching before. in this case, it's something we allow in an ItemInput but remove before the ItemInput becomes ItemProps.

src/stories/ActionList.stories.tsx
jfuchs: Just added a `behavesAsComponent` test

docs/content/ActionList.mdx
```diff
@@ -62,11 +64,36 @@ An `ActionList` is a list of items which can be activated or selected. `ActionLi
...
+        <NextJSLikeLink href="/about">
+          <ActionList.Item as="a" {...props} />
+        </NextJSLikeLink>
+      )
+    }
+  ]}
+/>
+```
+
+## ActionList props
```
colebemis, line 92: We just call this "Props" on other pages.
```suggestion
## Props
```

src/ActionList/Item.tsx
```diff
@@ -457,4 +478,6 @@ export function Item(itemProps: Partial<ItemProps> & {item?: ItemInput}): JSX.El
       </DividedContent>
     </StyledItem>
   )
-}
+}) as PolymorphicForwardRefComponent<'div', ItemProps>
+
+Item.displayName = 'Item'
```
colebemis, line 483: Should the display name actually be `ActionList.Item`?
```suggestion
Item.displayName = 'ActionList.Item'
```

colebemis (approved):
Looks great. Excellent work, @jfuchs ✨

PR #1466, opened 2021-09-27
Simple select button for colorschemes and themeswitcher component

### LINKED ISSUE #1433: Preview components in different themes

pksjce:
This is a documentation only change for primer/react documentation website.
Added a ThemeSwitcher to the doctocat custom components.
Select dropdown and logic to switch between colorSchemes.
Decide which colorMode based on colorSchemes.

*Questions*
1. I'd like to use the Dropdown component, still trying to figure out which component triggers the onChange event.
2. The relation between colorMode and colorSchemes is not well defined. The type is just a string. Is this going to be fixed in the future?

Closes #1433

### Screenshots

https://user-images.githubusercontent.com/417268/134846970-adc04d04-96b0-4a89-a57c-24864c961934.mp4

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


docs/src/@primer/gatsby-theme-doctocat/components/theme-switcher.js
```diff
@@ -0,0 +1,51 @@
...
+  //         dayScheme: 'light',
+  //         nightScheme
+  //       })
+  //     }}
+  //   ></Dropdown.Menu>
+  // </Dropdown>
+  return (
+    <select
+      style={{float: 'right'}} // sorry
+      onBlur={event => {
```
siddharthkp, line 29: Hi hi hi!

Should we use `onChange` instead of `onBlur` so that it changes instantly. (It does change in your demo gif though!

This is what I see on Chrome:
Kapture 2021-09-27 at 14 01 38
)


```suggestion
      onChange={event => {
```

siddharthkp:
> I'd like to use the Dropdown component, still trying to figure out which component triggers the onChange event.

⚠️ I have no idea what I'm talking about, trying to find solutions by looking at the docs and codebase. @colebemis should correct me when I'm wrong!

Yeah, it looks the Dropdown component doesn't have an onChange/onSelect and expects an onClick on the (list) item??

But, `DropdownMenu` does: https://primer.style/react/DropdownMenu

which expects items in the [Item format](https://github.com/primer/react/blob/main/src/ActionList/Item.tsx#L51) from [`ActionList`](https://primer.style/react/ActionList) 😅 

```jsx
<DropdownMenu
  renderAnchor={anchorProps => (
    <Button {...anchorProps}> {colorScheme} <Dropdown.Caret /> </Button>
  )}
  items={items}
  onChange={item => { }
/>
```

colebemis:
This is great, @pksjce! ✨ 

A couple notes:
* `Dropdown` is being replaced by [`DropdownMenu`](https://primer.style/react/DropdownMenu) (this is not obvious, sorry!). We should use `DropdownMenu` here instead of `Dropdown`
* The complexity around `color modes` vs `color schemes` is to support these kinds of complex theme preference UIs in github.com: 
CleanShot 2021-09-27 at 12 13 58

  Since we don't need a complex theme preference UI for these documentation examples, we can keep the `colorMode` and `nightScheme` fixed and just switch the `dayScheme`. Here's an example: https://codesandbox.io/s/elastic-tereshkova-t65hw?file=/src/App.tsx

docs/src/@primer/gatsby-theme-doctocat/components/theme-switcher.js
pksjce: oops sorry! There was a wierd eslint rule on `<select />` suggesting me to change to `onBlur` I remember. Anyway, this shouldn't be an issue with the `DropdownMenu` component.

pksjce:
@siddharthkp, @colebemis - Thanks for the `DropdownMenu` component tip! That one totally works.
@colebemis - I think I understand the colorSchemes a lot better. I had not customised Github themes before now.

This is what it looks like now.
Screenshot (14)

pksjce:
Hey @colebemis - I made a PR for the fix you proposed here https://github.com/primer/react/pull/1470/files

I'd still like to say, I like my initial solution without the `setDayScheme` simplification because,
1. I don't have to wait for above PR to merge
2. In the documentation, it would be nice if only the colors of the live preview changed and the dropdown remains like the rest of the page.

What do you think?

colebemis:
> In the documentation, it would be nice if only the colors of the live preview changed and the dropdown remains like the rest of the page.
> 
> What do you think?

I'm going to loop @mperrotti in on this for some design guidance 😄

mperrotti:
I don't like the way the theme switcher creates extra whitespace above the component examples. It makes the box feel imbalanced because it's top-heavy.

I'd recommend something like this:
https://user-images.githubusercontent.com/2313998/135145645-06b91fde-9bf4-407f-a665-8cb16d21b4ca.mp4


I shared this with @colebemis and he's into it.

pksjce:
I love this @mperrotti! great advice! will modify my PR to get this working :)

pksjce:
Here's what it looks like

https://user-images.githubusercontent.com/417268/135366284-72014d8a-31e0-431b-91da-6bda2a59124b.mp4

colebemis:
This is looking great! Nice work, @pksjce. And thanks for the design help, @mperrotti ❤️ 

Noticed a few quirks with showing and hiding the dropdown button:

* If your cursor leaves the dropdown menu, the dropdown button goes away but the menu doesn't. We probably shouldn't hide the button if the menu is open.

  CleanShot 2021-09-29 at 21 36 19

  A potential fix for this is to lift the state of the dropdown menu into the wrapper component (with the `open` and `onOpenChange` props) and make sure that we don't hide the button if the menu is open:

   ```jsx
   function ThemeSwitcher({open, onOpenChange}) {
     return (
       <DropdownMenu
         open={open}
         onOpenChange={onOpenChange}
         ...
       />
     )
   }
   
   function LivePreviewWrapper({children}) {
     const [isDropdownOpen, setIsDropdownOpen] = React.useState(false)
     const [showSwitcher, setShowSwitcher] = React.useState(false)
     // Now that we control this state
     // we can make sure to never call `setShowSwitcher(false)`
     // if `isDropdownOpen` is `true`
     return (
       ...
       <ThemeSwitcher open={isDropdownOpen} onOpenChange={setIsDropdownOpen} />
       ...
     )
   }
   ```
* I also noticed that the button doesn't go away when the focus leaves the example. I wouldn't call this a blocker. I just wanted to point it out.

   CleanShot 2021-09-29 at 21 47 10

docs/src/@primer/gatsby-theme-doctocat/components/live-preview-wrapper.js
```diff
@@ -1,38 +1,58 @@
...
+        }}
+        onMouseLeave={() => {
+          setShowSwitcher(false)
+        }}
+        width="100%"
+        bg="canvas.default"
+        position="relative"
+        sx={{borderTopLeftRadius: 2, borderTopRightRadius: 2}}
+      >
+        <Box p={2} display={showSwitcher ? '' : 'none'} zIndex="1" position="absolute" right="2">
```
colebemis, line 48: Let's adjust the spacing here to match the "copy to clipboard" button:

| Before | After |
| --- | --- |
| CleanShot 2021-09-29 at 21 51 24@2x | CleanShot 2021-09-29 at 21 50 37@2x |



```suggestion
        <Box p={2} display={showSwitcher ? '' : 'none'} zIndex="1" position="absolute" top={0} left={0}>
```

pksjce: Done!

pksjce:
@colebemis - 
1. Good catch with the dropdownOpen. Was able to fix it with the directions. ❇️ 
2. About the focus, yes I noticed it as I implemented it. I decided to avoid the complexity with onBlur as there was some flickering as each of the items in the live preview received and lost focus. Will give it a go once again with a fresh mind if there is a simpler fix. I also did not think it was a blocker because it happens for keyboard users only.

docs/src/@primer/gatsby-theme-doctocat/components/live-preview-wrapper.js
```diff
@@ -1,38 +1,60 @@
...
+        <DropdownButton variant="small" {...anchorProps}>
+          {children}
+        </DropdownButton>
+      )}
+      items={items}
+      selectedItem={selectedItem}
+      onChange={item => {
+        setDayScheme(item.key)
+        // If selected remove itself after few seconds.
+        setTimeout(() => setShowSwitcher(false), 3000)
```
colebemis, line 21: I noticed another quirk related to this setTimeout. If you select a theme, then reopen the dropdown menu before the end of the `setTimeout, the button will disappear while the dropdown menu is open:

https://user-images.githubusercontent.com/4608155/135488293-beeab3fd-e1df-417a-bcde-abaf6727e0dd.mp4

There are a couple of ways I think we could fix this:
1. We could remove the `setTimeout` and not worry about hiding the button after selection.
2. We could pass the `open` state to the `ThemeSwitcher` and check that the dropdown isn't open before calling `setShowSwitcher(false)` in the `setTimeout`.

   ```jsx
   function ThemeSwitcher({isDropdownOpen, setIsDropdownOpen, setShowSwitcher}) {
     ...
     return (
       <DropdownMenu
         open={isDropdownOpen}
         onOpenChange={setIsDropdownOpen}
         onChange={() => {
           // If selected remove itself after few seconds.
           setTimeout(() => {
             if (!isDropdownOpen) {
               setShowSwitcher(false)
             }
           }, 3000)
         }
       />
     )
   }
   ```

I don't have a strong opinion about which option we should choose. @pksjce or @mperrotti do you have a preference?

pksjce: 🤔 Lets unship that setimeout behaviour. Will keep an eye out for any feedback once this is published.

colebemis (approved):
This is _seriously_ awesome. Excellent work, @pksjce! 💖  And thanks for your help, @mperrotti 🙇

PR #1468, opened 2021-09-27
AvatarStack no longer accepts styled props

jfuchs:
See https://github.com/github/primer/issues/296

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/selfish-deers-change.md
+++ b/.changeset/selfish-deers-change.md
@@ -2,4 +2,4 @@
 '@primer/components': major
 ---
 
-AvatarStack no longer accepts styled props
+AvatarStack no longer accepts styled props or DOM props
--- a/.changeset/selfish-deers-change.md
+++ b/.changeset/selfish-deers-change.md
@@ -1,0 +1,5 @@
+---
+'@primer/components': major
+---
+
+AvatarStack no longer accepts styled props
--- a/docs/content/AvatarStack.mdx
+++ b/docs/content/AvatarStack.mdx
@@ -26,18 +26,10 @@ AvatarStack is used to display more than one Avatar in an inline stack.
 </AvatarStack>
 ```
 
-## System props
-
-<Note variant="warning">
-
-System props are deprecated in all components except [Box](/Box). Please use the [`sx` prop](/overriding-styles) instead.
-
-</Note>
-
-AvatarStack components get `COMMON` system props, AvatarStack.Item components do not get any system props. Read our [System Props](/system-props) doc page for a full list of available props.
-
 ## AvatarStack Component props
 
-| Name       | Type    | Default | Description                       |
-| :--------- | :------ | :-----: | :-------------------------------- |
-| alignRight | Boolean |         | Creates right aligned AvatarStack |
+| Name       | Type              | Default | Description                          |
+| :--------- | :---------------- | :-----: | :----------------------------------- |
+| alignRight | Boolean           |         | Creates right aligned AvatarStack    |
+| children   | React.Node        |         | The avatars                          |
+| sx         | SystemStyleObject |   {}    | Style to be applied to the component |
--- a/docs/content/AvatarStack.mdx
+++ b/docs/content/AvatarStack.mdx
@@ -31,4 +31,7 @@ AvatarStack is used to display more than one Avatar in an inline stack.
 
 ## Props
 
-<Props of={AvatarStack} />
+| Name       | Type              | Default | Description                          |
+| :--------- | :---------------- | :-----: | :----------------------------------- |
+| alignRight | Boolean           |         | Creates right aligned AvatarStack    |
+| sx         | SystemStyleObject |   {}    | Style to be applied to the component |
--- a/docs/content/AvatarStack.mdx
+++ b/docs/content/AvatarStack.mdx
@@ -2,6 +2,9 @@
 title: AvatarStack
 ---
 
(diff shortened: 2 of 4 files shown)


smockle (approved):

docs/content/AvatarStack.mdx
```diff
@@ -26,18 +26,10 @@ AvatarStack is used to display more than one Avatar in an inline stack.
...
-
 ## AvatarStack Component props
 
-| Name       | Type    | Default | Description                       |
-| :--------- | :------ | :-----: | :-------------------------------- |
-| alignRight | Boolean |         | Creates right aligned AvatarStack |
+| Name       | Type              | Default | Description                          |
+| :--------- | :---------------- | :-----: | :----------------------------------- |
+| alignRight | Boolean           |         | Creates right aligned AvatarStack    |
+| children   | React.Node        |         | The avatars                          |
```
smockle, line 34: Non-blocking: Do we typically include `children` (since every? React component supports it)?

jfuchs: Doesn't seem like we generally do. Would it be useful to include it in places where we have an expectation of a certain kind of component? e.g.,:

```
| children   | React.Node        |         | Generally an `Avatar`     |
```

colebemis: Does the `<Props>` component work for this component?

jfuchs:
If we want to get rid of styled props, but we still want to support `as`... do we also want to all PRC components to pass through props that the `AsComponent` supports?

A contrived example: 
```typescript
<AvatarStack as="a" href="/foo">{...}</AvatarStack>
```

docs/content/AvatarStack.mdx
jfuchs: Good question — I'll check.

jfuchs: Yep, will switch over to that.

src/AvatarStack.tsx
```diff
@@ -140,17 +137,18 @@ const transformChildren = (children: React.ReactNode) => {
...
-const AvatarStack = ({children, alignRight, ...rest}: AvatarStackProps) => {
+const AvatarStack = ({children, alignRight, sx: sxProp}: AvatarStackProps) => {
   const count = React.Children.count(children)
   const wrapperClassNames = classnames({
     'pc-AvatarStack--two': count === 2,
     'pc-AvatarStack--three-plus': count > 2,
     'pc-AvatarStack--right': alignRight
   })
   return (
-    <AvatarStackWrapper count={count} className={wrapperClassNames} {...rest}>
```
colebemis, line 153: This change prevents AvatarStack from accepting any DOM props (e.g. `id`, `className`, `data-*`). I would love for us to document how we decide when a component should accept DOM props. Maybe it's appropriate for some components to accept all DOM props (e.g. `Link`) and other components should only accept a subset of DOM props. 

We should also make it obvious in our component documentation pages if a component accepts DOM props or not. 

This is a bigger discussion that we can probably have outside of this PR.

For now, I'm fine with moving forward with this PR as-is. It might help us expose places where DOM props are necessary.

.changeset/selfish-deers-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/components': major
+---
+
+AvatarStack no longer accepts styled props
```
colebemis, line 5: Let's document that AvatarStack no longer accepts DOM props as well.

```suggestion
AvatarStack no longer accepts styled props or DOM props
```

colebemis (approved):

colebemis:
Hmm, looks like prop documentation isn't working 😢 

CleanShot 2021-09-28 at 10 45 41

src/AvatarStack.tsx
jfuchs: Good call to make it clear in docs.

Proposal: The set of components that take an `as` prop should be the same set that pass through DOM props. (I can't see any use case where we'd want those sets to diverge).

colebemis: That seems reasonable to me. I'm trying to think of components that might not accept the `as` prop but should still accept DOM props. What about form components? (like `TextInput`)

jfuchs:
> Hmm, looks like prop documentation isn't working 😢
> 
> CleanShot 2021-09-28 at 10 45 41

How strange... it's working locally. I'll investigate!

docs/content/AvatarStack.mdx
```diff
@@ -2,6 +2,9 @@
 title: AvatarStack
 ---
 
+import {Props} from '../src/props'
+import {AvatarStack} from '@primer/components'
+
 AvatarStack is used to display more than one Avatar in an inline stack.
```
colebemis, line 8: This is not necessary for this PR, but while we're here it might be nice to add some more metadata to the docs:
```suggestion
title: AvatarStack
status: Alpha
source: https://github.com/primer/react/blob/main/src/AvatarStack.tsx
description: Use avatar stack to display multiple avatars in an inline stack.
---

import {Props} from '../src/props'
import {AvatarStack} from '@primer/components'
```

jfuchs: 😍 I would like that very much. I'm happy to work toward that in a separate stream of work.

PR #1470, opened 2021-09-28
Fix item color to not depend on inherit

pksjce:
In https://github.com/primer/react/pull/1466, we ran into an issue with dropdown menu items inheriting colors wrongly instead of picking up the theme colors.
I could not reproduce it in the storybooks, but look at original issue [here](https://github.com/primer/react/pull/1466#issuecomment-928673630)

You can look at the reproduction here too
https://codesandbox.io/s/upbeat-moore-8mot8?file=/src/App.tsx

Screenshot (16)

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Questions

1. Need help to bump primer/react version.
2. Not sure why I can't reproduce this in storybooks.

### DIFF
--- a/.changeset/sixty-jokes-compete.md
+++ b/.changeset/sixty-jokes-compete.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Fix ActionList.Item color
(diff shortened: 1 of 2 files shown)


siddharthkp (approved):
This is perfect! ✨

colebemis (approved):
Woo 🎉

PR #1471, opened 2021-09-28
feat: Focus the primary action in 'ConfirmationDialog'

smockle:
Alternative to #1472. This PR makes [`focusTrap`’s `initialFocus`](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/behaviors/focusTrap.ts#L82-L84) handle action focus and retains the [hack in `Dialog.Buttons`](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/Dialog/Dialog.tsx#L345) which handles action focus in nested `focusTrap`s (such as the one in [`ShorthandHookFromActionMenu`](https://github.com/primer/react/blob/f5a7c2e7ec6ce884df61d35c1c0b5d59ab3da1ba/src/stories/ConfirmationDialog.stories.tsx#L83-L105)).

Fixes https://github.com/github/primer/issues/313

ℹ️ You might wonder–“Why wasn’t moving [the `autoFocus` attribute](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/Dialog/ConfirmationDialog.tsx#L121) from `cancelButton` to `confirmButton` in `ConfirmationDialog.tsx` sufficient?” Great question! I wondered the same thing at first. It turns out that [this special case in `focusTrap`](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/behaviors/focusTrap.ts#L91) causes the primary-action button (`confirmButton`) to receive `tabIndex: -1`, so the secondary-action button (`cancelButton`) still receives focus, even if `autoFocus: true` is moved to `confirmButton`.

### Screenshots

Screen recording demonstrating the primary action is focused when a 'ConfirmationDialog' opens:
Screen recording demonstrating the primary action is focused when a 'ConfirmationDialog' opens

We generally follow platform conventions, and primary-action-focus is the default in macOS confirmation dialogs (screen recording of Pages below):
Confirmation dialog in Pages, with the primary action focused by default

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


colebemis (approved):
Looks good to me 👍  Thanks, @smockle!

Can you add a changeset?

colebemis:
Is there a way we can write a test to check if the confirm button is focused when the dialog is opened? It be nice to catch any accidental regressions in CI.

smockle:
From @colebemis in https://github.com/primer/react/pull/1471#issuecomment-929540227:
> Is there a way we can write a test to check if the confirm button is focused when the dialog is opened? It be nice to catch any accidental regressions in CI.

Yes there is! Added tests in https://github.com/primer/react/pull/1471/commits/3abb6364f03f1899b9a7c2c50f77053608a5a11d.

Check it out—the #1471 CI check [passed](https://github.com/primer/react/runs/3737480709?check_suite_focus=true), and the #1472 CI check [failed](https://github.com/primer/react/runs/3737492361?check_suite_focus=true)—the new tests catch the problem with #1472!

dgreif (approved):
Changes look good.  For reference, I'm fairly certain that @T-Hugs set it up this way intentionally.  If I remember correctly (it's been a while), the mindset was to not have users _accidentally_ trigger the submit button, but instead require them to move focus to it and then trigger it.  Maybe he will weigh in since this is OSS 😁.  In any case, I'm fine with making the change as it does match the behavior of most  confirmation dialogs out there in the world.

T-Hugs:
Indeed, the old behavior was intentional for the reason @dgreif stated. That said, I was never passionate one way or the other!

src/__tests__/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+        </BaseStyles>
+      </SSRProvider>
+    </ThemeProvider>
+  )
+}
+
+describe('ConfirmationDialog', () => {
+  behavesAsComponent({
+    Component: ConfirmationDialog,
+    systemPropArray: [COMMON, LAYOUT],
```
colebemis, line 80:
```suggestion
```

src/__tests__/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+import {render as HTMLRender, cleanup, act, fireEvent} from '@testing-library/react'
+import {axe, toHaveNoViolations} from 'jest-axe'
+import React, {useCallback, useRef, useState} from 'react'
+
+import {ActionMenu} from '../ActionMenu'
+import BaseStyles from '../BaseStyles'
+import Box from '../Box'
+import Button from '../Button/Button'
+import {ConfirmationDialog, useConfirm} from '../Dialog/ConfirmationDialog'
+import {COMMON, LAYOUT} from '../constants'
```
colebemis, line 11:
```suggestion
```

PR #1472, opened 2021-09-28
feat: Focus the primary action in 'ConfirmationDialog', without hacks

smockle:
Alternative to #1471. This PR makes [`focusTrap`’s `initialFocus`](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/behaviors/focusTrap.ts#L82-L84) handle action focus and subsequently removes the (now redundant) [hack in `Dialog.Buttons`](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/Dialog/Dialog.tsx#L345) which previously handled focus.

⚠️ **Warning:** That hack may still be necessary to support cases I did not consider—additional review is appreciated. If we determine the hack is still necessary, please close this PR in favor of #1471.

Fixes https://github.com/github/primer/issues/313

ℹ️ You might wonder–“Why wasn’t moving [the `autoFocus` attribute](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/Dialog/ConfirmationDialog.tsx#L121) from `cancelButton` to `confirmButton` in `ConfirmationDialog.tsx` sufficient?” Great question! I wondered the same thing at first. It turns out that [this special case in `focusTrap`](https://github.com/primer/react/blob/ad0d5c6fee8ba31035f5ea1ebf75ba4c9baa52bd/src/behaviors/focusTrap.ts#L91) causes the primary-action button (`confirmButton`) to receive `tabIndex: -1`, so the secondary-action button (`cancelButton`) still receives focus, even if `autoFocus: true` is moved to `confirmButton`.

### Screenshots

Screen recording demonstrating the primary action is focused when a 'ConfirmationDialog' opens:
Screen recording demonstrating the primary action is focused when 'ConfirmationDialog' opens

We generally follow platform conventions, and primary-action-focus is the default in macOS confirmation dialogs (screen recording of Pages below):
Confirmation dialog in Pages, with the primary action focused by default

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Dialog/Dialog.tsx
+++ b/src/Dialog/Dialog.tsx
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 import Button, {ButtonPrimary, ButtonDanger, ButtonProps} from '../Button'
 import Box from '../Box'
 import {get, SystemCommonProps, SystemPositionProps, COMMON, POSITION} from '../constants'
-import {useOnEscapePress} from '../hooks'
+import {useOnEscapePress, useProvidedRefOrCreate} from '../hooks'
 import {useFocusTrap} from '../hooks/useFocusTrap'
 import sx, {SxProp} from '../sx'
 import StyledOcticon from '../StyledOcticon'
@@ -36,6 +36,12 @@ export type DialogButtonProps = ButtonProps & {
    * focus this button automatically when the dialog appears.
    */
   autoFocus?: boolean
+
+  /**
+   * A reference to the rendered Button’s DOM node, used together with
+   * `autoFocus` for `focusTrap`’s `initialFocus`.
+   */
+  ref?: React.RefObject<HTMLButtonElement>
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/Dialog/Dialog.tsx
```diff
@@ -338,29 +345,13 @@ const buttonTypes = {
...
-  let autoFocusCount = 0
-  const [hasRendered, setHasRendered] = useState(0)
-  useEffect(() => {
-    // hack to work around dialogs originating from other focus traps.
-    if (hasRendered === 1) {
-      autoFocusRef.current?.focus()
-    } else {
-      setHasRendered(hasRendered + 1)
-    }
-  }, [hasRendered])
```
colebemis, line 351: I'm a little hesitant to remove this code because I don't fully understand why it was added. Who would have the most context about why this was necessary?

smockle: It was added in https://github.com/primer/react/pull/1235, which @dgreif reviewed. I just noticed that PR includes [a Storybook story which demos the issue](https://github.com/primer/react/pull/1235/files#diff-1a7fe7df557c29e6117f7bbeade270a7ea033739ebe88ea1519399d94722eac7); I’m checking now whether this PR causing that story to regress.

smockle: This PR _does_ cause the `ShorthandHookFromActionMenu` story to regress—in #1471, the `OK` button is focused, but in #1472, the `Cancel` button is focused (even though the `OK` button has `autoFocus`). I’ll close this issue in favor of #1471. 👍

smockle:
Re-opening to test https://github.com/primer/react/pull/1472/commits/f9c93a7b713e0023545539efe47e05a9854b261e. If that test works, the check should fail. 🤞

PR #1474, opened 2021-09-28
Add color blind themes

colebemis:
Adds color blind color schemes published in [v4.8.0](https://github.com/primer/primitives/releases/tag/v4.8.0) of Primer Primitives.

Try it out in storybook: https://primer-components-4g2oegw1w-primer.vercel.app/react/storybook?path=/story/composite-components-button--primary-button&globals=dayScheme:light_protanopia

CleanShot 2021-09-28 at 13 43 10

Part of https://github.com/github/design-infrastructure/issues/1911

### DIFF
--- a/src/theme-preval.js
+++ b/src/theme-preval.js
@@ -48,58 +48,20 @@ const fontSizes = ['12px', '14px', '16px', '20px', '24px', '32px', '40px', '48px
 
 const space = ['0', '4px', '8px', '16px', '24px', '32px', '40px', '48px', '64px', '80px', '96px', '112px', '128px']
 
-const light = partitionColors(primitives.colors.light)
-const dark = partitionColors(primitives.colors.dark)
-const darkDimmed = partitionColors(primitives.colors['dark_dimmed'])
-const darkHighContrast = partitionColors(primitives.colors['dark_high_contrast'])
-
-// This file must be in vanilla JS to work with preval
-// but our temporary filter utils make it impossible for
-// our TypeScript to properly infer const object structure
-// so we need to use JSDoc comments.
-
 /**
- * @type Partial<typeof primitives.colors.light>
+ * @type Record<keyof typeof primitives.colors, Record<'colors' | 'shadows', Partial<typeof primitives.colors.light>>
  */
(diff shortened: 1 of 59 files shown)


src/theme-preval.js
```diff
@@ -48,58 +48,20 @@ const fontSizes = ['12px', '14px', '16px', '20px', '24px', '32px', '40px', '48px
...
-const darkHighContrastShadows = omitScale(darkHighContrast.shadows)
+const colorSchemes = Object.entries(primitives.colors)
+  .map(([name, variables]) => {
+    const {colors, shadows} = partitionColors(variables)
+    return [name, {colors: omitScale(colors), shadows: omitScale(shadows)}]
+  })
+  .reduce((acc, [name, variables]) => {
+    acc[name] = variables
+    return acc
+  }, {})
```
colebemis, line 62: I refactored this to get all color schemes from Primer Primitives instead of hard-coding references to them. This means that when future color schemes are published in Primer Primitives, we only need to bump the Primer Primitives version number in Primer React to make the color schemes available to Primer React consumers.

src/theme-preval.js
```diff
@@ -48,58 +48,20 @@ const fontSizes = ['12px', '14px', '16px', '20px', '24px', '32px', '40px', '48px
...
- */
-const darkHighContrastColors = omitScale(darkHighContrast.colors)
-
 /**
- * @type Partial<typeof primitives.colors.dark_high_contrast>
+ * @type Record<keyof typeof primitives.colors, Record<'colors' | 'shadows', Partial<typeof primitives.colors.light>>
  */
-const darkHighContrastShadows = omitScale(darkHighContrast.shadows)
+const colorSchemes = Object.entries(primitives.colors)
+  .map(([name, variables]) => {
```
jfuchs, line 55: Non-blocking: you could collapse this map into the reduce below, if you think it'd be more readable?

jfuchs (approved):

src/theme-preval.js
colebemis: Yup. Good call. I'll do that

PR #1487, opened 2021-09-30
jfuchs/props adr

jfuchs:
Needs wordsmithing, but I want to make sure the decisions are not wildly off before I keep going


contributor-docs/adrs/adr-003-prop-norms.md
```diff
@@ -0,0 +1,53 @@
...
+### 🟢 `ref`
+
+All components that ultimately render to the DOM should accept a `ref` prop. That `ref` prop should be passed to the root HTMLElement rendered by the component.
+
+### 🟡 `as`: Limited
+
+Only components with a clear need for polymorphism should accept an `as` prop. Reasonable cases include:
+
+- Components that need functionality from the `as` prop, like a button
+- Components whose accessibility are improved by using semantically appropriate HTML elements
```
colebemis, line 30: The reasoning behind when a component should accept the `as` prop is a little unclear here.

I think we should also mention restricting the allowed values for the `as` prop. Some questions to answer:
* When do should we allow anything to be passed to the `as` prop?
* When should the possible values be limited?
* How do we decide on a limited set of values?

contributor-docs/adrs/adr-003-prop-norms.md
```diff
@@ -0,0 +1,53 @@
...
+- Consistency is important
+- There’s a gap between where we want to be and where we
+- This document aims to codify our best practices for component APIs and unify disparate conversations
+  - In the course of some refactorings, we may have intermediate states where there will added be value in a clear documentation of which parts of our API reflect the best practices.
+
+## Decision
+
+### 🟢 `sx`
+
+All components that ultimately render to the DOM should accept an `sx` prop. The `sx` prop (of type `SystemStyleObject`) should set styles for the root HTML element rendered by the component. It may make sense to have an additional prop to style a descendant element (e.g., `innerSx`).
```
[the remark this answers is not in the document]
colebemis: Also, I think "renders something to the DOM" is a great way to decide if something should accept an `sx` prop. Nice!

contributor-docs/adrs/adr-003-prop-norms.md
```diff
@@ -0,0 +1,53 @@
...
+- Components that need functionality from the `as` prop, like a button
+- Components whose accessibility are improved by using semantically appropriate HTML elements
+
+### 🟡 DOM props: Limited
+
+All components that accept an `as` prop should accept the props for the element specified by the as prop. _Additionally_, some other elements (e.g., `<TextInput>`) should accept the props for their root HTML element when those props are fundamental to the component’s function.
+
+### 🔴 System props
+
+Components should not accept styled system props (except our utility components: `Box` and `Type`)
```
colebemis, line 38: I wonder if we can explain why utility components get system props and the rest of the components don't. 

This is worded badly but something like... The primary function of utility components is to provide a convenient API for writing custom styles using design system values so it doesn't make sense to bury that API in the `sx` prop. Non-utility components' purpose is to implement specific design patterns. Custom styling is not their primary function so we don't want them to have a first-class API (i.e. system props) for overriding styles.

colebemis (commented):
This is a great start! Thanks for taking this on, @jfuchs.

This document will provide a lot of clarity as we incrementally work toward more consistent APIs for Primer React components ❤️

contributor-docs/adrs/adr-003-prop-norms.md
siddharthkp: How do y'all feel about using sx prop in the case of utility components as well for the sake of the "one true way" to use tokens from the system?

colebemis: I'm open to that!

contributor-docs/adrs/adr-003-prop-norms.md
jfuchs: Oh interesting. I had assumed that _if_ we ever have an `as` prop, we would allow any `ElementType`.

contributor-docs/adrs/adr-003-prop-norms.md
jfuchs: Ooh, that's interesting. A couple questions that might inform whether `sx` is sufficient for our utility components:
- Would the slight loss of convenience lead to users taking shortcuts that they wouldn't otherwise take? Would a user decide to apply styles directly or reach down into styled-components to avoid the clunkiness of going through `sx`?
- Would a potential user decide not to use PRC because of the loss of convenience? Would they decide to use PRC because of the more consistent API?

contributor-docs/adrs/adr-003-prop-norms.md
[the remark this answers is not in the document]
jfuchs: > If consumers really need to apply styles to a descendant element, they can use nested selectors in the sx prop:
 
It seems like nested selectors would require more knowledge about the implementation of a component than `innerSx`?

Also the `innerSx` example was probably a bad one. Here's something more realistic I think: What if `Dialog` took an `sx` prop that applied to the dialog, and also took a `backdropSx`. If the name is more semantically meaningful ("backdrop" over "inner") does that change anything?

contributor-docs/adrs/adr-003-prop-norms.md
colebemis: Primer ViewComponents restricts it for some components: https://github.com/github/primer/discussions/130#discussioncomment-676056

contributor-docs/adrs/adr-003-prop-norms.md
```diff
@@ -0,0 +1,70 @@
...
+Only components with a clear need for polymorphism should accept an `as` prop. Reasonable cases include:
+
+- Components that need functionality from the `as` prop, like a button
+- Components whose accessibility are improved by using semantically appropriate HTML elements
+
+When passing an `as` prop to a component, it should be done in a way that is consistent with the component’s intended use.
+
+### 🟡 DOM props: Limited
+
+All components that accept an `as` prop should accept the props for the element specified by the as prop. _Additionally_, some other elements (e.g., `<TextInput>`) should accept the props for their root HTML element when those props are fundamental to the component’s function.
```
colebemis, line 42: How do we feel about components accepting some standard DOM props like `id`, `className`, `data-*`? Are these props that all components that render to the DOM should accept? I know having `id` and `data-*` can be helpful for testing purposes.

contributor-docs/adrs/adr-003-prop-norms.md
```diff
@@ -0,0 +1,70 @@
...
+
+### 🔴 `theme`
+
+Components should not accept a `theme` prop (with the exception of `ThemeProvider`).
+
+_Reasoning:_ The `theme` prop doesn't enable anything that can't be done with `<ThemeProvider>`, and promotes the anti-pattern of per-component theme overrides.
+
+### ❓ `children`
+
+TK What can we say about children? What conventions do we expect when a components accepts children, but of a certain type?
```
colebemis, line 60: This seems connected to the discussion around [content as data vs children](https://github.com/primer/react/blob/main/contributor-docs/component-contents-api-patterns.md) and @siddharthkp's [ActionList API proposal](https://github.com/github/primer/issues/307). I think this topic is big enough to warrant its own ADR. Let's leave it out of this ADR.

contributor-docs/adrs/adr-003-prop-norms.md
```diff
@@ -0,0 +1,70 @@
...
+
+All components that ultimately render to the DOM should accept an `sx` prop.
+
+The `sx` prop (of type `SystemStyleObject`) should generally set styles for the root HTML element rendered by the component. An exception would be components like `Dialog`, whose outermost HTML element is a backdrop. In that case, it would be appropriate for `sx` styles to apply to child of the backdrop that is more likely to need styling overrides.
+
+It may make sense to have an additional prop to style a descendant element (e.g., `Dialog` could accept a `backdropSx` prop).
+
+### 🟢 `ref`
+
+All components that ultimately render to the DOM should accept a `ref` prop. That `ref` prop should be passed to the root HTMLElement rendered by the component.
```
colebemis, line 29: Our `TextInput` component provides an interesting edge case for this rule because the `<input>` element is nested inside a `<span>`: https://github.com/primer/react/blob/main/src/TextInput.tsx#L156-L171

Nesting the `input` inside a `span` allows us to render an icon inside the input:

CleanShot 2021-10-01 at 17 09 25


However, this rule would lead me to pass the `ref` to the wrapping `<span>` but as a consumer, I'd expect the `ref` to be passed to the `<input>` element.

I don't necessarily think we need to update this rule though. Maybe the "correct" solution is adopt a more composable API for TextInput that would avoid this problem entirely (similar to Chakra UI):

CleanShot 2021-10-01 at 17 11 44

https://chakra-ui.com/docs/form/input#add-elements-inside-input

contributor-docs/adrs/adr-003-prop-norms.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Decision
+
+### 🟢 `sx`
+
+All components that ultimately render to the DOM should accept an `sx` prop.
+
+The `sx` prop (of type `SystemStyleObject`) should generally set styles for the root HTML element rendered by the component. An exception would be components like `Dialog`, whose outermost HTML element is a backdrop. In that case, it would be appropriate for `sx` styles to apply to child of the backdrop that is more likely to need styling overrides.
+
+It may make sense to have an additional prop to style a descendant element (e.g., `Dialog` could accept a `backdropSx` prop).
```
colebemis, line 25: `backdropSx` feels wrong to me for some reason 🤔  I'm unsure if we'd need to allow style overrides for the backdrop.

colebemis: For `Dialog`, should `ref` be passed to the dialog or the backdrop?

contributor-docs/adrs/adr-003-prop-norms.md
jfuchs: [Strong opinion, weakly held](https://www.saffo.com/02008/07/26/strong-opinions-weakly-held/): It is fine to add support for props that are one-by-one relayed to the DOM (i.e., not en masse via a spread) as they are needed.

In your example, if there was a need for a test attribute, I'd rather accept a prop like `data-test-id` than `id` so the naming reinforces the semantics of the field.

contributor-docs/adrs/adr-003-prop-norms.md
jfuchs: Makes sense! I might just leave a call-out that it is intentionally withheld.

contributor-docs/adrs/adr-003-prop-norms.md
jfuchs: Oooh yeah, I think this is a good argument for @siddharthkp's work on `children`-based APIs.

For this ADR, I might just relax the language to say:

> That `ref` prop should most often be passed to the root HTMLElement rendered by the component, although sometimes it may make more sense to pass it to a descendent node.

contributor-docs/adrs/adr-003-prop-norms.md
jfuchs: > backdropSx feels wrong to me for some reason 🤔 I'm unsure if we'd need to allow style overrides for the backdrop.

Fair. Can you think of any place where a second `sx` prop would feel right?

> For Dialog, should ref be passed to the dialog or the backdrop?

Good point. I'd expect it to go to the dialog. See my above note on weakening the language re: who gets the `ref`.

contributor-docs/adrs/adr-003-prop-norms.md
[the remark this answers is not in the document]
siddharthkp: okay, this is weird, but hear me out, how about:

```js
<Dropdown sx={{ [data-component="Overlay"]: { width: 'medium' } }} />
```

we give each of our components a data attribute like `data-primer` or `data-component`, this way you can safely target it without making assumptions about the dom structure or element tagname.

bonus: if we can add some syntactical sugar around sx, we could dot this:

```js
<Dropdown sx={{ Overlay: { width: 'medium' } }} />
```

Capital letter means component.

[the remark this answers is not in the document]
jfuchs: That's interesting, and I'd like us to explore it more. But it feels too experimental to tack on to this ADR. How about I remove any language about additional `*sx` props for now, and maybe later as we get a better sense of what happens with children-as-content we'll also get a better sense of if this is a real problem or hypothetical?

PR #1488, opened 2021-09-30
Token components

mperrotti:
**I will fill this out Friday 01 Oct**

---

Describe your changes here.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/Token.mdx
+++ b/docs/content/Token.mdx
@@ -1,0 +1,51 @@
+# Tokens
+
+A Token represents a piece of data. They are typically used to show a collection of related attributes.
+
+The default `Token` can be used for most cases, but specialized token components are provided for specific use cases (see below for more info).
+
+All types of tokens may have the same interactive qualities as links or buttons. They also have a sense of being "selected" and ready for user interaction such as being removed from a collection.
+
+## Token
+
+The default `Token` component
+
+### Basic example
+<!-- insert example of plain token that just has text -->
+```tsx live
+```
+
+### With a `leadingVisual`
+<!-- insert example of plain token that has an icon as the leadingVisual -->
+```tsx live
+```
+
+### Interactive token examples
+<!--
+insert multiple examples:
+- link
+- button
+- with internal remove button
+- selectable
+-->
+```tsx live
+```
+
+## Component props
+<!-- insert prop table here -->
+
+## TokenLabel
+
+Tokens that represent Issue Labels should use the `TokenLabel` component.
+
+<!-- insert row of issue labels -->
+```tsx live
+```
+
+## TokenProfile
+
+Tokens that represent GitHub users should use the `TokenProfile` component.
+
+<!-- insert row of profile tokens -->
+```tsx live
+```
--- a/src/Token/Token.tsx
+++ b/src/Token/Token.tsx
@@ -1,0 +1,98 @@
+import React, { forwardRef, KeyboardEventHandler, MouseEventHandler } from 'react'
+import styled from 'styled-components'
+import { get } from '../constants'
+import TokenBase, { isTokenInteractive, TokenBaseProps } from './TokenBase'
+import RemoveTokenButton from './_RemoveTokenButton'
+
+export interface TokenProps extends TokenBaseProps {
+    /**
+     * A function that renders a component before the token text
+     */
+    leadingVisual?: React.FunctionComponent<any>
+    /**
+     * Whether the remove button should be rendered in the token
+     */
+    hideRemoveButton?: boolean
+}
+
+const tokenBorderWidthPx = 1;
+
+const DefaultTokenStyled = styled(TokenBase)<TokenProps>`
+    background-color: ${get('colors.neutral.subtle')};
+    border-color: ${props => props.isSelected ? get('colors.fg.default') : get('colors.border.subtle')};
+    border-style: solid;
+    border-width: 1px;
+    color: ${props => props.isSelected ? get('colors.fg.default') : get('colors.fg.muted')};
+    max-width: 100%;
+    padding-right: ${props => !props.hideRemoveButton ? 0 : undefined};
+
+    &:hover {
+        background-color: ${props => isTokenInteractive(props) ? get('colors.neutral.muted') : undefined};
+        box-shadow: ${props => isTokenInteractive(props) ? get('colors.shadow.medium') : undefined};
+        color: ${props => isTokenInteractive(props) ? get('colors.fg.default') : undefined};
+    }
+`;
+
+const TokenTextContainer = styled('span')`
+    flex-grow: 1;
+    min-width: 0;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+`;
+
+const LeadingVisualContainer = styled('span')`
+    flex-shrink: 0;
+    line-height: 0;
+`;
+
+const Token = forwardRef<HTMLAnchorElement | HTMLButtonElement | HTMLSpanElement, TokenProps>((props, forwardedRef) => {
+    const {
+        as,
+        handleRemove,
+        id,
+        leadingVisual: LeadingVisual,
+        ref,
+        text,
+        variant,
+        hideRemoveButton,
+        ...rest
+    } = props;
+    const hasMultipleActionTargets = isTokenInteractive(props) && Boolean(handleRemove) && !hideRemoveButton
+    const handleRemoveClick: MouseEventHandler = (e) => {
+        e.stopPropagation()
+        handleRemove && handleRemove()
+    }
+
+    return (
+        <DefaultTokenStyled
+            as={as}
+            handleRemove={handleRemove}
+            hideRemoveButton={hideRemoveButton || !handleRemove}
+            id={id?.toString()}
+            text={text}
+            ref={forwardedRef}
+            variant={variant}
+            {...rest}
+        >
+            {LeadingVisual ? (
+                <LeadingVisualContainer>
+                    <LeadingVisual />
+                </LeadingVisualContainer>
+            ) : null}
+            <TokenTextContainer>{text}</TokenTextContainer>
+            {!hideRemoveButton && handleRemove ? (
+                <RemoveTokenButton
+                    borderOffset={tokenBorderWidthPx}
+                    parentTokenTag={as || 'span'}
+                    onClick={handleRemoveClick}
+                    variant={variant}
+                    parentTokenIsInteractive={isTokenInteractive(props)}
+                    aria-hidden={hasMultipleActionTargets ? "true" : "false"}
+                />
+            ) : null}
+        </DefaultTokenStyled>
+    )
+});
+
+export default Token;
--- a/src/Token/TokenBase.tsx
+++ b/src/Token/TokenBase.tsx
@@ -1,0 +1,92 @@
+import { KeyboardEvent } from 'react';
+import styled, { css } from 'styled-components'
+import { variant } from 'styled-system'
+import { get } from '../constants'
+
+export type TokenSizeKeys = 'sm' | 'md' | 'lg' | 'xl'
+
+export const tokenSizes: Record<TokenSizeKeys, number> = {
+    sm: 16,
+    md: 20,
+    lg: 24,
+    xl: 32,
+}
+
+export const defaultTokenSize = 'md'
+
+export interface TokenBaseProps extends Omit<React.HTMLProps<HTMLSpanElement | HTMLButtonElement | HTMLAnchorElement>, 'size' | 'id'> {
+    as?: 'button' | 'a' | 'span'
+    handleRemove?: () => void
+    isSelected?: boolean
+    tabIndex?: number
+    text: string
+    id?: number | string
+    variant?: TokenSizeKeys
+}
+
+export const isTokenInteractive = ({
+  as = 'span',
+  onClick,
+  onFocus,
+  tabIndex = -1
+}: TokenBaseProps) => Boolean(onFocus || onClick || tabIndex > -1 || ['a', 'button'].includes(as))
+
+const variants = variant<{fontSize: number, height: string, gap: number, paddingLeft: any, paddingRight: number}, TokenSizeKeys>({
+    variants: {
+      sm: {
+        fontSize: 0,
+        gap: 1,
+        height: `${tokenSizes.sm}px`,
+        paddingLeft: 1,
+        paddingRight: 1,
+      },
+      md: {
+        fontSize: 0,
+        gap: 1,
+        height: `${tokenSizes.md}px`,
+        paddingLeft: 2,
+        paddingRight: 2,
+      },
+      lg: {
+        fontSize: 0,
+        gap: 2,
+        height: `${tokenSizes.lg}px`,
+        paddingLeft: 2,
+        paddingRight: 2,
+      },
+      xl: {
+        fontSize: 1,
+        gap: 2,
+        height: `${tokenSizes.xl}px`,
+        paddingLeft: 3,
+        paddingRight: 3,
+      }
+    }
+  })
+
+const TokenBase = styled.span.attrs<TokenBaseProps>(({text, handleRemove, onKeyUp}) => ({
+  onKeyUp: (e: KeyboardEvent<HTMLSpanElement | HTMLButtonElement | HTMLAnchorElement>) => {
+    onKeyUp && onKeyUp(e)
+
+    if ((e.key === 'Backspace' || e.key === 'Delete') && handleRemove) {
+      handleRemove()
+    }
+  },
+  'aria-label': handleRemove ? `${text}, press backspace or delete to remove` : undefined
+}))<TokenBaseProps>`
+  align-items: center;
+  border-radius: 999px;
+  cursor: ${props => isTokenInteractive(props) ? 'pointer' : 'auto'};
+  display: inline-flex;
+  font-weight: ${get('fontWeights.bold')};
+  text-decoration: none;
+  white-space: nowrap;
+  ${variants}
+`
+
+TokenBase.defaultProps = {
+  as: 'span',
+  variant: defaultTokenSize,
+}
+
+export default TokenBase
--- a/src/Token/TokenLabel.tsx
+++ b/src/Token/TokenLabel.tsx
@@ -1,0 +1,191 @@
+import React, { forwardRef, MouseEventHandler } from 'react'
+import styled, { css } from 'styled-components'
+import TokenBase, { isTokenInteractive, TokenBaseProps } from './TokenBase'
+import RemoveTokenButton from './_RemoveTokenButton'
+import tinycolor from 'tinycolor2'
+import { useTheme } from '../ThemeProvider'
+
+interface ColorModeConfig {
+    bgOpacity: number
+    borderThreshold: number
+    borderOpacity: number
+    lightnessThreshold: number
+}
+
+const colorModeConfigs: Record<string, ColorModeConfig> = {
+    dark: {
+        bgOpacity: 0.18,
+        borderThreshold: 0,
+        borderOpacity: 0.3,
+        lightnessThreshold: 0.6,
+    },
+    light: {
+        bgOpacity: 1,
+        borderThreshold: 0.96,
+        borderOpacity: 1,
+        lightnessThreshold: 0.453,
+    }
+}
+
+export interface TokenLabelProps extends TokenBaseProps {
+    /**
+     * The color that corresponds to the label
+     */
+    fillColor?: string
+    /**
+     * Whether the remove button should be rendered in the token
+     */
+    hideRemoveButton?: boolean
+}
+
+interface LabelStyleProps {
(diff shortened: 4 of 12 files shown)


src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    if ((e.key === 'Backspace' || e.key === 'Delete') && handleRemove) {
+      handleRemove()
+    }
+  },
+  'aria-label': handleRemove ? `${text}, press backspace or delete to remove` : undefined
+}))<TokenBaseProps>`
+  align-items: center;
+  border-radius: 999px;
+  cursor: ${props => (isTokenInteractive(props) ? 'pointer' : 'auto')};
```
colebemis, line 79: `isTokenInteractive` feels a little brittle to me. Could we do this with pure CSS? Maybe something like:

```css
&:is(a, button) {
  cursor: pointer;
}
```

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+      gap: 2,
+      height: `${tokenSizes.xl}px`,
+      paddingLeft: 3,
+      paddingRight: 3
+    }
+  }
+})
+
+const TokenBase = styled.span.attrs<TokenBaseProps>(({text, handleRemove, onKeyUp}) => ({
+  onKeyUp: (e: KeyboardEvent<HTMLSpanElement | HTMLButtonElement | HTMLAnchorElement>) => {
```
colebemis, line 68: Is there a reason we're attaching this event handler to onKeyUp instead of onKeyDown?

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+  Boolean(onFocus || onClick || tabIndex > -1 || ['a', 'button'].includes(as))
+
+const variants = variant<
+  {fontSize: number; height: string; gap: number; paddingLeft: any; paddingRight: number},
+  TokenSizeKeys
+>({
+  variants: {
+    sm: {
+      fontSize: 0,
+      gap: 1,
```
colebemis, line 38: Looks like the flexbox gap property still doesn't have support in Safari <14.1. Do we need to support Safari <14.1? https://caniuse.com/flexbox-gap

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export interface TokenBaseProps
+  extends Omit<React.HTMLProps<HTMLSpanElement | HTMLButtonElement | HTMLAnchorElement>, 'size' | 'id'> {
+  as?: 'button' | 'a' | 'span'
+  handleRemove?: () => void
+  isSelected?: boolean
+  tabIndex?: number
+  text: string
+  id?: number | string
+  variant?: TokenSizeKeys
```
colebemis, line 25: IMO we should only use the prop name "variant" when no other name would be more clear. In this case, `size` feels more clear to me:

```suggestion
  size?: TokenSize
```

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+const variants = variant<
+  {fontSize: number; height: string; gap: number; paddingLeft: any; paddingRight: number},
+  TokenSizeKeys
+>({
+  variants: {
+    sm: {
+      fontSize: 0,
+      gap: 1,
+      height: `${tokenSizes.sm}px`,
```
colebemis, line 39: Is there a reason we don't store the token sizes in pixel values so we don't have to do this string manipulation?

```js
const tokenSizes = {
  small: '16px'
  ...
}
```

src/Token/TokenLabel.tsx
```diff
@@ -0,0 +1,178 @@
...
+      id={id?.toString()}
+      isSelected={isSelected}
+      ref={forwardedRef}
+      text={text}
+      variant={variant}
+      {...rest}
+    >
+      <TokenTextContainer>{text}</TokenTextContainer>
+      {!hideRemoveButton && handleRemove ? (
+        <RemoveTokenButton
```
colebemis, line 161: Is nesting a button inside another button or link an accessibility issue?

src/Token/_RemoveTokenButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+import {defaultTokenSize} from './TokenBase'
+import {TokenButtonProps, tokenButtonStyles, variants, getTokenButtonIconSize} from './_tokenButtonUtils'
+
+const RemoveTokenButton = styled.span.attrs<TokenButtonProps>(
+  ({borderOffset, parentTokenTag, variant, parentTokenIsInteractive, ...rest}) => {
+    delete rest.children
+
+    return {
+      borderOffset,
+      as: parentTokenIsInteractive ? 'span' : 'button',
```
colebemis, line 12: Curious when we would want this button to be rendered as a span?

src/Token/TokenBase.tsx
colebemis: Ah, I see that `isTokenInteractive` is used in many other places where we can't use CSS. Seems fine for now 👍 I'm just not sure if it would work if we allowed for ReactRouter-like Link components.

src/Token/_RemoveTokenButton.tsx
colebemis: Ah, this is to avoid rendering a button inside an interactive element.

colebemis: In the storybook, clicking the remove icon on the link token takes me to the link address. Is that what we want?
CleanShot 2021-10-04 at 12 00 03

docs/content/Token.mdx
```diff
@@ -0,0 +1,51 @@
...
+- with internal remove button
+- selectable
+-->
+```tsx live
+```
+
+## Component props
+<!-- insert prop table here -->
+
+## TokenLabel
```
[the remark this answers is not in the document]
mperrotti: I also prefer `LabelToken`, but I was following the pattern we use for `Button` (e.g.: `ButtonSuccess`, `ButtonPrimary`, etc).

More than happy to change to `LabelToken` if I misinterpreted the `Button` component names as some kind of naming convention. Let me know what you think.

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+  md: 20,
+  lg: 24,
+  xl: 32
+}
+
+export const defaultTokenSize = 'md'
+
+export interface TokenBaseProps
+  extends Omit<React.HTMLProps<HTMLSpanElement | HTMLButtonElement | HTMLAnchorElement>, 'size' | 'id'> {
+  as?: 'button' | 'a' | 'span'
```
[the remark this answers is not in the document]
mperrotti: Yes, probably

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+  handleRemove?: () => void
```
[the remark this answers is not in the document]
mperrotti: Yea, totally agreed. Not committing the suggestions because I'll also need to update Storybook stories

src/stories/Token.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+      <Box
+        display="flex"
+        sx={{
+          alignItems: 'start',
+          gap: get('space.2')
+        }}
+      >
+        <Token as="a" href="http://google.com/" text="Link" />
+        <Token as="button" onClick={() => console.log('clicked')} text="Button" />
+        <Token as="span" tabIndex={0} onFocus={() => console.log('focused')} text="Focusable Span" />
```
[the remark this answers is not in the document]
mperrotti: We discussed this in a Zoom call, but for those following along:
There could be times where we update something like the `isSelected` prop when a token gets focus. For example, the TextInput component that renders tokens does this

src/Token/TokenBase.tsx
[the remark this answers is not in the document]
mperrotti: Just realized something...

We change the way we render the "x" button if the token is a link or a button. For example, we don't let the "x" button get focused with the keyboard, and we don't give the "x" button an `aria-label`.

src/Token/TokenBase.tsx
mperrotti: When/if we add support for ReactRouter-like Link components, we should make this suggested change

src/Token/_tokenButtonUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+      width: `${tokenSizes.lg}px`
+    },
+    xl: {
+      height: `${tokenSizes.xl}px`,
+      width: `${tokenSizes.xl}px`
+    }
+  }
+})
+
+export const tokenButtonStyles = css`
```
colebemis, line 34: Is there a reason you pulled this into its own file?

src/Token/TokenBase.tsx
mperrotti: I don't think so. I think we actually want `onKeyDown`

src/Token/TokenBase.tsx
mperrotti: I'm not sure if we need to support Safari <14.1. Do we have some guidelines on Primer's browser support?

src/Token/TokenBase.tsx
mperrotti: I think I set this to `variant` because of something to do with `styled-components`. I also prefer `size`, so I'll try changing it.

src/Token/TokenBase.tsx
mperrotti: No reason. I agree with you - we should store the values with a unit.

src/Token/TokenLabel.tsx
mperrotti: You addressed this in your next comment: we specifically avoid rendering a button inside of a button or link.

src/Token/_RemoveTokenButton.tsx
mperrotti: > In the storybook, clicking the remove icon on the link token takes me to the link address. Is that what we want?

No, this is a bug. Will fix

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,92 @@
...
+})
+
+const TokenBase = styled.span.attrs<TokenBaseProps>(({text, handleRemove, onKeyUp}) => ({
+  onKeyUp: (e: KeyboardEvent<HTMLSpanElement | HTMLButtonElement | HTMLAnchorElement>) => {
+    onKeyUp && onKeyUp(e)
+
+    if ((e.key === 'Backspace' || e.key === 'Delete') && handleRemove) {
+      handleRemove()
+    }
+  },
```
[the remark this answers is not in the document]
mperrotti: Thanks @colebemis - I'll update my other PRs and keep this in mind for future work.

src/Token/_tokenButtonUtils.ts
mperrotti: I had an "add" button at some point, and I just forgot to move this back to the `RemoveTokenButton` component file.

docs/content/Token.mdx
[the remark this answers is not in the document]
colebemis: IMO the button API should be changed. We should just have one button component with multiple variants:

```
<ButtonPrimary> -> <Button variant="primary">
```

In other words, I don't think we need to base the token API on the button API.

src/Token/Token.tsx
```diff
@@ -0,0 +1,88 @@
...
+import styled from 'styled-components'
+import {get} from '../constants'
+import TokenBase, {isTokenInteractive, TokenBaseProps} from './TokenBase'
+import RemoveTokenButton from './_RemoveTokenButton'
+
+export interface TokenProps extends TokenBaseProps {
+  /**
+   * A function that renders a component before the token text
+   */
+  leadingVisual?: React.FunctionComponent<any>
```
colebemis, line 11: Does this need to a be a FunctionComponent or could it be `React.ComponentType`?

mperrotti: I think it could be `React.ComponentType`

src/Token/TokenLabel.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const TokenLabel = forwardRef<HTMLElement, TokenLabelProps>((props, forwardedRef) => {
+  const {as, fillColor, handleRemove, id, isSelected, ref, text, variant, hideRemoveButton, ...rest} = props
+  const {colorScheme} = useTheme()
+  const {bgOpacity, borderOpacity, borderThreshold, lightnessThreshold} = colorModeConfigs[colorScheme || 'light']
+  let bgColor = fillColor
+  let borderColor = fillColor
+  let textColor = '#FFF'
+  const perceivedLightness = tinycolor(fillColor).getLuminance()
+  const isFillColorLight = perceivedLightness >= lightnessThreshold
```
colebemis, line 95: Have you seen the Memex/dotcom implementation of issue labels? Maybe we can copy some of that code for LabelToken.

https://github.com/github/memex/blob/main/src/client/components/issue-label.tsx

mperrotti: I saw the dotcom implementation, but not the Memex implementation. I'm definitely going to copy some of that code.

src/Token/_tokenButtonUtils.ts
```diff
@@ -0,0 +1,61 @@
+import {css} from 'styled-components'
+import {variant} from 'styled-system'
+import {get} from '../constants'
+import {tokenSizes, TokenSizeKeys, defaultTokenSize} from './TokenBase'
+
+export interface TokenButtonProps {
+  borderOffset?: number
+  parentTokenTag: 'span' | 'button' | 'a'
+  variant?: TokenSizeKeys
```
colebemis, line 9:
```suggestion
  size?: TokenSizeKeys
```

src/Token/_tokenButtonUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+  parentTokenIsInteractive?: boolean
```
colebemis, line 10: Using `is` as a prefix for booleans feels more consistent here:

```suggestion
  isParentInteractive?: boolean
```

I'm also unsure if having "token" in the variable name is redundant or not.

src/Token/_RemoveTokenButton.tsx
```diff
@@ -0,0 +1,28 @@
+import {XIcon} from '@primer/octicons-react'
+import styled from 'styled-components'
+import {defaultTokenSize} from './TokenBase'
+import {TokenButtonProps, tokenButtonStyles, variants, getTokenButtonIconSize} from './_tokenButtonUtils'
+
+const RemoveTokenButton = styled.span.attrs<TokenButtonProps>(
+  ({borderOffset, parentTokenTag, variant, parentTokenIsInteractive, ...rest}) => {
+    delete rest.children
```
colebemis, line 8: Doing all of this inside the `styled` function feels a little off to me. What if we organized the component more like this:

```
const TokenButton = styled.button`
   ...all the styles here...
`

function RemoveTokenButton({isParentInteractive, ...}) {
  return <TokenButton as={isParentInteractive? : 'span' : 'button'} ...><XIcon /></TokenButton>
}
```

src/Token/TokenBase.tsx
mperrotti: This is possible: https://styled-system.com/variants/#custom-prop-name

src/Token/TokenLabel.tsx
mperrotti: I tried copying that code, but it's leading to bad contrast, I can't figure out how to get the `:after` pseudo-element styles working, and it's going to take a little more work to get the `darken` and `lighten` color functions right for the `isSelected` styles.

Screen Shot 2021-10-04 at 8 29 50 PM

If you still think we should still use the custom property implementation, I'll make it work. Otherwise, I'm just going to replace `tinycolor2` with `color2k` and keep everything in JavaScript.

mperrotti:
@colebemis - I think we should rename `LabelToken` to `IssueLabelToken`. What do you think?

colebemis:
> @colebemis - I think we should rename `LabelToken` to `IssueLabelToken`. What do you think?

Agreed. I was actually going to suggest that 👍

src/stories/Token.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+    <SingleExampleContainer label="isSelected">
+      <Token text="Default Token" isSelected={true} />
+    </SingleExampleContainer>
+  </ExampleCollectionContainer>
+)
+
+export const profileToken = () => (
+  <ExampleCollectionContainer>
+    <SingleExampleContainer label="Resting">
+      <TokenProfile size="lg" avatarSrc="https://avatars.githubusercontent.com/mperrotti" text="Mike Perrotti" />
```
siddharthkp, line 133: Should we use children instead of text here? 

```jsx
<TokenProfile size="lg" avatarSrc="https://avatars.githubusercontent.com/mperrotti">
   Mike Perrotti
</TokenProfile>
```

src/Token/TokenBase.tsx
```diff
@@ -0,0 +1,93 @@
+import {KeyboardEvent} from 'react'
+import styled from 'styled-components'
+import {variant} from 'styled-system'
+import {get} from '../constants'
+
+export type TokenSizeKeys = 'sm' | 'md' | 'lg' | 'xl'
```
siddharthkp, line 6: Should we use full names? small, medium, large, xlarge?

Looking around at other components, that seems to be the pattern: [CircleBadge](https://github.com/primer/react/blob/627cd3e95614ab4347e5acf456dd23588196ec87/src/CircleBadge.tsx#L8-L11), [Label](https://github.com/primer/react/blob/627cd3e95614ab4347e5acf456dd23588196ec87/src/Label.tsx#L18), [Spinner](https://github.com/primer/react/blob/627cd3e95614ab4347e5acf456dd23588196ec87/src/Spinner.tsx#L7-L11)

Part 2: How do you feel about using `variant` instead of `size` here, like we do for [Label](https://primer.style/react/Label#default-example).

src/stories/Token.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+      <LabelToken
+        text="good first issue"
+        fillColor="#656BFE"
+        isSelected={true}
+        onRemove={() => {
+          console.log('remove me')
+        }}
+      />
+    </SingleExampleContainer>
+    <SingleExampleContainer label="Automatically picks a readable text color based on fill color">
```
siddharthkp, line 212: This is really cool 👍👍👍

src/Token/TokenBase.tsx
mperrotti: If other components use "small", "medium", "large" - let's do that.

@colebemis specifically requested we use `size` instead of `variant` (which is what I originally had). I think `size` makes more sense, and we should keep it.

src/stories/Token.stories.tsx
mperrotti: When I use `children`, I expect to be able to pass a `ReactNode`, so it feels a little strange to limit it to `string`. Also, we use `text` instead of `children` in the [ActionList.Item component](https://github.com/primer/react/blob/9f39518224973dbc33dcc697fef1877ab2c4510c/src/ActionList/Item.tsx#L57).

This could just be a personal preference though, so I'm open to using `children` if other reviewers feel like this makes more sense.

siddharthkp: I'm okay with either, children feels like the way to put contents inside that already exists.

> When I use children, I expect to be able to pass a ReactNode, so it feels a little strange to limit it to string. 

That makes sense! I guess it's contextual. In the case of a label, you would pass a string in children. The type check is to warn us if it is stretched/misused beyond that.

> Also, we use text instead of children in the ActionList.Item component.

That's another one I'd like to change, proposal coming soon 😅

mperrotti: > That's another one I'd like to change, proposal coming soon 😅

Good to know!

src/Token/TokenBase.tsx
colebemis: +1 to using the full names for sizes (`small`, `medium`, `large`)

IMO `Label` should use `size` instead of `variant` as well.

src/Token/TokenLabel.tsx
colebemis: Hmm looks like this is consistent with how dotcom renders labels. Guess it does technically pass color contrast 
CleanShot 2021-10-05 at 13 43 33@2x

mperrotti: I just went with the same implementation we have for Memex

mperrotti:
> Hmm, I wonder why there's still a significant bundlesize increase

I'm not sure 🤔 - it could be the `color2k` dependency

mperrotti:
The `IssueLabelToken` is rendering with dark mode styles in Memex. I'm not sure if this is an issue with Memex or with the the primer-react implementation

PR #1489, opened 2021-09-30
Component that renders a text input with tokens

mperrotti:
------**_I will fill this out Friday 01 Oct_**------

---

Describe your changes here.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/stories/TextInputWithTokens.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const onTokenRemove: (tokenId: string | number) => void = (tokenId) => {
+      setTokens(tokens.filter(token => token.id !== tokenId))
+  };
+
+  return (
+    <Box display="flex" flexDirection="column" sx={{'gap': '1em'}}>
+      <TextInputTokens
+          tokens={tokens}
+          onTokenRemove={onTokenRemove}
+          tokenSizeVariant="sm"
```
colebemis, line 66: Could we just call this `size` since it affects the size of the text input as well?
```suggestion
          size="sm"
```

src/stories/TextInputWithTokens.stories.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { useCallback, useState } from 'react'
+import {Meta} from '@storybook/react'
+
+import { BaseStyles, Box, ThemeProvider } from '..'
+import TextInputTokens from '../TextInputWithTokens';
```
colebemis, line 5: Should this be TextInputWithTokens?
```suggestion
import TextInputWithTokens from '../TextInputWithTokens';
```

src/stories/TextInputWithTokens.stories.tsx
mperrotti: Good point. Agreed that we should update this prop name

colebemis:
Removing tokens doesn't seem to be working for the label story:

CleanShot 2021-10-04 at 15 23 45

docs/content/TextInputWithTokens.mdx
```diff
@@ -0,0 +1,40 @@
...
+There is no function that gets called to "add" a token, so the user needs to be provided with a UI to select tokens.
+
+For example, an [Autocomplete]() component may be used to select tokens:
+<!-- insert editable topic tags input? -->
+```tsx live
+```
+
+## Custom token rendering
+
+By default, the `Token` input is used to render the tokens in the input. If this component does not make sense for the kinds of tokens you're rendering, you can pass a component to the `tokenComponent` prop
```
colebemis, line 31:
```suggestion
By default, the `Token` component is used to render the tokens in the input. If this component does not make sense for the kinds of tokens you're rendering, you can pass a component to the `tokenComponent` prop
```

src/TextInputWithTokens.tsx
```diff
@@ -0,0 +1,247 @@
...
+  flex-grow: 1;
+`
+
+type AnyTokenProps = Partial<TokenProps & TokenLabelProps & TokenProfileProps>
+type TokenDatum = MandateProps<AnyTokenProps, 'id' | 'text'>
+type TextInputWithTokensInternalProps = {
+  /**
+   * The array of tokens to render
+   */
+  tokens: TokenDatum[]
```
colebemis, line 27: I made a quick prototype of how we could define the type of `tokens` based on the type of `tokenComponent`: https://codesandbox.io/s/frosty-swanson-y8vyq?file=/src/App.tsx

src/TextInputWithTokens.tsx
```diff
@@ -0,0 +1,247 @@
...
+  preventTokenWrapping?: boolean
+  /**
+   * The size of the tokens
+   */
+  tokenSizeVariant?: TokenSizeKeys
+  /**
+   * Whether the remove buttons should be rendered in the tokens
+   */
+  // TODO: confirm whether or not we're comfortable with allowing people to hide that button
+  hideTokenRemoveButtons?: boolean
```
colebemis, line 53: If we keep this option, should we add some check to ensure the the `tokenComponent` has a `hideRemoveButton` prop?

src/TextInputWithTokens.tsx
```diff
@@ -0,0 +1,247 @@
...
+      selectedTokenIdx,
+      setSelectedTokenIdx,
+      ...rest},
+    externalRef) => {
+    const ref = useProvidedRefOrCreate<HTMLInputElement>(externalRef as React.RefObject<HTMLInputElement>)
+    const { onFocus, onKeyDown, ...inputPropsRest } = omit(rest)
+
+    const handleTokenFocus: (tokenIdx: number) => FocusEventHandler = (tokenIdx) => () => {
+        setSelectedTokenIdx(tokenIdx)
+    }
```
colebemis, line 84:
```suggestion
    const handleTokenFocus: (tokenIndex: number) => FocusEventHandler = (tokenIndex) => () => {
        setSelectedTokenIdx(tokenIndex)
    }
```

src/TextInputWithTokens.tsx
```diff
@@ -0,0 +1,247 @@
...
+   * Whether the remove buttons should be rendered in the tokens
+   */
+  // TODO: confirm whether or not we're comfortable with allowing people to hide that button
+  hideTokenRemoveButtons?: boolean
+} & TextInputProps
+
+// using forwardRef is important so that other components (ex. Autocomplete) can use the ref
+const TextInputWithTokensComponent = React.forwardRef<HTMLInputElement, TextInputWithTokensInternalProps & {
+  selectedTokenIdx: number | undefined,
+  setSelectedTokenIdx: React.Dispatch<React.SetStateAction<number | undefined>>
```
colebemis, line 59: We generally prefer not to abbreviate shorter words like `index` and `event`:
```suggestion
  selectedTokenIndex: number | undefined,
  setSelectedTokenIndex: React.Dispatch<React.SetStateAction<number | undefined>>
```

src/TextInputWithTokens.tsx
```diff
@@ -0,0 +1,258 @@
...
+    onTokenRemove,
+    tokenComponent: TokenComponent,
+    preventTokenWrapping,
+    size,
+    hideTokenRemoveButtons,
+    selectedTokenIdx,
+    setSelectedTokenIdx,
+    ...rest
+  }: TextInputWithTokensInternalProps<TokenComponentType> & {
+    selectedTokenIndex: number | undefined
```
colebemis, line 74: Why is this state passed into the component instead of being managed internally?

mperrotti: Because it's being used in the `useFocusZone` hook in this `TextInputWithTokensComponent`

mperrotti:
> Is there a reason that the tokens are spans instead of buttons?

They felt more like spans than buttons semantically.

src/TextInput.tsx
```diff
@@ -165,13 +158,14 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
...
+        ref={wrapperRef}
       >
         {IconComponent && <IconComponent className="TextInput-icon" />}
-        <Input ref={ref} disabled={disabled} {...inputProps} />
+        <UnstyledTextInput ref={ref} disabled={disabled} {...inputProps} />
       </Wrapper>
     )
   }
-)
+) as Polymorphic.ForwardRefComponent<'input', TextInputInternalProps>
```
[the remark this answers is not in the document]
mperrotti: I added an `as` prop to `TextInput` so we didn't have to copy the wrapper styles. See https://github.com/primer/react/pull/1489/files#diff-769a5e1fabc6b7a64b7b20b3df7eebbe11778ec65cd6c9dd16d4c679b9d41347R220

src/TextInputWithTokens.tsx
mperrotti: `TextInputWithTokensInnerComponent` renders everything inside the grey border. 
`TextInputWithTokensComponent` renders the entire thing - grey border and all.

I'm a little foggy on _why_ I had to split it like this, but I think it had to do with passing refs. I'm going to try and combine them again to remind myself why I did it this way.

If there really is a reason they need to be separate, I'll document as clearly as possible using comments.

mperrotti: We may be able to merge these into one component, but I haven't figured it out quite yet.
Here's why these are separate:

1. To be able to pass a component with forwarded refs to the `as` prop of the `TextInput` component
2. To pass the ref from `useFocusZone` directly to the `TextInput`'s `Wrapper` component

PR #1490, opened 2021-09-30
Adds autocomplete components

mperrotti:
------**_I will fill this out Friday 01 Oct_**------

---

### Still needs to be done
- [ ] Update component API for `AutocompleteMenu`
    - [ ] add `onOpenChange` prop
    - [ ] add `onSelectedChange` prop to replace `onSelect` and `onDeselect`
- [ ] Improve Storybook stories
- [ ] Write tests
- [ ] Write docs

---

Describe your changes here.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp:
These are nit picks / tiny annoyances that I ran into while exploring the stories. I don't think any of them are deal breakers.


1. Chrome tries to autofill the input which leads to 2 autocompletes at the same time. We might be able to turn the native one off with `autocomplete="off"`

    autocomplete-annoying-chrome

&nbsp;

2. When using keyboard navigation, I (wrongly?) expected to open the menu with `Enter`, but found this edge case instead where the focus is still on the last element and it can be toggled on/off with `Enter`

    autcomplete-edge-case-return

docs/content/Autocomplete.mdx
```diff
@@ -0,0 +1,83 @@
+# Autocomplete
```
colebemis, line 1:
```suggestion
---
title: Autocomplete
status: Alpha
---
```

colebemis:
I noticed the deleting text in the text input requires two backspaces to delete one character. It also looks like the list in the menu doesn't update as your backspace:
CleanShot 2021-10-05 at 10 49 11

colebemis:
Looks like filtering is not working in any of the stories except Single Select. Is that expected?
CleanShot 2021-10-05 at 10 55 51

mperrotti:
Thanks @siddharthkp - great catch on both of those! I pushed up a fix

@colebemis - the issues you noticed should also be fixed. Those regressions were introduced when I memoized things.

colebemis:
Noticed a few odds behaviors in the live code examples:


1. In the [issue label example](https://primer-components-ba8otix1d-primer.vercel.app/react/Autocomplete#example-render-items-using-actionlistitem-props), the label colors reset when a new label is selected:

https://user-images.githubusercontent.com/4608155/137035488-997b67eb-7d48-4a67-989a-3109517e7407.m

2. In the [multiple values example](https://primer-components-ba8otix1d-primer.vercel.app/react/Autocomplete#select-multiple-values), selecting an item causes an error:

https://user-images.githubusercontent.com/4608155/137035808-461ef235-aca5-40ca-a854-7c39ed79e1d2.mp4

PR #1493, opened 2021-10-02
Adapt to right aligned items in live preview wrapper

SferaDev:
First of all, congratulations @pksjce on your great work with #1466! Just fixing a small visual bug I noticed with elements that either take the whole space or are right aligned (such as ``AvatarStack``).

### Screenshots

Before:

before

After:

after

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/src/@primer/gatsby-theme-doctocat/components/live-preview-wrapper.js
+++ b/docs/src/@primer/gatsby-theme-doctocat/components/live-preview-wrapper.js
@@ -1,13 +1,12 @@
 import {BaseStyles, Box, ThemeProvider, useTheme, DropdownMenu, DropdownButton} from '@primer/components'
 import React from 'react'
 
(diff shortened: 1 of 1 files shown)


pksjce:
@colebemis - I like this solution. I added a commit to remove all the hover behaviour. Happy to merge it but I don't seem to have permissions. 🤔

pksjce:
On second thought, not a fan of the varying widths based on which dropdown I select. 🤔

pksjce:
Hey @SferaDev , thanks for your contribution. But we already were in the process of changing this switcher due to internal feedback. I've raised #1502 which will address all the issues. Closing this in favor of that.

PR #1497, opened 2021-10-04
Updates ActionList's ItemInput type to accept DOM props for divs when renderItem is not provided

jfuchs:
This PR widens our `ActionList` component's `ItemInput` type so that DOM props for `div` are allowed. This is only a type change and should not change the function of the compiled code.

This surfaced when @smockle was upgrading PRC in an internal project from v30.0.0 to v30.1.0, and noticed that that project's usage of PRC included passing `onMouseDown` listeners in the `ItemInput` objects. This PR restores correct typing for that usage.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/ActionList/List.tsx
```diff
@@ -10,7 +10,9 @@ import {hasActiveDescendantAttribute} from '../behaviors/focusZone'
 
 type RenderItemFn = (props: ItemProps) => React.ReactElement
 
-export type ItemInput = ItemProps | ((Partial<ItemProps> & {renderItem: RenderItemFn}) & {key?: Key})
+export type ItemInput =
```
smockle, line 13: Since `ActionList`’s `items` type now diverges from `ItemProps`, does the [props documentation for `ActionList`](https://primer.style/react/ActionList#props) need to be updated?

smockle (approved):

src/ActionList/List.tsx
jfuchs: Good catch, thanks!

PR #1505, opened 2021-10-06
Introduces a Merge<> utility and a type testing pattern

jfuchs:
This PR:

- Adds a new utility type to merge two objects
- Introduces a pattern for testing our types
- Refactors `src/util/types.ts` into separate files per utility
- Narrows the test matching for jest to exclude type testing

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


pksjce:
🤔 Is it a pain point within the team that we are constantly running into typing errors during certain situations like integration? Or has it happened a lot that we've broken types during releases? 
AFAIK, types that a component library exposes are 
- Props
- Callback arguments
- Callback return values
These should get tested in tests and stories.(Ideally we should have integration testing too so that all interactions have sufficient coverage along with opportunity to automate visual testing.) 
Avoiding impossible states should be api design driven rather than only type restricted. But this is where I see most value in this tool. We can specifically test that we haven't opened up impossible states to our consumers. It's good dx and educational to an extent too.
We should add this to our contributor docs and keep an eye out in PRs so people can be made aware to use these tests.

PR #1517, opened 2021-10-15
ActionList Composable API

siddharthkp:
Ready for review :) 

----

Closes https://github.com/github/primer/issues/307 | [Interface guidelines](https://primer.style/design/components/action-list) | [Figma](https://www.figma.com/file/GCvY3Qv8czRgZgvl1dG6lp/Primer-Web?node-id=9677%3A44019)
Next: https://github.com/github/primer/issues/447 + https://github.com/github/primer/issues/448

<br> See [Docs](https://primer-components-git-refs-headsaction-list-2-primer.vercel.app/react/ActionList2)  for API surface area for ActionList

&nbsp;

<br>See [Storybook](https://primer-components-git-refs-headsaction-list-2-primer.vercel.app/react/storybook?path=/story/composite-components-actionlist2--simple-list-story) for use cases as standalone and as List inside Menu/Select

&nbsp;

### API opinions worth noting:

[Watch video instead](https://github.rewatch.com/video/wlhv8hgmzz4a8lq1-actionlist2-api-decisions)

1. Prefer composition over config: See https://github.com/github/primer/issues/307 for long version

2.  `selectionVariant` for `ActionList` is null by default. Item makes no assumptions about what kind of selection to render if it is not set on the List root or Group
   
3. `showDividers` is on `ActionList` root, not each `ActionList.Item`. Loosely held opinion, will test with memex. The Dividers are meant to add a separator between Items of the list and should be consistent in a given List and not customisable at the Item level. 
   - Should it also be on Group level? Can 2 groups inside a List have different divider settings?
   - Does the name `showDividers` cause confusion/conflict with `ActionList.Divider`?

4. Added `ActionList.LinkItem` which uses an internal API with `Item` to make sure we can give the semantics for free.
 

---

Progressive improvements: 
- [x] Try out rc version in memex https://github.com/github/memex/pull/5862
- [x] Play well with AnchoredOverlay for "build your own ActionMenu" mode in memex
- [x] replace custom item themes when https://github.com/primer/primitives/pull/263 is released
- [x] help wanted: typescript unhappy with polymorphic + forwardRef
- [x] ActionList.LinkItem
   - [x] Polymorphic Link component
   - [x] Padding at the container not root
- [x] Test suite for create-slots (it's where all the complex logic sits)

Questions:

Design
- [x]  padding: 8 or 6? x:8, y:6
- [x]  different size for icon and avatar, range?
- [x] Async results UX → https://github.com/github/primer/issues/457

Bundling
- [x] Bundle `ActionList2` in `@primer/components/unreleased`

While I'm confident about the API in this PR, I'd like to "release" these changes along with their composable counterparts ActionMenu and SelectPanel, so that the memex team can catch up with breaking API for Menus together.  This also gives us a change to tweak and polish the underlying ActionList to be accessible. The change is not code aesthetics, it's a more mature group of components ✨ 

Accessibility questions moved to [ActionList beta audit](https://github.com/github/primer/issues/447)

Merge checklist
- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

[^1]: Native checkbox styles image


colebemis:
@siddharthkp I love where this is going! ❤️ 

```js
const {Slots, Slot} = createSlots([
  'LeadingVisual', 'InlineDescription', 'BlockDescription', 'TrailingVisual'
])
```

How would an API like this handle a "renders many" situation? For example, an `ActionList` may want to render many `Item` slots. Maybe we could borrow a config similar to ViewComponent:

```js
const {Slots, Slot} = createSlots({
  rendersOne: ['Header'],
  rendersMany: ['Item']
})
```

Then we could store "Items" in an array (the pluralization bit might get a little tricky)


    <Slots>
      {slots => (
        <Box>
          {slots.Header}
          {slots.Items.map(item => <Box>{item}</Box>)}
        </Box>
      )}
    </Slots>

pksjce:
Interesting concept with the double render strategy! I checked it out in the new profiler. 
https://user-images.githubusercontent.com/417268/138023524-7cd08a34-c9bf-4db5-b0b0-981e59450305.mov

**Performance**
It re-renders all of the slots present like you mentioned. The first render takes a lot longer than the second as the re-render is on the `Slots` subset only. 

Another observation is how the `unregister` on the dynamic sections makes the whole `Slots` re-render selectively. So when the `children` prop changes in the dynamic `Slot`, it is unregistered from the `context` which causes a `forceUpdate`.

https://user-images.githubusercontent.com/417268/138023960-f54c8b33-3799-45f1-834a-df62fadbc242.mov

I do think the re-renders are a bit too generous in case of the dynamic example. Just something to keep in mind I guess. Perf degradation could stand out during edge cases in the consumer ui. 

The old component, of course, doesn't support such complex usecases. But worth noting it just renders everything once.

**Accessibility**

I wonder how this will be handled? The `Item` component is missing keyboard accessibility. As for dynamic components, its upto the component to announce changes I guess?

mperrotti:
Wow, this has come far since I last took a look! Thanks for going into detail about all the different possible directions.

package.json
```diff
@@ -124,6 +124,8 @@
     "lodash.isobject": "3.0.2",
     "prettier": "2.3.2",
     "react": "17.0.2",
+    "react-dnd": "14.0.4",
+    "react-dnd-html5-backend": "14.0.2",
```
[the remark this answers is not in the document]
mperrotti: If these are devDependencies, will they still be available in the [live Storybook deployment](https://primer.style/react/storybook)?

src/sx.ts
```diff
@@ -1,4 +1,5 @@
 import css, {SystemStyleObject} from '@styled-system/css'
+import merge from 'deepmerge'
```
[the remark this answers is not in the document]
mperrotti: FWIW, I think we're getting rid of sx props.

@jfuchs - correct me if I misunderstood what you meant by "getting rid of styled-components props" in this morning's Primer React planning.

src/ActionList2/Group.tsx
```diff
@@ -0,0 +1,33 @@
...
+    selectionVariant?: ListProps['selectionVariant'] | false
+  }
+
+type ContextProps = Pick<GroupProps, 'selectionVariant'>
+export const GroupContext = React.createContext<ContextProps>({})
+
+export const Group: React.FC<GroupProps> = ({title, variant, auxiliaryText, selectionVariant, sx = {}, ...props}) => {
+  return (
+    <Box
+      as="li"
```
mperrotti, line 18: Looks like Typescript isn't happy because of the `as="li"` 🤔

src/ActionList2/Header.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+/**
+ * Displays the name and description of a `Group`.
+ */
+export function Header({
+  variant = 'subtle',
+  title,
+  auxiliaryText,
+  children: _children,
```
mperrotti, line 36: Why not just do
`delete props.children`?

src/ActionList2/Header.tsx
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Secondary text which provides additional information about a `Group`.
+   */
+  auxiliaryText?: string
+}
+
+/**
+ * Displays the name and description of a `Group`.
+ */
+export function Header({
```
mperrotti, line 32: Why are we defining this as a named function, but we're defining other components as a `const`?

mperrotti (commented):
Loving where this is at right now. I've only had a chance to review actually usage in the MDX docs and Storybook stories.

Expect more feedback tomorrow. I think I'll have more helpful feedback after taking a little break and looking at usage in your [Memex Experiment PR](https://github.com/github/memex/pull/5862)

docs/content/ActionList2.mdx
```diff
@@ -0,0 +1,263 @@
+---
+title: ActionList
```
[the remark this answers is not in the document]
siddharthkp: yes I did! Had to clean up gatsby cache and build again

docs/content/ActionList2.mdx
```diff
@@ -0,0 +1,263 @@
...
+---
+
+import {BorderBox, Avatar} from '@primer/components'
+import {ActionList} from '@primer/components/experiments'
+import {Props} from '../src/props'
+
+import {ImageContainer} from '@primer/gatsby-theme-doctocat'
+import {LinkIcon, AlertIcon, ArrowRightIcon} from '@primer/octicons-react'
+
+<br />
```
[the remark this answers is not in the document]
siddharthkp: More white space in the docs, it's looking a bit crowded. 😅 We could remove some of these by increasing margin in doctocat

docs/content/ActionList2.mdx
```diff
@@ -0,0 +1,263 @@
...
+      primer-css
+      <ActionList.Description>GitHub Design Systems Bot</ActionList.Description>
+    </ActionList.Item>
+  </ActionList>
+)
+```
+
+<br />
+
+# Props / API reference
```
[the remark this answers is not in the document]
siddharthkp: I did, it's not very reliable 😢

Haven't tried to figure out why in this context though. Long term solution discussion here: https://github.com/primer/react/discussions/1492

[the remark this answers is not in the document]
siddharthkp: You're right! Will change this into h2

src/ActionList2/Description.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import Box from '../Box'
+import {SxProp, merge} from '../sx'
+import Truncate from '../Truncate'
+import {Slot, ItemContext} from './Item'
+
+export type DescriptionProps = {
+  variant?: 'inline' | 'block'
```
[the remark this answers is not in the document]
siddharthkp: Oh no no, we do want jsdoc in the code as well. I just missed it here. Will add it.

src/ActionList2/Divider.tsx
```diff
@@ -0,0 +1,23 @@
...
+import {Theme} from '../ThemeProvider'
+
+/**
+ * Visually separates `Item`s or `Group`s in an `ActionList`.
+ */
+export function Divider(): JSX.Element {
+  return (
+    <Box
+      as="li"
+      role="separator"
```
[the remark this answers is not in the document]
siddharthkp: Not sure if it's consistent across screen readers. Will have more info after a11y audit this week :)

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,197 @@
...
+   * - `"default"` - An action `Item`.
+   * - `"danger"` - A destructive action `Item`.
+   */
+  variant?: 'default' | 'danger'
+  /**
+   * Items that are disabled can not be clicked, selected, or navigated through.
+   */
+  disabled?: boolean
+  /**
+   * The ARIA role describing the function of `Item` component. `option` is a common value. |
```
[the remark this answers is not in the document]
siddharthkp: In the context of a list, it's not always an option. It can also be a list of links in some contexts. 

When it's used inside an ActionMenu, that is when we will pass a value for it without needing to ask

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,197 @@
...
+      paddingX: get('space.2'),
+      paddingY: '6px',
+      marginX: listVariant === 'inset' ? get('space.2') : 0,
+      minHeight: get('space.5'),
+      borderRadius: get('radii.2'),
+      transition: 'background 33.333ms linear',
+      color: getVariantStyles(variant, disabled).color,
+      textDecoration: 'none', // for as="a"
+
+      ':not([aria-disabled])': {cursor: 'pointer'},
```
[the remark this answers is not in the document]
siddharthkp: No need because we set aria-disabled ourselves :)

https://github.com/primer/react/pull/1517/files/a86b13b99321cb5b43159c96cb4b17f016e80b8d#diff-6b3f2a26879573bcf7068c2c16107ae50982e6aff5d735478c13ca231dd8e22bR152

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,197 @@
...
+          </Box>
+        )}
+      </Slots>
+    )
+  }
+) as PolymorphicForwardRefComponent<'li', ItemProps>
+
+Item.displayName = 'ActionList.Item'
+
+const ConditionalBox: React.FC<{if: boolean} & BoxProps> = props => {
```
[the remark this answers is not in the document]
siddharthkp: Maaaybe. I'm not too confident if this is a good idea to be honest

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const {if: condition, ...rest} = props
+
+  if (condition) return <Box {...rest}>{props.children}</Box>
```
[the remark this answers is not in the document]
siddharthkp: sure i don't mind either :)

src/ActionList2/List.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+type ContextProps = Omit<ListProps, 'sx'>
+export const ListContext = React.createContext<ContextProps>({})
+
+export const List = React.forwardRef<HTMLUListElement, ListProps>(
+  ({variant = 'inset', selectionVariant, showDividers = false, sx = {}, ...props}, forwardedRef): JSX.Element => {
+    const styles = {
+      margin: 0,
+      fontSize: get('fontSizes.1'),
+      lineHeight: '20px', // TODO: check if we replace this already
```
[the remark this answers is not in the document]
siddharthkp: yes! will add to my TODOs, something i have to check 🤔

src/ActionList2/List.tsx
```diff
@@ -0,0 +1,54 @@
...
+      paddingY: variant === 'inset' ? get('space.2') : 0
+    }
+
+    return (
+      <Box
+        as="ul"
+        sx={merge(styles, sx as SxProp)}
+        aria-multiselectable={selectionVariant === 'multiple' ? true : undefined}
+        {...props}
+        ref={forwardedRef}
```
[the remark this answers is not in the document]
siddharthkp: I thought that's what the `as PolymorphicForwardRefComponent<'ul', ListProps>` will help with. But i'm missing something here :(

package.json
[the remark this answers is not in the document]
siddharthkp: yep :) Check it out - [Memex story](https://primer-components-git-refs-headsaction-list-2-primer.vercel.app/react/storybook?path=/story/composite-components-actionlist2--memex-sortable)

src/ActionList2/Description.tsx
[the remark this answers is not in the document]
siddharthkp: Done!

src/ActionList2/Group.tsx
siddharthkp: Yeah, any idea why? :( 

cc @colebemis @jfuchs?

src/ActionList2/Header.tsx
siddharthkp: Because I copied this from the previous Item implementation 🤣 Changed now.

src/ActionList2/Header.tsx
siddharthkp: I think that's a no-no because props isn't a mutable object

src/ActionList2/Item.tsx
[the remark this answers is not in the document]
siddharthkp: Changed my mind! Personal opinion, it's harder to parse 🤔 

```js
return condition ? <Box {...rest}>{props.children}</Box> : <>{props.children}</>
```
vs.
```js
if (condition) return <Box {...rest}>{props.children}</Box>
else return <>{props.children}</>
```

src/ActionList2/Group.tsx
mperrotti: I wish I had some helpful guidance, but I'm stumped here too :/

docs/content/ActionList2.mdx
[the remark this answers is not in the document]
mperrotti: Thanks! I'll give that a try

docs/content/ActionList2.mdx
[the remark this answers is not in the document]
mperrotti: Agreed that increasing the margin in doctocat would be better, but this is totally fine for now

docs/content/ActionList2.mdx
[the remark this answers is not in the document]
mperrotti: Thanks for creating that issue! Hopefully we can get something more reliable someday soon.

src/ActionList2/List.tsx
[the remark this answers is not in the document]
siddharthkp: aww, we don't have a lineHeights scale in our theme. Will replace the magic number with a variable

src/ActionList2/Item.tsx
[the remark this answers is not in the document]
mperrotti: Ok, then let's keep what you have. Thanks for giving it a try :)

src/ActionList2/List.tsx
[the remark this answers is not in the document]
siddharthkp: Done!

src/ActionList2/Visuals.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+type VisualProps = SxProp & {children: React.ReactNode}
+
+export const LeadingVisualContainer: React.FC<SxProp> = ({sx = {}, ...props}) => {
+  return (
+    <Box
+      as="span"
+      sx={merge(
+        {
+          height: '20px', // match height of text row
```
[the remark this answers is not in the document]
siddharthkp: Done!

src/sx.ts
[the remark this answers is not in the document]
siddharthkp: i don't think he was talking about the sx prop 😮

src/stories/ActionList2.stories.tsx
```diff
@@ -0,0 +1,1278 @@
...
+          </ActionList.LinkItem>
+          <ActionList.LinkItem href="https://github.com/primer" target="_blank" rel="noopener noreferrer">
+            <ActionList.LeadingVisual>
+              <LinkIcon />
+            </ActionList.LeadingVisual>
+            ActionList.LinkItem with anchor attributes
+          </ActionList.LinkItem>
+          <ActionList.LinkItem
+            as={ReactRouterLikeLink}
+            to="?path=/story/composite-components-actionlist2--simple-list-story"
```
siddharthkp, line 593: help wanted: How do we add type support for arbitrary components and their props?

mperrotti:
I think we need to set `list-style: none` on the `<ul>`
Screen Shot 2021-11-02 at 2 55 31 PM

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,214 @@
...
+      annotationColor: 'fg.muted'
+    }
+  }
+
+  switch (variant) {
+    case 'danger':
+      return {
+        color: 'danger.fg',
+        iconColor: 'danger.fg',
+        annotationColor: 'fg.muted'
```
mperrotti, line 27: Are we sure we don't want the annotationColor to be `danger.fg`?

I like how `fg.muted` matches the other annotations
Screen Shot 2021-11-02 at 3 10 54 PM

However, I could also see some value in showing a "danger" color to reinforce that it's a dangerous action:
Screen Shot 2021-11-02 at 3 10 41 PM


I'd love to hear if other reviewers have any thoughts on this.

mperrotti (commented):
I spent some more time reviewing the docs, Storybook stories, and Memex experiment PR. I'm very happy with these changes, and I'll approve the PR as soon as we resolve the Typescript issues.

I'd be happy to pair with you to get those resolved, and I'd recommend trying to get @jfuchs to take a look. He was very helpful with Typescript issues in my Autocomplete work.

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,214 @@
...
+    },
+    forwardedRef
+  ): JSX.Element => {
+    const customItemTheme = customItemThemes[variant]
+    const {variant: listVariant, showDividers} = React.useContext(ListContext)
+
+    const {theme} = useTheme()
+
+    const styles = {
+      display: 'flex',
```
mperrotti, line 100: I think flexbox is perfectly fine for this, but @langermank is using grid with named areas in the @primer/css implementation of ActionList items

langermank: If we could confidently use `gap` with `flex` I might have gone that direction! But here's the grid action https://github.com/primer/css/blob/main/src/actionlist/action-list-item.scss#L291

mperrotti: Great point about `gap`. I just had to deal with that in tokens.

langermank: When `gap` has more Safari support you best believe I'll be spamming all our slack channels 😄

docs/content/ActionList2.mdx
```diff
@@ -0,0 +1,246 @@
...
+
+<br />
+<br />
+
+# Examples
+
+## Minimal example
+
+```javascript live noinline
+const {ActionList} = experiments
```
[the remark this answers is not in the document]
siddharthkp: Yep! I'd really like a way to hide that but not sure without calling making the jsx look like this:

```jsx
<ActionList2>
  <ActionList2.Item>New file</ActionList2.Item>
  <ActionList2.Item>Copy link</ActionList2.Item>
</ActionList2>
```

src/sx.ts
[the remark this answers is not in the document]
colebemis: We've moving from styled-system props (`mr={2}`) to the sx prop (`sx={{mr: 2}}`). `sx` isn't going anywhere. If we change the CSS-in-JS library we use, we'll probably still want a similar safe escape hatch.

src/theme-preval.js
```diff
@@ -1,5 +1,6 @@
 // @preval
 // This file needs to be a JavaScript file using CommonJS to be compatiable with preval
+// Cache bust: 2021-11-04 12:00:00 GMT (This file is cached by our deployment tooling, update this timestamp to rebuild this file)
```
colebemis, line 3: TIL

src/utils/create-slots.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react'
+import {useForceUpdate} from './use-force-update'
+
+/** createSlots is a factory that can create a
+ *  typesafe Slots + Slot pair to use in a component definition
+ *  For example: ActionList.Item uses createSlots to get a Slots wrapper
+ *  + Slot component that is used by LeadingVisual, Description
+ */
+const createSlots = <SlotNames extends string>(slotNames: SlotNames[]) => {
```
colebemis, line 9: This could be a useful standalone open source library eventually 💡

src/experiments.ts
```diff
@@ -0,0 +1,2 @@
+// Components
+export * from './ActionList2'
```
colebemis, line 2: I like the idea of keeping experiments out of the main bundle. A couple of questions:

* If we want to build experimental components in primer, should we update the definition of "experimental" in this document? https://primer.style/contribute/component-lifecycle#experimental (defines "experimental" as built outside of Primer)
* Should we rename this file `experimental` to align with that document? ☝️ 
* When we release breaking changes to experimental components, which semver number should we bump?

src/ActionList2/Visuals.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import Box from '../Box'
+import {SxProp, merge} from '../sx'
+import {get} from '../constants'
+import {getVariantStyles, Slot, ItemContext, TEXT_ROW_HEIGHT} from './Item'
+
+type VisualProps = SxProp & {children: React.ReactNode}
+
+export const LeadingVisualContainer: React.FC<SxProp> = ({sx = {}, ...props}) => {
```
colebemis, line 9:
```suggestion
export const LeadingVisualContainer: React.FC<VisualProps> = ({sx = {}, ...props}) => {
```

src/ActionList2/Visuals.tsx
```diff
@@ -0,0 +1,76 @@
...
+          display: 'flex',
+          justifyContent: 'center',
+          alignItems: 'center',
+          flexShrink: 0,
+          marginRight: 2
+        },
+        sx as SxProp
+      )}
+      {...props}
+    />
```
colebemis, line 27: Do we need to spread `...props` here? It looks like we only accept the `sx` prop

src/ActionList2/Visuals.tsx
```diff
@@ -0,0 +1,76 @@
...
+          height: TEXT_ROW_HEIGHT, // match height of text row
+          minWidth: get('space.3'),
+          maxWidth: TEXT_ROW_HEIGHT, // square (same as height)
+          display: 'flex',
+          justifyContent: 'center',
+          alignItems: 'center',
+          flexShrink: 0,
+          marginRight: 2
+        },
+        sx as SxProp
```
colebemis, line 24: Why is the type cast here necessary? Shouldn't this come from the type definition?

src/ActionList2/List.tsx
```diff
@@ -0,0 +1,55 @@
...
+   */
+  variant?: 'inset' | 'full'
+  /**
+   * Whether multiple Items or a single Item can be selected.
+   */
+  selectionVariant?: 'single' | 'multiple'
+  /**
+   * Display a divider above each `Item` in this `List` when it does not follow a `Header` or `Divider`.
+   */
+  showDividers?: boolean
```
colebemis, line 19: I agree that this `showDividers` name could be confusing when used with `ActionList.Divider`. But I can't think of a better name. Let's move forward with this for now 👍

src/ActionList2/List.tsx
```diff
@@ -0,0 +1,55 @@
...
+} & SxProp
+
+type ContextProps = Omit<ListProps, 'sx'>
+export const ListContext = React.createContext<ContextProps>({})
+
+const ListBox = styled.ul<SxProp>(sx)
+
+export const List = React.forwardRef<HTMLUListElement, ListProps>(
+  (
+    {variant = 'inset', selectionVariant, showDividers = false, sx: propsSx = {}, ...props},
```
colebemis, line 33: [nit] `sxProp` feels like a slightly better name than `propsSx`
```suggestion
    {variant = 'inset', selectionVariant, showDividers = false, sx: sxProp = {}, ...props},
```

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,228 @@
...
+} & SxProp
+
+const {Slots, Slot} = createSlots(['LeadingVisual', 'InlineDescription', 'BlockDescription', 'TrailingVisual'])
+export {Slot}
+export type ItemContext = Pick<ItemProps, 'variant' | 'disabled'> & {
+  inlineDescriptionId: string
+  blockDescriptionId: string
+}
+
+const LiBox = styled.li<SxProp>(sx)
```
colebemis, line 85: I read this as "lil box" haha 😍

colebemis (approved):
Love where this is going ❤️  Amazing work, @siddharthkp!

Happy to move forward with what you have here. Left a few minor comments.

src/experiments.ts
siddharthkp: Hmm, there's a name conflict here for sure.

In our component lifecycle maturity, components with experimental maturity can still be in the main bundle.

What I want to signal here is - this is a "unreleased" component that is not part of the public api yet (not in main bundle). We don't recommend using it in production yet. You can still import it with explicitly for experimentation/feedback.

alternate names: unreleased, next, future, private, internal

src/ActionList2/Visuals.tsx
siddharthkp: Good catch!

src/ActionList2/Visuals.tsx
siddharthkp: Oh my bad, it should accept html attributes as well. Added them to type.

src/ActionList2/Visuals.tsx
siddharthkp: deepmerge loses the definition on its way back :( Might be useful to create a custom deepmerge function that understands sx

src/ActionList2/List.tsx
siddharthkp: You're right. Changed it for List and Item

src/ActionList2/Item.tsx
siddharthkp: i mean, you're not wrong 😋

src/experiments.ts
siddharthkp: Update: changed to `unreleased`.

(`next` was a close second, but that might get confused with npm tag `primer/components@next`. `future` is romantic, but i'd like to put long running things in this scope that might not make it to main bundle at all)

@pksjce thoughts for new Button?

src/utils/create-slots.tsx
siddharthkp: One of these days 😉

src/theme-preval.js
siddharthkp: There has to be a better way of doing this, but I did not spend more than 5 minutes researching vercel's caching mechanism

src/experiments.ts
siddharthkp: (Useful to add, also added to PR description)

Why is ActionList2 in `unreleased` instead of replacing the current ActionList with a breaking change?

While I'm confident about the API in this PR (by testing it A LOT 😇), I'd like to "release" these changes along with their composable counterparts ActionMenu and SelectPanel, so that the memex team can catch up with breaking API for Menus together. This also gives us a change to tweak and polish the underlying ActionList to be accessible.

mperrotti (approved):

src/ActionList2/Item.tsx
```diff
@@ -0,0 +1,228 @@
...
+        hoverColor: 'fg.default'
+      }
+  }
+}
+
+export type ItemProps = {
+  /**
+   * Primary content for an Item
+   */
+  children?: React.ReactNode
```
mperrotti, line 44: Do we need to specifically define `React.ReactNode`?

mperrotti (approved):

src/ActionList2/Item.tsx
siddharthkp: I think so 🤔  Added this here because that's what we say in the docs

PR #1521, opened 2021-10-18
ActionList/ActionMenu API: Add trailingVisual alias

siddharthkp:
- [x] mark `trailingIcon` and `trailingText` as deprecated
- [x] prefer `trailingVisual` but support old api as fallback 
- [x] Update docs: add `trailingVisual` example

Use as text:

```jsx
<ActionList
  items={[
    {text: 'New file', trailingVisual: '⌘N'},
    {text: 'Edit file', trailingText: '⌘E'}, // backward compatible
    {text: 'Copy link'},
    {text: 'Delete file', variant: 'danger'}
  ]}
/>
```

---

Freeform, use with icon:

```jsx
<ActionList 
  items={[{
    leadingVisual: SearchIcon,
    text: 'repo:github/github',
    trailingVisual: () => (
      <>
        ⌘S<ArrowRightIcon />
      </>
    )
  },
  {
    leadingVisual: SearchIcon,
    text: 'repo:github/github',
    trailingText: '⌘S', // backward compatible
    trailingIcon: ArrowRightIcon // backward compatible
  }]}
/>
```

---

Question: Should we drop direct `string` support and only support a component type to keep it consistent with `leadingVisual`?

```diff
items={[
  {
    text: 'New file',
-   trailingText: '⌘N'
-   trailingVisual: '⌘N'
+   trailingVisual: () => <>⌘N</>
  }
]}

// types:
- trailingVisual?: React.FunctionComponent<IconProps> | string
+ trailingVisual?: React.FunctionComponent
```

### DIFF
--- a/docs/content/ActionList.mdx
+++ b/docs/content/ActionList.mdx
@@ -32,7 +32,7 @@ An `ActionList` is a list of items which can be activated or selected. `ActionLi
     {groupId: '4'}
   ]}
   items={[
-    {key: '1', leadingVisual: TypographyIcon, text: 'Rename', groupId: '0'},
+    {key: '1', leadingVisual: TypographyIcon, text: 'Rename', groupId: '0', trailingVisual: '⌘R'},
     {key: '2', leadingVisual: VersionsIcon, text: 'Duplicate', groupId: '0'},
     {key: '3', leadingVisual: SearchIcon, text: 'repo:github/github', groupId: '1'},
     {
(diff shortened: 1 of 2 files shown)


siddharthkp:
@dusave Oh right! Updated :)

pksjce:
I was wondering what's the benefit of combining these two props to one? Is it to reduce api surface area?

pksjce:
Ah I see, so if `leadingVisual` needed to have visual and text, we could just pass in a component to do that. I guess same could apply to `trailingVisual`. You could type it only as a react component and that could include text albeit slightly more markup?

siddharthkp:
> Ah I see, so if leadingVisual needed to have visual and text, we could just pass in a component to do that. I guess same could apply to trailingVisual. You could type it only as a react component and that could include text albeit slightly more markup?

`leadingVisual` is a bit more locked in. There's only space for either an Icon or Avatar. I merged the types for `trailingText` and `trailingIcon` for `trailingVisual`, but we could probably put `React.ReactNode` and be good - depending on how flexible we want to be with that space.

PR #1524, opened 2021-10-19
Update `Button` styles

moatorres:
I don't think that `@primer/primitives` has either `hoverText` or `hoverShadow` defined in its primary button color schemes.

`btn.danger` and `btn.outline` do so, and they're currently the same. We can check that [here](https://github.com/primer/primitives/blob/main/data/colors/vars/component_light.ts). 

```ts
// btn.outline.hoverText, btn.danger.hoverText
get('scale.white') 
// prints → #ffffff

// btn.outline.hoverText, btn.danger.hoverText
(theme: any) => `0 1px 0 ${alpha(get('scale.black'), 0.1)(theme)}`
// prints → '0 1px 0 rgba(27,31,36,0.1)'
```

Perhaps we should update `@primer/primitives` then update it here? We could also remove it altogether as it's probably not rendering anything.

I'm guessing that, if they're supposed to be the same, they could go up one level (`btn.hoverText`, `btn.hoverShadow`) and be shared across all buttons.

PS: Just found another "missing" style in `<ButtonDanger />`. Apparently there's no `btn.danger.disabledBorder` defined.

```ts
border-color: ${get('colors.btn.danger.disabledBorder')};
```

Let me know your thoughts. Happy to make any adjustments needed, either way. 

- [x] No tests were affected

### DIFF
--- a/src/Button/ButtonPrimary.tsx
+++ b/src/Button/ButtonPrimary.tsx
@@ -11,10 +11,10 @@ export const ButtonPrimary = styled(ButtonBase)<ButtonBaseProps & ButtonSystemPr
   box-shadow: ${get('shadows.btn.primary.shadow')};
 
   &:hover {
-    color: ${get('colors.btn.outline.hoverText')};
+    /* color: ${get('colors.btn.primary.hoverText')}; */
     background-color: ${get('colors.btn.primary.hoverBg')};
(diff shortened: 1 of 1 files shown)


moatorres:
No worries! Glad to help.

PR #1532, opened 2021-10-25
New Button component

pksjce:
Describe your changes here.

Building up the new button as per recommendations from https://github.com/github/primer/issues/382 and https://github.com/github/primer/issues/253

Things done

- Single button component
- No outline button
- Add icon only button
- Add LeadingIcon
- Add Counter
- Add Caret button
- Story for block button

Todos

- [x] Linter issues
- [x] Type issues with forwarding
- [x] Make all Button sizes correct
- [x] Dialog component uses the Button wierdly for close button. Try and fix this
- [x] Figure out all the spacings between slots from the audit
- [ ] How will we make button changes in watch/upvote more a11y friendly
- [ ] Attaching descriptions to buttons must be form group functionality
- [x] Edge cases of button variants + icons + caret with different button sizes
- [ ] Button as Link 

### Screenshots

TBD

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


pksjce:
Hi! I created a Loom to demo this component. 
Would be happy to receive any feedback! 🤞 

https://www.loom.com/share/7b870fd58d084ad7bedf4f8e9f57ed90

src/NewButton/button.tsx
```diff
@@ -0,0 +1,220 @@
...
+  }
+  return style[variant]
+}
+
+const getSizes = (size = 'medium', variant: VariantType = 'default', iconOnly: boolean) => {
+  let paddingTop, paddingLeft
+  switch (size) {
+    case 'small':
+      paddingTop = 3
+      paddingLeft = 12
```
siddharthkp, line 149: question: Are these both pixel values or is 3 a value on the scale?

src/stories/NewButton.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+  return (
+    <Button {...args} variant="invisible">
+      Invisible
+    </Button>
+  )
+}
+
+export const iconBeforeButton = (args: ButtonProps) => {
+  return (
+    <Button icon={() => <SearchIcon />} {...args}>
```
siddharthkp, line 65: Do we want to allow folks to pass props to the icon itself (size/color) or is that always dictated by the Button size+variant. 

If it's the latter, we might want to promote usage like this in the docs: `Button icon={SearchIcon}>`

src/stories/NewButton.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+    </>
+  )
+}
+
+export const WatchCounterButton = ({...args}: ButtonProps) => {
+  const [count, setCount] = useState(0)
+  return (
+    <Button onClick={() => setCount(count + 1)} {...args}>
+      Watch
+      <Button.Counter count={count} />
```
siddharthkp, line 87: Multiple suggestions here:

For the prop, we can can use children instead of introducing a prop `count`

```suggestion
      <Button.Counter>{count}</Button.Counter>
```

On the Counter component as a whole, curious to know your decision between `Button.Counter` component and `count` prop

theinterned:
Hello, I'm new and probably missed it, but I was wondering if there was a documented reason we are moving from separate components for button variants to `variant` prop?

src/NewButton/button.tsx
```diff
@@ -0,0 +1,220 @@
...
+    if (Icon && !children) {
+      iconOnly = true
+    }
+    const iconWrapStyles = {
+      display: 'inline-block',
+      ...(!iconOnly ? {mr: 2} : {})
+    }
+    return (
+      <ButtonBase ref={forwardedRef} {...props} iconOnly={iconOnly}>
+        {Icon && (
```
langermank, line 192: I have some thoughts on icon only buttons 😄 firstly, I love that you pointed out they should be square in your loom 🎉 

For a11y, this is my [favorite article](https://www.sarasoueidan.com/blog/accessible-icon-buttons/) on icon only buttons from Sara Soueidan. I have personally implemented `Technique #2: Accessible Visually Hidden Text with hidden and aria-labelledby` a few times and its always worked out nicely. The one caviet is that you need to wrap `children` with a span or div, which adds another layer of dom elements (there is something quite nice about a button with just text as a child)

As a quick example:

```
<ButtonBase aria-labelledby="button-label">
{Icon && (
// icons should always be hidden as they are decorative
// aria-hidden="true" focusable="false"
   <Box sx={iconWrapStyles}>
            <Icon size={size} aria-hidden="true" focusable="false" />
    </Box>
)}
        <span id="button-label" hidden={iconOnly}>{children}</span>
</ButtonBase>
```

With this implementation, `children` is _always_ required. If `iconOnly` is `true` the label will be visually hidden and `aria-labelledby` will go to work for a screen reader. The nice thing about this API is that the consumer doesn't need to do anything other than always provide a descriptive button label and a prop `iconOnly`. 

This could also be used for a loader button if we want to announce a loading message to the user.

For buttons that have icons and a visible label, these same props should be true for icons: `aria-hidden="true" focusable="false"`

src/NewButton/button.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <Box sx={iconWrapStyles} aria-hidden={!iconOnly}>
+            <Icon size={size} />
+          </Box>
+        )}
+        {children}
+        {caret && (
```
langermank, line 198: I'm wondering if this could be more generic, though perhaps we don't have a use case for a `trailingIcon` in button other than a dropdown menu? My instinct is that button should support `leading` and `trailing` icons (generic slots) and the Dropdown component should utilize `trailingIcon` for `<TriangleDownIcon>`.

src/NewButton/button.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+const Button = forwardRef<HTMLAnchorElement | HTMLButtonElement, ComponentProps<typeof ButtonBase>>(
+  ({children, ...props}, forwardedRef) => {
+    const {icon: Icon, caret, size} = props
+    let iconOnly = false
+    if (Icon && !children) {
+      iconOnly = true
+    }
+    const iconWrapStyles = {
+      display: 'inline-block',
```
langermank, line 187: The CSS would need some adjusting with my recommendation for icon handling. I could see CSS grid being used here to to handle slots, similar to what I'm doing with [Action List](https://github.com/primer/css/pull/1657/files#diff-1012f63983e65b582649d97a326609627c08f3ae717a127979884e3955a23151R291). I'm using grid-area to create slots, and allowing children to fill them- and if a child exists, `margin-right`. If you want to explore grid together I'm happy to pair on it!

langermank (commented):
This is awesome! I would love to be able to play with this in the React storybook- perhaps @siddharthkp could help setup deploy on PR for React land?!

siddharthkp:
@langermank There is a storybook deployment, it's just not linked as a deployment status :) https://primer-components-git-refs-headspknew-button-primer.vercel.app/react/storybook?path=/story/composite-components-new-button--default-button

langermank:
> @langermank There is a [storybook deployment](https://primer-components-git-refs-headspknew-button-primer.vercel.app/react/storybook?path=/story/composite-components-new-button--default-button), it's just not linked as a deployment status :)

Ahh thank you!

pksjce:
> Hello, I'm new and probably missed it, but I was wondering if there was a documented reason we are moving from separate components for button variants to `variant` prop?

@theinterned - I thought there was general consensus on this. I asked in #primer-react slack channel a while back and essentially got the idea that button being one of the first components made, it didnt get sophisticated handling in the component.

src/stories/NewButton.stories.tsx
pksjce: TBH, I went for this API because thats what I was able to type correctly. You are right though, we don't want consumers to provide styles. We want to add a few size and a11y params to the icon though. Like suggested by @langermank in the comment below.

src/stories/NewButton.stories.tsx
pksjce: So you are saying that I don't need a `Button.Counter` component and can simply rely on the existence of a `count` prop. I don't mind that tbh. The counter is created and styled specifically to the button component and I'm not looking to reuse it anywhere.

src/NewButton/button.tsx
pksjce: in the button audit anatomy here, there was no trailing icon specifically. https://github.com/github/primer/issues/253#issuecomment-949215306

Also there are special margin recommendations for the caret icon which wouldn't necessarily apply for a `trailingIcon`. Which is why I opted for this API.

src/NewButton/button.tsx
pksjce: Would love to pair on this!

theinterned:
@pksjce thank you for that recap! I suppose that Slack convo happened before my time. I guess I was just wondering about if there was general guidance about when to use a variant prop vs create a separate component? Looking through the docs of the various Primer platforms, it looks like there may be a strong preference for variants as props vs variants as components — I wonder if we have the same answer to this across all versions of primer?

On a related note: I notice that PVC uses [`scheme = primary | danger | outline | invisible | link`](https://primer.style/view-components/components/button#schemes) and [`variant = small | medium | large`](https://primer.style/view-components/components/button#variants) 

In contrast, this PR introduces [`variant = default | primary | invisible | danger`](https://github.com/primer/react/pull/1532/files#diff-323861e2e3e75fd205b9e764d36b3e2c8205503296dac08ec5c0903cff2f0851R27) and [`size =  small | medium | large`](https://github.com/primer/react/pull/1532/files#diff-323861e2e3e75fd205b9e764d36b3e2c8205503296dac08ec5c0903cff2f0851R32) to PRC.

... and looking a littler further, it looks like the Figma components call these [`type = Default | Outline | | Primary | Danger | Invisible`](https://primer.style/design/ui-patterns/button-usage#button-types) and [`Size = Small | Medium | Large`](https://primer.style/design/ui-patterns/button-usage#variation-options) ... although I am not sure if these variations are expressed as separate Figma components or as properties.

I wonder if it's possible to align these concepts across platforms?

src/NewButton/button.tsx
langermank: Thanks for linking that audit! The anatomy graphic Ash provided makes a lot of sense. I still think we could make this more generic, and maybe borrow some concepts from ActionList. Here's how I see Button based on that audit:

image

I'm also new to the team so I don't know if we have guidelines around this, but I've had success in the past by separating concerns. IMO `Button` shouldn't handle `Dropdown` specific controls- it should provide the tools that `Dropdown` needs to be a Dropdown (a slot for a caret, anything else necessary). 

Now that I've seen the audit, `counter` could be a `trailingVisual` or an action if its interactive in some way (I don't think it is?). I'm not sure how that works with with `Button.Counter` format, I'm not super familiar with dot notation. 

For the spacing comment: I think the audit is pointing out spacing inconsistencies (using whitespace instead of actual layout properties) and that could be handled in `trailingAction` even if its not "caret" specific.

@ashygee any thoughts here?

ashygee: @langermank I think that we could make the icon slot align to the `leadingVisual` naming convention for consistency. I can update that in my audit as it does make sense in the off chance we decide to allow for other leading visual accessories outside of octicons. The same for `label` too. For `trailingVisual` I could see us being more generic here but I would want to be explicit about what we would want to allow, at this point I think it would just be a counter or an octicon. 

I am hesitant about the `trailingAction` since that is primarily for dropdown carets only, I did not come across any additional actions in my original audit and can't imagine at this point what else we may want to pass.

@pksjce in my original audit I was just pointing out the different elements in the anatomy and not clearly recommending that those be the slot names so I apologize if that misled you. I can update the imagery with @langermank 's recommendation for clarity.

langermank: Thanks @ashygee! There's always a fine line between building what you need and what you might need in the future 😆 I think for the caret dropdown, we can be both "future-proof" (ish) and refrain from building too much logic without a use case by handling that second visual/action slot as such (keep the naming generic) while _also_ shipping Button with the caret functionality IN that slot. So essentially a private API, but it will allow for us to expand upon it later should a new use case come up. For example...

```
<Button>
{trailingAction && (<Caret />)}
</Button>

// which could eventually become

<Button>
{trailingAction && dropdownMenu ? (<Caret />) : trailingActionSlot}
</Button>
```

So offer `trailingAction` with a default for now, but allowing for flexibility later. Another idea would be to leave it completely open now, but treat it as a private API and only use it in other Primer components- basically just don't document that its an available slot.

```
// DropdownMenu

const DropdownMenu = (
<Button trailingAction="caret" />
<Menu>
)
```

mperrotti:
@langermank - we deploy storybook in the preview deployments too. This is the latest: https://primer-components-7ffewgcr6-primer.vercel.app/react/storybook

src/stories/NewButton.stories.tsx
mperrotti: @pksjce - we could just wrap the SVG in a container that has a width and height. Then, if somebody tries to pass an icon size that's too large, it would overflow the container and look obviously broken. If we want to make sure the icon never overflows the container, we could even do something like `.buttonContainer > svg { max-width: 100%; }`

src/stories/NewButton.stories.tsx
mperrotti: I agree that a trailing visual slot would make more sense. If I'm understanding @langermank correctly, that would look something like this:

```jsx
<Button>
    Watch
    <Button.TrailingVisual>
        <CounterLabel>42</CounterLabel>
    </Button.TrailingVisual>
</Button>
```

src/NewButton/button.tsx
```diff
@@ -0,0 +1,220 @@
...
+  size?: 'small' | 'medium' | 'large'
+  icon?: React.FunctionComponent<IconProps>
+  as?: 'button' | 'a' | 'summary' | 'input' | string | React.ReactType
+} & SxProp &
+  FontSizeProps
+
+const getVariantStyles = (theme: Theme, variant: VariantType = 'default') => {
+  const style = {
+    default: `
+      color: ${get('colors.btn.text')({theme})};
```
mperrotti, line 41: Instead of calling `get('my.primitive.path')({theme})`, we could use `css` from `styled-components`.

Example:
```

const style = {
    default: css`
        color: ${get('colors.btn.text')};
    `
}
```

src/stories/NewButton.stories.tsx
langermank: I think that could work! I haven't really worked with that format before (I think its called dot notation) so I'm not sure how the end user would feel about it. I've used props for children in the past, like this:

```
<Button trailingVisual={<CounterLabel>42</CounterLabel>} />
```

So the consumer can use the component props (which are easy to find with typeahead or hover, I think) rather than having to know to use `<Button.TrailingVisual>` etc. But maybe that isn't a format we use in Primer, or we're moving away from it (I'm lacking context). I think either way works so long as we're consistent with other React components

src/NewButton/button.tsx
mperrotti: I like what @siddharthkp did with his composable ActionList API. I'd recommend something like...

1. Specify `Button.Count` as it's own slot
```jsx
<Button>
    <Button.LeadingVisual><EyeIcon /><Button.LeadingVisual>
    Watch
    <Button.Count>42</Button.Count>
    <Button.TrailingVisual><CaretIcon /></Button.TrailingVisual>
</Button>
```

2. style `Button.TrailingVisual` with `display: flex` and allow multiple children:
```jsx
<Button>
    <Button.LeadingVisual><EyeIcon /><Button.LeadingVisual>
    Watch
    <Button.TrailingVisual>
        <CaretIcon />
        <CountLabel>42</CountLabel>
    </Button.TrailingVisual>
</Button>
```

src/NewButton/counter.tsx
```diff
@@ -0,0 +1,22 @@
...
+  display: 'inline-block',
+  padding: '0px 6px',
+  backgroundColor: 'neutral.muted',
+  borderRadius: '50%',
+  marginLeft: '8px',
+  marginRight: '-4px'
+}
+
+const Counter = ({count}: CounterProps) => {
+  // get this to announce new count. How? Change aria property for every render? or aria-live?
```
mperrotti, line 18: `aria-live` feels right to me, but we should check with the accessibility team

src/NewButton/counter.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return <Box sx={styles}>{count}</Box>
```
mperrotti, line 19: We already have a component for counters, so we could probably use that.

src/NewButton/counter.tsx
```diff
@@ -0,0 +1,22 @@
...
+  count: number
+}
+
+const styles = {
+  display: 'inline-block',
+  padding: '0px 6px',
+  backgroundColor: 'neutral.muted',
+  borderRadius: '50%',
+  marginLeft: '8px',
+  marginRight: '-4px'
```
mperrotti, line 14: This negative margin might create an issue if we have a count followed by an icon

src/constants.ts
```diff
@@ -5,7 +5,9 @@ import theme from './theme'
 
 const {get: getKey, compose, system} = styledSystem
 
-export const get = (key: string) => themeGet(key, getKey(theme, key))
+export const get = (key: string) => {
```
mperrotti, line 8: I don't think this change was intentionally included

mperrotti (commented):
@pksjce - let me know if you want to pair on the Typescript errors.

I'm more of a night person, so I'm sure we can find some overlapping time :)

mperrotti:
Disregard all of my feedback about switching from a props API (e.g.: `<Button leadingVisual={EyeIcon}>`) to children (e.g.: `<Button><Button.LeadingVisual><EyeIcon />Watch</Button>`).

I hadn't realized @pksjce had talked to other folks about this and decided that a props API was a better way to go.

src/stories/NewButton.stories.tsx
pksjce: Convo with @ashygee that icon size won't change between buttons, so we don't need to accept that prop. 
The a11y strategy is to hide icons since they are only a visual artifact. My discussion with @langermank is to mandate button having associated text always. Though I'm still exploring how this will work in iconOnly scenarios. 
Will we allow consumers to change other icon properties? like color and such? I don't think we need to do that. Everything should be globally themeable and not locally customisable right?

src/NewButton/button.tsx
pksjce: This was a great session @langermank. you are such a wizard with css and esp grid!

pksjce:
@theinterned 
> I wonder if it's possible to align these concepts across platforms?

So about `variant` being used as a prop here, we can't use `type` because that has semantic meaning for html elements. And `size` is a better name for a property that changes sizes. So we settled on `variant`. Its also used in a couple of other components in PRC like `Flash` and `Header` (not too widely used though)
I remember reading somewhere that PVC uses the prop `scheme` due to some kind of internal limitation. I'll try to find more on that.
It would be really good if the Figma components used `variant` instead of `type`. cc - @ashygee

src/NewButton/button.tsx
langermank: @pksjce I was thinking more on this after our pairing session! I love how you're thinking about the public API and working towards keeping it simple. I can see how `iconOnly` bool is a bit clunky.. but I think in all Button cases, there will always need to be some kind of check like this. We'll just have to decide which one makes the most sense for the end user! For example...

`iconOnly`:
`label`: button description text, required in all variants
`leadingVisual`: pass in an icon (and maybe this should be `leadingIcon` since we don't have a use case for other types of visuals in buttons)
`iconOnly`: consumer passes `iconOnly` prop, we hide `label` programmatically

`ariaLabel`:
`ariaLabel`: button description text
`leadingVisual`: pass in an icon
In this example, instead of always requiring one label and hiding it programmatically, we're asking the consumer to provide an aria-label if its an icon button. But how could we require that without knowing its iconOnly? Further, we shouldn't rely on aria-labels if there are more solid methods available for labeling. We could ask for `ariaLabel` and then treat it exactly like `label` above with visually hidden, but that seems like an extra step for little benefit (abstracting for what reason?)

The `iconOnly` prop starts to feel like `<IconButton>` doesn't it? Which is exactly what you're addressing in this PR 😅 what if instead of `iconOnly` we call the prop `iconButton` so it _feels_ like a different component almost, but is still a prop
`<Button iconButton icon={}` maybe introduce a simple `icon` prop that if used instead of `leadingIcon` or `trailingIcon` indicates that its an iconButton?? Either way, always a prop.. always a decision placed on the consumer.

Yay buttons!

src/NewButton/button.tsx
langermank: > style Button.TrailingVisual with display: flex and allow multiple children

I love this idea! Its just flexible enough, and we can still control the layout within the slot 😍

theinterned:
@pksjce cool! Thank you for the great explanation. 

I am pretty used to seeing `variant` used in design systems for this purpose (along with a `size` prop) so I definitely aligned with that choice! It would be nice if Figma and PVC could align on that naming as well if possible. I'm sure it is common for devs to switch between PRC and PVC and this kind of naming inconsistency would be confusing.

I'm definitely curious about the `scheme` limitation in view components if you can find it!

src/NewButton/button.tsx
pksjce: @langermank - I was thinking along the same lines!
The three options you gave are what I also found as I hunted around for what existing design systems do.
Here's a small decision matrix for this.

No | Solution | Pros | Cons
---|---|---|---
1 | hasIconOnly prop | Clear indication of icon only | no indication of if an icon is actually given, no indication of if the button has text description
2 | ariaLabel/allyTitle | Makes sure we always have a title | We only really need the prop if an icon only button is being rendered. useless in other cases
3 | icon/leadingIcon/trailingIcon props | rest of the Button props remain same including text description, `icon` is a nice indicator of it being iconOnly and still get hold of the actual icon component | three icon related props, what happens when all three are used

I love solution 3. It makes sure we are not asking for unnecessary information from the user. It serves as a good standard. It helps our mission of the button always having description. Its a good enough indication for us to internally style the button as a square and add any extra ally properties. I will refactor the implementation to reflect this.

src/NewButton/button.tsx
pksjce: thanks for this tip! worked perfectly!

src/NewButton/button.tsx
langermank: I think you nailed it! Solution 3 sounds great to me. For prop collisions, I wonder if we can safeguard against that with typescript somehow.. like `if {icon} remove {leadingIcon} & {trailingIcon} from props` or maybe utilize linting with helpful messaging "did you mean `leadingIcon`?"

pksjce: 🤔 I'm not sure if typescript can do something like that TBH. Technically they'd be able to use all of those props, but I might lean on making it look broken if they do it that way. linting may be possible. I will look at what I can come up with!

ashygee:
> I wonder if it's possible to align these concepts across platforms? Also sorry if this has already been covered or there's a plan for it again, newb here!

@theinterned Hi! This work is being covered in https://github.com/github/primer/issues/220 and specifically in https://github.com/github/primer/issues/253.

src/NewButton/button.tsx
```diff
@@ -0,0 +1,215 @@
...
+      <ButtonBase ref={forwardedRef} {...props} iconOnly={iconOnly}>
+        {LeadingIcon && (
+          <Box as='span' data-component='leadingIcon' sx={iconWrapStyles} aria-hidden={!iconOnly}>
+            <LeadingIcon />
+          </Box>
+        )}
+        {Icon &&  <Box as='span' sx={{display:'inline-block'}} aria-hidden={!iconOnly}>
+            <Icon />
+          </Box>}
+        <span data-component='text' hidden={Icon? true:false}>{children}</span>
```
langermank, line 195: I think `hidden` shouldn't be in the dom if its false, so would `hidden={Icon}` work here?
Also provide an `id` here for the `button`

output something like:
```
<button aria-labelledby="label-id">
<span id="label-id" hidden>label</span>
</button>
```

siddharthkp: I think `hidden={Icon ? true: undefined}` would achieve it that

pksjce:
## Ready for Review 🎉

The base for new button component is ready for review and merge. There are pending tasks which I will tend to in separate PRs.

Things to do

- [ ] Add documentation for new button
- [ ] Add Button Link component
- [ ] Complete Accessibility audit concerns

src/unreleased.ts
```diff
@@ -7,3 +7,4 @@
 
 // Components
 export * from './ActionList2'
+export * from './NewButton'
```
siddharthkp, line 10: This wouldn't be able to grab the `export default Button`, only `ButtonProps` :(

siddharthkp (approved):
Reviewed the API, I like it a lot!

The only place where I wasn't super confident is the counter. I don't have the whole context for the decision so I trust your judgement. Happy to pair if you're still iterating on it.

src/unreleased.ts
pksjce: Ah! Haven't tested this properly yet.

pksjce:
@siddharthkp - Thanks! I changed the Counter approach mostly because
1. The counter needs to respond to different variants and sizes.
2. I don't want to change the existing counter.
3. I can put counter styles in `button.tsx`. 

What do you think?

I would love some pairing on the sxProps. I'm struggling with typescript issues around that and right now button is not able to accept any custom styles aka Block button. But I'll have this in a followup PR. I think I'm happy to merge what I have.

siddharthkp:
> I would love some pairing on the sxProps. I'm struggling with typescript issues around that

I'm up for it. Not an expert at it, but will give it an honest try!

PR #1553, opened 2021-11-02
"fixes" a tsc error. Adds ts-toolbelt

jfuchs:
This was broken in main. See also https://github.com/primer/react/pull/1551. It's not a great fix, but after a few hours I decided to just `ts-ignore` this type I believe to be working correctly (and we've validated the output).

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/__tests__/KeyPaths.types.test.ts
+++ b/src/__tests__/KeyPaths.types.test.ts
@@ -1,3 +1,4 @@
+import {Union} from 'ts-toolbelt'
 import {KeyPaths} from '../utils/types/KeyPaths'
 
 type NestedObject = {
@@ -8,6 +9,6 @@ type NestedObject = {
   }
 }
 
-export function generatesKeyPathsFromObject(x: KeyPaths<NestedObject>): 'a' | 'b.b1' | 'b.b2' {
+export function generatesKeyPathsFromObject(x: Union.Diff<KeyPaths<NestedObject>, 'a' | 'b.b1' | 'b.b2'>): never {
   return x
 }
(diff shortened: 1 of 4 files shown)


src/__tests__/KeyPaths.types.test.ts
```diff
@@ -8,6 +9,6 @@ type NestedObject = {
   }
 }
 
-export function generatesKeyPathsFromObject(x: KeyPaths<NestedObject>): 'a' | 'b.b1' | 'b.b2' {
+export function generatesKeyPathsFromObject(x: Union.Diff<KeyPaths<NestedObject>, 'a' | 'b.b1' | 'b.b2'>): never {
```
colebemis, line 12: If I understand this correctly, we're asserting that the difference between the `KeyPaths<NestedObject>` and `'a' | 'b.b1' | 'b.b2'` is nothing. Seems reasonable to me 👍

src/utils/types/KeyPaths.ts
```diff
@@ -1,4 +1,10 @@
 // Produces a union of dot-delimited keypaths to the string values in a nested object:
 export type KeyPaths<O extends Record<string, unknown>> = {
-  [K in keyof O]: K extends string ? (O[K] extends string ? `${K}` : `${K}.${KeyPaths<O[K]>}`) : never
+  [K in keyof O]: K extends string
+    ? O[K] extends string
+      ? `${K}`
+      : // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore TypeScript has bested me, but the KeyPaths type is tested.
```
colebemis, line 7: I think the type test gives us enough confidence that this does what we want 👍

src/__tests__/KeyPaths.types.test.ts
jfuchs: that's right! Before, I was only testing that `KeyPaths<NestedObject>` extends `'a' | 'b.b1' | 'b.b2'`. Without ts-toolbelt I could've added a second test function with swapped argument/return types.

PR #1580, opened 2021-11-04
Split CI workflow into two

siddharthkp:
How do y'all feel about splitting the CI test into 2 - so that we can still run Lint + Test independent if the build succeeds of not (looking specifically at types)

ci.yml
- name: Install dependencies
- name: Build
- name: Lint → test.yml
- name: Test  → test.yml

The increase in total time is nothing in this PR because there were no dependency changes, that's the step that gets repeated in both

cc @jfuchs @colebemis


.github/workflows/ci.yml
```diff
@@ -21,11 +21,5 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Lint
-        run: npm run lint
-
       - name: Build
```
colebemis, line 24: Should we call this workflow `build.yml`?

colebemis (approved):

.github/workflows/ci.yml
siddharthkp: Done!

PR #1581, opened 2021-11-04
[do not merge]Proof-of-concept for Label component API changes

mperrotti:
This proof-of-concept came as a side effect of trying to explain how a "label" is different from a "token"

See the `NewLabel` and `NewStateLabel` components and their Storybook stories


src/NewLabel/NewLabel.tsx
```diff
@@ -0,0 +1,168 @@
...
+  | 'accent'
+  | 'success'
+  | 'attention'
+  | 'severe'
+  | 'danger'
+  | 'done'
+  | 'sponsors'
+
+interface Props {
+  color?: LabelColorOptions
```
mperrotti, line 19: Help wanted: `color` is not a good name for this prop. Something like `variant` or `appearance` feels better, but I don't like either of those either.

Any suggestions for this prop name?

PR #1583, opened 2021-11-08
[WIP]Use Isomorphic effect

### LINKED ISSUE #1453: Overlay causes `useLayoutEffect` warnings in SSR mode
**Describe the bug**

When using the `<Overlay />` and components that build upon it in SSR mode, React emits warnings about calling `useLayoutEffect` during SSR.

Of course, there are legitimate uses of `useLayoutEffect` in Primer and cases where it may be unavoidable. However, in `<Overlay />`, it looks like this particular call is not doing any DOM measurement or direct manipulation, but is kicking off an animation. This _seems_ like it's probably fine to defer until after a paint occurs, and I'm guessing that browser is probably going to do that anyway before this animation begins.

Given that, I wonder if we can move this to a `useEffect` call, instead or alternatively investigate using a `useIsomorphicLayoutEffect`, which @mattcosta7 reminded me about in Slack:

```ts
const useIsomorphicLayoutEffect = typeof document !== 'undefined' ? useLayoutEffect : useEffect
```

(That said, in this case, I think just using the simpler `useEffect` is fine.)

In order to avoid these warnings currently, I have to defer completely the rendering of overlays until we're on the client.

**To Reproduce**
1. Create a Next.js app
2. Render an `<Overlay />` on some page
3. Watch logs

pksjce:
Describe your changes here.

In https://github.com/primer/react/issues/1453, we see that `useLayoutEffect` causes SSR warnings. This is the nature of the function. One proven way of mitigating this is to use `useLayoutEffect` in the client mode and `useEffect` when called in the SSR mode.
Reference - https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a

Closes #1453 

### Screenshots

No changes

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp (approved):

siddharthkp:
Good fix! LGTM :)

PR #1588, opened 2021-11-09
Separate typechecking and building type declarations

jfuchs:
I believe because in a previous PR I've excluded stories from our built type definitions, we aren't typechecking stories. This PR separates typechecking _everything_ from building type definitions for _only_ the exported parts of primer.

I'm open to other ways to accomplish the same thing!

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.github/workflows/ci.yml
```diff
@@ -24,8 +24,11 @@ jobs:
       - name: Lint
         run: npm run lint
 
-      - name: Build
-        run: npm run build
+      - name: Typecheck
```
colebemis, line 27: Is type check two words? 🤷 
```suggestion
      - name: Type check
```

colebemis (approved):
Nice catch 👍

.github/workflows/ci.yml
jfuchs: It looks like the TypeScript docs show both, bit I think separate words might be a bit more prominent. I'll take you change!

PR #1592, opened 2021-11-10
Fix: TextInput styling regression with icon and block prop usage

rezrah:
### Summary

The `TextInput` component suffers from a visual regression when you use **both** `icon` and `block` props together.

This PR fixes the regression, while attempting to ensure that no further regressions are introduced.

### Changes

- [x] Fixed highlighted issue
- [x] Backfilled a missing Story for `TextInput`
 
### Screenshots and 🎥 

**Before:** 

https://user-images.githubusercontent.com/13340707/141128860-4ae6d25a-5ebd-43da-b93f-6d86f139f6c8.mov

**After:** 
Screenshot 2021-11-10 at 14 11 25

https://user-images.githubusercontent.com/13340707/141127745-2f5f1b54-966c-4eab-be51-5980812a7f00.mov

**New Storybook:**

https://user-images.githubusercontent.com/13340707/141128949-1126d993-22a5-44a5-99a0-3398bc5815e3.mov

### Reviewer notes

- This is implemented as a temporary fix, as the `TextInput` component will undergo `alpha` => `beta` refactoring shortly and the longer-term solution should perhaps, come from a `css` classname instead (I'm unsure if it already exists)

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/TextInput.tsx
+++ b/src/TextInput.tsx
@@ -37,6 +37,9 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
     // this class is necessary to style FilterSearch, plz no touchy!
     const wrapperClasses = classnames(className, 'TextInput-wrapper')
 
+    // temporarily fixes an issue using icon and block in tandem
+    const customStyles = IconComponent && block && {...sxProp, ...{display: 'flex'}}
+
     return (
       <TextInputWrapper
         block={block}
@@ -44,7 +47,7 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
         contrast={contrast}
         disabled={disabled}
         hasIcon={!!IconComponent}
-        sx={sxProp}
+        sx={customStyles || sxProp}
         theme={theme}
         width={widthProp}
         minWidth={minWidthProp}
--- a/src/TextInput.tsx
+++ b/src/TextInput.tsx
@@ -37,9 +37,6 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
     // this class is necessary to style FilterSearch, plz no touchy!
     const wrapperClasses = classnames(className, 'TextInput-wrapper')
 
-    // temporarily fixes an issue using icon and block in tandem
-    const customStyles = IconComponent && block && {...sxProp, ...{display: 'flex'}}
-
     return (
       <TextInputWrapper
         block={block}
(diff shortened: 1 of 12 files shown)


src/TextInput.tsx
```diff
@@ -37,14 +37,17 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
...
+
     return (
       <TextInputWrapper
         block={block}
         className={wrapperClasses}
         contrast={contrast}
         disabled={disabled}
         hasIcon={!!IconComponent}
-        sx={sxProp}
+        sx={customStyles || sxProp}
```
siddharthkp, line 50: nitpick for readability

```js
 sx={{
  display: IconComponent && block ? 'flex' : 'inline-block',
  ...sxProp
}}
```

src/stories/TextInput.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </div>
+        <div className="form-group-body">
+          <TextInput id={inputId} value={value} onChange={handleChange} {...args} />
+        </div>
+      </div>
+    </form>
+  )
+}
+
+export const WithLeadingIcon = (args: TextInputProps) => {
```
siddharthkp, line 79: Great addition!

siddharthkp (approved):

src/TextInput.tsx
rezrah: Thanks @siddharthkp! Yes, was thinking along these lines initially too. I found that approach however, would force `inline-block` as a default style, overriding the `TextArea` css class styling (making this component more opinionated than it needed to be). 

The approach I had taken here only applies the override styling if the right conditions are met, leaving the default styling intact and reducing chance of regressions in future. At least that was the thinking 😅 .

siddharthkp: Ah I didn't realise there is already a display up the chain. Good thinking!

I'd also try this approach to set a value only when there is a value:

```
display: IconComponent && block ? 'flex' : undefined
```

rezrah: 👍 ... In the end, there was a much simpler solution 😅. Would you mind reviewing the changes again please @siddharthkp?

siddharthkp (approved):
That's very smart!

colebemis (approved):

PR #1601, opened 2021-11-11
ActionList: Use icon instead of input for multiple selection

siddharthkp:
Part of https://github.com/github/primer/issues/447
Fixes half of https://github.com/github/primer/issues/228

Nested interactive controls are not announced by screen readers, the accessibility advice here is not use interactive elements for presentation.

- [x] Replace readonly `input type="checkbox"` with svg from [primer/css](https://primer.style/css/components/action-list#leading-action-multi-select)
- [x] Bonus: checkmark consistency across browsers :)

Future - The svg used here has a short life and will get replaced with the [Checklist component](https://github.com/primer/react/pull/1606) when it's ready

### Screenshots

look pretty ActionList:
image

also shared in AutoComplete:

Screenshot 2021-11-12 at 11 15 34 AM

&nbsp;

Accessibility violations:

Before: 

image

After: 

image

### Merge checklist

- NA Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/action-list-selection-a11y.md
+++ b/.changeset/action-list-selection-a11y.md
@@ -1,0 +1,5 @@
+---
+'@primer/components': patch
+---
+
+ActionList: Use icon instead of input for multiple selection in ActionList
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -333,8 +333,16 @@ const DescriptionContainer = styled.span`
   flex-basis: var(--description-container-flex-basis);
 `
 
-const MultiSelectInput = styled.input`
-  pointer-events: none;
+const MultiSelectIcon = styled.svg<{selected?: boolean}>`
+  rect {
+    fill: ${({selected}) => (selected ? get('colors.accent.fg') : get('colors.canvas.default'))};
+    stroke: ${({selected}) => (selected ? get('colors.accent.fg') : get('colors.border.default'))};
+  }
+  path {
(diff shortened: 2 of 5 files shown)


pksjce:
The solution is so interesting! I've never heard of someone using an svg instead of a checkbox for a11y reasons.  Why not use the solution proposed in https://github.com/github/primer/issues/228 where they recommended to make the role `presentational` ? Is that for checkbox browser consistency?

siddharthkp:
> The solution is so interesting!

credit to @langermank for introducing it in primer/css

> Why not use the solution proposed in github/primer#228 where they recommended to make the role presentational ? 

We are? 😅 The [example mentioned](https://www.w3.org/TR/wai-aria-1.2/#example-4) in the issue uses an image with role="presentation"

langermank:
> From what I understand, having an interactive element inside a selectable listbox option or menuitem isn't valid. [See violations on story](https://primer.style/react/storybook?path=/story/composite-components-actionlist2--multi-select-list-story)

Exactly. `input` needs to be in a `form` and ActionList isn't really a `form` scenario. I attended a popup a11y meeting last week and was able to confirm this with a screen reader user, which was super helpful!

mperrotti (commented):
LGTM, and it doesn't break the multiselect behavior in Autocomplete 🙌

mperrotti (approved):

PR #1607, opened 2021-11-15
ActionList: Add comment to avoid import confusion

siddharthkp:
In the docs, because we have multiple ActionLists now, we import it ActionList2 from a different scope:

```jsx
const {ActionList} = unreleased

render(
  <ActionList>
    ...
  </ActionList>
)
```

This causes confusion in where does `unreleased` come from?

The long term fix would be to modify the scope conditionally by tagging the code block with "unreleased": ` ```javascript live unreleased `.

For now, the bandaid is the add a comment to ignore the internal silliness and import from the path

```jsx
// import {ActionList} from '@primer/components/unreleased'
const {ActionList} = unreleased // ignore this, import like that ↑
```

We mention this import path at the [start of the page](https://primer-components-fasv5eaxn-primer.vercel.app/react/ActionList2) as well

### Screenshots

image

### Merge checklist

- NA Added/updated tests
- NA Tested in Chrome
- NA Tested in Firefox
- NA Tested in Safari
- NA Tested in Edge

### DIFF
--- a/docs/content/ActionList2.mdx
+++ b/docs/content/ActionList2.mdx
@@ -58,7 +58,8 @@ import {ActionList} from '@primer/components/unreleased'
 ## Minimal example
 
 ```javascript live noinline
-const {ActionList} = unreleased
+// import {ActionList} from '@primer/components/unreleased'
+const {ActionList} = unreleased // ignore this, import like that ↑
 
 render(
   <ActionList>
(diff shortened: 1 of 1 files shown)


siddharthkp:
Closed in favor of https://github.com/primer/react/pull/1609

PR #1609, opened 2021-11-15
Rename unreleased to drafts to avoid confusion

siddharthkp:
Rename `@primer/components/unreleased` to `@primer/components/drafts`

-- 

Picked a name that doesn't conflict with component maturity status - experimental/alpha/beta/stable 
& doesn't conflict with package publishing language - published/released/next/apha/beta

Thanks to @pksjce for the suggestion!

--

+ In the docs, because we have multiple ActionLists now, we import it ActionList2 from a different scope:

```jsx
const {ActionList} = drafts

render(
  <ActionList>
    ...
  </ActionList>
)
```

This causes confusion in where does `drafts` come from?

The long term fix would be to modify the scope conditionally by tagging the code block with "drafts": ` ```javascript live drafts `.

For now, the bandaid is the add a comment to ignore the internal silliness and import from the path

```jsx
// import {ActionList} from '@primer/components/drafts'
const {ActionList} = drafts // ignore docs silliness, import like that ↑
```

### DIFF
--- a/.changeset/young-ties-fix.md
+++ b/.changeset/young-ties-fix.md
@@ -1,0 +1,5 @@
+---
+"@primer/components": patch
+---
+
+Rename `@primer/components/unreleased` to `@primer/components/drafts` to avoid confusion when referring to it.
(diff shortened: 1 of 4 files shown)


siddharthkp:
@pksjce @jclem FYI!

colebemis (approved):
:shipit:

PR #1611, opened 2021-11-16
Input field components - text, checkbox, radio

### LINKED ISSUE #480: Dialog component
For our internal project, we will require a modal. I have to build it anyway, so I figured I'd kick off a discussion about upstreaming it to `@primer/components`!

Here is the API I was thinking about using:

```JSX
<Dialog title="String" isOpen={true} onDismiss={() => {}}>
  <Text>Content</Text>
</Dialog>
```

This would render a dialog like this as a React portal:

Screenshot 2019-06-24 at 09 56 46

Clicking either the X Octicon or the background overlay calls `onDismiss`, which should close the dialog.

Initial questions:

- Should the `title` prop be able to accept any React node, do we render e.g. icons in dialog titles anywhere?
- I specifically avoided adding padding automatically to the contents, as that would make it hard to e.g. render an alert in the dialog like we do in the "Delete repo" dialog (reference below). To make sure users set the spacing correctly, we could add a `DialogBody` component that does nothing but sets `padding: 16px` around its children?

  Screenshot 2019-06-24 at 09 58 21

- I would like to use [@reach/dialog](https://ui.reach.tech/dialog/) to make sure we get the a11y correct. Would that be alright?

mperrotti:
This component renders input along with its label, validation content, and hint content. It will correctly associate labels and other related content with the input.

### Screenshots
Value input
Screen Shot 2021-11-15 at 8 04 37 PM

Toggle (radio/checkbox) input
Screen Shot 2021-11-15 at 8 10 53 PM

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

Closes [Primer issue #480](https://github.com/github/primer/issues/480)


src/utils/use-force-update.tsx
```diff
@@ -0,0 +1,7 @@
+// Inspired from reach-ui: https://github.com/reach/reach-ui/blob/develop/packages/utils/src/use-force-update.ts
```
mperrotti, line 1: This wasn't meant to be added in this PR. Will remove

src/InputLabel.tsx
```diff
@@ -0,0 +1,49 @@
...
+    // TODO: fix typescript errors
+    <Box
+      fontWeight="bold"
+      fontSize={1}
+      as="label"
+      display="block"
+      color={disabled ? 'fg.muted' : 'fg.default'}
+      sx={
+        // TODO: create a VisuallyHidden component if this hiding technique is used in any other components
+        visuallyHidden
```
langermank, line 21: Love this! ❤️ 🎉

mperrotti:
I still need to write tests

mperrotti:
[helpwanted]

This component is similar to [FormGroup](https://primer.style/react/FormGroup). I considered just updating `FormGroup` instead of creating a new component, but there are two problems with that:
- It would cause a breaking change
- I think the name "FormGroup" is confusing. It sounds like a group of multiple fields in a form (like a fieldset), not a group of elements for a single form field

siddharthkp:
API question, I see each input component has it's own *.Label child in the style of compound components

As we want labels, descriptions, validation messages to be consistent across all these components, how do you feel about pulling them up on a `Form` component which can compose all the input fields we provide

Something like: 
```jsx
<Form>
  <Form.Field>
    <Form.Label>Option 1</Form.Label>
    <TextInput />
    <Form.Description>this line explains what will happen if you press this checkbox</Form.Label>
   <Form.Validation status="error">oh no no, please don't do that</Form.Validation>
  </Form.Field>
  <Form.Field>
    <Form.Label>Option 1</Form.Label>
    <Checkbox />
    <Form.Description>this line explains what will happen if you press this checkbox</Form.Label>
   <Form.Validation status="error">oh no no, please don't do that</Form.Validation>
  </Form.Field>
</Form>
```

mperrotti:
Thanks @siddharthkp. After thinking about your feedback, I'm second-guessing my decision to go with specific components for each kind of input. I actually started by going in a similar direction, but I went with the current approach after weighing the pros and cons of having generic field components.

## Pros of general `Form.Field` components:
- No redundant `.Label`, `.Caption`, `.Validation` components
- Less surface area to introduce styling inconsistencies with components that are shared between different fields (labels, captions, and validation messages)
- Less surface area for inconsistent form field component APIs
- Less surface area for bugs
- Less components to remember

## Cons of general `Form.Field` components:
- If we just pass the input component directly as a child, the parent `Form.Field` component has no way to set the correct attributes on the rendered `<input>` for accessibility
    - The `<input>`'s `id` attribute must match the `<label>`'s `for` attribute
    - The `<input>`'s `aria-describedby` attribute must match the `id` of the text node/s for the input's validation message and/or caption text
    - The `<input>`'s `aria-invalid` attribute needs to be set when the field fails validation
- Every kind of field might not accept the same children. For example: checkbox and radio fields can have a `.LeadingVisual` that is rendered between the input and the label, but text input fields don't have this.
- It felt a little too "magic" for the radio or checkbox field layout to automatically get applied if it has a child where the `type` was equal to primer-react's radio or checkbox component
- It felt strange to have children that didn't follow the `Field.{ChildComponent}` naming pattern
- If we have a generic form field component, it is more likely that we'll introduce breaking changes when/if we only want to make a change to a specific kind of form field
- If we just pass the input component directly as a child, they could pass whatever they want in that spot. This isn't so bad, but it's a small risk for inconsistent or unexpected usage of the component

---

The only cons that are blockers for me are the first two. If other reviewers agree that a more generic field component is a better component API, I'm confident I can come up with ways around those.

src/_InputField/InputField.tsx
```diff
@@ -0,0 +1,116 @@
...
+        validationMessageId
+      }}
+    >
+      {slots => {
+        const isLabelHidden = React.isValidElement(slots.Label) && slots.Label.props.visuallyHidden
+        if (slots.Input && !slots.Label) {
+          // eslint-disable-next-line no-console
+          console.error(
+            `The input field with the id ${id} MUST have a Label child (e.g.: <TextInputField.Label>).\n\nIf you want to hide the label, pass the 'visuallyHidden' prop to the Label component.`
+          )
```
siddharthkp, line 81: Would you like to throw an Error instead? 

```
throw ... raises an exception in the current code block and causes it to exit, or to flow to next catch statement if raised in a try block.

console.error just prints out a red message to the browser developer tools javascript console and does not cause any changes of the execution flow.
```
source: https://stackoverflow.com/a/25377160

I throw an Error in ActionList v2 here: https://github.com/primer/react/blob/main/src/ActionList2/Selection.tsx#L20

siddharthkp:
> If we just pass the input component directly as a child, the parent Form.Field component has no way to set the correct attributes on the rendered <input> for accessibility

Could we implement this with React Context to pass the generated id down? [Sandbox example](https://codesandbox.io/s/modest-visvesvaraya-umhju?file=/src/App.js), this would mean our input components are form context aware, which feels natural as long as they don't break outside a form.

> Every kind of field might not accept the same children. For example: checkbox and radio fields can have a .LeadingVisual that is rendered between the input and the label, but text input fields don't have this.

This is very interesting, would we like to restrict LeadingVisual with the API for some components?

src/_InputField/InputField.tsx
mperrotti: I was thinking that throwing an error might be too aggressive incase an unlabeled input makes it into production somehow.

Maybe we can safely assume that our CI will catch it though

siddharthkp: Fair enough :)

mperrotti:
Thanks again, @siddharthkp !

> Could we implement this with React Context to pass the generated id down? Sandbox example, this would mean our input components are form context aware, which feels natural as long as they don't break outside a form.

Thank you for putting in the effort to make a Sandbox! This feels cleaner than the ideas I had, but there are three minor trade-offs:
- we have two contexts going on in one component: the context from `createSlot`, and the `FieldContext`
- we have to consume context in every Primer input component
- PRC consumers could still pass in a custom input component that wouldn't be checking for context

None of these seem like blockers, but let me know if you or other reviewers think they are.

Just for transparency, here are the flawed ideas I had to get the right attributes on the input:
- Require an input ref be passed to the field component so we can just set attributes with JS
- Use `cloneElement` to clone the input component

---

> would we like to restrict LeadingVisual with the API for some components?

Yes, I'd like to restrict it only to components that are designed to show the LeadingVisual. Any ideas how we might do this?

siddharthkp:
Tradeoffs:

> we have two contexts going on in one component: the context from createSlot, and the FieldContext

Feels okay to me.

> we have to consume context in every Primer input component

I lowkey really like this 😅 because it means you can use the same component. Lower surface area. Forms are the most common use container for Inputs.

> PRC consumers could still pass in a custom input component that wouldn't be checking for context

Yes! This wouldn't work which is bad :( I'm hoping folks would pass their own id/for if they are opting out of system components. 🤔 We can automatically cover this case if we use the cloneElement approach you mentioned.

> Use cloneElement to clone the input component
This was my first implementation as well! What problems did you run into this? I want to say that contents of a field are fairly predictable? Do you expect nesting or composition where it will break?

> Yes, I'd like to restrict it only to components that are designed to show the LeadingVisual. Any ideas how we might do this?

I don't know how to do that with typescript, at least. We can figure out a hacky way of throwing a warning/error in dev mode. Not ideal, but worth the simpler API 🤔

mperrotti:
Re: cloneElement...
I haven't actually tried implementing it yet, but we'd have a similar problem as using context. If people pass a custom component, we wouldn't be able to reliably pass the correct attributes to the input because we wouldn't know how to identify the input component.

---

Re: LeadingVisual...
I'd be happy to pair on hacking something together

siddharthkp:
> Re: LeadingVisual...
> I'd be happy to pair on hacking something together

Let's do it 🙌

docs/content/Checkbox.md
```diff
@@ -2,12 +2,20 @@
...
-source: https://github.com/primer/react/blob/main/src/Checklist.tsx
+source: https://github.com/primer/react/blob/main/src/Checkbox.tsx
 storybook: '/react/storybook?path=/story/forms-checkbox--default'
 ---
 
 import {ComponentChecklist} from '../src/component-checklist'
 
+<Note variant="warning">
+
+**Use [ChoiceInputField](/ChoiceInputField) to render a checkbox with a label and other content related to the field.** This component is only meant to be used in the case that you're building a custom checkbox that is not yet supported by Primer. For example: the "Use your administrator privilages..." checkbox at the bottom of PR that doesn't meet the criteria to be merged without admin privilages.
```
rezrah, line 13: Suggestion on wording...

```suggestion
**Use [ChoiceInputField](/ChoiceInputField) to display an accessible checkbox form field**. This `Checkbox` component is intended only as an ingredient for other custom components, or as a drop-in replacement for native HTML checkboxes outside of form use-cases.

```

docs/content/ChoiceInputField.mdx
```diff
@@ -0,0 +1,155 @@
+---
+title: ChoiceInputField
+status: Alpha
```
rezrah, line 3: Could we add a `description` here. Perhaps move the initial paragraph here...

```
description: Use `ChoiceInputField`  to render a labelled checkbox or radio input, along with any corresponding hint text or validation feedback.
```

docs/content/ChoiceInputField.mdx
```diff
@@ -0,0 +1,155 @@
...
+TODO: add `validationStatus` examples when we have the validation styling done
+-->
+
+import {ComponentChecklist} from '../src/component-checklist'
+import {ChoiceInputField, Checkbox, Radio} from '@primer/react'
+import {MarkGithubIcon} from '@primer/octicons-react'
+
+The `ChoiceInputField` component is used to render a labelled checkbox or radio input and, optionally, associated hint text and graphic.
+
+## Anatomy
```
rezrah, line 18: Feels a bit heavy to go straight into this. Can we merge the descriptions here with the Props at the bottom, to save space and avoid repeating the headings? Anatomy feels more like design guidelines language too.

docs/content/ChoiceInputField.mdx
```diff
@@ -0,0 +1,155 @@
...
+  <ChoiceInputField.Label>Option one</ChoiceInputField.Label>
+  <Checkbox />
+</ChoiceInputField>
+```
+
+### Radio
+
+```jsx live
+<ChoiceInputField>
+  <ChoiceInputField.Label>Option one</ChoiceInputField.Label>
```
rezrah, line 53: Can we add more radios to this example? Just to emphasise that users shouldn't use a single radio ever, but only as part of a set.

docs/content/ChoiceInputField.mdx
```diff
@@ -0,0 +1,155 @@
...
+
+```jsx live
+<ChoiceInputField disabled>
+  <ChoiceInputField.Label>Option one</ChoiceInputField.Label>
+  <Checkbox />
+</ChoiceInputField>
+```
+
+### With a visually hidden label
+
```
rezrah, line 68: Is it worth adding another Note or external link here on when - and when not - to use this prop? I'm not sure myself 😅 . Wouldn't you use `Checkbox` if you need to hide the label?

docs/src/@primer/gatsby-theme-doctocat/nav.yml
```diff
@@ -111,6 +113,8 @@
       url: /Text
     - title: TextInput
       url: /TextInput
+    - title: TextInputField
+      url: /TextInputField
```
rezrah, line 117: Should this be `InputField` now?
Screenshot 2021-12-15 at 15 08 53

docs/content/InputField.mdx
```diff
@@ -0,0 +1,192 @@
...
+source: https://github.com/primer/react/blob/main/src/InputField.tsx
+storybook: '/react/storybook?path=/story/forms-inputfield--text-input-field'
+---
+
+import {InputField, TextInputWithTokens} from '@primer/react'
+import {ComponentChecklist} from '../src/component-checklist'
+
+The `InputField` component is used to render a labelled text input and, optionally, associated validation text and hint text.
+
+## Anatomy
```
rezrah, line 13: Same feedback as https://github.com/primer/react/pull/1611/files#r769697405

docs/content/InputField.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+### InputField
+
+The container that handles the layout and passes the relevant IDs and ARIA attributes it's children.
+
+### InputField.Label
+
+A `InputField.Label` must be passed, but it may be visually hidden.
```
rezrah, line 21: Nit. This doesn't feel clear why we'd want someone to use this, instead of using the base components

docs/content/InputField.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+### InputField.Caption
+
+If this field needs additional context, a `InputField.Caption` may be used to render hint text.
```
rezrah, line 25: Shall we switch to active for these descriptions? E.g. `InputField.Caption` can be used to render hint text for fields that require additional context`

docs/content/InputField.mdx
```diff
@@ -0,0 +1,192 @@
...
+<InputField>
+  <InputField.Label>Name</InputField.Label>
+  <TextInput />
+  <InputField.Caption>Hint: your first name</InputField.Caption>
+</InputField>
+```
+
+### With validation
+
+```javascript live noinline
```
rezrah, line 105: this example displays strangely for me, does it look like this for you @mperrotti? 

Screenshot 2021-12-15 at 15 16 44

docs/content/InputField.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+render(ValidationExample)
+```
+
+## Props
+
+### InputField
+
+| Name             | Type                                                                                                                  |     Default      | Description                                                                                                                                                                                                          |
+| :--------------- | :-------------------------------------------------------------------------------------------------------------------- | :--------------: | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
rezrah, line 147: Are you planning to switch these to tables?

docs/content/InputField.mdx
```diff
@@ -0,0 +1,192 @@
...
+| id               | `string`                                                                                                              | generated string | The unique identifier for this field. Used to associate the label, validation text, and caption text                                                                                                                 |
+| required         | `boolean`                                                                                                             |      false       | Whether this field must have a value for the user to complete their task                                                                                                                                             |
+| validationMap    | `Record<string, 'error'\|'sucess'>`                                                                                   |        -         | A map of validation statuses and their associated validation keys. When one of the validation keys is passed to the `validationResult` prop, the associated validation message will be rendered in the correct style |
```
rezrah, line 152: stray `\` in the union here
Screenshot 2021-12-15 at 15 20 33

docs/content/ChoiceInputField.mdx
mperrotti: I agree - these descriptions can just go under the Props section.

docs/content/ChoiceInputField.mdx
mperrotti: I think another note would be helpful. Every input should have an associated `<label>` for assistive technology, so instead of excluding it we'd prefer visually hiding it.

docs/src/@primer/gatsby-theme-doctocat/nav.yml
mperrotti: Ooops my mistake 🙈

docs/content/InputField.mdx
mperrotti: Agreed. I'll add an explanation.

docs/content/InputField.mdx
mperrotti: Yes. I wrote this before `PropTable` was merged

docs/content/InputField.mdx
mperrotti: It does. Easy fix though :)

src/__tests__/ChoiceInputField.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+      expect(
+        render(
+          <SSRProvider>
+            <ChoiceInputField disabled>
+              <ChoiceInputField.Label visuallyHidden>{CHECKBOXINPUTFIELD_LABEL_TEXT}</ChoiceInputField.Label>
+              <Checkbox />
+            </ChoiceInputField>
+          </SSRProvider>
+        )
+      ).toMatchSnapshot()
```
rezrah, line 73: General observation about snapshot tests: These may guard against regressions to the implementation detail, but not necessarily the actual user behaviour or the underlying html elements. I think tests like these ones should be done with testing library, which would give us greater confidence that the rendered behaviour is covered too.

mperrotti: Agreed. I'll update them.

docs/content/ChoiceInputField.mdx
mperrotti: I think I'm actually going to remove this feature from `ChoiceInputField.Label`. It makes more sense to use the `Checkbox` or `Radio` component directly and provide your own `<label>`.

docs/content/ChoiceInputField.mdx
```diff
@@ -0,0 +1,164 @@
...
+title: ChoiceInputField
+status: Alpha
+description: The ChoiceInputField component is used to render a labelled checkbox or radio input and, optionally, associated hint text and graphic.
+source: https://github.com/primer/react/blob/main/src/ChoiceInputField.tsx
+storybook: '/react/storybook?path=/story/forms-choiceinputfield--checkbox-input-field'
+---
+
+<!--
+TODO: add `validationStatus` examples when we have the validation styling done
+-->
```
rezrah, line 11: Do we need this? ☝️

docs/content/ChoiceInputField.mdx
```diff
@@ -0,0 +1,164 @@
+---
+title: ChoiceInputField
+status: Alpha
+description: The ChoiceInputField component is used to render a labelled checkbox or radio input and, optionally, associated hint text and graphic.
```
rezrah, line 4: Being a bit pedantic (sorry) but feels like there's an unnecessary comma here

```suggestion
description: The ChoiceInputField component is used to render a labelled checkbox or radio inputs with optional  hint text.
```

docs/content/ChoiceInputField.mdx
```diff
@@ -0,0 +1,164 @@
...
+---
+
+<!--
+TODO: add `validationStatus` examples when we have the validation styling done
+-->
+
+import {ChoiceInputField, Checkbox, Radio} from '@primer/react'
+import {MarkGithubIcon} from '@primer/octicons-react'
+import {ComponentChecklist} from '../src/component-checklist'
+import {PropsTable} from '../src/props-table'
```
rezrah, line 16: Some of these imports may not be needed anymore

docs/content/ChoiceInputField.mdx
mperrotti: I guess we could get rid of it

docs/content/ChoiceInputField.mdx
mperrotti: I'll see what I can get rid of. When I excluded the `PropsTable` import, it caused a deployment error in Vercel.

src/__tests__/ChoiceInputField.test.tsx
```diff
@@ -0,0 +1,213 @@
...
+        <ChoiceInputField>
+          <ChoiceInputField.Label>{CHECKBOXINPUTFIELD_LABEL_TEXT}</ChoiceInputField.Label>
+          <Checkbox />
+          <ChoiceInputField.Caption>{CHECKBOXINPUTFIELD_CAPTION_TEXT}</ChoiceInputField.Caption>
+        </ChoiceInputField>
+      </SSRProvider>
+    )
+    const results = await axe(container)
+    expect(results).toHaveNoViolations()
+    cleanup()
```
rezrah, line 211: Might be worth running `cleanup` and `jest.clearAllMocks();` on an `afterEach` block at the top of the test suite, as you're using spy a fair bit here

rezrah (approved):
💥 nice one @mperrotti

PR #1621, opened 2021-11-17
Fix TypeScript errors in the next_major branch

jfuchs:
Maybe this will work better?

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,7 +79,7 @@
         "eslint-plugin-jsx-a11y": "6.4.1",
         "eslint-plugin-mdx": "1.15.1",
         "eslint-plugin-prettier": "3.4.0",
-        "eslint-plugin-primer-react": "0.6.1",
+        "eslint-plugin-primer-react": "0.7.0",
         "eslint-plugin-react": "7.24.0",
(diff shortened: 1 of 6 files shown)


.github/workflows/release_canary.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches-ignore:
       - 'main'
       - 'next_major'
+      - '**-next_major'
```
colebemis, line 7:
```suggestion
```

colebemis (approved):
😈

jfuchs:
Chatted with @colebemis about this and we decided (or I convinced him?) that it's okay to merge this with broken — there are a few problems that aren't worth making work in this PR when the merge result will still be what we want:

- the `size` workflow is failing because typechecking in next_major is broken
- the `release_canary` workflow is failing because this branch, which needs to be based off next_major to fix type issues in next_major, includes a pre.json (and changesets won't let you do a canary release from a prerelease branch)

PR #1634, opened 2021-11-22
Docs for the new button

pksjce:
Added new documentation for the new button component.

https://primer-components-65wwy51f1-primer.vercel.app/react/NewButton

### Screenshots

Screen Shot 2021-11-22 at 4 28 31 pm

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
...
+import {Props} from '../src/props'
+import {ComponentChecklist} from '../src/component-checklist'
+import {NewButton} from '@primer/components'
+
+## Usage
+
+### Installation
+
+```js
+import {NewButton} from '@primer/components/drafts'
```
siddharthkp, line 18: Can we export `Button` instead? it's scoped in drafts

docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
...
+</NewButton>
+```
+
+### Styling a button
+
+A button can be styled in an appropriate manner using the `sx` prop. This maybe to change width, or to add margins etc.
+Here's an example of a block button which takes 100% of available width. Checkout [styled-system](https://styled-system.com/) to see what you can send in an `sx` prop.
+
+```jsx live
+<NewButton sx={{width: '100%'}}>Block</NewButton>
```
rezrah, line 125: Question: 100% width is presumably a common use case for many Button users, so should we instead turn that into a `block` prop? `sx` example could then be for custom margins, as you described?

docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
...
+  </NewButton>
+</>
+```
+
+### Icon only button
+
+This uses the `icon` prop and creates a nice square button with only the icon visible.
+
+```jsx live
+<NewButton icon={SearchIcon}>Search</NewButton>
```
siddharthkp, line 85: ah, I didn't pick up on this detail earlier! 

```jsx
<NewButton leadingIcon={SearchIcon}>Search</NewButton>
<NewButton icon={SearchIcon}>Search</NewButton>
```

Putting these two together, it wasn't obvious to me that one of them does not render the text 🤔 

How do you feel about an API like:

```jsx
<NewButton leadingIcon={SearchIcon} aria-label="Search files"/>
```

which renders only the icon in a square box based on the absence of text/children?

docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
...
+### Counter component
+
+A common use case for primer is a button with a counter component which shows count value as provided by the consumer.
+We provide `Button.Counter` as a composite component which requires you to provide a number as child.
+It renders a counter with the same `variant` styles as the parent Button component
+
+```jsx live
+<NewButton>
+  Watch
+  <NewButton.Counter>1</NewButton.Counter>
```
siddharthkp, line 115: Love to see it!

docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
+---
+title: Buttons
```
colebemis, line 2: Since we don't have multiple buttons components anymore:
```suggestion
title: Button
```

docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
...
+status: Alpha
+source: https://github.com/primer/react/tree/main/src/NewButton
+storybook: '/react/storybook?path=/story/composite-components-new-button--default-button'
+description: Button is core design component used for actions.
```
colebemis, line 6: Can we active voice here? [Related comment](https://github.com/primer/component-metadata/pull/3)

```
// Passive voice
Button is used for ...

// Active voice
Use button for ...
```

docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
...
+    adaptsToScreenSizes: true,
+    fullTestCoverage: true,
+    usedInProduction: null,
+    usageExamplesDocumented: true,
+    designReviewed: null,
+    a11yReviewed: null,
+    stableApi: false,
+    addressedApiFeedback: false,
+    hasDesignGuidelines: null,
+    hasFigmaComponent: null
```
colebemis, line 167: `null` here means "not applicable" but I think all of these "null" cases are things we want to eventually be "true"
```suggestion
    usedInProduction: false,
    usageExamplesDocumented: true,
    designReviewed: false,
    a11yReviewed: false,
    stableApi: false,
    addressedApiFeedback: false,
    hasDesignGuidelines: false,
    hasFigmaComponent: false
```

docs/content/NewButton.mdx
```diff
@@ -0,0 +1,169 @@
...
+| sx           | `BetterSystemStyleObject` |           | accepts all [styled system props](https://styled-system.com/table/).                                                                      |
+
+### Button.Counter
+
+| Prop name  | Type                      | Default | Description                                                          |
+| :--------- | :------------------------ | :-----: | :------------------------------------------------------------------- |
+| children\* | Number                    |         | The child for this has to be a number                                |
+| sx         | `BetterSystemStyleObject` |         | accepts all [styled system props](https://styled-system.com/table/). |
+
+## Component Status
```
colebemis, line 151:
```suggestion
## Component status
```

docs/content/NewButton.mdx
colebemis: What happens if people render multiple `Button.Counter`s?

docs/content/NewButton.mdx
pksjce: So we had a bit of a discussion around this specific API. I had marked the comments `Resolved` and that kinda hides the whole convo! Here's a link to why I concluded on this API.  @langermank convinced me that we should make button content text as mandatory.  Then I researched a few options including the `aria-label` one before settling on this. What do you think?

https://github.com/primer/react/pull/1532#discussion_r744338329

docs/content/NewButton.mdx
pksjce: Yeah.. The button api can be broken in multiple ways right now. You can basically provide any child component, or multiple `Button.Counter`. There's no design reason for it to have multiple counters. Don't think we need to account for those kind of usages.

docs/content/NewButton.mdx
pksjce: Yeah, this is what I had initially implemented. Then I felt that the `Button` api would unnecessarily bloat with a new prop. It's just a small style addition, we can document it as a recipe.

Yesterday though, I found a bug. Basically if an icon and text need to be shown in a block width button, they show up with a large gap between them. I'm thinking I should revisit this block type of button in the near future. I'll put up an issue on this ASAP.

docs/content/NewButton.mdx
pksjce: Well after migration this will be the only button and I need to update docs too at that time. Its okay if it stays this way right?

docs/content/NewButton.mdx
siddharthkp: First of all, loving the iterative approach to good API. I get paralysed with aligning all the decisions, I am liking this approach a lot more!

Read through the conversation! sorry for missing this on the previous PR.

I agree with the discussion here, pushing for a "label" and avoiding the clunkiness of something like `hasIconProp`, buuuut I find the `icon` approach because it's hard to predict what the code does when you see it:

If you run into this in the codebase, 

````jsx
<NewButton icon={SearchIcon}>Search</NewButton>
```

you would _have to know_ that using `icon` instead of `leadingIcon` means that children will be used for screen readers users but not sighted users. It's implied/implicit by the choice of using the `icon` prop. Personally, this isn't what I would expect/predict. 

Of course, you can come to the primer docs and see what it does, but that's not ideal. For me, it comes down to the expectations of using `children` prop without _visually rendering_ the children inside the component.

Moving towards a solution, my personal preference is [Solution 2 proposed in the discussion](https://github.com/primer/react/pull/1532#discussion_r744338329)

Rewinding a bit, 
For a classic button with text and icon, something like this looks nice:

```jsx
<Button icon={SearchIcon}>Search files</Button>
```

But, we have the option of leading and trailing icons, so we would like an explicit choice.


If you do not want to render text inside the button, you'd want to remove it:

```jsx
<Button icon={SearchIcon} />
```

This of course is terrible, and we need you to label the button. 

The two methods you can choose are, either of them are valid.

```jsx
1. <Button icon={SearchIcon} aria-label="Search files" />
2. <Button icon={SearchIcon}><span hidden>Search files</span></Button>
```

For 1, We can make `aria-label` required if there are no children. To simplify 2.

langermank: Thanks @siddharthkp for laying this all out! This is a great recap of the conversations from the original Button PR 🎉 

To my knowledge, relying on `aria-label` might not be the _best_ solution, but certainly not a bad solution if we decide to go with that. I just did a little research to see if this is still a valid assumption and came across this [article from Scott O'Hara](https://www.scottohara.me/blog/2021/11/02/names-and-labels.html). About halfway down there's this bit:

> Anyway, aria-label can be used to name things, but it should be used only when necessary and when other native labeling mechanisms cannot be relied upon. Also, you’d do well to read up on how well aria-label behaves with translation services. aria-labelledby is definitely the preferred of the two attributes, not only for potential translation woes, but also the ARIA specification even points out that use of aria-labelledby is preferred over aria-label, particularly in cases where the content that will provide the element’s “label” is visually available.

He's referencing a native `input` in this example though, so perhaps `aria-label` is just as good as `aria-labelledby` for a `button` element. Not sure!

Here's a few things I like about the `aria-labelledby` API for a Button:
- `label` is `label` no matter the case of a Button. I think the helps the consumer think about their design more thoroughly when `label` is required. Its common to see an `ariaLabel` prop where the consumer passes in a `key` because they see it as some technical requirement, not a label for humans to read. If we go with `aria-label` as a prop, we could continue making `label` required in all Button cases and just handle it internally `<Button aria-label={label}`. **Maybe we can steer folks to use buttons with visual labels by asking them to think about button intent by always requiring a label.** 
- This API is also nice for loading buttons, if we see a need to add one in the future. The markup is there for the `hidden` label span, and if we want to capture the attention of a screen reader with `aria-live="polite"` we could programmatically provide some specific text for that loading state

> As detailed above, I don't like the implicit icon prop here because you simply have to know about it.

I agree with you here, but I also understand how adding another boolean prop could be clunky for the consumer 😆 so here are some ideas for adding a billion more props (hehe) but keeping `icon` simple

```
// icon prop, additional props for design
<Button icon={someIcon} label={label} iconLeading />
<Button icon={someIcon} label={label} iconTrailing />
<Button icon={someIcon} label={label} iconOnly />

// could also reference the label instead
<Button icon={someIcon} label={label} labelRight />
<Button icon={someIcon} label={label} labelLeft />
<Button icon={someIcon} label={label} labelHidden />

// this feels weird to me because of the "leading"
<Button leadingIcon={someIcon} iconOnly />
```

siddharthkp:
```jsx
<Button icon={someIcon} iconTrailing />
```

I think we considered this for a hot second, but there are (rare but) valid use cases for having both left and right icons 🙃 

How do you feel about using "label" instead of "aria-label" and absorbing how it's implemented in the component? I would personally still prefer to make it required when there are no children though.

siddharthkp: Side note for @pksjce: Please feel free to merge the docs without a conclusion on this API! This PR documents the API we have.

langermank: @siddharthkp I think that's what I recommended above 🤔 `label` always required. The question is if it should be an `aria-label` or a `hidden` label with `aria-labelledby`. I think this question should go to the the accessibility team, maybe bring it up in the slack channel https://github.slack.com/archives/C0FSWLQ0Y

langermank: @pksjce per discussion today, it sounds like the a11y folks are recommending _all_ buttons have an `aria-label` in addition to a visual label (not just icon only buttons) as described [here](https://github.com/github/primer/issues/366#issuecomment-958141820). If I have that right, it probably makes sense to move forward with that here instead of `aria-labelledby`! 

I don't know if the plan for PVC is to have `aria-label` required in addition to `label` (maybe @manuelpuyol can confirm) but if its _not_ required, we could require it only if the Button is an IconButton (or `iconOnly`, whatever we move forward with).

I think CSS grid is still a solid choice for design and layout, this is more of a markup consideration :D

siddharthkp: @langermank Bah, I made it more confusing 😅 What I meant was

I don't like this style:

```jsx
<Button icon={someIcon} label={label} />
<Button icon={someIcon} label={label} iconOnly />
```

I like this style :
```jsx
<Button leadingIcon={someIcon}>text label in children</Button>
<Button leadingIcon={someIcon} label="text label in label prop when there are no children" />
```

We can own and change the label implementation (aria-label or aria-labelledby+hidden)

langermank: @siddharthkp ooOOooo... interesting 👀 had not thought of it this way before! 🤯 it makes sense!

siddharthkp:
I think we should merge this and then continue iterating! Documented component > perfect component

siddharthkp (approved):
Voting for merging this and then iterating on this!

mperrotti (approved):
LGTM. Just resolve merge conflicts.

PR #1643, opened 2021-11-22
Avatar, BranchName, Details, Heading, and ProgressBar no longer accept styled-system props.

jfuchs:
I went to clean up some docs and noticed I missed a few in the first pass.

See https://github.com/github/primer/issues/296.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/system-props.mdx
+++ b/docs/content/system-props.mdx
@@ -6,7 +6,7 @@ import {PropsList, COMMON, LAYOUT, BORDER, TYPOGRAPHY, FLEX, POSITION, GRID} fro
 
 <Note variant="warning">
 
-System props are deprecated in all components except [Box](/Box). Please use the [`sx` prop](/overriding-styles) instead.
+System props are deprecated in all components except [Box](/Box) and [Text](/Text). Please use the [`sx` prop](/overriding-styles) instead.
 
 </Note>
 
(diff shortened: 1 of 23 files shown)


colebemis (approved):

docs/content/system-props.mdx
```diff
@@ -6,7 +6,7 @@ import {PropsList, COMMON, LAYOUT, BORDER, TYPOGRAPHY, FLEX, POSITION, GRID} fro
 
 <Note variant="warning">
 
-System props are deprecated in all components except [Box](/Box). Please use the [`sx` prop](/overriding-styles) instead.
+System props are deprecated in all components except [Box](/Box) and [Text](/Text). Please use the [`sx` prop](/overriding-styles) instead.
```
pksjce, line 9: TIL. Why are these two components the exceptions?

pksjce (approved):

docs/content/system-props.mdx
jfuchs: As I understand it, the thought is that these two components are the basic building blocks of styling — whereas other components _should_ be opinionated about their styling and overriding their styles shouldn't be as dominant a part of their API.

I lean toward this approach but I don't think it would be the end of the world if we went with `sx` everywhere.

cc @colebemis who I think is on the fence about it.

PR #1648, opened 2021-11-23
Fixes Overlay types to allow DOM props

jfuchs:
Our v32 release caused a regression in our types — specifically that Overlay's `props` did not include DOM props. This PR fixes that using the Radix polymorphic type.

See https://github.slack.com/archives/C01L618AEP9/p1637702270252200

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


colebemis (approved):
Maybe we should remove polymorphism from Overlay eventually. But this seems like a good incremental step to get the types to match the implementation 👍

Issue #1635, opened 2021-11-22
Error: Cannot find module 'lodash.isempty'


peterbe:
**Describe the bug**

**Note:** As this is a public repo, please be careful not to include details or screenshots from unreleased GitHub products or features. In most cases, a good bug report should be able to describe the new component without including business logic or feature details, but if you must discuss context relating to an unreleased feature, please open an issue in the private [Design Systems repo](https://github.com/github/design-systems/issues/new/choose) and link to it here.

**To Reproduce**
1. Try to upgrade from 31.1.0 to 31.2.0
2. See error

```
Error occurred prerendering page "/". Read more: https://nextjs.org/docs/messages/prerender-error
Error: Cannot find module 'lodash.isempty'
Require stack:
- /home/runner/work/action-stats/action-stats/node_modules/@primer/components/lib/utils/theme.js
- /home/runner/work/action-stats/action-stats/node_modules/@primer/components/lib/theme-preval.js
- /home/runner/work/action-stats/action-stats/node_modules/@primer/components/lib/theme.js
- /home/runner/work/action-stats/action-stats/node_modules/@primer/components/lib/index.js
- /home/runner/work/action-stats/action-stats/.next/server/pages/index.js
- /home/runner/work/action-stats/action-stats/node_modules/next/dist/server/require.js
- /home/runner/work/action-stats/action-stats/node_modules/next/dist/server/load-components.js
- /home/runner/work/action-stats/action-stats/node_modules/next/dist/export/worker.js
- /home/runner/work/action-stats/action-stats/node_modules/jest-worker/build/workers/processChild.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/home/runner/work/action-stats/action-stats/node_modules/@primer/components/lib/utils/theme.js:4:17)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
info  - Exporting (3/3)
Error: Export encountered errors on following paths:
	/
[...]

PR #1652, opened 2021-11-23
Revert "Separate typechecking and building type declarations"

colebemis:
Reverts primer/react#1588 which was causing `theme-preval.js` to not actually be preval'd.  Reverting this PR fixes the `Error: Cannot find module 'lodash.isempty'` error when upgrading to Primer React >v31.2.0.

Fixes https://github.com/primer/react/issues/1635


jfuchs (approved):
😥

PR #1653, opened 2021-11-23
Snapshot test theme-preval.js

colebemis:
We recently released a version of Primer React that broke theme-preval.js (https://github.com/primer/react/issues/1635). This PR adds a test for theme-preval.js that will prevent us from making a similar mistake again.

### DIFF
--- a/src/__tests__/__snapshots__/themePreval.test.ts.snap
+++ b/src/__tests__/__snapshots__/themePreval.test.ts.snap
@@ -1,0 +1,3176 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`snapshot theme-preval.js 1`] = `
+"// this file was prevaled
+// This file needs to be a JavaScript file using CommonJS to be compatiable with preval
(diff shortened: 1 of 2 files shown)


jfuchs (approved):
I like it. Thoughts:

- What does it mean for someone testing locally who hasn't run a build? It would fail, right? Maybe that's okay, but we could make that experience more obvious?
- I wonder if this might help us catch theme changes ... but maybe folks don't always read snapshots?

siddharthkp (approved):

colebemis:
@jfuchs all good questions! Let's move forward with this and see if the snapshot test causes any problems.

rezrah:
> * What does it mean for someone testing locally who hasn't run a build? It would fail, right? Maybe that's okay, but we could make that experience more obvious?

🙏  yes please.. I just spent a good ten minutes trying to work out why the snapshot for theme-preval was failing despite being up-to-date, before realising I needed to run `build`. Tests - particularly snapshots - I assumed should run deterministically from checked in code, so if that's not the case here perhaps we can make `build` a prerequisite to `test` and create a separate `test:ci` that doesn't need to do this?

PR #1657, opened 2021-11-24
Adds a component to support checkbox and radio groups

mperrotti:
Adds a component to show a set of related checkbox or radio inputs

Closes https://github.com/github/primer/issues/496

#### selectionVariant="single" (default)
Screen Shot 2021-12-03 at 3 58 21 PM

#### selectionVariant="multiple"
Screen Shot 2021-12-03 at 3 57 57 PM

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


pksjce:
Review summary

Nice use of the slots api. This seems like a great use case to extend the same pattern. I tested out the stories and they seem to work pretty well. The api seems minimal wrt props used. Component construction seems a bit deeply nested.

Some nits I could think of
- Disabled state story is absent. Should disabled state also mute the colors on the label and caption?
- Should we throw an error if the slot for Input/Label inside a Field is missing? We have the info in the context 🤷‍♀️ 
- Naming nits - `ChoiceFieldset.ChoiceField` could be `ChoiceFieldset.Item` inline with our other components and the `List/Item` relation.

Other comments

- I noticed that the "required field" aria description is hardcoded. How does this affect changing language in the consumer app?
- One observation is that the Slot api makes the overall api a bit verbose. But I suppose it is better than passing props or array objects into the component like the old ActionList.

Structure seems to look like 
-- ChoiceFieldset
    |-- Legend (required?)
    |-- List
         |-- ChoiceField [] (disabled?)
              |-- Input (checked?)
              |-- LeadingVisual
              |-- Label
              |-- Caption
    |-- Validation
    |-- Caption

mperrotti:
Thanks for the quick feedback @pksjce !

> Disabled state story is absent. Should disabled state also mute the colors on the label and caption?

Great catch - I need to add that story. You can see what disabled fields look like in the Storybook deployment from my [input field PR](https://github.com/primer/react/pull/1611): https://primer-components-pi9cymoo5-primer.vercel.app/react/storybook?path=/story/forms-checkboxinputfield--with-caption&args=disabled:true

The label goes from `fg.default` to `fg.muted`, but the caption stays colored `fg.muted`. I'm going to try `fg.subtle` to maintain the contrast between the label and the caption.

---

> Should we throw an error if the slot for Input/Label inside a Field is missing? We have the info in the context 🤷‍♀️

I log a `console.error` if we do that. See [this line](https://github.com/primer/react/pull/1611/files#diff-66f1d7952f3b031ab8b62c9e465b5a1c658a224f9c14c68db7d8fb5ad57efc65R80) from my input field PR.

---

> Naming nits - ChoiceFieldset.ChoiceField could be ChoiceFieldset.Item inline with our other components and the List/Item relation.

I agree with you - `ChoiceFieldset.Item` is better.

---

> Component construction seems a bit deeply nested

> the Slot api makes the overall api a bit verbose

I agree with both of these, but I couldn't think of a more concise API that's easy to use and aligns with the direction we started going in the new `ActionList` component API.

---

pksjce (approved):

mperrotti:
As I've been working on the docs for this, I've been making improvements. Hold off on reviewing until you see that `ChoiceFieldset.mdx` has been pushed up.

mperrotti:
@pksjce - I made some updates to the design and API. Feel free to take another look

docs/content/ChoiceFieldset.mdx
```diff
@@ -0,0 +1,407 @@
...
+    </ChoiceFieldset>
+  )
+}
+
+render(<ChoiceFieldsetWithValidation />)
+```
+
+## ChoiceFieldset.Legend
+
+A title for the set of choices. A `ChoiceFieldset.Legend` must be passed, but it may be visually hidden.
```
mperrotti, line 217: Technically, a legend is optional in a fieldset: 
> Permitted content An optional <legend> element, followed by flow content.

—https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#technical_summary



I think maybe we could let consumers use a ChoiceFieldset without a legend. What do you think @alliethu ?

jfuchs (commented):
Small nit: if the directory is called `ChoiceFieldset` should `ChoiceFieldCaption` and `ChoiceFieldLabel` be renamed to `ChoiceFieldsetCaption` and `ChoiceFieldsetLabel`?

jfuchs:
Last time I spoke with @siddharthkp about `createSlots`, it sounded like there was some uncertainty about using it broadly. Are we good with bringing it into more components now?

siddharthkp:
> Last time I spoke with @siddharthkp about createSlots, it sounded like there was some uncertainty about using it broadly. Are we good with bringing it into more components now?

(Copied over to https://github.com/primer/react/issues/1690)

Context: The current slot implementation isn't ideal when it comes to server side rendering - It only renders the freeform main slot on the server and then adds other slots after mounting on client (with useEffect). Which is a non-problem for components in an overlay, but can be a bad loading experience for other components with layout shift.

I need to make changes to the implementation that makes it consistent across server and client. This however will introduce a limitation that a slot needs to be a direct child of the slot container, deep nesting not allowed. 

so this is allowed:
```diff
<ActionList>
+ <motion.div>
   <ActionList.Item>
     free text 
     <ActionList.Description>description</ActionList.Description>
   </ActionList.Item>
+ </motion.div>
</ActionList>
```

but this isn't:

```diff
<ActionList>
    <ActionList.Item>
     free text 
+   <motion.div> // deep nesting inside Item
       <ActionList.Description>description</ActionList.Description>
+   </motion.div>
   </ActionList.Item>
</ActionList>
```

I want to say that's a fair tradeoff to make inside an `ActionList.Item`. @mperrotti would be able to comment if that's safe for `ChoiceFieldset` as well. Happy to discuss this with potential use cases.

The API of how we create slots internally might change and I'll update the components as well along with it!

mperrotti:
@siddharthkp - thanks for the context.

This seems like a reasonable limitation to me, and I don't foresee this being a problem as long as I can still use [phrasing elements](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content) in `ChoiceFieldset.Description`.

Example:
```jsx
<ChoiceFieldset>
  <ChoiceFieldset.Description>
    This description has a <a href="google.com">link</a>
  </ChoiceFieldset.Description>
</ChoiceFieldset>
```

mperrotti:
> Small nit: if the directory is called ChoiceFieldset should ChoiceFieldCaption and ChoiceFieldLabel be renamed to ChoiceFieldsetCaption and ChoiceFieldsetLabel?

@jfuchs - I was worried that `ChoiceFieldsetCaption` and `ChoiceFieldsetDescription` would get mixed up, and `ChoiceFieldsetLabel` and `ChoiceFieldsetLegend` would get mixed up. This isn't something we'd have to worry about for consumers though, so it's probably fine

Here's what usage looks like:
```jsx
    <ChoiceFieldset>
      <ChoiceFieldset.Legend>Prefered Primer component interface</ChoiceFieldset.Legend>
      <ChoiceFieldset.Description>Pick your top two</ChoiceFieldset.Description>

      <ChoiceFieldset.List selectionVariant="multiple">
        {choices.map(choice => (
          <ChoiceFieldset.Item value={choice.value}>
            <Item.Label>{choice.displayName}</Item.Label>
            <Item.Caption>{choice.description}</Item.Caption>
          </ChoiceFieldset.Item>
        ))}
      </ChoiceFieldset.List>

      <ChoiceFieldset.Validation validationKey="tooManySelections">
        Only two selections are allowed
      </ChoiceFieldset.Validation>
    </ChoiceFieldset>
```

src/stories/ChoiceFieldset.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+)
+
+export const WithValidation = ({selectionVariant, ...restArgs}: Args) => (
+  <ChoiceFieldset {...restArgs}>
+    <ChoiceFieldset.Legend>Legend</ChoiceFieldset.Legend>
+    <ChoiceFieldset.List selectionVariant={selectionVariant}>
+      <ChoiceFieldset.Item value="labelOne">Label one</ChoiceFieldset.Item>
+      <ChoiceFieldset.Item value="labelTwo">Label two</ChoiceFieldset.Item>
+    </ChoiceFieldset.List>
+    <ChoiceFieldset.Validation validationKey="validChoice">You made the right selection</ChoiceFieldset.Validation>
```
rezrah, line 160: Might just be my eyes playing tricks on me, but I think there's some alignment issues with the icon?

Screenshot 2021-12-07 at 12 08 00

src/ChoiceFieldset/ChoiceFieldLabel.tsx
```diff
@@ -0,0 +1,14 @@
...
+import ChoiceFieldsetListContext from './ChoiceFieldsetListContext'
+
+const ChoiceFieldLabel: React.FC = ({children}) => {
+  const choiceFieldsetListContext = useContext(ChoiceFieldsetListContext)
+  if (choiceFieldsetListContext === null) {
+    throw new Error('ChoiceFieldsetListContext returned null')
+  }
+  const {fieldComponent: FieldComponent} = choiceFieldsetListContext
+
+  return <FieldComponent.Label>{children}</FieldComponent.Label>
```
rezrah, line 11: Potential scope-creep, so feel free to ignore in this PR... shall we add a cursor: pointer to this label so it's consistent with the input? As clicking it will trigger the associated input? Feels a bit strange it doesn't in the examples.

src/ChoiceFieldset/ChoiceFieldsetList.tsx
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        return selected.filter(selectedValue => selectedValue !== value)
+      }
+      return (
+        <ChoiceFieldsetListContext.Provider
+          value={{
+            disabled,
+            selected,
+            name: name || uniqueId(),
```
rezrah, line 43: Do we need to regenerate this unique id on each render? I think it be more performant to hoist this outside of the component render/return?

src/ChoiceFieldset/ChoiceFieldsetList.tsx
```diff
@@ -0,0 +1,66 @@
...
+            fieldComponent: selectionVariant === 'multiple' ? CheckboxInputField : RadioInputField,
+            onChange: (e: ChangeEvent<HTMLInputElement>) => {
```
rezrah, line 45: Similar to previous comment, can this be an assigned function to avoid the inline arrow being recreated on each render? I.e.

```jsx
const handleChange = ( {currentTarget : {value, checked}) => {
  ...
  if(onSelect) {
     ...
    // move private fns in here
     ....
  }
}


...

// this helps the readability of the render too IMO
onChange: handleChange
``

File: src/ChoiceFieldset/ChoiceFieldsetListItem.tsx [hunk 6]
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * The value that is being selected
+   */
+  value: string
+}
+
+const ChoiceFieldsetListItem: React.FC<ChoiceFieldProps> = ({children, id, disabled: disabledProp, value}) => {
+  const choiceFieldsetListContext = useContext(ChoiceFieldsetListContext)
+  if (choiceFieldsetListContext === null) {
+    throw new Error('ChoiceFieldsetListContext returned null')
```
rezrah, line 28: curious about this, do our users expect us to be throwing native stack errors like this? It'll trigger hits on their observability tooling, that they may not be expecting (or will they?). We use console.warn in other places too, should this be console.error instead?

src/stories/ChoiceFieldset.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+      )
+    }
+  ]
+} as Meta
+
+export const RadioGroup = ({selectionVariant, ...restArgs}: Args) => (
+  <ChoiceFieldset {...restArgs}>
+    <ChoiceFieldset.Legend>Legend</ChoiceFieldset.Legend>
+    <ChoiceFieldset.List selectionVariant={selectionVariant}>
+      <ChoiceFieldset.Item value="labelOne">Label one</ChoiceFieldset.Item>
```
rezrah, line 57: I can't seem to keyboard tab through these items, is this a known issue?

src/ChoiceFieldset/ChoiceFieldsetLegend.tsx
```diff
@@ -0,0 +1,39 @@
...
+        isVisible={!visuallyHidden}
+        title={required ? 'required field' : undefined}
+        sx={{
+          color: disabled ? 'fg.muted' : undefined,
+          fontSize: 2,
+          padding: 0
+        }}
+      >
+        {required ? (
+          <Box display="flex" as="span">
```
rezrah, line 27: It's a wider system props issue perhaps, but shall we use `sx` here instead? This creates some ugly and redundant html.
Screenshot 2021-12-07 at 12 51 48

src/ChoiceFieldset/ChoiceFieldset.tsx
```diff
@@ -0,0 +1,138 @@
...
+        required,
+        selected,
+        validationMessageId
+      }}
+    >
+      {slots => {
+        const isLegendVisible = React.isValidElement(slots.Legend) && slots.Legend.props.isVisible
+
+        return (
+          <div>
```
rezrah, line 93: Switch to fragment?

src/stories/ChoiceFieldset.stories.tsx
mperrotti: I think you're right. It looks like a `line-height` issue. I'll investigate

src/ChoiceFieldset/ChoiceFieldLabel.tsx
mperrotti: I could go either way. I don't think labels typically have `cursor: pointer`, but I think it's a nice enhancement.

src/ChoiceFieldset/ChoiceFieldsetListItem.tsx
mperrotti: I went with an `Error` because if the context returns null, the component function correctly. I think this will feel natural to consumers.

src/stories/ChoiceFieldset.stories.tsx
mperrotti: You're not supposed to be able to tab through radio inputs with the same name - you use the arrow keys.

src/ChoiceFieldset/ChoiceFieldsetLegend.tsx
mperrotti: Yea, this looks like an issue with system props. I could use `sx` instead.

@colebemis - is this a known issue?

src/ChoiceFieldset/ChoiceFieldset.tsx
mperrotti: I intentionally wrapped this in a `div` so that everything is grouped in 1 DOM node. This way, if this is in a flex or grid container, the whole component would turn into a flex item or grid cell.

src/ChoiceFieldset/ChoiceFieldsetList.tsx
mperrotti: I used a hook that doesn't calculate a new ID on each re-render.

src/ChoiceFieldset/ChoiceFieldsetList.tsx
mperrotti: I couldn't figure out how to hoist the `onChange` handler up, but I at least moved `getSelectedCheckboxes` outside of the `ChoiceFieldsetList` render.

rezrah (approved):

docs/content/CheckboxInputField.mdx
```diff
@@ -11,11 +11,13 @@ import {ComponentChecklist} from '../src/component-checklist'
 TODO: add `validationStatus` examples when we have the validation styling done
 -->
 
-import {CheckboxInputField, TextInputWithTokens} from '@primer/components'
+import {CheckboxInputField} from '@primer/components'
 import {MarkGithubIcon} from '@primer/octicons-react'
 
 The `CheckboxInputField` component is used to render a labelled checkbox input and, optionally, associated hint text and graphic.
 
+**Use [ChoiceFieldset](/ChoiceFieldset) to render a group of related checkbox input fields.** Do not use this component on it's own unless you're rendering a single checkbox field on it's own, and there will never be related checkboxes.
```
rezrah, line 19: minor typo here: Its not it's

PR #1658, opened 2021-11-26
Add alpha Radio component

rezrah:
### Summary

Adds a native `RadioButton` component in `alpha` status. 🔘 

A baseline abstraction on the native HTML `type="radio"` input element, providing a Primer-consistent component API and enough test coverage for resiliency against future refactoring.

Closes https://github.com/github/primer/issues/489

### Try it here
- Docs: https://primer-components-8int6jo87-primer.vercel.app/react/RadioButton
- Storybook: https://primer-components-8int6jo87-primer.vercel.app/react/storybook?path=/story/forms-radio-button--default

### Review strategy
- [ ] Manually test component API in Storybook and Docs
- [ ] Verify test coverage on the critical path
- [ ] Verify that no console warnings are present
- [ ] Verify docs for conciseness and grammatical errors

### Screenshots
Screenshot 2021-11-26 at 10 47 36
Screenshot 2021-11-26 at 10 48 37
Screenshot 2021-11-26 at 10 48 52
Screenshot 2021-11-26 at 10 48 57
Screenshot 2021-11-26 at 10 49 05

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/RadioButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+    ref
+  ): ReactElement => {
+    return (
+      <StyledRadioButton
+        type="radio"
+        value={value}
+        name={name}
+        ref={ref}
+        disabled={disabled}
+        aria-disabled={disabled ? 'true' : 'false'}
```
pksjce, line 61: Wondering if we need to set the aria- properties at all? Taking care of the main html properties should be enough? Keen to learn more.

src/stories/RadioButton.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+  &:first-child {
+    margin-left: 0;
+  }
+
+  &[aria-disabled='true'] {
+    pointer-events: none;
+    cursor: not-allowed;
+    color: ${get('colors.primer.fg.disabled')};
+  }
+  ${COMMON}
```
pksjce, line 61: we don't need to pass `COMMON` props anymore.

rezrah: Cheers @pksjce . Forgot to remove these from the stories.. that's done now 👍 . Also removed from Checkbox one, as it was there too.

src/RadioButton.tsx
rezrah: It's a good question. They are primarily here to provide additional context for screen readers.. but also:

- Some screen readers supposedly have [better support ](https://www.levelaccess.com/how-not-to-misuse-aria-states-properties-and-roles/)for aria attributes vs inferring values from semantic tags. E.g. `aria-required`
- Some aria roles aren't supported natively. E.g. `aria-invalid`. So adding those should hopefully provide screen readers with additional context on UI state 
- Some aria tags have better support for non-standard values.  E.g.  `aria-checked`, which supports `mixed` to represent a js-only indeterminate modes.
- Explicitly setting a string boolean value felt like a stricter implementation of html, which should provide better utility while writing css or tests, rather than checking for whether an aria attribute is set or not by the js framework
- There is test coverage added against aria roles to guard the component API against future breaking changes /  refactoring, in case we were ever to move away from native checkboxes and radios in favour of custom ones.

Admittedly, there are a lot of assumptions here though. For some of these aria attributes, it may be better to remove them. @alliethu - do you have any opinions on this too?

docs/content/RadioButton.md
```diff
@@ -0,0 +1,92 @@
+---
+title: RadioButton
+description: Use radio buttons when users need to select one option from a list
+status: Alpha
+source: https://github.com/primer/react/blob/main/src/RadioButton.tsx
+storybook: '/react/storybook?path=/story/forms-radio-button--default'
+---
+
+import {ComponentChecklist} from '../src/component-checklist'
```
colebemis, line 9: ComponentChecklist is globally available in markdown files now 🎉 
```suggestion
```

docs/content/RadioButton.md
```diff
@@ -0,0 +1,92 @@
...
+<>
+  <RadioButton value="active radio button value" name="radio-group-name" />
+  <RadioButton value="active radio button value" name="radio-group-name" />
+  <RadioButton disabled value="inactive radio button value" name="radio-group-name" />
+</>
+```
+
+<Note>
+Please use a <a href="/react/Checkbox">Checkbox</a> if the user needs to select more than one option in a list
+</Note>
```
colebemis, line 23: Weird MDX quirk: If you add newlines, you can use markdown syntax within component children:
```suggestion
<Note>

Please use a [Checkbox](/Checkbox) if the user needs to select more than one option in a list.

</Note>
```

.changeset/giant-timers-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/components': minor
+---
+
+Adds a RadioButton component
```
colebemis, line 5: Curious if we should call this `Radio` instead of `RadioButton`? Don't have strong feelings either way, just wanted to raise the question in case there is a "correct" answer.

colebemis (approved):

docs/content/RadioButton.md
mperrotti: Wow, I had no idea! This is great.

.changeset/giant-timers-ring.md
mperrotti: We call the checkbox input a `Checkbox`, so I'm with Cole on this one.

src/RadioButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+   */
+  ref?: React.RefObject<HTMLInputElement>
+  /**
+   * Indicates whether the radio button must be checked before the form can be submitted
+   */
+  required?: boolean
+  /**
+   * Indicates whether the radio button validation state is non-standard
+   */
+  validationStatus?: 'error' | 'success' // TODO: hoist to Validation typings
```
mperrotti, line 34: Don't forget this TODO. Here's where I put them in my PR: https://github.com/primer/react/pull/1611/files#diff-410797b1c4259d74820358a792a62f83cfc7a1c0a0eb26226c89e4621b88d42cR1

.changeset/giant-timers-ring.md
rezrah: I called this `Radio` initially  and later renamed to `RadioButton` 😅 , as that appeared to be the common naming convention used by many other component libraries for this type of input. There's [a bit history around the 'button' aspect to it too](https://www.nngroup.com/articles/radio-buttons-default-selection/#articleBody). Shall I change it to Radio? Or it may be worth checking in with wider group too, as we'd need to coordinate the same naming with PVC and CSS eventually too.

docs/content/RadioButton.md
rezrah: Thanks. Removed.

.changeset/giant-timers-ring.md
rezrah: Update: Changed to `Radio` now.

mperrotti (approved):

src/RadioButton.tsx
alliethu: @rezrah As long as you're using **native** HTML radio buttons, you should not need to include these additional ARIA roles. The ARIA roles are made to be used with custom radio buttons to mimic the native version.

krygierklaudia225-hue (commented):
Fixes #8170

PR #1659, opened 2021-11-26
Button: enhancements with ButtonLink and IconButton

pksjce:
Describe your changes here.

 Add two new components
- IconButton
- ButtonLink

Change implementation of button to better accomodate sharing of code.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,8 +129,6 @@
     "react-dnd": "14.0.4",
     "react-dnd-html5-backend": "14.0.2",
     "react-dom": "17.0.2",
-    "react-router": "6.0.2",
-    "react-router-dom": "6.0.2",
     "react-test-renderer": "17.0.2",
     "rollup": "2.60.2",
     "rollup-plugin-babel": "4.4.0",
--- a/package.json
+++ b/package.json
@@ -129,6 +129,8 @@
     "react-dnd": "14.0.4",
     "react-dnd-html5-backend": "14.0.2",
     "react-dom": "17.0.2",
+    "react-router": "6.0.2",
+    "react-router-dom": "6.0.2",
     "react-test-renderer": "17.0.2",
     "rollup": "2.60.2",
     "rollup-plugin-babel": "4.4.0",
--- a/src/NewButton/icon-button.tsx
+++ b/src/NewButton/icon-button.tsx
@@ -1,0 +1,31 @@
+import React, {forwardRef} from 'react'
+import {merge, SxProp} from '../sx'
+import {useTheme} from '../ThemeProvider'
+import Box from '../Box'
+import {IconButtonProps} from './types'
+import ButtonBase from './button-base'
+import {getBaseStyles, getSizeStyles, getVariantStyles} from './styles'
+
+const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>((props, forwardedRef): JSX.Element => {
+  const {variant = 'default', size = 'medium', sx: sxProp = {}, icon: Icon, iconLabel} = props
+  const {theme} = useTheme()
+  const styles = {
+    ...getBaseStyles(theme)
+  }
+  const sxStyles = merge.all([
+    styles,
+    getSizeStyles(size, variant, true),
+    getVariantStyles(variant, theme),
+    sxProp as SxProp
+  ])
+  return (
+    <ButtonBase sx={sxStyles} ref={forwardedRef} {...props}>
+      <span hidden={true}>{iconLabel}</span>
+      <Box as="span" sx={{display: 'inline-block'}}>
+        <Icon />
+      </Box>
+    </ButtonBase>
+  )
+})
+
+export default IconButton
--- a/src/NewButton/icon-button.tsx
+++ b/src/NewButton/icon-button.tsx
@@ -3,8 +3,8 @@ import {merge, SxProp} from '../sx'
 import {useTheme} from '../ThemeProvider'
 import Box from '../Box'
 import {IconButtonProps} from './types'
-import ButtonBase from './button-base'
 import {getBaseStyles, getSizeStyles, getVariantStyles} from './styles'
+import {Base} from './button-base'
 
 const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>((props, forwardedRef): JSX.Element => {
   const {variant = 'default', size = 'medium', sx: sxProp = {}, icon: Icon, iconLabel} = props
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -127,6 +127,8 @@
     "react-dnd": "14.0.4",
     "react-dnd-html5-backend": "14.0.2",
     "react-dom": "17.0.2",
+    "react-router": "6.0.2",
+    "react-router-dom": "6.0.2",
```
pksjce, line 131: Added for storybook example

src/NewButton/types.ts
```diff
@@ -27,10 +31,28 @@ export type ButtonProps = {
    * The trailing icon comes after button content
    */
   trailingIcon?: React.FunctionComponent<IconProps>
+  children: React.ReactNode
+  as?: React.ElementType
```
pksjce, line 35: I shouldn't need to add this if I use `PolymorphicForwardRefComponent` but typescript simply wouldn't recognise this without explicit prop definition.

src/NewButton/types.ts
```diff
@@ -27,10 +31,28 @@ export type ButtonProps = {
...
-  children: React.ReactNode
-} & SxProp &
-  HTMLAttributes<HTMLButtonElement>
+  icon: React.FunctionComponent<IconProps>
+  iconLabel: string
+} & ButtonBaseProps
+
+// adopted from React.AnchorHTMLAttributes
+export type LinkButtonProps = {
+  underline?: boolean
```
pksjce, line 48: [help-wanted[: accessibility audit mentions that alink should be underlined on focus. Not sure how important it is.

src/NewButton/icon-button.tsx
```diff
@@ -0,0 +1,31 @@
+import React, {forwardRef} from 'react'
```
rezrah, line 1: Should this - and the other files that export default React components - be renamed in PascalCase? Or is this intentional?

src/NewButton/icon-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+  const sxStyles = merge.all([
+    styles,
+    getSizeStyles(size, variant, true),
+    getVariantStyles(variant, theme),
+    sxProp as SxProp
+  ])
+  return (
+    <Base sx={sxStyles} ref={forwardedRef} {...props}>
+      <span hidden={true}>{iconLabel}</span>
```
rezrah, line 23: Is `hidden` needed anymore? If so, should it be `aria-hidden` instead?

src/NewButton/icon-button.tsx
pksjce: Ah I see! Will change up the file names in a followup PR 👍

src/NewButton/icon-button.tsx
pksjce: Since its an icon only component, the idea was to have a visibly hidden icon label but the screenreaders will be able to access it.

src/NewButton/styles.ts
```diff
@@ -0,0 +1,233 @@
...
+import {Theme} from '../ThemeProvider'
+
+export const TEXT_ROW_HEIGHT = '20px' // custom value off the scale
+
+export const getVariantStyles = (variant: VariantType = 'default', theme?: Theme) => {
+  const style = {
+    default: {
+      color: 'btn.text',
+      backgroundColor: 'btn.bg',
+      boxShadow: `${theme?.shadows.btn.shadow}, ${theme?.shadows.btn.insetShadow}`,
```
rezrah, line 11: is `theme` assumed to always be defined inside the fn? If so, do we need the optional chaining? If not, should we provide a fallback?

src/NewButton/styles.ts
```diff
@@ -0,0 +1,233 @@
...
+      },
+      '&:active:not([disabled])': {
+        backgroundColor: 'btn.primary.selectedBg',
+        boxShadow: `${theme?.shadows.btn.primary.selectedShadow}`
+      },
+      '&:disabled': {
+        color: 'btn.primary.disabledText',
+        backgroundColor: 'btn.primary.disabledBg'
+      },
+      '[data-component="ButtonCounter"]': {
```
rezrah, line 49: Minor observation: I can't see it in the test file, but given we're applying design logic against DOM node attributes, shall we add some extra tests against the `data-*` attributes to ensure that these styles are picked up?

src/NewButton/icon-button.tsx
rezrah: Okay, I believe the `hidden` html5 attribute will also hide it from screenreaders too though, because [it prevents the semantics from getting surfaced to the accessibility api  ](https://www.w3.org/TR/core-aam/#exclude_elements2). I think you'll either need to just visually hide it with styles, or explicitly add `aria-hidden="false"`, but the latter seems counter-productive.

src/NewButton/styles.ts
```diff
@@ -0,0 +1,233 @@
...
+        color: 'primer.fg.disabled'
+      }
+    },
+    outline: {
+      color: 'btn.outline.text',
+      boxShadow: `${theme?.shadows.btn.shadow}`,
+
+      '&:hover': {
+        color: 'btn.outline.hoverText',
+        backgroundColor: 'btn.outline.hoverBg',
```
rezrah, line 117: Potentially unrelated to this PR, but should this hover style be disabled? 

https://user-images.githubusercontent.com/13340707/145564523-34edd694-847b-4037-8f0d-5755ffe96aad.mov

src/__tests__/NewButton.test.tsx
```diff
@@ -63,7 +63,7 @@ describe('Button', () => {
     expect(render(<Button variant="danger">Danger</Button>)).toHaveStyleRule('background-color', '#f6f8fa')
   })
   it('styles icon only button to make it a square', () => {
-    const IconOnlyButton = render(<Button icon={SearchIcon}>Search icon only button</Button>)
+    const IconOnlyButton = render(<IconButton icon={SearchIcon} iconLabel="Search icon only button" />)
```
rezrah, line 66: One for another time, but wanted to put it on your radar. Can we move away from Enzyme to use the render from testing library instead? It's already imported here as `HTMLRender`. On the disabled test for example, we can instead look for `disabled` html attributes instead of testing the implementation props. What do you think?

rezrah (approved):
Looks great @pksjce. I've left additional questions/recommendations but would be happy for them to be resolved later if you want to move ahead with this.

src/NewButton/styles.ts
pksjce: Honestly I only used it to make typescript happy. This seems common usage across our codebase.

src/NewButton/styles.ts
pksjce: yes, some counter tests are much needed.

src/NewButton/styles.ts
pksjce: Thanks a ton for this catch! Fixed now.

src/__tests__/NewButton.test.tsx
pksjce: Good idea! Its much nicer to use testing library. Rewrote the tests in that.

PR #1661, opened 2021-11-29
Text Input enhancements

pksjce:
Describe your changes here.

Closes # https://github.com/github/primer/issues/501

### Screenshots

Some additions

Error condition
image

Warning condition
image

Trailing Icon
image

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/TextInput.tsx
+++ b/src/TextInput.tsx
@@ -61,13 +61,13 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
       >
         {IconComponent && <IconComponent className="TextInput-icon" />}
         {LeadingVisual && (
-          <span data-component="leadingVisual">
+          <span className="TextInput-icon">
             {typeof LeadingVisual === 'function' ? <LeadingVisual /> : LeadingVisual}
           </span>
         )}
         <UnstyledTextInput ref={ref} disabled={disabled} {...inputProps} data-component="input" />
         {TrailingVisual && (
-          <span data-component="leadingVisual">
+          <span className="TextInput-icon">
             {typeof TrailingVisual === 'function' ? <TrailingVisual /> : TrailingVisual}
           </span>
         )}
--- a/src/TextInput.tsx
+++ b/src/TextInput.tsx
@@ -6,10 +6,10 @@ import TextInputWrapper from './_TextInputWrapper'
 
 type NonPassthroughProps = {
   className?: string
-  /** @deprecated Use `leadingIcon` or `trailingIcon` prop instead */
+  /** @deprecated Use `leadingVisual` or `trailingVisual` prop instead */
   icon?: React.ComponentType<{className?: string}>
-  leadingIcon?: React.ComponentType<{className?: string}>
-  trailingIcon?: React.ComponentType<{className?: string}>
+  leadingVisual?: string | React.ComponentType<{className?: string}>
+  trailingVisual?: string | React.ComponentType<{className?: string}>
 } & Pick<
   ComponentProps<typeof TextInputWrapper>,
   'block' | 'contrast' | 'disabled' | 'sx' | 'width' | 'maxWidth' | 'minWidth' | 'variant' | 'size'
@@ -23,8 +23,8 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
   (
     {
       icon: IconComponent,
-      leadingIcon: LeadingIconComponent,
-      trailingIcon: TrailingIconComponent,
+      leadingVisual: LeadingVisual,
+      trailingVisual: TrailingVisual,
       block,
       className,
       contrast,
@@ -52,7 +52,6 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
         validationStatus={validationStatus}
         contrast={contrast}
         disabled={disabled}
-        hasIcon={!!IconComponent || !!(LeadingIconComponent || TrailingIconComponent)}
         sx={sxProp}
         size={sizeProp}
         width={widthProp}
@@ -60,10 +59,18 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
         maxWidth={maxWidthProp}
         variant={variantProp}
       >
-        {IconComponent && <IconComponent className="TextInput-leading-icon" />}
-        {LeadingIconComponent && <LeadingIconComponent className="TextInput-leading-icon" />}
+        {IconComponent && <IconComponent className="TextInput-icon" />}
+        {LeadingVisual && (
+          <span data-component="leadingVisual">
+            {typeof LeadingVisual === 'function' ? <LeadingVisual /> : LeadingVisual}
+          </span>
+        )}
         <UnstyledTextInput ref={ref} disabled={disabled} {...inputProps} data-component="input" />
-        {TrailingIconComponent && <TrailingIconComponent className="TextInput-trailing-icon" />}
+        {TrailingVisual && (
+          <span data-component="leadingVisual">
+            {typeof TrailingVisual === 'function' ? <TrailingVisual /> : TrailingVisual}
+          </span>
+        )}
       </TextInputWrapper>
     )
   }
--- a/src/TextInput.tsx
+++ b/src/TextInput.tsx
@@ -6,13 +6,13 @@ import TextInputWrapper from './_TextInputWrapper'
 
 type NonPassthroughProps = {
   className?: string
-  // deprecate icon prop
+  /** @deprecated Use `leadingIcon` or `trailingIcon` prop instead */
   icon?: React.ComponentType<{className?: string}>
   leadingIcon?: React.ComponentType<{className?: string}>
   trailingIcon?: React.ComponentType<{className?: string}>
 } & Pick<
   ComponentProps<typeof TextInputWrapper>,
-  'block' | 'contrast' | 'disabled' | 'sx' | 'theme' | 'width' | 'maxWidth' | 'minWidth' | 'variant' | 'size'
+  'block' | 'contrast' | 'disabled' | 'sx' | 'width' | 'maxWidth' | 'minWidth' | 'variant' | 'size'
 >
 
 // Note: using ComponentProps instead of ComponentPropsWithoutRef here would cause a type issue where `css` is a required prop.
@@ -29,9 +29,8 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
       className,
       contrast,
       disabled,
-      status,
+      validationStatus,
       sx: sxProp,
-      theme,
       size: sizeProp,
       // start deprecated props
       width: widthProp,
@@ -50,12 +49,11 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
       <TextInputWrapper
         block={block}
         className={wrapperClasses}
-        status={status}
+        validationStatus={validationStatus}
         contrast={contrast}
         disabled={disabled}
         hasIcon={!!IconComponent || !!(LeadingIconComponent || TrailingIconComponent)}
         sx={sxProp}
-        theme={theme}
         size={sizeProp}
         width={widthProp}
         minWidth={minWidthProp}
--- a/src/TextInput.tsx
+++ b/src/TextInput.tsx
@@ -6,10 +6,13 @@ import TextInputWrapper from './_TextInputWrapper'
 
 type NonPassthroughProps = {
   className?: string
+  // deprecate icon prop
   icon?: React.ComponentType<{className?: string}>
+  leadingIcon?: React.ComponentType<{className?: string}>
+  trailingIcon?: React.ComponentType<{className?: string}>
 } & Pick<
   ComponentProps<typeof TextInputWrapper>,
-  'block' | 'contrast' | 'disabled' | 'sx' | 'theme' | 'width' | 'maxWidth' | 'minWidth' | 'variant'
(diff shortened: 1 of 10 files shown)


src/TextInput.tsx
```diff
@@ -6,10 +6,13 @@ import TextInputWrapper from './_TextInputWrapper'
 
 type NonPassthroughProps = {
   className?: string
+  // deprecate icon prop
   icon?: React.ComponentType<{className?: string}>
```
colebemis, line 10: I think using a JSDoc comment here will give us some nice in-editor warnings for free.
```suggestion
  /** @deprecated Use `leadingIcon` prop instead */
  icon?: React.ComponentType<{className?: string}>
```

src/TextInput.tsx
```diff
@@ -20,16 +23,22 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
...
       icon: IconComponent,
+      leadingIcon: LeadingIconComponent,
+      trailingIcon: TrailingIconComponent,
       block,
       className,
       contrast,
       disabled,
+      status,
       sx: sxProp,
       theme,
```
colebemis, line 34: We should also deprecated the `theme` prop

src/TextInput.tsx
```diff
@@ -41,18 +50,22 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
       <TextInputWrapper
         block={block}
         className={wrapperClasses}
+        status={status}
```
colebemis, line 53: Should this be called `validationStatus` to be consistent with @rezrah's new checkbox and radio components?

colebemis (commented):
What do you think of having a composable API for leading and trailing visuals similar to action list? Maybe something like this:

```jsx
<InputGroup>
  <InputGroup.LeadingVisual><SearchIcon /></InputGroup.LeadingVisual>
  <TextInput placeholder="Search" />
</InputGroup>
```

This is similar to Chakra UI's API: https://chakra-ui.com/docs/form/input#add-elements-inside-input

cc @siddharthkp

src/TextInput.tsx
pksjce: Nice will do that!

src/TextInput.tsx
pksjce: Got it!

pksjce:
Yep! Definitely looks like the composite component is our way forward seeing all the other component implementations too. My only qualm is that it gets verbose when composed at higher levels. If its a pattern we'll follow then our users should probably get used to it. 
However, in this case I'm only fixing up current issues with `TextInput`. Do you think we want to change this api for beta?

siddharthkp:
> What do you think of having a composable API for leading and trailing visuals similar to action list?

```
<InputGroup>
  <InputGroup.LeadingVisual><SearchIcon /></InputGroup.LeadingVisual>
  <TextInput placeholder="Search" />
</InputGroup>
```

What's the benefit of using that API here?

> Yep! Definitely looks like the composite component is our way forward seeing all the other component implementations too. My only qualm is that it gets verbose when composed at higher levels.

I agree with @pksjce here, the additional verbosity doesn't seem to give us any benefits here.

I don't think composite components should not be our go-to API choice. They are useful when we're composing application components within primer components with. In TextInput, we tightly control "how" the Icon renders (size, color, spacing, etc) so `<TextInput leadingVisual={SearchIcon}` makes a lot of sense to me.

src/TextInput.tsx
```diff
@@ -6,10 +6,13 @@ import TextInputWrapper from './_TextInputWrapper'
 
 type NonPassthroughProps = {
   className?: string
+  /** @deprecated Use `leadingIcon` or `trailingIcon` prop instead */
   icon?: React.ComponentType<{className?: string}>
+  leadingIcon?: React.ComponentType<{className?: string}>
+  trailingIcon?: React.ComponentType<{className?: string}>
```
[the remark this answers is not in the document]
mperrotti: Another thing to consider for this space: we might want to render a button inside the input. For example: in the project board search, we show a button to clear the input.

Screen Shot 2021-11-30 at 2 58 18 PM

src/TextInput.tsx
```diff
@@ -41,18 +49,21 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
       <TextInputWrapper
         block={block}
         className={wrapperClasses}
+        validationStatus={validationStatus}
         contrast={contrast}
         disabled={disabled}
-        hasIcon={!!IconComponent}
+        hasIcon={!!IconComponent || !!(LeadingIconComponent || TrailingIconComponent)}
```
mperrotti, line 55: Non-blocking nitpick: I prefer `Boolean(isTruthy)` over `!!isTruthy` just because I find it more readable.

src/_TextInputWrapper.tsx
```diff
@@ -25,45 +43,50 @@ type StyledWrapperProps = {
   hasIcon?: boolean
   block?: boolean
   contrast?: boolean
+  validationStatus?: 'error' | 'warning'
```
mperrotti, line 46: Heads up: I defined this as a type in utils/types in one of my PRs: https://github.com/primer/react/pull/1611/files#diff-410797b1c4259d74820358a792a62f83cfc7a1c0a0eb26226c89e4621b88d42cR1

mperrotti: I didn't include `'warning'` because I couldn't figure out where we use it, but I'm fine with keeping it.

src/_TextInputWrapper.tsx
```diff
@@ -25,45 +43,50 @@ type StyledWrapperProps = {
...
   background-repeat: no-repeat; // Repeat and position set for form states (success, error, etc)
   background-position: right 8px center; // For form validation. This keeps images 8px from right and centered vertically.
   border: 1px solid ${get('colors.border.default')};
   border-radius: ${get('radii.2')};
   outline: none;
   box-shadow: ${get('shadows.primer.shadow.inset')};
   cursor: text;
+  padding: 6px 12px;
+  display: grid;
+  grid-template-areas: 'leadingIcon input trailingIcon';
```
mperrotti, line 72: We should add `grid-template-columns: auto 1fr auto;` so that the empty grid cells collapse when no content is passed.

Stuff like this will happen if we don't:
Screen Shot 2021-11-30 at 3 29 10 PM

src/_TextInputWrapper.tsx
pksjce: Does that mean I have to handle `success` case in `TextInput`?

src/TextInput.tsx
```diff
@@ -6,10 +6,13 @@ import TextInputWrapper from './_TextInputWrapper'
 
 type NonPassthroughProps = {
   className?: string
+  /** @deprecated Use `leadingIcon` or `trailingIcon` prop instead */
   icon?: React.ComponentType<{className?: string}>
+  leadingIcon?: React.ComponentType<{className?: string}>
```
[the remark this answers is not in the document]
pksjce: Ah I see. So there's text and icons we have to support.

The term leadingVisual is so heavy for me too. I'd never heard of such a prop before using primer. I was used to `before` and `after` props. I think we need to decide whats a standard prop name we'll use for what we call `leadingVisual` and `trailingVisual`. I used it in the new button and its being used in Item components too I think. Using `prefix`/`suffix` would make TextInput the odd one out.

src/TextInput.tsx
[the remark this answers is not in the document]
pksjce: Will update to trailingVisual property for the first comment.

Yes, I see in this 
image
from https://github.com/github/primer/issues/156
I was wondering if your work in https://github.com/primer/react/pull/1611/files would cover TextInput with actions. I'm still reviewing your PR to find that out.

src/_TextInputWrapper.tsx
pksjce: Yeah, I've been trying to fix this as well.
But your solution results in 
image

pksjce: Adding 
```
grid-template-columns: auto 1fr auto;
justify-items: end;
```
seems to help. WDYT?

src/TextInput.tsx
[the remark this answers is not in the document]
mperrotti: You're right - we should just keep with the current standard: `leadingVisual`/`trailingVisual`

src/TextInput.tsx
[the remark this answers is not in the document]
mperrotti: No, my work in https://github.com/primer/react/pull/1611 doesn't touch anything inside the inputs. It just handles the layout and passing the correct attributes for screenreader accessibility (e.g.: matching the label's `for` attribute to the input's `id` attribute)

src/_TextInputWrapper.tsx
mperrotti: We don't have a special style for inputs with a validation status of `success`. We just style any success message we might show.

For example:
Screen Shot 2021-12-02 at 8 35 57 AM

src/_TextInputWrapper.tsx
mperrotti: It looks like the text input is getting sized with `auto` when it's the first child. We might need to change the `grid-template-columns` based on what children are there.

#### `leadingVisual` + input + `trailingVisual`
`grid-template-columns: auto 1fr auto;`

#### `leadingVisual` + input
`grid-template-columns: auto 1fr;`

#### input + `trailingVisual`
`grid-template-columns: 1fr auto;`

#### input
`grid-template-columns: 1fr;`
or
`grid-template-columns: unset;`

mperrotti: I was hoping grid would have some fancy way of handling this without having to specify different column templates for each case, but I couldn't figure it out :(

Maybe another reviewer will know?

mperrotti: Or maybe we could just use flexbox instead of grid? Is there something we need from grid that can't be done with flexbox?

src/_TextInputWrapper.tsx
```diff
@@ -25,39 +43,47 @@ type StyledWrapperProps = {
...
   vertical-align: middle;
+  background-color: ${get('colors.canvas.default')};
   background-repeat: no-repeat; // Repeat and position set for form states (success, error, etc)
   background-position: right 8px center; // For form validation. This keeps images 8px from right and centered vertically.
   border: 1px solid ${get('colors.border.default')};
   border-radius: ${get('radii.2')};
   outline: none;
   box-shadow: ${get('shadows.primer.shadow.inset')};
   cursor: text;
+  padding: 6px 12px;
```
vdepizzol, line 70: i'm afraid there's too much horizontal padding for a text input in here. We should aim for more compact UIs where possible 😇.

> Screen Shot 2021-12-06 at 15 22 06


```suggestion
  padding: 6px 8px;
```

vdepizzol: Also, because this padding is being set in the wrapper, and not in the `<input>` itself, the surroundings of the input element aren't interactive.

https://user-images.githubusercontent.com/293280/144939127-644cab4b-d115-4cf5-92f1-286ef13d5bcd.mov

We should make sure the `<input>` area covers as much room as possible (saving for leading and trailing elements) within the wrapper.

vdepizzol: One last thing :)

The vertical padding of `6px` is ignoring the border, resulting in an input field height of 34px[^1]. We should make sure the component height as a whole matches 32px.

🙇 

[^1]: 20px of line height, 6px+6px of vertical padding, 1px+1px of border

src/TextInput.tsx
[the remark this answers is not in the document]
pksjce: Hmmm then I think we should discuss this in a separate PR. It's common to have `trailingVisual` and `trailingAction` in our components. Possibly we can think of adding `trailingAction` onto the TextInput?

src/TextInput.tsx
pksjce: Removed this check. It was not required after all

src/_TextInputWrapper.tsx
pksjce: Isn't it unnecessary for me to use that type then since it has `success` in it too.

src/_TextInputWrapper.tsx
pksjce: Yeah I'm not particularly attached to the grid way. Do you want to pair on this for a bit tomorrow?

src/TextInput.tsx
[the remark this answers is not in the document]
mperrotti: I'm happy with addressing this in a separate PR. Agreed - `trailingAction` should be on the TextInput

src/_TextInputWrapper.tsx
mperrotti: Sounds great! I love a CSS pairing sesh :)

mperrotti (commented):
Everything is looking good! After @pksjce and I pair on the layout of the leading/trailing visuals and the input, this should be ready to merge.

src/_TextInputWrapper.tsx
vdepizzol: > The height on that component is a minimum of 34px and the padding is the same as this component.

We should probably fix it in there too. We should make sure the form components match sizes with buttons at a 32px height for the default size variant. c/c @mperrotti

mperrotti: @vdepizzol - I'm going to take care of this in a follow-up PR. It's not a trivial amount of work to add to these changes.

mperrotti (approved):
Add a changeset, and then it's good to go!

PR #1669, opened 2021-11-30
Rename "component statuses" to "component status"

colebemis:
From @yaili:

> can we rename the page to “Component status” - I think the plural of status can be both, and it looks nicer :smile:


siddharthkp (approved):

siddharthkp:
lol suspicious - why does the size limit change here on a docs update 😅 

57.31 KB (-0.12% 🔽)

colebemis:
> why does the size limit change here on a docs update 😅

looks like merging main fixed this 😅

PR #1671, opened 2021-11-30
Introduces a PropsTable docs component

jfuchs:
Describe your changes here.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


docs/content/TextInput.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: TextInput
+status: Alpha
+---
+
+import {PropsTable} from '../src/props-table'
```
colebemis, line 6: You can make PropsTable globally available in markdown files by exporting it here: https://github.com/primer/react/blob/main/docs/src/%40primer/gatsby-theme-doctocat/mdx-components.js

docs/content/TextInput.mdx
```diff
@@ -0,0 +1,61 @@
...
+<PropsTable>
+  <PropsTable.Row name="aria-label" type="string" description="Required. Allows input to be accessible." />
+  <PropsTable.Row name="block" type="boolean" description="Adds `display: block` to element" />
+  <PropsTable.Row name="contrast" type="boolean" description="Changes background color to a higher contrast color" />
+  <PropsTable.Row
+    name="variant"
+    type="string"
+    description="Can be either `small` or `large`. Creates a smaller or larger input than the default."
+  />
+  <PropsTable.Row name="width" type="string | number" description="Set the width of the input" />
```
colebemis, line 44: I wonder if we should accept an array for enum types. Example:

```
type={['string', 'number']}
```

Requiring people to manually type `|` could introduce some inconsistency. For example people might write, `string or number`, or `string, number`, instead of `string | number`

docs/content/TextInput.mdx
jfuchs: It looks like when I try that, I get this error:

CleanShot 2021-11-30 at 14 52 53@2x

docs/content/TextInput.mdx
```diff
@@ -0,0 +1,61 @@
...
+  <PropsTable.Row name="maxWidth" description="Set the maximum width of the input">
+    <PropsTable.Row.Type>
+      string | number | <Link href="https://styled-system.com/guides/array-props">Array</Link>
+    </PropsTable.Row.Type>
+  </PropsTable.Row>
```
colebemis, line 49: It feels a bit odd to me to have two ways of defining the type. I'd vote for a single `type` prop that accepts a string or a JSX element:

```
<PropsTable.Row ... type={<>string | number | <Link href="https://styled-system.com/guides/array-props">Array</Link></>} ...
```

docs/content/TextInput.mdx
colebemis: I don't feel too strongly about this. Just an idea. Happy to ship what you have here and see if we end up with inconsistencies

docs/content/TextInput.mdx
colebemis: I think we can use the same batter for description too

jfuchs: Yeah, I was experimenting a little bit here with the `createSlots` API, but something more straightforward should be fine.

docs/content/TextInput.mdx
jfuchs: I like the mindset of reducing inconsistency (ultimately I want us to be able to do that through generating the docs). What would you expect to happen when one of the types needs to be linked?

colebemis: I was thinking elements in the array could be strings of JSX elements:
```
type={['string', <Link href='#>number</Link>]}
```

jfuchs: Gotcha, thanks. I think I'm going to hold off on that for now — in part because `|` isn't even the only type operator that could apply there and I want to think about that, and in part because this still feels like forward progress.

docs/src/props-table.js
```diff
@@ -0,0 +1,84 @@
...
+      <tbody>{children}</tbody>
+    </Table>
+  )
+}
+
+function Row({name, type, defaultValue, description, required}) {
+  return (
+    <tr>
+      <Box as="td" fontFamily="mono" fontSize={1} fontWeight="bold" sx={{whiteSpace: 'nowrap'}}>
+        {typeof name === 'function' ? name() : name}
```
colebemis, line 33: When would `name` be a function?

docs/src/props-table.js
```diff
@@ -0,0 +1,84 @@
...
+  )
+}
+
+function AsRow({defaultValue}) {
+  return (
+    <Row
+      name="as"
+      defaultValue={defaultValue}
+      type="ReactElement"
+      description="The underlying element to render — either a DOM element or a React component."
```
colebemis, line 58: Should we add a note here about also accepting all props associated with the value of the as prop?

docs/src/props-table.js
```diff
@@ -0,0 +1,84 @@
...
+    </tr>
+  )
+}
+
+function AsRow({defaultValue}) {
+  return (
+    <Row
+      name="as"
+      defaultValue={defaultValue}
+      type="ReactElement"
```
colebemis, line 57: Is this type correct?

I think the type definition for `as` was something like `keyof JSX.IntrinsicElements | React.ComponentType`

Also, I'm wondering if we should allow people to override the type so they restrict the values that can be passed to `as`

docs/content/Link.mdx
```diff
@@ -0,0 +1,36 @@
...
+<PropsTable>
+  <PropsTable.Row name="href" type="String" description="URL to be used for the Link" />
+  <PropsTable.Row
+    name="muted"
+    type="Boolean"
+    defaultValue="false"
+    description="Uses a less prominent shade for Link color, and the default link shade on hover"
+  />
+  <PropsTable.Row name="underline" type="Boolean" defaultValue="false" description="Adds underline to the Link" />
+  <PropsTable.Row name="hoverColor" type="String" description="Color used when hovering over link" />
```
colebemis, line 33: Nit: In typescript, in `string` and `boolean` are lowercase, right?

```suggestion
  <PropsTable.Row name="href" type="string" description="URL to be used for the Link" />
  <PropsTable.Row
    name="muted"
    type="boolean"
    defaultValue="false"
    description="Uses a less prominent shade for Link color, and the default link shade on hover"
  />
  <PropsTable.Row name="underline" type="boolean" defaultValue="false" description="Adds underline to the Link" />
  <PropsTable.Row name="hoverColor" type="string" description="Color used when hovering over link" />
```

jfuchs: yes! At least, the lowercase types are the ones you want 99% of the time. Good catch!

docs/src/props-table.js
jfuchs: Ahh I think you're right, and there's `React.ElementType`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L73 as shorthand (with some other special stuff we don't need?)

docs/src/props-table.js
jfuchs: I was thinking maybe for an `(all DOM props)` row, we might format that name in that row without the monospaced font.

docs/src/props-table.js
jfuchs: Since sometimes we'll accept DOM props but _not_ an `as` prop, I've started another helper row that explains DOM props, and additionally it will point out that its value changes according to the `as` prop. WDYT?

jfuchs:
I'm experimenting more with what `Link`'s props table could be, since it ticks all the boxes for our common props:

CleanShot 2021-12-01 at 16 57 29@2x

jfuchs:
Another shot at a succinct explanation of DOM/passthrough props:
CleanShot 2021-12-02 at 10 15 21@2x

docs/content/TextInput.mdx
```diff
@@ -0,0 +1,72 @@
...
+  />
+</>
+```
+
+## Component props
+
+Native `<input>` attributes are forwarded to the underlying React `input` component and are not listed below.
+
+<PropsTable>
+  <PropsTable.Row name="aria-label" type="string" description="Required. Allows input to be accessible." />
```
colebemis, line 37:
```suggestion
  <PropsTable.Row name="aria-label" required={true} type="string" description="Allows input to be accessible." />
```

colebemis (approved):
🚢 🚢 🚢

PR #1675, opened 2021-12-01
ActionMenu.Overlay for ActionMenu v2

siddharthkp:
```diff
- <ActionMenu overlayProps={{width: 'medium', onClickOutside: handleClickOutside}}>
+ <ActionMenu>
    
    <ActionMenu.Button>Menu</ActionMenu.Button>

+   <ActionMenu.Overlay width="medium" onClickOutside={handleClickOutside}>
      <TextInput  />
      <ActionList>
        <ActionList.Item>Copy link</ActionList.Item>
        <ActionList.Item>Quote reply</ActionList.Item>
      </ActionList>
+   </ActionMenu.Overlay>
  </ActionMenu>
```

- [ ] Added/updated tests - nope, going to follow up in [branch/siddharth/actionmenu-tests](https://github.com/primer/react/tree/siddharth/actionmenu-tests)
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/ActionMenu2.tsx
```diff
@@ -90,5 +87,9 @@ const MenuButton = React.forwardRef<AnchorRef, ButtonProps>((props, anchorRef) =
   )
 })
 
+/** this component is syntactical sugar 🍭 */
+type MenuOverlayProps = Partial<OverlayProps> & {children: React.ReactElement[] | React.ReactElement}
+const Overlay: React.FC<MenuOverlayProps> = props => <>{props.children}</>
```
colebemis, line 92: Hmm, I think I understand what's going on here, but I wonder if it'll be confusing that...
```
<ActionMenu>
  <ActionMenu.Overlay>
  ...
  </ActionMenu.Overlay>
</ActionMenu>
```

... will give you the same result as... 
```
<ActionMenu>
   ...
</ActionMenu>
```

In Autocomplete, rendering without the Overlay gives you a different result: https://primer.style/react/Autocomplete#example-without-autocompleteoverlay

siddharthkp: Oh yeah, this is a (probably evil) first draft. I want to move most of that Menu logic to MenuOverlay if it's possible.

siddharthkp: Why do we have this as an example?: https://primer.style/react/Autocomplete#example-without-autocompleteoverlay 🤔 

Do we see a use case for this?

colebemis (approved):

Issue #1204, opened 2021-05-03
Make @primer/octicons-react a peer dependency


daaku:
My package uses @primer/octicons-react v13 but this package pulls in v11 resulting in unnecessary increase in my bundle size. Maybe a peer dependency is the right solution here?

rezrah:
To be completed at the same time as https://github.com/github/primer/issues/241

colebemis:
Upgrading to `@primer/octicons-react` `v16.x` should solve this issue ✅

PR #1677, opened 2021-12-01
Make @primer/octicons-react a peer dependency

colebemis:
## Summary
Makes the [`@primer/octicons-react`](https://primer.style/octicons/packages/react) package a peer dependency

## Impact

* Consumers don't have to worry about version mismatch issues. A version mismatch could result in increased bundle size and inconsistent icons
* Consumers can update all the Octicons in their app without having to wait for Primer React to update its version of Octicons.

---

Closes https://github.com/primer/react/issues/1204
Discussed in https://github.com/primer/react/discussions/821

### DIFF
--- a/.changeset/ten-colts-rule.md
+++ b/.changeset/ten-colts-rule.md
@@ -1,0 +1,11 @@
+---
+"@primer/components": major
+---
+
+Make the `@primer/octicons-react` package a peer dependency
+
+To upgrade, run:
+
+```shell
+npm install @primer/octicons-react
+```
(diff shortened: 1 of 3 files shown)


jfuchs (approved):
Should we also include directions in the getting started and the README?

colebemis:
@dusave pointed out that using loose versioning for octicons-react (e.g. `^16.0.0` vs `16.0.0`) should prevent the version mismatch issue ([slack discussion](https://github.slack.com/archives/C01L618AEP9/p1638388148320900?thread_ts=1638385643.318700&cid=C01L618AEP9)).

Seems like this change is probably not needed.

PR #1678, opened 2021-12-01
Propose a new file structure (ADR)

jfuchs:
This PR adds a proposed ADR establishing a file convention.

There are two reasons I'm doing this:
- I think the result would be good (more to come on that in the ADR)
- I'm interested in seeing if we can keep ADRs fast and light for smaller questions

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/contributor-docs/adrs/adr-XXX-file-structure.md
+++ b/contributor-docs/adrs/adr-XXX-file-structure.md
@@ -18,7 +18,7 @@ primer-react/
 │  ├─ Breadcrumbs/
 │  │  ├─ index.ts                    // Just re-exporting?
 │  │  ├─ Breadcrumbs.tsx             // Primary component
-│  │  ├─ BreadcrumbsItem.tsx         // Subcomponent (include item name for findability in most IDEs)
+│  │  ├─ BreadcrumbsItem.tsx         // Subcomponent (include parent component name to increase findability in most IDEs)
 │  │  ├─ Breadcrumbs.mdx             // Documentation. Always .mdx, not .md
 │  │  ├─ Breadcrumbs.stories.tsx
 │  │  ├─ Breadcrumbs.test.tsx        // Unit tests
--- a/contributor-docs/adrs/adr-XXX-file-structure.md
+++ b/contributor-docs/adrs/adr-XXX-file-structure.md
@@ -32,3 +32,6 @@ Questions I would like to settle in this ADR:
   - Should subcomponents always have their own test files? What about type tests and stories?
 - Would this structure help if we wanted to move toward one-package-per-component like react-aria does?
 - Could we set a standard for replacement components? E.g., NewButton vs. Button2
+- Migration
+  - Can we implement this quickly? Can we script the migration?
+  - Can we do this without changing any of our exports from src/index.ts?
--- a/contributor-docs/adrs/adr-XXX-file-structure.md
+++ b/contributor-docs/adrs/adr-XXX-file-structure.md
@@ -6,9 +6,11 @@ Proposed
 
 ## Context
 
-Components might be more grokable if they were structured consistently
+Components might be more grokable if they were structured consistently. This ADR proposes conventions
 
 ## Decision
+
+TL;DR:
 
 ```
 primer-react/
@@ -16,22 +18,25 @@ primer-react/
 │  ├─ BreadCrumbs/
 │  │  ├─ index.ts                    // Just re-exporting?
 │  │  ├─ BreadCrumbs.tsx             // Primary component
-│  │  ├─ BreadCrumbsItem.tsx         // Subcomponent
+│  │  ├─ BreadCrumbsItem.tsx         // Subcomponent (include item name for findability in most IDEs)
 │  │  ├─ BreadCrumbs.mdx             // Documentation. Always .mdx, not .md
 │  │  ├─ BreadCrumbs.stories.tsx
-│  │  ├─ BreadCrumbs.test.tsx
-│  │  └─ BreadCrumbs.types.test.tsx
+│  │  ├─ BreadCrumbs.test.tsx        // Unit tests
+│  │  ├─ BreadCrumbs.types.test.tsx  // Type tests
+│  │  ├─ BreadCrumbs.yml             // Component metadata (Possible future)
+│  │  └─ __snapshots__/
 ┆  ┆
 ```
 
-Questions I would like to settle in this ADR:
+### Rules
 
-- On subcomponents:
-  - Do subcomponents filenames need to be namespaced with the parent component? (e.g., `BreadCrumbsItem.tsx` vs `Item.tsx`)
-  - Should the `index.ts` in each component directory export every subcomponent under its own name, or should we prefer `Object.assign`ing subcomponents to the parent component?
-  - Should subcomponents always have their own test files? What about type tests and stories?
-- Would this structure help if we wanted to move toward one-package-per-component like react-aria does?
-- Could we set a standard for replacement components? E.g., NewButton vs. Button2
-- Migration
-  - Can we implement this quickly? Can we script the migration?
-  - Can we do this without changing any of our exports from src/index.ts?
+- Every component should have its own PascalCased directory directly under `src/`
+- Subcomponents meant to be used as children of their parent component should be properties of the exported component (e.g., `BreadCrumbs.Item`)
+- Subcomponents meant to be used on their own should be exported as a named export (e.g., `ButtonDanger`)
+- Replacements of existing components should use an incrementing number (e.g., `BreadCrumbs2` rather than `NewBreadCrumbs`)
+
+## Implementation
+
+- [ ] Migrate components into this structure
+- [ ] Set up tests for the component structure
+- [ ] Mark this ADR as adopted
--- a/contributor-docs/adrs/adr-XXX-file-structure.md
+++ b/contributor-docs/adrs/adr-XXX-file-structure.md
@@ -15,15 +15,15 @@ TL;DR:
 ```
 primer-react/
 ├─ src/
-│  ├─ BreadCrumbs/
+│  ├─ Breadcrumbs/
 │  │  ├─ index.ts                    // Just re-exporting?
-│  │  ├─ BreadCrumbs.tsx             // Primary component
-│  │  ├─ BreadCrumbsItem.tsx         // Subcomponent (include item name for findability in most IDEs)
-│  │  ├─ BreadCrumbs.mdx             // Documentation. Always .mdx, not .md
-│  │  ├─ BreadCrumbs.stories.tsx
-│  │  ├─ BreadCrumbs.test.tsx        // Unit tests
-│  │  ├─ BreadCrumbs.types.test.tsx  // Type tests
-│  │  ├─ BreadCrumbs.yml             // Component metadata (Possible future)
+│  │  ├─ Breadcrumbs.tsx             // Primary component
+│  │  ├─ BreadcrumbsItem.tsx         // Subcomponent (include item name for findability in most IDEs)
+│  │  ├─ Breadcrumbs.mdx             // Documentation. Always .mdx, not .md
+│  │  ├─ Breadcrumbs.stories.tsx
+│  │  ├─ Breadcrumbs.test.tsx        // Unit tests
+│  │  ├─ Breadcrumbs.types.test.tsx  // Type tests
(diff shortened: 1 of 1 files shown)


contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,34 @@
+# ADR XXX: Isolating behaviors through custom clements and vanilla JavaScript
```
colebemis, line 1:
```suggestion
# ADR XXX: File structure
```

colebemis:
Having a consistent file structure would also lay the groundwork for us to create a `npm run new:component` script to scaffold a new component 😄

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,34 @@
...
+primer-react/
+├─ src/
+│  ├─ BreadCrumbs/
+│  │  ├─ index.ts                    // Just re-exporting?
+│  │  ├─ BreadCrumbs.tsx             // Primary component
+│  │  ├─ BreadCrumbsItem.tsx         // Subcomponent
+│  │  ├─ BreadCrumbs.mdx             // Documentation. Always .mdx, not .md
+│  │  ├─ BreadCrumbs.stories.tsx
+│  │  ├─ BreadCrumbs.test.tsx
+│  │  └─ BreadCrumbs.types.test.tsx
```
pksjce, line 23: @colebemis also talked about possibly adding a yaml file to export things like component status? Though with the mdx file already in there, we may not need it.

colebemis: Looks like Braid Design System does something similar: https://github.com/seek-oss/braid-design-system/tree/master/lib/components/Accordion

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,37 @@
...
+┆  ┆
+```
+
+Questions I would like to settle in this ADR:
+
+- On subcomponents:
+  - Do subcomponents filenames need to be namespaced with the parent component? (e.g., `BreadCrumbsItem.tsx` vs `Item.tsx`)
```
rezrah, line 30: ➕ : Makes searching for related files easier in the IDE

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,37 @@
...
+  - Should the `index.ts` in each component directory export every subcomponent under its own name, or should we prefer `Object.assign`ing subcomponents to the parent component?
```
rezrah, line 31: Could it depend on whether the subcomponent is private to its parent and not intended for composition? I.e. With a compound component, I'd expect it the parent component to be the only one exported from the folder. Might be misunderstanding this point though.

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,37 @@
...
+  - Should subcomponents always have their own test files? What about type tests and stories?
+- Would this structure help if we wanted to move toward one-package-per-component like react-aria does?
```
rezrah, line 33: This approach totally opens doors to this and makes codesplitting much easier to maintain. Big ➕

rezrah:
Pretty hyped for this. It'll improve the experience across the board, for _both_ contributors and consumers. Thanks for proposing it @jfuchs.

contributor-docs/adrs/adr-XXX-file-structure.md
jfuchs: Yes! I think I like the idea of a separate `BreadCrumbs.yml` even though we could use front matter. I imagine it might inform some unit tests and behaviors outside of docs.

contributor-docs/adrs/adr-XXX-file-structure.md
jfuchs: I think I've seen two different styles:
1. `BreadCrumbsItem`, which isn't exported on their own and is only accessed as `BreadCrumbs.Item`
2. `ButtonDanger`, which is available as its own export.
it is reasonable to me that those are different, as `BreadCrumbsItem` only makes sense as a child of `BreadCrumbs`.... and maybe we could name that as a rule? Alternately, we could consolidate to one style, right?

contributor-docs/adrs/adr-XXX-file-structure.md
jfuchs: I think that makes sense — personally I don't know enough about what makes tree shaking possible or not possible to be confident that separate packages are actually necessary.

siddharthkp:
Love it! :shipit:

siddharthkp (approved):
Love it! ship it! :shipit:

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,37 @@
...
+- Could we set a standard for replacement components? E.g., NewButton vs. Button2
```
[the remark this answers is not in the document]
jfuchs: Ohh interesting — can you say more about why you like that?

I hesitate because it'd mean advancing the status of a component from propsed -> alpha would have to be a breaking change, right?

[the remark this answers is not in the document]
siddharthkp: > I hesitate because it'd mean advancing the status of a component from proposed -> alpha would have to be a breaking change, right?

Don't think so. we should use `drafts` for replacement components which can be at a more advanced maturity than their current counterpart. For example, ActionList v1 is alpha but ActionList v2 can be at beta maturity while being in drafts. This is why we chose drafts over other names. More context here: https://github.com/primer/react/pull/1609#issue-1053897027

I like it more because it matches the bundle scope. `drafts` are not part of the main bundle.

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,37 @@
...
+│  │  └─ BreadCrumbs.types.test.tsx
+┆  ┆
+```
+
+Questions I would like to settle in this ADR:
+
+- On subcomponents:
+  - Do subcomponents filenames need to be namespaced with the parent component? (e.g., `BreadCrumbsItem.tsx` vs `Item.tsx`)
+  - Should the `index.ts` in each component directory export every subcomponent under its own name, or should we prefer `Object.assign`ing subcomponents to the parent component?
+  - Should subcomponents always have their own test files? What about type tests and stories?
```
mperrotti, line 32: I don't think so. There are going to be subcomponents that rely on some context to get passed down from the parent.

contributor-docs/adrs/adr-XXX-file-structure.md
siddharthkp: Have to say first - super happy to go with what the majority of the team prefers. very loosely help opinions.

Aw, I was going to vote ➖ on this because it is scoped inside the component folder: `src/Breadcrumb/Item.tsx`.

For example, in VS Code, i would search for "action item" to look for `ActionList/Item`

image

pksjce (approved):

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,42 @@
...
+│  │  ├─ BreadCrumbs.yml             // Component metadata (Possible future)
+│  │  └─ __snapshots__/
+┆  ┆
+```
+
+### Rules
+
+- Every component should have its own PascalCased directory directly under `src/`
+- Subcomponents meant to be used as children of their parent component should be properties of the exported component (e.g., `BreadCrumbs.Item`)
+- Subcomponents meant to be used on their own should be exported as a named export (e.g., `ButtonDanger`)
```
colebemis, line 35: Is this a pattern we want to encourage?

I'd argue that it's not a "subcomponent" if it's exported on its own. If it's not a property of a parent component, I think it should probably have it's own directory.

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,42 @@
...
+│  │  ├─ BreadCrumbs.types.test.tsx  // Type tests
+│  │  ├─ BreadCrumbs.yml             // Component metadata (Possible future)
+│  │  └─ __snapshots__/
+┆  ┆
+```
+
+### Rules
+
+- Every component should have its own PascalCased directory directly under `src/`
+- Subcomponents meant to be used as children of their parent component should be properties of the exported component (e.g., `BreadCrumbs.Item`)
```
colebemis, line 34: Does this mean that the `Object.assign` happens in `Breadcrumbs.tsx` or `index.tsx`?

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,42 @@
...
+Components might be more grokable if they were structured consistently. This ADR proposes conventions
+
+## Decision
+
+TL;DR:
+
+```
+primer-react/
+├─ src/
+│  ├─ BreadCrumbs/
```
colebemis, line 18: Nit: `Breadcrumbs` is one word
```suggestion
│  ├─ Breadcrumbs/
```

https://primer.style/react/Breadcrumbs

contributor-docs/adrs/adr-XXX-file-structure.md
jfuchs: I didn't mean to specify — but if you think this should I'm fine with that!

contributor-docs/adrs/adr-XXX-file-structure.md
jfuchs: I like simple rules.... but thinking about that with `ButtonDanger`.... that doesn't feel like it merits its own component, and I'm not sure I'd want that to just be a prop on Button (i.e., `<Button variant="danger"...`). Would you say `ButtonDanger` should have its own directory?

colebemis: >  I'm not sure I'd want that to just be a prop on Button (i.e., <Button variant="danger"...)

`ButtonDanger` is going away in favor of a `variant` prop. @pksjce's `Button2` uses that API.

contributor-docs/adrs/adr-XXX-file-structure.md
colebemis: We can add more rules to this doc as we form opinions. So I guess it's fine to be vague now.

IMO, the `Object.assign` should happen in `Breadcrumbs.tsx`. `index.tsx` seems mostly like an implementation detail to make imports nicer

contributor-docs/adrs/adr-XXX-file-structure.md
jfuchs: Okay cool. That settles it then :-)

contributor-docs/adrs/adr-XXX-file-structure.md
jfuchs: Should index.tsx export default or named?

colebemis: Hmm. I lean towards named export. In general, I'm not a big fan of default exports (unless they're necessary for some reason)

contributor-docs/adrs/adr-XXX-file-structure.md
```diff
@@ -0,0 +1,41 @@
...
+
+TL;DR:
+
+```
+primer-react/
+├─ src/
+│  ├─ Breadcrumbs/
+│  │  ├─ index.ts                    // Just re-exporting?
+│  │  ├─ Breadcrumbs.tsx             // Primary component
+│  │  ├─ BreadcrumbsItem.tsx         // Subcomponent (include item name for findability in most IDEs)
```
colebemis, line 21: Not sure what "item name" means here
```suggestion
│  │  ├─ BreadcrumbsItem.tsx         // Subcomponent (include parent component name to increase findability in most IDEs)
```

colebemis (approved):

contributor-docs/adrs/adr-XXX-file-structure.md
mperrotti: FWIW I prefer the longer, more descriptive file name.

mperrotti: FWIW, I prefer the longer more descriptive name.

PR #1679, opened 2021-12-02
adds resolved color scheme to theme context

### LINKED ISSUE #1664: useTheme should know which colorscheme is currently applied
When an unknown colorscheme is applied, the ThemeProvider falls back to a base theme, which is nice for ensuring a user sees something reasonable, but may not be the expected theme. 

This behavior is a little hard to test for, since there's no property returned from `useTheme` describing which colorScheme is currently applied in the context.

```jsx
<ThemeProvider {...{colorMode: night, nightScheme: dark_colorblind}}>...</ThemeProvider>
```

for instance, gives a `useTheme` return of 

```js
const theme = useTheme()
=> {
   colorMode: "night"
   colorScheme: "dark_colorblind"
   dayScheme: "light"
   nightScheme: "dark_colorblind"
   resolvedColorMode: "night"
   theme: {}
}
```

but when I pass an invalid colorScheme like

```jsx
<ThemeProvider {...{colorMode: night, nightScheme: dark_protanopia}}>...</ThemeProvider>
```

```js
const theme = useTheme()
=> {
   colorMode: "night"
   colorScheme: "dark_protanopia"
   dayScheme: "light"
   nightScheme: "dark_protanopia"
   resolvedColorMode: "night"
   theme: {}
}
```

This does `console.error`, however since themes can be provided at runtime it might be helpful to be able to determine which theme is currently applied in a given context.

Based on the above, the current way to do that seems like

```js
theme.theme && theme.colorScheme && theme.colorScheme in theme.theme.colorSchemes ? theme.colorScheme : _some other color scheme - the default one _
```

A `resolvedColorScheme` property returned from useTheme, that returns which color scheme is actually active in a context might be a nice way to internalize that check _and_ provide a nice way to validate that the correct them is applied from a consumer.

```jsx
const themeProps = {colorMode: night, nightScheme: dark_protanopia}
<ThemeProvider {...themeProps}>
   <Component />
</ThemeProvider>

const Component = () => {
   const theme = useTheme()
   return <div id="testme" data-test-colorscheme={theme.resolvedColorScheme} />
}

assertEquals(window.testme.dataset.testColorscheme, theme.resolvedColorScheme)
```

mattcosta7:
Describe your changes here.

Closes #1664 

This just adds the 'colorscheme' that was selected in the theme provider to the context

### Screenshots

No visual change

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/red-bottles-prove.md
+++ b/.changeset/red-bottles-prove.md
@@ -1,0 +1,5 @@
+---
+'@primer/components': minor
+---
+
+Add `resolvedColorScheme` property to the object returned by `useTheme()
--- a/.changeset/red-bottles-prove.md
+++ b/.changeset/red-bottles-prove.md
@@ -2,4 +2,4 @@
 '@primer/components': minor
 ---
 
-Add `resolvedColorScheme` property to the object returned by `useTheme()
+Add `resolvedColorScheme` property to the object returned by `useTheme()`
--- a/src/ThemeProvider.tsx
+++ b/src/ThemeProvider.tsx
@@ -24,6 +24,7 @@ const ThemeContext = React.createContext<{
   colorScheme?: string
   colorMode?: ColorModeWithAuto
   resolvedColorMode?: ColorMode
+  resolvedColorScheme?: string
   dayScheme?: string
   nightScheme?: string
   setColorMode: React.Dispatch<React.SetStateAction<ColorModeWithAuto>>
@@ -52,7 +53,10 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
   const systemColorMode = useSystemColorMode()
   const resolvedColorMode = resolveColorMode(colorMode, systemColorMode)
   const colorScheme = chooseColorScheme(resolvedColorMode, dayScheme, nightScheme)
-  const resolvedTheme = React.useMemo(() => applyColorScheme(theme, colorScheme), [theme, colorScheme])
+  const {resolvedTheme, resolvedColorScheme} = React.useMemo(
+    () => applyColorScheme(theme, colorScheme),
+    [theme, colorScheme]
+  )
 
   // Update state if props change
   React.useEffect(() => {
@@ -74,6 +78,7 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
         colorScheme,
         colorMode,
         resolvedColorMode,
+        resolvedColorScheme,
         dayScheme,
         nightScheme,
         setColorMode,
@@ -156,9 +161,15 @@ function chooseColorScheme(colorMode: ColorMode, dayScheme: string, nightScheme:
   }
 }
 
-function applyColorScheme(theme: Theme, colorScheme: string) {
+function applyColorScheme(
+  theme: Theme,
+  colorScheme: string
+): {resolvedTheme: Theme; resolvedColorScheme: string | undefined} {
   if (!theme.colorSchemes) {
-    return theme
+    return {
+      resolvedTheme: theme,
+      resolvedColorScheme: undefined
+    }
   }
 
   if (!theme.colorSchemes[colorScheme]) {
@@ -167,10 +178,16 @@ function applyColorScheme(theme: Theme, colorScheme: string) {
 
     // Apply the first defined color scheme
     const defaultColorScheme = Object.keys(theme.colorSchemes)[0]
-    return deepmerge(theme, theme.colorSchemes[defaultColorScheme])
+    return {
+      resolvedTheme: deepmerge(theme, theme.colorSchemes[defaultColorScheme]),
+      resolvedColorScheme: defaultColorScheme
+    }
   }
 
-  return deepmerge(theme, theme.colorSchemes[colorScheme])
+  return {
+    resolvedTheme: deepmerge(theme, theme.colorSchemes[colorScheme]),
+    resolvedColorScheme: colorScheme
+  }
 }
 
 export default ThemeProvider
--- a/src/__tests__/ThemeProvider.test.tsx
+++ b/src/__tests__/ThemeProvider.test.tsx
@@ -439,3 +439,119 @@ describe('useColorSchemeVar', () => {
     expect(screen.getByText('Hello')).toHaveStyleRule('background-color', 'blue')
   })
 })
+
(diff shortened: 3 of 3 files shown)


src/ThemeProvider.tsx
```diff
@@ -158,7 +163,7 @@ function chooseColorScheme(colorMode: ColorMode, dayScheme: string, nightScheme:
 
 function applyColorScheme(theme: Theme, colorScheme: string) {
```
colebemis, line 164: Should we be explicit about the return type here?
```suggestion
function applyColorScheme(theme: Theme, colorScheme: string): [Theme, string | undefined] {
```

src/ThemeProvider.tsx
```diff
@@ -167,10 +172,10 @@ function applyColorScheme(theme: Theme, colorScheme: string) {
 
     // Apply the first defined color scheme
     const defaultColorScheme = Object.keys(theme.colorSchemes)[0]
-    return deepmerge(theme, theme.colorSchemes[defaultColorScheme])
+    return [deepmerge(theme, theme.colorSchemes[defaultColorScheme]), defaultColorScheme] as const
```
colebemis, line 175: Probably out of scope for this PR, but should we have smarter defaults for color schemes?

For example, we could default `light_foo` to `light` but `dark_foo` to `dark`?

.changeset/polite-colts-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/components': minor
+---
+
+Adds 'resolvedColorScheme' to the ThemeContext'
```
colebemis, line 5:
```suggestion
Add `resolvedColorScheme` property to the object returned by `useTheme()`
```

colebemis:
Something strange is going on with the package-lock.json 🤔  Probably an npm version issue

mattcosta7:
> Something strange is going on with the package-lock.json 🤔 Probably an npm version issue

I saw that too - what version lock/npm are y'all running, I can try to switch

src/ThemeProvider.tsx
mattcosta7: agreed, I think that would be nice for a future tweak here!

colebemis:
I'm running node 16, npm 8

mattcosta7:
> I'm running node 16, npm 8

ahh cool.  the codespace I opened was 14 and 6.   The [nvmrc](https://github.com/primer/react/blob/main/.nvmrc ) says 14.

I'll upgrade and see if that helps

src/ThemeProvider.tsx
```diff
@@ -52,7 +53,10 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
   const systemColorMode = useSystemColorMode()
   const resolvedColorMode = resolveColorMode(colorMode, systemColorMode)
   const colorScheme = chooseColorScheme(resolvedColorMode, dayScheme, nightScheme)
-  const resolvedTheme = React.useMemo(() => applyColorScheme(theme, colorScheme), [theme, colorScheme])
+  const [resolvedTheme, resolvedColorScheme] = React.useMemo(
+    () => applyColorScheme(theme, colorScheme),
+    [theme, colorScheme]
+  )
```
colebemis, line 59: Curious what you're thoughts are about returning an object here instead of an array:

```
const {resolvedTheme, resolvedColorScheme} = ...
// vs
const [resolvedTheme, resolvedColorScheme] = ...
```

mattcosta7: Either works for me, I was keeping the diff small by returning the tuple, but I don't think there's a reason to do that instead of the object

colebemis (approved):
Looks great! Thanks for the PR, @mattcosta7 ❤️

colebemis:
Oh, would you mind adding a test for this in https://github.com/primer/react/blob/main/src/__tests__/ThemeProvider.test.tsx?

mattcosta7:
> Oh, would you mind adding a test for this in https://github.com/primer/react/blob/main/src/__tests__/ThemeProvider.test.tsx?

Sure thing, I didn't see one for resolvedColorMode, so I let it slide, but will add. 

I think we should also probably read the outer scope, an pass down the resolvedColorScheme to children themes too, which I believe won't get that returned from applyColorSchemes

colebemis:
> Sure thing, I didn't see one for resolvedColorMode, so I let it slide, but will add.

Just noticed that too. Oops 😅 

> I think we should also probably read the outer scope, an pass down the resolvedColorScheme to children themes too, which I believe won't get that returned from applyColorSchemes

Not sure I understand. Could you elaborate? I think the logic for nested themes still works with this change.

mattcosta7:
> > I think we should also probably read the outer scope, an pass down the resolvedColorScheme to children themes too, which I believe won't get that returned from applyColorSchemes
> 
> Not sure I understand. Could you elaborate? I think the logic for nested themes still works with this change.

Ahh nevermind on this.  I misread the merge, and thought we wouldn't have the 'colorSchemes' object for them, but we do, so all good :-D

mattcosta7:
> Oh, would you mind adding a test for this in https://github.com/primer/react/blob/main/src/__tests__/ThemeProvider.test.tsx?

Done.   Added a handful of cases

src/__tests__/ThemeProvider.test.tsx
```diff
@@ -439,3 +439,119 @@ describe('useColorSchemeVar', () => {
...
+  })
+
+  it('is the value of the fallback colorScheme applied when attempting to apply an invalid colorScheme', () => {
+    const Component = () => {
+      const {resolvedColorScheme} = useTheme()
+
+      return <Text data-testid="text">{resolvedColorScheme}</Text>
+    }
+
+    const schemeToApply = 'totally-invalid-colorscheme'
```
colebemis, line 498: 😆

.changeset/red-bottles-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/components': minor
+---
+
+Add `resolvedColorScheme` property to the object returned by `useTheme()
```
colebemis, line 5:
```suggestion
Add `resolvedColorScheme` property to the object returned by `useTheme()`
```

PR #1681, opened 2021-12-02
Update to node 16

colebemis:
Using older versions of node (<16) changes the structure of the package-lock.json in this repo.


rezrah (approved):
Looks good. Can we also lock it with [package.json engines](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines) to enforce it @colebemis?

Specifically the `npm` version, where we'll need >7 to avoid the package-lock.json issues.

PR #1691, opened 2021-12-03
Use reusable release_canary workflow

colebemis:
Moved `release_canary.yml` into a [reusable workflow](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows) that can be shared across the Primer org

https://github.com/primer/.github/blob/main/.github/workflows/release_canary.yml


dgreif (approved):
Looks great!  Awesome use of the new shared workflows!

jfuchs (approved):

PR #1695, opened 2021-12-04
Update component docs

colebemis:


docs/src/props-table.js
```diff
@@ -6,6 +6,12 @@ import InlineCode from '@primer/gatsby-theme-doctocat/src/components/inline-code
 function PropsTable({children}) {
   return (
     <Table>
+      <colgroup>
```
jfuchs, line 9: TIL about `colgroup`. Nice.

docs/src/props-table.js
```diff
@@ -27,15 +33,28 @@ function PropsTable({children}) {
   )
 }
 
-function Row({name, type, defaultValue, description, required}) {
+function Row({name, type, defaultValue, description, required, deprecated}) {
```
jfuchs, line 36: Doesn't look like this PR adds usage of `deprecated` — that's fine by me; I'm just calling it out in case that wasn't intended.

docs/src/props-table.js
```diff
@@ -70,8 +89,8 @@ function BasePropRows({passthroughPropsLink, elementType, isPolymorphic, refType
 function PassthroughPropsRow({elementName, isPolymorphic, passthroughPropsLink}) {
   return (
     <tr>
-      <Box as="td" colSpan={4} fontSize={1} verticalAlign="top" fontStyle="italic">
-        Additional props are passed through to the <InlineCode>&lt;{elementName}&gt;</InlineCode> element. See{' '}
+      <Box as="td" colSpan={4} fontSize={1} verticalAlign="top">
+        Additional props are passed to the <InlineCode>&lt;{elementName}&gt;</InlineCode> element. See{' '}
```
jfuchs, line 93: What do you think about adding something like a ⚠️ to call out this row?

docs/content/drafts/ActionList2.mdx
```diff
@@ -373,3 +472,13 @@ render(<SelectFields />)
...
   }}
 />
+
+## Further reading
+
+- [Interface guidelines: Action List](https://primer.style/design/components/action-list)
+
+## Related components
+
+- [ActionMenu](/drafts/ActionMenu2)
```
jfuchs, line 482: Should we call this "ActionMenu2" in the link text?

docs/content/drafts/ActionList2.mdx
```diff
@@ -264,97 +251,208 @@ const SelectFields = () => {
...
-
-## Related components
-
-- [ActionMenu](/drafts/ActionMenu2)
-- [DropdownMenu](/DropdownMenu)
-- [SelectPanel](/SelectPanel)
-
-## Component status
+<PropsTable>
+  <PropsTableRow name="children" required={true} type="ActionList.Item[] | ActionList.LinkItem[]" />
```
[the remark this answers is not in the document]
jfuchs: I wonder if we should also have a `PropsTableChildrenRow` — perhaps with overrides for when components expect a more limited set of nodes?

Something like:

```
<PropsTableChildrenRow acceptableChildElements={[ActionList2.Item, ActionList2.LinkItem, ActionList2.Group]} />
```

?

jfuchs (approved):
I'm liking this 😻

docs/src/props-table.js
colebemis: Yup. I don't think we have many deprecated props. But it felt like something the `Row` component should support

docs/src/props-table.js
colebemis: Hmm, not sure. ⚠️ makes me think is a "bad" thing but I'm not sure it is 🤷

docs/content/drafts/ActionList2.mdx
colebemis: I'll leave that up to @siddharthkp

docs/src/props-table.js
jfuchs: I think that's fair. 👍🏻

PR #1703, opened 2021-12-07
ADR 004: Strict props or Composite components

siddharthkp:
See [rendered version](https://github.com/primer/react/blob/siddharth/adr-children-api/contributor-docs/adrs/adr-004-children-as-api.md)

Closes https://github.com/github/primer/issues/490

Would love to get feedback from everyone in @primer/react-reviewers


pksjce:
Really love this! Kudos for putting this together. Feel very validated with the Button approach being used as a case study.

While I agree with what the ADR has to say, it's not clear to me the goal or the conclusion of this ADR. I feel like you want to talk about when to use Composable API vs when to use props vs when to use `children` as API. But don't conclude on concrete rules at the end of the ADR. WDYT?

siddharthkp:
> I feel like you want to talk about when to use Composable API vs when to use props vs when to use children as API. But don't conclude on concrete rules at the end of the ADR. WDYT?

Yes! I don't think there are any hand wavy rules here :( It really depends on the component and what components its often used with. My goal here is to help pick a good starting point, instead of a definitive flowchart. Thoughts?

pksjce (approved):
Love this breakdown of api decisions. What I'm taking away is
1. Regardless of API decision, have consistent naming
2. Deliberate over each prop and composite component you need to add to a component. Make examples of both usages and try to overlap with all known usecases.
3. Go for a conservative API. Its easy to open an API for future usecases than close it down because of misuse.
4. Multiple props for a single functionality, object data inside a single prop , parent level props are smells that point to reconsidering API.

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,405 @@
...
+  </Flash>
+)
+```
+
+<br/>
+
+### Pros of this approach here:
+
+1. The component is aware of recommended use cases and comes with those design decisions backed-in. For example, using an `Icon` with `Flash` is a recognised use case. We don’t lock-in a specific icons, but we do set the size, variant-compatible color and the margin between the icon and text.
+
```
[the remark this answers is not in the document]
siddharthkp: Include styles for `svg` in `Flash` - https://github.com/primer/react/blob/main/src/Flash.tsx#L8-L15

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,405 @@
...
+<br/>
+
+You can see this pattern used in `NewButton`:
+
+The icon gets its color and margin based on the variant and size of the `Button`. This is the happy path we want folks to be on, so we ask for the icon component instead of asking the developer to render the icon.
+
+<img width="141" alt="image 9" src="https://user-images.githubusercontent.com/1863771/144945219-a853ed1c-f21d-412e-a388-6d74ec436645.png">
+
+```jsx
```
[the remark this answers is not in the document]
siddharthkp: Ouch, was supposed to be NewButton. Fixed!

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,405 @@
...
+
+_Note: These are in no way rules, these should be treated as a starting point for API conversations._
+
+_Note #2: Consumer is used multiple times on this page. It refers to the developers consuming the component API and not end users._
+
+<br/>
+
+## Prefer using children for “content”
+
+With React, `children` is the out-of-the-box way for putting _content_ inside your component. By using `children` instead our own custom prop, we can make the API “predictable” for its consumers.
```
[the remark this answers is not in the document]
siddharthkp: thanks! fixed

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+
+## Status
+
+Proposed
+
+<br/>
+
+_Note: These are in no way rules, these should be treated as a starting point for API conversations._
+
+_Note #2: Consumer is used multiple times on this page. It refers to the developers consuming the component API and not end users._
```
mperrotti, line 11: Good choice for clarity 🙌

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+
+<br/>
+
+## Prefer using children for “content”
+
+With React, `children` is the out-of-the-box way for putting _content_ inside your component. By using `children` instead of our own custom prop, we can make the API “predictable” for its consumers.
```
mperrotti, line 17: > ...for putting _content_ inside...

Do we want to be more specific about what we mean by "content"?

> ...for putting [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content) inside...

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+<Flash variant="success">
+  <CheckIcon /> Changes saved!
+</Flash>
+```
+
+_Note: We need to assume good intent here, developers using the components aren’t trying to break the system. They are either trying to implement something that the system’s happy path does not support OR there are multiple ways of doing something with primer and they have unintentionally picked the approach that is not recommended._
+
+<br/>
+
+The general wisdom is to _Make the right (recommended) thing easy to do and the wrong (not recommended) hard to do_. When going off the happy path, developers should feel some friction, some weight, code that “feels hacky” or feels like a workaround.
```
mperrotti, line 105: 💯  😍

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+```jsx
+<Flash variant="success">
+  <Flash.Icon icon={CheckIcon} variant="filled" />
+  Changes saved!
+</Flash>
+```
+
+<br/>
+
+The `Parent.Child` syntax signals that this component is tied to the parent.
```
mperrotti, line 194: Do we have composite components with nesting deeper than 2 levels? It might be good to have a recommendation for naming deeply nested children.

For example:
```
<ChoiceFieldset>
  <ChoiceFieldset.List>
    <ChoiceFieldset.Item value="labelOne">Label one</ChoiceFieldset.Item>
    <ChoiceFieldset.Item value="labelTwo">Label two</ChoiceFieldset.Item>
  </ChoiceFieldset.List>
</ChoiceFieldset
```

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+
+We can look at `Flash.Icon` as a stricter version of `Icon` that automatically works with different variants of `Flash`. It does not need to support all the props of `Icon` like size or color, only the ones that are compatible in the context of a `Flash`.
+
+It can be tempting to "future proof" our API and adopt this pattern early, but we should resist that until the use case presents itself. _It is always easier to open up the API later than to close it down._
```
mperrotti, line 198: > _It is always easier to open up the API later than to close it down._

Preach! 🙌

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+
+---
+
+<br/>
+
+### 2. Exposing customisation options for internal components:
+
+Another place where composite patterns lead to aesthetic predictable APIs is when we want to expose customisation options for internal components.
+
+For Example, `ActionMenu` accepts `overlayProps` and `anchorContent` to pass it down to the implementation details:
```
mperrotti, line 242: I don't think "Example" needs to be capitalized

mperrotti (commented):
This is a very clear and detailed explanation of the direction we want to take component APIs.

Should we try and bring some of the high-level ideas to PVC? For example: "make the easy thing easy and the hard thing hard", "its easier to open up the API later than to close it down

contributor-docs/adrs/adr-004-children-as-api.md
siddharthkp: ooh TIL phrasing content!

contributor-docs/adrs/adr-004-children-as-api.md
siddharthkp: we do! 

```jsx
<ActionList>
  <ActionList.Group>
      <ActionList.Item>
        <ActionList.LeadingVisual />
      </ActionList.Item>
  </ActionList.Group>
</ActionList>
```

the recommendation would be to attach components on the root level component and not middle ones, right?
`ActionList.LeadingVisual` and not `ActionList.Item.LeadingVisual`

siddharthkp:
Oh no bot, please don't

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+     </span>
+     <Button variant="invisible" icon={CheckIcon} aria-label="Hide flash message" onClick={onDismiss} />
+   </Flash>
+   ```
+
+<br/>
+
+### Tradeoffs of this approach / When not to use children
+
+Our goal isn't to put all content inside children though. Composition offers flexibility to the consumer of the component, this flexibility, however, can also lead to inconsistencies.
```
lesliecdubs, line 86:
```suggestion
Our goal isn't to put all content inside children though. Composition offers flexibility to the consumer of the component; this flexibility, however, can also lead to inconsistencies.
```

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+  // oh oh, we don't know if that color or icon size is the right choice!
+  <CheckIcon size={20} color="success.emphasis" /> Changes saved!
+</Flash>
+
+<Flash variant="success">
+  <CheckIcon /> Changes saved!
+</Flash>
+```
+
+_Note: We need to assume good intent here, developers using the components aren’t trying to break the system. They are either trying to implement something that the system’s happy path does not support OR there are multiple ways of doing something with primer and they have unintentionally picked the approach that is not recommended._
```
lesliecdubs, line 101:
```suggestion
_Note: We need to assume good intent here, developers using the components aren’t trying to break the system. They are either trying to implement something that the system’s happy path does not support OR there are multiple ways of doing something with Primer and they have unintentionally picked the approach that is not recommended._
```

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+
+```jsx
+<Flash variant="success" icon={CheckIcon}>
+  Changes saved!
+</Flash>
+```
+
+We are still using `children` for text content, but we have moved the `icon` back as a prop with reduced flexibility.
+
+When intentionally going off the happy path, developers can still drop down an abstraction level add an `Icon` to `children`, they would have to pick up the additional effort of setting compatible color, size and margin themselves. However, when it’s unintentional, it would feel like way too much work that the component should be doing automatically.
```
lesliecdubs, line 117:
```suggestion
When intentionally going off the happy path, developers can still drop down an abstraction level to add an `Icon` to `children`, though they would have to pick up the additional effort of setting compatible color, size and margin themselves. However, when it’s unintentional, it would feel like way too much work that the component should be doing automatically.
```

Feel free to edit my suggestion as needed if I didn't hit the right tone, just wanted to try to clear up some confusion from the wording.

lesliecdubs:

**ADR decision maker:** @colebemis 
**Decision deadline:** Thursday, 5 May

colebemis (approved):
As the ADR decision-maker, I'm approving this ADR ✅  The whole team seems aligned on this API direction.

These guidelines will serve as a great resource for making consistent and usable API decisions. Thank you, @siddharthkp!

A few minor edits I'd love to see before this is merged:
* A brief "Context" section at the beginning of the ADR (see the [TypeScript ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-001-typescript.md#context) as an example)
* Remove references to `v2` and `New` components, as they have now officially replaced the legacy components

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
...
+Proposed
+
+<br/>
+
+_Note: These are in no way rules, these should be treated as a starting point for API conversations._
+
+_Note #2: Consumer is used multiple times on this page. It refers to the developers consuming the component API and not end users._
+
+<br/>
+
```
colebemis, line 14: Could we add a "Decision" heading here for consistency with other ADRs?

lesliecdubs:
Friendly reminder that we're a week past the deadline on this one. @siddharthkp let's try to get this updated and merged this week 🙏🏻

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,416 @@
...
+_Note: Consumer is used multiple times on this page. It refers to the developers consuming the component API and not end users._
+
+<br/>
+
+## Decision:
+
+
+1. Prefer using children for “content”
```
siddharthkp, line 16: Should this be its own ADR instead of being in this one?

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -0,0 +1,403 @@
+# ADR 004: Strict props or Composite components
+
+## Status
+
+Proposed
+
+<br/>
```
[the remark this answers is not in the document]
colebemis: Looks good to me 👍

contributor-docs/adrs/adr-004-children-as-api.md
colebemis: I think it's fine to include it in this one 🤷

PR #1705, opened 2021-12-08
Move deprecated components to a folder

siddharthkp:
Where does this show up:

1. In search, the folder is showed in the search result, which would point people to the right component

image

2. When you're in the component, in additional to the `Deprecated` label next to the title, it shows up in the URL as well.

image

### Merge checklist


colebemis:
Should we also move them into the deprecated section of the nav?
CleanShot 2021-12-08 at 08 35 47

siddharthkp:
@colebemis Yes! I forgot about those! my bad, fixed now!

colebemis (approved):
Thanks for cleaning this up, @siddharthkp! 

Maybe not necessary for this PR, but we should link to the replacement components on deprecated component pages (e.g. `Dialog` should link to `Dialog v2`)

siddharthkp:
> Maybe not necessary for this PR, but we should link to the replacement components on deprecated component pages (e.g. Dialog should link to Dialog v2)

Great idea, will create another PR

Issue #1701, opened 2021-12-06
Update component docs to use `PropsTable` and `ComponentChecklist`


colebemis:
## Problems

* Component props are not documented consistently across pages on https://primer.style/react
* Not all component documentation pages include a [component status checklist](https://github.com/primer/react/pull/1626)

## Goals

* Use the new [`PropsTable`](https://github.com/primer/react/pull/1671) component, to render prop documentation consistently across all component pages on https://primer.style/react
* The props table should not reference internal types. For example, if you have a string union called `Variants`, you should show each string (`'danger' | 'success' | 'warning'`)
* Add a component status checklist to every component page

### Consistency nice-to-haves ✨ 
- Every component page should have `source` frontmatter that links to the source code for that component ([example](https://github.com/primer/react/blob/main/docs/content/Box.mdx))
- Every component page should have an `Examples`, `Props`, and `Status` heading (in that order)
- The top of every component page should show how to import the component ([example](https://primer.style/react/Box))
  ```
  import {Box} from '@primer/react'
  ```

## Non-goals

The props API of many of our components could be improved. However, changing the props of any component is out of scope for this issue.

## Example

An example of how to use `PropsTable` and `ComponentChecklist` in an `.mdx` file:

```mdx
---
title: ExampleComponent
status: Alpha
---

...

## Props

<PropsTable>
  {/* normal prop */}
  <PropTableRow
    name="variant"
    type="'default' | 'danger'"
    defaultValue="'inset'"
    description={<>...use a string or JSX to describe this prop...</>}
  />

  {/* required prop */}
  <PropTableRow
    name="children"
    required={true}
    type="React.ReactNode"
    description="..."
  />

  {/* deprecated prop */}
  <PropTableRow
    name="scheme"
    deprecated
    type="'default' | 'danger'"
    defaultValue="'inset'"
    description={<>Use <InlineCode>variant</InlineCode> instead</>}
  />

  {/* sx prop */}
  <PropsTableSxRow />

  {/* as prop */}
  <PropsTablePassthroughProps
    defaultElementType="a"
  />

  {/* ref prop */}
  <PropsTablePassthroughProps
    elementType="a"
    isPolymorphic={true}
    refType="HTMLAnchorElement"
  />

  {/* passthrough props */}
  <PropsTablePassthroughProps
    elementType="a"
    isPolymorphic={true}
    passthroughPropsLink={
[...]

PR #1710, opened 2021-12-09
Adding datepicker API doc

dusave:
Describe your changes here.

Adds the Datepicker API doc

### Screenshots

Base

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/Datepicker.md
+++ b/docs/content/Datepicker.md
@@ -38,8 +38,6 @@ With a range-select datepicker, the user is able to select a range of dates by s
 
 ## Component props
 
-Native `<input>` attributes are forwarded to the underlying React `input` component and are not listed below.
-
 | Name                  | Type                                 |     Default      | Description                                                                                                                                                |
 | :-------------------- | :----------------------------------- | :--------------: | :--------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | anchoredOverlayProps  | Object                               |    undefined     | Optional. Additional properties passed to the overlay                                                                                                      |
--- a/docs/content/Datepicker.md
+++ b/docs/content/Datepicker.md
@@ -1,0 +1,89 @@
+---
+title: Datepicker
+description: Use datepicker to select a single date, multiple dates, or a range of dates
+status: Alpha
+source: https://github.com/primer/react/blob/main/src/Datepicker/index.tsx
+storybook: '/react/storybook?path=/story/composite-components-datepicker--default-date-picker'
+---
+
+## Default example
+
+The `Datepicker` component offers a variety of usages, such as single-select, multi-select, and range-select. The default usage is single-select
+
+```jsx live
+<>
+  <DatePicker />
+</>
+```
+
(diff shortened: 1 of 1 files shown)


siddharthkp:
Played around with the [storybook from the other PR](https://primer-components-1015mjqb0-primer.vercel.app/react/storybook?path=/story/composite-components-datepicker--start-day-picker), this is so well built!

Initial thoughts on API:

There is a wip ADR that talks about the API direction https://github.com/primer/react/pull/1703, you can see some of this in action in Autocomplete and [ActionMenu v2](https://primer.style/react/drafts/ActionMenu2)

When there are props on the top level component `DatePicker` which are passed down to children like `anchoredOverlayProps` and `iconPlacement`, we should expose that detail in the API as compound components: `DatePicker.Button` and `DatePicker.Overlay`

```jsx
<DatePicker>
  <DatePicker.Button>Choose Date</DatePicker.Button>
  <DatePicker.Overlay />
<DatePicker>
```

`DatePicker.Overlay`: `anchoredOverlayProps` like `align` and `side` become props for `DatePicker.Overlay`.
`DatePicker.Button`: `placeholder` becomes `children`, `iconPlacement` gets replaced with `leadingIcon` and `trailingIcon` based on [new button props](https://primer-components-65wwy51f1-primer.vercel.app/react/NewButton#icon-and-button). For this component, `leadingIcon` has a default value of `CalendarIcon`

What do you think?

dusave:
@siddharthkp Thanks for taking a look!

I really like the idea of using the compound props for this, it would definitely simplify the more complex implementations. One concern I do have though is with the simpler implementation, where the consumer just wants the basic datepicker, no props. Would that require them to insert the `Button` and `Overlay` components as children, even when they don't want to override any of the default behavior?

docs/content/Datepicker.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Datepicker
+description: Use datepicker to select a single date, multiple dates, or a range of dates
+status: Alpha
+source: https://github.com/primer/react/blob/main/src/Datepicker/index.tsx
+storybook: '/react/storybook?path=/story/composite-components-datepicker--default-date-picker'
+---
+
+## Default example
```
mperrotti, line 9: Minor nitpick: let's structure the headings to match what has been defined in https://github.com/primer/react/issues/1701

rezrah:
Hi @dusave - sorry it's taken so long to get back to you on this. I've spoken to a few Primer folks about our roadmap and unfortunately a date picker isn't on the near-term agenda for patterns that we plan to support. Reason being that the level of niche product focus and complexity balanced against the maintenance overhead means we'll struggle to support it. 

Do you see any issues in keeping this component internal for the time-being @dusave?. This is just the current state of play too, so we can revisit this decision down the line. 

cc. @gaknoll

PR #1711, opened 2021-12-09
Adding Emoji Picker API doc

dusave:
Describe your changes here.

Adds the Emoji Picker API doc

### Screenshots

emoji picker

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/EmojiPicker.md
+++ b/docs/content/EmojiPicker.md
@@ -1,0 +1,73 @@
+---
+title: Emoji Picker
+description: Use Emoji Picker to select an emoji
+status: Alpha
+source: https://github.com/primer/react/blob/main/src/EmojiPicker/index.tsx
+storybook: '/react/storybook?path=/story/composite-components-emoji-picker--default-emoji-picker'
+---
+
+## Default example
+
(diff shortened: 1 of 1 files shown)


pksjce:
This is so cool. 
The custom category is the tabs on the top? And it converts icons to emojis?

dusave:
@pksjce The custom categories are for any additional categories the user might need in addition to the standard unicode emoji set, such as our GitHub category. Each category contains an icon for the tab bar on top as well as an array of emojis (in the form of React Nodes) which can be images, other components, etc.

pksjce:
Would be good to document what arguments are returned in `onSelect`? I'm not sure what to expect out of that callback.
Also do you think any of the parts will be optional? Like the searchbar? And do you think we'll need a lite down version of the picker too?

dusave:
Interesting questions 🤔 I could see both. Toggling the search bar would be easy enough, but a compact view would require thought and designs. That said, I agree that a compact view could be valuable for certain use cases.

PR #1712, opened 2021-12-09
Remove "engines" field from package.json

colebemis:
The `engines` field is breaking builds for projects using node <16. Since node 16 is not required to use `@primer/react`, let's remove this field from the package.json.

[CleanShot 2021-12-09 at 14 53 32@2x](https://vercel.com/primer/doctocat/41ukG3xkShTswtBLLsHJnKrS1x54)

### DIFF
--- a/.changeset/tidy-maps-care.md
+++ b/.changeset/tidy-maps-care.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Remove "engines" field from package.json
--- a/package.json
+++ b/package.json
@@ -155,9 +155,5 @@
       "webpack": false,
       "running": false
     }
-  ],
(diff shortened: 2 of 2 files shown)


jfuchs (commented):
Is `engines` supposed to be about what's needed to run the distributed package? Or is it about what's needed to build and contribute?

colebemis:
I think it's "what's needed to run the distributed package" because I couldn't install `@primer/react` without node 16

colebemis:
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines
> You can also use the "engines" field to specify which versions of npm are capable of properly installing your program.

jfuchs (approved):

PR #1717, opened 2021-12-11
Update component docs (CircleBadge, Flash)

colebemis:
Part of https://github.com/primer/react/issues/1701

- [x] CircleBadge
- [x] Flash


docs/src/props-table.js
```diff
@@ -107,10 +107,12 @@ function PassthroughPropsRow({elementName, isPolymorphic, passthroughPropsLink})
 }
 
 function AsRow({defaultElementType}) {
+  // Element is a component if the first letter is uppercase (e.g. Button)
+  const isComponent = defaultElementType[0].toUpperCase() === defaultElementType[0]
   return (
     <Row
       name="as"
-      defaultValue={`"${defaultElementType}"`}
+      defaultValue={isComponent ? defaultElementType : `"${defaultElementType}"`}
```
colebemis, line 115: This prevents component names from being displayed as strings (`"Overlay"` vs `Overlay`)

docs/content/CircleBadge.mdx
```diff
@@ -0,0 +1,93 @@
+---
+componentId: circle_badge
```
rezrah, line 2: Total nitpick, but why do we put `componentId` at the top? I feel like `title` should always go first and `componentId` at the bottom as it's the least useful for those scanning this file.

rezrah (approved):

docs/content/CircleBadge.mdx
colebemis: I'm fine with either :) The only reason I put it on the top was to try to avoid merge conflicts with other PRs 😅

PR #1700, opened 2021-12-06
Use extracted @primer/behaviors

dgreif:
As discussed in [ADR 2](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-002-behavior-isolation.md#providing-behaviors-through-vanilla-javascript), all of the "behaviors" in PRC have been written in vanilla javascript, with React Hooks that wrap the vanilla js behaviors.  One of the reasons for this approach was that the behaviors could eventually be extracted and shared with other projects, like PVC and github/github.  We have seen an increased need for a few of these behaviors on the github/github side, and have extracted the behaviors to a new [@primer/behaviors](https://github.com/primer/behaviors) package that can be shared.  This PR removes all of the behaviors from the PRC project and pulls in @primer/behaviors as a replacement.

Closes https://github.com/github/primer/issues/483

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated documentation
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #1719, opened 2021-12-11
Add side effects false to package.json

mattcosta7:
Describe your changes here.

Fixes a bundler related bug introduced via #1700 where the sideEffects declaration was removed from package.json, potentially causing bundlers to not properly treeshake unused modules. 

#1332 introduced this fix initially, but had to declare specific files as containing sideEffects, due to their implementation.  When removing these files to a dependency, this should have been left in

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/dull-dots-act.md
+++ b/.changeset/dull-dots-act.md
@@ -1,5 +1,0 @@
----
-'@primer/react': patch
----
-
-adds sideEffects false back to package.json
--- a/.changeset/dull-dots-act.md
+++ b/.changeset/dull-dots-act.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+adds sideEffects false back to package.json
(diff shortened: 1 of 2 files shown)


pksjce (approved):
This is awesome!

.changeset/dull-dots-act.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+adds sideEffects false back to package.json
```
colebemis, line 5: I don't think we need a changeset on this PR because it patches unreleased changes

mattcosta7: removed it, but can't add the label

PR #1722, opened 2021-12-13
ADR: Parallel drafts track & plan for deprecated components

siddharthkp:
[Read the rendered markdown](https://github.com/primer/react/blob/siddharthkp/migration-adr/contributor-docs/adrs/adr-005.md)

Context: As we work on maturity of our components, we will sometimes need to deprecate components that are replaced with new components. Here's something to start that conversation and converge on a strategy.

tl;dr: We should detangle "drafts" from component lifecycle.

~~1. Rename `Experimental` to `Candidate` in [Component maturity lifecycle](https://primer.style/contribute/component-lifecycle): Candidate, Alpha, Beta and Stable.~~
~~2. Rename `drafts` to `experimental`. A component can have `beta` maturity in the `experimental` bundle.~~

(scrapped the above 2 suggestions for the churn they'll cause in the ecosystem, [see this comment](https://github.com/primer/react/pull/1722#discussion_r788000582) for more context. We do have the optional of changing `drafts` to `next` or `labs` if that's more descriptive)
3. Draft components should not be collocated with "main bundle" components in the documentation or status page. They should have their own section, as draft components are not recommended as an alternative yet.

&nbsp;

Flowchart: 
drafts flowchart

📺 Here's a video of a dry run of the flowchart: https://github.rewatch.com/video/arc9fb2sivdkm44j-draft-parallel-track-dry-run

Code example for the stages:

```js
// Stage 0: 
import { PrimaryButton } from '@primer/react'


// Stage 1: new component starts in drafts

// you can continue using current component
import { PrimaryButton } from '@primer/react'
// or try the new one
import { Button } from '@primer/react/lib-esm/drafts'


// Stage 2: new component goes from drafts to main

// you can migrate to the new component (breaking change, recommended)
import { Button } from '@primer/react'
// or continue to use the old, now deprecated one by changing the import
import { PrimaryButton } from '@primer/react/lib-esm/deprecated'


// Stage 3: deleting the deprecated component
// you need to migrate to the new component (breaking change)
import { Button } from '@primer/react'
```


contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+As we work on maturity of our components, we will sometimes need to deprecate components that are replaced with new components.
+
+## Here are the 3 proposed stages:
+
+### Stage 1
+
+Start new component outside the main bundle. Folks who want to try it, need to explicitly import it from the `drafts` bundle.
+
+`import { ActionMenu} from '@primer/react/drafts'`
```
siddharthkp, line 17: sidenote: thinking about these stages, I started thinking about the name again. 

In this context, I really like the name `'@primer/react/labs'` but that's a tiny detail

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+3. Grid
+4. Position
+
+Replaced with newer components
+
+5. Dialog - replaced with Dialog2, ready for stage 2
+6. Dialog2 - ready for Stage 2 with rename, already in main bundle
+7. Dropdown - replaced with DropdownMenu, ready for stage 2
+8. SelectMenu - replaced with SelectPanel?, ready for stage 2
+9. SelectPanel - unclear if it should be renamed, already in main bundle
```
siddharthkp, line 56: This is a TODO: Should we rename it? It's already in the main bundle, used in production and doesn't seem like a 1:1 successor to SelectMenu. Maybe, it doesn't belong to this list.

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+Future deprecations, replaced with newer components
+
+10. ActionList - replaced with ActionList2, need to plan for stage 2
+11. ActionList v2 - in drafts bundle, need to plan for stage 2
+12. ActionMenu - replaced with ActionList2, need to plan for stage 2
+13. ActionList v2 - in drafts bundle, need to plan for stage 2
+14. Button - replaced with NewButton, need to plan for stage 2
```
siddharthkp, line 64: suggested name change here for consistent strategy

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+`import { ActionMenu } from '@primer/react/deprecated'`
+
+The deprecated component does not accept new features requests.
+
+Reason: Because we have a single bundle for all components, you can not pick the components you want to upgrade. This can results in additional unrelated work.
+
+### Stage 3
+
+After 3 months of living in the `deprecated` bundle, a component is retired/deleted from the codebase.
```
siddharthkp, line 37: `3 months` is an example, do you feel this should have a longer timeline?

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Context
+
+As we work on maturity of our components, we will sometimes need to deprecate components that are replaced with new components.
+
+## Here are the 3 proposed stages:
+
+### Stage 1
+
+Start new component outside the main bundle. Folks who want to try it, need to explicitly import it from the `drafts` bundle.
```
colebemis, line 15: Should drafts be listed in the docs sidenav?

contributor-docs/adrs/adr-005.md
colebemis: I like `/labs` too ❤️

contributor-docs/adrs/adr-005.md
siddharthkp: yep! 💯

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+`import { ActionMenu} from '@primer/react/drafts'`
+
+The contract with consumers is - you are opting into a rewrite of the old component that might not cover all the cases of the old component yet. If you are using both the old and new version of the component in different pages, you are paying some additional bundlesize cost.
+
+Note: If it's a 1:1 replacement, it's useful to keep the same name.
```
mperrotti, line 21: So we would stop adding "2" as a suffix? For example: `TextInput2` would just be called `TextInput` if it was a 1-to-1 replacement.

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Stage 2
+
+After we have the confidence that this component is better than the old version of it, we swap the components and move the old component out of the main bundle.
```
mperrotti, line 25: What factors determine how confident we are? If it's in `main`, it's already been code reviewed.

contributor-docs/adrs/adr-005.md
mperrotti: I think 3 months is fine for most cases. We can always make an exception if we need to.

contributor-docs/adrs/adr-005.md
mperrotti: SelectPanel seems different from SelectMenu. It's just unfortunate that the names are so similar.

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+5. Dialog - replaced with Dialog2, ready for stage 2
+6. Dialog2 - ready for Stage 2 with rename, already in main bundle
+7. Dropdown - replaced with DropdownMenu, ready for stage 2
+8. SelectMenu - replaced with SelectPanel?, ready for stage 2
+9. SelectPanel - unclear if it should be renamed, already in main bundle
+
+Future deprecations, replaced with newer components
+
+10. ActionList - replaced with ActionList2, need to plan for stage 2
```
mperrotti, line 60: Deprecating `ActionList` would probably affect any components that use `ActionList`. For example: `Autocomplete`

contributor-docs/adrs/adr-005.md
siddharthkp: I think it's a mixed bag,

In the shipped bundle:  `import { TextInput } from '@primer/react/labs'`

but the file name might still be `TextInput2.tsx` and the docs would still say `TextInput v2`. Would be nice to have some consistency around this.

contributor-docs/adrs/adr-005.md
siddharthkp: Good question! I think it's when we think we are happy with the API and it have internally tested that it covers all the use cases of the previous component. 

For example, I'd like to ship all the Menu components that compose ActionList together, so I'm waiting till I finish SelectPanel

contributor-docs/adrs/adr-005.md
siddharthkp: Yep! Autocomplete can still keep using ActionList v1, however we increase our bundlesize by a tiny bit by a few kbs by including both ActionLists (which is totally acceptable in the short term)

contributor-docs/adrs/adr-005.md
colebemis: ☝️  Could you add that to this doc for clarity?

siddharthkp: Done :)

mperrotti (approved):

contributor-docs/adrs/adr-005.md
rezrah: While I like labs too, would this make sense for users to import from `labs` as it feels more like a function. Would something like `import { ActionMenu} from '@primer/react/experimental'` make more sense? (I feel like I'm having deja vu, has this already been proposed? 😅 )

Might also be worth syncing with ViewComponents on this naming convention for ecosystem consistency. I saw a PVC thread where they were thinking about calling it `experimental`. Will try find it.

contributor-docs/adrs/adr-005.md
rezrah: For easier discovery, could we consider co-locating them somehow to the original component? Not sure how, maybe tabs? But would help reduce IA complexity too.

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Stage 1
+
+Start new component outside the main bundle. Folks who want to try it, need to explicitly import it from the `drafts` bundle.
+
+`import { ActionMenu} from '@primer/react/drafts'`
+
+The contract with consumers is - you are opting into a rewrite of the old component that might not cover all the cases of the old component yet. If you are using both the old and new version of the component in different pages, you are paying some additional bundlesize cost.
+
+Note: If it is a 1:1 replacement, it's useful to keep the component name the same for consumers. Internally, we would want to call the filename `ActionMenu2.tsx` and call it `ActionMenu2 v2` in the docs.
```
rezrah, line 21: After stage 2, where the component gets moved to `/deprecated`, do we rename this back to `ActionMenu`? Even if it's a breaking change in major version, it feels cleaner to take `v2` out of the naming.

contributor-docs/adrs/adr-005.md
rezrah: Shall we move these into an issue instead? I can imagine this changing very frequently so doesn't feel like it should be in an ADR?

contributor-docs/adrs/adr-005.md
siddharthkp: Sooo, the import name for consumers is always `ActionMenu` but the import path would change.

```diff
- import { ActionMenu } from '@primer/react/lib-esm/drafts'
+ import { ActionMenu } from '@primer/react'
```

The filename internally and the docs would drop the `v2` suffix

contributor-docs/adrs/adr-005.md
pouretrebelle: I can imagine that automating migrations (to maintain the same functionality without upgrading anything) when you make the breaking change would be really messy if a file has other `@primer/react` or `@primer/react/drafts` imports.

Is performance the main reason to bundle drafts separately? I'd have thought that with tree-shaking that wasn't much of a concern but idk how our js is all set up, and given we want to include the draft components in the docs we're not exactly trying to hide them.

If they were all in the main bundle you could have `ActionMenu` and `ActionMenuDraft`/`ActionMenuNext`, then renaming them to `ActionMenuDeprecated` and `ActionMenu` when you make the breaking change would be trivial. Just a thought.

siddharthkp: Hi! 

Sorry for chiming in late! I don't think automating migrations is a goal here. Most of the components in drafts right now have breaking API changes, so optimising for import paths isn't a goal here. 

> Is performance the main reason to bundle drafts separately?

Separation of tracks was the main reason here. Seeing `ActionMenu` and `ActionMenuNext` can be confusing to choose from. (We did not do ourselves any favours in the status page with v1 and v2).

The correct way to title them in the docs might be `ActionMenu` and `Experimental/ActionMenu`. I updated the PR description and ADR to add the usage of parallel tracks

Parallel tracks

contributor-docs/adrs/adr-005.md
siddharthkp: this was done :)

contributor-docs/adrs/adr-005.md
siddharthkp: See updated suggestions! (I think we should not collocate them at all)

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,73 @@
...
+After 3 months of living in the `deprecated` bundle, a component is retired/deleted from the codebase.
+
+At this point, consumers are expected to plan migration work.
+
+## Suggested changes:
+
+We should detangle "experimental" from component lifecycle.
+
+1. Rename `Experimental` to `Candidate` in [Component maturity lifecycle](https://primer.style/contribute/component-lifecycle): Candidate, Alpha, Beta and Stable.
+2. Rename `drafts` to `experimental`. A component can have `beta` maturity in the `experimental` bundle.
```
rezrah, line 46: Major ➕

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,73 @@
...
+3. Experimental components should not be collocated with "main bundle" components in the documentation or status page. They should have their own section, as experimental components are not recommended as an alternative yet.
```
rezrah, line 47: Makes sense, but it makes discovery and IA much more challenging. I think it's a wider discussion, so happy to go with this for now. This may make more sense if we can create a dedicated Labs section in our docs.

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,73 @@
...
+
+After 3 months of living in the `deprecated` bundle, a component is retired/deleted from the codebase.
+
+At this point, consumers are expected to plan migration work.
+
+## Suggested changes:
+
+We should detangle "experimental" from component lifecycle.
+
+1. Rename `Experimental` to `Candidate` in [Component maturity lifecycle](https://primer.style/contribute/component-lifecycle): Candidate, Alpha, Beta and Stable.
```
pouretrebelle, line 45: `Experimental` is a namespace used for components being developed in .com that have the intention of being upstreamed to PVC ([examples](https://cs.github.com/?scopeName=github%2Fgithub&scope=repo%3Agithub%2Fgithub&q=Primer%3A%3AExperimental)). It would be quite a pain to migrate these to `Candidate` - and I think in that context the word experimental is more meaningful because candidate implies a level of rigour that they don't necessarily have. Although it's a candidate in the eyes of our component lifecycles to consumers it's more important that they see it's unstable. I'd prefer to leave that as experimental and use a different term for this track (fwiw I like Cole's `lab` 😄)

siddharthkp: Fair enough, makes a lot of a sense. 

I guess we can keep `drafts` because it has some inertia 😅

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Stage 1
+
+Start new component outside the main bundle. Folks who want to try it, need to explicitly import it from the `drafts` bundle.
+
+`import { ActionMenu} from '@primer/react/drafts'`
+
+The contract with consumers is - you are opting into a rewrite of the old component that might not cover all the cases of the old component yet. If you are using both the old and new version of the component in different pages, you are paying some additional bundlesize cost.
+
+Note: If it is a 1:1 replacement, it's useful to keep the component name the same for consumers. Internally, we would want to call the filename `ActionMenu2.tsx` and call it `ActionMenu2 v2` in the docs.
```
colebemis, line 21:
```suggestion
Note: If it is a 1:1 replacement, it's useful to keep the component name the same for consumers. Internally, we would want to call the filename `ActionMenu2.tsx` and call it `ActionMenu v2` in the docs.
```

siddharthkp:
This ADR feels silly because we already ran this whole pipeline once 😅

Should we merge it as accepted 🤷‍♀️ and keep improving it?

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,71 @@
...
+
+After we have the confidence that this component is better than the old version of it, we swap the components and move the old component out of the main bundle.
+
+This is a breaking change! It’s recommended that you start using the new component, but if you’d like to push that effort to the future, we give you an easy way out -
+
+`import { ActionMenu } from '@primer/react/deprecated'`
+
+The deprecated component does not accept new features requests.
+
+Reason: Because we have a single bundle for all components, you can not pick the components you want to upgrade. This can results in additional unrelated work.
```
lesliecdubs, line 33:
```suggestion
Reason: Because we have a single bundle for all components, you can not pick the components you want to upgrade. This can result in additional unrelated work.
```

lesliecdubs:

**ADR decision maker:** @pksjce 
**Decision deadline:** Thursday, 5 May

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Stage 3
+
+After 3 months of living in the `deprecated` bundle, a component is retired/deleted from the codebase.
+
+At this point, consumers are expected to plan migration work.
+
+## Suggested changes:
+
+We should detangle "drafts" from component lifecycle.
```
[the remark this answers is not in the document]
colebemis: `/drafts` is a great place to put components that we want to build incrementally. Components can live in `/drafts` until they are feature complete.

colebemis (approved):
Thanks for documenting this process, @siddharthkp! 

Left a few comments but nothing blocking.

pksjce (approved):
This has already been put into practice in v35. While we haven't deleted any deprecated component yet, all of the other recommendations are working out just fine for us and our consumers. We have both `/drafts` and `/deprecated` paths setup for import and for documentation. LGTM 👍

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,71 @@
...
+
+The deprecated component does not accept new features requests.
+
+Reason: Because we have a single bundle for all components, you can not pick the components you want to upgrade. This can results in additional unrelated work.
+
+### Stage 3
+
+After 3 months of living in the `deprecated` bundle, a component is retired/deleted from the codebase.
+
+At this point, consumers are expected to plan migration work.
```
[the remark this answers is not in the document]
siddharthkp: There's [a long list](https://github.com/primer/react/pull/1898#:~:text=Deprecate%20Dropdown%20v1) for "deprecating a component" 😅 Should we add that to the contribution docs instead of ADR?

contributor-docs/adrs/adr-005.md
```diff
@@ -0,0 +1,71 @@
...
+
+Start new component outside the main bundle. Folks who want to try it, need to explicitly import it from the `drafts` bundle.
+
+`import { ActionMenu} from '@primer/react/drafts'`
+
+The contract with consumers is - you are opting into a rewrite of the old component that might not cover all the cases of the old component yet. If you are using both the old and new version of the component in different pages, you are paying some additional bundlesize cost.
+
+Note: If it is a 1:1 replacement, it's useful to keep the component name the same for consumers. Internally, we would want to call the filename `ActionMenu2.tsx` and call it `ActionMenu v2` in the docs.
+
+### Stage 2
```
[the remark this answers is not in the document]
siddharthkp: Is that separate from `@primer/react/next`? 😅

contributor-docs/adrs/adr-005.md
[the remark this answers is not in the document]
colebemis: > Should we add that to the contribution docs instead of ADR?

Yup, makes sense to me 👍

PR #1733, opened 2021-12-14
Button v2: Fix icon positions for block buttons

siddharthkp:
With a block button, the icon and text alignment is a bit strange

### Screenshots

image

image

The chromatic tests (see checks below)  show that there are no visual regressions :)

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/NewButton/styles.ts
+++ b/src/NewButton/styles.ts
@@ -225,11 +225,13 @@ export const getButtonStyles = (theme?: Theme) => {
       gridArea: 'leadingIcon'
     },
     '[data-component="text"]': {
-      gridArea: 'text',
-      textAlign: 'left'
+      gridArea: 'text'
     },
     '[data-component="trailingIcon"]': {
       gridArea: 'trailingIcon'
+    },
+    '[data-component="leadingIcon"] + [data-component="text"]': {
+      textAlign: 'left'
     }
   }
   return styles
--- a/src/NewButton/styles.ts
+++ b/src/NewButton/styles.ts
@@ -217,6 +217,7 @@ export const getButtonStyles = (theme?: Theme) => {
     ...getBaseStyles(theme),
     display: 'grid',
     gridTemplateAreas: '"leadingIcon text trailingIcon"',
+    gridTemplateColumns: 'min-content 1fr min-content',
     '& > :not(:last-child)': {
       mr: '2'
     },
@@ -224,7 +225,8 @@ export const getButtonStyles = (theme?: Theme) => {
       gridArea: 'leadingIcon'
     },
     '[data-component="text"]': {
-      gridArea: 'text'
+      gridArea: 'text',
+      textAlign: 'left'
     },
     '[data-component="trailingIcon"]': {
       gridArea: 'trailingIcon'
--- a/.changeset/button2-block-position.md
+++ b/.changeset/button2-block-position.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Button 2: Fix icon positions for block buttons
--- a/src/__tests__/__snapshots__/NewButton.test.tsx.snap
+++ b/src/__tests__/__snapshots__/NewButton.test.tsx.snap
@@ -23,6 +23,7 @@ exports[`Button renders consistently 1`] = `
   text-align: center;
   display: grid;
   grid-template-areas: "leadingIcon text trailingIcon";
+  grid-template-columns: min-content 1fr min-content;
   padding-top: 5px;
   padding-bottom: 5px;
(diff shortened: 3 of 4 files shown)


pksjce:
Interesting..
So, if the trailingIcon exists, then we get the text back to the middle?

image

src/NewButton/styles.ts
```diff
@@ -227,6 +228,9 @@ export const getButtonStyles = (theme?: Theme) => {
     },
     '[data-component="trailingIcon"]': {
       gridArea: 'trailingIcon'
+    },
+    '[data-component="leadingIcon"] + [data-component="text"]': {
+      textAlign: 'left'
```
pksjce, line 233: Not sure if this leftAlign is required tbh.

siddharthkp:
> Interesting.. So, if only the trailingIcon exists, then we get the text back to the middle?

> image

If leadingIcon exists, then we want leadingIcon and text to be together instead of far apart. cc @ashygee

src/NewButton/styles.ts
siddharthkp: fair, how do we make it easier for the user to override this with left align?

this would be the override, not sure if requires knowing the internal details?

```css
[data-component="text"] {
  text-align: left
}
```

pksjce: True :( This is where composite components would be handy. Maybe such an override usecase means we want to open up the API?

mperrotti (commented):
I think left-aligning the `leadingVisual` and label when we have a `trailingVisual` is a bit strange. It will especially look weird if we have two stacked full-width buttons where 1 has left-aligned text, and the other has center-aligned text.

I think the "current" column of [the mock you posted](https://user-images.githubusercontent.com/1863771/146175063-efa2a002-4c67-4496-ba17-65d7434439b9.png) is what we want.

Curious to hear what @ashygee thinks

src/NewButton/styles.ts
```diff
@@ -217,6 +217,7 @@ export const getButtonStyles = (theme?: Theme) => {
     ...getBaseStyles(theme),
     display: 'grid',
     gridTemplateAreas: '"leadingIcon text trailingIcon"',
+    gridTemplateColumns: 'min-content 1fr min-content',
```
mperrotti, line 220: If we decide to go with the center-aligned approach I recommended, we can remove this style and just add `justify-content: center;`

src/NewButton/styles.ts
mperrotti: If we go with the center-aligned approach I recommended, we can get rid of this style block.

siddharthkp:
For context, this is the pattern I'm chasing:

Projects button

This is what the code looks like for positioning it correctly (ignoring the hover state for now):

To be able to override it correctly, you have to know some implementation details about it, which feels wrong :(

```jsx
<NewButton
    variant="invisible"
    trailingIcon={GearIcon}
    sx={{
      width: "100%",
      justifyContent: "space-between",
      gridTemplateAreas: '"text trailingIcon"'
    }}
  >
```

pksjce:
@mperrotti - If there is a product use case then I think we should work on supporting the new mocks.
@siddharthkp - Would a boolean prop for `block` which takes care of this situation better? Though I'm still not sure how the alignment will work.

pksjce:
Since @ashygee is on holidays, I was wondering if @vdepizzol could help us with this?

pksjce:
This is what primer-viewcomponents looks like with block button and visuals. 

image

And in primer-css
image

Center-align seems standard in our system.

siddharthkp:
Context:

> If there is a product use case then I think we should work on supporting the new mocks.

only [issues-app](https://github.com/github/issues-app) for now I think, but feels like a generic use case, we can also decide we want this alignment to be an override in the product instead of supporting it out of the box.



> @siddharthkp - Would a boolean prop for block which takes care of this situation better? Though I'm still not sure how the alignment will work.

That could work!

> Since @ashygee is on holidays, I was wondering if @vdepizzol could help us with this?

That would be nice. This can also wait, it's not a december problem :)

mperrotti:
This came up during a Primer design pattern working session, and I like how @langermank decided to handle it

Screen Shot 2022-01-25 at 4 09 55 PM

Storybook preview deployment link: https://primer-css-goebhg4xi-primer.vercel.app/css/storybook?path=/story/explorations--example-sheet
WIP PR: https://github.com/primer/css/pull/1874

siddharthkp:
This isn't the direction in which we're going ✌️

PR #1734, opened 2021-12-14
Button v2 - Fix font family

siddharthkp:
Describe your changes here.

Button v1 [has `font-family: inherit`](https://github.com/primer/react/blob/main/src/Button/ButtonStyles.tsx#L8), this was missing from NewButton which made it use Arial on Chrome instead of system-font

### Screenshots

before:
Before fix, Button has Arial

after:
Before fix, Button has system font

### DIFF
--- a/.changeset/tiny-cows-fold.md
+++ b/.changeset/tiny-cows-fold.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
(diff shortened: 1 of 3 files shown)


colebemis (approved):

pksjce (approved):
Thanks!

PR #1731, opened 2021-12-14
Select Input react component

pksjce:
Describe your changes here.

This is one of the components implemented for the form components epic.
Things to do
 - [ ] Style the carets properly
 - [ ] Optgroup component
 - [ ] discuss textinput wrapper height
 - [ ] Add doc

Closes https://github.com/github/primer/issues/539

### Screenshots

TBD

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #1736, opened 2021-12-14
Select input component

mperrotti:
Adds a styled native `<select>` input component

Closes https://github.com/github/primer/issues/540

### Screenshots

https://user-images.githubusercontent.com/2313998/146074318-9f773146-c19f-4871-b785-c4c98fe4cc4f.mp4

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


docs/content/Select.mdx
```diff
@@ -0,0 +1,100 @@
+---
+title: Select
```
mperrotti, line 2: I'm thinking maybe "SelectInput" is a better name. I'd love to hear what other reviewers think.

pksjce:
There seems to be a clash here 😅 

https://github.com/primer/react/pull/1731

I'll close mine, as I'm yet to implement optgroups and review yours.

docs/content/Select.mdx
pksjce: Yep! I think so too. It kinda cements it as a Form component and not any fancy menu/overlay interface

src/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+  </StyledSelect>
+)
+
+const Option: React.FC<React.HTMLProps<HTMLOptionElement> & {value: string}> = props => <option {...props} />
+
+const Group: React.FC<React.HTMLProps<HTMLOptGroupElement>> = props => <optgroup {...props} />
+
+export default Object.assign(Select, {
+  Option,
+  Group
```
pksjce, line 53: Can we name this `OptGroup`?

src/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+    as="select"
+    ref={ref as RefObject<HTMLSelectElement>}
+    required={required || Boolean(placeholder)}
+    disabled={disabled}
+    aria-required={required}
+    aria-disabled={disabled}
+    {...rest}
+  >
+    {placeholder ? (
+      <option value="" disabled={required} selected hidden={required}>
```
pksjce, line 39: I think we need a storybook to show what this looks like?

src/stories/Select.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  ],
+  argTypes: {
+    block: {
+      defaultValue: false,
+      control: {
+        type: 'boolean'
+      }
+    },
+    contrast: {
```
pksjce, line 28: This shouldn't be relevant for this component

src/stories/Select.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+        type: 'boolean'
+      }
+    },
+    sx: {
+      table: {
+        disable: true
+      }
+    },
+    variant: {
+      name: 'Variant',
```
pksjce, line 52: can we call this `size`?

pksjce:
Some comments
1. Should it show required * symbol?
2. All the dark mode colors seem broken? 

Dark mode
image

Dark high contrast
image

Dark color blind
image

docs/content/Select.mdx
rezrah: We only recently decided to avoid using `*Input` suffix on form components, so I'm in favour of keeping it as `Select` here. Reason: a) To match the underlying HTML and b) be consistent with other form components like Radio, Checkbox.

docs/content/Select.mdx
```diff
@@ -0,0 +1,100 @@
...
+description: Use a select input when a user needs to select one option from a long list
+status: Alpha
+source: https://github.com/primer/react/blob/main/src/Select.tsx
+storybook: '/react/storybook?path=/story/forms-select--default'
+---
+
+import {ComponentChecklist} from '../src/component-checklist'
```
rezrah, line 9: Don't think we need to import this anymore as it's global

src/Select.tsx
rezrah: ➕ . Can we call it `OptionGroup` to avoid the abbreviation?

src/Select.tsx
```diff
@@ -0,0 +1,54 @@
+import React, {RefObject} from 'react'
+import styled from 'styled-components'
+import {get} from './constants'
+import TextInputWrapper, {StyledWrapperProps} from './_TextInputWrapper'
+
+type SelectProps = Omit<React.HTMLProps<HTMLSelectElement> & StyledWrapperProps, 'multiple' | 'size' | 'hasIcon' | 'as'>
+
+const StyledSelect = styled(TextInputWrapper)<SelectProps>`
+  appearance: none;
+  background-image: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTYiIGhlaWdodD0iMTYiIHZpZXdCb3g9IjAgMCAxNiAxNiIgZmlsbD0iIzU4NjA2OSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48cGF0aCBkPSJNNC40MjcgOS40MjdsMy4zOTYgMy4zOTZhLjI1MS4yNTEgMCAwMC4zNTQgMGwzLjM5Ni0zLjM5NkEuMjUuMjUgMCAwMDExLjM5NiA5SDQuNjA0YS4yNS4yNSAwIDAwLS4xNzcuNDI3ek00LjQyMyA2LjQ3TDcuODIgMy4wNzJhLjI1LjI1IDAgMDEuMzU0IDBMMTEuNTcgNi40N2EuMjUuMjUgMCAwMS0uMTc3LjQyN0g0LjZhLjI1LjI1IDAgMDEtLjE3Ny0uNDI3eiIgLz48L3N2Zz4=);
```
rezrah, line 10: Is there a way we can show the icon without using a base64 uri? It's quite expensive on the runtime and AFAIK a new patter to PRC that could set a slippery precedent. Also, this image in its current form seems incompatible with other colour modes. Worth looking at pulling this in as a composable icon instead?

src/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+  background-repeat: no-repeat;
+  background-position: right 4px center;
+  background-size: 16px;
+  padding-right: 20px;
+  cursor: pointer;
+
```
rezrah, line 16: worth setting the background-color here too?

src/Select.tsx
```diff
@@ -0,0 +1,54 @@
...
+  <StyledSelect
+    as="select"
+    ref={ref as RefObject<HTMLSelectElement>}
+    required={required || Boolean(placeholder)}
+    disabled={disabled}
+    aria-required={required}
+    aria-disabled={disabled}
+    {...rest}
+  >
+    {placeholder ? (
```
rezrah, line 38: Do we need the ternary here? Could shorten it with `&&` instead?

```suggestion
{placeholder && <option value="" disabled={required} selected hidden={required}>{placeholder}</option>}
``

File: src/Select.tsx (hunk 2 above)
Comment (mperrotti) [reply src/Select.tsx:53]: I think if we're going to call the parent component `Select`, `OptGroup` might make more sense because it's closer to the HTML tag name: `<optgroup>`

File: src/Select.tsx (hunk 3 above)
Comment (mperrotti) [reply src/Select.tsx:39]: You can see it by using the Controls. Do you think we need something more prominent?

File: src/stories/Select.stories.tsx (hunk 4 above)
Comment (mperrotti) [reply src/stories/Select.stories.tsx:28]: Ok. I can remove it and we can always add it later if we decide we want parity with `TextInput`

File: src/stories/Select.stories.tsx (hunk 5 above)
Comment (mperrotti) [reply src/stories/Select.stories.tsx:52]: Yes, absolutely. I was waiting for https://github.com/primer/react/pull/1661 to merge which deprecates `variant` and adds `size`.

**Comment by mperrotti** [comment]
> Should it show required * symbol?

That will be handled by the InputFieldComponent added in https://github.com/primer/react/pull/1611

> All the dark mode colors seem broken?

That's no good. I'll take a look.

File: docs/content/Select.mdx (hunk 6 above)
Comment (mperrotti) [reply docs/content/Select.mdx:9]: That's what I thought too, but if I don't import it, it doesn't render.

File: src/Select.tsx (hunk 7 above)
Comment (mperrotti) [reply src/Select.tsx:10]: Those are both great points. I'll investigate alternatives.

File: src/Select.tsx (hunk 8 above)
Comment (mperrotti) [reply src/Select.tsx:16]: That should be handled in the `TextInputWrapper` component.

File: docs/content/Select.mdx (hunk 6 above)
Comment (rezrah) [reply docs/content/Select.mdx:9]: Could your `/docs` node_modules need updating?

File: src/Select.tsx [hunk 10]
```diff
@@ -0,0 +1,74 @@
...
+    >
+      <StyledSelect
+        ref={ref}
+        required={required || Boolean(placeholder)}
+        disabled={disabled}
+        aria-required={required}
+        aria-disabled={disabled}
+        {...rest}
+      >
+        {placeholder ? (
```
rezrah, line 55: Nit: Feels like an unnecessary ternary when you can use an AND / && instead?

docs/content/Select.mdx
```diff
@@ -0,0 +1,100 @@
+---
```
rezrah, line 1: `Select` wasn't added to `nav.yml` and is therefore not present in the sidebar menu. Was this intentional @mperrotti? You can only find this doc page via the search atm.

src/Select.tsx
mperrotti: Great point. I'll switch it

docs/content/Select.mdx
mperrotti: Not intentional. Adding now...

docs/content/Select.mdx
```diff
@@ -0,0 +1,100 @@
...
+    <Select.Option value="pvc">Primer ViewComponents</Select.Option>
+  </Select>
+</>
+```
+
+## Props
+
+### Select
+
+The `Select` component accepts the same props as a native HTML [`<select>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) except for `multiple` and `size`.
```
rezrah, line 70: There used to be prop tables here, have they been removed? 🤔

docs/content/Select.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+### Select.OptGroup
+
+The `Select.OptGroup` component accepts the same props as a native HTML [`<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup).
+
+### Select.Option
+
+The `Select.Option` component accepts the same props as a native HTML [`<option>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option).
+
+<Props of={Select.Option} />
```
rezrah, line 80: This doesn't seem to be working
Screenshot 2021-12-17 at 14 36 54

src/__tests__/Select.test.tsx
```diff
@@ -0,0 +1,136 @@
+import React from 'react'
+import {Select} from '..'
+import {render} from '@testing-library/react'
+import {toHaveNoViolations} from 'jest-axe'
+import 'babel-polyfill'
+import '@testing-library/jest-dom'
+
+expect.extend(toHaveNoViolations)
+
+describe('Radio', () => {
```
rezrah, line 10: Should be `Select` here

docs/content/Select.mdx
mperrotti: I don't think there ever were

src/__tests__/Select.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <Select.Option value="four">Choice four</Select.Option>
+          <Select.Option value="five">Choice five</Select.Option>
+          <Select.Option value="six">Choice six</Select.Option>
+        </Select>
+      </>
+    )
+
+    const select = getByLabelText('Choice')
+
+    expect(select.getAttribute('aria-disabled')).toBeTruthy()
```
rezrah, line 133: ❤️ . Personal preference, but`@testing-library/jest-dom` has some nice, more easily-readable matchers for this type of test. E.g. 

```suggestion
    expect(select).toHaveAttribute('aria-disabled')
```

rezrah (approved):
Looks good to go, nice one @mperrotti. Just a heads up, this won't appear on the primer.style/status page until you add `comonentId` to the front-matter. Might be worth sticking that into this PR or in a follow-up.

Issue #1753, opened 2021-12-20
Dialog2 should use h1 and h2 for title and description


owenniblock:
**Describe the bug**

For accessibility (and semantic correctness) our dialog should present the `title` and `description` parameters in `<h1>` and `<h2>` tags.

Currently the example in the docs shows that this uses `<div>` [see this link to the docs for Dialog2](https://primer.style/react/Dialog2)

**To Reproduce**

1. Go to https://primer.style/react/Dialog2#example
2. Click on `Open`
3. Inspect the title and description text
4. See that they're `<div>` elements

PR #1754, opened 2021-12-20
Fix title and subtitle tag for Dialog2 and ConfirmationDialog

siddharthkp:
Fixes https://github.com/primer/react/issues/1753

title should use h1 and subtitle should use h2 instead of div

### Merge checklist

- NA Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/Dialog/Dialog.tsx
+++ b/src/Dialog/Dialog.tsx
@@ -314,17 +314,18 @@ const Header = styled.div.attrs<SxProp>({as: 'header'})`
   flex-shrink: 0;
 `
 
-const Title = styled.div<SxProp>`
+const Title = styled.h1<SxProp>`
   font-size: ${get('fontSizes.1')};
   font-weight: ${get('fontWeights.bold')};
-
+  margin: 0; /* override default margin */
   ${sx};
 `
 
-const Subtitle = styled.div<SxProp>`
+const Subtitle = styled.h2<SxProp>`
   font-size: ${get('fontSizes.0')};
-  margin-top: ${get('space.1')};
   color: ${get('colors.fg.muted')};
+  margin: 0; /* override default margin */
+  margin-top: ${get('space.1')};
 
   ${sx};
 `
--- a/.changeset/dialogs-title-tag.md
+++ b/.changeset/dialogs-title-tag.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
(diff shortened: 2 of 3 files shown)


pksjce:
LGTM. Why the margin overrides though?

src/Dialog/Dialog.tsx
```diff
@@ -314,17 +314,18 @@ const Header = styled.div.attrs<SxProp>({as: 'header'})`
...
+const Title = styled.h1<SxProp>`
   font-size: ${get('fontSizes.1')};
   font-weight: ${get('fontWeights.bold')};
-
+  margin: 0; /* override default margin */
   ${sx};
 `
 
-const Subtitle = styled.div<SxProp>`
+const Subtitle = styled.h2<SxProp>`
```
rezrah, line 324: Wondering if this makes semantic sense. Should it be a `p` instead?

siddharthkp:
> Why the margin overrides though?

h1 and h2 have default margins in the browser's stylesheet, so we we override it with the margin we want.

src/Dialog/Dialog.tsx
siddharthkp: Following it from https://github.com/primer/react/issues/1753, should be h2

rezrah (approved):
Nice one @siddharthkp ⚡. Would be nice to have some test coverage on the heading roles, but we can follow-up on that later.

src/Dialog/Dialog.tsx
owenniblock: Hmm, I've been thinking about this and it's a good question. I naturally leaned towards `<h2>` but I'm not sure that's actually the right thing. Maybe @bolonio will have opinions here?

Sorry if that turns out to be the wrong tag for the description!

bolonio: First of all, we can agree that Dialog title headings should be `<h1>`. I would go with `<h2>` for the subtitle as well, since modal dialogs are considered areas that are distinct from their parent document, so they have their own heading level structure. I wouldn't say that using a `<p>` is wrong, since not having a `<h2>` heading is possible. I've been reading about it and I couldn't find anything about subtitle or `<h2>` in dialog (of course there are some information about the dialog title  to be `<h1>`). 

Related issues:
- https://github.com/github/accessibility/issues/450#issuecomment-988037997

PR #1759, opened 2021-12-22
ActionMenu2: Add aria-labelledby to ActionList

siddharthkp:
- Part of https://github.com/github/primer/issues/518

Changes:
- [x] AnchoredOverlay: Add support for passing an id to the anchor
- [x] AnchoredOverlay: Remove unnecessary aria-labelledby on anchor (which was the same as the id on the anchor)
- [x] ActionMenu v2: Add aria-labelledby for ActionList

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


docs/content/AnchoredOverlay.mdx
```diff
@@ -81,6 +81,11 @@ See also [Overlay positioning](/Overlay#positioning).
       </>
     }
   />
+  <PropsTableRow
+    name="anchorId"
+    type="string"
+    description="Optional: An override to the internal id that will be passed to the anchor."
```
rezrah, line 87: Don't think you'll need to add optional here, as it's assumed they are all optional unless flagged as required.

rezrah (approved):

PR #1770, opened 2022-01-05
Add aria-hidden to InputLabel required asterisk

PeterYangIO:
The required asterisk needs the `aria-hidden="true"` attribute so that it doesn't get announced by screen readers.

I discovered this issue since switching my input components to `<InputField/>` caused testing-library's `getByLabelText` to fail. That selector uses the accessibility tree (as pictured in screenshots) and looks for exact name matches, so `getByLabelText("Name")` would fail as the name would be "Name *" and not "Name".

I also fixed the broken InputField source link in the docs.

### Screenshots
Before, asterisk was included in name property:
image

After, asterisk is no longer included in name property but still displays properly:
image

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [ ] Tested in Safari -> Unable to since I'm on Windows
- [X] Tested in Edge

### DIFF
--- a/.changeset/breezy-coats-play.md
+++ b/.changeset/breezy-coats-play.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Adds aria-hidden="true" to InputLabel required asterisk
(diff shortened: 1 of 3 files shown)


siddharthkp (approved):
Great spot! Thank for taking the time to create a PR :)

Issue #1768, opened 2022-01-04
Multiple `Alpha` components can't be used despite being available in the documentation.


maximedegreve:
## Problem

While trying to prototype new layouts using several Alpha components I ran into issues. 
I understand those components are still drafts but I think they should be accessible for prototyping in early stages to discover missing patterns and especially when they can be found in our documentation.
If they are not ready at all we should probably hide them from the documentation.

Internally: If we remove/hide them in the package we should find a way to expose those components for early prototyping.

## [SelectPanel](https://primer.style/react/SelectPanel)

- `import { SelectPanel } from '@primer/react'` doesn't work
- `import { SelectPanel } from '@primer/react/drafts'` doesn't work
- `import { SelectPanel } from '@primer/react/lib-esm/drafts'` doesn't work
- There is no information how to import this at all

## [IconButton](https://primer.style/react/drafts/IconButton)

- Documentation is incorrect as `import {IconButton} from '@primer/react/drafts'` doesn't work
- `import { IconButton } from '@primer/react/lib/Button2';` loses the styling
- Only `import { IconButton } from "@primer/react/lib-esm/drafts";` works. [Sandbox](https://codesandbox.io/s/objective-dirac-ihz1h?file=/src/App.js:67-125)
- SelectMenu is also totally broken when used with IconButton. [Sandbox](https://codesandbox.io/s/happy-faraday-vfq6v?file=/src/App.js)

See also: [Internal Slack thread](https://github.slack.com/archives/C01L618AEP9/p1640866149490200)

## [SelectMenu](https://primer.style/react/deprecated/SelectMenu)
- Is deprecated so I assumed `SelectPanel` should be used now despite it being in Alpha and you have to search it to find it.

## To Reproduce
- Use the latest release in your package.json
- Import any of the above components

## Expected behavior
Provide clear instructions on how to import the components or remove it from the documentation when they don't work and aren't ready yet. Including links in our [design guidelines](https://primer.style/design/components/action-list).

PR #1771, opened 2022-01-05
Can now export from @primer/react/drafts

pksjce:
Describe your changes here.

Right now, we need to import draft components from `@primer/react/lib-esm/drafts`. This is less than ideal.
In our docs, we have aliased `@primer/react` to the src folder relatively. So it works for our documentation site. 
But it clearly does not work elsewhere as demonstrated in https://github.com/primer/react/issues/1768

Our module definition right now defines `lib-esm/index.js` as the main module import in `package.json`. We need to alias `drafts` to work as expected.

One option is to use workspaces. I think this is unwieldy for our current usecase.
This [blogpost](https://arunmichaeldsouza.com/blog/aliasing-module-paths-in-node-js) suggested `module-alias`, but there's a new update to go for subpath-imports.

I'm trying it out with this PR. I need a build package to verify if it's working

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


pksjce:
# Update 

This does seem to work! Try it for yourself by using this canary version `@primer/react@0.0.0-20220572435`
https://codesandbox.io/s/objective-dirac-ihz1h?file=/src/App.js

siddharthkp (approved):
Yes!! Thank you for this <3

maximedegreve (approved):

maximedegreve:
Not 100% sure if it's working or if I'm doing this wrong but...

In my `package.json`:
`"@primer/react": "primer/react#pk/exports",`.

But then I get an import error:
`Module not found: Can't resolve '@primer/react' in '/workspaces/search/src'`

I also can't see a the `lib-esm` folder in `./node_modules/@primer/react` that is defined in `exports`.

maximedegreve (dismissed):
Not 100% sure if it's working or if I'm doing this wrong but...

In my `package.json` I've added:
`"@primer/react": "primer/react#pk/exports",`.

But then I get an import error:
`Module not found: Can't resolve '@primer/react' in '/workspaces/search/src'`

I also can't see a the `lib-esm` folder in `./node_modules/@primer/react` that is defined in `exports`.

**UPDATE:** @siddharthkp explained me that we don't support pointing to branches right now and I need to use the generated id however the same error still happens.

siddharthkp:
@pksjce Should we add IconButton to drafts as well?

siddharthkp:
Update: Aww, tried to use this with webpack and it doesn't pick it up automatically :( (I think it works on codesandbox because it uses sandbpack instead of webpack)

pksjce:
@siddharthkp - `IconButton` is already in there

Also yeah, this exports syntax seems to not be fully supported by webpack 5 and not at all supported in webpack 4. I wonder how the other bundlers will react to this either. Will investigate more.

Until then `import {..} from '@primer/react/lib-esm/drafts'` is what will work unfortunately.

maximedegreve:
Sounds good. Before we close this issue can we make sure we update all our docs to have those new instructions on how to import those components?

siddharthkp:
> can we make sure we update all our docs to have those new instructions on how to import those components?

I can create a PR :)

package.json
```diff
@@ -4,7 +4,6 @@
   "description": "An implementation of GitHub's Primer Design System using React",
   "main": "lib/index.js",
   "module": "lib-esm/index.js",
-  "type":"module",
```
siddharthkp: the commit message 🤣

pksjce:
hi @maximedegreve @siddharthkp. I got this to a working state! it works in both nextjs and cra on my testing.

`npm i @primer/react@0.0.0-202201893255`

drafts/package.json
```diff
@@ -0,0 +1,9 @@
+{
+    "_comment1": "this is required only for typescript. once this is fixed https://github.com/microsoft/TypeScript/issues/33079 we can remove this hack",
+    "name": "@primer/react/drafts",
+    "types": "../lib-esm/drafts/index.d.ts",
+    "main": "../lib-esm/drafts/index.js",
+    "type": "module",
+    "sideEffects": false
+  }
+
```
siddharthkp, line 9: ohhhhh, this is big brain time! 🧠

siddharthkp:
Tested in memex and a fresh next app,

This works now: 🎉 

```js
import { Button, ActionList, ActionMenu } from '@primer/react/drafts';
```

These old imports however break:

```js
import { Button, ActionList, ActionMenu } from '@primer/react/lib-esm/drafts';
import { ThemeProvider } from '@primer/react/lib/ThemeProvider';

// Module not found: Error: Package path ./lib-esm/drafts is not exported from package 
```

Is there a way to make this change not a breaking change? If not, we can look into shipping a codemod along with the release to make the migration easier.

pksjce:
@siddharthkp , Is this common usage? `import { AnchoredOverlay } from '@primer/react/lib-esm/AnchoredOverlay'` ?
I didn't know we allowed to dip into the build folder like that. I thought everyone had to export from `@primer/react` and we were trying to fix it such that `drafts` is accessible directly. 

My first thought would be that we shouldn't let it be common place to get components from `lib-esm` but I'm not sure if there are any pros to this. If we want each component to be available as it's own package, we should look into getting `lerna` to do that for us.

Else, I'm good to push out a codemod for this. I'm thinking I make it a folder in our codebase?

drafts/package.json
pksjce: Yeah, so without these all the imports work except the typescript build fails. This hack I found in @primer/behavior simply worked here.

siddharthkp:
@pksjce Unfortunately, yes [searching in memex](https://cs.github.com/?scopeName=All+repos&scope=&q=%40primer%2Freact%2Flib-esm%2F)

Unless this is for some specific reason, we could change ask to change the import statements (or automate them)

pksjce:
@siddharthkp , after much trial and error, I seem to have fixed this PR to make it backward compatible with the `lib-esm` folder imports. Adding 
```
    "./lib-esm/*": {
      "node": ["./lib/*.js","./lib/*/index.js"],
      "default": ["./lib-esm/*.js","./lib-esm/*/index.js"]
    }
```
seems to have done the trick. 
I tested this in memex in this [codespace](https://pksjce-github-memex-x554jw4c6prv.github.dev/) and in a nextjs app too.
I have also updated the documentation for the files in https://github.com/primer/react/pull/1785/files

siddharthkp (approved):
Let's gooooo!

PR #1772, opened 2022-01-05
Export SelectPanel from the package

pksjce:
As described in https://github.com/primer/react/issues/1768, SelectPanel is not exported nor is the documentation displayed. 
But the [component status page](https://primer.style/react/status) lists it as alpha.
Since it's quite an old component, I'm thinking we missed exporting it.

lmk if its otherwise.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -137,6 +137,8 @@ export type {
   SelectMenuTabPanelProps,
   SelectMenuLoadingAnimationProps
 } from './SelectMenu'
+export {SelectPanel} from './SelectPanel'
+export type {SelectPanelProps} from './SelectPanel'
 export {default as SideNav} from './SideNav'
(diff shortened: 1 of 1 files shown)


maximedegreve (approved):

src/index.ts
```diff
@@ -137,6 +137,8 @@ export type {
   SelectMenuTabPanelProps,
   SelectMenuLoadingAnimationProps
 } from './SelectMenu'
+export {SelectPanel} from './SelectPanel'
+export type {SelectPanelProps} from './SelectPanel'
```
siddharthkp, line 141: not blocker, up to you: should we skip adding SelectPanel v1 to the main bundle and directly add v2. One less deprecation?

maximedegreve:
This works! Just tested it. 🎉

PR #1778, opened 2022-01-07
Docs: change import to lib-esm for drafts

siddharthkp:
Fix import path in docs/examples. draft components cannot be imported from `@primer/react/drafts`, you have to mention the format as well: `@primer/react/lib-esm/drafts` or `@primer/react/lib/drafts`.

More context: https://github.com/primer/react/issues/1768

```js
// before: 
import {Button} from '@primer/react/drafts'

// after: 
import {Button} from '@primer/react/lib-esm/drafts'
```

```jsx
// before: 

<drafts.Button>Default</drafts.Button>

// after:

// import {Button} from '@primer/react/lib-esm/drafts'
const {Button} = drafts // ignore docs silliness; import like that ↑

render(<Button>Default</Button>)
```

### DIFF
--- a/docs/content/drafts/ActionList2.mdx
+++ b/docs/content/drafts/ActionList2.mdx
@@ -43,7 +43,7 @@ import InlineCode from '@primer/gatsby-theme-doctocat/src/components/inline-code
 </Box>
 
 ```js
-import {ActionList} from '@primer/react/drafts'
+import {ActionList} from '@primer/react/lib-esm/drafts'
 ```
 
 ## Examples
@@ -51,7 +51,7 @@ import {ActionList} from '@primer/react/drafts'
 ### Minimal example
 
 ```javascript live noinline
-// import {ActionList} from '@primer/react/drafts'
+// import {ActionList} from '@primer/react/lib-esm/drafts'
 const {ActionList} = drafts // ignore docs silliness; import like that ↑
 
 render(
@@ -71,7 +71,7 @@ Leading visuals are optional and appear at the start of an item. They can be oct
 
 <!-- prettier-ignore -->
 ```javascript live noinline
-// import {ActionList} from '@primer/react/drafts'
+// import {ActionList} from '@primer/react/lib-esm/drafts'
 const {ActionList} = drafts // ignore docs silliness; import like that ↑
 
 render(
@@ -97,7 +97,7 @@ render(
 Trailing visual and trailing text can display auxiliary information. They're placed at the right of the item, and can denote status, keyboard shortcuts, or be used to set expectations about what the action does.
 
 ```javascript live noinline
-// import {ActionList} from '@primer/react/drafts'
+// import {ActionList} from '@primer/react/lib-esm/drafts'
 const {ActionList} = drafts // ignore docs silliness; import like that ↑
 
 render(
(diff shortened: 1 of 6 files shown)


docs/content/drafts/ActionList2.mdx
```diff
@@ -43,15 +43,15 @@ import InlineCode from '@primer/gatsby-theme-doctocat/src/components/inline-code
...
+import {ActionList} from '@primer/react/lib-esm/drafts'
 ```
 
 ## Examples
 
 ### Minimal example
 
 ```javascript live noinline
-// import {ActionList} from '@primer/react/drafts'
+// import {ActionList} from '@primer/react/lib-esm/drafts'
```
rezrah, line 54: Can we remove these lines altogether, as it's duplicated in the Installation section at the top of this page?

docs/src/@primer/gatsby-theme-doctocat/nav.yml
```diff
@@ -129,6 +129,18 @@
       url: /Truncate
     - title: UnderlineNav
       url: /UnderlineNav
+- title: Drafts
```
rezrah, line 132: How do you feel about the IA here @siddharthkp? For easier discovery, would it make sense to co-locate it to the original component docs page under an experimental tab/panel/label?

rezrah (approved):
Looks good. 

Only other question I have is what the plan is for the CommonJS version? Your docs only cover `lib-esm` import methods now, which may be most popular mechanism but raises others concerns.

The DX is quite poor if we need to ask user to add `lib-esm` to path each time. If we don't export these components in CJS, or if we have low utilisation, shouldn't we scrap the CommonJS folder or flip them so that `lib-esm` is the default format.

docs/content/drafts/ActionList2.mdx
siddharthkp: I've kept them around for now because the next line isn't how you would import in the application, but that's how we do it in the docs.

```js
const {ActionList} = drafts // ignore docs silliness; import like that ↑
```

I'd like to hide that implementation detail silliness altogether. Don't know how to do it yet, would work on it on another PR

siddharthkp:
> Your docs only cover lib-esm import methods now, which may be most popular mechanism but raises others concerns.

We do have it export it as commonjs as well ([package on unpkg](https://unpkg.com/browse/@primer/components@33.1.0/)), but we don't have that in the docs. I think we can add it to the top in the first import?

docs/content/drafts/ActionList2.mdx
rezrah: Yeah, was suggesting we remove both import lines here.. as the live code snippet works fine without them.

Something like this 👇 , although would be nice to get rid of the render function too if we can.

Screenshot 2022-01-11 at 13 05 54

siddharthkp: If we remove both those lines, it picks Button v1 from `@primer/components`, not the one from drafts 😅, that's we import from drafts.

This PR would help remove the lines though :) https://github.com/primer/react/pull/1785

docs/src/@primer/gatsby-theme-doctocat/nav.yml
siddharthkp: Ooh, I like that idea 🤔 

We loose the section heading of "Drafts", but that's probably okay

| Different section        | Collocated           |
| ------------- |-------------|
| image | image |

siddharthkp:
Closed in favor of https://github.com/primer/react/pull/1785

docs/src/@primer/gatsby-theme-doctocat/nav.yml
siddharthkp: I thought about it a bit more and decided that we should put all the drafts together. When a draft component is ready to be used in production (design and a11y reviewed, then we should move it out of drafts and deprecate the previous one.)

PR #1781, opened 2022-01-07
Fix: restore color prop functionality for StyledOcticon

rezrah:
Restores `color` prop functionality to `StyledOcticon`, which now works with both native and functional colours.

### Other changes:

- Docs updated to reflect correct usage
  - Updated prop tables to new format
  - Added component status

#### Before
Screenshot 2022-01-07 at 14 24 36

#### After
Screenshot 2022-01-07 at 14 24 47

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


maximedegreve (approved):

siddharthkp (approved):
Nice! Ship it

siddharthkp:
Sidenote: How does https://github.com/primer/eslint-plugin-primer-react know that it has to ignore `color` on this specific component?

rezrah:
> Sidenote: How does https://github.com/primer/eslint-plugin-primer-react know that it has to ignore `color`, which resembles a styled-system prop, on this specific component?

Looking at the rule source, there's an [exclusion](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/rules/no-system-props.js#L20) but interestingly `color` isn't defined there. It also looks like it's looking for a type signature that matches the one exported by styled-system, which this doesn't. It would probably be a bit limiting if we couldn't use props like `color` anymore tbh.

siddharthkp:
> It also looks like it's looking for a type signature that matches the one exported by styled-system, which this doesn't.

Trying to understand the source, are you referring to this line? [rules/no-system-props.js#L66](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/rules/no-system-props.js#L66):

```js
let systemProps = Object.values(pick(propsByNameObject))
```

definition of pick: [@styled-system/props/index.js#L45-L52](https://github.com/styled-system/styled-system/blob/master/packages/props/src/index.js#L45-L52)

PR #1786, opened 2022-01-11
CI: disable visual testing

siddharthkp:
Remove visual testing action till our limit is reset [more context on slack](https://app.slack.com/client/T0CA8C346/C02NUUQ9C30).

To enable it again, we can revert this PR :)


rezrah:
To make it easier to renable this without a code change in future, can we keep the workflow in repo and disable it via click-ops instead? https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

siddharthkp:
That's cool too! We don't need this PR :)

PR #1799, opened 2022-01-14
Storybook: Create 2 category of stories for ActionList2: Examples & Fixtures)

siddharthkp:
Context: We use storybook for multiple purposes
- isolated dev environment
- documenting edge cases
- fixtures for visual testing
- fixtures for automated accessibility testing (axe)

For design and accessibility reviews, this is a lot of input, most of which isn't very useful.

Solution: Split stories into `examples` and `fixtures`. Examples are scenarios that might see in production. Fixtures are everything else that are only interesting to maintainers of primer-react.

Bonus points if we can match examples documented in primer-css's storybook

Example: ActionList2 has 6 examples and 21 fixtures.

List of 6 examples inside ActionList2

List of 21 fixtures inside ActionList2


langermank:
This is cool Sid! What do you think about calling fixtures "features"? In Primer CSS we create one base "Playground" story that includes all the base props as controls. Then we split specific prop combinations into a Features folder (especially because it tends to be long.) We use Patterns the same way I think, more like contextual or compound examples :D 

storybook nav

alliethu:
@siddharthkp Thank you for making this change and for collaborating with @langermank! I hope we can carry this pattern through all instances of the components!

siddharthkp:
@langermank Hi! Renamed patterns to examples 👍 

> What do you think about calling fixtures "features"? 

I went with `fixtures` because we use these as fixtures for test coverage and axe violations; and for testing edge cases (for example, [this story 1](https://primer.style/react/storybook?path=/story/composite-components-actionlist2--with-sx) and [this story 2](https://primer.style/react/storybook?path=/story/composite-components-actionlist2--all-combinations) are to test edge cases, we don't recommend folks use the component like this 😅)

That being said, I'm okay with calling it features, as long as we all know what that means to us. What do you think?

pksjce:
Hey Sid, this is super cool. This could also help integration testing. Having too many of them is a pain, so we could just run tests over the production scenarios.
I'm also thinking of adding a separate section for perf analysis storybooks. WDYT?

rezrah (approved):
Love it ✨ 

I think `fixtures` make sense here, particularly if it becomes the input data for automation runs eventually. 

Could we also capture this new IA and naming convention in the `Contributing.md` please? Maybe also record it in a small ADR, so that we can refer back to this supporting documentation when/if we need to backdate the other stories at some point.

siddharthkp:
> This could also help integration testing. Having too many of them is a pain, so we could just run tests over the production scenarios.

Makes a lot of sense!

> I'm also thinking of adding a separate section for perf analysis storybooks. WDYT?

Do you want to collocate them with the component or have a different group with fixtures for performance testing?

PR #1804, opened 2022-01-17
Fixes PointerBox caret not styling correctly using sx props

### LINKED ISSUE #1508: PointerBox needs to support `sx` props
**Describe the bug**
`PointerBox` only allows use of `bg`, `borderColor`, etc. props to style it, but as advised, we are migrating to use `sx` instead. In this instance, it breaks the styling because `sx` props are not accounted for. Also, unless `position: relative` is supplied in the `sx` object, the Caret will render in  the top left of the page as it blows away the local `sx` prop in the component, rather than doing a spread/object merge.

**To Reproduce**
Instantiate a `PointerBox` component using `sx={{bg: '#ff0000'}}` and notice that the background does not go red.

**Expected behavior**
It should either exclusively look for `sx` styling _or_ use an either/or logic block to style the Caret

**Screenshots**
image

rezrah:
Doing some housekeeping to close down old issues 🧹 

Closes #1508

- Fixes bug by passing custom styles through to the `Caret`.
  - Removed dependency on  `BorderedBox` including broken documentation link
  - Kept API backwards-compatible for now to avoid a breaking change.
- Updated docs for `PointerBox` to new layout for consistency
- Updated docs to remove mentions of system props, although they are still available on the API

### Screenshots

Before:
Before

After:
After

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/PointerBox.tsx
```diff
@@ -1,30 +1,45 @@
...
+import {SxProp} from './sx'
+
+// FIXME: Make this work with BetterStyledSystem types
+type MutatedSxProps = {
+  sx?: {
+    bg?: string
+    backgroundColor?: string
+    borderColor?: string
+  }
+} & SxProp
```
pksjce, line 13: wow i wonder if this can be removed?

rezrah: 🤷‍♂️  any ideas? I can't seem to coerce the types to match with sx without doing this. Couldn't find any other similar patterns either, so stuck a FIXME on it for now to come back to later. Would appreciate any suggestions 🙏

docs/content/PointerBox.md
```diff
@@ -72,12 +76,37 @@ function CaretSelector(props) {
 render(<PointerBoxDemo />)
 ```
 
-## System props
+## Props
+
+<PropsTable>
+  <PropsTableRow
+    name="caret"
+    type="top| top-left | top-right | right | right-top | right-bottom | bottom | bottom-left | bottom-right | left | left-top | left-bottom"
```
colebemis, line 84:
```suggestion
    type={`| 'top'
| 'top-left'
| 'top-right'
| 'right'
| 'right-top'
| 'right-bottom'
| 'bottom'
| 'bottom-left'
| 'bottom-right'
| 'left'
| 'left-top'
| 'left-bottom'`}
```

docs/content/PointerBox.md
```diff
@@ -72,12 +76,37 @@ function CaretSelector(props) {
...
+    defaultValue="bottom"
```
colebemis, line 85:
```suggestion
    defaultValue="'bottom'"
```

rezrah: Fixed. Thanks @colebemis. Is this g2g now?
Screenshot 2022-01-24 at 21 47 00

mperrotti (approved):

PR #1806, opened 2022-01-18
Temporarily hide form field layout components from React docs sidebar

mperrotti:
There has been some feedback that the components' names/API are unclear. We want to avoid consumers using these components until we resolve any issues that will cause breaking changes.

This is going to be discussed in the 01/18/2022 Primer Patterns sync.


siddharthkp (approved):
Fair play 👍

PR #1810, opened 2022-01-19
ActionMenu2 + DropdownMenu2: Initial focus is based on key used to open menu

siddharthkp:
- Fixes https://github.com/github/primer/issues/586

Checklist:
- [x] Created new hook
- [x] Updated ActionMenu2 to use hook
- [x] Updated DropdownMenu2 to use hook
- [x] Added tests for hook

<strong>Notes for reviewer:</strong> Would have been really cool to use `focusInStrategy` in [`@primer/behaviors/focusZone`](https://github.com/primer/behaviors/blob/main/docs/focus-zone.md) but unfortunately that only kicks once the focus zone (overlay in this case) is open and initialised; And because of that, it doesn't have the information about the trigger keyboard event. To make that happen, we would need to rewire how AnchoredOverlay works, so that it can attach listeners even when it's closed. I didn't feel super confident about doing that, so create a separate hook that only cares about initial focus.


src/hooks/useMenuFocus.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  React.useEffect(() => {
+    if (!open) return
+    if (!openingKey || !containerRef.current) return
+
+    const iterable = iterateFocusableElements(containerRef.current)
+    if (['ArrowDown', 'Space', 'Enter'].includes(openingKey)) {
+      const firstElement = iterable.next().value
+      /** We push imperative focus to the next tick to prevent React's batching */
+      setTimeout(() => firstElement?.focus())
```
siddharthkp, line 31: As the comment suggests, we push this to next tick so that the menu has opened.

colebemis (approved):
Looks great 👍 

Tested in Storybook and everything seems to work 🌟

PR #1812, opened 2022-01-20
Add Textarea component

rezrah:
Adds a `Textarea` component.

Try it here: [Docs](https://primer-react-1jd0ylb9h-primer.vercel.app/react/Textarea) | [Storybook](https://primer-react-1jd0ylb9h-primer.vercel.app/react/storybook?path=/story/forms-textarea--default)

### Screenshots
Docs

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/TextArea.tsx
```diff
@@ -0,0 +1,164 @@
...
+   */
+  block?: boolean
+  /**
+   * Allows resizing of the textarea
+   */
+  resize?: 'none' | 'both' | 'horizontal' | 'vertical'
+  /**
+   * Applies a custom width to the textarea
+   */
+  width?: number
```
mperrotti, line 31: We don't have a `width` prop for with Select or TextInput. Why would we have it for TextArea?

src/TextArea.tsx
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Applies a custom height to the textarea
+   */
+  height?: number
```
mperrotti, line 35: Instead of `height`, maybe we should stick closer to the HTML API and use `rows?: number`.

mperrotti: We could take it a step further and say that the textarea would render with the specified number of `rows`, but if the user input goes beyond that, the height would adjust to fit the text.

If we wanted to prevent a textarea from growing too high, we could have a `maxRows` prop that limits the height of the textarea. After the user inputs a number of lines that is greater than `maxRows`, it would just scroll vertically.

src/TextArea.tsx
```diff
@@ -0,0 +1,164 @@
...
+} & TextareaHTMLAttributes<HTMLTextAreaElement> &
+  SxProp
+
+const StyledTextArea = styled.textarea<TextAreaProps>`
+  min-height: 124px;
```
mperrotti, line 40: Where does "124px" come from?

src/TextArea.tsx
```diff
@@ -0,0 +1,164 @@
...
+  font-family: inherit;
+  font-size: ${get('fontSizes.1')};
+  line-height: 20px;
```
mperrotti, line 43: Should we have primitives for line-height? cc @colebemis 

(feedback isn't specifically relevant to this PR)

src/TextArea.tsx
```diff
@@ -0,0 +1,164 @@
...
+   */
+  width?: number
+  /**
+   * Applies a custom height to the textarea
+   */
+  height?: number
+} & TextareaHTMLAttributes<HTMLTextAreaElement> &
+  SxProp
+
+const StyledTextArea = styled.textarea<TextAreaProps>`
```
mperrotti, line 39: The Select and TextInput component both just use the private `TextInputWrapper` component to maintain consistent styling. I'd recommend we do that here too.

src/__tests__/__snapshots__/TextArea.test.tsx.snap
```diff
@@ -0,0 +1,67 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
mperrotti, line 1: Why do we need snapshots?

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+    name="block"
+    type="boolean"
+    defaultValue="false"
+    description="Expands with width of the component to fill the parent elements"
+  />
+  <PropsTableRow
+    name="block"
+    type="boolean"
+    defaultValue="false"
+    description={<>Creates a full width input element</>}
```
mperrotti, line 134: 1. Why does this need to be in a fragment?
2. Let's use this description in the JSDoc comment in the TextArea's `Props` type

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
+---
+componentId: textarea
+title: TextArea
```
colebemis, line 3: Is textarea one or two words?

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+```jsx live
+<TextArea resize="none" height={400}></TextArea>
+```
+
+### Custom styling
+
+```jsx live
+<>
+  <TextArea sx={{marginBottom: 2}}></TextArea>
```
colebemis, line 108:
```suggestion
  <TextArea sx={{marginBottom: 2}} />
```

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+### Resizing and constraining dimensions
+
+By default, `TextArea` can be resized by the user vertically and horizontally.
+
+Resizing can be prevented by setting `resize` to `none`, with `height` and `width` props available for granular control of its appearance.
+Setting an explicit `height` or `width` can prevent its respective `resize` options from being applied.
+
+```jsx live
+<TextArea resize="none" height={400}></TextArea>
```
colebemis, line 101:
```suggestion
<TextArea resize="none" height={400} />
```

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+}
+
+render(<TextAreaExample />)
+```
+
+### Displaying form validation state
+
+```jsx live
+<>
+  <Box as="label" for="success-state" sx={{display: 'block'}}>
```
colebemis, line 74:
```suggestion
  <Box as="label" htmlFor="success-state" sx={{display: 'block'}}>
```

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+    Success state:
+  </Box>
+  <TextArea id="success-state" validationStatus="success" />
+  <Box as="label" for="error-state" sx={{display: 'block', mt: 5}}>
```
colebemis, line 78:
```suggestion
  <Box as="label" htmlFor="error-state" sx={{display: 'block', mt: 5}}>
```

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+```javascript live noinline
+const TextAreaExample = () => {
+  // Running in controlled mode (recommended)
+  const [value, setValue] = React.useState('')
+
+  const handleChange = event => {
+    setValue(event.target.value)
+  }
+
+  return <TextArea onChange={handleChange}>{value}</TextArea>
```
colebemis, line 36: Is `<TextArea>{value}</TextArea>` equivalent to `<TextArea value={value} />`?

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+  <PropsTableRow
+    name="block"
+    type="boolean"
+    defaultValue="false"
+    description={<>Creates a full width input element</>}
+  />
+  <PropsTableRow
+    name="resize"
+    type="'both' | 'horizontal' | 'vertical' | 'none'"
+    defaultValue="both"
```
colebemis, line 139:
```suggestion
    defaultValue="'both'"
```

src/TextArea.tsx
colebemis: +1 for `rows/cols` (which is natively supported by textarea). If a consumer wants to explicitly set the `width/height`, they can use the `sx` prop

src/TextArea.tsx
colebemis:
```suggestion
  line-height: ${get('lineHeights.condensed')};
```

We have lineHeights available in the theme (These should be lifted into primitives eventually)

colebemis:
```
"lineHeights": {
    "condensedUltra": 1,
    "condensed": 1.25,
    "default": 1.5
  },
```

src/TextArea.tsx
```diff
@@ -0,0 +1,164 @@
...
+    return (
+      <StyledTextArea
+        ref={ref}
+        disabled={disabled}
+        sx={sxProp}
+        validationStatus={validationStatus}
+        resize={resize}
+        width={width}
+        height={height}
+        aria-required={required ? 'true' : 'false'}
```
colebemis, line 154: What's the difference between `aria-required` and [`required`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-required)?

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+<PropsTable>
+  <PropsTableRow
+    defaultValue="undefined"
+    name="required"
+    type="boolean"
+    description="Indicates to the user and assistive technologies that the field value is required"
+  />
+  <PropsTableRow
+    name="block"
```
colebemis, line 125: I wonder if we should remove the `block` prop in favor of `sx={{width: '100%'}}` (in both TextArea and TextInput) cc @mperrotti

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+    defaultValue="undefined"
+    type="number"
+    description="Apply an optional, explicit width"
+  />
+  <PropsTableRow name="validationStatus" 
+    defaultValue="undefined" 
+    type="'success' | 'error' | undefined" 
+    description="Style the textarea to match the current form validation status"
+  />
+  <PropsTablePassthroughPropsRow
```
colebemis, line 159:
```suggestion
  />
  <PropsTableRefRow
    elementType="textarea"
    refType="HTMLTextAreaElement"
  />
  <PropsTablePassthroughPropsRow
```

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+</>
+```
+
+## Props
+
+### TextArea
+
+<PropsTable>
+  <PropsTableRow
+    defaultValue="undefined"
```
colebemis, line 119:
```suggestion
    defaultValue="false"
```

Feels a little strange for the default value to be `undefined` when the type is `boolean`

docs/content/TextArea.mdx
mperrotti: I'd be down with that if that's all `block` is doing

src/TextArea.tsx
mperrotti: I think they have different levels of screenreader support. I think `aria-required` is the preferred one. I bet somebody on the accessibility team would know.

cc @alliethu

docs/content/TextArea.mdx
mperrotti: In the React element types, they call it "TextArea". I also thought the CamelCasing looked funny here 🤷‍♂️

src/TextArea.tsx
rezrah: So rows/cols are still supported natively and for most users would be perfectly fine. It's definitely a bit confusing here because it’s not explicitly listed in the prop table, so it might appear like I was recommending one over the other. I don't agree that they are equivalent to explicit height and width however, which provide more granular control over appearance. 

Why not just use `sx` for this? Maybe 🤷 . I feel like we over-rely on `sx` too much already to handle common use-cases, which is a sentiment I’ve heard on a few occasions by users. So height and width were really just there for improved ergonomics and convenience.

src/TextArea.tsx
rezrah: It's the height that's set on the text area component in Figma. Took it from there.

src/TextArea.tsx
rezrah: I did think about this, and decided against it as that component had styling that wasn't relevant for this component, like warning state, size variants etc. Do we want to inherit those here? It feels like we'd be creating a dependency here for the sake of convenience, but potentially more detrimental to the user as we're loading unnecessary css and subsequently delivering a less performant component.

src/__tests__/__snapshots__/TextArea.test.tsx.snap
rezrah: We don't. I can remove this. Are we all happy to remove snapshots going forward? I asked this previously and majority asked to keep it for visual regression purposes.

docs/content/TextArea.mdx
rezrah: 👍 it doesn't, will remove it. I was copy/pasting too fast 😓

docs/content/TextArea.mdx
rezrah: I think it should be two words really, but HTML has perhaps conditioned us to believe it's one? I went with `TextArea`  because 1) If they are two words, pascalcase makes sense. 2) We call a similar component `TextInput`. 

Not precious either way. As @mperrotti pointed out, React internally refers to it in PascalCase too.

src/TextArea.tsx
rezrah: This was an oversight on my part, I was actually meant to include `required` too. I don't think it hurts to add both, but I have seen some data indicating `aria-*` offers improved support too.

docs/content/TextArea.mdx
rezrah: Feels like a very common use-case. Are we over-relying on `sx` here to handle scenarios many users would expect to be done via props?

docs/content/TextArea.mdx
rezrah: It's because it's optional, which becomes by default equivalent to `undefined`. I asked this on a different PR, but don't think there was a definitive answer. Happy to remove it.

docs/content/TextArea.mdx
colebemis: I'm okay with a prop for this use case if we think it's pretty common. I'm not the biggest fan of `block` as a name for this though. What do y'all think?

docs/content/TextArea.mdx
siddharthkp, line 36: 2 notes on the default example:

1. We should have a placeholder
2. Should it be a simple uncontrolled component?  just `<TextArea placeholder="something something">`

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+  </Box>
+  <TextArea id="success-state" validationStatus="success" />
+  <Box as="label" for="error-state" sx={{display: 'block', mt: 5}}>
+    Error state:
+  </Box>
+  <TextArea id="error-state" validationStatus="error" />
+</>
+```
+
+### Inactive
```
siddharthkp, line 85: Can we call this disabled?

docs/content/TextArea.mdx
```diff
@@ -0,0 +1,187 @@
...
+    description="Indicates to the user and assistive technologies that the field value is required"
+  />
+  <PropsTableRow
+    name="block"
+    type="boolean"
+    defaultValue="false"
+    description="Expands with width of the component to fill the parent elements"
+  />
+  <PropsTableRow
+    name="block"
```
siddharthkp, line 131: heads up: There are 2 `block` rows  

image

rezrah: thanks @siddharthkp - removed it 🙇

docs/content/TextArea.mdx
rezrah: Can do. It's for more inclusive language. This came up on product-design slack around accessible language, for example: https://polaris.shopify.com/content/accessible-and-inclusive-language#enable-disable. I think it's a good idea here, but the underlying `disabled` attribute hasn't changed, just the contextual semantics.

docs/content/TextArea.mdx
rezrah: Should we encouraging using placeholders by putting them in our examples? @alliethu [wrote up some feedback](https://github.com/primer/figma-post-to-github/issues/3#issue-1052084644) previously suggesting we shouldn't use them.

On point 2, I think we should encourage controlled over uncontrolled, but both examples exist. I also think our examples could generally be a bit more copy/paste-able so I went with full implementation detail. 

Good point though, I'll add a ultra minimal example without state handling too 👍

siddharthkp: re: placeholder: I read @alliethu's feedback as "don't use placeholders as labels, use labels." I think we should still use placeholders for giving examples. cc @alliethu what do you think?

src/TextArea.tsx
rezrah: I've now removed this min-height, so the user is forced to set their own. Was the textarea in Figma just an example, or intentional min-height @mperrotti?

src/TextArea.tsx
alliethu: @rezrah has it right. It's suggested to use both for wider browser/screen reader support. :)

docs/content/TextArea.mdx
alliethu: @siddharthkp, I think it's context dependent. Lots of times placeholders used as examples, are unnecessary as the label is straightforward enough. For text areas, however, I can imagine a placeholder being helpful in guiding the user on what type of content may be expected. 

@rezrah this would be a great example of how shifting left could solve for these types of questions early on. 😄

src/TextArea.tsx
mperrotti: I still prefer `rows` over `height` because it sets the height to fit a specific number of lines.

I agree that `cols` feels weird, and I'd be happy with using `sx` to set width. I assume `cols` was added to support vertical writing modes.

src/TextArea.tsx
mperrotti: If we end up using `rows` instead of `height`, I'd set the a default value for rows instead of specifying a min-height.

Otherwise, just make sure 124px perfectly fits {n} lines of text and doesn't cut off any lines of text in their vertical middle.

src/TextArea.tsx
mperrotti: You're right - the size variants aren't relevant. Most of the others are though:
| Styling prop | Used for TextArea? |
| ------------ | ------------------- |
| disabled | ✅ |
| hasLeadingVisual | ❌ |
| hasTrailingVisual | ❌ |
| block | ✅ |
| contrast | ✅ |
| validationStatus | ✅ |
| size | ❌ |

src/__tests__/__snapshots__/TextArea.test.tsx.snap
mperrotti: I think now that we're using Chromatic, they're not necessary.

docs/content/TextArea.mdx
mperrotti: I agree that `block` is a little unclear, but I'd prefer we change the naming for all components that use `block` if we choose to make it different here.

src/__tests__/__snapshots__/TextArea.test.tsx.snap
rezrah: Chromatic is disabled atm

docs/content/TextArea.mdx
rezrah: Thanks all. Updated the basic example with a placeholder.

src/__tests__/__snapshots__/TextArea.test.tsx.snap
mperrotti: Ah that's right. Ok, then let's keep them until we re-enable Chromatic.

src/TextArea.tsx
rezrah: height has been removed 👍

src/TextArea.tsx
mperrotti: I still think we should DRY this by using `TextInputWrapper` or something that replaces it. It's unlikely somebody is going to load the CSS for a `TextArea` and not also have a component on the page that loads `TextInputWrapper` styles (e.g.: `TextInput` or `Select`). Then, we'd downloading nearly identical styles twice.

mperrotti (commented):
Everything is looking good, there's just that one styling sugggestion.

src/_TextInputWrapper.tsx
```diff
@@ -38,66 +39,55 @@ const sizeVariants = variant({
...
+  variant?: 'small' | 'large'
+  size?: 'small' | 'large'
+} & StyledBaseWrapperProps
 
-const TextInputWrapper = styled.span<StyledWrapperProps>`
-  min-height: 32px;
+const textInputBasePadding = '12px'
+export const textInputHorizPadding = textInputBasePadding
+
+export const TextInputBaseWrapper = styled.span<StyledBaseWrapperProps>`
```
rezrah, line 63: In order to reuse `TextInputWrapper`, I've abstracted the base styles into `TextInputBaseWrapper`, which apply universally to all types of text input, including textarea.

src/utils/types/FormValidationStatus.ts
```diff
@@ -1 +1 @@
-export type FormValidationStatus = 'error' | 'success'
+export type FormValidationStatus = 'error' | 'success' | 'warning' // TODO: remove warning as a breaking change. Added for backwards compatibility with TextArea
```
rezrah, line 1: Not ideal, but this avoids a breaking change to TextInput by leaving `warning` in play. I'd recommend we use the upcoming form validation work to remove `warning` altogether and defer the breaking change.

src/TextArea.tsx
```diff
@@ -0,0 +1,102 @@
...
+ */
+const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>(
+  (
+    {
+      value,
+      disabled,
+      sx: sxProp,
+      required,
+      validationStatus,
+      rows = 7,
```
mperrotti, line 73:
```suggestion
      rows = 4,
```

7 seems a bit big for the default

rezrah: It was the closest to the previous `124px` in the Figma file, 4 is basically going back to default. Is that okay?

src/TextArea.tsx
```diff
@@ -0,0 +1,102 @@
...
+    props.resize &&
+    css`
+      resize: ${props.resize};
+    `}
+
+  ${props =>
+    props.disabled &&
+    css`
+      resize: none;
+      cursor: not-allowed;
```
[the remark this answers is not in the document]
rezrah: We do this for checkbox and radio too 😄

src/TextArea.tsx
mperrotti: Yea, that's ok

src/TextArea.tsx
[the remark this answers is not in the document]
mperrotti: Ah ok. Should we move this style to `TextInputBaseWrapper` then? That way we'll also get this behavior in the Select and TextInput components.

rezrah:
@mperrotti @colebemis FYI, I've renamed everything to `Textarea`. Seems there's ecosystem alignment for this too, so happy to make that change.

Issue #1783, opened 2022-01-10
Button state selected


YousefED:
How do I use the React components to show a button as Selected?

With the CSS Primer, this is demonstrated here: https://primer.style/css/components/buttons#selected

![image](https://user-images.githubusercontent.com/368857/148786436-48fe4371-3fd4-4ed4-8dcc-2ece827d9c26.png)

mperrotti:
We don't currently have a way to do this.

@pksjce - this is something to consider with the work you're doing on buttons right now

siddharthkp:
Hi @YousefED, (just curious) where do you use this pattern?

pksjce:
We didn't consider it originally. Though the solution seems to be just to add background and box shadow, do some testing and push it both for Button and Button2

PR #1819, opened 2022-01-24
Add selected state to Button

pksjce:
Describe your changes here.

I have added `selected` state as described in https://primer.style/css/components/buttons#selected. I used the same styles as https://github.com/primer/css/blob/e8d5402a26f0e4428cb0cd0694569da3c2ed505e/src/buttons/button.scss#L86

I have implemented this for both `Button` and `Button2`.

Closes # (https://github.com/github/primer/issues/639)

### Screenshots

Checkout the storybook in 

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/drafts/Button2.mdx
+++ b/docs/content/drafts/Button2.mdx
@@ -65,6 +65,20 @@ The `invisible` variant of `Button` indicates that the action is a low priority 
 </>
 ```
 
+### Selected button
+
+`Button` can be set to in a "selected" state in two ways. You can use a `selected` class or set `aria-selected ="true"`.
+This is in many cases similar to the active state of the button. Usually used in `ButtonGroup` when like tabs.
+
+```jsx live
+<>
+  <drafts.Button aria-selected="true">Search</drafts.Button>
+  <drafts.Button aria-selected="true" variant="danger">
+    Delete
+  </drafts.Button>
+</>
+```
+
 ### Appending an icon
 
 We can place an inside the `Button` in either the leading or the trailing position to enhance the visual context.
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -31,6 +31,11 @@ const Button = styled(ButtonBase)<ButtonBaseProps & SxProp>`
     background-color: ${get('colors.btn.bg')};
     border-color: ${get('colors.btn.border')};
   }
+  &.selected,
+  &[aria-selected='true'] {
+    background-color: ${get('colors.btn.selectedBg')};
+    box-shadow: ${get('colors.primer.shadow.inset')};
+  }
 
   ${sx};
 `
--- a/src/Button/ButtonDanger.tsx
+++ b/src/Button/ButtonDanger.tsx
@@ -22,7 +22,9 @@ const ButtonDanger = styled(ButtonBase)<ButtonBaseProps & SxProp>`
     box-shadow: ${get('shadows.btn.danger.focusShadow')};
   }
 
-  &:active {
+  &:active,
+  &.selected,
+  &[aria-selected='true'] {
     color: ${get('colors.btn.danger.selectedText')};
     background-color: ${get('colors.btn.danger.selectedBg')};
     box-shadow: ${get('shadows.btn.danger.selectedShadow')};
--- a/src/Button/ButtonInvisible.tsx
+++ b/src/Button/ButtonInvisible.tsx
@@ -20,7 +20,9 @@ const ButtonInvisible = styled(ButtonBase)<ButtonBaseProps & SxProp>`
   &:hover {
     background-color: ${get('colors.btn.hoverBg')};
   }
-  &:active {
+  &:active,
+  $.selected,
+  &[aria-selected='true'] {
     background-color: ${get('colors.btn.selectedBg')};
   }
 
--- a/src/Button/ButtonOutline.tsx
+++ b/src/Button/ButtonOutline.tsx
@@ -22,7 +22,9 @@ const ButtonOutline = styled(ButtonBase)<ButtonBaseProps & SxProp>`
     box-shadow: ${get('shadows.btn.outline.focusShadow')};
   }
 
-  &:active {
+  &:active,
+  $.selected,
+  &[aria-selected='true'] {
     color: ${get('colors.btn.outline.selectedText')};
     background-color: ${get('colors.btn.outline.selectedBg')};
     box-shadow: ${get('shadows.btn.outline.selectedShadow')};
--- a/src/Button/ButtonPrimary.tsx
+++ b/src/Button/ButtonPrimary.tsx
@@ -22,7 +22,9 @@ export const ButtonPrimary = styled(ButtonBase)<ButtonBaseProps & SxProp>`
     box-shadow: ${get('shadows.btn.primary.focusShadow')};
   }
 
-  &:active {
+  &:active,
+  $.selected,
+  &[aria-selected='true'] {
     background-color: ${get('colors.btn.primary.selectedBg')};
     box-shadow: ${get('shadows.btn.primary.selectedShadow')};
   }
--- a/src/Button2/Button2.stories.tsx
+++ b/src/Button2/Button2.stories.tsx
@@ -25,6 +25,12 @@ export default {
         type: 'radio',
         options: ['small', 'medium', 'large']
       }
+    },
+    'aria-selected': {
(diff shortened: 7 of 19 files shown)


docs/content/drafts/Button2.mdx
```diff
@@ -65,6 +65,20 @@ The `invisible` variant of `Button` indicates that the action is a low priority
 </>
 ```
 
+### Selected button
+
+`Button` can be set to in a "selected" state in two ways. You can use a `selected` class or set `aria-selected ="true"`.
```
rezrah, line 70: Is it worth adding a `selected` prop that handles both styling and setting aria attributes? May make consumer-side implementation easier. Also, I don't usually ask for users to set classes elsewhere to enable features, so could this be confusing?

docs/content/drafts/Button2.mdx
```diff
@@ -65,6 +65,20 @@ The `invisible` variant of `Button` indicates that the action is a low priority
...
+This is in many cases similar to the active state of the button. Usually used in `ButtonGroup` when like tabs.
```
rezrah, line 71: Should this capability be restricted to `ButtonGroup`?  I can see two risks with enabling this for individual buttons:

1) Opens door to misuse of standalone button appearance
2) A standalone button shouldn't need to be pre-selected

docs/content/drafts/Button2.mdx
pksjce: Same! this was the exact confusion I was trying to discuss on slack as well. prop or class? Class is just not how we'd do this in react

docs/content/drafts/Button2.mdx
pksjce: 🤔 I don't understand what misuse of this pattern could look like.
How would you restrict to `ButtonGroup`? By parent class?

mperrotti: We talked about this in the Primer React planning meeting.

I don't think we want to restrict it to `ButtonGroup` buttons, especially because we use the "selected" styling when a button controls a dropdown and the dropdown is open.

https://user-images.githubusercontent.com/2313998/151037767-e41232dc-e1db-46cd-9724-4814d5575a8f.mp4

docs/content/drafts/Button2.mdx
mperrotti: Should it be`aria-pressed` instead? 🤔 

> The aria-selected attribute indicates the current "selected" state for gridcell, option, row and tab roles.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected

> The aria-pressed attribute indicates the current "pressed" state of a toggle button.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

cc @alliethu - what do you think the correct ARIA attribute might be?

mperrotti: This came up today in the Primer patterns working session. We decided this should be purely visual and not set any ARIA attributes.

In the case where a button controls a dropdown, it would need to be passed `aria-haspopup` and `aria-expanded` attributes. These attributes would be passed by the component that renders the dropdown button and menu.

cc @langermank

docs/content/drafts/Button2.mdx
langermank: I _think_ what we decided in the meeting is to utilize the `active` state for dropdowns (this is all forward thinking, not talking about existing functionality) and restrict `aria-pressed` to a future **new** Segmented Control component (a more specific version of a ButtonGroup). The confusing piece of this is that we are simultaneously updating an existing Button component while refactoring a future Button component.

@pksjce I may have missed this, but do you have the use case that prompted this ask to add `selected`? We may need to continue supporting it in the original Button while we redesign the future Button.. but my recommendation there will depend on the use case for adding it now.

rezrah: @langermank - this was the original issue where this came up https://github.com/primer/react/issues/1783

langermank: Thanks @rezrah.. that's an interesting use case 🤔 I still think this would be a `SegmentedControl` though I imagined that to be two/three options max not six 😆 I guess in that case to unblock them use the `aria-selected` styles from Primer CSS but map it to `aria-pressed`? But I would only add this to the existing Button component, not the refactor.

docs/content/drafts/Button2.mdx
pksjce: Two questions here
- How do we set the pressed styles in this case?
- How do we support the toggles in the tabs use case in the issue https://github.com/primer/react/issues/1783 ?

I'm thinking we need to implement `aria-pressed` styles and scrap the `selected` prop.

docs/content/drafts/Button2.mdx
pksjce: We are planning to deprecate the existing component as soon as possible. I think for parity we'd need to have it in the new one as well? The new component is not a refactor in terms of functionality I feel, only the API. Once you are done with https://github.com/primer/css/pull/1874, we'd have to work on absorbing it into primer-react.

docs/content/drafts/Button2.mdx
siddharthkp: The two use cases here:

1. Sounds like #1783 is using Button as a "toggle button", which isn't just a visual update but also a semantic change? 

From MDN (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed): 

> Adding aria-pressed to an element with a role of button turns the button into a toggle button. The aria-pressed attribute is only relevant for toggle buttons. It represents the button's current "pressed" state.

It looks like it makes sense to add `aria-pressed` for toggle button. cc @alliethu for confirmation

2. For buttons that open a Menu, we add the aria attributes `aria-haspopup="true" aria-expanded="true"`, _we can_ modify button styles if those are present OR to keep the two concerns independent, we can pass a `pressed` prop or className to the Button from the Menu components.

docs/content/drafts/Button2.mdx
langermank: @pksjce right, thanks for pointing that out. How do you see the new Button API changes rolling into PRC? In CSS and PVC, I think we'll likely end up creating a new component similar to how you've handled `Button2`. If you follow that pattern in PRC, you would end up with `Button3` 😆 Although, the markup in `Button2` is almost identical to where we're going with the Button API changes, so perhaps it wouldn't be a big deal to wrap it into `Button2` once `Button` is deprecated (this is getting a little confusing!)

pksjce: Yes, I'm not planning to have a `Button3` for sure. I'd just update `Button2` which would become the only `Button` and release it going forward. I don't think there'd be any major API changes. If any breaking then we'd go for a major version change I believe.

docs/content/drafts/Button2.mdx
pksjce: Got it, so @mperrotti @siddharthkp I'm adding a `pressed` prop to the component and we can handle the css based on that. This will address both the above use cases.

docs/content/drafts/Button2.mdx
```diff
@@ -65,6 +65,20 @@ The `invisible` variant of `Button` indicates that the action is a low priority
...
+
+```jsx live
+<>
+  <drafts.Button aria-selected="true">Search</drafts.Button>
+  <drafts.Button aria-selected="true" variant="danger">
+    Delete
+  </drafts.Button>
```
siddharthkp, line 78: Would be nice to have examples for both ways of doing it :)

```suggestion
  <drafts.Button aria-selected="true">Search</drafts.Button>
  <drafts.Button className="selected" variant="danger">
    Delete
  </drafts.Button>
```

src/Button/Button.tsx
```diff
@@ -31,6 +31,11 @@ const Button = styled(ButtonBase)<ButtonBaseProps & SxProp>`
     background-color: ${get('colors.btn.bg')};
     border-color: ${get('colors.btn.border')};
   }
+  &.selected,
+  &[aria-selected='true'] {
```
siddharthkp, line 35: 👍 good call on the `=true`! I've done the mistake of just putting `aria-selected` in the selector and catching `=false coz react 😅

siddharthkp (approved):

Issue #1802, opened 2022-01-17
Overlay: Events bubbling up to the page can conflict with global shortcuts


siddharthkp:
via @dusave on [slack](https://app.slack.com/client/T0CA8C346/C01L618AEP9): Is there a way to stop propagation of Escape when an anchored overlay is open? I just want the anchored overlay to close, but it's also bubbling up outside of the anchored overlay

Outdated repo (This feature was removed in memex, [watch this video instead](https://www.loom.com/share/962646dc566a41329c5d660c932ba985))
 
https://github.com/orgs/github/projects/4205/settings/fields/55540

1. Click on the dates of a given iteration, make the date picker show
2. Hit Escape
3. Notice it closes the date picker and settings. We only want it to close the date picker

---

### Problem: 

By default, pressing Escape in an Overlay bubbles up to the page outside the Overlay. Adding `event.stopPropagation` inside `onEscape` does not stop the events from bubbling up because of the way the handlers are wired.

```jsx
const Page = () => {
  React.useEffect(() => {
    // global keydown listener on page.
    document.addEventListener('keydown', console.log('global handler:', event.key))
  }, [])

  return (
    <Overlay
      onEscape={event => {
        closeOverlay()
        // You'd expect this to prevent the event from bubbling 
        // to the global handler on the page, but it doesn't
        event.stopPropogation() 
      }}
    >
      ...
    </Overlay>
  )
```

### Scope 

This issue of course isn't just limited to "Escape". If an application has global shortcuts, events bubbling out of Overlay can fire global shortcuts. 

### Why does this happen?

When escape is pressed, you would expect it to be before because Overlay is a child of the Page, but this isn't the case. Overlay uses the `useOnEscapePress` hook which [attaches an event handler on the document](https://github.com/primer/react/blob/main/src/hooks/useOnEscapePress.ts#L70), this is great because no matter where the focus is, pressing Escape will close the top most Overlay.

However, because this event listener is on the document, we cannot predict the order in which it will be fired - will it be before the global handler on the page or after it? And that's why `stopPropagation` and `preventDefault` will not stop the global handler on the Page from catching this event. [Here is a loom with a demo of this](https://www.loom.com/share/962646dc566a41329c5d660c932ba985)

[...]

PR #1824, opened 2022-01-26
Overlay: Attach escape handler to overlay container

siddharthkp:
Fixes https://github.com/primer/react/issues/1802

### Screenshots

Explainer: https://www.loom.com/share/962646dc566a41329c5d660c932ba985

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/hooks/useOnEscapePress.ts
```diff
@@ -38,16 +38,24 @@ function handleEscape(event: KeyboardEvent) {
  * @param callbackDependencies {React.DependencyList} The dependencies of the given
  * `onEscape` callback for memoization. Omit this param if the callback is already
  * memoized. See `React.useCallback` for more info on memoization.
+ *
+ * @param containerRef {React.RefObject<HTMLElement>} The overlay element to attach the
```
rezrah, line 42: ❤️  great idea

src/stories/Overlay.stories.tsx
```diff
@@ -211,3 +229,103 @@ export const OverlayOnTopOfOverlay = ({anchorSide}: OverlayProps) => {
     </Box>
   )
 }
+
+export const OverlayOnOverlay = () => {
```
rezrah, line 233: Suggestion on naming: NestedOverlays?

rezrah (approved):
🙌  amazing @siddharthkp - thanks for fixing this.

rezrah:
Out of curiosity, did you test the canary on memex and verify this works?

siddharthkp:
> Out of curiosity, did you test the canary on memex and verify this works?

Yes, but the solution isn't ideal. You would have to add `event.stopPropagation` to all the overlay components on the settings page. I feel like that should probably be the default when you are inside an Overlay and we can handle that on our end, making it easier for users 🤔

PR #1826, opened 2022-01-29
Remove unnecessary label title

PeterYangIO:
Remove unnecessary title attribute on `<InputLabel/>`. Currently there is a title that says "required field" when the input is marked as required which can end up overriding the accessible name for the input. The first screenshot shows how the a11y labels are parsed when trying to use `getByRole('textbox', {name: TEXTINPUTFIELD_LABEL_TEXT})`. As you can see, the title overrides the label text which will be particularly confusing on a page with multiple required inputs as the a11y tree will show multiple text inputs all with the "required field" name.

In this PR I simply remove the title attribute because the necessary "required" information is already passed to the `<input/>` itself with the `required` attribute.

### Screenshots
image

No visual changes (unless you count the lack of a title when hovering as one).

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [ ] Tested in Safari // I'm on windows
- [X] Tested in Edge


PeterYangIO:
Bonus side-effect showing how it fixes the issue with Testing Playing (for testing-library):

For required fields with fix applied:
image

Trying to do the same without the fix applied:
image

PeterYangIO:
@pksjce Do you think you could take a look or assign the right person? Thanks!

rezrah:
LGTM. Thanks for raising this PR @PeterYangIO.

@alliethu @mperrotti  - Do you see any issues if we stop the `title` attribute automatically being applied to required fields? It also seems to prevent users from applying custom `title` values in the `InputField` component.

Screenshot 2022-02-09 at 12 48 41

rezrah (approved):

PeterYangIO:
Hi @rezrah I do not have merge permissions for the repo, if you think it looks good do you think you could merge it?

Thanks!

PR #1831, opened 2022-02-01
Input validation styles

mperrotti:
These changes render validation styles on form inputs that didn't previously have them, and shows those validation styles in Storybook.

Closes https://github.com/github/primer/issues/495

### Merge checklist

- [n/a] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


docs/content/TextInputWithTokens.mdx
```diff
@@ -213,12 +213,19 @@ render(MaxHeightExample)
     description="Whether tokens should render inline horizontally. By default, tokens wrap to new lines"
   />
   <PropsTable.Row name="size" type="TokenSizeKeys" defaultValue="extralarge" description="The size of the tokens" />
+  <PropsTable.Row
+    name="inputSize"
+    type="'small' | 'large'"
+    defaultValue="small"
```
rezrah, line 219:
```suggestion
    defaultValue="'small'"
```

docs/content/TextInputWithTokens.mdx
```diff
@@ -213,12 +213,19 @@ render(MaxHeightExample)
...
+    description="The size of the text input"
+  />
   <PropsTable.Row
     name="hideTokenRemoveButtons"
     type="boolean"
     defaultValue="false"
     description="Whether the remove buttons should be rendered in the tokens"
   />
+  <PropsTableRow name="validationStatus" type="'warning' | 'error'" description="Style the input to match the status" />
```
rezrah, line 228: If we're not removing 'warning' as part of this issue, do we need to create another one or does it already exist?

rezrah: Think you're missing `'success'` from here too?

```suggestion
  <PropsTableRow name="validationStatus" type="'warning' | 'error' | 'success'" description="Style the input to match the status" />
```

src/stories/TextInputWithTokens.stories.tsx
```diff
@@ -30,7 +32,53 @@ export default {
...
+    },
+    size: {
+      name: 'size (token size)',
+      defaultValue: 'extralarge',
+      options: ['small', 'medium', 'large', 'extralarge'],
+      control: {
+        type: 'radio'
+      }
+    },
+    inputSize: {
```
rezrah, line 57: Heads up.. this control doesn't seem to work as intended?

https://user-images.githubusercontent.com/13340707/152334520-8970f2da-9b7e-42b7-a59b-ff05bf8eba98.mov

src/TextInputWithTokens.tsx
```diff
@@ -45,6 +45,10 @@ type TextInputWithTokensInternalProps<TokenComponentType extends AnyReactCompone
    * The size of the tokens
    */
   size?: TokenSizeKeys
+  /**
+   * The size of the text input
+   */
+  inputSize?: StyledWrapperProps['size']
```
rezrah, line 51: Curious about this one, could you provide some context on it please? I'm a bit confused about relationship to native `size` attribute on `<input`, which is a number? When I tried `small` or `large` it doesn't seem to make any noticeable difference either.

rezrah: Also think that given the component it named as a `**TextInput**WithTokens`, `size` prop should perhaps relate to input instead of tokens

docs/content/TextInputWithTokens.mdx
rezrah: Another one docs related one (sorry): Can we add an example showing `validationStatus` being applied?

src/TextInputWithTokens.tsx
```diff
@@ -246,7 +252,9 @@ function TextInputWithTokensInnerComponent<TokenComponentType extends AnyReactCo
       width={widthProp}
       minWidth={minWidthProp}
       maxWidth={maxWidthProp}
-      variant={variantProp}
+      size={inputSize}
+      validationStatus={validationStatus}
```
rezrah, line 256: can we add an `aria-invalid` attribute here too if `validationStatus === 'error'`?

src/stories/TextInputWithTokens.stories.tsx
```diff
@@ -30,7 +32,53 @@ export default {
...
+      }
+    },
+    visibleTokenCount: {
+      defaultValue: undefined,
+      control: {
+        type: 'number'
+      }
+    },
+    validationStatus: {
+      options: ['warning', 'error', undefined],
```
rezrah, line 77: missing 'success' here?

docs/content/TextInputWithTokens.mdx
mperrotti: > If we're not removing 'warning' as part of this issue, do we need to create another one or does it already exist?

We need to create one

mperrotti: > Another one docs related one (sorry): Can we add an example showing validationStatus being applied?

Yes

src/stories/TextInputWithTokens.stories.tsx
mperrotti: Not sure what happened there. Will fix.

src/TextInputWithTokens.tsx
mperrotti: I agree, `size` should refer to the input. When this component was made, we used `variant` for size.

I'll change it and update `size` to be `tokenSize`.

src/TextInputWithTokens.tsx
mperrotti: Yes, we should.

src/stories/TextInputWithTokens.stories.tsx
mperrotti: We don't use 'success' for `TextInput`, so I didn't add it for `TextInputWithTokens`.

src/TextInputWithTokens.tsx
mperrotti: Or should this be a separate issue? It would technically be a breaking change

src/stories/TextInputWithTokens.stories.tsx
mperrotti: The size difference isn't noticeable with most token sizes unless you have text in the input.

Maybe we should get rid of the `inputSize` prop altogether, and rely just on the `size` prop. Then we can manually pass the size to the `TextInputWrapper` depending on what was passed to `size`.

I'll push something up and you can let me know what you think.

src/TextInputWithTokens.tsx
mperrotti: This won't be an issue if we just remove `inputSize` as I'm recommending in https://github.com/primer/react/pull/1831#discussion_r802156523

src/stories/TextInputWithTokens.stories.tsx
mperrotti: Before:
https://user-images.githubusercontent.com/2313998/153096054-74a5d87e-646e-4a5c-84b0-539d87a428eb.mp4

After:
https://user-images.githubusercontent.com/2313998/153096410-3d8c7d3d-4f3a-4c2c-83b6-35ad8cc5dc38.mp4

src/stories/TextInputWithTokens.stories.tsx
rezrah: Does it use `TextInputWrapper` under the hood? If so, it does support `success` now in the base component: https://github.com/primer/react/blob/ce7bef34b9a4024ee63ee79da880f9b9aac7f74e/src/_TextInputWrapper.tsx#L46

docs/content/TextInputWithTokens.mdx
```diff
@@ -213,12 +244,19 @@ render(MaxHeightExample)
     description="Whether tokens should render inline horizontally. By default, tokens wrap to new lines"
   />
   <PropsTable.Row name="size" type="TokenSizeKeys" defaultValue="extralarge" description="The size of the tokens" />
+  <PropsTable.Row
+    name="inputSize"
+    type="'small' | 'large'"
+    defaultValue="'small'"
+    description="The size of the text input"
+  />
```
rezrah, line 252: This can be removed now with latest changes?

rezrah: Can we also update `TokenSizeKeys` to the actual string values like you did for `inputSize`? 
Screenshot 2022-02-09 at 13 34 04

src/stories/TextInputWithTokens.stories.tsx
rezrah: Yep, I think this is a big improvement 👍 . Thanks for making this change.

On a side note, are there usage guidelines for this particular prop? Curious why / where you'd need some of these larger sizes. Could we be encouraging an anti-pattern?

mperrotti: I think we would mostly use the larger sizes because they have a larger touch target. You would use different sizes depending on where you're rendering the input.

src/stories/TextInputWithTokens.stories.tsx
mperrotti: Ok, then let's accept "success" in both `TextInput` and `TextInputWithTokens`.

We're going to be discussing the "warning" and "success" statuses in next week's Primer patterns working session, so we can update if we need to.

docs/content/TextInputWithTokens.mdx
mperrotti: > Can we also update TokenSizeKeys to the actual string values like you did for inputSize?

I agree that makes more sense, but in other docs we use the type name. For example: https://primer.style/react/ActionList

mperrotti: I talked with @colebemis about this, and we're ok with a little inconsistency until we fix this in the other docs. I've updated https://github.com/primer/react/issues/1701 with this task.

rezrah: Hmm, okay.. it's just expensive to defer something that can be easily resolved now to avoid a heavier context switching cost later. 

It also seems that we have existing precedents, [where we can link directly to the interface/type definitions](https://primer.style/react/Flash) like this:
Screenshot 2022-02-10 at 13 33 57

Can we do the same thing here?

rezrah (approved):

docs/content/TextInputWithTokens.mdx
mperrotti: Yup, I changed it to be the whole string union.

Issue #1690, opened 2021-12-03
[Batch] Improve slots implementation in PRC to enable React SSR compatibility & performance gain


siddharthkp:
## Background

The current slot implementation isn't compatible with server-side rendering (SSR). It only renders the freeform main slot on the server and then adds other slots after mounting on the client (with useEffect) which is not a problem in client-only components like Overlay but can be a bad loading experience for other components with layout shift.

The tradeoffs considered at the time of building this component are mentioned in this comment: https://github.com/github/primer/issues/1224#issuecomment-1222450075

## Task 

Update the slots implementation to render consistently across the server and client. This will introduce a limitation that a slot needs to be a direct child of the slot container, deep nesting is not allowed. 

### Components to migrate

- [x] PageLayout: https://github.com/primer/react/pull/3036
- [x] TreeView: https://github.com/primer/react/pull/3097
- [x] CheckboxGroup: https://github.com/primer/react/pull/3146
- [x] RadioGroup: https://github.com/primer/react/pull/3146
- [x] FormControl https://github.com/primer/react/pull/3149
- [x] ActionList https://github.com/primer/react/pull/3173
- [x] NavList https://github.com/primer/react/pull/3173
- [x] MarkdownEditor (draft) https://github.com/primer/react/pull/3174 https://github.com/github/primer/issues/2090

### Related work

- [x] https://github.com/github/primer/issues/1936

### Clean up
- [x] Remove legacy slots implementation https://github.com/primer/react/pull/3205

siddharthkp:
Another use case for this - https://github.com/orgs/github/issues/issues (behind feature flag)

https://user-images.githubusercontent.com/1863771/180964965-df092a4e-06c3-47bd-abf7-58fe41f6adac.mp4

As a stop gap, we moved the contents of leading visual inside text

siddharthkp:
Another example where this falls short: https://github.com/github/primer/issues/1224

lesliecdubs:
Hey @siddharthkp, would it be helpful to keep this one open for reference or should we go ahead and close? I have a draft in the Primer Roadmap backlog to revisit slots so it's definitely still on the radar. https://github.com/orgs/github/projects/2759/views/32?filterQuery=-no%3Afocus-area+-status%3A%22Inbox+%F0%9F%93%A5%22+-label%3Ainitiative+slots

siddharthkp:
@lesliecdubs Late reply, but I think we should keep this open :)

[...]

PR #1836, opened 2022-02-03
Add FormControl component

mperrotti:
The `FormControl` component combines the functionality of `InputField` and `ChoiceInputField`, and will replace `FormGroup`, `InputField`, and `ChoiceInputField`.

Relates to https://github.com/github/primer/issues/679

### Screenshots
Screen Shot 2022-02-03 at 2 43 12 PM

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


docs/content/FormControl.mdx
```diff
@@ -0,0 +1,429 @@
...
+                {text: 'design-systems', id: 6}
+              ]}
+              selectedItemIds={[]}
+            />
+          </Autocomplete.Overlay>
+        </Autocomplete>
+      </FormControl>
+      <FormControl>
+        <FormControl.Label>Select</FormControl.Label>
+        <Select>
```
rezrah, line 73: Think there may be an issue with `Select` here, as the dropdown isn't triggered unless you press the text. Even pressing arrows doesn't work.

https://user-images.githubusercontent.com/13340707/152513393-00d6afd2-5173-48bd-9cce-c42d9f7d5803.mov

rezrah:
@mperrotti - not sure if it's just me but on the preview url I can't see any labels. Any ideas why?

https://primer-react-efgapobji-primer.vercel.app/react/FormControl
no labels in form control

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,429 @@
...
+    <FormControl
+      validationMap={{
+        noSpaces: 'error',
+        validName: 'success'
+      }}
+      validationResult={validationResult}
+    >
+      <FormControl.Label>GitHub handle</FormControl.Label>
+      <TextInput block value={value} onChange={handleInputChange} />
+      <FormControl.Validation validationKey="noSpaces">GitHub handles cannot contain spaces</FormControl.Validation>
```
rezrah, line 264: Would this be a good opportunity to revisit the `validationKey` prop?

Wondering if this pattern - raised at the workshop - is more familiar and intuitive for end users... shall we switch? what do you think @mperrotti?

```suggestion
      {noSpaces && <FormControl.Validation type="success">GitHub handles cannot contain spaces</FormControl.Validation>}
```

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,429 @@
...
+    <Checkbox />
+  </FormControl>
+</Box>
+```
+
+### With a visually hidden label
+
+<Note>
+
+Every input must have a corresponding label to be accessible to assistive technology. That's why you'd use `FormControl` instead of using [`TextInput`](/TextInput) directly.
```
rezrah, line 179: Just FYI, I don't believe we should use contractions like `you'd` in content. https://brand.github.com/content/grammar#contractions

maybe..

```suggestion
We encourage using `FormControl` alongside all standalone form components like [`TextInput`](/TextInput), as every input must have a corresponding label to be accessible to assistive technology. 
``

File: docs/content/FormControl.mdx (hunk 2 above)
Comment (mperrotti) [reply docs/content/FormControl.mdx:264]: This was accidentally copied over from the InputField example. I switched over to the pattern you described.

File: src/FormControl/FormControl.tsx [hunk 4]
```diff
@@ -0,0 +1,232 @@
...
+   */
+  id?: string
+  /**
+   * Whether this field must have a value for the user to complete their task
+   */
+  required?: boolean
+  /**
+   * Changes the layout of the form control
+   */
+  variant?: 'stack' | 'choice' // TODO: come up with a better name for 'stack'
```
mperrotti, line 30: Should we just go with "stack" for the vertically stacked label+input+caption/validation? Or does anybody have an idea for a better name?

src/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,232 @@
...
+            )}
+            {slots.Caption}
+          </Box>
+        )
+      }}
+    </Slots>
+  )
+}
+
+FormControl.defaultProps = {
```
rezrah, line 222: Do you need this here, if default value is already declared as part of param destructuring?

src/FormControl/FormControl.tsx
rezrah: Could it be more abstract like `vertical` and `horizontal` with prop called `alignment` instead?

Taking that further, do we need a variant prop at all?

E.g. 

```jsx
      <FormControl variant="stack">
        <FormControl.Label>Checkbox option one</FormControl.Label>
        <Checkbox />
      </FormControl>

```

☝️ This is incorrect and breaks, but is the default value that user needs to manually change to `choice`. 

Can we read children via `React.Children` to see if there's a type of `Checkbox` or `Radio`, which automatically applies `choice` for them without needing it via prop?

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+  return (
+    <Box
+      display="flex"
+      flexDirection="column"
+      sx={{
+        '> * + *': {
+          marginTop: 4
+        }
+      }}
+    >
```
colebemis, line 46: I think this will give you a similar result with less code:
```suggestion
    <Box display="grid" gridGap={1}>
```

`<Box sx={{display: 'grid', gap: 1}}>` should also work, depending on your preference.

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+### Basic
+
+```jsx live
+<FormControl>
+  <FormControl.Label>Name</FormControl.Label>
+  <TextInput />
+</FormControl>
+```
+
+### `variant="stack"` (default) inputs
```
colebemis, line 24: ```variant="stack" (default) inputs``` didn't make sense to me at first because it wasn't clear to me what `variant="stack"` meant. How about this?

```suggestion
### With complex inputs
```

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+      </FormControl>
+    </Box>
+  )
+}
+
+render(DifferentInputs)
+```
+
+### `variant="choice"` inputs
+
```
colebemis, line 92: `variant="choice"` seems like an ambiguous heading. What if we added a description to explain why someone would use `variant="choice"`? 

```suggestion
### With checkbox and radio inputs

Use `variant="choice"` to position checkbox and radio inputs to the left of the label.
```

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+```javascript live noinline
```
colebemis, line 93: Is there a reason we need to use `noinline` in this example?

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+  Checkbox and radio form controls cannot be required individually. Instead, you can require a selection from the entire
+  group of checkboxes or radios.
+</Note>
+
+```jsx live
+<FormControl required>
+  <FormControl.Label>Name</FormControl.Label>
+  <TextInput />
+</FormControl>
+```
```
colebemis, line 148: This note feels like it could come after the example:
````suggestion
```jsx live
<FormControl required>
  <FormControl.Label>Name</FormControl.Label>
  <TextInput />
</FormControl>
```

<Note variant="warning">
  Checkbox and radio form controls cannot be required individually. Instead, require a selection from the entire
  group of checkboxes or radios.
</Note>
````

File: docs/content/FormControl.mdx [hunk 11]
```diff
@@ -0,0 +1,414 @@
...
+    </FormControl>
+  )
+}
+
+render(ValidationExample)
+```
+
+### With a leading visual
+
+<Note>Only a checkbox or radio form control may have a leading visual.</Note>
```
colebemis, line 272: Actually, I think most of these notes could go after the code example and should maybe be marked as warnings?

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+    name="disabled"
+    type="boolean"
+    defaultValue="false"
+    description="Whether the field is ready for user input"
+  />
+  <PropsTableRow
+    name="id"
+    type="string"
+    defaultValue="a generated string"
+    description="The unique identifier for this field. Used to associate the label, validation text, and caption text"
```
colebemis, line 323: Should we be using `field` or `control` in these descriptions?
```suggestion
    description="The unique identifier for this control. Used to associate the label, validation text, and caption text"
```

src/FormControl/FormControl.tsx
colebemis: @rezrah We discussed that after you left one of our meetings. We decided that reading from `React.Children` felt like a little too much magic and would prevent people from using custom checkboxes/radios

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+<PropsTable>
+  <PropsTableRow
+    name="children"
+    type="FormControl.Label | FormControl.Caption | FormControl.Validation | Autocomplete | TextInput | TextInputWithTokens | Select"
+    required
+  />
+  <PropsTableRow
+    name="variant"
+    defaultValue="'stack'"
+    type="'stack' | 'choice'"
```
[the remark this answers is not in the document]
colebemis: Oh, looks like you already mentioned this [here](https://github.com/primer/react/pull/1836/files#r799547693)

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,414 @@
...
+    defaultValue="a generated string"
+    description="The unique identifier for this field. Used to associate the label, validation text, and caption text"
+  />
+  <PropsTableRow
+    name="required"
+    type="boolean"
+    defaultValue="false"
+    description="Whether the field is ready for user input"
+  />
+</PropsTable>
```
colebemis, line 331: Do you intentionally omit the `sx` prop from the `FormControl.*` components?

src/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,232 @@
...
+  variant: 'stack'
+}
+
+export type FormControlComponentProps = ComponentProps<typeof FormControl>
```
colebemis, line 226: I don't think `ComponentProps<>` adds much here since this is not a styled-component. Let's just export the props type where it's defined.

```suggestion
```

src/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,232 @@
...
+import {ComponentProps} from '../utils/types'
+import FormControlCaption from './_FormControlCaption'
+import FormControlLabel from './_FormControlLabel'
+import FormControlValidation from './_FormControlValidation'
+import {Slots} from './slots'
+import ValidationAnimationContainer from '../_ValidationAnimationContainer'
+import {get} from '../constants'
+import FormControlLeadingVisual from './_FormControlLeadingVisual'
+
+export interface Props {
```
colebemis, line 13: We typically use `type` instead of `interface` for props.

```suggestion
export type FormControlProps = {
```

src/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,232 @@
...
+  variant?: 'stack' | 'choice' // TODO: come up with a better name for 'stack'
+}
+
+type FormControlValidationProps = ComponentProps<typeof FormControlValidation>
+export interface FormControlContext extends Pick<Props, 'disabled' | 'id' | 'required'> {
+  captionId: string
+  validationMessageId: string
+}
+
+const FormControl = ({children, disabled, id: idProp, required, variant = 'stack'}: Props) => {
```
colebemis, line 39:
```suggestion
const FormControl = ({children, disabled, id: idProp, required, variant = 'stack'}: FormControlProps) => {
```

src/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,232 @@
...
+                  disabled,
+                  ['aria-describedby']: captionId
+                })}
+              {React.Children.toArray(children).filter(
+                child =>
+                  React.isValidElement(child) &&
+                  ![Checkbox, Radio].some(inputComponent => child.type === inputComponent)
+              )}
+            </Box>
+            {slots.LeadingVisual && (
```
colebemis, line 159: Not a blocker for this PR:

@mperrotti and @siddharthkp how do y'all feel about not capitalizing slot names (`slots.leadingVisual` instead of `slots.LeadingVisual`)? The PascalCase makes me think I can do something like this `<slots.LeadingVisual>` (which wouldn't work)

src/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,232 @@
...
+  )?.filter(Boolean)
+  const labelChild: React.ReactNode | undefined | null = React.Children.toArray(children).find(
+    child => React.isValidElement(child) && child.type === FormControlLabel
+  )
+  const validationMessageId = validationChild ? `${id}-validationMsg` : ''
+  const validationStatus = React.isValidElement(validationChild) ? validationChild.props.appearance : undefined
+  const captionId = captionChildren?.length ? `${id}-caption` : undefined
+  const InputComponent = React.Children.toArray(children).find(child =>
+    expectedInputComponents.some(inputComponent => React.isValidElement(child) && child.type === inputComponent)
+  )
```
colebemis, line 56: Does this implementation prevent people from passing custom input components?

colebemis (commented):
This is excellent! ✨ 

I left a lot of comments but they're mostly nitpicks or questions. I didn't notice any serious blockers

src/FormControl/FormControl.tsx
```diff
@@ -0,0 +1,232 @@
...
+      console.warn('The Checkbox or Radio components are only intended to be used with the "choice" variant')
+    }
+
+    if (
+      React.Children.toArray(children).find(
+        child => React.isValidElement(child) && child.type === FormControlLeadingVisual
+      )
+    ) {
+      // eslint-disable-next-line no-console
+      console.warn(
```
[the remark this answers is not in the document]
mperrotti: Jest specifically tests to make sure these are logged, so I don't think we can.

src/FormControl/FormControl.tsx
siddharthkp: that's a good point 🤔 

I call it `slots.LeadingVisual` in slots because I wanted to call it the exact same name from the component API, which is `ActionList.LeadingVisual`. 

```js
const {Slots, Slot} = createSlots(['LeadingVisual', 'InlineDescription', 'BlockDescription', 'TrailingVisual'])
```
https://github.com/primer/react/blob/main/src/ActionList2/Item.tsx#L79

mperrotti: I'm cool with that change, but I'd rather make it across _all_ components instead of just this one.

src/FormControl/FormControl.tsx
mperrotti: Yes. We can open it up later if we need to.

docs/content/FormControl.mdx
```diff
@@ -0,0 +1,372 @@
...
+    {text: 'zero', id: 0},
+    {text: 'one', id: 1},
+    {text: 'two', id: 2}
+  ])
+  const onTokenRemove = tokenId => {
+    setTokens(tokens.filter(token => token.id !== tokenId))
+  }
+
+  return (
+    <Box display="grid" gridGap={1}>
```
colebemis, line 38: Nitpick: Can we use a gap of `3` instead of `1` in these examples? `gridGap={1}` feels a little cramped IMO


CleanShot 2022-02-04 at 18 35 40

```suggestion
    <Box display="grid" gridGap={3}>
```

src/FormControl/FormControl.tsx
rezrah: Thanks @colebemis. Should the values be more explicit to the type of child instead then? It's not immediately clear (to me at least) which values (stacked/choice) to use for `variant`, even though there are restrictions on types of input to variant. E.g. You can't use stacked with checkbox children.

E.g.
```jsx
      <FormControl variant="checkbox">
        <FormControl.Label>Checkbox option one</FormControl.Label>
        <Checkbox />
      </FormControl>
```

I personally find `for` even more readable...

```jsx
      <FormControl for="checkbox">
        <FormControl.Label>Checkbox option one</FormControl.Label>
        <Checkbox />
      </FormControl>
```

docs/src/@primer/gatsby-theme-doctocat/nav.yml
```diff
@@ -53,12 +53,10 @@
       url: /Buttons
     - title: Checkbox
       url: /Checkbox
-    # Temporarily hiding form field layout components due to
-    # feedback about the names being unclear
+    # Temporarily hiding ChoiceFieldset because the API
+    # is being refactored with breaking changes
```
rezrah, line 57: Can we delete these from here now?

src/FormControl/FormControl.tsx
siddharthkp: I'd like to keep `slots.LeadingVisual` until we know what changes in https://github.com/primer/react/issues/1690 in the hopes we can actually have `<Slots.LeadingVisual props="too">` 😅

src/FormControl/FormControl.tsx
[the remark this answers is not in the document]
rezrah: Rather than testing the global, can you do something like: 

```js

const mockWarningFn = jest.fn();

beforeAll(() => {
  jest.spyOn(global.console, 'warn').mockImplementation(mockFn);
});

afterAll(() => {
  jest.clearAllMocks();
});


...

// where you're checking it...
expect(mockWarningFn).toHaveBeenCalled();
```

src/FormControl/FormControl.tsx
mperrotti: @colebemis - I went with Reza's suggestion to remove `variant`. I won't merge this until you either say this is ok, or we come up with a solution everybody is comfortable with.

src/__tests__/FormControl.test.tsx
```diff
@@ -0,0 +1,434 @@
...
+const LABEL_TEXT = 'Form control'
+const CAPTION_TEXT = 'Hint text'
+const ERROR_TEXT = 'This field is invalid'
+
+describe('FormControl', () => {
+  const mockWarningFn = jest.fn()
+  const mockErrorFn = jest.fn()
+
+  beforeAll(() => {
+    jest.spyOn(global.console, 'warn').mockImplementation(mockWarningFn)
```
rezrah, line 18: ❤️

docs/content/FormControl.mdx
mperrotti: No, that was just a silly copy/paste mistake

docs/content/FormControl.mdx
mperrotti: I had them marked as warnings at first, but it felt aggressive

docs/content/FormControl.mdx
mperrotti: Yes

src/FormControl/FormControl.tsx
siddharthkp: > Given this component only supports internal form inputs, I think it makes sense to remove variant.

When we were looking at the API, we talked about supporting custom form inputs and that's where an explicit layout `variant` came from, because there is no non-hacky way for the user to get a vertical layout on their own.

If we are choosing to not support custom input elements for now, we can remove the variant prop. I'm always up for a little magic. I imagine the need for custom inputs will come up at some point and then we can bring back the `variant` prop, it's nice that we already have a solution.

rezrah (approved):

PR #1840, opened 2022-02-04
Add package engines

jclem:
This adds the following to package.json and docs/package.json:

```json
"engines": {
  "node": ">=12",
  "npm": ">=7"
}
```

Node 12 works with this project and since it's [the oldest maintenance LTS version of Node.js](https://nodejs.org/en/about/releases/), it seems like a reasonable choice.

The docs/package-lock.json has also been updated to v2, now that we're using npm>=7, which uses the newer lock file format.


colebemis (approved):
Thanks, @jclem!

PR #1841, opened 2022-02-04
Add history dependency

jclem:
Since the "history" package is used by [UnderlineNav](https://github.com/primer/react/blob/1c4786c7d99fe083b47902acff5326c44a7c8288/src/UnderlineNav.tsx#L2), it should be in @primer/react's direct production dependencies. This ensures that its types are properly exposed to consumers of @primer/react that do not have `skipLibCheck` enabled.

Since this doesn't actually change anything in @primer/react at runtime or even in development, I don't believe this is worth a changelog.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "color2k": "1.2.4",
         "deepmerge": "4.2.2",
         "focus-visible": "5.2.0",
+        "history": "5.0.0",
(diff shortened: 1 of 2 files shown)


colebemis (approved):
Thanks, @jclem!

We should probably eventually update UnderlineNav to remove its dependency on `history` but this seems like a reasonable fix for now 👍

PR #1842, opened 2022-02-04
Make KeyPaths not require @ts-ignore

jclem:
Currently, `KeyPaths` requires a `@ts-ignore` line, because TypeScript can't verify that `O[K]` is a `Record<string, unknown>`.

In order to fix this, we can change `KeyPaths<O extends Record<string, unknown>>` to just `KeyPaths<O>`, but internally still verify that when recursing through `O`, we only recurse when `O[K] extends Record<string, unknown>`.

This makes it so that `KeyPaths` can be called on _any_ value and I'm not sure if the repercussions of that are relevant. However, we should avoid `@ts-ignore` in types that are potentially consumed by projects using @primer/react, otherwise we essentially require that they have `skipLibCheck: true` in their TS config files.

There may be some better way of solving this, but my TypeScript knowledge isn't strong enough to figure it out 😄

### DIFF
--- a/src/__tests__/KeyPaths.types.test.ts
+++ b/src/__tests__/KeyPaths.types.test.ts
@@ -1,5 +1,5 @@
 import {Union} from 'ts-toolbelt'
-import {KeyPaths} from '../utils/types/KeyPaths'
+import {KeyPaths, KP} from '../utils/types/KeyPaths'
 
 type NestedObject = {
   a: string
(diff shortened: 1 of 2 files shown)


colebemis (approved):
LGTM ✨

rezrah (approved):
Looks great, thanks @jclem!

PR #1846, opened 2022-02-08
PageLayout: Prevent content region from wrapping on wide viewports

colebemis:
## Problem

If the contents of `PageLayout.Content` are too wide, `PageLayout.Content` wraps onto a different line than `PageLayout.Pane`.

## Solution

This PR fixes this issue with a combination of `flex-grow: 1` and `flex-basis: 0%` to ensure that `PageLayout.Content` never wraps on wide viewports.

## Screenshots

| Before | After |
| --- | --- |
| CleanShot 2022-02-08 at 14 57 24 | CleanShot 2022-02-08 at 14 56 22 |

### Browsers

- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari

Part of https://github.com/github/primer/issues/565


colebemis:
🚫  Looks like this broke the other stories. Investigating...

colebemis:
Fixed with `flex-basis` ✅

mperrotti (approved):

src/PageLayout/PageLayout.tsx
```diff
@@ -229,8 +229,21 @@ const contentWidths = {
...
-    <Box sx={merge<BetterSystemStyleObject>({order: REGION_ORDER.content, flexGrow: 1}, sx)}>
-      <Box sx={{maxWidth: contentWidths[width], marginX: 'auto'}}>{children}</Box>
+    <Box
+      sx={merge<BetterSystemStyleObject>(
+        {
+          order: REGION_ORDER.content,
+          // Set flex-basis to 0% to allow flex-grow to control the width of the content region.
+          // Without this, the content region could wrap onto a different line
+          // than the pane region on wide viewports if its contents are too wide.
+          flexBasis: '0%',
```
mperrotti, line 239: Minor nitpick: we could just say `flexBasis: 0`. We don't need the unit

colebemis: good catch!

colebemis, line 239:
```suggestion
          flexBasis: 0,
```

Issue #1849, opened 2022-02-09
Storybook dependency changes types in build output


jclem:
**Describe the bug**

In the types output by Primer React, many components are augmented by the Emotion library, which is a transitive dependency of Storybook. This means that an app consuming Primer React can not be compiled unless `skipLibCheck: true` is enabled or the app contains the same Emotion dependency via some other means.

For example, in the build output, `lib/Button/ButtonClose.d.ts` contains the following type augmentation:

```tsx
declare const ButtonClose: React.ForwardRefExoticComponent<Pick<{
  /* etc */
  css?: import("@emotion/core").InterpolationWithTheme<any>;
  /* etc */
}>>
```

The reason that this augmentation appears is that the `@emotion/core` package, specifically, [globally augments the `DOMAttributes` interface exported by React](https://github.com/emotion-js/emotion/blob/9d7e25ae33133454d240b6616c80dd5bb1529a6d/packages/core/types/index.d.ts#L82-L86
):

```ts
declare module 'react' {
  interface DOMAttributes<T> {
    css?: InterpolationWithTheme<any>
  }
}
```

**To Reproduce**

1. Install and use @primer/react in any project without other dependencies.
2. Run the TypeScript compiler with `skipLibCheck: false`.

**Expected behavior**

The compiler will pass.

**Additional context**

I haven't searched exhaustively, but I believe there are other types imported by surprise. For example, `hoist-non-react-statics` also appears in exported types.

**Actual behavior**

The compiler complains that it can't find types for `@emotion/core`.

---

I think the correct solution for this would be to move Primer React's storybook stories and all of the related dependencies into a separate package within this repository.

PR #1852, opened 2022-02-09
Extract Storybook content into subdirectory

jclem:
🚨 🏗️ **WIP:** Work in progress 🚧 🚨

Closes https://github.com/primer/react/issues/1849 by extracting Storybook files and configuration to the stories/ directory.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


jclem:
All tests are passing except for the `checkStoriesForAxeViolations` tests, which `require` the stories themselves.

Now that the root and stories/ directories have separate package.json manifests and node_modules, it appears that we may have two separate instances of React installed. Temporarily, this can be resolved by `rm -r`-ing the `stories/node_modules` directory, which wipes out the additional React package installation, but I'm not sure what a permanent fix looks like, yet.

colebemis:
👋  I know this isn't done yet, but I just wanted to mention that this approach would move us away from the [file structure ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-XXX-file-structure.md) that @jfuchs proposed last quarter. Not a dealbreaker but something to consider.

jclem:
@colebemis I don't think we have a choice, unfortunately, unless we want to continue having Primer React ship types from Storybook in its build output. This happens _just_ because those types are in our `node_modules` directory when we compile.

The only other solution I might try is to restrict the `typeRoots` in `tsconfig.build.json`, but it'll mean listing each build dependency's `node_modules` types path specifically.

jclem:
Closed in favor of https://github.com/primer/react/pull/1855

PR #1853, opened 2022-02-10
Stories: Add example of onEscape for Overlay

siddharthkp:
Add story to override `onEscape` with Overlay

Story: https://primer-react-gqp6t9wxq-primer.vercel.app/react/storybook?path=/story/internal-components-overlay--override-on-escape


siddharthkp:
Covered in https://github.com/primer/react/pull/1861

Issue #1854, opened 2022-02-10
Element-bound events with `useOnEscapePress` are triggering all `escape` events globally


iansan5653:
A change introduced in `34.4.0` was to add support for `useOnEscapePress` to bind events to specific elements instead of just the document root (and to implement that new functionality in the `useOverlay` hook):

> https://github.com/primer/react/pull/1824 [4eab65e5](https://github.com/primer/react/commit/4eab65e57571d64a0f8a4fcb4aafbf47d3e98a2d) Thanks [@siddharthkp](https://github.com/siddharthkp)! - Overlay: Attach escape handler to overlay container instead of document to fix stopPropagation

This makes sense, because the events need to be bound to the overlay container so that the overlay can `stopPropogation` and keep its events contained.

However, this hook is using a global function with a global array of event handlers, even with the element-bound events. So when an escape event is triggered on a top-level overlay, it still calls every escape event handler ever registered:

https://github.com/primer/react/blob/96473f398bb0fe5f2eff1b3e8bb9110a009f1894/src/hooks/useOnEscapePress.ts#L3-L19

An example of the undesirable behavior that this bug causes is visible in the [Nested Overlays](https://primer-react-68602yp8g-primer.vercel.app/react/storybook?path=/story/internal-components-overlay--nested-overlays) story:

- Click the dropdown arrow next to "Star"
- Click "+ Create List"
- Press <kbd>Escape</kbd>
  - Only the top overlay should close, but both do instead
- Click the dropdown arrow next to "Star" again
  - Both overlays open in incorrect order because the top overlay was never cleaned up properly

I see two possible ways to fix this:

1. For element-bound `useOnEscapePress` events, bind/unbind the passed handler function directly. This seems simple but I'm not sure why the behavior is the way it is in the first place - there may be a reason for it. Example of this change:
```ts
    if (element) element.addEventListener('keydown', escapeCallback)
    else document.addEventListener('keydown', escapeCallback)
    return () => {
      if (element) element.removeEventListener('keydown', escapeCallback)
      else document.removeEventListener('keydown', escapeCallback)
    }
```
2. Or, only use `useOnEscapePress` for global (root-bound) events, and just bind the escape event using react's native `onKeyPress` property instead.

cc @siddharthkp @mattcosta7

siddharthkp:
[...]

PR #1861, opened 2022-02-11
Escape on nested Overlays

siddharthkp:
Fixes https://github.com/primer/react/issues/1854

- [x] Add failing tests
- [x] Add stories for repro
- [x] Only close on overlay at a time (#1854)
- [x] Test ^ in memex

Does not fix in this PR, need to experiment more:
- Don't leak events to global (#1802) (memex doesn't use this anymore, but someone else might in the future). There is a [skipped test](https://github.com/primer/react/pull/1861/files?diff=unified&w=0#diff-d3d3fcb267046e3d15cd1df1a8296520db7954a8a559bfa78ebec1cd606fd793R180) for future fixes.


src/hooks/useOnEscapePress.ts
```diff
@@ -1,23 +1,34 @@
...
 /**
  * Calls all handlers in reverse order
  * @param event The KeyboardEvent generated by the Escape keydown.
  */
 function handleEscape(event: KeyboardEvent) {
-  if (event.key === 'Escape' && !event.defaultPrevented) {
-    for (let i = handlers.length - 1; i >= 0; --i) {
-      handlers[i](event)
+  if (!event.defaultPrevented) {
+    for (const handler of Object.values(registry).reverse()) {
```
colebemis, line 9: Is the order of `Object.values(registry)` consistent/predictable?

colebemis: Would it make more sense to push and pop from a stack (array)?

src/stories/Overlay.stories.tsx
```diff
@@ -211,3 +231,267 @@ export const OverlayOnTopOfOverlay = ({anchorSide}: OverlayProps) => {
...
+                }}
+                onKeyDown={(event: React.KeyboardEvent<HTMLInputElement>) => {
+                  if (event.key === 'Enter') {
+                    setEditing(false)
+                    setTitle((event.target as HTMLInputElement).value)
+                  } else if (event.key === 'Escape') {
+                    setEditing(false)
+                    setTitle(title)
+                    event.preventDefault() // prevent Overlay from closing, this is what we recommend
+                    // event.stopPropagation() // this also works and feels nicer to use
```
colebemis, line 461: Maybe we should update this comment to explain that `event.stopPropagation()` doesn't work here yet

src/hooks/useOverlay.tsx
```diff
@@ -29,6 +29,12 @@ export const useOverlay = ({
   const overlayRef = useProvidedRefOrCreate<HTMLDivElement>(_overlayRef)
   useOpenAndCloseFocus({containerRef: overlayRef, returnFocusRef, initialFocusRef, preventFocusOnOpen})
   useOnOutsideClick({containerRef: overlayRef, ignoreClickRefs, onClickOutside})
-  useOnEscapePress(onEscape)
+
+  /** We only want one overlay to close at a time */
```
colebemis, line 33: Nitpick: do we need a JSDoc-style comment here?
```suggestion
  // We only want one overlay to close at a time
```

colebemis (approved):
This was a tricky bug. Nice fix ✨ Just left a few non-blocking comments

I love that we're building real-world examples in our storybook ❤️

src/hooks/useOnEscapePress.ts
siddharthkp: Yep! I found it more consistent than an array implementation I tried to build myself.

src/stories/Overlay.stories.tsx
siddharthkp: In this case, it does work though because the order of handlers is predictable 😅

src/hooks/useOverlay.tsx
siddharthkp: yep, not sure what I had here. fixed now.

PR #1867, opened 2022-02-16
Add tests for PageLayout

colebemis:
Adds Jests test for PageLayout

Part of https://github.com/github/primer/issues/565


jest.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
     '<rootDir>/src/utils/test-deprecations.tsx',
     '<rootDir>/src/utils/test-helpers.tsx'
   ],
-  testMatch: ['<rootDir>/(src|codemods)/__tests__/**/*.test.[jt]s?(x)', '!**/*.types.test.[jt]s?(x)'],
+  testMatch: ['<rootDir>/(src|codemods)/**/*.test.[jt]s?(x)', '!**/*.types.test.[jt]s?(x)'],
```
siddharthkp, line 12: ooooh collocation 👍

siddharthkp: We might want to exclude stories from showing up in coverage with [`collectCoverageFrom`](https://jestjs.io/docs/configuration#collectcoveragefrom-array)

image

docs/content/drafts/PageLayout.mdx
```diff
@@ -252,7 +252,7 @@ See [storybook](https://primer.style/react/storybook?path=/story/layout-pagelayo
     noUnnecessaryDeps: true,
     adaptsToThemes: true,
     adaptsToScreenSizes: true,
-    fullTestCoverage: false,
+    fullTestCoverage: true,
```
siddharthkp, line 255: Solid work here!

Would be nice to get more coverage on the branches here with all the different options

image

siddharthkp (changes requested):
Looking good! This component is progressing fast! 🚀 

Couple of changes requested:
- Jest config update for coverage
- More branch coverage for props

docs/content/drafts/PageLayout.mdx
colebemis: Adds more tests to cover these branches 👍  There are a few uncovered branches in some internal components (dividers) but I'd have to write some hacky code to cover those cases. Let me know if you comfortable with this level of code coverage

siddharthkp (approved):
ship it!

Issue #1773, opened 2022-01-05
Flash component renders differently in server vs client


peterbe:
**Describe the bug**

We use it like this:
```jsx
import { Flash } from '@primer/components'

...

<Flash variant="warning">
   <p>
      Bla
   </p>
</Flash>
```

When rendered in NextJS on the server, the HTML it generates becomes:
```html
<div class="Flash-sc-4y90bn-0 bIFBCA">
  <p>
    Bla
  </p>
</div>
```

But in the browser, it becomes:
```html
<div class="Flash-sc-4y90bn-0 eTFOFY">
  <p>
    Bla
  </p>
</div>
```

If you have access, please see: https://github.com/github/docs-engineering/issues/1388
If not, open https://docs.github.com/en/rest/reference/repos and note how the "We've recently moved some of the REST API documentation ..." flash message changes color as the page re-renders in the client. 

**Note:** As this is a public repo, please be careful not to include details or screenshots from unreleased GitHub products or features. In most cases, a good bug report should be able to describe the new component without including business logic or feature details, but if you must discuss context relating to an unreleased feature, please open an issue in the private [Design Systems repo](https://github.com/github/design-systems/issues/new/choose) and link to it here.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

siddharthkp:
This is a great bug report! Thanks for taking the time, i'll find out what's up.

PR #1868, opened 2022-02-17
Fixes mismatch in rendered output for theming with server side rendering

siddharthkp:
- Improves SSR for theming
- Fixes https://github.com/primer/react/issues/1773
- Does not completely fix but reduces the surface area of https://github.com/github/docs-internal/pull/24123
- [x] TODO: Update ThemeProvider docs

There's a lot going on here visually, which is hard to explain in text, so here's a [Loom](https://www.loom.com/share/a38a8fc125804dc5a74946e2eaf114ef) (12mins)

### Bug

When using `colorMode=auto` with `ThemeProvider`, 

```jsx
<ThemeProvider colorMode="auto">
  <Flash variant="warning">Warning Flash</Flash>
</ThemeProvider>
```

`colorMode` resolves differently on the server and client:

```js
function getSystemColorMode() {
  if (typeof window !== 'undefined' && window.matchMedia?.('(prefers-color-scheme: dark)')?.matches) {
    return 'night'
  } else {
    return 'day'
  }
}
```

On the client, we have access to the system preferences, so "auto" resolves to the OS settings. On the server, however, we don't have any of that so it always resolves to `day`. So if you have dark mode on your OS (or it's after sunset on automatic) you get different `resolvedColorMode` leading to the infamous warning of mismatch between server and client.

In development environment, this would simply break the page. On production, the error would be suppressed and the page would be stuck on what the server provided (always day/light mode).

Before:

https://user-images.githubusercontent.com/1863771/154459524-56e8cbc1-56e4-4a65-a809-7d062b962d96.mov

### Short term fix (this pull request): 

The short term fix for this to: 
1) Remove the mismatch of output between client and server: Pass the `resolvedColorMode` from the server to the client and use that as the initial color mode. [diff](https://github.com/primer/react/pull/1868/files#diff-693c129d15d32a073c701ec06dc3d82242e49ff8f561501e2bd9f87d33892882R61)
2) Apply the correct theme after resolving color mode on client: Run an `useEffect` on the client to see if the `resolvedColorMode` on the client should be different than what we initially set and fix it. [diff](https://github.com/primer/react/pull/1868/files#diff-693c129d15d32a073c701ec06dc3d82242e49ff8f561501e2bd9f87d33892882R68-R89)

I have made this feature opt in instead of the default so that we don't inject anything into the DOM by default, would like some feedback if that makes sense and what the prop should be called:

```jsx
<ThemeProvider colorMode="auto" preventSSRMismatch>
  <Flash variant="warning">Warning Flash</Flash>
</ThemeProvider>
```

After:

https://user-images.githubusercontent.com/1863771/154459543-91ff865c-61ee-456b-9d6e-6930025e5a3b.mov

### Long term fix (not in this PR):

The core of the problem is that we are not able to respond the client's resolved color mode instantly because we need to wait for the javascript bundle to download and generate new styles on runtime.

The fix for that would be to move towards css variables in components which can respond to changes in theme instantly without waiting for javascript.
To trigger that change, we would need to ship a tiny chunk of javascript embedded in the initial server rendered html that updates the color mode the moment it reaches the browser.

&nbsp;


src/ThemeProvider.tsx
```diff
@@ -17,6 +17,7 @@ export type ThemeProviderProps = {
   colorMode?: ColorModeWithAuto
   dayScheme?: string
   nightScheme?: string
+  preventServerClientMismatch?: boolean
```
siddharthkp, line 20: Start here: 1) Add opt-in feature to prevent output mismatch between server and client. If it's a client only application, they don't need to bother.

src/ThemeProvider.tsx
```diff
@@ -47,21 +48,50 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
...
+    // @ts-ignore This custom variable does not exist on window because we set it outselves
+    typeof window !== 'undefined' ? window.__PRIMER_RESOLVED_SERVER_COLOR_MODE : undefined
+  )
+
   const [colorMode, setColorMode] = React.useState(props.colorMode ?? fallbackColorMode ?? defaultColorMode)
   const [dayScheme, setDayScheme] = React.useState(props.dayScheme ?? fallbackDayScheme ?? defaultDayScheme)
   const [nightScheme, setNightScheme] = React.useState(props.nightScheme ?? fallbackNightScheme ?? defaultNightScheme)
   const systemColorMode = useSystemColorMode()
-  const resolvedColorMode = resolveColorMode(colorMode, systemColorMode)
+  const resolvedColorMode = resolvedColorModePassthrough.current || resolveColorMode(colorMode, systemColorMode)
```
siddharthkp, line 62: Step 3. If there was a `resolvedColorMode` passed on from the server, just use that instead of resolving color mode on the client. This will ensure consistent output between client and server.

siddharthkp, line 62: Step 4. If there was a `resolvedColorModePassthrough` passed on from the server, just use that instead of resolving color mode on the client. This will ensure consistent output between client and server.

src/ThemeProvider.tsx
```diff
@@ -47,21 +48,50 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
 
   // Initialize state
   const theme = props.theme ?? fallbackTheme ?? defaultTheme
+
+  const resolvedColorModePassthrough = React.useRef(
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore This custom variable does not exist on window because we set it outselves
+    typeof window !== 'undefined' ? window.__PRIMER_RESOLVED_SERVER_COLOR_MODE : undefined
+  )
```
siddharthkp, line 56: Step 3. On the client side, check if there was a variable set by the server, pick it up and set it in a `ref`.

src/ThemeProvider.tsx
```diff
@@ -47,21 +48,50 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
...
+            // immediately after that, set the colorMode to what the user passed to respond to system color mode changes
+            setColorMode(colorMode)
+          })
+        }
+
+        resolvedColorModePassthrough.current = null
+      }
+    },
+    [colorMode, systemColorMode]
+  )
```
siddharthkp, line 89: Step 5/5: After mount - If the `resolvedColorMode` passed from the server and the `resolvedColorMode` resolved on the client do not match, update the color mode to respect the client's settings (example: If the color mode is set to "auto" and it resolved to "night" on the client side but was sent as "day" from the server)

colebemis (approved):
This is a great short-term fix! Thanks for the thorough PR description and comments. They were really helpful ❤️

colebemis:
Let's update the ThemeProvider docs in this PR too

siddharthkp:
@colebemis Added docs, also renamed the prop to a nicer `preventSSRMismatch`

pksjce:
Great explanation of the issue! I've seen similar fixes for rehydration issues where we pass data btw server and client through the html.
One thing is I don't understand the need for a boolean. I guess you want to use it as a feature flag? 
Either ways great job! ✅

pksjce (approved):

siddharthkp:
> One thing is I don't understand the need for a boolean. I guess you want to use it as a feature flag?

Yep, making it opt-in instead of enabling it by default for the imperfect short term solution.

src/ThemeProvider.tsx
```diff
@@ -47,21 +48,50 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
...
 
+  // this effect will only run on client
+  React.useEffect(
+    function updateColorModeAfterServerPassthorugh() {
+      const resolvedColorModeOnClient = resolveColorMode(colorMode, systemColorMode)
+
+      if (resolvedColorModePassthrough.current) {
+        // if the resolved color mode passed on from the server is not the resolved color mode on client, change it!
+        if (resolvedColorModePassthrough.current !== resolvedColorModeOnClient) {
+          window.setTimeout(() => {
```
rezrah, line 77: Do you need to clear this timeout?

src/ThemeProvider.tsx
```diff
@@ -47,21 +48,50 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
...
+            // override colorMode to whatever is resolved on the client to get a re-render
+            setColorMode(resolvedColorModeOnClient)
+            // immediately after that, set the colorMode to what the user passed to respond to system color mode changes
+            setColorMode(colorMode)
```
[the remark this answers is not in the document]
rezrah: Also just curious, this feels like a potential race condition. Rather than relying on the same effect to trigger the second rerender, could this have its own?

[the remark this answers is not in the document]
siddharthkp: I honestly don't know, something to watch out for when the time comes 🤔 

The fix would be to use `flushSync` then.

See "What if I don’t want to batch?" under https://github.com/reactwg/react-18/discussions/21

src/ThemeProvider.tsx
siddharthkp: Don't think so :) This is only called once, ever.

rezrah:
Agree with the others. Great PR @siddharthkp.. really appreciated the detailed descriptions too 🙇

src/ThemeProvider.tsx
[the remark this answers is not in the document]
rezrah: Yeah, I suspect we (and everyone else) won't have a choice to begin with. We've been working around it for years now. In longer term, I imagine we'd want to enable it though.

docs/content/theming.md
```diff
@@ -176,6 +176,16 @@ function Example() {
 }
 ```
 
+#### `preventSSRMismatch` prop
+
+For usage with server side rendering, it is recommended to pass `preventSSRMismatch` to ensure the rendered output from the server and browser match even when they resolve "auto" color mode differently.
```
colebemis, line 181: We generally try to avoid passive voice (i.e. "it is recommended")
```suggestion
If you are doing server-side rendering, pass the `preventSSRMismatch` prop to ensure the rendered output from the server and browser match even when they resolve "auto" color mode differently.
```

colebemis (approved):
🚢

mikesurowiec:
@siddharthkp great job on this! I really appreciate the fix 🙏

PR #1881, opened 2022-02-22
Move deprecated components to deprecated folder

pksjce:
Describe your changes here.
Create a deprecated folder and move all existing deprecated components there.
Setup exports subpaths as well

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.changeset/short-spoons-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
```
rezrah, line 2: Can we change to major here?

src/deprecated/index.ts
```diff
@@ -0,0 +1,30 @@
...
+export type {
+  DropdownProps,
+  DropdownCaretProps,
+  DropdownButtonProps,
+  DropdownItemProps,
+  DropdownMenuProps
+} from '../Dropdown'
+export {default as FormGroup} from '../FormGroup'
+export type {FormGroupProps, FormGroupLabelProps} from '../FormGroup'
+export {default as SelectMenu} from '../SelectMenu'
```
rezrah, line 30: Think you're missing the props for `SelectMenu`, which are still in root index.ts

rezrah:
Thanks for raising this @pksjce. Here's some initial feedback.

- I expect Docs will fail to build due to the webpack alias being incorrect for the `@primer/components` dependency in `@primer/gatsby-theme-doctocat`. It will need to be [changed in `gatsby-config.js` here](https://github.com/primer/react/blob/1195336e475441f6f7bc266630070834ad564503/docs/gatsby-config.js#L22). 
  - It would have been easier if we could [pass multiple alias paths](https://webpack.js.org/configuration/resolve/#resolvealias), but this webpack 5 feature doesn't seem to be supported in `gatsby-plugin-alias-imports` plugin. Probably some other options worth exploring. cc. @colebemis  
- There are a few scattered references to `Flex` and other deprecated components in `Getting Started` pages. Can we remove these too please?
- Shall we leave `FormGroup` to @mperrotti to handle as part of his deprecation issues? Easier to potentially roll them back together as they are related.

package.json
```diff
@@ -67,7 +74,7 @@
     "npm": ">=7"
   },
   "dependencies": {
-    "@primer/behaviors": "1.1.0",
+    "@primer/behaviors": "1.0.3",
```
siddharthkp, line 77: :)

mperrotti:
> Shall we leave FormGroup to @mperrotti to handle as part of his deprecation issues? Easier to potentially roll them back together as they are related.

@rezrah - do you want me to push commits to this PR? or open a new one?

package.json
pksjce: lol not sure how that happened. i did pull the latest next branch.

rezrah:
> > deprecation issues? Easier to potentially roll them back together as they are related.
> 
> @rezrah - do you want me to push commits to this PR? or open a new one?

Easier to roll back if you could open a new one please. Get feedback on that, and when it's ready to go, merge it into this branch. Would that work for you?

src/deprecated/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export {default as BorderBox} from '../BorderBox'
+export type {BorderBoxProps} from '../BorderBox'
+export {default as Flex} from '../Flex'
+export type {FlexProps} from '../Flex'
+export {default as Grid} from '../Grid'
+export type {GridProps} from '../Grid'
+export {default as Position, Absolute, Fixed, Relative, Sticky} from '../Position'
+export type {PositionProps, AbsoluteProps, FixedProps, RelativeProps, StickyProps} from '../Position'
+export {default as Dialog} from '../Dialog'
```
rezrah, line 17: I don't think we want to deprecate Dialog, as there isn't technically a Dialog2 component yet. I'm not actually sure what's going on with this tbh, but it seems like Dialog2 is just a docs-specific alias for Dialog. So maybe this was deprecated mistakenly? 

cc. @colebemis do you have any historical context on Dialog/Dialog2?

siddharthkp: Yes! This one gets to live! There is no Dialog1 or Dialog2, there is only Dialog in our codebase 😓

rezrah: Update: @pksjce - we've confirmed that this is a hangover from a previous deprecation. 

We need to do some additional triage into this, so for the time being can we please remove Dialog altogether from this PR?

src/deprecated/index.ts
```diff
@@ -0,0 +1,30 @@
+/** This is the place where we keep components that are deprecated.
+ *  We don't recommend using it in production.
+ *  If you already use them, please move to the suggested alternative components
+ *
+ *  But, they are published on npm and you can import them.
+ *  example: import {FormGroup} from '@primer/react/deprecated
+ */
+
+export {default as BorderBox} from '../BorderBox'
```
mperrotti, line 9: Are we intentionally importing these from `src` instead of moving them to `src/deprecated/`?

pksjce: Well we don't have all the draft component in a separate folder, so I kept it the same for deprecated

src/deprecated/index.ts
pksjce: Done!

.changeset/smooth-cameras-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": major
+---
+
+Move deprecated components to deprecated folder
```
rezrah, line 5: A TODO for later @pksjce, we'll need to put more specific breaking change release notes for the components that have moved in this PR.

docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
```diff
@@ -51,6 +52,7 @@ export default function resolveScope(metastring) {
     ...doctocatComponents,
     ...primerComponents,
     ...(metastring.includes('drafts') ? drafts : {}),
+    ...deprecated,
```
rezrah, line 55: Out of curiosity, why is deprecated different to drafts in that we aren't checking metadata like the preceding line?

siddharthkp: I think it needs to be conditional, otherwise deprecated components will override components with the same name and you will always see deprecated components in scope

```js
...
...primerComponents, // has Button, ActionMenu
...
...deprecated // also has Button, ActionMenu
```

siddharthkp, line 55:
```suggestion
    ...(metastring.includes('deprecated') ? deprecated : {}),
```

rezrah (approved):
Looks good @pksjce ✨ 

Few things we'll need to do in follow-up PRs:

- SelectMenu should have a `@deprecated` [JSDoc annotation](https://jsdoc.app/tags-deprecated.html)

- [Dropdown](https://primer.style/react/deprecated/Dropdown) will need docs updated to remove the 'Use DropdownMenu instead` line. Or @siddharthkp - do you want to do this bit in your PRs? 

- We'll need to update the changeset to mention each component that's deprecated, and provide the migration guides there.

PR #1882, opened 2022-02-22
Export PageLayout from main bundle

colebemis:
```diff
- import {PageLayout} from '@primer/react/drafts'
+ import {PageLayout} from '@primer/react'
```

Part of https://github.com/github/primer/issues/724


pksjce (approved):

.changeset/odd-apes-guess.md
```diff
@@ -0,0 +1,10 @@
+---
+"@primer/react": major
+---
+
+`PageLayout` has been moved from the `drafts` bundle to the main bundle. To upgrade, rewrite your imports accordingly:
```
rezrah, line 5: Minor formatting suggestion:

```suggestion
`PageLayout` is being graduated from the `drafts` bundle to the `main` bundle. 

To upgrade, rewrite your imports accordingly:
```

I think we should opt to use graduated as it asserts the intention behind the move.

rezrah:
Can we also update docs (frontmatter + nav.yml) in this PR @colebemis?
Screenshot 2022-02-23 at 11 05 13

colebemis:
@rezrah Nice catch. Thanks! Updated ✅

pksjce (approved):

PR #1902, opened 2022-02-25
fix control value for anchored position storybook

khiga8:
The storybook control for `anchorAlignment` is broken for the[ useAnchoredPosition storybook.](https://primer.style/react/storybook?path=/story/hooks-useanchoredposition--use-anchored-position).

It appears the `anchorAlignment` value should be `start`, `center`, `end`, based on [this line of code in primer/behaviors](https://github.com/primer/behaviors/blob/main/src/anchored-position.ts#L1).


siddharthkp (approved):
Thanks for catching this!

PR #1909, opened 2022-02-28
Add next major to release workflow

colebemis:
## Motivation

We want to reduce the frequency of major releases of `@primer/react` without slowing down the release cadence of minor and patch releases. 

## Solution

This PR implements workflow changes that allow us to merge breaking changes into a `next-major` branch (instead of `main`) that can be tracked and released independently of minor and patch versions.

image

### DIFF
--- a/.github/workflows/release_candidate.yml
+++ b/.github/workflows/release_candidate.yml
@@ -4,11 +4,23 @@ on:
   push:
     branches:
       - 'changeset-release/main'
+      - 'changeset-release/next-major'
 
 jobs:
   release-candidate:
-    name: Candidate
+    if: github.ref_name == 'changeset-release/main'
+    name: Candidate (@next)
     uses: primer/.github/.github/workflows/release_candidate.yml@main
+    secrets:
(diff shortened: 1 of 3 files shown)


jonrohan:
If this nice workflow diagram is intended to stay in docs somewhere, there's one small update to this block 
image

`Create branch called 'next-major' from 'main' branch`

.github/workflows/release_candidate.yml
```diff
@@ -4,11 +4,23 @@ on:
   push:
     branches:
       - 'changeset-release/main'
+      - 'changeset-release/next-major'
```
jonrohan, line 7:
```suggestion
      - 'changeset-release/**'
```

Not necessary but these could be `**`

jonrohan (approved):
Nice, this has generally worked for us in primer/css. 👍🏻

.github/workflows/release_candidate.yml
colebemis: I wanted to be explicit here because I'm referencing those branch names directly in the workflow `if` statements

PR #1913, opened 2022-03-01
Fix theming with SSR with CSRF safe approach

siddharthkp:
Following up on https://github.com/primer/react/pull/1868, the approach I chose relied on injecting an inline script to set a variable. This doesn't work in [github/docs-internal](https://github.com/github/docs-internal/pull/24123) because secure applications prevent executing inline scripts injected in the DOM 😇 

Inspired by `__NEXT_DATA__` from next.js, I've changed the implementation to inject a `<script type="application/json">` instead of `<script type="text/javascript">`.

[Tested with docs-internal](https://github.com/github/docs-internal/pull/24123/commits/625e6b4f6268e39e836375ac494a3f26f2c89dce#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R16), it works!

### DIFF
--- a/src/ThemeProvider.tsx
+++ b/src/ThemeProvider.tsx
@@ -39,12 +39,11 @@ const ThemeContext = React.createContext<{
 
 // inspired from __NEXT_DATA__, we use application/json to avoid CSRF policy with inline scripts
 const getServerHandoff = () => {
-  if (typeof document !== 'undefined' && document.getElementById('__PRIMER_DATA__')?.textContent) {
-    try {
-      return JSON.parse(document.getElementById('__PRIMER_DATA__')?.textContent as string)
-    } catch (error) {
+  try {
+    const serverData = document?.getElementById('__PRIMER_DATA__')?.textContent
+    if (serverData) return JSON.parse(serverData)
+  } catch (error) {
       // if JSON is invalid, supress error
-    }
   }
   return {}
 }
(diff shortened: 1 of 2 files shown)


src/ThemeProvider.tsx
```diff
@@ -37,6 +37,18 @@ const ThemeContext = React.createContext<{
   setNightScheme: () => null
 })
 
+// inspired from __NEXT_DATA__, we use application/json to avoid CSRF policy with inline scripts
+const getSeverHandoff = () => {
```
pksjce, line 41: typo :)

siddharthkp: loooool, fixed!

src/ThemeProvider.tsx
```diff
@@ -37,6 +37,18 @@ const ThemeContext = React.createContext<{
...
 
+// inspired from __NEXT_DATA__, we use application/json to avoid CSRF policy with inline scripts
+const getServerHandoff = () => {
+  if (typeof document !== 'undefined' && document.getElementById('__PRIMER_DATA__')?.textContent) {
+    try {
+      return JSON.parse(document.getElementById('__PRIMER_DATA__')?.textContent as string)
+    } catch (error) {
+      // if JSON is invalid, supress error
+    }
+  }
```
colebemis, line 48: A possible way to clean up this code:

```suggestion
  try {
    const serverData = document?.getElementById('__PRIMER_DATA__')?.textContent
    if (serverData) {
      return JSON.parse(serverData)
    } 
  } catch (error) {
      // if JSON is invalid, supress error
  }
```

colebemis (approved):
Looks good! This is a clever trick 😅

src/ThemeProvider.tsx
siddharthkp: I like it a lot!

PR #1914, opened 2022-03-01
Menu: Fix typeahead with leading visuals

siddharthkp:
When there is a leading visual present, the `textContent` has an extra space which means we are not able to find the matching item. Add a `trim()` solves this issue.

image

Without trim:  `' Text'`, `' Number'`
With trim: `'Text'`, `'Number'`

### DIFF
--- a/.changeset/slimy-garlics-wash.md
+++ b/.changeset/slimy-garlics-wash.md
@@ -2,4 +2,4 @@
 "@primer/react": patch
 ---
 
-Menu: Fix typeahead with leading visuals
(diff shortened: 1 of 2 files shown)


.changeset/slimy-garlics-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Menu: Fix typeahead with leading visuals
```
colebemis, line 5: Should this say `ActionMenu`?

colebemis (approved):

PR #1915, opened 2022-03-01
Button: Update active styles + use them for aria-expanded

siddharthkp:
2 changes here:

1. Update active style for default variant of Button to match new designs & primer-css. The other variants are the same

| Before | After |
|--------|--------|
| Button with pressed shadow | Button with pressed background |

2. Add `aria-expanded` to reuse active styles. This would be useful for all Buttons that open an Overlay. For now, focused on ActionMenu

| Before | After |
|--------|--------|
button stays in it's default state when menu is open | button is active or pressed when the menu is opened |

#### Question for the reviewer: Is the right way to add a pressed state for Button triggers?

### DIFF
--- a/.changeset/healthy-dots-chew.md
+++ b/.changeset/healthy-dots-chew.md
@@ -1,0 +1,6 @@
+---
+"@primer/react": patch
+---
+
+- Update styles for default variant of Button's active state
+- Use active state for Button when it is used to open an Overlay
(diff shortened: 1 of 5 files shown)


colebemis:
Do we want a hover style on the button when the menu is open?

https://user-images.githubusercontent.com/4608155/156823789-5a5f2767-fdd5-4799-8a75-3e5dd384321f.mp4

siddharthkp:
> Do we want a hover style on the button when the menu is open?

Yep, because you can click the button to close the menu as well, so some feedback would be nice

pksjce:
I like it! :shipit:

pksjce (approved):

PR #1917, opened 2022-03-02
Move remaining deprecated components to src/deprecated

siddharthkp:
BorderBox, Flex, Gird, Position, Dropdown, Position, SelectMenu


rezrah (approved):
nice catch.. :shipit:

PR #1922, opened 2022-03-04
ActionMenu.Button: Fix spacing between text and caret

siddharthkp:
Long story:

Here's a `<Button trailingIcon={TriangleDownIcon}>` which is used as the default trigger for Action menu.

ActionMenu.Button

2 issues here:

1. The horizontal padding around the button contents seems unequal
2. The space between the text and caret is a lot!

These issues arise because caret isn't your usual trailing icon, so the default margins don't work well. @ashygee has a spec for this use case 🙌  [github/primer#253](https://github.com/github/primer/issues/253#:~:text=appear%20off%2Dcenter.-,For%20dropdowns,-%2C%20if%20we%20have)

### 1. The horizontal padding around the button contents seems unequal

The padding looks unequal because optically, it is :) This is because the caret octicon isn't edge-to-edge. It has 4px of padding baked into it, effectively making it a 20px padding!

unequal horizontal padding

To balance that, we apply "margin-right: -4px" to the trailing icon:

image

### 2. The space between the text and caret is a lot!

It is a lot because both the text and icon apply their 8px margin, add to that the additional 4px padding inside the icon.

20px optical margin between text and caret

The fix for this to remove to override margin-left to -4 for the icon:

8px optical margin

Here's a combined before and after: 
before has too much space, after has just the right amount

### DIFF
--- a/.changeset/actionmenu-button-caret-spacing.md
+++ b/.changeset/actionmenu-button-caret-spacing.md
@@ -2,4 +2,4 @@
 '@primer/react': patch
 ---
 
-Fix spacing between Text and caret for ActionMenu.Button
+ActionMenu.Button: Fix spacing between text and caret
--- a/.changeset/actionmenu-button-caret-spacing.md
+++ b/.changeset/actionmenu-button-caret-spacing.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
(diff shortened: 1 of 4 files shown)


ashygee (approved):
🚀  Love this! Ship it!

colebemis (approved):

PR #1933, opened 2022-03-08
Add ToggleSwitch component

mperrotti:
Adds a toggle switch for turning settings on and off and immediately save the change.

Closes https://github.com/github/primer/issues/690

### Screenshots
ToggleSwitchDemo

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
What do we think of calling it `ToggleSwitch` instead of `Switch`? 🤔

docs/content/Switch.mdx
```diff
@@ -0,0 +1,225 @@
...
+  statusLabelPosition?: 'start' | 'end'
+
+<PropsTable>
+  <PropsTableRow name="aria-describedby" type="string" description="The id of the DOM node that describes the switch" />
+  <PropsTableRow
+    name="aria-labelledby"
+    type="string"
+    required
+    description="The id of the DOM node that labels the switch"
+  />
```
colebemis, line 175: What's the difference between "labeling" and "describing"?

docs/content/Switch.mdx
```diff
@@ -0,0 +1,225 @@
...
+  <PropsTableRow name="defaultOn" type="boolean" description="Uncontrolled - whether the switch is turned on" />
+  <PropsTableRow name="disabled" type="boolean" description="Whether the switch is ready for user input" />
+  <PropsTableRow name="isLoading" type="boolean" description="Whether the switch is ready for user input" />
```
colebemis, line 178: Can we change this to `loading`?
```suggestion
  <PropsTableRow name="loading" type="boolean" description="Whether the switch is ready for user input" />
```

Also, is it intentional that the description is the same as the `disabled` prop?

docs/content/Switch.mdx
```diff
@@ -0,0 +1,225 @@
...
+    name="onChange"
+    type="(on: boolean) => void"
+    description="The callback that is called when the switch is toggled on or off"
+  />
+  <PropsTableRow
+    name="onClick"
+    type="(e: MouseEvent) => void"
+    description="The callback that is called when the switch is clicked"
+  />
+  <PropsTableRow name="size" type="'small' | 'medium'" description="Size of the switch" />
```
colebemis, line 190: What's the default size?

docs/content/Switch.mdx
```diff
@@ -0,0 +1,225 @@
...
+  <PropsTableRow
+    name="statusLabelPosition"
+    type="'start' | 'end'"
```
colebemis, line 193: What's the default position?

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+  /** The id of the DOM node that describes the switch */
+  ['aria-describedby']?: string
+  /** The id of the DOM node that labels the switch */
+  ['aria-labelledby']: string
+  /** Uncontrolled - whether the switch is turned on */
+  defaultOn?: boolean
+  /** Whether the switch is ready for user input */
+  disabled?: boolean
+  /** Whether the switch's value is being calculated */
+  isLoading?: boolean
```
colebemis, line 24:
```suggestion
  loading?: boolean
```

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+  disabled,
+  isLoading,
+  on: onProp,
+  onChange,
+  onClick,
+  size,
+  statusLabelPosition,
+  sx: sxProp
+}) => {
+  const isControlled = typeof defaultOn === 'undefined' && typeof onProp !== 'undefined'
```
colebemis, line 223: If `defaultOn` and `on` are defined is the component uncontrolled?

I would expect that the `on` prop is the only prop we check to determine if we're in "controlled" mode

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+  const {theme, colorScheme} = useTheme()
+  const [onState, setOnState] = useProvidedStateOrCreate<boolean>(onProp, onChange, Boolean(defaultOn))
```
colebemis, line 225: `isOn` is a little clearer to me. `onState` makes me think it's an event handler

```suggestion
  const [isOn, setIsOn] = useProvidedStateOrCreate<boolean>(onProp, onChange, Boolean(defaultOn))
```

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+        mx={2}
+        aria-hidden="true"
+        sx={{position: 'relative'}}
+      >
+        <Box textAlign="right" sx={onState ? null : hiddenTextStyles}>
+          On
+        </Box>
+        <Box textAlign="right" sx={onState ? hiddenTextStyles : null}>
+          Off
+        </Box>
```
colebemis, line 264: Is there a reason something like this wouldn't work?
```suggestion
        <Box textAlign="right">
          {isOn ? 'On' : 'Off'}
        </Box>
```

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+            flexBasis="50%"
+            color={acceptsInteraction ? 'fg.default' : 'fg.subtle'}
+            lineHeight="0"
+            sx={{
+              transform: `translateX(${onState ? '100%' : '0'})`,
+              transitionProperty: 'transform',
+              transitionDuration: TRANSITION_DURATION,
+              '> svg': {
+                fill: 'currentcolor'
+              }
```
colebemis, line 309: Is this necessary? Looks like the SVG already has a fill of `currentColor`

```suggestion
```

docs/content/Switch.mdx
```diff
@@ -0,0 +1,225 @@
...
+```jsx live
+<Box display="flex" maxWidth="300px">
+  <Box flexGrow={1} fontSize={2} fontWeight="bold" id="switchLabel">
+    Notifications
+  </Box>
+  <Switch aria-labelledby="switchLabel" />
+</Box>
+```
+
+### Uncontrolled (on)
```
colebemis, line 23:
```suggestion
### Uncontrolled with default value
```

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+      onChange(Boolean(onProp))
+    }
+  }, [onChange, onProp, isControlled])
+
+  return (
+    <Box
+      display="inline-flex"
+      alignItems="center"
+      flexDirection={statusLabelPosition === 'start' ? 'row' : 'row-reverse'}
+      mb={4}
```
colebemis, line 248: Why does this component have a margin bottom?

docs/content/Switch.mdx
```diff
@@ -0,0 +1,225 @@
...
+  <PropsTableRow
+    name="aria-labelledby"
+    type="string"
+    required
+    description="The id of the DOM node that labels the switch"
+  />
+  <PropsTableRow name="defaultOn" type="boolean" description="Uncontrolled - whether the switch is turned on" />
+  <PropsTableRow name="disabled" type="boolean" description="Whether the switch is ready for user input" />
+  <PropsTableRow name="isLoading" type="boolean" description="Whether the switch is ready for user input" />
+  <PropsTableRow name="on" type="boolean" description="Whether the switch is turned on" />
```
colebemis, line 179: Should we call this prop `checked` instead of `on` to align with the `<input type="checkbox">` API?

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+  )
+
+  useEffect(() => {
+    if (onChange && isControlled) {
+      onChange(Boolean(onProp))
+    }
+  }, [onChange, onProp, isControlled])
+
+  return (
+    <Box
```
colebemis, line 244: Should this component render a visually hidden checkbox in case people plan to use this inside a form?

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+
+  &:after {
+    content: '';
+    box-sizing: border-box;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    border-radius: 6px;
```
colebemis, line 126:
```suggestion
    border-radius: ${get('radii.2')};
```

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+          borderColor={
+            onState
+              ? acceptsInteraction
+                ? 'accent.emphasis'
+                : 'neutral.emphasis'
+              : acceptsInteraction
+              ? 'border.default'
+              : 'border.subtle'
+          }
+          borderRadius="6px"
```
colebemis, line 329:
```suggestion
          borderRadius={2}
```

docs/content/Switch.mdx
```diff
@@ -0,0 +1,225 @@
...
+### Left-aligned with label
+
+```jsx live
+<>
+  <Text fontSize={2} fontWeight="bold" id="switchLabel" display="block" mb={1}>
+    Notifications
+  </Text>
+  <Switch statusLabelPosition="end" aria-labelledby="switchLabel" />
+</>
+```
```
colebemis, line 158: Are people allowed to render the toggle with a text label ('on' or 'off')?

src/Switch.tsx
```diff
@@ -0,0 +1,358 @@
...
+        onClick={handleToggleClick}
+        aria-labelledby={ariaLabelledby}
+        aria-describedby={ariaDescribedby}
+        aria-checked={onState}
+        aria-disabled={!acceptsInteraction}
+        role="switch"
+        on={onState}
+        size={size}
+        disabled={!acceptsInteraction}
+        isHighContrast={Boolean(colorScheme?.includes('high_contrast'))}
```
colebemis, line 276: This feels a little wrong but I'm not I can't think of a better alternative at the moment

docs/content/Switch.mdx
mperrotti: The label identifies what the switch is, and the description gives more info about what the switch affects.

For example, if the [wiki setting](https://user-images.githubusercontent.com/2313998/157947699-e2029327-6cd8-4841-b1ab-532719c65daf.png) was converted to use a switch instead of a checkbox, "Wikis" would be the label, and "Wikis host documentation for your repository."

docs/content/Switch.mdx
mperrotti: Not intentional. Thanks for flagging that.

docs/content/Switch.mdx
mperrotti: "medium". I'll add a `defaultValue` prop

src/Switch.tsx
mperrotti: I always want the label to take up the same amount of space whether it's "on" or "off". This isn't much an an issue for English, but it could be for other languages.

For example, French:
on = "au"
off = "désactivé"

src/Switch.tsx
mperrotti: You're right. I forgot to remove this when I made that change.

src/Switch.tsx
mperrotti: I have no idea where this came from. Removing.

docs/content/Switch.mdx
mperrotti: I'm good with that.

src/Switch.tsx
mperrotti: We intentionally used a button with `role="switch"` because this component _should not_ be used in a form. The accessibility team doesn't want checkbox changes to be saved unless the user explicitly clicks a "save" or "submit" button.

cc @alliethu or @ichelsea to keep me correct

docs/content/Switch.mdx
mperrotti: I don't understand what you mean. There's already an "on" and "off" label.

src/Switch.tsx
mperrotti: Yea, this feels hacky, but I couldn't think of anything either. Maybe we could use a shared enum to define theme names, and then I could just check against that?

For example:
```
enum Themes {
  LIGHT = 'light'
  LIGHT_HC = 'light_high_contrast'
  // ...etc
}
```

docs/content/Switch.mdx
colebemis: Ah makes sense. Thank you!

src/Switch.tsx
colebemis: Ah okay. Should we mention that in the docs?

ichelsea: `<button role=switch>` sounds like the right thing to do.

mperrotti: Sure, I'll include something in the docs.

mperrotti: @colebemis - I'm not sure where in the React docs this should go. Could I add a new heading for accessibility?

ichelsea (changes requested):
It's hard to pin point the specific colors in the moving GIFs, but I'd like to make sure we get the colors down before we launch this. 

My specific concerns are the border of the whole element against the background (both on and off). If we have a border that meets 3:1 contrast around the whole toggle then we don't have to worry about the background colors of the switch itself. 

Let me know if a visual would help! Essentially we should have the boundaries of the interactive element hit 3:1 contrast against the background of the main page to identify the element. 

We should also make sure that the icons indicating on / off are hitting 3:1 contrast (I'm skeptical about the blue `|` without checking with a dropper). 

It'd be nice to have the switch nob's have enough of a defined border against the switch's background, but it's not necessary because we have the Off / On text _and_ the off / on symbols.

mperrotti:
@ichelsea - yea, a GIF isn't a great way to tell. If it's easier, you can take a look at the preview deployment: https://primer-react-m6t3qcj4o-primer.vercel.app/react/Switch

I'll do my own testing an make sure it hits 3:1

mperrotti:
Thanks for the reminder @alliethu . I forgot that we made that change.

mperrotti:
> For the off switch I used gray scale 3 that hits 3:1 against the white background (#8B949E)

I was trying to stick to functional variables, but I guess we could go with grey scale 3 as long as it works for all color themes

---

> I feel like the off switch could use something else extra, maybe a darker grey on the knob?

I'm using the same grey we use for our default buttons (`variant="secondary"`). Maybe that doesn't work anymore if we have to darken the border color.

mperrotti:
@ichelsea - the contrast-y borders in the "Off" state looked pretty severe, so I darkened the inner part of the switch too.

Kapture 2022-03-16 at 16 01 35

vdepizzol:
> For the on switch, I bumped up the opacity to 0.75 which brought it to the minimum 3:1 contrast

@ichelsea hm, I don't think we can assume that the color tokens coming from different themes will have the same color contrast? I'm not sure we should be using custom opacity for these color decisions... Looking at https://primer.style/primitives/colors, neither `Neutral` nor `Accent` have 3:1-friendly values. `Neutral.emphasis` gives us 4.54, while `Accent.emphasis` 5.19. We should probably align these color contrast requirements in the Primitives level if we want to drive AA-level compliance. Looping @simurai and @colebemis for color token context.

> [](https://user-images.githubusercontent.com/40274682/158642866-fbf05b53-5eed-4118-b1b6-ddb592f538fa.png)For the off switch I used gray scale 3 that hits 3:1 against the white background (#8B949E)
> Toggle off

I know that in order for us to support 3:1 contrast we'll need darker outlines, but this is looking quite... _wireframey_ 🤔. I talked with @mperrotti and I understand his line of thought in making everything darker to compensate:

> Darker toggle off

But at the same time, an **Off** state that is almost as dark as an **On** state also feels a little odd. Most toggle switches I've seen in the [wild](https://www.justinmind.com/blog/toggle-design-patterns-examples/) have quite a variation of contrast between on/off states. It feels it'd be harder for a colorblind person to scan through different toggles and quickly recognize their states, maybe? 🤔 🤔 🤔 

@mperrotti I wonder if going with an inset/button-style gray would help alleviate things a little bit? For reference, I've [rendered](https://www.figma.com/file/y3RWZloGSkvw3yjRY60ITO/FY22-Q3-%7C-Toggle-switch?node-id=158%3A8242) (only available for hubbers) a piece of the Settings page with updated contrast borders in other controls as well[^1].

Screen Shot 2022-03-16 at 14 45 29


[^1]: will we have to use 3:1 contrast for buttons borders too? 😱

ichelsea:
@vdepizzol yeah I totally get the `off` feeling severe. Those were just my first, simplest options. I agree it feels wireframe-y 😆 . 

> will we have to use 3:1 contrast for buttons borders too? 😱

No, it's [recommended for cognitive reasons](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html#:~:text=If%20a%20button,completion%20of%20activities.), but not a requirement.

We will need to meet border contrast for text input borders, though. Planning on going over that in my form inputs audit.

simurai:
> @vdepizzol I'm not sure we should be using custom opacity for these color decisions...
> We should probably align these color contrast requirements in the Primitives level if we want to drive AA-level compliance.

Yeah, I agree that we shouldn't use opacity to tweak contrast ratios directly in components. If something is missing or needs changes, it should be addressed in Primitives.

It seems most Primer components suffer from not having enough contrast since the [`border.default`](https://primer.style/primitives/colors#border) is only `1.45:1`. And the "how can we have a `3:1` contrast ratio without making it too wireframe-y" is probably a bigger discussion. One thought would be to:

1. Rename the current `Light default` to `Light dimmed`
2. Make `Light default` 100% accessible

Then people that like a more subtle/dimmed look&feel can switch to `Light dimmed`. And `Light default` is free to become 100% AA-level compliant.

Anyways, how can we unblock this PR? 😅  Maybe keep the opacity (for now)? Or another somewhat short term option is to add `switch` colors to [component](https://github.com/primer/primitives/blob/main/data/colors/vars/component_light.ts) variables. E.g. `switch.border: get('scale.gray.4'),`.

src/ToggleSwitch.tsx
```diff
@@ -0,0 +1,333 @@
...
+import styled, {css} from 'styled-components'
+import {variant} from 'styled-system'
+import {Box, Spinner, Text} from '.'
+import {get} from './constants'
+import {useProvidedStateOrCreate} from './hooks'
+import sx, {BetterSystemStyleObject, SxProp} from './sx'
+import VisuallyHidden from './_VisuallyHidden'
+import {useTheme} from './ThemeProvider'
+
+const TRANSITION_DURATION = '80ms'
```
rezrah, line 11: It may just be me, but this felt surprisingly fast. Is this aligned to our other motion-related timings across Primer? Wondering if we have a design token for it somewhere too 🤔 

If it's staying at 80ms, can we include a slower version for users that `prefers-reduced-motion`? https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

docs/content/Switch.mdx
```diff
@@ -0,0 +1,226 @@
...
+  <Text fontSize={2} fontWeight="bold" id="switchLabel" display="block" mb={1}>
+    Notifications
+  </Text>
+  <ToggleSwitch statusLabelPosition="end" aria-labelledby="switchLabel" />
+</>
+```
+
+## Props
+
+/\*\* Whether the "on" and "off" labels should appear before or after the switch.
```
rezrah, line 162: FYI

Screenshot 2022-03-17 at 09 49 19

mperrotti:
I was misunderstanding what @vdepizzol meant about the contrast. I thought he _just_ meant the contrast of the "track" background. I now see that the dark grey track background for the "off" state and the dark blue knob for the "on" state have a similar contrast.

I agree that it would be better to have a bigger difference between the "on" and "off" state. I'm about to push something up...

src/ToggleSwitch.tsx
mperrotti: 80ms is being proposed for the "fast" duration in our new tokens. See [Transition section of the doc](https://docs.google.com/document/d/1Yc5WKc0R0mAQFDPOsS9HgLM7ZFB0lNJHWeeycbi5mvY/edit#heading=h.2q6g3fn7042u) (only available to hubbers).

Good call on `prefers-reduced-motion`. We should turn off the transition entirely for those users.

mperrotti:
I just talked to @ichelsea, and I'm going to create component primitives to define colors that work for each theme instead of relying only on the functional color system.

Using component primitives will let us have more depth, and will simplify the CSS because I won't have to use a pseudo-element with opacity to "hack" different shades of `accent.muted`.

docs/content/Switch.mdx
colebemis: Ah, I meant "without" 🤦 

Are people allowed to render the toggle _without_ a text label ('on' or 'off')?

mperrotti: No - we intentionally don't allow a switch without the "on" or "off" label.

mperrotti:
Here's what we have when using component color primitives defined in [this PR](https://github.com/primer/primitives/pull/309):

Kapture 2022-03-21 at 14 08 09

src/Switch.tsx
mperrotti: This is no longer an issue now that we're using component primitives.

docs/content/ToggleSwitch.mdx
```diff
@@ -187,10 +181,11 @@ render(LoadingToggle)
     type="(e: MouseEvent) => void"
     description="The callback that is called when the switch is clicked"
   />
-  <PropsTableRow name="size" type="'small' | 'medium'" description="Size of the switch" />
+  <PropsTableRow name="size" type="'small' | 'medium'" defaultValue="medium" description="Size of the switch" />
   <PropsTableRow
     name="statusLabelPosition"
     type="'start' | 'end'"
+    defaultValue="start"
```
colebemis, line 188:
```suggestion
    defaultValue="'start'"
```

docs/content/ToggleSwitch.mdx
```diff
@@ -187,10 +181,11 @@ render(LoadingToggle)
     type="(e: MouseEvent) => void"
     description="The callback that is called when the switch is clicked"
   />
-  <PropsTableRow name="size" type="'small' | 'medium'" description="Size of the switch" />
+  <PropsTableRow name="size" type="'small' | 'medium'" defaultValue="medium" description="Size of the switch" />
```
colebemis, line 184:
```suggestion
  <PropsTableRow name="size" type="'small' | 'medium'" defaultValue="'medium'" description="Size of the switch" />
```

src/ToggleSwitch.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+type SwitchButtonProps = {
+  disabled?: boolean
+  checked?: boolean
+  size?: SwitchProps['size']
+} & SxProp
+
+type InnerIconProps = {size?: SwitchProps['size']}
+
+const CircleIcon: React.FC<InnerIconProps> = ({size}) => (
```
colebemis, line 56: Should these be added to Octicons? cc @ashygee @edokoa

colebemis (approved):
Looks good! 👍

src/ToggleSwitch.tsx
mperrotti: I didn't include them because I didn't think we'd ever want to use them in isolation. More than happy to add them if you and other folks think we should.

mperrotti:
@colebemis - could you also have a look at https://github.com/primer/primitives/pull/309 ?

It blocks this PR

mperrotti:
Help wanted: the `themePreval` snapshot test is failing, but I don't understand why.

simurai:
> the `themePreval` snapshot test is failing, but I don't understand why.

Ohh.. the same happened to me here: https://github.com/primer/react/pull/1961#issuecomment-1067659196. You could try:

1. Update the date [here](https://github.com/primer/react/pull/1961/files#diff-672c68aad1badf049e9675d289b68d838a5d80dd44d0256d627a8d5a8603f5a8R3)
2.  Then run `test:update` https://github.com/primer/react/blob/1bfb1284b540457cafa59ef7d39778fefccad484/package.json#L46

mperrotti:
A PR has also been opened for documenting interface guidelines: https://github.com/primer/design/pull/249

mperrotti:
@simurai - that doesn't seem to have worked 🤔 

Could you check my [last commit](https://github.com/primer/react/pull/1933/commits/5aee49df6b0b614746e1ba8e43a0012f7f97f8e7) and let me know if I did something wrong?

I updated the snapshots, but `themePreval` won't update. I even tried running `rm -rf node_modules && npm i` just to be sure.

simurai:
> @mperrotti that doesn't seem to have worked

Ahh, yess.. sorry. I missed the `build` step. So it would be:

1. Update the date [here](https://github.com/primer/react/pull/1961/files#diff-672c68aad1badf049e9675d289b68d838a5d80dd44d0256d627a8d5a8603f5a8R3)
2. Run `build` https://github.com/primer/react/blob/1bfb1284b540457cafa59ef7d39778fefccad484/package.json#L38
3.  Run `test:update` https://github.com/primer/react/blob/1bfb1284b540457cafa59ef7d39778fefccad484/package.json#L46
4. Commit changes

Here the commit https://github.com/primer/react/pull/1933/commits/fa9ff75b24771375e4aaa79565cef008e4b4337f that seems to now pass CI.

Something weird in https://github.com/primer/react/pull/1933/commits/fa9ff75b24771375e4aaa79565cef008e4b4337f are the changes to `package-lock.json`. Those changes were made by simply opening the `mp/switch-component` branch in a Codespace. So I thought maybe ok to commit that too. 🤷

PR #1935, opened 2022-03-08
Add monospace prop to textinput

pksjce:
Describe your changes here.

Added a monospace prop to TextInput component

image

`TextInput` now accepts a prop called `monospace` which is boolean. When this prop is present, the textinput changes `font-family` to `fonts.mono`. All other functionality remains same.

Closes https://github.com/github/primer/issues/757

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/TextInput.mdx
+++ b/docs/content/TextInput.mdx
@@ -84,7 +84,13 @@ TextInput is a form component to add default styling to the native text input.
 ## Monospace text input
 
 ```jsx live
-<TextInput monospace aria-label="personal access token" name="pat" placeholder="github_pat_abcdefg123456789" />
+<TextInput
+  monospace
+  aria-label="personal access token"
+  name="pat"
+  leadingVisual={CheckIcon}
+  placeholder="github_pat_abcdefg123456789"
+/>
 ```
 
 ## Props
--- a/docs/content/TextInput.mdx
+++ b/docs/content/TextInput.mdx
@@ -81,6 +81,12 @@ TextInput is a form component to add default styling to the native text input.
 <TextInput contrast aria-label="Zipcode" name="zipcode" placeholder="Find user" autoComplete="postal-code" />
 ```
 
+## Monospace text input
+
+```jsx live
+<TextInput monospace aria-label="Zipcode" name="zipcode" placeholder="Find user" autoComplete="postal-code" />
+```
+
 ## Props
 
 ### TextInput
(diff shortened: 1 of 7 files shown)


docs/content/TextInput.mdx
```diff
@@ -81,6 +81,12 @@ TextInput is a form component to add default styling to the native text input.
 <TextInput contrast aria-label="Zipcode" name="zipcode" placeholder="Find user" autoComplete="postal-code" />
 ```
 
+## Monospace text input
+
+```jsx live
+<TextInput monospace aria-label="Zipcode" name="zipcode" placeholder="Find user" autoComplete="postal-code" />
```
rezrah, line 87: Although it's just an example, shall we switch to using personal access token instead? I don't think we'd ever want post code in monospace 🤔 , so best to avoid giving that impression.

```suggestion
<TextInput monospace aria-label="personal access token" name="pat" placeholder="github_pat_abcdefg123456789" />
```

pksjce: Ah thanks! This is helpful!

docs/content/TextInput.mdx
```diff
@@ -81,6 +81,12 @@ TextInput is a form component to add default styling to the native text input.
 <TextInput contrast aria-label="Zipcode" name="zipcode" placeholder="Find user" autoComplete="postal-code" />
 ```
 
+## Monospace text input
+
+```jsx live
+<TextInput monospace aria-label="personal access token" name="pat" placeholder="github_pat_abcdefg123456789" />
```
rezrah, line 87: after checking out the deployment preview I realised my earlier suggestion wasn't quite right, my bad.

I think you should switch `placeholder` to value so it looks more like product version. Would be nice to get the CheckIcon in there too!

```suggestion
<TextInput monospace leadingVisual={CheckIcon} aria-label="personal access token" name="pat" value="ghp_abcdefg123456789" />
```

src/_TextInputWrapper.tsx
```diff
@@ -104,6 +105,12 @@ export const TextInputBaseWrapper = styled.span<StyledBaseWrapperProps>`
       background-color: ${get('colors.input.disabledBg')};
       border-color: ${get('colors.border.default')};
     `}
+
+    ${props =>
+    props.monospace &&
+    css`
+      font-family: ${get('fonts.mono')};
```
rezrah, line 112: do we need to tweak the size here too? the new type face looks a bit larger than usual inputs so may look odd adjacent to them. May just be the fixed-pitch of the the monospace font playing tricks on my eyes though. cc. @mperrotti 
Screenshot 2022-03-08 at 12 43 37

pksjce: Yeah I feel the size issue too. Would appreciate help from @mperrotti on this design.

pksjce: Took another detailed look and it may just be how monospace fonts look. Its the same size as the one on the developer settings > PAT generator page. I think I'll go ahead with what we have.

docs/content/TextInput.mdx
pksjce: done 👍

PR #1936, opened 2022-03-08
ThemeProvider: Fix setColorMode

siddharthkp:
https://github.com/primer/react/pull/1868 breaks `setColorMode` by [overriding `colorMode`](https://github.com/primer/react/pull/1868/files#diff-693c129d15d32a073c701ec06dc3d82242e49ff8f561501e2bd9f87d33892882R94) immediately after it is set with `setCodeMode`.  (my bad!)

The fix is straight forward, these are 2 unrelated effects and should be separate.

Before:  

https://user-images.githubusercontent.com/1863771/157272418-590a1635-6cae-4ded-9846-dd5ca79e2ae5.mov

After:

https://user-images.githubusercontent.com/1863771/157272410-6e5e08b1-8e1a-418d-be27-1197155c6d9d.mov

### DIFF
--- a/.changeset/pretty-cougars-walk.md
+++ b/.changeset/pretty-cougars-walk.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+ThemeProvider: Fix `setColorMode`. Broken in `34.6.0`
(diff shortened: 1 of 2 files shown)


rezrah (approved):
LGTM

PR #1937, opened 2022-03-08
v35: Remove type aliases for ActionList

siddharthkp:
Cleaner exports for ActionList types


rezrah (approved):
👍  thanks @siddharthkp

PR #1942, opened 2022-03-09
v35 component doc page fixes

mperrotti:


src/FilteredSearch.tsx
```diff
@@ -7,7 +7,8 @@ const FilteredSearch = styled.div<SxProp>`
   display: flex;
   align-items: stretch;
 
-  summary {
+  summary,
+  > button {
```
rezrah, line 11: Would we need to also guard against someone using `as` prop here to change the underlying tag to an anchor or something else?

rezrah (approved):
Looks good, just one observation.

docs/content/deprecated/Dialog.md
```diff
@@ -1,11 +1,11 @@
 ---
-title: Dialog
+title: Dialog (legacy)
```
pksjce, line 2: Do you think using `legacy` will add to any confusion about which component we are talking about?

pksjce (approved):

src/FilteredSearch.tsx
mperrotti: I don't think there would ever be an anchor or something else rendered here

docs/content/deprecated/Dialog.md
mperrotti: If anything, I think it makes it clearer

PR #1945, opened 2022-03-10
Icon button fixes: Removes iconLabel and adds aria-label to the type

pksjce:
Describe your changes here.

The documentation and stories for `IconButton` were not consistent.
The `iconLabel` is not actually required.
The `aria-label` should be mandatory for `IconButton`

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/Button.mdx
+++ b/docs/content/Button.mdx
@@ -85,7 +85,7 @@ It is recommended to use an octicon here.
 A separate component called `IconButton` is used if the action shows only an icon with no text. This button will remain square in shape.
 
 ```jsx live
-<IconButton icon={SearchIcon}>Search</IconButton>
+<IconButton aria-label="Search" icon={SearchIcon} />
 ```
 
 ### Different sized icon buttons
@@ -94,15 +94,9 @@ A separate component called `IconButton` is used if the action shows only an ico
 
 ```jsx live
 <>
-  <IconButton icon={SearchIcon} size="small">
-    Search
-  </IconButton>
-  <IconButton sx={{ml: 2}} icon={SearchIcon}>
-    Search
-  </IconButton>
-  <IconButton sx={{ml: 2}} icon={SearchIcon} size="large">
-    Search
-  </IconButton>
+  <IconButton aria-label="Search" size="small" icon={SearchIcon}  />
+  <IconButton aria-label="Search" icon={SearchIcon} sx={{ml: 2}} />
+  <IconButton aria-label="Search" size="large" icon={SearchIcon} sx={{ml: 2}} />
 </>
 ```
 
--- a/docs/content/Button.mdx
+++ b/docs/content/Button.mdx
@@ -94,7 +94,7 @@ A separate component called `IconButton` is used if the action shows only an ico
 
 ```jsx live
 <>
-  <IconButton aria-label="Search" size="small" icon={SearchIcon}  />
+  <IconButton aria-label="Search" size="small" icon={SearchIcon} />
   <IconButton aria-label="Search" icon={SearchIcon} sx={{ml: 2}} />
   <IconButton aria-label="Search" size="large" icon={SearchIcon} sx={{ml: 2}} />
 </>
--- a/docs/content/Button.mdx
+++ b/docs/content/Button.mdx
@@ -2,26 +2,26 @@
 componentId: button
 title: Button
 status: Alpha
-source: https://github.com/primer/react/tree/main/src/Button2
-storybook: '/react/storybook?path=/story/composite-components-button2'
+source: https://github.com/primer/react/tree/main/src/Button
+storybook: '/react/storybook?path=/story/composite-components-button'
 description: Use button for the main actions on a page or form.
 ---
 
-import {Button, IconButton, LinkButton} from '@primer/react/drafts'
+import {Button, IconButton, LinkButton} from '@primer/react'
 
 ## Usage
 
 ### Installation
 
 ```js
-import {Button} from '@primer/react/drafts'
+import {Button} from '@primer/react'
 ```
 
 ### Default button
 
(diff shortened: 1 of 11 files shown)


src/Button/types.ts
```diff
@@ -45,7 +45,7 @@ export type IconButtonProps = {
    * This is to be used if it is an icon-only button. Will make text visually hidden
```
rezrah, line 45: do you need this comment anymore?

src/Button/types.ts
```diff
@@ -45,7 +45,7 @@ export type IconButtonProps = {
...
    */
   icon: React.FunctionComponent<IconProps>
-  iconLabel: string
+  'aria-label': string
```
rezrah, line 48: Should be this required? Also, what would the user do if they wanted to apply `aria-labelledby` instead, should we have some logic to only apply the `aria-label` if `aria-labelledby` hasn't been passed in so that takes precedence?

src/__tests__/Button.test.tsx
```diff
@@ -91,8 +91,9 @@ describe('Button', () => {
   })
 
   it('styles icon only button to make it a square', async () => {
-    const container = render(<IconButton icon={SearchIcon} iconLabel="Search icon only button" />)
+    const container = render(<IconButton icon={SearchIcon} aria-label="Search button" />)
```
rezrah, line 94: I think we should have a separate test for the aria-label / aria-labelledby. thoughts?

also a suggestion for your matcher, to be more specific around looking for the aria-label attributes:

```diff
- findByRole('button')
+ getByRole('button', {name: 'Search button'} ) 
```

I don't think you need use an async matcher? 🤔

src/__tests__/Button.test.tsx
```diff
@@ -91,8 +91,9 @@ describe('Button', () => {
...
     const IconOnlyButton = await container.findByRole('button')
     expect(IconOnlyButton).toHaveStyleRule('padding-right', '8px')
+    expect(IconOnlyButton).toMatchSnapshot()
```
rezrah, line 97: I feel like aria attributes fall under logic, so I think it's worth adding a separate test for it like..

```
expect(IconOnlyButton).toHaveAttribute('aria-label', labelThingy)
```

src/Button/types.ts
siddharthkp: Related: What's the right way to add a tooltip to an IconButton, like the one on a file?

image

When composable Tooltip + IconButton, does the aria-label go on both of them or just one?

image

cc @alliethu

mperrotti: @siddharthkp - I actually think we'd want the `aria-label` on the button but not the tooltip.

Here's what I think would give a nice experience for screenreaders

### When the tooltip is acting like a label
```jsx
<button aria-label="Edit this file"><svg>{/* paths and stuff */}</svg></button>
<div>Edit this file</div> // <--- this is a tooltip
```
OR
```jsx
<button aria-labelledby="editTooltip"><svg>{/* paths and stuff */}</svg></button>
<div id="editTooltip">Edit this file</div> // <--- this is a tooltip
```

### When the tooltip is acting like helper/caption text
```jsx
<button aria-label="Edit this file" aria-describedby="editTooltip"><svg>{/* paths and stuff */}</svg></button>
<div id="editTooltip">Only people in this organization can edit this file</div> // <--- this is a tooltip
```

@alliethu - please keep me correct

mperrotti: Also, I just realized our tooltips aren't displayed on focus, only hover. :(

We probably need an a11y audit on these

alliethu: @mperrotti Yes, the `aria-label` should be on the `button`. Also, heads-up that @khiga8 just did a recent release to fix tooltips on our comment box toolbars. Hopefully, you should be able to reference her implementation for the fixes on the PRC end. 

https://github.com/github/accessibility/issues/594#issuecomment-1034219464

alliethu: @mperrotti even more current, @hectahertz is working on Upstreaming tooltip to PVC. The two of you should coordinate! https://github.com/github/accessibility/issues/802

khiga8: Here is the [upstreamed tooltip component](https://primer.style/view-components/components/alpha/tooltip).

pksjce: For this PR, I'll try and support both `aria-label` and `aria-labelled-by` in the types. Hope that works 🤞

siddharthkp (changes requested):
Can we update the first example on the docs as well to use aria-label instead of children: [IconButton#icon-only-button](https://primer-react-mn73fh7d2-primer.vercel.app/react/IconButton#icon-only-button) and on the page for Button: [Button#icon-only-button](https://primer-react-mn73fh7d2-primer.vercel.app/react/Button#icon-only-button)

sorry for not catching this one on the first pass!

we can also remove the "drafts" meta string here :)

siddharthkp (approved):

siddharthkp (approved):
Found a missing spot, fixed it: https://github.com/primer/react/pull/1945/commits/40f8d81de55d76866d291d2ea281ec97483711a2

Ship it :shipit:

docs/content/Button.mdx
```diff
@@ -94,7 +94,7 @@ A separate component called `IconButton` is used if the action shows only an ico
 
 ```jsx live
 <>
-  <IconButton aria-label="Search" size="small" icon={SearchIcon}  />
+  <IconButton aria-label="Search" size="small" icon={SearchIcon} />
```
siddharthkp, line 97: oops, my bad!

PR #1920, opened 2022-03-02
Text input loading state

mperrotti:
Issue: https://github.com/github/primer/issues/156

Adds a loading state to the text input. The loading indicator is positioned to avoid layout shifts in the component when it hides and shows.

#### `loadingIndicatorPosition="auto"` (default behavior)
**Does not have a `leadingVisual` OR `trailingVisual`?**
Show at the end of the input

**Has a `leadingVisual`?**
Show in place of the `leadingVisual`

**Has a `trailingVisual`?**
Show in place of the `trailingVisual`

**Has a `leadingVisual` _and_ `trailingVisual`?**
Show in place of the `trailingVisual`

**Does not have a `leadingVisual`?**
Show at the end of the input

#### `loadingIndicatorPosition="leading"`
Show in place of `leadingVisual` if there is one, or show at the start of the input

#### `loadingIndicatorPosition="trailing"`
Show in place of `trailingVisual` if there is one, or show at the end of the input

### Screenshots

TextInputLoadingDemo

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #1947, opened 2022-03-10
Add trailingAction to TextInput

mperrotti:
Adds the capability to render a trailing action in a text input.

Closes https://github.com/github/primer/issues/748

### Screenshots

Kapture 2022-03-10 at 09 36 01

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


mperrotti:
https://github.com/primer/react/pull/1920 should merge before this

docs/content/TextInput.mdx
```diff
@@ -160,61 +212,63 @@ render(<WithIconAndLoadingIndicator />)
...
+    type={<>string | React.ComponentType</>}
+    description="Visual positioned on the left edge inside the input"
+  />
+  <PropsTableRow name="monospace" type="boolean" defaultValue="false" description="Shows input in monospace font" />
+  <PropsTableRow
+    name="trailingVisual"
+    type={<>string | React.ComponentType</>}
+    description="Visual positioned on the right edge inside the input"
+  />
+  <PropsTableRow
```
mperrotti, line 256: This is the only thing that was added. The rest of the changes are just indentation.

src/_InputLabel.tsx
```diff
@@ -21,6 +21,7 @@ const InputLabel: React.FC<Props & SxProp> = ({children, disabled, required, vis
         display: 'block',
         color: disabled ? 'fg.muted' : 'fg.default',
         cursor: disabled ? 'default' : 'pointer',
+        alignSelf: 'flex-start',
```
mperrotti, line 24: Not relevant to this PR. I just noticed the label was clickable even if the user's cursor wasn't on the text.

docs/content/TextInput.mdx
```diff
@@ -108,6 +108,58 @@ const WithIconAndLoadingIndicator = () => {
...
+        <TextInput.Action
+          onClick={() => {
+            alert('clear input')
+          }}
+          icon={XIcon}
+          iconLabel="Clear input"
+          sx={{color: 'fg.subtle'}}
+        />
+      }
+    />
```
colebemis, line 128: I'm not super confident about this but what do you think about an API like this:

```
```
```suggestion
    <TextInput>
        <TextInput.TrailingAction
          onClick={() => {
            alert('clear input')
          }}
          icon={XIcon}
          iconLabel="Clear input"
          sx={{color: 'fg.subtle'}}
        />
    </TextInput>
```

docs/content/TextInput.mdx
```diff
@@ -108,6 +108,58 @@ const WithIconAndLoadingIndicator = () => {
...
+      trailingAction={
+        <TextInput.Action
+          onClick={() => {
+            alert('clear input')
+          }}
+          icon={XIcon}
+          iconLabel="Clear input"
+          tooltipMessage="Clear"
+          sx={{color: 'fg.subtle'}}
+        />
```
colebemis, line 142: Is there a reason we couldn't wrap TextInput.Action with a Tooltip?

```jsx
<Tooltip aria-label="Clear"><TextInput.Action ... /></Tooltip>
```

docs/content/TextInput.mdx
colebemis: Could we add a props table for `TextInput.Action`?

src/_TextInputInnerAction.tsx
```diff
@@ -0,0 +1,69 @@
...
+  condition: boolean
+  wrapper: (children: React.ReactNode) => React.ReactNode
+}> = ({condition, wrapper, children}) => <>{condition ? wrapper(children) : children}</>
+
+const TextInputAction = forwardRef<HTMLButtonElement, TextInputActionProps>(
+  ({children, icon, iconLabel, sx: sxProp, tooltipMessage, variant, ...rest}, forwardedRef) => {
+    const sx = {
+      ...buttonStyleOverrides,
+      ...sxProp
+    }
```
colebemis, line 31: Should we do a deep merge of the sx prop? https://github.com/primer/react/blob/main/src/PageLayout/PageLayout.tsx#L59

docs/content/TextInput.mdx
colebemis: I imagine consumers don't to want to think about when they should use a tooltip

docs/content/TextInput.mdx
siddharthkp: Oh yeah! Me and @mperrotti went through a bunch of options and this was one of them! 

What I like:
1. It makes TextInput feel like a compound component and then putting children inside it makes sense

What I don't like:
1. Doesn't match the prop for trailingVisual (we _could_ move that inside as well along with leadingVisual)
2. It's not intuitive to put contents inside a TextInput because input can't have children.

docs/content/TextInput.mdx
colebemis: I vote we replace `iconLable` and `tooltipMessage` with an `aria-label` prop that is required if no children are passed

docs/content/TextInput.mdx
mperrotti: @siddharthkp said exactly what I was going to 😅

docs/content/TextInput.mdx
mperrotti: @siddharthkp - I thought there would be a case where the `iconLabel` and `tooltipMessage` are different, but now I'm second-guessing that... 🤔 

I'm going to move forward with @colebemis's suggestion, and we can change that later if we need to.

mperrotti: Should we start without allowing a tooltip for buttons with text labels? This feels like a reasonable limitation, but I don't like that we would just silently swallow the `aria-label`

siddharthkp: > Should we start without allowing a tooltip for buttons with text labels? This feels like a reasonable limitation

That sounds reasonable!

> , but I don't like that we would just silently swallow the aria-label

For actions with text labels, is there a reason not to pass aria-label through to the Button if it's provided?

mperrotti: I decided not to limit it

PR #1952, opened 2022-03-11
Use loose versioning for @primer/octicons-react

colebemis:
Use loose versioning for @primer/octicons-react dependency to prevent consumers from accidentally installing two versions of octicons ([slack thread for context](https://github.slack.com/archives/C01L618AEP9/p1647040084073399?thread_ts=1647036308.411569&cid=C01L618AEP9))


colebemis:
Closing in favor of https://github.com/primer/react/pull/1953/

PR #1953, opened 2022-03-11
Upgrade to @primer/octicons-react v17.0.0

colebemis:


siddharthkp (commented):
Can this result in a breaking change? (related to the [breaking change in octicons](https://github.com/primer/octicons/releases/tag/v17.0.0))

Also, does this mean folks using octicons-react@16 would get 2 copies of octicons now? Example: [blackbird-fe](https://github.com/github/blackbird-fe/blob/be91ddf09525f47c94998c1ef6835a50a2ac9465/package.json) and [others](https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Agithub+path%3A**%2Fpackage.json+%40primer%2Focticons-react+%22%5E16%22)

What's a good way of handling these changes? 🤔

PR #1955, opened 2022-03-14
Button.counter does not have disabled styles

pksjce:
Describe your changes here.

Add disabled color and background to `Button.Counter`.

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/Button/styles.ts
```diff
@@ -22,7 +22,11 @@ export const getVariantStyles = (variant: VariantType = 'default', theme?: Theme
       },
       '&:disabled': {
         color: 'primer.fg.disabled',
-        backgroundColor: 'btn.disabledBg'
+        backgroundColor: 'btn.disabledBg',
+        '[data-component="ButtonCounter"]': {
```
siddharthkp, line 26: If it helps to make the code look nicer, you don't need the quotes:

`[data-component="ButtonCounter"]` is equivalent to `[data-component=ButtonCounter]`

src/Button/styles.ts
```diff
@@ -22,7 +22,11 @@ export const getVariantStyles = (variant: VariantType = 'default', theme?: Theme
       },
       '&:disabled': {
         color: 'primer.fg.disabled',
-        backgroundColor: 'btn.disabledBg'
```
siddharthkp, line 25: `btn.disabledBg` doesn't exist 🤔 

image

src/Button/styles.ts
```diff
@@ -89,6 +96,7 @@ export const getVariantStyles = (variant: VariantType = 'default', theme?: Theme
         backgroundColor: 'btn.danger.disabledBg',
         borderColor: 'btn.danger.disabledBorder',
         '[data-component="ButtonCounter"]': {
+          color: 'btn.danger.disabledText',
```
siddharthkp, line 99: Optional: As this is the same as the disabled button color, we could use `color: inherit` so that matches the parent even when we change the color value for the parent

siddharthkp (changes requested):
Looks good! 

Mentioned 1 bug (not user facing though) and a couple of optional changes

src/Button/Button2.stories.tsx
```diff
@@ -151,9 +157,16 @@ export const blockButton = ({...args}: ButtonProps) => {
   )
 }
 
-export const disabledButton = ({...args}: ButtonProps) => {
+export const DisabledButton = ({...args}: ButtonProps) => {
```
siddharthkp, line 160: Nice! 

Can we also sneak in a quick rename to `Button.stories.tsx`? 😄

src/Button/styles.ts
pksjce: Nice! I guess the quotes are only needed if there's special characters in the selector there.

src/Button/Button2.stories.tsx
pksjce: did that in the other PR I have open!

siddharthkp (approved):
:shipit:

PR #1958, opened 2022-03-14
ActionList: a11y fixes

siddharthkp:
- Closes https://github.com/github/primer/issues/447 (last fix!)

1. Group container should not have role=none when there is no role on the List.

2. Added example for Links with Heading

3. High contrast mode:

    Copying what we do in primer/css:
    
    Before:
    
    https://user-images.githubusercontent.com/1863771/158215934-5603f5a3-1e08-46b9-8678-020b2a0c8b42.mov
    
    After: (the color is chosen by the browser)
    
    https://user-images.githubusercontent.com/1863771/158215947-444b0bc0-e2de-45fc-ac95-3e0f60d7e66a.mov


.changeset/actionlist-a11y-fixes.md
```diff
@@ -0,0 +1,6 @@
+---
+"@primer/react": patch
+---
+
+ActionList: Add focus styles for Windows high contrast mode
+ActionList: Fix incorrect role for ActionList.Group outside ActionMenu
```
colebemis, line 6: FYI you can add multiple changesets in one PR

src/ActionList/Item.tsx
```diff
@@ -159,6 +159,14 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         }
       },
 
+      '@media (forced-colors: active)': {
+        ':focus': {
+          // we set color to be transparent and let the high contrast rules
+          // decide what color with contrast should that be corrected to
+          outline: 'solid 1px transparent !important'
```
colebemis, line 166: Is there documentation about this browser behavior anywhere?

colebemis (approved):

src/ActionList/Item.tsx
pksjce: Also what happens with safari with this media tag?

siddharthkp: > Is there documentation about this browser behavior anywhere?

Yep! Here's the [spec](https://www.w3.org/TR/css-color-adjust-1/#forced-colors-properties), [windows blog post](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/#:~:text=you%E2%80%99ll%20get%20a%20visual%20outline%20on%20the%20element). I got this trick from primer/css [action-list-item.scss](https://github.com/primer/css/blob/main/src/actionlist/action-list-item.scss#L56)

> Also what happens with safari with this media tag?

Safari doesn't support it, so it skips it. (Not sure if there's a way to make this work with windows high contrast + safari)

Issue #1948, opened 2022-03-10
No props typings for `TextInput`


jclem:
**Describe the bug**

The `TextInput` component has no props other than `key` and `ref`:

<img width="457" alt="CleanShot 2022-03-10 at 12 18 59@2x" src="https://user-images.githubusercontent.com/412282/157719002-e8847f45-c05c-4f1f-8fe9-1669b5e8e0ac.png">

Note that the type of `onKeyDown` is inferred as `true` because the effective types for `TextInput` appear to be essentially `any & React.RefAttributes & React.Attributes` (meaning `any` along with `key` and `ref`).

**To Reproduce**

Edit `TextInput` components anywhere, such as in the `TextInput.test.tsx` file.

PR #1959, opened 2022-03-14
Fix TextInput types

colebemis:
## Problem

The type for `TextInput` props was resolving to `any`, causing a bad developer experience for consumers. 

After doing some digging, I found that the issue was caused by this line:

```ts
type TextInputInternalProps = Merge<React.ComponentPropsWithoutRef<typeof UnstyledTextInput>, NonPassthroughProps>
// any & React.RefAttributes & React.Attributes
```

Passing a [styled](https://github.com/styled-components/styled-components) component type to [`React.ComponentPropsWithoutRef`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/190ce439349e1fe23fd467f3f76d66be80d12fa0/types/react/index.d.ts#L845) yields an `any` type.

## Solution

I fixed the type issue by rewriting the `TextInputProps` to avoid using `React.ComponentPropsWithoutRef` with a styled component.

Fixing the `TextInput` type revealed 50+ type errors in our codebase 😅 I fixed those errors in this PR as well.

Closes https://github.com/primer/react/issues/1948


src/TextInput.tsx
```diff
@@ -1,25 +1,34 @@
...
+  | 'width'
+  | 'maxWidth'
+  | 'minWidth'
+  | 'variant'
+  | 'size'
+  | 'validationStatus'
 >
 
-// Note: using ComponentProps instead of ComponentPropsWithoutRef here would cause a type issue where `css` is a required prop.
-type TextInputInternalProps = Merge<React.ComponentPropsWithoutRef<typeof UnstyledTextInput>, NonPassthroughProps>
```
colebemis, line 19: This line was the root of the problem. Passing a [styled](https://github.com/styled-components/styled-components) component type to [`React.ComponentPropsWithoutRef`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/190ce439349e1fe23fd467f3f76d66be80d12fa0/types/react/index.d.ts#L845) yields an `any` type

```ts
import React from 'react'

const StyledComponent = styled.div`...`

type Props = 
  React.ComponentPropsWithoutRef<typeof StyledComponent>
// any
```

src/TextInput.tsx
```diff
@@ -1,25 +1,34 @@
...
+  | 'contrast'
+  | 'disabled'
+  | 'monospace'
+  | 'sx'
+  | 'width'
+  | 'maxWidth'
+  | 'minWidth'
+  | 'variant'
+  | 'size'
+  | 'validationStatus'
```
colebemis, line 25: This type was missing `validationStatus`

src/TextInput.tsx
```diff
@@ -1,25 +1,34 @@
...
 type NonPassthroughProps = {
-  className?: string
   /** @deprecated Use `leadingVisual` or `trailingVisual` prop instead */
   icon?: React.ComponentType<{className?: string}>
   leadingVisual?: string | React.ComponentType<{className?: string}>
   trailingVisual?: string | React.ComponentType<{className?: string}>
 } & Pick<
-  ComponentProps<typeof TextInputWrapper>,
-  'block' | 'contrast' | 'disabled' | 'monospace' | 'sx' | 'width' | 'maxWidth' | 'minWidth' | 'variant' | 'size'
+  StyledWrapperProps,
```
colebemis, line 14: `ComponentProps<>` wasn't necessary here. We should avoid using `ComponentProps<>` when possible because it can yield unexpected results somtimes.

src/TextInput.tsx
```diff
@@ -78,13 +87,12 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputInternalProps>(
       </TextInputWrapper>
     )
   }
-)
+) as PolymorphicForwardRefComponent<'input', TextInputProps>
```
colebemis, line 90: Using the polymorphic type utility from Radix here to implement correct types for the `as` prop.

Eventually, we should deprecate the `as` prop for TextInput. I can't think of a good use case for it now that we have the `Textarea` component

src/TextInputWithTokens.tsx
```diff
@@ -53,7 +52,7 @@ type TextInputWithTokensInternalProps<TokenComponentType extends AnyReactCompone
    * The number of tokens to display before truncating
    */
   visibleTokenCount?: number
-} & TextInputProps
+} & Omit<TextInputProps, 'size'>
```
colebemis, line 55: The `size` prop from `TextInputProps` was clashing with the `size` prop defined above

src/TextInputWithTokens.tsx
```diff
@@ -88,10 +86,7 @@ function TextInputWithTokensInnerComponent<TokenComponentType extends AnyReactCo
     variant: variantProp, // deprecated. use `size` instead
     visibleTokenCount,
     ...rest
-  }: TextInputWithTokensInternalProps<TokenComponentType> & {
-    selectedTokenIndex: number | undefined
-    setSelectedTokenIndex: React.Dispatch<React.SetStateAction<number | undefined>>
-  },
```
colebemis, line 94: I don't think `selectedTokenIndex` and `setSelectedTokenIndex` are props that we use in this component

src/TextInput.tsx
mperrotti: Good to know so we don't make this mistake in the future.

src/TextInputWithTokens.tsx
mperrotti: You're right. I'm not sure how these types made it here 🤔

src/stories/Overlay.stories.tsx
```diff
@@ -311,7 +311,7 @@ export const NestedOverlays = () => {
                   Create a list to organize your starred repositories.
                 </Text>
                 <TextInput placeholder="Name this list" sx={{mb: 2}} />
-                <TextInput as="textarea" placeholder="Write a description" rows="3" sx={{mb: 2, textarea: {p: 2}}} />
+                <TextInput as="textarea" placeholder="Write a description" rows={3} sx={{mb: 2, textarea: {p: 2}}} />
```
mperrotti, line 314: Could we just change this to a Textarea?

mperrotti (approved):
Thanks for fixing all of these. Code looks good, and everything in the preview deployment looks good

src/stories/Overlay.stories.tsx
colebemis: Yes, but until we deprecate the `as` prop for TextInput I think consumers will still do this. If people do this in their apps, I like that we are type-checking this use-case in our codebase.

mperrotti: Ah ok. Let's keep it as-is then.

colebemis:
@dusave tested these changes in Memex and didn't run into any build or type errors

PR #1962, opened 2022-03-15
Stop styled components from swallowing variant type

pksjce:
Use `ComponentProps` from utils/types instead of `ComponentPropsWithRef`.

However, this hack does not work for `IconButton`. Since `IconButton` is new, this won't be blocker hopefully


colebemis:
Could you add type tests for `IconButton` to catch any regressions in the future? Similiar to the TextInput type tests I added here: https://github.com/primer/react/pull/1959/files#diff-3e5c6bd4a5da42e27b8bfec548d9654e67ddc1ed08da5e32b32df2f3b7b8a146

pksjce:
Closing this as I might go for an alternative solution and change the way Button types work

PR #1965, opened 2022-03-15
Add changesets flowchart to contributor docs

colebemis:
Updates our contributor docs to include a flowchart that explains our changesets release workflow

image


rezrah:
Looks great @colebemis.... just a minor few comments:

- The rebase strategy might need some extra explanation, so it's really clear why they need to rebase to avoid the reattribution problem. Maybe using an asterisk and at the bottom `* Use rebase instead of merge to prevent Changesets from altering the original author attributions in the final Changelog` or similar.

- Unless we're planning to do it via workflow (probably better), right now they'll also need to manually close the Release Tracking (next-major) PR

rezrah (approved):
Minor feedback but still good to go 👍

PR #1969, opened 2022-03-16
Fix code example renders

mperrotti:
We were exporting octicons' deprecated `default` export.

### DIFF
--- a/docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
+++ b/docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
@@ -12,13 +12,15 @@ const ReactRouterLink = ({to, ...props}) => {
   return <a href={to} {...props} />
 }
 
+delete octicons.default
+
 export default function resolveScope(metastring) {
   return {
     ...doctocatComponents,
     ...primerComponents,
-    ...octicons,
     ...(metastring.includes('drafts') ? drafts : {}),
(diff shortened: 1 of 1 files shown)


docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
```diff
@@ -12,6 +12,8 @@ const ReactRouterLink = ({to, ...props}) => {
   return <a href={to} {...props} />
 }
 
+delete octicons.default
```
colebemis, line 15: Could you add a comment about why this is necessary?

colebemis (approved):
Thank you for fixing this!

docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
```diff
@@ -12,6 +12,9 @@ const ReactRouterLink = ({to, ...props}) => {
   return <a href={to} {...props} />
 }
 
+// don't attempt to export this below because octicons-react's default export is deprecated
+delete octicons.default
```
colebemis, line 16: This `delete` statement is breaking the build. Going to see if this works
```suggestion
// Exclude octicons-react's default export because it's deprecated
const {default: _, ...octiconComponents} = octicons
```

colebemis (approved):

PR #1973, opened 2022-03-17
PageLayout: Use HTML5 landmark elements

colebemis:
Based on @alliethu's recommendation in her [accessibility review](https://github.com/github/primer/issues/782), the `PageLayout` component now renders [HTML5 landmark elements](https://web.dev/use-landmarks/) (`header`, `aside`, `footer`) to improve the navigation experience for people using assistive technologies (like screen readers). 

html5 landmarks in pagelayout component

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/PageLayout/PageLayout.tsx
+++ b/src/PageLayout/PageLayout.tsx
@@ -114,7 +114,6 @@ const HorizontalDivider: React.FC<DividerProps> = ({variant = 'none', variantWhe
   const {padding} = React.useContext(PageLayoutContext)
   return (
     <Box
-      role="separator"
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       sx={(theme: any) =>
         merge<BetterSystemStyleObject>(
@@ -156,7 +155,6 @@ const verticalDividerVariants = {
 const VerticalDivider: React.FC<DividerProps> = ({variant = 'none', variantWhenNarrow = 'inherit', sx = {}}) => {
   return (
     <Box
-      role="separator"
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       sx={(theme: any) =>
         merge<BetterSystemStyleObject>(
--- a/src/PageLayout/PageLayout.tsx
+++ b/src/PageLayout/PageLayout.tsx
@@ -191,6 +191,7 @@ const Header: React.FC<PageLayoutHeaderProps> = ({
   const {rowGap} = React.useContext(PageLayoutContext)
   return (
     <Box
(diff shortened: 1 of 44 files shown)


src/PageLayout/PageLayout.tsx
```diff
@@ -286,6 +286,7 @@ const Pane: React.FC<PageLayoutPaneProps> = ({
   const computedDividerWhenNarrow = dividerWhenNarrow === 'inherit' ? divider : dividerWhenNarrow
   return (
     <Box
+      as="aside"
```
colebemis, line 289: Waiting on an answer to this question: https://github.com/github/primer/issues/782#issuecomment-1069606557

src/PageLayout/PageLayout.tsx
```diff
@@ -114,7 +114,6 @@ const HorizontalDivider: React.FC<DividerProps> = ({variant = 'none', variantWhe
   const {padding} = React.useContext(PageLayoutContext)
   return (
     <Box
-      role="separator"
```
pksjce, line 117: ooh interesting, what does removing this do?

rezrah (approved):
👍  nice

src/PageLayout/PageLayout.tsx
siddharthkp: If doesn't have role= separator should this have aria-hidden=true instead?

siddharthkp (approved):
other than the Divider, everything looks good to me!

src/PageLayout/PageLayout.tsx
alliethu: Good callout @siddharthkp. I believe my recommendation was to use both `role="presentation"` and `aria-hidden="true"`. @colebemis, I'll update the recommendation in the a11y review issue.

colebemis: Is this necessary to do on a `div` element?

colebemis: Without adding `role` or `aria-hidden`, it like looks divider already does not appear in the accessibility tree: 

CleanShot 2022-03-17 at 16 25 07@2x

What do `role` and `aria-hidden` do for us in this case?

siddharthkp: Ooh because it's an empty div, it would probably just get "Ignored" in the tree

Issue #1972, opened 2022-03-17
404 on React 'LinkButton' page


BlockArchitech:
**Describe the bug**
When opening `https://primer.style/react/LinkButton` - it returns a 404 instead of the LinkButton page

**Note:** As this is a public repo, please be careful not to include details or screenshots from unreleased GitHub products or features. In most cases, a good bug report should be able to describe the new component without including business logic or feature details, but if you must discuss context relating to an unreleased feature, please open an issue in the private [Design Systems repo](https://github.com/github/design-systems/issues/new/choose) and link to it here.

**To Reproduce**
1: Open a web browser
2: Navigate to `https://primer.style/react/LinkButton`
3: Observe

**Expected behavior**
Page to open LinkButton documentation

**Screenshots**
- https://i.imgur.com/dCouOkj.png

**Desktop (please complete the following information):**
 - OS: macOS 12.1
 - Browser Safari 15.2 + Chrome 98.0.4758.102
 - Version (see above)

**Smartphone (please complete the following information):**
 - Device: Google Pixel 5
 - OS: Android 12 QPL Beta
 - Browser: Chrome
 - Version: 99.0.4844/58

BlockArchitech:
This may be important: `ID: cle1:cle1::fnlzz-1647475841792-48e887ff9813`

siddharthkp:
Hi! 

`LinkButton` has been removed in favor of `<Button as="a">Clicky</Button>`. It's a mistake that it's showing up in the nav. Fixed in https://github.com/primer/react/pull/1975

PR #1975, opened 2022-03-17
Docs: Remove LinkButton from navigation

siddharthkp:
Closes https://github.com/primer/react/issues/1972

LinkButton was removed in https://github.com/primer/react/pull/1880, but is still visible in navigation


pksjce (approved):
Thanks!

PR #1977, opened 2022-03-17
ActionMenu: Update memex story with overrides for new button

siddharthkp:
Update memex example with new button to catalog an example of the overrides

Keeping this in our storybook as a repro

### DIFF
--- a/.changeset/large-owls-dance.md
+++ b/.changeset/large-owls-dance.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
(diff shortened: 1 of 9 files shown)


pksjce (approved):

pksjce:
Maybe I should have a button specific story for this.

siddharthkp:
> Maybe I should have a button specific story for this.

@pksjce Up to you! I think the Button gets used in all sorts of ways we can't predict. Would be nice to bring those back to our repo as stories, gives us a chance to iterate over common use cases and make them "official" if we need to

PR #1981, opened 2022-03-17
Firefox bug - fix contrast of <select> options in dark mode

mperrotti:
The options were inheriting the color from the `<select>` parent.

### Screenshots

Before:
Screen Shot 2022-03-17 at 6 16 09 PM

After:
Screen Shot 2022-03-17 at 6 13 54 PM

### Merge checklist

- [x] Added/updated tests
- [na] Added/updated documentation
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Select.tsx
+++ b/src/Select.tsx
@@ -16,6 +16,10 @@ const StyledSelect = styled.select`
   outline: none;
   width: 100%;
 
+  option {
+    color: initial;
+  }
+
   /* colors the select input's placeholder text */
   &:invalid {
     color: ${get('colors.fg.subtle')};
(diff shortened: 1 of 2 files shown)


colebemis (approved):

src/Select.tsx
```diff
@@ -16,6 +16,10 @@ const StyledSelect = styled.select`
   outline: none;
   width: 100%;
 
+  option {
+    color: initial;
+  }
+
```
colebemis, line 22: Now that option is set to `color:initial`, can we delete the styles on line 28?
```
  &:invalid option:not(:first-child) {
    color: ${get('colors.fg.default')};
  }
```

mperrotti: I believe so. Let me test it out...

PR #1984, opened 2022-03-18
ActionMenu: Remove focus trap

siddharthkp:
For ActionMenu to be accessible, it needs [a couple of features](https://github.com/github/primer/issues/488#:~:text=high%20contrast%20mode.-,%5BNew!%5D,-Pressing%20Tab%20while) that negate the focus trap that we have right now:

1. Pressing Tab while Menu is open should close menu and move focus to next focusable element (Shift + Tab should move it to previous)

    <details>
    <summary>See video</summary>    

    https://user-images.githubusercontent.com/1863771/159462569-aaf9ba21-f061-42cc-adb4-43d08995dbf7.mov  
    </details>

2. When Menu is opened with mouse click, the focus should stay on the anchor.

    <details>
    <summary>See video</summary>    

    https://user-images.githubusercontent.com/1863771/159461309-af8890f3-8b18-4364-be2e-677c59533eec.mov
    </details>
    

To make this happen,
1. We need to disable focus trap for ActionMenu 
~2. Render the menu's overlay inline instead of in a portal for natural focus management with Tab to work~

Features that need to be re-implemented after disabling focus trap:
- [x] Check for missing features from [spec](https://github.com/github/primer/issues/518#issuecomment-999104848)
- [x] Add tests
- [x] When the menu is open with click:
  - [x] Pressing Arrow key after menu with click should start keyboard activation [video](https://github.com/primer/react/pull/1984/files#r832003581)
  - [x] Typeahead should activate when the menu is open [video](https://github.com/primer/react/pull/1984/files#r831999178)

~Fixes required after rendering Menu.Overlay inline:~
~- [x] Allow overlay contents to overflow outside so that menu can open with `overflow:visible`. [Broken](https://primer-react-aamc3jrl4-primer.vercel.app/react/storybook?path=/story/internal-components-overlay--memex-nested-overlays) → [Fixed](https://primer-react-git-refs-headssiddharthactionmenu-wi-b12425-primer.vercel.app/react/storybook?path=/story/internal-components-overlay--memex-nested-overlays)~
~- [x] ActionMenu.Overlay should have z-index > 0~
  ~- [ ] Are there reasons to regret [`overflow:visible`](https://github.com/primer/react/pull/1984/files#diff-6e710f790dd517cb3cf843980eb869fe0e48798dee39d99e351594341771e12cR66) on Overlay? [doesn't look like it](https://github.com/primer/react/pull/1984/files#r831992805)~


src/__tests__/ActionMenu.test.tsx
```diff
@@ -58,9 +59,7 @@ describe('ActionMenu', () => {
     const component = HTMLRender(<Example />)
     const button = component.getByText('Toggle Menu')
 
-    // We pass keycode here to navigate a implementation detail in react-testing-library
-    // https://github.com/testing-library/react-testing-library/issues/269#issuecomment-455854112
-    fireEvent.keyDown(button, {key: 'Enter', charCode: 13})
+    fireEvent.keyDown(button, {key: 'Enter'})
```
siddharthkp, line 62: Don't actually need to pass charCode here

src/__tests__/ActionMenu.test.tsx
```diff
@@ -83,6 +82,9 @@ describe('ActionMenu', () => {
 
     fireEvent.click(button)
     const menuItems = await waitFor(() => component.getAllByRole('menuitem'))
+
+    // We pass keycode here to navigate a implementation detail in react-testing-library
+    // https://github.com/testing-library/react-testing-library/issues/269#issuecomment-455854112
     fireEvent.keyPress(menuItems[0], {key: 'Enter', charCode: 13})
```
siddharthkp, line 88: Do need to pass charCode with keyPress though

src/ActionMenu.tsx
```diff
@@ -3,17 +3,19 @@ import {useSSRSafeId} from '@react-aria/ssr'
 import {TriangleDownIcon} from '@primer/octicons-react'
 import {AnchoredOverlay, AnchoredOverlayProps} from './AnchoredOverlay'
 import {OverlayProps} from './Overlay'
-import {useProvidedRefOrCreate, useProvidedStateOrCreate, useMenuInitialFocus, useTypeaheadFocus} from './hooks'
+import {useProvidedRefOrCreate, useProvidedStateOrCreate, useMenuKeyboardNavigation} from './hooks'
```
siddharthkp, line 6: there's a lot happening with keyboard navigation and focus, so moved the effects to `useMenuKeyboardNavigation` which can call other hooks

src/ActionMenu.tsx
```diff
@@ -3,17 +3,19 @@ import {useSSRSafeId} from '@react-aria/ssr'
...
 import {SxProp, merge} from './sx'
 
-type MenuContextProps = Pick<
+export type MenuContextProps = Pick<
   AnchoredOverlayProps,
-  'anchorRef' | 'renderAnchor' | 'open' | 'onOpen' | 'onClose' | 'anchorId'
->
+  'anchorRef' | 'renderAnchor' | 'open' | 'onOpen' | 'anchorId'
+> & {
+  onClose?: (gesture: 'anchor-click' | 'click-outside' | 'escape' | 'tab') => void
```
siddharthkp, line 17: new gesture: 'tab'

src/__tests__/__snapshots__/AnchoredOverlay.test.tsx.snap
```diff
@@ -175,7 +175,7 @@ exports[`AnchoredOverlay should render consistently when open 1`] = `
   height: auto;
   width: auto;
   border-radius: 12px;
-  overflow: hidden;
+  overflow: visible;
```
siddharthkp, line 178: comment: this is just the snapshot, check Overlay for big story

src/hooks/useMenuInitialFocus.ts
```diff
@@ -25,7 +32,11 @@ export const useMenuInitialFocus = (open: boolean, onOpen?: Callback, providedRe
     if (!openingKey || !containerRef.current) return
 
     const iterable = iterateFocusableElements(containerRef.current)
-    if (['ArrowDown', 'Space', 'Enter'].includes(openingKey)) {
+
+    if (openingKey === 'mouse-click') {
```
siddharthkp, line 36: If the menu is opened with a click, we should not focus any items. The focus should remain on the anchor for natural focus management. For example: If you press Tab, it would move to the next element after the Menu instead of entering the Menu

src/Overlay.tsx
```diff
@@ -63,7 +63,7 @@ const StyledOverlay = styled.div<StyledOverlayProps>`
   max-height: ${props => props.maxHeight && heightMap[props.maxHeight]};
   width: ${props => widthMap[props.width || 'auto']};
   border-radius: 12px;
-  overflow: hidden;
+  overflow: visible;
```
siddharthkp, line 66: question: Will this bite us back?

It would be really nice if this _just works_ and doesn't have to be done on case by case basis.

See also, in a completely different context: https://github.com/github/details-dialog-element/issues/82

src/hooks/useMenuInitialFocus.ts
```diff
@@ -25,7 +32,11 @@ export const useMenuInitialFocus = (open: boolean, onOpen?: Callback, providedRe
...
+      if (anchorRef.current) anchorRef.current.focus()
```
siddharthkp, line 37: comment: If the menu is opened with a click, we should not focus any items. The focus should remain on the anchor for natural focus management. For example: If you press Tab, it would move to the next element after the Menu instead of entering the Menu

src/hooks/useMenuKeyboardNavigation.ts
```diff
@@ -0,0 +1,88 @@
...
+import {useTypeaheadFocus} from './useTypeaheadFocus'
+import {MenuContextProps} from '../ActionMenu'
+
+/**
+ * Keyboard navigation is a mix of 4 hooks
+ * 1. useMenuInitialFocus
+ * 2. useTypeaheadFocus
+ * 3. useCloseMenuOnTab
+ * 4. useMoveFocusToMenuItem
+ */
```
siddharthkp, line 13: comment: There is a lot happening in the menu, so I've tried to call the hooks from a single hook as the entry point so that it's easier to read and maintain.

There is a tiny overhead of attaching multiple event listeners. As long as we clean them up after the menu closes, the readability it provides feels like a good tradeoff

src/hooks/useTypeaheadFocus.ts
```diff
@@ -4,12 +4,20 @@ import {useProvidedRefOrCreate} from './useProvidedRefOrCreate'
 
 export const TYPEAHEAD_TIMEOUT = 1000
 
-export const useTypeaheadFocus = (open: boolean, providedRef?: React.RefObject<HTMLElement>) => {
-  const containerRef = useProvidedRefOrCreate(providedRef)
+export const useTypeaheadFocus = (
+  open: boolean,
+  providedContainerRef: React.RefObject<HTMLElement>,
+  providedAnchorRef?: React.RefObject<HTMLElement>
```
siddharthkp, line 10: comment: typeahead should activate when the menu is opened with keyboard and mouse both.

https://user-images.githubusercontent.com/1863771/159458072-396e6226-9d05-4e53-abde-b9b486be053c.mov

src/hooks/useMenuKeyboardNavigation.ts
```diff
@@ -0,0 +1,88 @@
...
+      anchor?.removeEventListener('keydown', handler)
+    }
+  }, [open, onClose, containerRef, anchorRef])
+}
+
+/**
+ * When Arrow Keys are pressed and the focus is on the anchor,
+ * focus should move to a menu item
+ */
+const useMoveFocusToMenuItem = (
```
siddharthkp, line 60: comment: When menu is open with keyboard, the focus moves to the menu item based on the key pressed. However, when menu is opened with click, the focus stays on the anchor. 

Pressing ArrowDown or ArrowUp should move the focus to the right menu item

https://user-images.githubusercontent.com/1863771/159458645-806e05cc-c264-45e4-b9eb-c4828cb8a7f7.mov

src/hooks/useMenuKeyboardNavigation.ts
```diff
@@ -0,0 +1,88 @@
...
+  useMoveFocusToMenuItem(open, containerRef, anchorRef)
+
+  return {openWithFocus}
+}
+
+/**
+ * When Tab or Shift+Tab is pressed, the menu should close
+ * and the focus should naturally move to the next item
+ */
+const useCloseMenuOnTab = (
```
siddharthkp, line 33: before, ignore `Tab` when the menu is open:

https://user-images.githubusercontent.com/1863771/159462565-f0a7527b-4cf1-4a0c-a21b-5c1ada74e14d.mov

after, pressing `Tab` should close the Menu and let the focus go to the next item (or previous item for `Shift+Tab`)

https://user-images.githubusercontent.com/1863771/159462569-aaf9ba21-f061-42cc-adb4-43d08995dbf7.mov

src/ActionMenu.tsx
```diff
@@ -123,8 +124,13 @@ const Overlay: React.FC<MenuOverlayProps> = ({children, align = 'start', ...over
...
       onClose={onClose}
       align={align}
-      overlayProps={overlayProps}
+      overlayProps={{
+        usePortal: false, // render inline for natural focus management
+        sx: merge({zIndex: 1}, propsSx as SxProp), // render above overlay contents (story/internal-components-overlay--memex-nested-overlays)
+        ...overlayProps
+      }}
       focusZoneSettings={{focusOutBehavior: 'wrap'}}
+      focusTrapSettings={{disabled: true}}
```
siddharthkp, line 133: Menu should not have focus trap: Pressing Tab (or Shift + Tab) should close the menu.

(The keyboard navigation with arrow keys is handled by focusZone, not focusTrap)

docs/content/Overlay.mdx
```diff
@@ -240,6 +240,13 @@ See the W3C accessibility recommendations for modals [here](https://www.w3.org/T
     }
   />
   <PropsTableSxRow />
+  <PropsTableRow
+    name="usePortal"
```
colebemis, line 244: `usePortal` makes me think it accepts a hook 🤔 Not sure what would be a better name though. Maybe `withPortal`? 🤷‍♂️

colebemis (approved):
Looks great! Thank you for such a thorough PR description ❤️ 

Just left a minor comment about prop naming

docs/content/Overlay.mdx
siddharthkp: good point!

how do you feel about `<Overlay inPortal={false}>` or `<Overlay insidePortal={false}>` instead of withPortal?

siddharthkp: good point!

how do you feel about these:

`<Overlay inPortal={false}>` or 
`<Overlay renderInPortal={false}>`

I'm tending towards the explicit `renderInPortal`

src/ActionMenu.tsx
```diff
@@ -123,8 +124,13 @@ const Overlay: React.FC<MenuOverlayProps> = ({children, align = 'start', ...over
       onOpen={openWithFocus}
       onClose={onClose}
       align={align}
-      overlayProps={overlayProps}
+      overlayProps={{
+        renderInPortal: false, // render inline for natural focus management
+        sx: merge({zIndex: 1}, propsSx as SxProp), // render above overlay contents (story/internal-components-overlay--memex-nested-overlays)
```
siddharthkp, line 129: this should be > 1 to give room for application folks to put stuff under it 🤔 

will change to 9

siddharthkp:
Found a place where `renderInPortal=false` backfires :(

You can't render it inline with a container that has `position:relative` :(

https://user-images.githubusercontent.com/1863771/159739764-b7be6fe7-1431-4a51-83c5-cb4e697b1bf7.mov

docs/content/Overlay.mdx
colebemis: I like `renderInPortal` 👍

src/Overlay.tsx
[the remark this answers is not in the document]
siddharthkp: update: removed inline rendering!

siddharthkp: update: It did in fact bite :) removed inline rendering!

siddharthkp:
@colebemis Can I get another review. Removed `renderInPortal` and `overflow: visible` from Overlay, it created more problems than it solved. Don't need that for focus mangement to work correctly :)

colebemis (approved):
Poked around in the Storybook and it seems to work as expected 👍

jsartisan:
@siddharthkp Hi, I just tried the actionmenu in the docs. There is one issue i found that the whole pages scrolls down when we press down arrow key. Is that expected or a bug?

jsartisan:
https://user-images.githubusercontent.com/6636360/185539825-0ec1bf4a-e307-4d1a-9c3d-84ee9b117a57.mov

siddharthkp:
Ouch, that's a bit janky. 

🤔 We might need to prevent the scroll event when moving focus for the first time in [useMoveFocusToMenuItem](https://github.com/primer/react/blob/main/src/hooks/useMenuKeyboardNavigation.ts#L86)

siddharthkp:
Tracking in https://github.com/primer/react/issues/2250

PR #1991, opened 2022-03-23
Update snapshots for theme preval

siddharthkp:
After updating primitives, we need to update snapshot


src/__tests__/__snapshots__/themePreval.test.ts.snap
```diff
@@ -3,7 +3,7 @@
 exports[`snapshot theme-preval.js 1`] = `
 "// this file was prevaled
 // This file needs to be a JavaScript file using CommonJS to be compatible with preval
-// Cache bust: 2022-03-14 12:00:00 GMT (This file is cached by our deployment tooling, update this timestamp to rebuild this file)
+// Cache bust: 2021-11-04 12:00:00 GMT (This file is cached by our deployment tooling, update this timestamp to rebuild this file)
```
siddharthkp, line 6: hmm that's not right either

PR #1994, opened 2022-03-23
Add plugin to find source code from storybook

siddharthkp:
Adds a way of jumping from storybook to source code

📺 As seen on design sync: https://github.rewatch.com/video/1cqzuowd7jt5a0ee-di-design-sync-march-31-2022?t=482


pksjce:
Super cool! So Alt+click to open the code? Is there better discoverability of the shortcut?

pksjce (approved):

mperrotti (approved):
I love this _so much_. Because it's not easily discoverable, should we mention it somewhere in the contributor docs?

We don't mention any other Storybook plugins, so I'm totally fine merging this as-is.

siddharthkp:
> We don't mention any other Storybook plugins, so I'm totally fine merging this as-is.

taking the easy path and merging it 😅 

> Is there better discoverability of the shortcut? — @pksjce 
> Because it's not easily discoverable, should we mention it somewhere in the contributor docs? — @mperrotti 

I need ideas on how to make this more discoverable, one option is to add a button in the top bar which activates it instead of pressing Alt. Only slightly better though 😅   

plugins area in the topbar storybook

PR #1998, opened 2022-03-24
Update token components' APIs

mperrotti:
Summary of changes:
- Add `'xlarge'` size option to eventually replace `'extralarge'` to align with our [size naming ADR](https://github.com/github/primer/blob/main/adrs/2022-02-09-size-naming-guidelines.md)
- Update docs to instruct people not to use an Octicon as a `leadingVisual` in small tokens because it's rendered <16px
- Update docs to instruct people not to use an AvatarToken as a small or medium token because it's rendered smaller than our smallest avatar size
- Add `hideRemoveButton` prop to AvatarToken

### Merge checklist

- [x] Added/updated tests
- [n/a] Tested in Chrome
- [n/a] Tested in Firefox
- [n/a] Tested in Safari
- [n/a] Tested in Edge


src/Token/TokenBase.tsx
```diff
@@ -4,13 +4,18 @@ import {variant} from 'styled-system'
 import {get} from '../constants'
 import sx, {SxProp} from '../sx'
 
-export type TokenSizeKeys = 'small' | 'medium' | 'large' | 'extralarge'
+// TODO: remove invalid "extralarge" size name in next breaking change
+// ADR: https://github.com/github/primer/blob/main/adrs/2022-02-09-size-naming-guidelines.md
+export type TokenSizeKeys = 'small' | 'medium' | 'large' | 'extralarge' | 'xlarge'
```
rezrah, line 9: What do you think about adding a `@deprecated` flag to the `extralarge` value here? 

We'd need to bump to minor instead, but feels like we can give early notice that we'll remove it more formally later.

rezrah (approved):
This is great. Love how you've avoided the breaking change here 🙌

src/Token/TokenBase.tsx
colebemis: Is it possible to add the `@deprecated` flag to specific values?

rezrah: It's not supported end-to-end yet, but you can get it working for anyone browsing the codebase by hoisting it into a separate type and then reapplying it in the union or as an intersection. Maybe someone will find it useful 🤷 

E.g. something like:

```suggestion
/** @deprecated to be removed in v36 or something**/
type ExtraLarge = 'extralarge'
export type TokenSizeKeys = 
  | 'small' 
  | 'medium' 
  | 'large'  
  | 'xlarge' 
  | ExtraLarge
```
will show this..
Screenshot 2022-03-29 at 14 29 53

PR #2002, opened 2022-03-28
Update primer/behaviors for esm compatibility

dgreif:
In v`1.1.1`, @primer/behaviors adds better support of imports with ESM.  This resolves issues for projects attempting to import @primer/react in environments that support ESM, but do not recognize the `module` export type. It also allows us to remove the override that caused Jest to tranform the behaviors files before using them.

### Merge checklist

- [x] Added/updated tests


pksjce:
Would you think its useful to add a similar `imports` to @primer/react as well?

pksjce (approved):

dgreif:
I think it has `default` which may give you the same behavior in the end. The down-stream use case only seem to have issues with the behaviors imports, so I think we are good to go

pksjce:
:shipit:

PR #2017, opened 2022-04-11
Replace UnderlineNav component

pksjce:
Describe your changes here.

This PR implements new version of UnderlineNav as per https://github.com/github/primer/issues/892

Why is this work being done?
The current UnderlineNav is outdated and not adequate in the following ways.
1. Uses css class names. These clash with existing implementations.
2. Not accessible as it doesnt set `aria-label` and `aria-current-page`.
3. Does not support icons
4. Has no support for smaller screens

My proposal with this PR is
- We can keep the existing API for `UnderlineNav` while changing all of its internals and adding new API for icons and accessibility
- No deprecation and drafts required.
- Refer to https://primer.style/view-components/components/alpha/underlinenav/#

Pending things to merge this PR.

- [ ] Accessibility updates. Add label and current page
- [ ] Add actions - Ability to add buttons?
- [ ] Allow right align of the nav
- [ ] Sophisticated links
- [ ] Add tests and types
- [ ] Add documentation

Closes https://github.com/github/primer/issues/892

### Screenshots

image

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


langermank:
Hi @pksjce! We have an upcoming major release for PCSS that includes some visual improvements for UnderlineNav. Its nothing drastic, but does improve the experience when items are focused. It also upstreams some CSS from dotcom that uses CSS `content` to create a "fake" bold state, which prevents layout shift when an item becomes active (which I noticed it happening in PRC currently.) https://github.com/primer/css/pull/1821

Let me know if you want to sync on any of this, happy to chat! I'm not sure the scope of this work either but wanted to make sure this release from PCSS is on your radar.

src/UnderlineNav/UnderlineNavLink.jsx
```diff
@@ -0,0 +1,46 @@
...
+    display: 'inline-flex',
+    color: 'fg.default',
+    textAlign: 'center',
+    borderBottom: '2px solid transparent',
+    borderColor: selected ? 'primer.border.active' : 'transparent',
+    textDecoration: 'none',
+    paddingX: get('space.2'),
+    paddingY: get('space.3'),
+    marginRight: get('space.3'),
+    fontSize: get('fontSizes.1'),
```
colebemis, line 24: `get()` shouldn't be necessary since this object is passed to the sx prop

src/UnderlineNav/UnderlineNavLink.jsx
```diff
@@ -0,0 +1,46 @@
...
+    '&:hover, &:focus': {
+      borderColor: 'neutral.muted',
+      transition: '0.2s ease'
+    }
+  }
+  const iconWrapStyles = {
+    display: 'inline-block',
+    marginRight: '8px'
```
colebemis, line 32:
```suggestion
    marginRight: 2
```

src/UnderlineNav/UnderlineNavLink.jsx
```diff
@@ -0,0 +1,46 @@
...
+import {merge} from '../sx'
+import {get} from '../constants'
+
+export const UnderlineNavLink = ({
+  sx: sxProp = {},
+  href = '#',
+  children,
+  selected = false,
+  leadingIcon: LeadingIcon,
+  ...props
```
colebemis, line 12: Should we support the `as` prop to allow people to use React Router-like link components?

src/UnderlineNav/UnderlineNavLink.jsx
pksjce: I'm not sure. whats the alternative? It doesn't seem to work without the get.

pksjce:
Here's a loom of me demo-ing what's possible with the new implementation of `UnderlineNav` https://www.loom.com/share/0a5432ffced946c5901109882e045b8d

maximedegreve:
Hi @pksjce 

I've watched your video and one thing thing I noticed was that the responsive behaviour (more button) behaves a lot different to what we do right now.

In repositories: 
Screenshot 2022-05-13 at 14 32 40

We only show the 3 dots more button when the content overflows for example. 

You might want to sync up with @langermank or @vdepizzol on this behaviour as I saws designs where we just add an additional more item to the end if it doesn't fit. So even the 3 dots behaviour right now in repositories might not be the correct one.

maximedegreve:
Hi @pksjce is there any update on above?

pksjce:
Closing this as I'm continuing the work in https://github.com/primer/react/pull/2142

PR #2020, opened 2022-04-12
ADR: Use Box + sx to author components - DEADLINE: Friday 22 April

siddharthkp:
There are  2 common ways to create components with primer/react:
1. with styled-components
2. with Box + sx

This ADR recommends using the 2nd one exclusively. [See rendered markdown →](https://github.com/primer/react/blob/adr-box-sx/contributor-docs/adrs/adr-005-box-sx.md)

Next step after ADR is accepted: Update [contributor docs](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#component-patterns).


contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,85 @@
...
+
+Proposed
+
+## Context
+
+There are multiple ways to create components within primer/react and composing components from primer/react.
+
+The 2 popular methods are:
+
+1.  Creating components with styled-components
```
colebemis, line 13: How do you feel about this rewording? Feels a bit more clear to me.
```suggestion
In Primer React and consuming applications, we use many different patterns for creating React components. Two common patterns are:

1. Creating components with styled-components
```

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,85 @@
...
+    }))<StyledAvatarProps>`
+      display: inline-block;
+      overflow: hidden;
+      line-height: ${get('lineHeights.condensedUltra')};
+      border-radius: ${props => getBorderRadius(props)};
+      ${sx}
+    `
+    ```
+
+    [show full code example →](https://github.com/primer/react/pull/2019/files?diff=split&w=0)
```
colebemis, line 28:
```suggestion
    [Show full code example →](https://github.com/primer/react/pull/2019/files?diff=split&w=0)
```

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,85 @@
...
+          as="img"
+          alt={alt}
+          sx={merge(styles, sx as SxProp)} // styles needs to merge with props.sx
+          {...props}
+        />
+      )
+    }
+    ```
+
+    [show full code example →](https://github.com/primer/react/pull/2019/files?diff=split&w=0)
```
colebemis, line 52:
```suggestion
    [Show full code example →](https://github.com/primer/react/pull/2019/files?diff=split&w=0)
```

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,85 @@
...
+
+&nbsp;
+
+## Decision
+
+Prefer using method #2: Creating components with Box for the following reasons:
+
+- Better authoring experience with typescript. With Box, we can improve the API and autocomplete for consuming primitives.
+- The styling library used used is an implementation detail and we should be able to replace it. (Avoid leaky abstractions)
```
colebemis, line 61:
```suggestion
- The styling library (i.e. styled-components) becomes an implementation detail and can be changed later with minimal breaking changes for consumers. (Avoids leaky abstractions)
```

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,85 @@
...
+- We have had issues with exporting types, we can increase confidence by keeping the exported types close to what we author.
```
colebemis, line 62: Can you expand on the issues we've had with exporting types?

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,85 @@
...
+
+&nbsp;
+
+This conversation can be extended to overriding stlyes composing components and adding styles to them. We want to use the `sx` prop to add these styles.
```
colebemis, line 66: I'm not sure I understand what you're trying to say here.

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,85 @@
+# ADR 005: Using Box for building components
```
colebemis, line 1: How do you feel about this phrasing?
```suggestion
# ADR 005: Use Box as a building block for all other components
```

colebemis (commented):
Thanks for starting writing this, @siddharthkp! 💖

contributor-docs/adrs/adr-005-box-sx.md
siddharthkp: I like it :)

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,83 @@
...
+
+    [Show full code example →](https://github.com/primer/react/pull/2019/files?diff=split&w=0)
+
+&nbsp;
+
+## Decision
+
+Prefer using method #2: Creating components with Box for the following reasons:
+
+- Better authoring experience with typescript. With Box, we can improve the API and autocomplete for consuming primitives.
```
talum, line 58:
```suggestion
- Better authoring experience with Typescript. With Box, we can improve the API and autocomplete for consuming primitives.
```

What API do we mean here? 🤔 As an outside contributor, this seems confusing to me.

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,83 @@
...
+- The styling library (i.e. styled-components) becomes an implementation detail and can be changed later with minimal breaking changes for consumers. (Avoids leaky abstractions)
```
talum, line 59: In what cases what `styled-comopnents` be appropriate?

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,83 @@
...
+- We have had issues with exporting types, we can increase confidence by keeping the exported types close to what we author.
+
+&nbsp;
+
+This conversation can be extended to overriding stlyes composing components and adding styles to them. We want to use the `sx` prop to add these styles.
```
talum, line 64:
```suggestion
This conversation can be extended to overriding styles composing components and adding styles to them. We want to use the `sx` prop to add these styles.
```

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,83 @@
...
+    }
+    ```
+
+    [Show full code example →](https://github.com/primer/react/pull/2019/files?diff=split&w=0)
+
+&nbsp;
+
+## Decision
+
+Prefer using method #2: Creating components with Box for the following reasons:
```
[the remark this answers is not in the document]
siddharthkp: I'd recommend using the same method in dotcom as well.

We want to improve the way product specific React components can consume primer primitives and those improvements will ship in Box. ([more here](https://github.com/github/primer/discussions/755#discussioncomment-2318144))

Also, replacing styled-components with a more performant styling framework is on our radar, although not soon.

contributor-docs/adrs/adr-005-box-sx.md
siddharthkp: > In what cases what styled-comopnents be appropriate?

I want to say that if you're using primer primitives, then Box + sx is always recommended over styled-components. 

There is of course a slight syntax difference between string with styled-components and an object with Box, which can be a tradeoff. The planned improvements with Box should make that tradeoff easy to make though.

@colebemis any use cases where this wouldn't be true?

contributor-docs/adrs/adr-005-box-sx.md
siddharthkp: Going to remove this and move into an ADR of it's own!

lesliecdubs:

**ADR decision maker:** @rezrah
**Decision deadline:** Friday, 22 April

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,62 @@
...
+
+- Better authoring experience with Typescript. With Box, we can improve the API and autocomplete for consuming primitives.
```
rezrah, line 58: @siddharthkp - can you link to your [research](https://github.com/github/primer/discussions/755#discussioncomment-2318144), which you mentioned in [this comment](https://github.com/primer/react/pull/2020/files#r849468656) from the ADR itself for easier discovery? I think it will help to elaborate why working with Primitives might be easier for anyone reading it later.

contributor-docs/adrs/adr-005-box-sx.md
rezrah: Could the additional [runtime overheads ](https://github.com/github/primer/discussions/755#discussioncomment-2452177) to using Box + sx be highlighted in this ADR as a known risk? Particularly as we're recommending authors use the costlier implementation approach at this time.

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,62 @@
...
+2.  Creating components with Box
+
+    ```tsx
+    const Avatar: React.FC<AvatarProps> = ({size = 20, alt = '', square = false, sx = {}, ...props}) => {
+      const styles = {
+        display: 'inline-block',
+        overflow: 'hidden',
+        lineHeight: 'condensedUltra',
+        borderRadius: getBorderRadius({size, square})
+      }
```
dusave, line 37: Our goals as customers/consumers are:
1. Ease-of-use in terms of authorship (Be it specifying colors, styles, customization, etc)
2. Implementing something that is performant with little to no friction
3. Strongly typed guardrails, so it's easier to implement ("Do I preface this color with `colors.`? Do I pass a theme here? What does 'space.2' mean?")

I see some pros and cons to this approach:
**Pros**:
- This allows consumers to more easily define their styles without as much overhead
- It's coupled with the component, so there is less mental friction
- It still provides a solution that is responsive based on props (i.e. dynamic styles)

**Cons**:
- The biggest drawback I see is that these styles, because they are implemented inside of the component, would be re-computed every single render cycle. While this might be improved with memoization, I still think there's a good deal of perf impact with this approach
- I'm not a huge fan of each component needing to run `merge()` to get styles to work properly, not to mention the added dependency that each app would need to install
- I'm concerned about the mental friction of implementing everything as a `Box` and transmuting them into other components by way of `as`. I always thought of that as an escape hatch more than a common pattern.

I definitely think there are improvements to be made on the implementations of styles, specifically styled-components, but I'm also curious to see what styled-components' plans are now that React 18 is released with all of their improvements and DX hooks for these sorts of libraries.

mattcosta7: I  don't think the intention is to require a dependency on merge @dusave but just that it's the consumers job to merge local and props sx if necessary?

if `Box` is a `Memo` component, could we avoid some overhead by doing a deeper comparison on sx props

```
memo(function Box(props) {...}, function shouldComponentRender(prevProps, nextProps) {
   compare sx deepy, but maintain as object.is equality otherwise
})
```

which might avoid some/many performance differences from inline definitions of sx?

contributor-docs/adrs/adr-005-box-sx.md
```diff
@@ -0,0 +1,62 @@
...
+    `
+    ```
+
+    [Show full code example →](https://github.com/primer/react/pull/2019/files?diff=split&w=0)
+
+2.  Creating components with Box
+
+    ```tsx
+    const Avatar: React.FC<AvatarProps> = ({size = 20, alt = '', square = false, sx = {}, ...props}) => {
+      const styles = {
```
mattcosta7, line 32: should we avoid `styles` here, since it's very close to sounding like inline styles vs `localSx`? or something like that

should we add types here too?

```
const localSx: BetterStyledSystemObject (is this still the type to use) = {}
```

contributor-docs/adrs/adr-005-box-sx.md
mattcosta7: > I'm concerned about the mental friction of implementing everything as a Box and transmuting them into other components by way of as. I always thought of that as an escape hatch more than a common pattern.

this, i think would be just something to extract to module scope and use from there?

mattcosta7 (commented):
Is there any consideration for using `sx` but recommending it be defined in a static scope, with nested classnames, etc being updated instead of the actual sx object?

I think that avoids consumers using styled-components directly, but also avoids the runtime overhead of sx object equality?

contributor-docs/adrs/adr-005-box-sx.md
siddharthkp: `componentStyles` or `baseStyles` can be a better variable here. I don't want to prescriptive about the variable name in this ADR here because it depends on the context. (for example, Button has `baseStyles` + `variantStyles` + `sizeStyles`)

`BetterStyledSystemObject` is the correct type to use, however it could use some improvements as well. [Example sandbox](https://codesandbox.io/s/proud-glitter-q3z5rh?file=/src/Avatar.tsx:884-1309)

mattcosta7: Totally!  Just wanted to make sure we don't cause confusion with inline styles here, more than variable naming!

contributor-docs/adrs/adr-005-box-sx.md
siddharthkp: > Cons:

I'm glad we're talking about this. the sx implementation could use a lot of improvements!

> The biggest drawback I see is that these styles, because they are implemented inside of the component, would be re-computed every single render cycle. While this might be improved with memoization, I still think there's a good deal of perf impact with this approach

You're not wrong. We need work performance improvements to make sure the page doesn't slow down over time because of re-renders.

However, injecting styles is the biggest task in the whole process. Because Box still uses styled-components to inject styles in the end, the same styles are not injected twice. The exception to this is prop-based styles which need to be re-evaluated and the additional work required by styled-system on top of styled-components could be a bad fit for use cases like javascript-based animations.

> I'm not a huge fan of each component needing to run merge() to get styles to work properly, not to mention the added dependency that each app would need to install

Same. I have a few API ideas on how to simplify the API and abstract the merge away from the component.

For the merge dependency, right now it comes from `import {merge} from '@primer/react'`, but with API changes we should be able to remove it from the public API.

> I'm concerned about the mental friction of implementing everything as a Box and transmuting them into other components by way of as. I always thought of that as an escape hatch more than a common pattern.

Can you expand on the friction a bit more.

I think you only need to drop down to Box when you're building a new atomic component. `<Box as="img">` is equivalent to `styled.img`.

For most use cases, you should be able to start with a primer/react component, especially as we get more layout components like Stack. However, to give better html semantics, `as` prop would be the way to go: `<Stack as="section"/>`

siddharthkp:
> Is there any consideration for using sx but recommending it be defined in a static scope, with nested classnames, etc being updated instead of the actual sx object?
> I think that avoids consumers using styled-components directly, but also avoids the runtime overhead of sx object equality?

Not yet, I only have that in a prototypes outside of primer/react 😅

rezrah (approved):

contributor-docs/adrs/adr-005-box-sx.md
lesliecdubs: Would it be useful to amend the ADR to include some of these additional details for context?

siddharthkp: Going to create a new ADR for types all together, there's more to discuss there because we don't have a standard way right now to propose :)

contributor-docs/adrs/adr-005-box-sx.md
mattcosta7: @siddharthkp @colebemis related to 

> I don't think there's anything you can do with styled-components that you can't do with Box + sx.

Should Primer expose a replacement for the styled-components built in version of `keyframes` and `createGlobalStyle` (even if they're just re-exports for now)?  I believe they may be the only apis from styled-components that consumers would still need to call directly.

This might enable applications that are "all in" on using primer primitives for styling, and allow for things like an eslint plugin that bans internal usage of styled-components directly, directing consumers to the primer-based alternatives?

siddharthkp: I like that idea 🤔 

Just to make sure, could this result in multiple versions of styled-components on the page because of version mismatch?

mattcosta7: As long as it's still just a peer of primer I don't think there's any risk of multiple versions for sc.

I think that the only limit on styled components be via an eslint rule saying, use box and ..... Instead of using styled components directly, which should provide a nice way to nudge other sc usage towards primer apis

PR #2027, opened 2022-04-18
[WIP] NavigationList API

colebemis:


docs/content/NavList.mdx
```diff
@@ -0,0 +1,536 @@
...
+
+</details>
+
+### With sub-items
+
+```jsx
+<NavList>
+  <NavList.Item href="/branches">Branches</NavList.Item>
+  <NavList.Item href="/tags">Tags</NavList.Item>
+  <NavList.Item>
```
siddharthkp, line 286: Should we give this item a special name? `NavList.Collapsable`, `NavList.Section`, `NavList.Accordion`, `NavList.Details`



```jsx
<NavList>
  <NavList.Item href="/branches">Branches</NavList.Item>
  <NavList.Item href="/tags">Tags</NavList.Item>
  <NavList.Collapsable title="Actions">
    <NavList.Item href="/branches">Branches</NavList.Item>
    <NavList.Item href="/tags">Tags</NavList.Item>
  </NavList.Collapsable/>
</NavList>
```

Section sounds nice but easy to confuse it with Group

siddharthkp: Looking at other primer implementations for inspiration,

primer/css uses [`.ActionList-item .ActionList-item--hasSubItem`](https://github.com/primer/css/blob/50221a5486d28c54dc1f578b459addd867bd01d3/src/actionlist/action-list-item.scss#L34), primer/view-components seems to [infers it from children](https://cs.github.com/github/github/blob/646833c1794e59114a06f223bf7dbe1d2ac3190f/app/components/primer/experimental/navigation_list.rb#L77)

docs/content/NavList.mdx
```diff
@@ -0,0 +1,536 @@
...
+<details>
+  <summary>Rendered HTML</summary>
+
+```html
+<nav>
+  <ul role="list">
+    <li><a href="/">Dashboard</a></li>
+    <li><a href="/pulls">Pull requests</a></li>
+    <li><a href="/issues">Issues</a></li>
+    <div role="separator"></div>
```
langermank, line 340: I think you need `aria-hidden="true"` here as well

docs/content/NavList.mdx
```diff
@@ -0,0 +1,536 @@
...
+        The URL that the item navigates to. <InlineCode>href</InlineCode> is passed to the underlying{' '}
+        <InlineCode>&lt;a&gt;</InlineCode> element. If <InlineCode>as</InlineCode> is specified, the component may need
+        different props. If the item contains sub-items, the item is rendered as a{' '}
+        <InlineCode>&lt;button&gt;</InlineCode> and <InlineCode>href</InlineCode> is ignored.
+      </>
+    }
+  />
+  <PropsTableRow
+    name="current"
+    type="boolean"
```
langermank, line 432: `aria-current` can be `page` or `location` (string)

langermank (commented):
Looking good!

docs/content/NavList.mdx
colebemis: When would we use `page` vs `location`?

colebemis: From [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#values):

> `location`:
> Represents the current location within an environment or context such as the image that is visually highlighted as the current component of a flow chart.

I'm curious when we'd use NavList to link to a "location," as defined above.

docs/content/NavList.mdx
colebemis: I'm curious what purpose `role` would serve if `aria-hidden` is set to `true` cc @ichelsea

docs/content/NavList.mdx
langermank: `location` would be used for hash links. Here's a nice [definition](https://www.digitala11y.com/aria-current-state/):

> page = Represents the current page within a set of pages
> location = Represents the current location within an environment or context

There are more options for `aria-current`, but I think these are the only two that make sense for a navigational component.

docs/content/NavList.mdx
langermank: Good question, I think I added that based on the [a11y audit](https://github.com/github/primer/issues/651) but didn't research it myself.

docs/content/NavList.mdx
colebemis: Ah I see. Thanks for explaining! So, for example, the current link in a table of contents would have `aria-current="location"`?

I wonder if it makes sense to just have this component accept `aria-current` directly, instead of abstracting it behind a `current` prop. What do you think?

docs/content/NavList.mdx
colebemis: This is amazing feedback. Thank you @siddharthkp! I've scheduled some time to discuss the nuances of these API options. Feel free to move the calendar event if another time is more convenient :)

Another API consideration: TreeView/TreeGrid will need to support similar expand/collapse functionality. I would love to pick an API that we can share across all three components

colebemis: > /* assuming that's good, and subitem does not need to restricted to just text

Good question! @vdepizzol @langermank Should subitem support all the same functionality as item or is it more restrictive?

langermank: It's kind of a long story, _but_, I refactored `NavList` to use semantic `<button>` elements for the collapse items (which originally had a click handler on `<li>`) allowing us to remove a bunch of JS. After I made that same change to `TreeView`, I got feedback from the a11y team that we actually need to move that click handler back up to the `<li>` for trees. It has to do with the `role="tree"` and child "role=treeitem"` markup and navigating the dom with a screen reader. If the child `<li>` is `treeitem` we want the screenreader to focus that item, so having a click handler elsewhere is confusing. In PVC that change requires another PCSS release and more JS refactor, so I have a [followup ticket](https://github.com/github/primer/issues/910) for that. 

So basically.. I think we should try and maintain a `<button>` for `NavList` and remove `<button>` from `TreeView`, meaning those two items have slightly different markup.

> restricted to just text

Both items can have all the same slots, but I've added two additional slots that `TreeView` utilizes: `leadingAction` and `trailingAction`. I created slots with CSS grid, so `ActionList-content` always has these slots available, but so far only `TreeView` and collapsible items have used the action slots. 

`grid-template-areas: 'leadingAction leadingVisual label trailingVisual trailingAction'`

docs/content/NavList.mdx
langermank: > the current link in a table of contents would have aria-current="location"

Yup! In some of the Settings pages we link to subheadings, and we still want to show that "page" as active even if the link in the NavList is linking to #some-id

> I wonder if it makes sense to just have this component accept aria-current directly, instead of abstracting it behind a current prop

I think that makes more sense too. I wonder if there's a way to help the end user decide which option to use (page or location) since I think its a lesser known aria label. Or maybe we can infer it based on what kind of link is passed in?

docs/content/NavList.mdx
colebemis: @siddharthkp and I tentatively decided to move forward with option 2:

```jsx
<NavList>
  <NavList.Item href="/branches">Branches</NavList.Item>
  <NavList.Item href="/tags">Tags</NavList.Item>
  <NavList.Item>
      Actions
      <NavList.LeadingVisual></ActionIcon/></NavList.LeadingVisual>
      <NavList.SubNav>
        <NavList.Item href="/actions">General</NavList.Item>
        <NavList.Item href="/actions/runners">Runners</NavList.Item>
      </NavList.SubNav>
  </NavList.Item/>
</NavList>
```

If anyone has strong opinions against this API, please let me know

docs/content/NavList.mdx
colebemis: I removed the `current` prop in favor of the native `aria-current` prop 👍

docs/content/NavList.mdx
camertron: As @siddharthkp mentioned, the PVC `NavigationList` infers whether or not it should expand/collapse by determining if it has any sub-items, i.e. any item that has sub-items can be expanded/collapsed. That seemed like a reasonable assumption at the time. Will there ever be a need for a list of non-collapsible sub-items?

The API you've chosen makes sense, but I do wonder if simply using `<NavList.SubItem ...>` would be closer to the PVC API. For example:

```jsx
<NavList>
  <NavList.Item href="/branches">
    Branches
    <NavList.SubItem href="/actions">General</NavList.SubItem>
    <NavList.SubItem href="/actions/runners">Runners</NavList.SubItem>
  </NavList.Item/>
</NavList>
```

My understanding is navigation lists only ever have one nesting level.

docs/content/NavList.mdx
```diff
@@ -0,0 +1,523 @@
...
+</nav>
+```
+
+</details>
+
+### With groups
+
+```jsx
+<NavList>
+  <NavList.Group title="Overview">
```
camertron, line 241: I assume groups in `NavList` are analogous to the concept of sections in `NavigationList`? If so, why not call them sections in `NavList` as well?

camertron (approved):
Looks good to me, just left a few non-blocking comments. One remaining question: why is the component called `NavList` instead of `NavigationList`?

siddharthkp (approved):
:shipit:

docs/content/NavList.mdx
```diff
@@ -0,0 +1,523 @@
...
+  <NavList.Item href="/" aria-current="page">
+    Dashboard
+  </NavList.Item>
+  <NavList.Item href="/pulls">Pull requests</NavList.Item>
+  <NavList.Item href="/issues">Issues</NavList.Item>
+</NavList>
+```
+
+<details>
+  <summary>Rendered HTML</summary>
```
mperrotti, line 26: Is the `<details>` and `<summary>` content meant to be inside the `html` codefence?

docs/content/NavList.mdx
```diff
@@ -0,0 +1,523 @@
...
+
+```html
+<nav>
+  <ul role="list">
```
mperrotti, line 30: Do we need to render `role="list"` if we're using a `<ul>`?

docs/content/NavList.mdx
mperrotti: > My understanding is navigation lists only ever have one nesting level.

I think you're correct. We're working on a `TreeView` component to support more levels of nesting.

mperrotti (approved):

docs/content/NavList.mdx
colebemis: I called it `Group` because that what it's called in`ActionList`: https://primer.style/react/ActionList#with-groups

It looks like "section" and "group" are used interchangeably in [some of our docs](https://primer.style/css/components/action-list#divider-with-label-text). I propose that we standardize on "group" in Primer React and ViewComponents because it aligns with the ARIA role ([group](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/group_role)). Here's some [more context](https://github.com/github/primer/issues/650#issuecomment-1029340402)

docs/content/NavList.mdx
colebemis: No. These were a temporary way to get review on the rendered HTML without cluttering the document too much. I'll remove these `<details>` sections when I implement the component.

docs/content/NavList.mdx
colebemis: There's a bug in some browsers that removes the list semantics from `ul` if the element has `list-style: none` ([article](https://www.scottohara.me/blog/2019/01/12/lists-and-safari.html)). `role="list"` restores the list semantics so it's properly announced by screenreaders. @alliethu or @ichelsea please correct me if I'm wrong 🙏

colebemis:
> why is the component called NavList instead of NavigationList?

@camertron we had a conversation about "NavigationList" vs "NavList" in [Slack](https://github.slack.com/archives/C02V74FB6KS/p1650400317433559). The consensus was to rename NavigationList to NavList in both React and ViewComponents to save characters and align with UnderlineNav and the `<nav>` element

docs/content/NavList.mdx
colebemis: > I do wonder if simply using <NavList.SubItem ...> would be closer to the PVC API

@siddharthkp and I are leaning towards the `<NavList.SubNav>` component because it will allow us to create a predictable and consistent API across `NavList`, `TreeView`, and `TreeGrid`. Would a similar API work for ViewComponents?

camertron: @colebemis yeah that works for me 😄

docs/content/NavList.mdx
ichelsea: cc: @alliethu I'm not entirely sure the nuances on this one

docs/content/NavList.mdx
ichelsea: Yes @colebemis! That's correct 🙂 . It's a failsafe for the browsers that remove it.

PR #2033, opened 2022-04-19
TextInput trailing action design and API update

mperrotti:
Changes:
- Inner action's hover bg should not touch the text input edges
- Increases touch target area of the inner action button
- Deprecation: we only want to allow inner action buttons to be icon buttons
- Deprecation: we only want to allow inner action buttons to be the 'invisible' variant

### Screenshots

Before:
innerAction-before

After:
innerAction-after

### Merge checklist

- [n/a] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/_TextInputInnerAction.tsx
+++ b/src/_TextInputInnerAction.tsx
@@ -5,13 +5,14 @@ import {ButtonProps} from './Button'
 import {BetterSystemStyleObject, merge, SxProp} from './sx'
 
 type TextInputActionProps = Omit<React.HTMLProps<HTMLButtonElement>, 'aria-label' | 'size'> & {
-  /** @deprecated Text input action buttons will only use icon buttons */
+  /** @deprecated Text input action buttons should only use icon buttons */
   children?: React.ReactNode
   /** Text that appears in a tooltip. If an icon is passed, this is also used as the label used by assistive technologies. */
   ['aria-label']?: string
   /** The icon to render inside the button */
   icon?: React.FunctionComponent<IconProps>
   /**
+   * @deprecated Text input action buttons should only use the 'invisible' button variant
    * Determine's the styles on a button one of 'default' | 'primary' | 'invisible' | 'danger'
    */
   variant?: ButtonProps['variant']
--- a/src/_TextInputInnerAction.tsx
+++ b/src/_TextInputInnerAction.tsx
@@ -16,7 +16,29 @@ type TextInputActionProps = Omit<React.HTMLProps<HTMLButtonElement>, 'aria-label
 } & SxProp
 
 const invisibleButtonStyleOverrides = {
-  color: 'fg.default'
+  color: 'fg.default',
+  margin: 1,
+  paddingTop: '2px',
+  paddingRight: '4px',
+  paddingBottom: '2px',
+  paddingLeft: '4px',
+  position: 'relative',
+
+  '@media (pointer: fine)': {
+    ':after': {
+      content: '""',
+      position: 'absolute',
+      left: 0,
+      right: 0,
+      transform: 'translateY(-50%)',
+      top: '50%',
+      minHeight: '44px'
+    }
+  }
+}
+
+const solidButtonStyleOverrides = {
+  margin: 1
 }
 
 const ConditionalTooltip: React.FC<{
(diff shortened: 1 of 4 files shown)


siddharthkp:
> Inner action's hover bg should not touch the text input edges

The tooltip container seems to be positioned awkwardly here. You can hover and get the tooltip but you're not on the button yet so a click doesn't trigger the button.

https://user-images.githubusercontent.com/1863771/164051492-a74a8845-3f30-4f72-894b-9bb0bcf4b6a9.mp4

&nbsp;

> Increases touch target area of the inner action button

Does this meet our minimum button size for a11y? How do we solve for this when it's limited by the container's size?

image

mperrotti:
> The tooltip container seems to be positioned awkwardly here

Great catch! I'll fix that.

---

> 24x24, does this meet our minimum button size for a11y? 

We increase the size of the clickable area for `pointer: coarse`: https://github.com/primer/react/pull/2033/files#diff-6d020864785e702f00d29d3c98e3259b410704ba0090553a47b510456871c7c2R30

siddharthkp:
> We increase the size of the clickable area for pointer: coarse

TIL!

src/_TextInputInnerAction.tsx
```diff
@@ -5,18 +5,38 @@ import {ButtonProps} from './Button'
...
 
 type TextInputActionProps = Omit<React.HTMLProps<HTMLButtonElement>, 'aria-label' | 'size'> & {
+  /** @deprecated Text input action buttons should only use icon buttons */
+  children?: React.ReactNode
   /** Text that appears in a tooltip. If an icon is passed, this is also used as the label used by assistive technologies. */
   ['aria-label']?: string
   /** The icon to render inside the button */
   icon?: React.FunctionComponent<IconProps>
   /**
+   * @deprecated Text input action buttons should only use the 'invisible' button variant
```
siddharthkp, line 15: Does this make it a breaking change? Would there be new warnings when someone upgrades primer/react?

How do we usually deprecate props?

_Side note: Are there any users of `TextInput.Action`? Can we get away with this in patch_

src/Button/IconButton.tsx
```diff
@@ -12,6 +12,10 @@ const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>((props, forwar
     getBaseStyles(theme),
     getSizeStyles(size, variant, true),
     getVariantStyles(variant, theme),
+    // when `size !== 'medium'`, vertical alignment of the icon is thrown off
+    // because changing the font size draws an em-box that does not match the
+    // bounding box of the SVG
+    {fontSize: 1},
```
siddharthkp, line 18: Should we move this to `getSizeStyles` where fontSize is set?

https://github.com/primer/react/blob/4fd9b74a38f313f829e5d201e9b0d782c5cf7b80/src/Button/styles.ts#L213-L237

.changeset/tall-dancers-exercise.md
```diff
@@ -0,0 +1,8 @@
+---
+'@primer/react': patch
+---
+
+- Inner action's hover bg should not touch the text input edges
+- Increases touch target area of the inner action button
+- Deprecation: we only want to allow inner action buttons to be icon buttons
+- Deprecation: we only want to allow inner action buttons to be the 'invisible' variant
```
siddharthkp, line 8: Should we rephrase to look more like a change log? 

"we only want to allow inner action buttons to be icon buttons" sounds a bit rude in a changelog 😅

siddharthkp (approved):
Looking good overall! Left a few comments on tiny details. 

Approving in advance :)

src/_TextInputInnerAction.tsx
mperrotti: > Does this make it a breaking change?

No, if they're still using `variant`, it will still work

---

> Would there be new warnings when someone upgrades primer/react?

I _think_ it would just be in their IDE

---

> How do we usually deprecate props?

I don't think we have a defined process yet. Maybe @colebemis knows

---

> Side note: Are there any users of TextInput.Action? Can we get away with this in patch

I don't think so. Even if there were, I think we can get away with this in a patch.

src/Button/IconButton.tsx
mperrotti: Good call!

.changeset/tall-dancers-exercise.md
mperrotti: Well that's no good. Changing!

PR #2038, opened 2022-04-20
Select component a11y fixes

mperrotti:
These changes address the feedback from @alliethu in [this comment](https://github.com/github/primer/issues/751#issuecomment-1072733921) (repo only visible internally)

Closes https://github.com/github/primer/issues/540

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


mperrotti:
> The spacing between the label and input should be consistent with the rest of the FormControl components.

✅ Fixed. The spacing is correct when using [FormControl](https://primer.style/react/FormControl). I've updated Storybook examples to use FormControl.

---

> In certain high-contrast modes on windows, it can be difficult and sometimes nearly impossible to see the select arrows.

❓ Need to test in Windows. I think I have a solution, but I need to test it on Windows. If that doesn't work, we'll just stick with the browser-native dropdown arrow.

---

> In order to more visually convey the control is not enabled, I would suggest updating the color of the select arrows to match the selectInput color value, primer-fg-disabled.

✅ Fixed

---

> I noticed you are supplementing disabled and required attributes with: `aria-disabled`, `aria-required`, and `aria-invalid`...the normal attributes should do the job just fine on their own

✅ Fixed. I got guidance from our internal accessibility Slack channel in [this thread](https://github.slack.com/archives/C0FSWLQ0Y/p1650486844547309). `aria-invalid` did not need to be removed.

---

> I can't think of a reason for not providing a visual cue for required inputs.

✅ Fixed. This is provided when using [FormControl](https://primer.style/react/FormControl). I'll update Storybook examples to use FormControl.

---

> The color contrast for placeholder text and the SelectInput border should be adjusted for the following three color modes: dark, dark dimmed, dark colorblind

⚠️ I did not change the placeholder text. We require a `<label />`, so does a placeholder need to pass color contrast? A placeholder is just a "nice to have", and the form control is still usable without it.

⚠️ I did not change the border because we're going to do that with all affected controls as part of [this issue](https://github.com/github/primer/issues/836)

---

> I noticed the placeholder content is being treated differently across browsers when the menu is open. I'm not sure if this is a result of native use, or if this is a deliberate decision on your end. When I look at how other system's <select>s render across different browsers, like Aviva's <select> , the placeholder placement is consistent.

⚠️ Did not change current behavior. When a select input is required to have a value, we intentionally disable and hide the placeholder `<option>` to prevent it from being selected. Safari doesn't allow `<option>`s to be hidden, so this is a result of native use.

---

> I'm not sure if this is a true issue, but I noticed when <optgroup> is disabled, it's being rendered as a null value (disabled="" ) vs. being disabled. MDN docs for reference.

⚠️ Did not change current behavior. This isn't an issue, and it only happens in Firefox. I think it's something React is doing, not PRC.

---

> The menu options under the disabled group still look enabled on Firefox.

✅ Fixed

---

> The disabled group label in Windows (all color modes) look enabled.

❓ I need to test in Windows to see if we have control over this, or if it's an OS-native behavior that we can't change.

src/Select.tsx
```diff
@@ -10,20 +10,38 @@ export type SelectProps = Omit<
...
+  */
+  background-color: inherit;
+
+  /* 3. Prevents visible overlap of partially transparent background colors.
+  
+     'colors.input.disabledBg' happens to be partially transparent in light mode, so we use a
+     transparent background-color on a disabled <select>. */
+  &:disabled {
+    background-color: transparent;
+  }
```
mperrotti, line 44: Help wanted: I don't like this solution, but I've yet to come up with anything better.

lindseywild:
Thanks for doing this, exciting work! I noticed a few things, I am not sure if the `id` one is related to this work but I did want to call it out.

1. In the "With a placeholder" example, when selecting an item, it seems that it has the text color is still the dimmed placeholder color (this was tested in Chrome and Safari)
screen shot of the select example with placeholder showing that the color of the selected item is dimmed like the placeholder styling
 
2. Maybe it's just in the example documentation, but the `<label for="id" ...>` doesn't match the `id` of the `<select>` element:
screen shot of the select field with dev tools expanded, highlight around the `for` attribute and `id` attribute not matching

mperrotti:
> In the "With a placeholder" example, when selecting an item, it seems that it has the text color is still the dimmed placeholder color (this was tested in Chrome and Safari)

Yup, I think I broke this while addressing the feedback about removing `aria-required` in favor of `required`. We may not have an accessible way to have placeholder text that's a different color than the other options, but I'll try.

---

> Maybe it's just in the example documentation, but the <label for="id" ...> doesn't match the id of the <select> element

Good catch! This _is_ a documentation bug, and I'll fix it

lindseywild (approved):
This looks good to me! I don't know if @alliethu wanted to take another look since she called out the issues originally? Thank you for this!

pksjce (approved):

PR #2039, opened 2022-04-21
Create ThemeDecorator for stories

siddharthkp:
Useful for decorating storybook stories, used in https://github.com/primer/react-scripts/blob/main/storybook/preview.js,

Not included in public API

- Part of https://github.com/github/primer/issues/774

### DIFF
--- a/src/ThemeDecorator.tsx
+++ b/src/ThemeDecorator.tsx
@@ -1,0 +1,13 @@
+import React from 'react'
+import {ThemeProvider} from './ThemeProvider'
+import BaseStyles from './BaseStyles'
+
+export const ThemeDecorator = (Component: React.FC) => {
+  return (
+    <ThemeProvider>
+      <BaseStyles>
+        <Component />
+      </BaseStyles>
+    </ThemeProvider>
+  )
+}


src/ThemeDecorator.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import {ThemeProvider} from './ThemeProvider'
+import BaseStyles from './BaseStyles'
+
+export const ThemeDecorator = (Component: React.FC) => {
```
colebemis, line 5: Could you add a comment here explaining what this component is and how it should be used?

siddharthkp:
I'm going to bring this back slightly better, brb 😅

src/ThemeDecorator.tsx
siddharthkp: This isn't the right location for this (closed PR, will open new one soon), but the idea was to export what we have in [.storybook/preview.js](https://github.com/primer/react/blob/main/.storybook/preview.js#L116) from primer/react so that it can be used to decorate stories with [primer/react-scripts](https://github.com/primer/react-scripts)

PR #2044, opened 2022-04-22
Move theme decorator into helper util

siddharthkp:
Logistic changes:

Moved the decorator logic to `src/utils` so that it can be used by other packages like [primer/react-scripts](https://github.com/primer/react-scripts)

Visual changes:

| Before | After |
| --- | --- |
| Pick a `color mode` & `day scheme` + `night scheme` | Pick a `color scheme` |
| before | after |
| | |
| `All` shows the day scheme and night scheme  | `All` shows all color schemes |
| all, before | all, after |


.storybook/main.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
   addons: [
     '@storybook/addon-a11y',
     '@storybook/addon-links',
-    '@storybook/addon-essentials',
+    {name: '@storybook/addon-essentials', options: {backgrounds: false}},
```
siddharthkp, line 6: This removes the background switcher which isn't useful for us:

image

src/stories/ThemeProvider.stories.tsx
```diff
@@ -8,6 +8,7 @@ import {createGlobalStyle} from 'styled-components'
 export default {
   title: 'Generic behaviors/ThemeProvider',
   component: ThemeProvider,
+  parameters: {disableThemeDecorator: true},
```
mperrotti, line 11: I'm still seeing the the theme switcher in toolbar, but it doesn't do anything. Not a blocker, but it feels a bit strange.

Screen Shot 2022-04-28 at 4 36 43 PM

mperrotti (approved):

src/stories/ThemeProvider.stories.tsx
siddharthkp: 👀 damn, that's bad. There doesn't seem to be a way for customising the toolbar without writing a custom addon 😓

colebemis (approved):
💖

PR #2046, opened 2022-04-25
Use primer/react-scripts toolchain

siddharthkp:
Part of the [toolchain for component development](https://github.com/github/primer/issues/774#toolchain)

puppers eating food

_Before we start asking other teams to start using the toolchain, 
we should use it in internal repos to test the experience_

&nbsp;

Replaces local storybook setup with [primer/react-scripts](https://github.com/primer/react-scripts). Still allows for project specific customisation. See [`.storybook/main.js`](https://github.com/primer/react/blob/siddharth/primer-scripts/.storybook/main.js) and [`.storybook/preview.js`](https://github.com/primer/react/blob/siddharth/primer-scripts/.storybook/preview.js)

~The canary version is intentional, [@primer/react-scripts](https://github.com/primer/react-scripts) is still a proof of concept and does not publish a stable version yet. The package name and script name might change.~ Now using stable version :)

Similar PR in memex: https://github.com/github/memex/pull/9638


src/Button/Button.stories.tsx
```diff
@@ -2,23 +2,10 @@ import {EyeClosedIcon, EyeIcon, SearchIcon, TriangleDownIcon, XIcon} from '@prim
...
-    Story => {
-      return (
-        <ThemeProvider>
-          <BaseStyles>
-            <Story />
-          </BaseStyles>
-        </ThemeProvider>
-      )
-    }
-  ],
```
siddharthkp, line 21: We do this in every story, which creates nested ThemeProviders! Switching theme at the top level doesn't go all the way down :/

Needs some clean up after this PR is merged.

src/utils/story-helpers.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react'
```
siddharthkp, line 1: New file :)

To review this file alone, head over to the other PR: https://github.com/primer/react/pull/2044

siddharthkp:
Circular dependency problem:

`@primer/react-scripts` has a peer dependency on `@primer/react`, so when we import something from `@primer/react` in this repo (primer/react) it wouldn't exist because we don't want to install another copy of primer/react 😅 

```
ModuleNotFoundError: Module not found: Error: Can't resolve '@primer/react/lib-esm/utils/story-helpers' in '/Users/sid/code/primer/react/node_modules/@primer/react-scripts/storybook'
```

colebemis (approved):
:shipit:

siddharthkp:
@colebemis Do you know why CI is failing on this, looks like an unrelated change but it doens't fail on main 🤔

Could be because of a potential eslint version change?

colebemis:
> @colebemis Do you know why CI is failing on this, looks like an unrelated change but it doens't fail on main 🤔
> 
> Could be because of a potential eslint version change?

Huh, I wonder if it's related to these changes: https://github.com/primer/react/pull/2083

siddharthkp:
Decided to use `--legacy-peer-deps` for docs until https://github.com/primer/doctocat/pull/454 is merged, will create another PR after doctocat release :)

Issue #2045, opened 2022-04-22
Theme provider is not respecting `colorMode="auto"` 


jfuchs:
**Describe the bug**

**To Reproduce**
1. Go to https://codesandbox.io/s/clever-volhard-w8csh2?file=/src/App.tsx
2. Toggle your system color mode
3. Note button not changing

**Expected behavior**
Button changes color mode

**Screenshots**
![CleanShot 2022-04-22 at 11 41 06@2x](https://user-images.githubusercontent.com/21195/164774964-03d73943-efe9-41d0-93c3-ed737df978b3.png)
![CleanShot 2022-04-22 at 11 42 23@2x](https://user-images.githubusercontent.com/21195/164775158-44b3e88a-39b7-49cc-8ce2-28e1896cf8bf.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome latestish

**Additional context**
- I believe the issue was introduced in https://github.com/primer/react/pull/1936
- Reloading the page will give the correct auto color mode initially

siddharthkp:
Can confirm! The theme is correct on first render based on system settings but isn't responding to changes to system settings after that


https://user-images.githubusercontent.com/1863771/165057963-147ebdbb-ebe2-4d27-b2a1-6766142accc1.mov

PR #2048, opened 2022-04-26
ThemeProvider: Fix theme sync with system settings

siddharthkp:
- Fix https://github.com/primer/react/issues/2045
- Dependency: https://github.com/primer/behaviors/pull/93

Bug: ThemeProvider with `colorMode=auto` only respects system settings on first load. It does not sync with system settings after that.

Manually verified that this works with server side rendering as well. There isn't an easy way to add tests for this because it's a system setting :(

before:

https://user-images.githubusercontent.com/1863771/165322367-37ed0ef5-497c-4473-9c21-cb4718d3cd56.mov

after:

https://user-images.githubusercontent.com/1863771/165322314-e2949f80-e9f3-4d6b-b694-0c9185b010e2.mov


.storybook/preview.js
```diff
@@ -71,6 +71,9 @@ export const globalTypes = {
 
 // context.globals.X references items in globalTypes
 const withThemeProvider = (Story, context) => {
+  // used for testing ThemeProvider.stories.tsx
+  if (context.parameters.disableThemeDecorator) return <Story {...context} />
```
siddharthkp, line 75: Had to disable the theme decorator so that I can actually test ThemeProvider 😅

src/ThemeProvider.tsx
```diff
@@ -101,10 +101,6 @@ export const ThemeProvider: React.FC<ThemeProviderProps> = ({children, ...props}
     setColorMode(props.colorMode ?? fallbackColorMode ?? defaultColorMode)
   }, [props.colorMode, fallbackColorMode])
 
-  React.useEffect(() => {
-    setColorMode(resolvedColorMode)
-  }, [resolvedColorMode])
-
```
siddharthkp, line 107: I split this effect in https://github.com/primer/react/pull/1936, now this effect was just syncing state to colorMode 🤔  which is a bug

mperrotti: I don't think I understand this bug. Everything seems to be working as expected 🤔

mperrotti (approved):

src/ThemeProvider.tsx
siddharthkp: The bug was that changes in system settings did not change the resolved theme in ThemeProvider

repro:

https://user-images.githubusercontent.com/1863771/165322367-37ed0ef5-497c-4473-9c21-cb4718d3cd56.mov

PR #2053, opened 2022-05-02
Namespace UnderlineNav

langermank:
PRC and PCSS are conflicting since these classes aren't hashed. I believe UnderlineNav is being refactored right now, so I feel like this temporary solution could help unblock memex?

I simply namespaced UnderlineNav classes with "PRC"

### Screenshots

This is in prod:

memex board navigation

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


colebemis (approved):
👍 it looks like we need to update snapshots (`npm run test:update`) and add a changeset

PR #2063, opened 2022-05-10
Update issue templates for this repo to include redirects to pattern …

lesliecdubs:
This PR updates the issue templates for this repo that live at https://github.com/primer/react/issues/new/choose as follows:
-  Removed the "Component Proposal" template, as we prefer proposals to start at the design layer for holistic consideration of a pattern before prioritizing implementation in Primer React or Primer ViewComponents. GitHub staff can provide more context in a pattern proposal issue about which framework they need the component for.
- Added links to the Issues template page which redirect folks to propose new component ideas in an internal Discussion, or request new patterns via an internal pattern proposal issue.


yaili (approved):

colebemis (approved):
✨

PR #2076, opened 2022-05-16
NavList.Item: Add `as` prop

colebemis:
Adds support for the `as` prop on `NavList.Item`. This will allow consumers to use `NavList` with router-specific link components (e.g., NextJS, React Router)

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


colebemis:
The current implementation of `NavList.Item` uses [`React.Children` to check if the current item is nested inside a `NavList.SubNav`](https://github.com/primer/react/blob/main/src/NavList/NavList.tsx#L52). It uses this information to change the styles and default open state of the parent `NavList.SubNav`. This implementation relies on `NavList.Item` being a direct child of `NavList.SubNav`. 

With the introduction of the `as` prop to `NavList.Item`, consumers are likely to wrap `NavList.Item` in a component that integrates with their router. Example:

```jsx
function NavLink({to, children}) {
  const isCurrent = ... // Some router-specific logic
  return <NavLink.Item as={RouterLink} to={to} aria-current={isCurrent ? 'page' : false}>{children}</NavLink.Item>
}
```

However, because we rely on `NavList.Item` being a direct child of `NavList.SubNav`, wrapping `NavList.Item` like this breaks the `NavList.SubNav` behavior, as demonstrated by this [failing test](https://github.com/primer/react/runs/6460340796?check_suite_focus=true#step:7:1069).

**Open question: How can we check if a `SubNav` contains a current `Item` without relying on `Item` being a direct child of `SubNav`? Is it possible to implement it in an SSR-compatiable way?**

siddharthkp:
> How can we check if a SubNav contains a current Item without relying on Item being a direct child of SubNav? Is it possible to implement it in an SSR-compatible way?

Doesn't look like there's a way without requiring input from the user. Explored the implementation here: https://github.com/primer/react/pull/2079

```jsx
<NavList.SubNav>
 <CustomNavLink href="/4-1">
   On first render, we have no idea of knowing if there's a
   aria-current inside CustomNavLink
 </CustomNavLink>
 <CustomNavLink href="/4-2" aria-current={pathname === '/4-1'}>
   We can ask the developer to pass it on the direct child of SubNav
   even if it isn't NavList.Item, feels awkward though
 </CustomNavLink>
</NavList.SubNav>
```

Ways to make this work:
- Add a "Extending NavList.Item" section to the docs
- Add warning in dev mode which throws a warning in dev environment if `aria-current` isn't found on mount

colebemis:
Hmm, after doing more research into router APIs, I don't think requiring `aria-current` to be passed to a direct child of `SubNav` is a viable API :( 

For example, Gatsby Link (which uses Reach Router under-the-hood) exposes a [`getProps`](https://reach.tech/router/api/Link) prop to give access to the `isCurrent` state of the link. 

Developers trying to use NavList with Gatsby would need to do something like this:

```jsx
<NavList>
  <NavList.Item>
    Item
    <NavList.SubNav>
      <NavList.Item as={Link} to="#" getProps={({isCurrent}) => ({'aria-current': isCurrent})}>Sub-item</NavList.Item>
    </NavList.SubNav>
  </NavList.Item>
</NavList>
```

In this example, it would be hard for us to determine if a child of `SubNav` is current using `React.Children` because `aria-current` isn't passed to `NavList.Item` directly.

colebemis:
Using DOM APIs + `useLayoutEffect` might be the best path forward for now ([commit](https://github.com/primer/react/pull/2076/commits/f02f63b0d0864700c87068a767feb7287269d1c6)). Here are the pros and cons of this approach:

Pros:
* Allows us to keep the API of NavList simple
* Allows consumers to freely wrap NavList.Item with router-specific logic without breaking NavList functionality and styles

Cons:
* We cannot determine if a SubNav contains the current item on a server render. This may cause some layout jank when a page is loaded from the server and hydrated on the client. However, this only will only affect applications that use SSR on pages that are nested in a sub-nav which seems like an edge case, especially because no GitHub React apps use SSR yet.

siddharthkp:
> We cannot determine if a SubNav contains the current item on a server render. This may cause some layout jank when a page is loaded from the server and hydrated on the client. However, this only will only affect applications that use SSR on pages that are nested in a sub-nav which seems like an edge case, especially because no GitHub React apps use SSR yet.

I don't love it, but I can get behind solving this in the future instead of now.

Alternate solution: How do you feel about adding an explicit `open` prop on SubNav and passing the responsibility down to the developer?

colebemis:
During our engineering sync today, we decided to move forward with the DOM API + useLayoutEffect implementation for now ✅

PR #2077, opened 2022-05-16
NavList: Add `sx` prop

colebemis:
Adds support for the `sx` prop on `NavList` and every subcomponent of `NavList` (e.g. `NavList.Item`).

### DIFF
--- a/src/NavList/NavList.tsx
+++ b/src/NavList/NavList.tsx
@@ -38,7 +38,7 @@ export type NavListItemProps = {
 
 // TODO: as prop
 const Item = React.forwardRef<HTMLAnchorElement, NavListItemProps>(
-  ({href, 'aria-current': ariaCurrent, children, sx: sxProp}, ref) => {
+  ({href, 'aria-current': ariaCurrent, children, sx: sxProp = {}}, ref) => {
     const {depth} = React.useContext(SubNavContext)
 
     // Get SubNav from children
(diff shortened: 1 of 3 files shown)


src/NavList/NavList.tsx
```diff
@@ -1,23 +1,27 @@
...
 // ----------------------------------------------------------------------------
 // NavList
 
 export type NavListProps = {
   children: React.ReactNode
-} & React.ComponentProps<'nav'>
+} & SxProp &
+  React.ComponentProps<'nav'>
+
+const NavBox = styled.nav<SxProp>(sx)
```
mperrotti, line 18: Does this go against ADR 005? https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-005-box-sx.md

cc @siddharthkp

mperrotti (approved):
Approving on the condition that we're not violating [ADR 005](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-005-box-sx.md)

src/NavList/NavList.tsx
colebemis: I copied this from `ActionList` ([code](https://github.com/primer/react/blob/main/src/ActionList/List.tsx#L30)) as a workaround for a ref type error I ran into.

```jsx
// TypeScript doesn't like this
<Box as="nav" ref={ref} sx={{...}}>

// But this works
const NavBox = styled.nav(sx)
<NavBox ref={ref} sx={{...}}>
```

If there's another way to workaround that type issue, I'm open to updating this!

mperrotti: I've also struggled with this type issue, and I'm not sure how to fix it 😅 

Let's leave this overnight and see if any of our coworkers in other timezones know a solution.

siddharthkp: > I copied this from ActionList ([code](https://github.com/primer/react/blob/main/src/ActionList/List.tsx#L30)) as a workaround for a ref type error I ran into.

Oh yes, I remember this! I don't have a better solution for this either :( 

I guess we should update the ADR with a note for components that accept a ref?

colebemis: I'm going to move forward with this for now. A long-term solution might be to move away from the `as` prop entirely 😅

PR #2064, opened 2022-05-11
Implement NavList.SubNav

colebemis:
https://user-images.githubusercontent.com/4608155/168161016-bb1b8d9b-a2e1-4902-85aa-10470ba75426.mp4

* Implements `NavList.SubNav` component
* Changes behavior of `NavList.Item` when `NavList.SubNav` is nested inside it
* Adds tests for `NavList.Item` and `NavList.SubNav`
* Adds `active` prop to `ActionList.Item` so active styles can be shared among all `ActionList`-related components

## Checklist
- [x] Updated documentation
- [x] Wrote tests
- [x] Tested in Safari
- [x] Tested in Chrome
- [x] Tested in Firefox

PR #2080, opened 2022-05-17
ActionMenu: adjust width based on padding

siddharthkp:
Changed in https://github.com/primer/react/pull/2064, skipping changeset because we haven't published this change yet.

before:
image

after: 
image


colebemis (approved):
Nice catch!

Issue #2070, opened 2022-05-12
Docs use unreleased release candidate instead of published version


siddharthkp:
**Describe the bug**
Our docs deploy `main` branch which isn't the same as the latest published release from npm. Which means, when looking at the docs, you can't be sure that the props and examples will actually work in your application even when you're on the latest version.

In practice, [primer.style/react](https://primer.style/react) points at the [latest release candidate](https://github.com/primer/react/pull/2056), not at the [latest published version](https://github.com/primer/react/releases). We usually release a new path/minor version every couple of weeks, so the chances of you running in this are not low.

Related issue: There is no easy way to see docs for an old version of primer/react that your application might be using.

**To Reproduce**
1. Go to [Tooltip examples](https://primer.style/react/Tooltip#examples)
2. Copy the example into your application
3. It wouldn't work as documented because the latest Tooltip changes are in a [release candidate](https://github.com/primer/react/pull/2056) and not published yet.

**Expected behavior**

1. [primer.style/react](https://primer.style/react) should point at the latest published version, not the version from `main`.
2. Bonus: We should introduce a version switcher on the docs which can take you the published version for older versions (and the latest release candidate)

PR #2091, opened 2022-05-23
Uncomment docs deployment guard

siddharthkp:
Fixes https://github.com/primer/react/issues/2070

We already had a deployment guard based on changesets, but it was commented out 😅 

Tested with and without changesets in https://github.com/primer/react/pull/2089

image

---

Is there a way to mention a reason why the job was skipped? Update: There isn't without creating an additional check, which leads to duplicate checks, not ideal either.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -35,7 +35,7 @@ jobs:
   deploy:
     name: Production
     needs: [guard]
-    #if: ${{ needs.guard.outputs.should_deploy == true }}
+    if: ${{ needs.guard.outputs.should_deploy == 'true' }}
     uses: primer/.github/.github/workflows/deploy.yml@main
     with:
       node_version: 16


.github/workflows/deploy_production.yml
```diff
@@ -35,7 +35,7 @@ jobs:
   deploy:
     name: Production
     needs: [guard]
-    #if: ${{ needs.guard.outputs.should_deploy == true }}
+    if: ${{ needs.guard.outputs.should_deploy == 'true' }}
```
siddharthkp, line 38: @colebemis Turns out the culprit was string comparison :)

colebemis (approved):
😅 😅 😅

siddharthkp:
> 😅 😅 😅

Now, let's hope it actually works on `main` too 😅 I've tested it on a branch till now.

siddharthkp:
It works as intended on main with and without changesets :)

image

PR #2095, opened 2022-05-24
Set aria-multiselectable in SelectPanel

hectahertz:
Part of the `SelectPanel` changes that will make based on [the A11y eng review](https://github.com/github/primer/issues/659).

This small PR communicates the multi-select capability to assistive technologies by setting `aria-multiselectable="true"` on the listbox element.

### Screenshots

From the Chrome accessibility tree

## Before

Multi select
Screen Shot 2022-05-24 at 15 04 36

## After

Multi select
Screen Shot 2022-05-24 at 15 00 18

Single select
Screen Shot 2022-05-24 at 15 01 25


siddharthkp (approved):
Works like a charm! :shipit:

PR #2101, opened 2022-05-25
ActionMenu: Inline label

siddharthkp:
- Based on https://github.com/github/primer/discussions/696
- Fixes https://github.com/github/primer/issues/994
- Fixes https://github.com/github/primer/issues/995
- Fixes https://github.com/github/primer/issues/993

.
- [See new section about accessibility in docs →](https://primer-01ca87129d-13348165.drafts.github.io/ActionMenu#accessibility)

---

Open questions:

1. Is the prop `label` too similar to `aria-label`, can that cause confusion? Should we call it `inlineLabel` instead to add some weight?
2. Is there something we can do to simplify the a11y needs with an external label as well? [Right now, we defer this responsibility to the user](https://primer-01ca87129d-13348165.drafts.github.io/ActionMenu#:~:text=If%20you%20have%20an%20external%20label%2C). For example, we could split the `label` prop into `label` & `showLabel`. `label` would be used to create `aria-label` on button and `aria-labelledby` for action list. it will only be shown visually if `showLabel` or `withInlineLabel` is passed.

---

### Explanation

&nbsp;

Goal:

For menus with selection, it's common to show the selected value inside the button. To give context, it's important to show purpose of the menu as well. Visually, this can be an external label above the button or an inline label inside the button (new!)

In both cases, the label for screen readers (with aria-label or aria-labelledby) should have both purpose and value, example: "Field type: Number".

We want to add the option to add an inline number (left) as an alternative to [external label](https://primer-01ca87129d-13348165.drafts.github.io/storybook/?path=/story/composite-components-actionmenu-examples--single-selection) (right)

image

a11y tree:
```
► combobox "Field type: Number"
► menu "Field type"
```

&nbsp;

To achieve this today, users have to write something like this:

```jsx
<ActionMenu>
  <ActionMenu.Button 
    aria-label={'Field type: ' + selectedOption.name}
    leadingIcon={selectedOption.value}
  >
    <Text sx={{ color: 'fg.muted' }}>Field type:</Text>
    {selectedOption.name}
  </ActionMenu.Button>
  <ActionList
    selectionVariant="single"
    aria-label="Field type"
   >
    ...
   </ActionList>
</ActionMenu>
```

Suggested API in this pull request: 

1. use `label` instead of `aria-label` on `ActionMenu.Button`
2. labels for button and menu are created automatically

```diff
<ActionMenu>
  <ActionMenu.Button 
-   aria-label={'Field type: ' + selectedOption.name}
+   label="Field type"
    leadingIcon={selectedOption.value}
  >
-   <Text sx={{ color: 'fg.muted' }}>Field type:</Text>
    {selectedOption.name}
  </ActionMenu.Button>
  <ActionList
    selectionVariant="single"
-   aria-label="Field type"
   >
    ...
   </ActionList>
</ActionMenu>
```

html output:

```html
<button aria-labelledby="purpose divider value"> 
  <span id="purpose">Field type</span><span id="divider">:</span> 
  <svg aria-hidden="true"></svg>
  <span id="value">Iteration</span>
<button>
<div role="menu" aria-labelledby="purpose"></div>
```

a11y tree:
```
► combobox "Field type : Number"
► menu "Field type"
```

### DIFF
--- a/.changeset/odd-bobcats-cheer.md
+++ b/.changeset/odd-bobcats-cheer.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": minor
+---
+
+ActionMenu: Option to add inline label for accessible menus. ([Read more](https://primer.react/ActionMenu#accessibility))
--- a/docs/content/ActionMenu.mdx
+++ b/docs/content/ActionMenu.mdx
@@ -289,11 +289,13 @@ render(
 
 ### Single Selection
 
-It is recommended to have a visual label for `ActionMenu.Button`. This can either be an external label or an inline label.
+For menus with selection, it's common to show the selected value in the button. To give context to the user, it's important to show the purpose of the menu as well.
 
-If you have an external label, it's important to add an `aria-label` to `ActionMenu.Button` which contains both the purpose and current selected value, example: _"Field type: Number"_. In addition to that, you must add `aria-label` to the `ActionList` with the purpose of the menu, example: _"Field type"_.
+Visually, this can be an external label above the button or an inline label inside the button. In both cases, the label for screen readers should have both purpose and value, example: "Field type: Number".
 
-Alternatively, you can add an inline label by adding `label`, example: _"Field type"_ on `ActionMenu.Button` instead of `aria-label` which will add `aria-label` on the button and `aria-labelledby` on the menu for you.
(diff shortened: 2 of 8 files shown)


colebemis (approved):

siddharthkp:
happy-dusting

lesliecdubs:
👋🏻 @siddharthkp anything I can do to help get this one over the line? Were you waiting on an accessibility specialist to review?

PR #2102, opened 2022-05-25
SelectPanel: Add title prop for a11y purposes

hectahertz:
Part of the `SelectPanel` changes that we will make based on [the A11y eng review](https://github.com/github/primer/issues/659).

This PR adds a **required** `title` prop to be both used as the overlay title and on `aria-labelledby` on the listbox element.

It is also rendered visually, so it matches the new SelectPanel designs.

## Before
Screen Shot 2022-05-25 at 17 49 37
Screen Shot 2022-05-25 at 17 52 59

## After
Screen Shot 2022-05-25 at 17 49 04
Screen Shot 2022-05-25 at 17 53 54


.changeset/rare-falcons-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': major
```
hectahertz, line 2: This is a breaking change as the prop is required; please do let me know if a different release strategy would be better.

hectahertz: Only the `title` prop being required for now

siddharthkp (approved):
Overall looks good! Left a tiny comment about `styled.h1`

The other part we might want to consider is if how do we want to merge this PR:
A. merge this to next-major with a breaking change 
B. split it into 2 PRs, one without breaking change in the next-minor and one making title required in the next-major

_Sidenote: You might have to "Update branch" with master to make CI green again_

PR #2117, opened 2022-06-06
Allow minor version updates for prod dependencies

dgreif:
When using PRC + PVC + Primer CSS (eg in github/github), it's difficult to keep the dependency versions perfectly in-line.  We should allow _minor_ version differences for prod dependencies to account for discrepancies.


colebemis (approved):
🚢

PR #2123, opened 2022-06-14
Fix CI for Node 16.15.1

siddharthkp:
Something changed between 16.15.0 (npm 8.5.5, [see CI action logs](https://github.com/primer/react/runs/6803999761?check_suite_focus=true)) and 16.15.1 (npm 8.11.0, [see CI action logs](https://github.com/primer/react/runs/6821886462?check_suite_focus=true))

- Upgraded local node to 16.15.1 (`nvm install --lts`)
- ran `npm install`
- committed changed `package-lock.json`

Doing the same didn't fix docs & preview deployment, so as a hot fix, changed the node version to Node 14 (npm 6.14.17), 

We can swap it out with Node 16 after fixing other issues.

---

@colebemis looks like doctocat expects `gatsby: 2.x` as [peer dependency](https://github.com/primer/doctocat/blob/main/theme/package.json#L26) 🤔 

<details>
<summary>install logs</summary>

```
● docs:node-16.15.1 $ npm install

npm WARN config init.license Use `--init-license` instead.
npm WARN config init.author.name Use `--init-author-name` instead.
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: @primer/gatsby-theme-doctocat@4.0.0
npm ERR! Found: gatsby@3.7.2
npm ERR! node_modules/gatsby
npm ERR!   gatsby@"^3.7.2" from the root project
npm ERR!   peer gatsby@">2.0.0" from gatsby-plugin-alias-imports@1.0.5
npm ERR!   node_modules/gatsby-plugin-alias-imports
npm ERR!     gatsby-plugin-alias-imports@"^1.0.5" from the root project
npm ERR!   4 more (babel-plugin-remove-graphql-queries, ...)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer gatsby@"2.x" from @primer/gatsby-theme-doctocat@4.0.0
npm ERR! node_modules/@primer/gatsby-theme-doctocat
npm ERR!   @primer/gatsby-theme-doctocat@"^4.0.0" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: gatsby@2.32.13
npm ERR! node_modules/gatsby
npm ERR!   peer gatsby@"2.x" from @primer/gatsby-theme-doctocat@4.0.0
npm ERR!   node_modules/@primer/gatsby-theme-doctocat
[...]

PR #2125, opened 2022-06-15
Enforce correct semantics for TabNav

owenniblock:
Enforces the containing `div` to have `role="tablist"` and the item elements to have `role="tab"` - tabs should avoid having focusable elements as children.

### Screenshots

Before:
Before screenshot showing 3 tabs

After:
After screenshot showing 3 tabs that are styled the same

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/TabNav.tsx
+++ b/src/TabNav.tsx
@@ -10,25 +10,27 @@ import getGlobalFocusStyles from './_getGlobalFocusStyles'
 const ITEM_CLASS = 'TabNav-item'
 const SELECTED_CLASS = 'selected'
 
-const TabNavBase = styled.div<SxProp>`
-  margin-top: 0;
-  border-bottom: 1px solid ${get('colors.border.default')};
-  ${sx}
-`
-
-const TabNavBody = styled.nav`
+const TabNavTabList = styled.div<SxProp>`
   display: flex;
   margin-bottom: -1px;
   overflow: auto;
 `
 
-export type TabNavProps = ComponentProps<typeof TabNavBase>
+const TabNavBody = styled.nav`
+  margin-top: 0;
+  border-bottom: 1px solid ${get('colors.border.default')};
+  ${sx}
+`
+
+export type TabNavProps = ComponentProps<typeof TabNavTabList>
 
 function TabNav({children, 'aria-label': ariaLabel, ...rest}: TabNavProps) {
   return (
-    <TabNavBase {...rest}>
-      <TabNavBody aria-label={ariaLabel}>{children}</TabNavBody>
-    </TabNavBase>
+    <TabNavBody aria-label={ariaLabel}>
+      <TabNavTabList role="tablist" {...rest}>
+        {children}
+      </TabNavTabList>
+    </TabNavBody>
   )
 }
 
@@ -39,7 +41,8 @@ type StyledTabNavLinkProps = {
 
 const TabNavLink = styled.a.attrs<StyledTabNavLinkProps>(props => ({
   activeClassName: typeof props.to === 'string' ? 'selected' : '',
-  className: classnames(ITEM_CLASS, props.selected && SELECTED_CLASS, props.className)
+  className: classnames(ITEM_CLASS, props.selected && SELECTED_CLASS, props.className),
+  role: 'tab'
 }))<StyledTabNavLinkProps>`
   padding: 8px 12px;
   font-size: ${get('fontSizes.1')};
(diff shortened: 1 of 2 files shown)


siddharthkp:
Hi! Just to let you know that CI/deployments breaking is not because of the changes in this PR and is something on our radar! https://github.com/primer/react/pull/2123

owenniblock:
Was chatting with Joel last night and now have a better handle on what `As` does. Maybe we *don't* want to support `As` here and we should have the top level HTML starting with the `<nav>`.

I also wonder whether we really want to support `sx` functionality (I see those tests disabled on a lot of other components).

As these are breaking changes, I wanted to get some feedback before pushing up a change in that direction.

siddharthkp:
Hi! I haven't had the chance to look at this PR yet, I will come back to it with opinions/suggestions this week :)

tl;dr: 
- we want to support sx on every component
- we try to avoid breaking changes, but if there's a good case for them, we can schedule it for the next-major

owenniblock:
Thanks @siddharthkp - my opinions below 🙂 

> it might be okay to skip the as prop here if we want to enforce the container is a `nav`

I think we do, it makes sense for TabNav to enforce the `nav` element to allow screen readers to navigate to the tabs in navigation mode. So if I try to support `As` - all I'm doing is wrapping it in a `div` that's not going to be required 99% of the time. Let me know if that sounds sensible and I'll update this PR to remove the `nav` and not test for `As` compatibility.

> we want to support sx on every component

No problem. I'll leave this in 🙂 

> we try to avoid breaking changes, but if there's a good case for them, we can schedule it for the next-major

No problem, I think there's no avoiding it here in any case (currently the HTML is invalid so we have to sort that out).

siddharthkp:
Hi again! This looks good! 

I don't think we need to mark this as a breaking change. However, I did find 2 instances that we might want to check with canary build before we merge:

- https://github.com/github/memex/blob/main/src/client/components/view-navigation.tsx#L99
- https://github.com/github/docs-internal/blob/main/components/article/PlatformPicker.tsx#L163

owenniblock:
Thanks for your attention to detail @siddharthkp!

For this one:

> in [memex/components/view-navigation.tsx#L99](https://github.com/github/memex/blob/main/src/client/components/view-navigation.tsx#L99), <TabNav role="tablist" would get attached to the base creating duplicate role=tablist

We're adding this change to fix an issue where the roles are incorrectly surfaced (because the `tablist` can only contain `tab` elements but currently the `tab` elements are wrapped in a `nav` so it created invalid semantic HTML.

> in [docs-internal/components/article/PlatformPicker.tsx#L163](https://github.com/github/docs-internal/blob/main/components/article/PlatformPicker.tsx#L163), data-test-id is passed down to the container and used for tests. With a quick look, this seems to be fine, but just to be sure, we should run the CI suite with a canary build

Sounds like a good idea! Is there a doc I can look at for how to do this? Or is it something you can help with?

siddharthkp:
> We're adding this change to fix an issue where the roles are incorrectly surfaced (because the tablist can only contain tab elements but currently the tab elements are wrapped in a nav so it created invalid semantic HTML.

This would require a change in memex as well, right?

> Sounds like a good idea! Is there a doc I can look at for how to do this? Or is it something you can help with?

This isn't documented, I can write down the steps here as a start:

1. Grab the canary version from the pull request builds:
    image

2. Install canary version in a branch:

    ```
    npm install @primer/react@0.0.0-2022520124925
    ```

However, docs-internal is one major version behind (breaking changes), so I don't think we would be able to test this change in isolation :(

siddharthkp:
> However, docs-internal is one major version behind (breaking changes), so I don't think we would be able to test this change in CI :(

Update: I made local changes in `node_modules` to test if tests will pass with the changes from this PR, they do :)

owenniblock:
@siddharthkp I've created a draft PR here: https://github.com/github/memex/pull/10664 - I'll have to get some help to work out how best to test it!

owenniblock:
OK, I'm getting an error on the `<TabNav.Link` line on memex - let me dig in to that and try and work out what's going on there!

owenniblock:
@siddharthkp looks good with the canary build: https://github-1f17a67ba6-12535706.drafts.github.io/orgs/app/projects/1/views/1

Let me know when we should merge this (or if you think we need any further changes) 🙂

siddharthkp:
Looks good, I'll merge it :)

Issue #2132, opened 2022-06-20
NavList.Group should pass all props through to the ActionList.Group


mattcosta7:
**Describe the bug**

https://github.com/primer/react/blob/53713b2f3ab7dd7084ce3e602c01c3f66ccd7579/src/NavList/NavList.tsx#L244

(as a nit here - using `sx = {}` in the function definition means every render _by default_ supplies a new sx object.  `sx = defaultSx`, where previously `const defaultSx = {}` would avoid that

NavList.Group only accepts a handful of props from ActionList.Group, and should probably accept almost/all and pass them through, specifically for things like aria attributes and `data` props, for test ids, etc

**Note:** As this is a public repo, please be careful not to include details or screenshots from unreleased GitHub products or features. In most cases, a good bug report should be able to describe the new component without including business logic or feature details, but if you must discuss context relating to an unreleased feature, please open an issue in the private [Design Systems repo](https://github.com/github/design-systems/issues/new/choose) and link to it here.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #2133, opened 2022-06-20
Navlist passthrough action list group props

mattcosta7:
Describe your changes here.

Closes https://github.com/primer/react/issues/2132

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/NavList.mdx
+++ b/docs/content/NavList.mdx
@@ -317,6 +317,16 @@ function App() {
   <PropsTableRow name="title" type="string" description="Title of the group." /> {/* Should `title` be required? */}
   <PropsTableSxRow />
   <PropsTableRefRow refType="HTMLElement" />
+  <PropsTableBasePropRows
+    passthroughPropsLink={
+      <>
+        the <Link href="/ActionList#actionlistgroup">ActionList.Group docs</Link>
+      </>
+    }
+    elementType="a"
+    isPolymorphic
+    refType="HTMLAnchorElement"
+  />
 </PropsTable>
 
 ### NavList.Divider
--- a/docs/content/NavList.mdx
+++ b/docs/content/NavList.mdx
@@ -315,15 +315,8 @@ function App() {
 
 <PropsTable>
   <PropsTableRow name="title" type="string" description="Title of the group." /> {/* Should `title` be required? */}
+  <PropsTableSxRow />
   <PropsTableRefRow refType="HTMLElement" />
-  <PropsTableBasePropRows
-    passthroughPropsLink={
-      <>
-        the <Link href="/ActionList#actionlistgroup">ActionList.Group docs</Link>
-      </>
-    }
-    elementType="ActionList.Group"
-  />
 </PropsTable>
 
 ### NavList.Divider
--- a/docs/content/NavList.mdx
+++ b/docs/content/NavList.mdx
@@ -323,9 +323,7 @@ function App() {
         the <Link href="/ActionList#actionlistgroup">ActionList.Group docs</Link>
       </>
     }
-    elementType="a"
-    isPolymorphic
-    refType="HTMLAnchorElement"
+    elementName="ActionList.Group"
   />
 </PropsTable>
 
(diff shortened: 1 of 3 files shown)


colebemis (approved):
Thank you, @mattcosta7! 💖

src/NavList/NavList.tsx
```diff
@@ -235,18 +236,18 @@ Divider.displayName = 'NavList.Divider'
 // ----------------------------------------------------------------------------
 // NavList.Group
 
-export type NavListGroupProps = {
+export type NavListGroupProps = Omit<ActionListGroupProps, 'selectionVariant' | 'variant' | 'auxiliaryText'> & {
```
colebemis, line 239: Could you update the props documentation for NavList.Group to show that props are passed through to ActionList.Group? This should do the trick:

```
...
### NavList.Group

<PropsTable>
  ...
  <PropsTablePassthroughPropsRow
    elementName="ActionList.Group"
    passthroughPropsLink={
      <>
        the <Link href="/ActionList#actionlistgroup">ActionList.Group docs</Link>
      </>
    }
  />
</PropsTable>
```

It should render something similar to this: 
CleanShot 2022-06-20 at 11 13 56@2x

src/NavList/NavList.tsx
```diff
@@ -235,18 +236,18 @@ Divider.displayName = 'NavList.Divider'
...
   children: React.ReactNode
-  title?: string
 } & SxProp
 
+const defaultSx = {}
 // TODO: ref prop
-const Group = ({title, children, sx: sxProp = {}}: NavListGroupProps) => {
+const Group: React.VFC<NavListGroupProps> = ({title, children, sx: sxProp = defaultSx, ...props}) => {
```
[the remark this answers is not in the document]
mattcosta7: `VFC` is `FC` but `children` is not implied, which makes it much easier to transition to react 18, where `FC` doesn't have children by default - since it ensures we add them manually

src/NavList/NavList.tsx
mattcosta7: can do

docs/content/NavList.mdx
```diff
@@ -317,6 +317,14 @@ function App() {
...
   <PropsTableSxRow />
   <PropsTableRefRow refType="HTMLElement" />
+  <PropsTableBasePropRows
+    passthroughPropsLink={
+      <>
+        the <Link href="/ActionList#actionlistgroup">ActionList.Group docs</Link>
+      </>
+    }
+    elementName="ActionList.Group"
+  />
```
colebemis, line 327:
```suggestion
  <PropsTableRefRow refType="HTMLElement" />
  <PropsTableBasePropRows
    passthroughPropsLink={
      <>
        the <Link href="/ActionList#actionlistgroup">ActionList.Group docs</Link>
      </>
    }
    elementType="ActionList.Group"
  />
```

colebemis:
I just realized that the only different between the NavList.Group props and ActionList.Group is role. Would it be simpler to explicitly add the role prop instead of spreading all passthrough props? I worry a little bit that passing through all the props might accidentally expose APIs we don’t expect in the future

mattcosta7:
> > I just realized that the only different between the NavList.Group props and ActionList.Group is role. Would it be simpler to explicitly add the role prop instead of spreading all passthrough props? I worry a little bit that passing through all the props might accidentally expose APIs we don’t expect in the future
> 
> Would passing through just dom props here work better
> 
> NavListGroupProps = htmlelementattributes & {} & SxProp
> 
> That should cover the usecase of passing through things like data and aria attributes, plus things like style and class names/IDs/etc

Ended up just doing what ActionList does, which is pass through props, but not explicitly document them or accept typings.

this mostly provides a way to pass through attributes like `aria-xxx` and `data-xxx`, since these don't cause typescript to complain, but other ActionList.Group or Box props wouldn't be accepted.  this should cover the need without allowing the abstractions to leak

colebemis (approved):

PR #2138, opened 2022-06-23
SelectPanel: Accessibility remediation

bolonio:
> Co-authored-by: @hectahertz

## Context

This PR addresses the changes derived from:
- The [Accessibility design review](https://github.com/github/primer/issues/659) of [the Figma files](https://www.figma.com/file/KBfDd7lnhjPDWxmZWilh5c/Scrap-paper?node-id=516%3A41331)
- The [Accessibility engineering review](https://github.com/github/primer/issues/1021)

The [Accessibility engineering review](https://github.com/github/primer/issues/1021) has more details on the reasoning behind all changes.

## Screenshots

| Before                | After                 |
| ------------------- | -------------------- |
| Screenshot of the previous version of the component | Screenshot of the proposed layout for the component, which adds a title, close button and save/cancel buttons |

## Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/SelectPanel/SelectPanel.tsx
```diff
@@ -72,6 +100,13 @@ export function SelectPanel({
     [externalOnFilterChange, setInternalFilterValue]
   )
 
+  const selectedItems = React.useMemo(
+    () => (Array.isArray(selected) ? selected : [...(selected ? [selected] : [])]),
+    [selected]
+  )
+
+  const [finalItemsSelected, setFinalItemsSelected] = useState(selectedItems)
```
camchenry, line 108: @bolonio I think I found the issue that we were running into with the selected state and why it wasn't reproducing well in Storybook. In our app, we're passing in the selected state immediately and then asynchronously loading all of the options. But this selected state inside `SelectPanel` doesn't get reset when that happens, so it appears that nothing is initially selected. I was able to fix that by patching our `@primer/react` built file to have this line:

```suggestion
  const [finalItemsSelected, setFinalItemsSelected] = useState(selectedItems)
  React.useEffect(() => setFinalItemsSelected(selectedItems), [selectedItems])
```

I think partially why our code worked before was because there was no `useState` here, so the state was automatically reset on every render, which is not true anymore.

hectahertz: Thank you for debugging it! I'll add the fix and reproduce this use case in one of the stories.

hectahertz (approved):

.changeset/grumpy-carrots-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': major
+---
+
+Accessibility fixes for SelectPanel. This is a **breaking change** for the SelectPanel because there are two new required props (`title` and `inputLabel`).
```
siddharthkp, line 5: @colebemis Should we make these optional until next major?

docs/content/deprecated/ActionList.mdx
```diff
@@ -14,25 +14,26 @@ Use [new version of ActionList](/ActionList) with composable API, design updates
...
 
 **After**
 
 ```jsx
-<ActionList>
-  <ActionList.Item>New file</ActionList.Item>
-  <ActionList.Item>Copy link</ActionList.Item>
-  <ActionList.Item>Edit file</ActionList.Item>
+<ActionList role="listbox">
+  <ActionList.Item role="option">New file</ActionList.Item>
```
colebemis, line 32: Should `role="option"` be the default value?

src/SelectPanel/SelectPanel.tsx
```diff
@@ -26,7 +32,8 @@ interface SelectPanelBaseProps {
     open: boolean,
     gesture: 'anchor-click' | 'anchor-key-press' | 'click-outside' | 'escape' | 'selection'
   ) => void
-  placeholder?: string
+  title: string
+  inputLabel: string
```
colebemis, line 36: Let's make these optional for now to avoid a breaking release. cc @siddharthkp 
```suggestion
  // TODO: Make these required in the next major release
  title?: string
  inputLabel?: string
```

src/deprecated/ActionMenu.tsx
```diff
@@ -18,7 +18,7 @@ interface ActionMenuBaseProps extends Partial<Omit<GroupedListProps, keyof ListP
   /**
    * A callback that triggers both on clicks and keyboard events. This callback will be overridden by item level `onAction` callbacks.
    */
-  onAction?: (props: ItemProps, event?: React.MouseEvent<HTMLDivElement> | React.KeyboardEvent<HTMLDivElement>) => void
+  onAction?: (props: ItemProps, event?: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>) => void
```
colebemis, line 21: Could these type changes break anything?

vdepizzol:
@bolonio @hectahertz a few design comments:
- Can we keep the `placeholder` attribute with the same `aria-label` contents? For users who don't need accessibility tools, that information should still be useful, and it's quite jarring to see an empty search field.
- Let's use a `large` height going forward
- the `x` close icon button should be dark gray, not blue ;)

Thank you!

bolonio:
> * Can we keep the `placeholder` attribute with the same `aria-label` contents? For users who don't need accessibility tools, that information should still be useful, and it's quite jarring to see an empty search field.

No, we shouldn't. The `placeholder` attribute should include an example of what the user need to insert in the input, while the `aria-label` attribute add more context about the purpose of the input. Technically, the `placeholder` attribute is optional, so it can always be included with their own content, but the content should not be the same as the `aria-label` attribute.

The sentence `For users who don't need accessibility tools` doesn't make any sense.

src/SelectPanel/SelectPanel.tsx
siddharthkp: nice! ship it

hectahertz:
> @bolonio @hectahertz a few quick design comments:
> 
> * Can we keep the `placeholder` attribute with the same `aria-label` contents? For users who don't need accessibility tools, that information should still be useful, and it's quite jarring to see an empty search field.
> * Let's use a `large` height going forward
> * the `x` close icon button should be dark gray, not blue ;)
> * Let's use `small` buttons in the footer 🙌
> 
> Thank you!

Thank you for your comments!

I added an explicit `inputPlaceholder` attribute so that we don't have empty search fields. I do agree with @bolonio on the fact that having the `aria-label` as a placeholder will encourage incorrect `aria-label` values, so I think this is a good compromise!

I addressed all the other visual changes but I wanted to ask about the close icon button.

[Our current `IconButton` variants](https://primer.style/react/storybook/?path=/story/composite-components-button--icon-button) don't support this, and [there is a `ButtonClose` component](https://primer.style/react/storybook/?path=/story/deprecated-components-button--close-button) but it's deprecated.

We probably shouldn't custom-style all our close buttons. Should we bring back the `ButtonClose` component?

hectahertz:
> Also! I just noticed the `hover` state no longer appears when you mouseover while an item is focused:
> 
>  Screen.Recording.2022-07-18.at.13.45.13.mov

Good catch! This one is an old `ActionList` bug. It will get fixed once we upgrade to the new one.

src/deprecated/ActionMenu.tsx
hectahertz: I don't expect any consumers will have coded any checks for these types, but if they have I think they should get the warning as the elements have changed.

src/SelectPanel/SelectPanel.tsx
hectahertz: Added!

.changeset/grumpy-carrots-admire.md
hectahertz: Done, I added reasonable default values so it remains accessible until then.

colebemis (approved):
Awesome work! Thank you, @hectahertz. Looks like there are a few linting errors that need to be fixed. Approving in advance 👍

PR #2144, opened 2022-06-23
SplitPageLayout API

colebemis:
### Summary 
This PR includes documentation for the proposed API of the [SplitPageLayout]( https://github.com/github/primer/issues/793) component.  I'm looking for feedback on the API before I begin implementation.

Here's a preview of the SplitPageLayout documentation: https://primer-48c11ddef3-13348165.drafts.github.io/SplitPageLayout

Here's a diagram of the proposed API:
image
([FigJam file](https://www.figma.com/file/PVzPcwfSMmSv9HtVLeKjqc/SplitPageLayout?node-id=0%3A1))

### Note about responsive props

SplitPageLayout adopts the responsive prop API that @vdepizzol proposed in [this ADR](https://github.com/github/primer/blob/main/adrs/2022-04-15-responsive-design-api-guidelines.md).

Any prop marked "responsive" will optionally accept an object mapping viewport ranges to prop values. Example:

```jsx
// Use "line" divider for all viewport ranges
<SplitPageLayout.Header divider="line">

// Use "line" divider on regular viewports and "filled" for narrow viewports
<SplitPageLayout.Header dividier={{regular: 'line', narrow: 'filled'}}>
```

Part of https://github.com/github/primer/issues/793


docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
+---
+title: SplitPageLayout
+componentId: split_page_layout
+status: Draft
+# description: TODO
```
colebemis, line 5: @vdepizzol Do you have a concise description for SplitPageLayout that we can use here?

vdepizzol: @colebemis wdyt?

```suggestion
description: Provides structure for a split page layout, providing independent scrolling for the pane and content regions. Useful for responsive list/detail patterns, when an item in the pane updates the page content on selection.
```

colebemis: I like it 👍

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+    <Placeholder label="Footer" height={64} />
+  </SplitPageLayout.Footer>
+</SplitPageLayout>
+```
+
+### Without padding
+
+```jsx
+<SplitPageLayout>
+  <SplitPageLayout.Header padding="none">
```
siddharthkp, line 111: nit: Do we need custom padding on every child or could we match the padding prop on the parent like [PageLayout](https://primer.style/react/PageLayout#with-condensed-spacing)?

`padding` prop on every child feels a lot like system props, this looks like you can customise the padding yourself with a value like `padding={0}` or `padding={1}`

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+
+```jsx
+<SplitPageLayout>
+  <SplitPageLayout.Pane>
+    <Placeholder label="Pane" height={120} />
+  </SplitPageLayout.Pane>
+  <SplitPageLayout.Content>
+    <Placeholder label="Content" height={240} />
+  </SplitPageLayout.Content>
+</SplitPageLayout>
```
siddharthkp, line 136: yay! makes so much sense

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+```
+
+### With non-sticky pane
+
+```jsx
+<SplitPageLayout>
+  <SplitPageLayout.Header>
+    <Placeholder label="Header" height={64} />
+  </SplitPageLayout.Header>
+  <SplitPageLayout.Pane sticky={false}>
```
siddharthkp, line 146: brb

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+  <SplitPageLayout.Pane>
+    <Placeholder label="Pane" height={120} />
+  </SplitPageLayout.Pane>
+  <SplitPageLayout.Content>
+    <Placeholder label="Content" height={240} />
+  </SplitPageLayout.Content>
+  <SplitPageLayout.Footer>
+    <Placeholder label="Footer" height={64} />
+  </SplitPageLayout.Footer>
+</SplitPageLayout>
```
siddharthkp, line 28: Really like that this matches PageLayout very closely!

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+  <SplitPageLayout.Pane divider="none">
+    <Placeholder label="Pane" height={120} />
+  </SplitPageLayout.Pane>
+  <SplitPageLayout.Content>
+    <Placeholder label="Content" height={240} />
+  </SplitPageLayout.Content>
+  <SplitPageLayout.Footer divider="none">
+    <Placeholder label="Footer" height={64} />
+  </SplitPageLayout.Footer>
+</SplitPageLayout>
```
siddharthkp, line 85: for later (and for PageLayout): can we change this example to show how to customise dividers (filled, line, none) instead of simply turning them off.

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+  <PropsTableRow
+    name="divider"
+    // TODO: Document ResponsiveValue
+    // This prop accepts a viewport range map (i.e. {regular: ..., narrow: ...}) in addition to a single value.
+    // The `regular` and `wide` keys of the viewport range map can be 'none' or 'line'.
+    // The `narrow` key of the viewport range map can be 'none' or 'line' or 'filled'.
+    type={`ResponsiveValue<
+  'none' | 'line',
+  'none' | 'line' | 'filled'
+>`}
```
siddharthkp, line 178: (for later): Can we bring responsive divider prop to PageLayout as well, i don't like `dividerWhenNarrow` 😅

docs/content/SplitPageLayout.mdx
colebemis: > the default for divider is line in SplitPageLayout, but none in PageLayout, is this am intentional design inconsistency?

I believe this is intentional. @vdepizzol feel free to correct me if I'm wrong

> for later: can we change this example to show how to customise dividers (filled, line, none) instead of simply turning them off.

👍

docs/content/SplitPageLayout.mdx
colebemis: Yup. That's the plan!

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+```
+
+### With pane hidden on narrow viewports
+
+```jsx
+<SplitPageLayout>
+  <SplitPageLayout.Header>
+    <Placeholder label="Header" height={64} />
+  </SplitPageLayout.Header>
+  <SplitPageLayout.Pane visible={{narrow: false}}>
```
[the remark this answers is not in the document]
jonrohan: I think there was push back because it matched html spec. In the PVC case `hidden: true` would end up being attributes on the html tag.

docs/content/SplitPageLayout.mdx
vdepizzol: It's intentional.

docs/content/SplitPageLayout.mdx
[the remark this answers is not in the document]
vdepizzol: I also prefer `hidden`, but like @jonrohan mentioned, there's a big confusion in the way PVC is designed, since it handles element attributes and component properties the same way.

Through [`system_arguments`](https://primer.style/view-components/system-arguments#html-attributes), a boolean `hidden` HTML attribute can be passed to any PVC already. Since the visibility prop also needs to accept responsive values (narrow/regular/wide), and doesn't manipulate the hidden attribute directly, we tried to find an alternative.

If you folks think this is not a problem, I'm happy to use `hidden` instead. I'd just like to make sure the API is the same across PRC and PVC.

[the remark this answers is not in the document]
mperrotti: +1 for "hidden". For hidden FormControl labels, we use a prop named `visuallyHidden`.

docs/content/SplitPageLayout.mdx
mperrotti, line 146: Same issue as `visible` - I think it would be better to have a prop that needs to be set to true to override the default behavior.

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+    defaultValue="'normal'"
+  />
+  <PropsTableSxRow />
+</PropsTable>
+
+### SplitPageLayout.Content
+
+<PropsTable>
+  <PropsTableRow
+    name="width"
```
mperrotti, line 201: Should this be called "maxWidth" if it's the **maximum** width of the content area?

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+
+## Props
+
+### SplitPageLayout
+
+<PropsTable>
+  <PropsTableSxRow />
+</PropsTable>
+
+### SplitPageLayout.Header
```
mperrotti, line 166: I'm not seeing a row for the "height" prop

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+<Note variant="warning">Not implemented yet</Note>
+
+## Examples
+
+### Default
+
+```jsx
+<SplitPageLayout>
+  <SplitPageLayout.Header>
+    <Placeholder label="Header" height={64} />
```
mperrotti, line 17: What happens if I don't pass `height` to any of the placeholders? Is `Placeholder` a real component?

colebemis: Placeholder is a documentation-only component. The Header adapts the height of its content.

docs/content/SplitPageLayout.mdx
colebemis: That makes sense to me. We'd need to update this in PageLayout as well. @vdepizzol @jonrohan How you feel about `width` -> `maxWidth`?

docs/content/SplitPageLayout.mdx
```diff
@@ -0,0 +1,313 @@
...
+### SplitPageLayout.Header
+
+<PropsTable>
+  <PropsTableRow
+    name="divider"
+    // TODO: Document ResponsiveValue
+    // This prop accepts a viewport range map (i.e. {regular: ..., narrow: ...}) in addition to a single value.
+    // The `regular` and `wide` keys of the viewport range map can be 'none' or 'line'.
+    // The `narrow` key of the viewport range map can be 'none' or 'line' or 'filled'.
+    type={`ResponsiveValue<
```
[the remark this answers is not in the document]
colebemis: Agreed. `ResponsiveValue<>` doesn't explain much. The full type signature looks like this:

```tsx
'none' | 'line' | { narrow?: 'none' | 'line' | 'filled'; regular?: 'none' | 'line'; wide?: 'none' | 'line' }
```

Do you think it'd be overwhelming to show the whole thing?

My original plan was to add a separate documentation page about responsive props that explains what `ResponsiveValue<>` means and link to it from here.

docs/content/SplitPageLayout.mdx
colebemis: We're planning to add a `sticky` prop to `PageLayout` as well. For `PageLayout`, `sticky` will default to `false`. For consistency, it feels better to me to have the same prop name for both components instead of having a `sticky` prop on PageLayout and a `static/nonsticky` prop on SplitPageLayout. What do you think?

docs/content/SplitPageLayout.mdx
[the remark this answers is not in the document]
vdepizzol: > There would also be no way to actually set `<div hidden class="Component--hiddenWhenNarrow">` hidden attribute on components that support the responsive structure. Which is something that's done for a11y reasons currently

@jonrohan From the MDN docs, it looks like `hidden` as a property and `display: none` as a style are interchangeable. `hidden` can even be overridden by CSS:

> Note: Changing the value of the CSS [display](https://developer.mozilla.org/en-US/docs/Web/CSS/display) property on an element with the hidden attribute overrides the behavior. For instance, elements styled display: flex will be displayed despite the hidden attribute's presence.
> — https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden

For accessibility, the `aria-describedby` [page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) doesn't describe _how_ the element could be hidden to remain working as an accessible description.

@alliethu @ichelsea quick a11y question: is the `hidden`[^1] required for accessibility reasons, or would hiding an element with CSS directly (using `display: none`, per instance) work the same way?

[^1]: `hidden`, not `aria-hidden` 😉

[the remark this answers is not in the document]
vdepizzol: > in general I'd like system_argument support to be turned off by default, I think moving that way for any new components is the way to go

@jonrohan _yes please_ <3

docs/content/SplitPageLayout.mdx
vdepizzol: I think it should remain as `width`. As a prop, it's a characteristic of the region, not a reference to the CSS attribute. Responsive components are expected to adjust on smaller viewport sizes anyway.

docs/content/SplitPageLayout.mdx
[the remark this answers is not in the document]
colebemis: Thank y'all for weighing in. Based on everyone's input so far, I'm going to move forward with the `hidden` prop for this API draft. It seems like we agree that calling the prop `hidden` is better than `visible` from a component API perspective, but we still need to figure out the implementation details. We can discuss those details in future PRs.

colebemis:
I'm going to go ahead and merge this draft to keep SplitPageLayout moving forward. Feel free to leave any additional feedback about the SplitPageLayout or PageLayout API in the #primer-react channel in Slack.

docs/content/SplitPageLayout.mdx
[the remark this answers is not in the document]
joelhawksley: Chiming in a bit late here, but I'd like to consider using something besides `hidden`. I think we need to be careful of overloading terms in our glossary, and `hidden` is also an HTML attribute. 

While I do generally agree that having system arguments coexist with HTML attributes, in this case I think it's actually a good thing! IMO we should be leaving room for HTML to supercede our component systems as it it ultimately the underlying technology we're building on.

PR #2147, opened 2022-06-27
Fix new select panel issues

camchenry:
@bolonio @hectahertz While integrating with our app, I found a couple of issues that should be addressed.

1. The internal selected state does not synchronize the selected props that are passed in. So, if the selected state changes asynchronously after the `SelectPanel` has already started rendering, then the selected state does not appear as expected. Adding an effect to ensure that the internal selection state stays synchronized if the external selections change makes this work as expected.
2. The click handlers being passed to the close/cancel buttons were passing events as the first argument instead of a gesture string as expected. Creating two separate handlers, one for the `AnchoredOverlay`, and one for the close/cancel buttons ensures that we always pass the gesture string correctly.

### DIFF
--- a/docs/content/SelectPanel.mdx
+++ b/docs/content/SelectPanel.mdx
@@ -44,11 +44,7 @@ function DemoComponent() {
 
   return (
     <SelectPanel
-      renderAnchor={({children, 'aria-labelledby': ariaLabelledBy, ...anchorProps}) => (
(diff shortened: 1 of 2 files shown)


hectahertz (approved):
Looks good to me! Thank you for contributing these fixes 🙂.

bolonio (commented):
Thanks for the help :)

PR #1157, opened 2021-04-09
Remove useMouseIntent in favor of focus-visible

dgreif:
useMouseIntent was manually watching for keyboard/mouse clicks to determine if focus was triggered by the mouse.  There is a new CSS pseudo-class, [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible), which exhibits this exact same behavior.  Also pulled in the `focus-visible` polyfill since browser support isn't complete yet. 

### Merge checklist
- [x] Added/updated tests
- [ ] Added/updated documentation
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

PR #2149, opened 2022-06-27
Use import for `focus-visible` instead of require

### LINKED ISSUE #1207: Usage of `require` breaks rollup builds for consumers
The `require` statement introduced in https://github.com/primer/components/pull/1157/files#diff-5cad94b1fb0ac48b138379a37927067430301335853ae33503a9a64a626fec16R37 can cause issues for consumers if they have Rollup configured for ESM only.  Rather than using `require`, we should use a dynamic `import('focus-visible')`.  This will require a `.d.ts` file be created for `focus-visible` since one does not exist yet.  

Thanks for pointing out this issue @daaku!

dgreif:
In #1157, we added a new import for `focus-visible` but did so using `require`.  Unfortunately this does not translate well to ESM with our current build setup, so it will break in environments that aren't able to handle commonjs.

To fix this, I've switched the `require` to an `import` and moved it to the top of the file.  This means we will apply this polyfill in _all cases_, even if BaseStyles isn't used.  Are we ok with that?

Closes #1207 

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/rude-buckets-fold.md
+++ b/.changeset/rude-buckets-fold.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Converted commonjs `require` of `focus-visible` to an esm `import`
--- a/@types/focus-visible/index.d.ts
+++ b/@types/focus-visible/index.d.ts
@@ -1,0 +1,1 @@
+declare module 'focus-visible'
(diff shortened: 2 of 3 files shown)


colebemis (approved):
Thanks for fixing this, @dgreif!

> To fix this, I've switched the require to an import and moved it to the top of the file. This means we will apply this polyfill in all cases, even if BaseStyles isn't used. Are we ok with that?

I think this is fine 👍

PR #2150, opened 2022-06-27
Convert commonjs code to esm during build

### LINKED ISSUE #2148: ESM Version of `theme-preval.js` is actually CommonJS
**Describe the bug**

PRC is currently built with exports for both commonjs and esm, with the esm version located in `lib-esm`.  Unfortuantely the `theme-preval.js` file included in `lib-esm` is still using commonjs export syntax, which causes it to break environments that only understand ESM.

**To Reproduce**
1. Run `npm run build`
2. Observe the `lib-esm/theme-preval.js` file has commonjs exports like the following:

```js
module.exports = {
  "theme": {
    // theme output here
  }
}
```

**Expected behavior**
The build should output an esm compatible version of `theme-preval.js` with a named export 

```js
export const theme = {
  // theme output here
}
```

dgreif:
When shipping PRC as esm, we want to ensure there is no commonjs code in the `lib-esm` directory.  I've fixed one case of this in #2149, but `theme-preval` is another that is a little tougher to address because it is built as part of the babel build script.  Rather than writing something custom for this one file, I decided to pull in `babel-plugin-transform-commonjs` which will automatically convert files to esm if commonjs imports/exports are detected.  This affects the following files:

image

* `BaseStyles` is addressed in #2149 and will drop off the list once that PR is merged
* theme and theme-preval are the files we are targeting.  I had to change from a named `theme` export to just using the default export because of the way `preval` outputs results.  This change shouldn't have an impact on consumers of the library unless they are reaching all the way into the `lib-esm/theme-preval.js` file.
* The rest are all test/type utility files that have some commonjs code in them, but are not actually intended to be used by consumers

Closes #2148


colebemis:
@dgreif is there a good way to test that this plugin is working correctly and doesn't break anything?

dgreif:
Good question @colebemis, I did do a test where I committed the `lib-esm` directory and then compared the before/after (which is where the screenshot in the description came from).  The output looked as I expected for theme-preval, and also lead me to the issue in BaseStyles (which is now fixed/merged).  I'd say the next step is to pull this into one of our other projects and give it a try.  I'll try it out in dotcom now.

colebemis (approved):
🚢

PR #2157, opened 2022-06-29
Add `InlineAutocomplete` component

iansan5653:
_Writeup adapted from github/memex#10580[^hubber-only]:_

### Component Description

A new `InlineAutocomplete` component that provides support for inline completion suggestions (for example, emojis) in `textarea` and `input` elements.

_Screen recording of the component being used to mention a suggested user in a multiline text box:_

https://user-images.githubusercontent.com/2294248/176513250-b3b50a5f-8dcd-4591-96fe-0116b227d72b.mov

The component is similar in some ways to the existing [`Autocomplete`](https://primer.style/react/Autocomplete) component, but this is a somewhat different pattern in that, when a suggestion is accepted, the value is inserted into the text (hence 'inline') rather than replacing the content. This pattern is closer to an IDE-style autocomplete, where the existing Primer component uses more of a 'search suggestions' pattern.

#### Supported Features

- Stateless: suggestion visibility is fully controlled by the parent component through events and props
- Can wrap any `input` or `textarea` element as well as PRC `Input` and `TextArea` components
- Suggestions have a default rendering, but can be fully customized (ie, to display user avatars)
- Supports multiple trigger types so that different suggestion types can be combined (ie, to support both user mentions and emojis in the same input)
- Predefined hooks provide reusable, mixable behavior for the most common types of suggestions: user mentions, issue references, and emojis
- Keyboard control: select items with arrow keys, apply items with <kbd>Tab</kbd> or <kbd>Enter</kbd>
- Mouse control: click to select items (maintains focus on the input)
- Undo history is preserved when applying suggestions
- Caret position is preserved when applying suggestions

### Additional Work

While adding this component, I also added some new hooks to abstract out some of the underlying logic:

#### `useCombobox` hook

The `useCombobox` hook provides a React-compatible wrapper around the [`@github/combobox-nav`](https://github.com/github/combobox-nav) class. This provides keyboard control and accessible roles for comboboxes (lists of items that can be selected to apply to an input). The hook provides support for comboboxes that can be opened and closed like this one, and calls an easy to use event callback when an option is selected.

#### `useSyntheticChange` hook

When a suggestion is applied, we need to emit a change event from the underlying input. It should look to the API consumer as though the input was typed in - they shouldn't have to add any additional event handlers to react to suggestion application. In addition, we want to preserve the undo stack and caret position - if we have a separate event handler and just update the input's value, the caret position will be reset.

So this hook provides an easy way to simulate a change that is as close as possible to the real thing. The change event preserves the undo stack and caret position and calls the `onChange` handler that is already bound to the underlying input.

### Background / Context

In projects (beta), we are using this component to provide emoji suggestions across the entire application and also to provide support for multiple types of inline suggestions in the Markdown editor.

This component is a dependency of the [`MarkdownEditor`](https://github.com/github/memex/issues/10460)[^hubber-only] (AKA [`CommentBox`](https://github.com/github/primer/issues/715)[^hubber-only]) component as it provides the inline suggestions for `#` references, `@` mentions, and emojis (`:`). Rather than open one giant PR, I wanted to split this work out into a separate task.

🚧 IMO, this component should be added as [experimental](https://primer.style/contribute/component-lifecycle#experimental) status, just like the Markdown editor itself.

Closes github/memex#10580[^hubber-only]. There is not (at this time) a Primer tracking issue for this component.

#### New Dependencies

Two new external dependencies are introduced:

- [`@github/combobox-nav`](https://github.com/github/combobox-nav): See the `useCombobox` hook note above.
  - [ ] **Note:** Due to the way this library uses the `navigator` global, it's incompatible with SSR so it doesn't work with the docs. I have [an open PR](https://github.com/github/combobox-nav/pull/53
) to fix this.
- [`@koddsson/textarea-caret`](https://github.com/koddsson/textarea-caret-position): Detects the position of a character index in a textarea or input element. This is used to position the suggestions list and is already used in GitHub for the existing Markdown editor.

#### Accessibility

This component dynamically adds and removes a combobox as the user is typing. A combobox that is always visible is accessible, but one that is added and removed is not - the suggestions will likely never be read by a screen reader. This issue is already present in GitHub with the existing Markdown editor. When we have the resources to solve this, we should do so by fixing the `@github/combobox-nav` library so the fix will cascade down to the existing Markdown editor in addition to this component.

#### Storybook error

The story works in Storybook, but there is a console error being thrown that I have not been able to fix. I'm not familiar with Slots and could use some help here if anyone has any thoughts. I can't figure out where `data-source` might be getting applied. Is this a Storybook-specific issue?

> Warning: Invalid prop `data-source` supplied to `React.Fragment`. React.Fragment can only have `key` and `children` props.
>
>     at Slots
>     at UserSuggestion
>     at SuggestionListItem
>     at ul
>     at StyledComponent
>     at List__ListBox
>     ...

### Merge checklist

- [x] Added/updated tests (working on this - it's tested as part of `MarkdownEditor` in projects beta; I need to extract those tests and make them specific to this component)
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

Need to merge and update combobox-nav library to fix [docs issue](#new-dependencies):

- [x] https://github.com/github/combobox-nav/pull/53

[^hubber-only]: :octocat: Link only accessible to GitHub employees


src/InlineAutocomplete/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+  // the possibility of it being 'normal' (another bug). In that case, fall back to a rough guess
+  // of 1.2 based on MDN: "Desktop browsers use a default value of roughly 1.2".
+  if (isNaN(coords.height)) coords.height = parseInt(getComputedStyle(input).fontSize) * 1.2
+
+  // For some single-line inputs, the rightmost character can be accidentally wrapped even with the
+  // wordWrap fix above. If this happens, go back to the last usable index
+  let adjustedIndex = index
+  while (input instanceof HTMLInputElement && coords.top > coords.height) {
+    coords = getCaretCoordinates(input, --adjustedIndex)
+  }
```
iansan5653, line 75: Admittedly we shouldn't just be hacking around these bugs in [the library](https://github.com/koddsson/textarea-caret-position). I really should open a PR there and fix them but I've been lazy. The library hasn't been updated in a while but the owner is staff :octocat: so we should be able to get it done. I will definitely try and get to that before we merge this.

src/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -0,0 +1,177 @@
...
+ */
+const InlineAutocomplete = ({
+  triggers,
+  suggestions,
+  onShowSuggestions,
+  onHideSuggestions,
+  sx,
+  children,
+  // Forward accessibility props so it works with FormControl
+  ...forwardProps
```
iansan5653, line 77: The `FormControl` component clones its child and adds an `id` prop to it. Since the child in this case will be `InlineAutocomplete` instead of `Input`, we need to forward the ID on down.

src/InlineAutocomplete/types.ts
```diff
@@ -0,0 +1,56 @@
...
+       * across all currently visible suggestions.
+       */
+      value: string
+      /**
+       * Optional key. If not provided, the `value` will be used. Setting a `key` allows
+       * for non-unique `value`s.
+       */
+      key?: string
+      /** This must return an `ActionList.Item` instance. */
+      render: (props: ActionListItemProps) => React.ReactElement
```
iansan5653, line 41: I don't love `render` functions and I know that's not really the Primer way, but I was unable to think of a better API. We could theoretically use composition here, but the list is pretty much always going to be an array that is then mapped to components - I feel like that would generally be less readable. I'm definitely open to suggestions.

src/hooks/useSyntheticChange.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Returns a function that will synthetically change the input, attempting to maintain caret
+ * position and undo history as though the user had typed using a keyboard.
+ *
+ * Will first attempt to use the non-standard browser `execCommmand` API to simulate a typing
+ * action. Failing this (ie, in test environments or certain browsers), the fallback handler
+ * will be called with a fake constructed `ChangeEvent` that looks like a real event.
+ */
+export const useSyntheticChange = ({inputRef, fallbackEventHandler}: UseSyntheticChangeSettings) =>
```
iansan5653, line 91: Heavily inspired by the [`markdown-toolbar-element`](https://github.com/github/markdown-toolbar-element/blob/main/src/index.ts#L376)

src/InlineAutocomplete/utils.ts
iansan5653: @koddsson would you mind enabling issues on that repo so I can catalog the bugs there?

src/hooks/useSyntheticChange.ts
mperrotti, line 91: Nice. This might help us resolve [this Autocomplete bug](https://github.com/primer/react/blob/main/src/Autocomplete/AutocompleteInput.tsx#L133-L135)

src/InlineAutocomplete/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+ * @param adjustForScroll Control whether the returned value is adjusted based on scroll position.
+ */
+export const getCharacterCoordinates = (
+  input: HTMLTextAreaElement | HTMLInputElement | null,
+  index: number,
+  adjustForScroll = true
+): Coordinates => {
+  if (!input) return {top: 0, left: 0}
+
+  // word-wrap:break-word breaks the getCaretCoordinates calculations (a bug), and word-wrap has
```
mperrotti, line 59: Thanks for working around these `getCaretCoordinates` bugs. It would be nice to fix them in the library, but that's definitely not a blocker for this PR.

mperrotti (commented):
Nice work. I'll keep an eye on this PR and give it a closer look when it's ready for review.

src/stories/InlineAutocomplete.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+    {user.name}
+    <ActionList.Description>{user.login}</ActionList.Description>
+  </ActionList.Item>
+)
+
+export const Default = () => {
+  const [suggestions, setSuggestions] = useState<Suggestions | null>(null)
+
+  const onShowSuggestions = (event: ShowSuggestionsEvent) => {
+    const filteredUsers = sampleUsers.filter(
```
mperrotti, line 51: In the other Autocomplete component we default to an alphanumeric filter, but allow that to be overridden using the `filterFn` prop. https://github.com/primer/react/blob/c7bbd06fd7c67690ed7a612737f79662444d388b/src/Autocomplete/AutocompleteMenu.tsx#L206

src/InlineAutocomplete/types.ts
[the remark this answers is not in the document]
iansan5653: Yeah at this point we don't - we let them use anything.

src/InlineAutocomplete/types.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export interface ShowSuggestionsEvent {
+  /** The trigger that caused this query. */
+  trigger: Trigger
+  /** The query string. */
+  query: string
+}
+
+export type Suggestion =
```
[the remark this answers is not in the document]
iansan5653: That API will definitely increase the complexity of finding and cloning the child input though. I guess I will need to learn Slots 🙂

src/hooks/useSyntheticChange.ts
iansan5653: Yes, I think it would!

src/stories/InlineAutocomplete.stories.tsx
iansan5653: Personally, I think this component is a lot easier to use if it doesn't manage its own state - in other words, the parent component handles filtering and then tells this component what to display.

That allows a lot of flexibility without a ton of props - the parent component can use a custom filtering function, control the sorting (very important when using fuzzy filtering), control the number of visible suggestions, and manage multiple sets of suggestions (when there are multiple types of triggers).

If we want to keep it consistent, I think I would eventually prefer to move `Autocomplete` in this stateless direction as well.

iansan5653:
> Could we move this work to drafts/ to start? We can move it back to the main bundle once we're ready to mature InlineAutocomplete to an "Alpha" status

I don't mind this, but what's the difference between "Draft" and "Experimental"? Would we be able to start transitioning projects beta (memex) to use the Primer component if the Primer component is a draft?

owenniblock:
@iansan5653 I would love to do some work on the screen reader support on this component. Currently if I activate the menu, because the screen reader is not aware that the menu appears, it's not announced by the screen reader and if the user presses down, because the menu is not associated with the input, the screenreader hijacks the down arrow and items cannot be selected.

Give me some time to discuss this in the Accessibility team but we might be able to resolve this by using [aria-activedesendant](https://www.w3.org/WAI/GL/wiki/Using_aria-activedescendant_to_allow_changes_in_focus_within_widgets_to_be_communicated_to_Assistive_Technology#:~:text=The%20aria-activedescendant%20attribute%20is%20generally%20used%20in%20composite,the%20current%20active%20child%20element%20within%20the%20widget.)

I'll have to double check if it's compatible with this usecase.

iansan5653:
@owenniblock thanks for the accessibility notes. I am 100% on board with fixing the accessibility issues.

At the moment, I'm using the [`combobox-nav`](https://github.com/github/combobox-nav) tool (GitHub-owned) to bind keyboard shortcuts and accessible roles. This is the same pattern we are using on the existing Rails Markdown editor across dotcom. I know that the problem also exists in the Rails editor

If I understand correctly from reading the docs, it seems that `auto-complete-element` is specifically designed for pulling autocomplete results from a server endpoint, which is not something we are necessarily doing here. Under the hood, the `auto-complete-element` is still dependent on `combobox-nav`.

I _think_ the solution to this problem may lie in improving that `combobox-nav` utility, which would allow the changes to cascade down to the whole site. I'm not 100% sure though - we may also have to change some things on this React end to avoid unmounting/remounting the menu?

---

Just for reference, expand below to see the (simplified) HTML when the menu is closed. If we can figure out exactly what needs to change, we can then easily figure out whether the fix needs to be done inside `combobox-nav` or within this component (or both).

<details>
<summary>View HTML</summary>

_When the list is not visible:_
```html
<label for="react-aria3864888683-1">Inline Autocomplete Demo</label>

<textarea
  aria-required="false"
  aria-invalid="false"
  id="react-aria3864888683-1"
  aria-describedby="react-aria3864888683-1-caption"
></textarea>

<span id="react-aria3864888683-1-caption"
  >Try typing '@' to show user suggestions.</span
>

<div id="portal-root"></div>
```

_When the list is visible:_
```html
<label for="input-id">Inline Autocomplete Demo</label>

<textarea
  aria-required="false"
  aria-invalid="false"
  id="input-id"
  aria-describedby="input-caption"
  role="combobox"
  aria-controls="combobox-7750"
  aria-expanded="true"
  aria-autocomplete="list"
  aria-haspopup="listbox"
></textarea>

<span id="input-caption">Try typing '@' to show user suggestions.</span>

<div id="portal-root">
  <ul role="listbox" id="combobox-7750">
    <li
      tabindex="0"
      aria-labelledby="monalisa monalisa--inline-description"
      role="option"
      id="combobox-react-aria3864888683-2__option-0"
      aria-selected="true"
    >
      <span id="monalisa">Monalisa Octocat</span>
      <div id="monalisa--inline-description" title="monalisa">monalisa</div>
    </li>
    <li
      tabindex="0"
      aria-labelledby="primer primer--inline-description"
      role="option"
      id="combobox-react-aria3864888683-2__option-1"
      aria-selected="false"
    >
      <span id="primer">Primer</span>
      <div id="primer--inline-description" title="primer">primer</div>
    </li>
    <li
      tabindex="0"
      aria-labelledby="github github--inline-description"
      role="option"
      id="combobox-react-aria3864888683-2__option-2"
      aria-selected="false"
    >
      <span id="github">GitHub</span>
      <div id="github--inline-description" title="github">github</div>
    </li>
  </ul>
</div>

```
</details>

iansan5653:
Oh, it's my bad! It is actually setting it, but I had blurred the input (after disabling hiding the menu) to copy the HTML and that must have cleared it. When I focus the input and press down the activedescendant is set.

src/InlineAutocomplete/types.ts
iansan5653: Resolving - discussed elsewhere

docs/content/InlineAutocomplete.mdx
```diff
@@ -0,0 +1,71 @@
+---
+componentId: inline_autocomplete
+title: InlineAutocomplete
+status: Experimental
```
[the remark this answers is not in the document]
iansan5653: What's the difference in 'Experimental' and 'Draft'?

src/hooks/index.ts
```diff
@@ -11,4 +11,6 @@ export type {UseOverlaySettings} from './useOverlay'
 export {useRenderForcingRef} from './useRenderForcingRef'
 export {useProvidedStateOrCreate} from './useProvidedStateOrCreate'
 export {useMenuInitialFocus} from './useMenuInitialFocus'
+export type {ComboboxCommitEvent} from './useCombobox'
+export {useCombobox} from './useCombobox'
```
iansan5653, line 15: I didn't export `useSyntheticChange` because it feels more like an internal tool, but that may be debatable. It definitely could theoretically be useful outside of Primer.

iansan5653:
@owenniblock I've updated the component to (hopefully) make it work well with screen readers. I think based on my testing it should actually work _better_ than the Markdown editor that is in production now.

Here's a list of decisions I made:

- Set the `role=combobox` _only_ when suggestions are available. This is because the `combobox` role is technically incompatible with multi-line inputs, so if apply it all the time then the input will not look multiline. By dynamically changing it when the menu appears, we can be correct as much of the time as possible while still providing useful roles.
- Don't set `aria-activedescendant` when the suggestions first appear, even though the first item will be applied if <kbd>Enter</kbd> is pressed. This is because setting the `activedescendant` makes it look like the input is not focused anymore, which stops the screen reader from reading out the typed characters. It's also really disruptive to the user.
- Insert an `aria-live` element with a status message that indicates when suggestions are available, how many there are, what the first one is, and how to apply it. This is `assertive` to prevent VoiceOver from reading the input label when the role changes, which it does try to do. `assertive` works well here since we don't want to wait until the user stops typing to read it.

I would appreciate any testing/feedback you can give me! I'd like to make this as ergonomic as possible for screen reader users.

Note - I do recommend testing this in a tab outside the Storybook with the "Open in new tab" button because VoiceOver may double-read aria-live regions inside iframes (it's a bug).

owenniblock:
There are enough nuances with these changes that I think we should get a screen reader user to check them. Some of these decisions may be disruptive / confusing for screen reader users but I can't comment on that personally because I am not a screen reader user myself.

@inkblotty is there anything specific I need to do to engage James on a PR like this? Or shall I just add him as an Assignee and explain what we need?

inkblotty:
@owenniblock Go ahead and ping him in Slack 👍

iansan5653:
FWIW I did discuss this with James ([Slack thread 🧵](https://github.slack.com/archives/C0FSWLQ0Y/p1657658386295259)) already so he should already have some context - but I definitely agree with running the final product by him. I wish there was a way to deploy a staging Storybook from this PR to make it easier to share

iansan5653:
Here's another Slack thread with @jscholes and @ichelsea about this component's accessibility. To summarize here:

- The feedback message (the `aria-live` message that indicates when suggestions are available) is too long. @jscholes suggested changing it:
  ```diff
  - 3 autocomplete suggestions available. The first is "monalisa"; press Enter or Tab to insert it.
  + 3 autocomplete suggestions; monalisa is highlighted. Press Enter to insert.
  ```
- We discussed the use of <kbd>Tab</kbd> to insert a suggestions. It's not ideal to hijack tab-navigation behavior but this may be expected functionality when used by software developers who are used to IDE suggestions (and are used to the current state of the Markdown editor). We'd like to eventually introduce a user setting to control this.
- When you press <kbd>Down Arrow</kbd> one time, the `activedescendant` should be the _first_ item, not the second. This is despite the first item getting inserted by default when you press <kbd>Enter</kbd>. @ichelsea linked to [some documentation](https://github.com/github/accessibility/blob/main/docs/coaching-recommendations/command-palette/list-semantics.md) describing this exact scenario. In short, we should style the default-insert element differently than the currently active descendant.

I'll address these items today.

docs/content/InlineAutocomplete.mdx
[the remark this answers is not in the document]
mperrotti: Yes, you'd be able to use the Primer component as a draft.

I honestly forget the difference, but I think @colebemis or @siddharthkp might be able to explain it.

src/InlineAutocomplete/types.ts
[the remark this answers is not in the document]
mperrotti: Yup, we're ok with unstable APIs in experimental/draft components.

src/stories/InlineAutocomplete.stories.tsx
mperrotti: You make a great point. Go ahead and keep the filtering as-is.

We could reconsider changing the Autocomplete API in the next breaking change.

andrialexandrou:
We're waiting for these pieces to come together, and once they do I'll do a behavioral test to be sure the requested changes all work as expected. 👍

docs/content/InlineAutocomplete.mdx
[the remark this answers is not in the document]
siddharthkp: From https://primer.style/contribute/component-lifecycle: 

> Experimental are Proof-of-concept built outside of Primer.
>
> The component exists as a proof-of-concept built outside of Primer, often in a GitHub application such as GitHub.com.

`drafts` is where we keep components in primer/react that are not part of the public API because we are not ready to commit to the html structure or API yet but we'd like to start testing them out in applications (perfect for your use case). These can be in any stage of the [maturity lifecycle](https://primer.style/contribute/component-lifecycle).

iansan5653:
Okay, should finally be ready now! Just to summarize, the accessibility fixes since the last review include:

- Shortened the status message
- Made <kbd>Tab</kbd> hijacking optional and disabled by default. When enabled, the status message indicates that <kbd>Tab</kbd> will insert
- When suggestions appear, the first item in the list is highlighted using a lighter background to indicate that it is default, but not selected. When the user presses the down arrow, the first item becomes selected, with a darker background.

iansan5653:
I see the failing "Check for changeset" action. Should I add a changeset to this PR considering it's targeting draft status? If so, how do I do that?

andrialexandrou (approved):
Alright, did a test on Storybook and confirmed:

- When enabled, tab completion inserts first item
- When not enabled, tab exits the textarea
- When the component appears, the first item is highlighted and inserts on enter[/tab]
- When the component appears and press down arrow, the first item is active
- SR announcements are the revised feedback message

So, looks good!

I did notice that the SR announcement of each item read both the handle and the display name. This seems preferable to reading only handle or display name. Just taking note here so we're all on the same page. 
Inline Autocomplete demo with 3 menuitems, with visible text reading, in order, "Monalisa Octocat monalist", "Primer primer", "GitHub github".

Thanks Ian

iansan5653:
Thanks @andrialexandrou for taking a look at this!

> I did notice that the SR announcement of each item read both the handle and the display name. This seems preferable to reading only handle or display name. Just taking note here so we're all on the same page.

Agreed, this is intentional. Just noting here that this HTML is controlled by the consumer (not the component) so the component doesn't have much control over how the items are read.

docs/content/InlineAutocomplete.mdx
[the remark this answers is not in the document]
iansan5653: I suppose this is both an experimental and a draft component then? This component does exist as a POC in Memex.

iansan5653:
I think we should be good to go with this now!

src/InlineAutocomplete/_AutocompleteSuggestions.tsx
```diff
@@ -0,0 +1,119 @@
...
+    <Overlay
+      onEscape={onClose}
+      onClickOutside={onClose}
+      returnFocusRef={inputRef}
+      preventFocusOnOpen
+      portalContainerName={portalName}
+      sx={{position: 'fixed'}}
+      {...{top, left}}
+    >
+      <ActionList selectionVariant="single" ref={setList}>
```
siddharthkp, line 103: Hi! What's the significance of using selectionVariant? would we like to show selected option with a checkmark?

iansan5653: Hmm, that might be leftover from when I first wrote this and I may have thought that `selectionVariant` was required for `ActionList`. I will take a look and see if I can remove that.

iansan5653: Yeah I removed it and it looks better (no leading space in front of the items!). I have no idea why I did that originally 🙃

siddharthkp (approved):
Good to go for drafts 👍

siddharthkp:
Is this empty file intentional? https://github.com/primer/react/pull/2157/files#diff-e5c61c2ea86f2684b47c752adc87f089e3c2c06efb57b23deea6158fea599114

iansan5653:
> Is this empty file intentional?

Yes, it doesn't compile without it. I guess every folder in that directory needs an `index.d.ts` file.

PR #2158, opened 2022-06-30
Selected tab doesn't need to be <a> tag

owenniblock:
After testing memex against the new TabNav it became clear that my TypeScript was skipping the case where the tab is changed from the default `<a>` to a `<div>` - this change should fix that issue.

I also noticed that `wrap` seems to be broken on the memex side but I cannot replicate that here so I'm going to use the build from the PR to do some more digging.  EDIT: Confirmed that `wrap` appears to work correctly once the issue was resolved.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/thick-pets-shop.md
+++ b/.changeset/thick-pets-shop.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Fixes issue when tabs are not links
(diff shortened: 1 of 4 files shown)


siddharthkp (approved):
Looks good to me 😄

PR #2131, opened 2022-06-20
Release tracking

primer-css:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   [#2108](https://github.com/primer/react/pull/2108) [`e5be3db3`](https://github.com/primer/react/commit/e5be3db3112db20efef5e49ebe89ea3af17fd486) Thanks [@mperrotti](https://github.com/mperrotti)! - Adds a draft component to render a basic segmented control.

-   [#2139](https://github.com/primer/react/pull/2139) [`f17446e0`](https://github.com/primer/react/commit/f17446e0dbfb9021291207972ece47e90a5656e0) Thanks [@owenniblock](https://github.com/owenniblock)! - Changes focus rules of TabNav to match WAI-ARIA rules for tablist

### Patch Changes

-   [#2099](https://github.com/primer/react/pull/2099) [`40da598e`](https://github.com/primer/react/commit/40da598e8d41b688eba53f4c594bff66b269f5f4) Thanks [@siddharthkp](https://github.com/siddharthkp)! - AnchoredOverlay: `aria-expanded` attribute is removed from anchor when overlay is not open

-   [#2155](https://github.com/primer/react/pull/2155) [`003cbcf0`](https://github.com/primer/react/commit/003cbcf0e5530d580636e4ce09e539317a95dc73) Thanks [@mperrotti](https://github.com/mperrotti)! - Update Checkbox component to useIsomorphicLayoutEffect instead of useLayoutEffect to support SSR

-   [#2154](https://github.com/primer/react/pull/2154) [`96ad635b`](https://github.com/primer/react/commit/96ad635b3063af03e31fcac06cfe8bdea248d0d3) Thanks [@mperrotti](https://github.com/mperrotti)! - Allow "falsely/empty" Autocomplete.Input values

-   [#2153](https://github.com/primer/react/pull/2153) [`ce45de30`](https://github.com/primer/react/commit/ce45de308f1aaaf760a8e3f18ee39f20610f896b) Thanks [@willglas](https://github.com/willglas)! - Add AutocompleteContext to Autocomplete component exports

[...]

PR #2159, opened 2022-06-30
Keep anchoredoverlay tabindex

siddharthkp:
This was removed in https://github.com/primer/react/pull/2099/commits/e7f2d03df8d40364270c1af750b86f0941238dbe to fix https://github.com/github/primer/issues/989 but this causes tests a test in memex to fail: [table-focus-test.tsx#L365-L380](https://github.com/github/memex/blob/main/src/tests/components/react-table/table-focus-test.tsx#L365-L380)

The use case is a bit complicated and would probably need a change in memex as well. To unblock https://github.com/primer/react/pull/2131, I'm taking this commit out and reopening https://github.com/github/primer/issues/989

### DIFF
--- a/.changeset/anchored-overlay-clever-mails-roll.md
+++ b/.changeset/anchored-overlay-clever-mails-roll.md
@@ -1,8 +1,5 @@
 ---
-"@primer/react": patch
+'@primer/react': patch
 ---
 
-AnchoredOverlay accessibility fixes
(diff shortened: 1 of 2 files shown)


owenniblock (approved):
We should probably revisit this once the release is out - I assume we added it for a good reason but lack the context to comment on the wider aspects of removing it 😅

Issue #2118, opened 2022-06-07
Buttons should support href as a prop


jdrush89:
**Describe the bug**
According to https://github.com/primer/react/issues/1972, LinkButtons have been removed in favor of `<Button as="a" />`.  If this is the recommended way of doing link buttons, then passing an href to the Button component should forward that prop to the a element.  Otherwise, an onClick that sets the window.location.href has to be used which is more code and clunkier.

**To Reproduce**
1. Go to https://primer.style/react/Button
2. Update the first example to `<Button as="a" href="http://google.com">Default</Button>`
3. Click on the Button in the first example
![image](https://user-images.githubusercontent.com/955442/172441354-62c3df2f-43f8-4123-a379-071932b15351.png)

**Expected behavior**
The user should be navigated to the url passed in as the href.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser chrome
 - Version 101.0.4951.67 (Official Build) (32-bit)

AdamShwert:
This bug also causes an issue where when you middle mouse click on a button, nothing happens because the href prop is not passed through. We are currently still using LinkButtons as a result so as to not lose functionality until this issue is resolved

PR #2166, opened 2022-07-01
Button should respect polymorphic as prop

mattcosta7:
Describe your changes here.

Closes https://github.com/primer/react/issues/2118

It seems like `LinkButton` was removed under the expectation that `Button` would act in apolymorphic way and handle styling links via the `As` prop, however it seems around the same time that functionality was blocked, since any `as` prop passed to button is overridden by the default prop here [see details](https://github.com/primer/react/pull/1880)

I've updated the types to account for this, and updated the test to avoid skipping the `as` prop component test.

This should allow projects to avoid using the now undocumented `LinkButton` and instead allow for properly passing as props to `Button` directly

### Screenshots

n/a

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mattcosta7:
Will take a look at ci here a bit later, if the changes seems reasonable (or at least the spirit of this change)

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@primer/react",
-  "version": "35.3.0",
```
mattcosta7, line 3: looks like publishing the last version didn't bump this

src/Button/Button.tsx
```diff
@@ -5,7 +5,7 @@ import {ButtonBase} from './ButtonBase'
 const ButtonComponent = forwardRef<HTMLButtonElement, ButtonProps>(
   ({children, ...props}, forwardedRef): JSX.Element => {
     return (
-      <ButtonBase ref={forwardedRef} {...props} as="button">
+      <ButtonBase ref={forwardedRef} as="button" {...props}>
```
mattcosta7, line 8: allow `as` to be overwritten by passed in props

src/Button/types.ts
```diff
@@ -13,7 +13,10 @@ export type VariantType = 'default' | 'primary' | 'invisible' | 'danger' | 'outl
 
 export type Size = 'small' | 'medium' | 'large'
 
-type StyledButtonProps = ComponentPropsWithRef<typeof StyledButton>
+/**
+ * Remove styled-components polymorphic as prop, which conflicts with radix's
+ */
+type StyledButtonProps = Omit<ComponentPropsWithRef<typeof StyledButton>, 'as'>
```
mattcosta7, line 19: the crux of the fix here seems like the `as` prop applied by styled-components differs from the types of that applied using the Radix polymorphic helper types an an incompatible way.  By omitting the styled component version of this type from the types that get passed around, we can rely on the radix version and avoid weird issues when trying to apply the as prop to the button base to the left of the spread restProps

jdrush89 (approved):

hebertcisco (approved):

pksjce (approved):
Added back a few stories using this feature. LGTM 👍

Issue #2008, opened 2022-03-31
`IconButton` should include `Tooltip` by default


iansan5653:
## Summary

There is no documented example of how to render an icon-only button with an accessible tooltip in a way that would meet all of the following standard requirements:

- Render as an icon-only button with a single focus target
- Read the label one time in a manner that ties it to the button, for screen reader users
- Show the label on hover, for mouse users
- Show the label on focus, for keyboard users

If this is a supported behavior, then some documentation is needed to show how to do it. If this is not supported, it would be very useful functionality to add to Primer.

The [`IconButton`](https://primer.style/react/IconButton) documentation provides an example of how to label the component in an accessible manner:
```tsx
<IconButton aria-label="Search" icon={SearchIcon} />
```

And the [`Tooltip`](https://primer.style/react/Tooltip) documentation has the same:
```tsx
<Tooltip aria-label="Hello, Tooltip!">Text with a tooltip</Tooltip>
```

Both of these examples use `aria-label` however, which provides no obvious way to create an icon button that shows its accessible label on hover as a tooltip. 

## Use Cases

This is a very common requirement. See for example these instances on github.com:

**Notifications bell**:
<img width="391" alt="Screenshot of the GitHub notifications bell button showing a tooltip that says 'you have unread notifications'" src="https://user-images.githubusercontent.com/2294248/161074778-dd755940-5cbe-440a-8712-840227937c52.png">

**Markdown editor toolbar**:
<img width="192" alt="Screenshot of the GitHub markdown editor 'bold' button showing a tooltip" src="https://user-images.githubusercontent.com/2294248/161074856-157f1b16-4023-48e6-8b18-6e59275d408d.png">

**Code view**:
<img width="192" alt="Screenshot of the GitHub pull request code view, showing the 'Expand all' button with a tooltip" src="https://user-images.githubusercontent.com/2294248/161075094-686ed8d6-b6a0-4fd1-aa76-a0bedee80a60.png">

This is recommended as a best practice by [Primer's own tooltip guide](https://primer.style/design/accessibility/tooltip-alternatives):

> <img width="473" alt="Do: Use a tooltip to give a text label to an icon button" src="https://user-images.githubusercontent.com/2294248/162039835-56f30f63-0ea9-4780-9bf0-fce995c2a964.png">

## Attempts

I've tried all of the following examples but I haven't been able to find a satisfying way to do this:

**Label on tooltip**: Tooltip does not appear on button focus and the button appears as unlabelled in the accessibility tree.
[...]

PR #2182, opened 2022-07-22
Add `MarkdownEditor` and `MarkdownViewer` components

iansan5653:
Screenshot of the editor component

Closes https://github.com/github/memex/issues/10460 [^hubber-only]

#### `MarkdownEditor`

`MarkdownEditor` is an input for writing Markdown and previewing the result. Features include:

- A built-in toolbar with Markdown formatting tools and support for customization
- Standard keyboard shortcuts for formatting (ie, <kbd>⌘B</kbd> to bold selected text)
- Inline autocompletion for `@`-mentions, `#`-references, and emojis (`:`)
- Previewing Markdown using an async function (ie, an API request)
- Optional file upload through drag-and-drop, pasting, or clicking a button
- Pasting links and tables
- List editing: new list items are automatically inserted upon <kbd>Enter</kbd> when editing a list
- Indenting and dedenting selected text with <kbd>Tab</kbd> and <kbd>Shift + Tab</kbd>
- Automatic resizing of the editor height as the user types
- Formatting and other actions preserve the undo history
- Inserting saved replies
- Compositional API for custom toolbar buttons, action buttons, and label

:x: Not planned for this PR:
- Reordering (drag and drop) task list items & converting task items to issues
- Custom (non-Unicode) emoji suggestions, like :octocat: (requires [`g-emoji`](https://github.com/github/g-emoji-element) element; proposal: https://github.com/github/primer/issues/1152)
- Emoji skin tone preference (also requires `g-emoji` element)
- Sanitizing HTML
- Tooltips for toolbar buttons (depends on https://github.com/primer/react/issues/2008)

##### Screen Recording of Usage

https://user-images.githubusercontent.com/2294248/180266205-64225adf-e254-4c5f-8b82-cefb5e81953e.mov

##### Props

https://github.com/primer/react/blob/c167ff0429da32bd58bfc8dc795ef9214a9bc61a/src/drafts/MarkdownEditor/MarkdownEditor.tsx#L28-L98

##### Sub-Components

https://github.com/primer/react/blob/c167ff0429da32bd58bfc8dc795ef9214a9bc61a/src/drafts/MarkdownEditor/index.ts#L9-L34

##### Usage Example

https://github.com/primer/react/blob/c167ff0429da32bd58bfc8dc795ef9214a9bc61a/src/drafts/MarkdownEditor/MarkdownEditor.stories.tsx#L251-L284

#### `MarkdownViewer`

`MarkdownViewer` is a wrapper for displaying rendered Markdown HTML. Features:

- Checking and unchecking task list items
- Optionally links to open in a new tab
- Intercept links to change routing behavior

### Additional Work

In addition to the two components, this PR contains many dependent changes & new functionality.

#### New Hooks

- `useUnifiedFileSelect`: This hook adds an easy to use API for unifying various methods of selecting files (dragging and dropping, pasting, and clicking a button). It combines all the necessary events into a single file select event
- `useIgnoreKeyboardInputWhileComposing`: When listening for <kbd>Enter</kbd> keypresses (ie, to submit forms), it is important to filter out keypresses made while the user is inputting using [IME](https://docs.microsoft.com/en-us/windows/apps/design/input/input-method-editors). This is frustratingly difficult to do using browser APIs because the key press events look the same, so this hook aims to make the process much easier. For more details, see https://github.com/github/memex/issues/10800 [^hubber-only]
- `useDynamicTextareaHeight`: Calculates the optimal height for a `textarea` based on the contents, automatically adjusting as the user types
- `useSafeAsyncCallback`: Makes it easier to manipulate state in async callbacks without memory leaks or accidentally calling outdated references.

#### Other Changes

- Extracted the character coordinates utilities from the `InlineAutocomplete` component so they can be used by other code (this is necessary for `useDynamicTextareaHeight`)
- Updated `useResizeObserver` to allow attaching the observer to elements other than the document root
- Updated `createSlots` to use layout effects when getting slots (reduces the number of renders)

### New Dependencies

- `fzy.js`: Powers fuzzy matching in inline suggestions
- `@github/markdown-toolbar-element`: Provides Markdown formatting tools (I eventually want to remove this dependency and just build the logic into the component; I don't think this element is a great implementation for the use case)
- `@github/paste-markdown`: Markdown pasting support (pasting tables & links)

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

[^hubber-only]: :octocat: Link only accessible to GitHub employees


src/MarkdownEditor/index.ts
```diff
@@ -0,0 +1,38 @@
...
+   */
+  Actions,
+  /** A button to show in the editor footer. */
+  ActionButton
+})
+export default MarkdownEditor
+
+export {MarkdownEditorProps} from './MarkdownEditor'
+export {Emoji, Mentionable, Reference} from './_useSuggestions'
+export {FileUploadResult} from './_useFileHandling'
```
Lukeghenco, line 38: these need to have export type on them. 

```suggestion
export type {MarkdownEditorProps} from './MarkdownEditor'
export type {Emoji, Mentionable, Reference} from './_useSuggestions'
export type {FileUploadResult} from './_useFileHandling'
```

src/drafts/index.ts
```diff
@@ -7,3 +7,15 @@
...
+export type {
+  InlineAutocompleteProps,
+  ShowSuggestionsEvent,
+  Suggestion,
+  Suggestions,
+  Trigger
+} from '../InlineAutocomplete'
+
+export {default as MarkdownViewer} from '../MarkdownViewer'
+export type {MarkdownViewerProps, InteractiveMarkdownViewerProps} from '../MarkdownViewer'
```
Lukeghenco, line 21: This is missing the exports to access the MarkdownEditor

```suggestion
export type {MarkdownViewerProps, InteractiveMarkdownViewerProps} from '../MarkdownViewer'
export {default as MarkdownEditor} from '../MarkdownEditor'
export type {MarkdownEditorProps} from '../MarkdownEditor'
```

src/MarkdownEditor/Toolbar.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+import '@github/markdown-toolbar-element'
+import {isMacOS} from '@primer/behaviors/dist/esm/utils'
+import Box from '../Box'
+import {IconButton, IconButtonProps} from '../Button'
+import {useFocusZone} from '../hooks/useFocusZone'
+import {MarkdownEditorSlot} from './MarkdownEditor'
+import {MarkdownEditorContext} from './_MarkdownEditorContext'
+
+export const ToolbarButton = forwardRef<HTMLButtonElement, IconButtonProps>((props, ref) => {
```
Lukeghenco, line 25: 🔥

src/utils/create-slots.tsx
```diff
@@ -82,7 +83,7 @@ const createSlots = <SlotNames extends string>(slotNames: SlotNames[]) => {
   }> = ({name, children}) => {
     const {registerSlot, unregisterSlot, context} = React.useContext(SlotsContext)
 
-    React.useEffect(() => {
+    useLayoutEffect(() => {
```
iansan5653, line 86: I was running into really weird issues with the file upload where the editor content wouldn't get updated when the file upload completed. I traced it down to the commit where I introduced slots. I have to admit I don't full understand why this was happening or why this fix works, but I think it is because the file upload handler is asynchronous and so is `useEffect`; there must have been some clashing here where a render got 'lost'. Changing this to `useLayoutEffect` fixed the problem and slots still seem to work fine in every component that uses them.

src/utils/character-coordinates.ts
```diff
@@ -0,0 +1,67 @@
+import getCaretCoordinates from '@koddsson/textarea-caret'
```
siddharthkp, line 1: Can we move this to `drafts/utils`?

src/_InputLabel.tsx
```diff
@@ -3,25 +3,41 @@ import {Box} from '.'
 import {SxProp} from './sx'
 import VisuallyHidden from './_VisuallyHidden'
 
-interface Props extends React.HTMLProps<HTMLLabelElement> {
+type BaseProps = SxProp & {
```
siddharthkp, line 6: (sorry, typescript noob here): Would removing `HTMLProps` throw warning for other html attributes like form or className?

siddharthkp (commented):
Almost there, left a couple a question for clarification

src/_InputLabel.tsx
iansan5653: It would, but those props weren't getting forwarded to the underlying element anyway - so if you passed `className` you'd expect it to work but it wouldn't have actually worked. In that sense, this type was actually misleading. 

In addition, this is a private (internal) component and there wasn't anywhere we were using those types of props internally, so this shouldn't have any effect.

src/utils/character-coordinates.ts
iansan5653: ✅ Done!

siddharthkp (approved):
Feedback implemented, ship it :shipit:

PR #2189, opened 2022-07-26
Make SegmentedControl uncontrolled by default

mperrotti:
The SegmentedControl can now be functional without handling state externally.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


siddharthkp (dismissed):
This feels really nice to use!! ✨ 

A few questions about the API here:

1. When the component can change the selected button by itself (uncontrolled), should the prop `selected` be called `defaultSelected` instead? (like value and defaultValue for input)

2. Do we have a recommended way of firing actions? `onChange` on the parent vs `onClick` on the buttons? We should reflect the preferred approach in the examples.

3. Depending on the answer for 1 + 2, should we have both `selected` and `defaultSelected` similar to `value` + `defaultValue` where `value` is the controlled version that needs to be managed via `onChange`

mperrotti:
> When the component can change the selected button by itself (uncontrolled), should the prop selected be called defaultSelected instead? (like value and defaultValue for input)

Yup, we should add `defaultSelected` as a prop.

---

> Do we have a recommended way of firing actions? onChange on the parent vs onClick on the buttons? We should reflect the preferred approach in the examples.

`onChange` on the parent. I'll update the examples.

---

> Depending on the answer for 1 + 2, should we have both selected and defaultSelected similar to value + defaultValue where value is the controlled version that needs to be managed via onChange

Yes, we should have `selected` and `defaultSelected`, and `selected` will be managed via `onChange` for controlled components.

mperrotti:
Thanks again @siddharthkp! I've addressed your feedback.

siddharthkp (approved):
Testing on local machine, looking good!

quick question: I don't see keyboard navigation behavior from https://github.com/primer/react/pull/2220 in this pull request. 🤔  I hope that's okay?

mperrotti:
@siddharthkp - that doesn't seem right... 🤔

I'll investigate

mperrotti:
@siddharthkp - I'm seeing that behavior in the preview deployment

PR #2190, opened 2022-07-26
Upgrade `user-event` to v14

iansan5653:
We are currently on `@testing-library/user-event@13.1.9`, which is [deprecated and not recommended](https://testing-library.com/docs/ecosystem-user-event). As I've been working on the Markdown editor tests, I've run into a number of issues with this outdated version. Rather than hack around them, I figured I would just update the dependency to v14+, which is far more reliable.

There are some breaking changes here. Notably:

- Instead of firing events off of the import directly, you call events off of the result of `setup`.
- Events always return promises.

In code, this means all tests need to be updated slightly:

```diff
- it('example test', () => {
+ it('example test', async () => {
+   const user = userEvent.setup()
    const target = doSomeSetup()
-   userEvent.click(target)
+   await user.click(target)
  })
```

### Merge checklist

- [x] Added/updated tests
- ~~[ ] Tested in Chrome~~
- ~~[ ] Tested in Firefox~~
- ~~[ ] Tested in Safari~~
- ~~[ ] Tested in Edge~~


src/__tests__/Autocomplete.test.tsx
```diff
@@ -276,49 +278,50 @@ describe('Autocomplete', () => {
       fireEvent.blur(inputNode)
 
       // wait a tick for blur to finish
-      setTimeout(() => {
-        expect(inputNode.getAttribute('aria-expanded')).not.toBe('true')
-      }, 0)
+      await waitFor(() => expect(inputNode.getAttribute('aria-expanded')).not.toBe('true'))
```
iansan5653, line 281: This strategy of waiting a tick was actually failing silently because the assertion would fail after the test completes (in a different 'thread'). Introducing async tests exposed this problem because the failure would actually happen during the next test, because async calls get put at the end of the stack. `await waitFor` is what we wanted for here anyway.

src/__tests__/Autocomplete.test.tsx
```diff
@@ -354,29 +359,29 @@ describe('Autocomplete', () => {
...
+        await user.type(inputNode, 'ze')
         // eslint-disable-next-line github/no-blur
         fireEvent.blur(inputNode)
       }
 
       // wait a tick for blur to finish
-      setTimeout(() => {
-        expect(sortOnCloseFnMock).toHaveBeenCalledTimes(mockItems.length)
-      }, 0)
+      await waitFor(() => expect(sortOnCloseFnMock).toHaveBeenCalled())
```
iansan5653, line 368: Here fixing the `setTimeout` actually exposed an assertion failure that was previously hidden. `Array.sort` does not call the sort function once per element like `filter` or `map` would; a sorting algorithm has to run more than one single pass over the array to fully sort it. In this case, it was calling the sort function 18 times on an array of 10 elements. Rather than test the number of calls, which depends on the JS engine's `sort` implementation, I changed it to just test that at least one call was made.

Lukeghenco (approved):
Quality improvement here for better testing.

colebemis (approved):
This is a nice upgrade. Thank you, @iansan5653 ✨

PR #2197, opened 2022-07-27
Improve the `useCombinedRefs` hook by using a `Proxy`

iansan5653:
> **Note**: This was part of https://github.com/primer/react/pull/2182, but I decided to pull it out into a separate PR to make reviewing easier.

The `useCombinedRefs` hook is intended to provide a single ref that cascades down to several other refs. This is useful when, for example, you have a forwarded ref (that could be an object or a function) and you want to also have a local ref available as well:

```tsx
const Example = forwardRef(({}, forwardedRef) => {
  const combinedRef = useCombinedRefs(forwardedRef)

  return <div ref={combinedRef} />
}
```

When the `combinedRef` is updated, the `forwardedRef` is updated as well. There can even be multiple `forwardedRef`s that will all get updated.

The current implementation uses `useLayoutEffect` to synchronize all the refs on every render. This typically works well, but can have edge cases because refs can update without causing a re-render. This is particularly likely if the ref is being used to hold a value that is not an element or if the ref refers to some element that is a deep child (children could update without re-rendering the parent).

This PR changes the implementation to use a `Proxy` as the target ref instead, so that any time the target ref is updated, all dependent refs are updated immediately. No layout effect is required using this strategy.

We could have used a function as the target ref instead - this would have made change detection easy without use of a `Proxy`. But there are many use cases where the consumer also wants to get the current value of the ref using `.current` and that's not available with a function.

This change brings `useCombinedRefs` up to date with the [latest code](https://github.com/github/memex/blob/6cc009bc900f15d2c170ac76f4d8a907cbd1b8db/src/client/hooks/common/use-combined-refs.ts) used in production in Projects. I also added the hook to `hooks/index` so that we can update that codebase to use the PRC version of the hook and remove some duplication.


Lukeghenco (approved):
This looks good to me (as someone testing the component that uses this right now on the PRs React team), but I don't have real permissions to approve here.

colebemis (approved):
Looks good to me 👍 

It is possible to unit test this hook ensure that it works as we expect? Also, could you add a `patch` changeset to this PR?

iansan5653:
@colebemis I can definitely add unit tests. How would I add a changeset?

colebemis:
@iansan5653 You can click this link:
CleanShot 2022-08-01 at 09 00 30

That will create a markdown file for you where you can describe your changes.

mattcosta7:
Curious about whether we can basically remove this hook entirely (or just wrap the native react implemetation of it)

```
useImperativeHandle(ref, () => internalRef.current)
```

What an imperative handle does

https://github.com/facebook/react/blob/49f8254d6dec7c6de789116e20ffc5a9401aa725/packages/react-reconciler/src/ReactFiberHooks.old.js#L1769-L1860

While combined refs allows passing many refs, where a hook might be a little less ergonomic since it won't handle being called in a loop well, it gets some of this out of this library

iansan5653:
Closing this in favor of https://github.com/primer/react/pull/2204 - thanks @mattcosta7!

Issue #2184, opened 2022-07-25
`SelectPanel` not scrolling with anchoring element


reneexeener:
**Describe the bug**
`SelectPanel` not scrolling with anchoring element when parent containers have different overflow properties. This also happens with `ActionMenu`.

**To Reproduce**
In memex list view:
1. Go to an existing memex project (list view)
2. Click on any downward carets to open the `SelectPanel`
3. Start scrolling the table
4. `SelectPanel` does not scroll with the anchoring element that is the table cell

In bulk add side panel:
1. Go to an existing memex project and click on the `+` icon in the omnibar at the bottom of the page
2. Click on the repository selector to open the `SelectPanel`
3. Start scrolling the table
4. `SelectPanel` does not scroll with the repository selector button

In Project settings (implements `ActionMenu`):
1. Make sure you have admin permissions and go to an existing project
2. Click on the three dots on the top right corner -> Settings -> go to bottom of the page
3. Click on the button next to 'Visibility'
4. Start scrolling the page
5. `ActionMenu` does not scroll with the button

**Expected behavior**
`SelectPanel` and `ActionMenu` should scroll correctly with their anchoring elements.

**Screenshots**
![selectpanel](https://user-images.githubusercontent.com/13259331/180819503-af2f8e3c-3de4-4287-b02a-f8d514508724.gif)

**Desktop (please complete the following information):**
 - OS: macOS Montery 12.4
 - Browser: Chrome 
 - Version 103.0.5060.114

**Additional context**
After doing some deductions with @siddharthkp, we suspect that the scrolling problem is caused by wrapping `SelectPanel` or `ActionMenu` in nested containers with different overflow properties, which causes the `getClippingRect` function to find the wrong parent node ([code](https://github.com/primer/behaviors/blob/main/src/anchored-position.ts#L195)).

iansan5653:
I've looked into this a bit before because we see the same thing with [`InlineAutocomplete`](https://github.com/primer/react/pull/2157) suggestions when you scroll the Markdown editor textarea in the side panel.

The cause is that all of our overlays render into a `Portal` at the document root, so the portals all are be positioned relative to the document root. In this case, the table grid is scrolling but the document is not. The only way to fix this without changing Primer is to put a portal inside the grid and render all in-grid overlays into that portal. That would be a nightmare to maintain and we'd have to do it for every single scrollable container in the app. It's actually impossible for the Markdown editor since you can't put a portal inside of a `textarea` element.

Here's an example: 

[...]

PR #2200, opened 2022-07-28
AnchoredOverlay + useAnchoredPosition: close overlay when menu goes out of view 

siddharthkp:
#### Changes

1. Adds an optional argument to `useAnchoredPosition` to observe changes to anchor element: `observe`
2. AnchoredPosition now uses  useAnchoredPosition with `observe` which means overlay re-positions based on changes to anchor's position (scroll, resize, etc)
3. AnchoredOverlay closes when anchor goes out of view (includes ActionMenu and SelectPanel, but not Autocomplete)

- Fixes https://github.com/primer/react/issues/2184

video:

https://user-images.githubusercontent.com/1863771/187722377-b98399fd-61ce-4065-b061-207e0674e5cf.mov

#### Notes for reviewer: 

1. Have kept the change in primer/react and not upstreamed to primer/behaviours because I'm not sure about the long term tradeoffs yet.
2. Using [reach/observe-rect](https://github.com/reach/observe-rect) which calls getBoundingRect on every requestAnimationFrame. Added 2 performance optimisations: 
  2.1 Observer is only initiated when overlay is open
  2.2 Callback is debounced (16ms) to make sure we're not updating state too fast
3. Considered alternate approach of adding scroll listeners on every parent up the tree, but decided against it because it attaches events on application elements

Checklist:
- [x] Check if new dependency works across build systems (library has support for [cjs, esm and umd](https://unpkg.com/browse/@reach/observe-rect@1.2.0/dist/))
- [ ] Is [this an okay way to load lodash](https://github.com/primer/react/pull/2200/files#diff-b302f9f2813f17720fbfab313cac4c110b2929452f2ecc1091d78d30e04d5f1bR3)?
- [ ] Check if there is a performance regression because of observer (using debounce but need to test with memex)
- [ ] Check if this approach is accessible (a11y discussion on [slack for github staff](https://github.slack.com/archives/C0FSWLQ0Y/p1661954660218019))
- [ ] Check if the type changes make sense for OnCloseGesture (maybe remove them from this PR)
- [ ] Check if Autocomplete should also add observe with useAnchoredPosition


owenniblock:
[Slack discussion around the SelectPanel functionality here](https://github.slack.com/archives/C0FSWLQ0Y/p1661954660218019)

inkblotty:
👋 Hi @siddharthkp - Would you cc @primer/accessibility-reviewers when this is out of Draft? 🙏

siddharthkp:
cc @primer/accessibility-reviewers for review :)

smockle:
> cc @primer/accessibility-reviewers for review :)

Caption: Screen recording of an open menu closing after its associated cell bounces back into view post-scrolling:
https://user-images.githubusercontent.com/3104489/189673468-5aec1f77-cec5-40a9-9cee-e4d9466040c0.mov

smockle:
Other menus on github.com remain open as the page is scrolled, even if they are scrolled out of view. Here is the issue comment menu, for example:

Caption: Screen recording showing that the issue comment three-dot menu remains open even after it scrolls out-of-view:

https://user-images.githubusercontent.com/3104489/189692010-67f98bba-4f1f-473c-9099-8ae086eb3aaa.mov

siddharthkp:
> Other menus on github.com remain open as the page is scrolled, even if they are scrolled out of view. Here is the issue comment menu, for example (screen recording below). I’d expect all menus to behave consistently.

Yes! that was my concern as well. The suggestion to close it came after my first iteration where it stays open:

Description: There is a sticky header above the table so the select panel awkwardly floats on when the table cell is "tucked under" the header. 

https://user-images.githubusercontent.com/1863771/189927290-f14caf6b-46dc-4983-8501-ddd7a888f28d.mov

(in the video, ignore the select panel jumping around, that's not the recommendation)

lesliecdubs:
👋🏻 @siddharthkp just doing some triage on the PRC board and this one has been in "ready for review" for a while. Is this still active?

siddharthkp:
> Is this still active?

We still need it 😅 but I haven't touched this in few weeks since the a11y feedback

PR #2204, opened 2022-08-01
Replace `useCombinedRefs` with `useRefObjectAsForwardedRef`

iansan5653:
The `useCombinedRefs` hook only has one real use-case: combining forwarded refs with ref objects so that the value of the ref can be accessed with `ref.current`. Elements can only take one ref, hence we 'combine' the refs into a single ref.

In #2197 I attempted to improve the hook to make it more useful. However, @mattcosta7 [noted](https://github.com/primer/react/pull/2197#issuecomment-1201447467) that we could completely replace this one use case with a `useImperativeHandle` call, which is absolutely correct. By wrapping that call in a simple hook and replacing `useCombinedRefs` with the new hook, we reach a number of advantages:

- It's easier to use (and more importantly, harder to use incorrectly)
- It's simpler and less bug-prone
- The intention is more explicit
- ESLint only recognizes `useRef` and forwarded refs as refs. By not creating a new ref through a hook, we avoid having to pass ref objects as dependencies to effects & memos

So this PR supersedes #2197, and I feel that it's a major improvement.

> **Note**: I don't think this is a breaking change because this hook is not exported through `hooks/index`.

### Merge checklist

- ~~[ ] Added/updated tests~~
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/hooks/useRefObjectAsForwardedRef.ts
```diff
@@ -0,0 +1,12 @@
+import {ForwardedRef, RefObject, useImperativeHandle} from 'react'
+
+/**
+ * Use a ref object as the imperative handle for a forwarded ref. This can be used to
+ * synchronize the ref object with the forwarded ref and allow local access the reference
+ * instance with `.current`.
+ *
+ * **NOTE**: The `refObject` should be passed to the underlying element, NOT the `forwardedRef`.
+ */
+export function useRefObjectAsForwardedRef<T>(forwardedRef: ForwardedRef<T>, refObject: RefObject<T>): void {
```
iansan5653, line 10: I'm open to suggestions for naming this. I think this name works but it could maybe be clearer. Other options I considered include:
- `useForwardedRefObject`
- `useForwardedRefAsRefObject`
- `useRefObjectAsHandle`

colebemis (approved):
Clean ✨ I like it!

PR #2207, opened 2022-08-02
Deployment previews for forks

siddharthkp:
Right now, forks don't get deployment previews because secrets/tokens can not be shared with forks safely.

```
Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
```

image

However, there's an event called `pull_request_target`, which runs github actions on the target repo (primer/react) but runs the code from the source repo (example: user/primer-react-fork), this is the perfect solution for 99% of contributions (1% is when you are making changes to github actions workflows)

Reference: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows

On forks, both `pull_request` and `pull_request_target` events are fired, so we can safely skip the `pull_request` based on the source repository of the pull request.

Tested on this pull request from a fork: https://github.com/primer/react/pull/2210

image


.github/workflows/deploy_preview_forks.yml
```diff
@@ -0,0 +1,23 @@
+name: Deploy (fork)
+on:
+  pull_request_target:
```
siddharthkp, line 3: only runs on forks

.github/workflows/deploy_preview.yml
```diff
@@ -9,7 +9,7 @@ permissions:
 
 jobs:
   deploy-preview:
-    if: ${{ github.repository == 'primer/react' }}
+    if: ${{ github.event.pull_request.head.repo.full_name == 'primer/react' }} # source repository for pull_request
```
siddharthkp, line 12: if it's a fork, skip the default deploy

.github/workflows/deploy_preview_forks.yml
colebemis: This is helpful context. Could you add this as a code comment to the yml file?

.github/workflows/deploy_preview.yml
colebemis: Could you also add this as a code comment to the yml file?

colebemis (approved):
Love it! This will make it so much easier to review and merge outside contributions. Thanks, @siddharthkp 💖

.github/workflows/deploy_preview.yml
siddharthkp: done :)

PR #2212, opened 2022-08-02
PageLayout.Pane: Add `sticky` prop

colebemis:
## Motivation

Pages with a lot of content (examples: files changed tab, settings page, etc.) might want to display a sticky side pane that stays in view as you scroll through the content. This is not currently supported by the `PageLayout` component

## Summary

This PR adds a `sticky` prop to the `PageLayout.Pane` component that provides a convenient way for consumers to make side panes sticky:

```diff
<PageLayout>
- <PageLayout.Pane>...</PageLayout.Pane>
+ <PageLayout.Pane sticky>...</PageLayout.Pane>
  <PageLayout.Content>...</PageLayout.Content>
</PageLayout>
```

Here's a quick demo of the sticky behavior:

https://user-images.githubusercontent.com/4608155/182470609-8ea88a18-af50-4ba3-be6f-22d76595577a.mp4

## Implementation

The sticky is more complex than you might expect so I recorded a quick demo to explain why the complexity exists and how it works: 

🎥 🍿  https://share.cleanshot.com/O9uSpc

### Merge checklist

- [ ] ~~Added/updated tests~~
  - Will add in a separate PR: https://github.com/primer/react/pull/2224 (Thanks @pksjce 🙌 )
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge
- [x] Test in Projects


src/PageLayout/useStickyPaneHeight.ts
```diff
@@ -0,0 +1,65 @@
+import React from 'react'
+import {useInView} from 'react-intersection-observer'
+
+/**
+ * Calculates the height of the sticky pane such that it always
+ * fits into the viewport even when the header or footer are visible.
+ */
+// TODO: Handle sticky header
```
colebemis, line 8: @vdepizzol How you feel about adding a prop called `stickyTop` (open to other name ideas) that would allow you to push the side pane down to make room for sticky headers if necessary?

src/PageLayout/PageLayout.tsx
```diff
@@ -55,9 +60,27 @@ const Root: React.FC<React.PropsWithChildren<PageLayoutProps>> = ({
...
+        contentTopRef,
+        contentBottomRef
+      }}
+    >
+      <Box
+        style={{
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore TypeScript doesn't know about CSS custom properties
+          '--sticky-pane-height': stickyPaneHeight
+        }}
```
vdepizzol, line 81: @colebemis since this value needs to update in real time while the user is scrolling, I wonder if we can measure how this jsx rendering (where you apply a variable to a `<Box style={{...}}` compares in terms of performance to updating the CSS custom property directly using [setProperty](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty) ([example](https://css-tricks.com/updating-a-css-variable-with-javascript/)).

src/PageLayout/useStickyPaneHeight.ts
```diff
@@ -0,0 +1,65 @@
...
+export function useStickyPaneHeight() {
+  // Default the height to the viewport height
+  const [height, setHeight] = React.useState('100vh')
```
vdepizzol, line 11: @colebemis we should use `100vh` by default, but if the browser supports 100dvh, we should use that instead. `dvh` units take into consideration any dynamic browser viewport resizing. For example:

image

([Source](https://twitter.com/jensimmons/status/1499441043930062854))

This is how I define it in the Primer CSS [implementation](https://github.com/primer/css/blob/main/src/layout/page-layout.scss#L114-L124):

```scss
.PageLayout-pane--sticky {
  position: sticky;
  top: 0;
  max-height: 100vh;
  overflow: auto;
  scrollbar-width: thin;

  @supports (max-height: 100dvh) {
    max-height: 100dvh;
  }
}
```

(Also note I'm using `max-height`, not `height`. Not sure how that'd compare to your implementation)

src/PageLayout/useStickyPaneHeight.ts
vdepizzol: I think it makes sense. One thing to keep in mind is that this needs to remain functional as we move to `rem` for all of our UI.

src/PageLayout/PageLayout.tsx
colebemis: @vdepizzol and I tried this out and it seems like the performance is roughly the same.

src/PageLayout/useStickyPaneHeight.ts
colebemis: Sounds good. I'll tackle this in a separate PR. Created an issue for it here: https://github.com/github/primer/issues/1200

src/PageLayout/useStickyPaneHeight.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  // Create intersection observers to track the top and bottom of the content region
+  const [contentTopRef, contentTopInView, contentTopEntry] = useInView()
```
dgreif, line 14: Looks like `useInView` takes a margin in the options, so it should be easy to add `stickyTop` down the road

dgreif (commented):
Looks good! Turned out very clean doing it the "react" way 😄. Glad the content sentinels worked out 🎉

src/PageLayout/useStickyPaneHeight.ts
colebemis: Since dynamic viewport units are only supported in mobile browsers right now, and we disable sticky behavior at narrow viewports, I'm going to postpone using `dvh` for now

src/PageLayout/PageLayout.tsx
```diff
@@ -254,12 +277,14 @@ const Content: React.FC<React.PropsWithChildren<PageLayoutContentProps>> = ({
...
 }) => {
   const isHidden = useResponsiveValue(hidden, false)
+  const {contentTopRef, contentBottomRef} = React.useContext(PageLayoutContext)
   return (
     <Box
       as="main"
       hidden={isHidden}
       sx={merge<BetterSystemStyleObject>(
         {
+          position: 'relative',
```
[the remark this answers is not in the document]
colebemis: Good catch, thank you! I just removed `position: relative` 👍

src/PageLayout/useStickyPaneHeight.ts
vdepizzol: @colebemis Safari on iPad also has that behavior.

src/PageLayout/useStickyPaneHeight.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    const topRect = contentTopEntry?.target.getBoundingClientRect()
+    const bottomRect = contentBottomEntry?.target.getBoundingClientRect()
+
+    const topOffset = topRect ? Math.max(topRect.top, 0) : 0
+    const bottomOffset = bottomRect ? Math.max(window.innerHeight - bottomRect.bottom, 0) : 0
+
+    // Safari's elastic scroll feature allows you to scroll beyond the scroll height of the page.
+    // We need to account for this when calculating the offset.
+    const overflowScroll = Math.max(window.scrollY + window.innerHeight - document.body.scrollHeight, 0)
```
pksjce, line 31: Do you think this should be calculated on the scrollcontainer? Or is the `PageLayout` always meant to be used as a top level component?

pksjce (commented):
I don't have a lot of comments regarding implementation. `react-interaction-observer` seems like a nifty little library. I wonder why the size-limit did not alert us of the additional 4kb 🤔 

Anyways, noticed that this seemed like a great opportunity for some interactive tests!
Here's a PR with a couple of tests on the storybooks
https://github.com/primer/react/pull/2224

For further info check my ADR https://github.com/primer/react/pull/2223/files

src/PageLayout/useStickyPaneHeight.ts
colebemis: PageLayout should always be the top-level component. But this is a good consideration. I'll keep it as `window` for now and adjust it if needed.

src/PageLayout/useStickyPaneHeight.ts
colebemis: Ah good point, @vdepizzol. I'll see how we can use `dvh` units with this implementation.

pksjce (approved):

src/PageLayout/useStickyPaneHeight.ts
colebemis: Actually, it looks like memex uses PageLayout inside a scroll container that isn't `document.body`. I'll fix this in this PR

siddharthkp: > PageLayout should always be the top-level component.

Can you say more about this, would the top level navigation be inside PageLayout.Header? 

Right now, I think issues has top level navigation and page layout as siblings, so there's some margin on top

src/PageLayout/useStickyPaneHeight.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    setHeight(`calc(100vh - ${topOffset + bottomOffset - overflowScroll}px)`)
+  }, [contentTopEntry, contentBottomEntry])
```
siddharthkp, line 34: I like this optimisation is a lot!

siddharthkp (approved):
Suggested some improvements for the docs, approving in advance 👍

src/PageLayout/useStickyPaneHeight.ts
colebemis: Yeah, I was wrong about PageLayout always being top-level. @joshblack and I just updated the implementation to work if PageLayout is nested in a scroll container: https://github.com/primer/react/pull/2212/commits/52e174fd535c5e29c552e685003a3cac7c578137

docs/content/PageLayout.mdx
```diff
@@ -147,6 +147,25 @@ See [storybook](https://primer.style/react/storybook?path=/story/layout-pagelayo
 </PageLayout>
 ```
 
+### With a sticky pane
+
+```jsx live
+<PageLayout>
```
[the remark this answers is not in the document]
colebemis: Great idea! Just updated ✅

docs/content/PageLayout.mdx
```diff
@@ -310,6 +329,7 @@ See [storybook](https://primer.style/react/storybook?path=/story/layout-pagelayo
     defaultValue="'medium'"
     description="The width of the pane."
   />
+  <PropsTableRow name="sticky" type="boolean" defaultValue="false" description="Whether the pane is sticky." />
```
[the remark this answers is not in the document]
colebemis: Done ✅

src/PageLayout/useStickyPaneHeight.ts
colebemis: This PR is getting a little too big. I'll break it into a new PR. Just made a tracking issue: https://github.com/github/primer/issues/1211

src/PageLayout/PageLayout.stories.tsx
```diff
@@ -7,237 +7,443 @@ import {PageLayout} from './PageLayout'
 
 const meta: Meta = {
   title: 'Layout/PageLayout',
-  component: PageLayout,
+  // component: PageLayout,
```
colebemis, line 10:
```suggestion
```

PR #2214, opened 2022-08-04
ADR 008: Lowering the barrier to entry for experimental components

siddharthkp:
[See rendered markdown](https://github.com/primer/react/blob/adr-experimental-package/contributor-docs/adrs/adr-008-experimental-components.md)


contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,62 @@
...
+
+The monolith already has a place where reusable react components live, however these cannot be shared by react projects outside the monolith.
+
+We can publish "react-shared" as a npm package but that would introduce additional work for non-monolith projects and components built within the context of the monolith might not work in other projects without making them more flexible.
+
+2. Experimental components should live in a new repository `github/react-shared` or `github/primer-react-candidates` instead of `primer/react-candidates`
+
+This is not a bad option, it helps in sharing components between projects outside of monolith as well and the ownership and responsibility of maintenance is shared by multiple teams.
+
+While primer would be one of the teams maintaining this repository, the lack of a primary "owner" might mean that this repository only gets additions and the maintenance of components gets neglected.
```
siddharthkp, line 62: this is a bit speculative, it might not be a problem at all.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,62 @@
...
+- Components in drafts are companied by a design spec and go through the process of accessibility reviews. A component can check multiple boxes in the component lifecycle.
+- The contract with consumers is that you are opting into a component in which we don't have 100% confidence yet.
+
+[^1]: Calling this a prototype and not experimental, because that already means something else in the maturity lifecycle.
+
+---
+
+#### Problems with drafts
+
+1. **Incompatible with semver**: Because the expectation with drafts is that we are not 100% confident about the API or structure of a component, we slip in breaking changes to draft components as minor releases to primer/react. On the consumer side, this makes every upgrade, even a patch, a potentially breaking upgrade. So, you are effectively opting out of your semver.
```
siddharthkp, line 28: I think we have been very careful about not making big breaking changes without also helping with the change.

This is likely to become more difficult because of more projects using primer/react + more contributors upstreaming components

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,70 @@
...
+
+[^1]: Calling this a prototype and not experimental, because that already means something else in the maturity lifecycle.
+
+&nbsp;
+
+#### Problems with drafts
+
+1. **Incompatible with semver**: Because the expectation with drafts is that we are not 100% confident about the API or structure of a component, we slip in breaking changes to draft components as minor releases to primer/react. On the consumer side, this makes every upgrade, even a patch, a potentially breaking upgrade. So, you are effectively opting out of your semver.
+
+1. **Confusion about maturity**: Because of the generic name, "drafts" causes a lot of confusion, especially with "experimental". It is a prototype or experiment, but NOT in the way the [component lifecycle]([https://primer.style/contribute/component-lifecycle]) defines "Experimental" maturity.
```
lesliecdubs, line 32:
```suggestion
2. **Confusion about maturity**: Because of the generic name, "drafts" causes a lot of confusion, especially with "experimental". It is a prototype or experiment, but NOT in the way the [component lifecycle]([https://primer.style/contribute/component-lifecycle]) defines "Experimental" maturity.
```

I think the automatic numbering in this list is off.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,70 @@
...
+
+3. To support different conventions and processes for experimental components, we should keep them in a new repository `primer/react-candidates`
+
+(other name ideas: primer/react-experimental, primer/react-candidates, primer/react-proposals, primer/react-contrib)
+
+&nbsp;
+
+#### Risks:
+
+The maintenance of components while they are still candidates (bugs, a11y remedial, etc.) will be a new parallel workstream for the primer team. With our current team size, we might not be able to give it the required attention.
```
colebemis, line 54: Should we add some guidance about how code-owners should work for candidate components? Maybe the team that created the component owns it (via CODEOWNERS) until the component is upstreamed into Primer?

colebemis (commented):
Thanks for writing this, @siddharthkp! Left a few comments and questions :)

lesliecdubs:
Thanks for getting this written up, @siddharthkp!

**ADR decision maker:** @mperrotti 
**Decision deadline:** EOD Monday, 15 August. This deadline is a bit quicker than usual so we can unblock active contributions from the Memex team. Mike, if this is too quick a timeline for you, please let me know 🙏🏻 

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. **Need for experimental candidates**: As more teams build React components and look to upstream reusable components, it is natural to look at drafts as the place to put them. However, the barrier to entry into drafts isn't low (because we still follow design reviews, dev conventions, API reviews, etc.)
+
+Upstreaming a component that was built in the context of a product requires re-evaluating the baked-in assumptions and most likely a change in the API to make it useful for other contexts.
+
+&nbsp;
+
+## Decision
+
+1. "Experimental" components, as defined by the component lifecycle, should not be part of semantically versioned npm package for `@primer/react`.
```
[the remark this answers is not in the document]
alebourne: Would the "Experimental" components become the same as the current "Drafts"? or will they still be "Experimental" but live outside of Primer?

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: Letting [rendered markdown](https://github.com/primer/react/blob/adr-experimental-package/contributor-docs/adrs/adr-008-experimental-components.md#problems-with-drafts) take care of that :)

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
siddharthkp: > If we move "experimental" components out of @primer/react into a new package, would we stop using the drafts directory?

yes. we would only move them to `@primer/react` package once we can commit to non-breaking changes

> Would the "Experimental" components become the same as the current "Drafts"? or will they still be "Experimental" but live outside of Primer?

I'd like to detach them from current "drafts" because "draft components" go through the primer workflow of design, API, sometimes a11y review. "Experimental" components should have a lower barrier to entry.

From: https://github.com/primer/react/pull/2214/files#diff-5e30a807664dafbee9f227901a2459809a2c2226e183880f8bf6e05b1f52a7ceR20-R21:

> - A draft has experimental API, but follows the development conventions of primer/react
> - Components in drafts are accompanied by a design spec and go through the process of accessibility reviews. A component can check multiple boxes in the component lifecycle.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,70 @@
...
+
+2. Each experimental component should be independently versioned with semver. Breaking changes should be tagged as major versions.
+
+3. To support different conventions and processes for experimental components, we should keep them in a new repository `primer/react-candidates`
```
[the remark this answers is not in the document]
siddharthkp: > Could we publish a @primer/react-candidates (or whatever we want to call it) from the primer/react repo?

Sorry, some nitpicking first 😅

The key is that 
1) experimental components are not part of the npm package `@primer/react`
2) each component is independently versioned like `@primer/react-commentbox`, not in a combined package like `@primer/react-candidates`

We can publish these components independently while keeping them in primer/react repository (repo, not npm package) or from another repository like primer/react-candidates or both depending on the component.

I personally think it's harder to maintain different dev conventions/practices and code ownership from a single repository, but I can go either way on this.

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: Yes, i think that's a great point!

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
pksjce: Maybe we update the expectation for drafts components esp regards to. Perhaps add an ETA for drafts to move to main component directory. Have a pre-alpha checklist that does not include accessibility review?

pksjce (commented):
I think this is a good direction for primer-contributions in the future. While we accept ownership of these new components, we do want them to adhere to conventions that we are strictly following in our team. In case of urgent upstreaming, we will need to come up with some compromise. And I believe this is a good one. Contributors will still have the opportunity to contribute directly to `primer-react`, provided they follow our guidelines and component sizes are minimal. 
In the specific case of this ADR ie a markdown editor component, I would want our design team/ally teams to be involved in it. It could also happen that we can breakdown this large component into smaller parts and have them in primer-react. Either way, a buffer for incoming components with shared ownership between product teams and primer-react is a great idea IMO

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
inkblotty: I do prefer using the `candidates` name instead of `shared`, because it contributes to the idea that it's not fully adopted within Primer. 👍 

Am I right in thinking that we don't want a `@primer/react/candidates` in order to mitigate the versioning issue? I do agree with @colebemis  that it'd be easier to publish candidates with the main package if not.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,70 @@
...
+&nbsp;
+
+#### Problems with drafts
+
+1. **Incompatible with semver**: Because the expectation with drafts is that we are not 100% confident about the API or structure of a component, we slip in breaking changes to draft components as minor releases to primer/react. On the consumer side, this makes every upgrade, even a patch, a potentially breaking upgrade. So, you are effectively opting out of your semver.
+
+1. **Confusion about maturity**: Because of the generic name, "drafts" causes a lot of confusion, especially with "experimental". It is a prototype or experiment, but NOT in the way the [component lifecycle]([https://primer.style/contribute/component-lifecycle]) defines "Experimental" maturity.
+
+1. **Need for experimental candidates**: As more teams build React components and look to upstream reusable components, it is natural to look at drafts as the place to put them. However, the barrier to entry into drafts isn't low (because we still follow design reviews, dev conventions, API reviews, etc.)
+
```
inkblotty, line 35: I'm not sure if this is the right place to put this. I think a little more detail here could be helpful.

```suggestion

Feature teams using Primer may not be intimately familiar with Primer component lifecycle criteria or have the time and/or knowledge to meet it. Often, feature teams create their own components outside of Primer if they find it too difficult to contribute. This leads to lower Primer adoption and conflicting versions of similar functionality throughout GitHub products.

We want to enable feature teams to contribute to Primer. We also want to make it clear which contributions haven't been evaluated by our high standards for Primer components.

```

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,70 @@
...
+Upstreaming a component that was built in the context of a product requires re-evaluating the baked-in assumptions and most likely a change in the API to make it useful for other contexts.
```
inkblotty, line 36: I agree with this statement, but it doesn't feel like `candidates` fixes this problem. It seems like a persistent one.

inkblotty (commented):
Some thoughts from a sister-team to Primer. I love that we're talking about making contributions easier 🙏

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: I think @pksjce articulated it really well:

> I would want our design team/a11y teams to be involved in it. It could also happen that we can breakdown this large component into smaller parts and have them in primer-react. Either way, a buffer for incoming components with shared ownership between product teams and primer-react is a great idea IMO

We need _a place_ where experimental components live that should not be `@primer/react/*` (because of the semver issues mentioned)

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: I really like this, can I edit pieces from this and add to the doc?

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
siddharthkp: > I do prefer using the candidates name instead of shared, because it contributes to the idea that it's not fully adopted within Primer. 👍

💯 

> Am I right in thinking that we don't want a @primer/react/candidates in order to mitigate the versioning issue?

Yep.

I think I haven't done a good job at distinguishing the strategy for npm package and code repository. 

Right now, we have 1 code repository [github.com/primer/react](http://github.com/primer/react) and 1 npm package: [`@primer/react`](https://www.npmjs.com/@primer/react).

For the npm package, the proposal is publish experimental components outside of the npm package for `@primer/react` to keep the confidence/trust that there are no breaking changes in minor or patch releases for `@primer/react`.

There are 2 approaches we can take:
1. (proposed in this ADR) Each experimental component is published as it's own npm package, example: `@primer/react-commentbox` (versioned individually)
2. Publish all experimental components together in a package `@primer/react-candidates` (Pro: One package to install and import from, less dependency management overhead. Con: A breaking change in one component is a breaking/major release for the whole package, can't update one component without making changes for the rest as well. Relevant because we expect more breaking changes in experimental components)

For the code repository, there are 2 approaches we can take:
1. Keep experimental components in [github.com/primer/react](http://github.com/primer/react), publish two or multiple packages from one repository. (Less overhead, one place where all the code lives)
2. (proposed in this ADR) Keep all experimental packages in one new repository [github.com/primer/react-candidates](http://github.com/primer/react-candidates). (this repository can have different conventions and processes with shared ownership between product teams and primer)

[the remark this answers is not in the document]
siddharthkp: I think I haven't done a good job at distinguishing the strategy for npm package and code repository. So here we go:

Right now, we have 1 code repository [github.com/primer/react](http://github.com/primer/react) and 1 npm package: [`@primer/react`](https://www.npmjs.com/@primer/react).

For the npm package, the proposal is publish experimental components outside of the npm package for `@primer/react` to keep the confidence/trust that there are no breaking changes in minor or patch releases for `@primer/react`.

There are 2 approaches we can take:
1. (proposed in this ADR) Each experimental component is published as it's own npm package, example: `@primer/react-commentbox` (versioned individually)
2. Publish all experimental components together in a package `@primer/react-candidates` (Pro: One package to install and import from, less dependency management overhead. Con: A breaking change in one component is a breaking/major release for the whole package, can't update one component without making changes for the rest as well. Relevant because we expect more breaking changes in experimental components)

For the code repository, there are 2 approaches we can take:
1. Keep experimental components in [github.com/primer/react](http://github.com/primer/react), publish two or multiple packages from one repository. (Less overhead, one place where all the code lives)
2. (proposed in this ADR) Keep all experimental packages in one new repository [github.com/primer/react-candidates](http://github.com/primer/react-candidates). (this repository can have different conventions and processes with shared ownership between product teams and primer)

contributor-docs/adrs/adr-008-experimental-components.md
inkblotty: Ah let me clarify:

I think line 36 is 100% true, but it belongs in its own section. Maybe like "continued risks" or just a section on "this proposal doesn't solve this problem" or something.

contributor-docs/adrs/adr-008-experimental-components.md
inkblotty: Please do!

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: I think I'm missing the point, can you say more about this?

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
siddharthkp: > Maybe we update the expectation for drafts components esp regards to. Perhaps add an ETA for drafts to move to main component directory.

Can you say more about this, who is the ETA for?

siddharthkp:
@pksjce Thank you for the articulation! This is really helpful, I'm going to start quoting that 😄

contributor-docs/adrs/adr-008-experimental-components.md
inkblotty: Let's Slack if I'm still being obtuse 😅

siddharthkp: Ohhh, nah you're totally right. That part should be under "Incompatible with semver" instead of "Need for experimental components"

siddharthkp: I was trying to make the point that upstreamed components will go through more API changes than (our current usage of) draft components

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
siddharthkp: Added more clarity to `Decision` and `Other options considered` based on the comment above ^

mperrotti:
@siddharthkp is going to make some updates and get feedback from other teams with similar challenges.

Monday Aug 15 is going to be too soon to freeze this decision cc @lesliecdubs

siddharthkp:
Update! 

I have updated the location where the source code of experimental components would live from `github.com/primer/react-candidates` to `https://github.com/github/github/tree/master/app/assets/modules/react-shared`

```diff
### Decision
- 3. code repository: To support different conventions and processes for experimental components and convey shared ownership between primer and product teams, we should keep them in a new repository `github.com/primer/react-candidates`
+ 3. The code for experimental components should live in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) and published as individual packages as suggested in point 2.
+
+  The monolith already has a place where reusable react components live. This has the lowest barrier to entry for folks working in the monolith.
+
+  Acknowledged risk: It would introduce some additional work for non-monolith projects when upstreaming reusable components. We should work on improving the development and publishing workflow for react-shared.
```

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: Hi! 

After collecting feedback from the team, I have updated decision no. 3 which is relevant to this conversation

https://github.com/primer/react/pull/2214#issuecomment-1212252442

colebemis:
Some notes from my conversation with @dgrief about publishing the react-shared directory as a package the memex can consumer:

- Overall, Dusty thinks publishing react-shared seems like reasonable solution to our current problems and prefers that approach to the alternatives (i.e., creating a new repo)
- We could/should publish the package to the GitHub Package registry (instead of npm). It would be easier to keep the package private and we already have the infrastructure set up for it in dotcom (i think)
- Ownership in dotcom is already handled via CODEOWNERS so we wouldn’t have to set that up ourselves
- Memex bundles all their dependencies so we run the risk of bundling react-shared components twice in both memex and dotcom. However, this is an existing issue. react, react-dom, and @primer/react are already being duplicated between memex and dotcom so this doesn’t seem like a blocker

— [Slack thread](https://github.slack.com/archives/C02NUUQ9C30/p1660243078505589)

lesliecdubs:
Thanks for the great discussion around this proposal!

@mperrotti I'll push out the decision deadline to Wednesday 24 August, although please feel free to complete the decision-maker duties prior to that date if you're able. Please ping me again if more time is needed.

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
iansan5653: As far as whether to put it in a new vs same repo: It seems like a major consideration here is whether we want these experimental components to be open-source. Are we trying to experiment publicly, or would we rather keep the experimenting internal and only publicize polished work? If we want closed-source candidates, then we need a new, private, repo.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Risks:
+
+The maintenance of components while they are still candidates (bugs, a11y remedial, etc.) will be a new parallel workstream for the primer team. With our current team size, we might not be able to give it the required attention.
+
+&nbsp;
+
+#### Other options considered
+
+1. Experimental components should live in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared)
```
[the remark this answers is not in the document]
iansan5653: > Depending on how hard it would be to share those components with Memex

I think that would be _very_ difficult without moving Memex back into the monolith, unfortunately

contributor-docs/adrs/adr-008-experimental-components.md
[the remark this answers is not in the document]
siddharthkp: > Are we trying to experiment publicly, or would we rather keep the experimenting internal and only publicize polished work?

I don't think we have considered that as a factor because we don't mind putting unpolished work in primer/react repo (open source)

There are times when a component might give a way a secret feature we are working on, but that's an exception not a rule

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,77 @@
...
+
+&nbsp;
+
+## Decision
+
+1. npm package: "Experimental" components, as defined by the component lifecycle, should not be part of semantically versioned npm package for `@primer/react`.
+
+2. npm package: Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@github/react-commentbox`
+
+3. The code for experimental components should live in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) and published as individual packages as suggested in point 2.
```
iansan5653, line 52: I think this will be a huge improvement, thank you for working on this!

I do have a few concerns with placing these components inside gh/gh instead of in their own repo:

1. Tooling
  - Will we be able to setup Storybook and linting with this strategy?
  - Will we be able to setup a staging environment for PRs?
3. Publishing
  - Will it be automatic, and if so how will we set the versions?
  - Are we going to have to use the dotcom deploy queue to publish these packages, and if we don't, how will we ensure that the code gets merged into `master` (due to the deploy-then-merge setup)?
  - How do we ensure that, when code in this directory changes, a new version gets published instead of just merging it to `master`?
5. Usage
  - When code inside the monolith uses these components, will they import directly or will they install the package?
  - If they import directly, how can they take advantage of semver to manage breaking changes?

Personally it feels like a lot of these problems would be solved with a new repository that has its own tooling configured. Then everyone everywhere will depend on the components using the NPM packages. And development would be easier in a smaller repo with a tailored Codespaces setup.

siddharthkp: All very good questions! I appreciate it!

> 1. Tooling
>   Will we be able to setup Storybook and linting with this strategy?
>   Will we be able to setup a staging environment for PRs?

Yes and yes :) Improving tooling would be the first step if we go forward with this ADR. I'm hoping we can utilise [primer/react-scripts](https://github.com/primer/react-scripts) work for the react-shared

> 2. Publishing
>   Will it be automatic, and if so how will we set the versions?
>   Are we going to have to use the dotcom deploy queue to publish these packages, and if we don't, how will we ensure that the code gets merged into master (due to the deploy-then-merge setup)?
>  How do we ensure that, when code in this directory changes, a new version gets published instead of just merging it to master?

I have some opinions but I don't think I can make that decision, I'm happy to collab on a followup ADR in github/github for these questions with someone who works on github/github regularly.

> 3. Usage
> When code inside the monolith uses these components, will they import directly or will they install the package?

I want to say yes, that's how it is setup currently and has the lowest barrier to contribution.

> If they import directly, how can they take advantage of semver to manage breaking changes?

There is a risk of introducing breaking changes for non-monolith projects and not realising it. I haven't spent enough time thinking about that, we should address it in the ADR about publishing react-shared

iansan5653: Thanks!

It feels like it's going to be very hard to avoid a complex situation where we have some code importing and some using the published package. It will be very challenging to ensure that importing code (teams inside the monolith) must always publish all the changes they make, particularly if there is no project actively using the published package version. It will also be very difficult for non-monolith teams to make changes if there are in-monolith teams importing the code, because they will have to address all breaking changes in other projects before they can publish anything.

Additionally - if the components are built inside the monolith, what's to stop them from importing code that also exists in the monolith, outside the react-shared folder? If they do that, how will we publish the code without blowing up the package to include _all_ monolith JS/TS code?

Maybe all these problems have solutions through linting and tooling and policies, but it seems like a lot of complexity - maybe the 'low barrier to entry' of an in-monolith package is really an illusion.

siddharthkp: > It feels like it's going to be very hard to avoid a complex situation where we have some code importing and some using the published package. It will be very challenging to ensure that importing code (teams inside the monolith) must always publish all the changes they make, particularly if there is no project actively using the published package version.

(This needs more thought, but...) Would it be fair to say if it's merged into main/master, we should publish it for other teams to use? We could make changelog required for react-shared ([just like primer](https://github.com/primer/react/blob/main/.github/workflows/check_for_changeset.yml)) and publish on merge.

cc @joshblack has some experience with this.

> It will also be very difficult for non-monolith teams to make changes if there are in-monolith teams importing the code, because they will have to address all breaking changes in other projects before they can publish anything.

This is an issue with experimental components in general, no matter where you keep them 😓 Because, there is no commitment to stability, one team can make changes without the knowledge (or testing) that it breaks in a different non-monolith product.

> Additionally - if the components are built inside the monolith, what's to stop them from importing code that also exists in the monolith, outside the react-shared folder? If they do that, how will we publish the code without blowing up the package to include all monolith JS/TS code?
> Maybe all these problems have solutions through linting and tooling and policies

This is definitely a linting opportunity.


> Maybe all these problems have solutions through linting and tooling and policies, but it seems like a lot of complexity - maybe the 'low barrier to entry' of an in-monolith package is really an illusion.

That's a good point! The barrier to entry is relative to primer/react, even then, the barrier is lowered for devs working on the monolith, but not as much for non-monolith projects like memex. (this is something i call out in the ADR, we'll have to find out how bad it is)

Another option (mentioned in [options considered in this ADR](https://github.com/primer/react/blob/adr-experimental-package/contributor-docs/adrs/adr-008-experimental-components.md#other-options-considered)) was to keep experimental components in a repository + package outside the monolith (example: github.com/github/react-shared) which adds some friction for monolith but solves the process/governance problem by creating the same environment for monolith and non-monolith projects. Ultimately, we decided the best tradeoff is to optimise for the monolith were react-shared lives today and then add tooling + processes to reduce friction for non-monolith projects.

contributor-docs/adrs/adr-008-experimental-components.md
dusave: For experimental (aka pre-1.0.0) components, this is where having them in a mono-repo would be extremely helpful. You can deploy components on a per-item basis and, optionally, you can have a `complete` package that just bundles and re-exports them too. Very useful.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+
+&nbsp;
+
+## Decision
+
+1. npm package: "Experimental" components, as defined by the component lifecycle, should not be part of semantically versioned npm package for `@primer/react`.
+
+2. npm package: Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@github/react-commentbox`
```
dusave, line 50: I would question wanting to dump more in the `@github` scope. We might want to make a private new team to house and manage these components. `@drywall`? Since it comes before Primer? 😛 Either way, we'll be publishing a number of components I'm sure, so scoping them would be wise, but not under `@github` or `@primer` IMHO

contributor-docs/adrs/adr-008-experimental-components.md
dusave: Something else to consider as well is if we choose the non-monolith route, we would import this package into gh/gh and A) guarantee that all parts of GitHub are working with the same code and B) will de-dupe any package imports of this package.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,77 @@
...
+#### Risks:
+
+
+&nbsp;
+
+#### Other options considered
+
+1. The code for experimental components should live in a new repository code `github.com/primer/react-candidates` to support different conventions and processes for experimental components and convey shared ownership between primer and product teams. 
+ 
+   Keeping experimental components in primer _org_ suggests that the primer _team_ would take up maintenance of these components while they are still candidates (bugs, a11y remedial, etc.). This will be a new parallel workstream for the team and with our current team size, we might not be able to give it the required attention.
```
dusave, line 69: I like this idea in theory, but candidly, does primer have the bandwidth to handle something like this? Perhaps longer term, but (perhaps falsely) it feels like the primer team is overloaded at the moment. You all do amazing work and I don't want this to cause burnout

siddharthkp:
@howdyray Not yet. The follow up work after this ADR gets approved would be on docs and discoverability of these components, which should help us spot emerging patterns and new system components

contributor-docs/adrs/adr-008-experimental-components.md
iansan5653: >Something else to consider as well is if we choose the non-monolith route, we would import this package into gh/gh and A) guarantee that all parts of GitHub are working with the same code and B) will de-dupe any package imports of this package.

That's exactly my thinking as well - it has the huge benefit of enforcing that everyone uses the code the same way, which could save us some serious headaches.

langermank: > Are we going to have to use the dotcom deploy queue to publish these packages, and if we don't, how will we ensure that the code gets merged into master (due to the deploy-then-merge setup)?

IMO it would be best to explore what the build, deploy and merge story would be for a new package in dotcom before committing to that vs. an external package. To me, it is less than ideal currently for Primer Rails Experimental components, as we adhere to all the same mechanics as everyone else working in dotcom. The main issues with that for component development are:
- No preview (lab) deploys until tests are passing and PR has been approved. Maybe this would be different for a Storybook deploy that doesn't use production data?
- Builds take forever and often fail for things unrelated to a given PR
- Need to be present for merging (trains) even if the thing you're shipping isn't used anywhere. 

I suppose you would need to go through all those steps to bump a package version as well. So maybe it doesn't matter?

joshblack: > (This needs more thought, but...) Would it be fair to say if it's merged into main/master, we should publish it for other teams to use? We could make changelog required for react-shared ([just like primer](https://github.com/primer/react/blob/main/.github/workflows/check_for_changeset.yml)) and publish on merge.
>
> cc @joshblack has some experience with this.


Hey! Definitely think this would be super helpful, especially if it's intended for other teams outside the monolith to use this work.

If it helps, here's a rough overview of a setup I've used in the past:

| Package | Owner | Version Policy | Release cadence | Repo |
| :------ | :---- | :------------- | :-------------- | :------- |
| `@ds/styles` | Core | semver | release train, biweekly | DS Monorepo |
| `@ds/react` | Core | semver | release train, biweekly | DS Monorepo |
| `@ds-incubator/react` | Maintainer(s) of contributed code | pre-1.0 | publish-on-merge | Incubator monorepo |
| `@ds-incubator/<component>` | Maintainer(s) of contributed code | pre-1.0 | publish-on-merge | Incubator monorepo |
| `@local-system/react` | Local system core | semver | publish-on-merge | Local system monorepo |
| `@local-system/<component>` | Maintainer(s) | semver or pre-1.0 | release train, as-needed, or publish-on-merge | Either in monorepo or standalone repo |

Generally:

- Each publish would generate a changelog
- The components of each package had some form of documentation site (designsystem.com/local-system) or was flagged as experimental in a peer docs site, like storybook
- A good number of teams chose a publish-on-merge strategy for faster iterations
- Pre-1.0 was for experimentation and alignment, the road to 1.0 generally involved a commitment to maintenance by core or a community member along with stabilization of the API
- "Monorepos" generally referred to projects that:
  - Published multiple packages (typically via lerna)
  - Tended to have one docs/storybook site
  - Used Yarn Workspaces

Incubator was used for opportunities where several teams may want to share some work and could use shared tooling (linting, formatting, storybook) and processes (releases, deploys) in order to focus on building components.

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: We don't 😅 

That's why this is under other options considered but not selected as decision

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: Love drywall! 🤣 

> I would question wanting to dump more in the @github scope. 

I don't think I understand the tradeoff here, what's the reason for not publishing under github scope?

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: While we don't need to make the decisions about publishing react-shared in this ADR, it's been really useful to collect concerns and challenges before approving this ADR to make sure we're not running into a wall.

Next steps: We need an ADR in gh/gh for publishing react-shared because there are other folks that should be involved. There is a big chunk of work between "Accepting" the current ADR (008) and "Adopting" it.

If we decide to move react-shared out of gh/gh, we can revisit this ADR and update the decision, all part of the process.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,77 @@
...
+
+1. **Need for experimental candidates**: As more teams build React components and look to upstream reusable components, it is natural to look at drafts as the place to put them. However, the barrier to entry into drafts isn't low (because we still follow design reviews, dev conventions, API reviews, etc.)
+
+   Feature teams using Primer may not be intimately familiar with Primer component lifecycle criteria or have the time and/or knowledge to meet it. Often, feature teams create their own components outside of Primer if they find it too difficult to contribute. This leads to lower Primer adoption and conflicting versions of similar functionality throughout GitHub products.
+
+   We want to enable feature teams to contribute to Primer. We also want to make it clear which contributions haven't been evaluated by our high standards for Primer components.
+
+
+
+&nbsp;
```
owenniblock, line 44: Unnecessary space?

contributor-docs/adrs/adr-008-experimental-components.md
owenniblock, line 52: This sounds like a great decision to me, having worked on experimental components in Primer View Components directly (we eventually moved the code to github/github) I'm aware of some of the friction being discussed here.

Building it inside the monolith repository will really help to reduce friction there - it would be great to document the steps you need to take when developing experimental components for projects outside the monolith (it doesn't have to be part of this ADR).

As @langermank pointed out, the build, deploy and merge story could do with some consideration - the hardest part once we'd moved the development to the monolith was getting it merged.

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: the new lines definitely aren't doing anything

contributor-docs/adrs/adr-008-experimental-components.md
iansan5653: Are there any current examples of packages actively getting published from inside the gh/gh monolith repository? If so, that might alleviate some of the concern. If not, that might be a warning sign.

contributor-docs/adrs/adr-008-experimental-components.md
mperrotti: > I don't think I understand the tradeoff here, what's the reason for not publishing under github scope?

I have the same question.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,77 @@
...
+
+3. The code for experimental components should live in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) and published as individual packages as suggested in point 2.
+
+   The monolith already has a place where reusable react components live. This has the lowest barrier to entry for folks working in the monolith.
+
+   Acknowledged risk: It would introduce some additional work for non-monolith projects when upstreaming reusable components. We should work on improving the development and publishing workflow for react-shared.
+
+&nbsp;
+
+#### Risks:
```
mperrotti, line 60: I think the comments from the past two days gave us content that we should add to this section.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Decision
+
+1. npm package: "Experimental" components, as defined by the component lifecycle, should not be part of semantically versioned npm package for `@primer/react`.
+
+2. npm package: Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@github/react-commentbox`
+
+3. The code for experimental components should live in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) and published as individual packages as suggested in point 2.
+
+   The monolith already has a place where reusable react components live. This has the lowest barrier to entry for folks working in the monolith.
```
mperrotti, line 54: What about folks working in Memex? Would they have their own `react-shared`? Or would they have to build in the monolith?

contributor-docs/adrs/adr-008-experimental-components.md
dusave: by publishing under the `@github` scope, especially on a per-component basis, this will flood the scope with repos and namespaces. Since there will be a number of components, it would (IMHO) make sense to be bundled up into its own org for cleanliness.

contributor-docs/adrs/adr-008-experimental-components.md
dusave: This is my concern as well. I can speak for myself that that would also drastically reduce contribution if we'd need to work and merge in the monolith

siddharthkp: >  I can speak for myself that that would also drastically reduce contribution if we'd need to work and merge in the monolith

Can you elaborate on that a little? Is the lack of component tooling or is it the general development workflow of the monolith?

(Added this to the ["Risks" section](https://github.com/primer/react/pull/2214/files#diff-5e30a807664dafbee9f227901a2459809a2c2226e183880f8bf6e05b1f52a7ceR58) as well)

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -0,0 +1,75 @@
...
+
+&nbsp;
+
+#### Other options considered
+
+1. The code for experimental components should live in a new repository code `github.com/primer/react-candidates` to support different conventions and processes for experimental components and convey shared ownership between primer and product teams. 
+ 
+   Keeping experimental components in primer _org_ suggests that the primer _team_ would take up maintenance of these components while they are still candidates (bugs, a11y remedial, etc.). This will be a new parallel workstream for the team and with our current team size, we might not be able to give it the required attention.
+
+2. The code for experimental components should live in a new repository `github/react-shared` or `github/primer-react-candidates` instead of `primer/react-candidates`
```
siddharthkp, line 69: Note: This is a very close second and we should keep this in option in mind

Cons: 
1. More work for folks working on the monolith to contribute to an external package

Pros:
1. A lot more accessible for non-monolith projects to upstream to this (vs the react-shared in monolith)
2. Cleaner versioning and dependency management for both monolith and non-monolith projects, easier to avoid mistakes.
3. Easier to make tooling, publishing workflow and documentation primer-adjacent (compared to the monolith)

contributor-docs/adrs/adr-008-experimental-components.md
siddharthkp: Hi!

I added this, but it felt strange to make and keep that decision in this repo, will include it in the [ADR for gh/gh](https://github.com/primer/react/pull/2214#discussion_r945915211)

contributor-docs/adrs/adr-008-experimental-components.md
iansan5653: Personally, it's the latter. The general development workflow of the monolith is a pain. It's big and slow-moving. Experimental components should be small and fast-moving.

mperrotti (approved):
Per @siddharthkp's suggestion ([Slack link](https://github.slack.com/archives/C02NUUQ9C30/p1661332526550329?thread_ts=1661272488.995289&cid=C02NUUQ9C30) for Hubbers), we could merge this ADR as "accepted" but not adopted.

We should collaborate with somebody who regularly works in the monolith, and create a second ADR there to address the feedback and recommendations collected here.

contributor-docs/adrs/adr-008-experimental-components.md
dusave: > More work for folks working on the monolith to contribute to an external package

Could you elaborate? There is the obvious "would require NPM linking to integrate and test", but that could be mitigated by having proper testing (e.g. Storybook) in that repo, which would allow for rapid testing and prototyping.

From a contribution standpoint, I would actually say it'd be easier. Assuming the bar for experimental components would be lower than that of Primer proper, it would allow for teams to contribute components and rapidly iterate on them.

There is, however, an un-answered question (possibly what you're referring to): How do we manage versions in the monolith? Specifically, if the Issues App team needs `Component@3.0.0` and PRs needs `Component@4.0.0`, how can we make sure that these components are able to live alongside each other? For breaking changes, this could cause major issues in the monolith and shifts the burden of responsibility on the person or team that is upgrading the component to _also_ upgrade all other instances that might be broken by the new version if there is a single version available in the monolith.

PR #2216, opened 2022-08-04
Change `createSlots` to use layout effects when registering slots

iansan5653:
Currently, `createSlots` registers slots and re-renders in asynchronous `useEffect` hooks. This works eventually, but I was running into strange and unpredictable edge cases where changes wouldn't take effect because the component would re-render often and in unpredictable orders.

Changing this to use `useLayoutEffect` fixed the problem and I think this makes more sense - we want to move the slotted content into place _before_ the element paints, which is what `useLayoutEffect` will do.

Further confirming that this is the right move is the snapshot tests for components using slots - previously they were nearly empty because the snapshot was being taken before the slotted content was fully rendered. Now, the snapshots have the full correct contents.

This was previously part of https://github.com/primer/react/pull/2182 but I decided to pull it into a separate PR to justify and link the snapshot updates to this change.

### Screenshots

No visual changes.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


colebemis (approved):
Much better ✨  Thanks, @iansan5653!

siddharthkp (approved):
This is really good!

PR #2217, opened 2022-08-04
Remove duplicate jest-matchmedia-mock

jclem:
This removes a duplicate dependency in package.json.


colebemis (approved):
Nice catch. Thank you!

PR #2219, opened 2022-08-05
Fix slots infinite rendering when no `context` prop is provided

### LINKED ISSUE #2187: Using `Slots` without a `context` prop results an infinite render loop
The `context` prop on the `Slots` component generated by `createSlots` is marked as optional:

https://github.com/primer/react/blob/2b5c86e59f16d697433e3302e78d623a177a061e/src/utils/create-slots.tsx#L31

However, not providing this prop causes an infinite render loop. The result will look fine, but a check of the console will show that it never stops rendering:

> Error: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.

This bug is really hard to figure out since there's no hints as to what causes the issue.

Even just setting the prop to an empty object (`<Slots context={{}}>`) fixes the problem.

iansan5653:
First, updated the slots tests to that the `context` prop is not always defined. This caused the tests to fail due to infinite rendering, as expected from #2187.

Then, fixed the bug by defaulting `context` to a statically referenced frozen object instead of a new object on every render. This fixes #2187 and makes the tests pass ✅.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


iansan5653:
Just realized I totally missed the existing slots test file. Going to put this back in draft for now until I fix that 😅

iansan5653:
Fixed - deleted my new test and updated the existing ones to not always define the `context` object.

siddharthkp (approved):
LGTM!

PR #2224, opened 2022-08-11
Add interaction tests for pagelayout sticky

pksjce:
Add a couple of interactive tests

To test this, check PageLayout/interactions.stories


src/PageLayout/interaction.stories.tsx
```diff
@@ -0,0 +1,399 @@
+import {Meta} from '@storybook/react'
+import {StickyPane} from './PageLayout.stories'
+import {within} from '@storybook/testing-library'
+import {expect} from '@storybook/jest'
+
+const meta: Meta = {
```
broccolinisoup, line 6: Do you think we can move this meta data out of the interaction stories and can be shared? As PageLayout.stories.tsx using the same data.

src/PageLayout/interaction.stories.tsx
```diff
@@ -0,0 +1,399 @@
...
+StickyPane.play = async ({canvasElement}: {canvasElement: HTMLElement}) => {
+  const canvas = within(canvasElement)
+  const content3 = await canvas.getByTestId('content3')
+  content3.scrollIntoView()
+  const paragraph0 = await canvas.getByTestId('paragraph0')
+  const paragraphRect = paragraph0.getBoundingClientRect()
+  expect(isInViewPort(paragraphRect)).toBe(true)
+  const paragraph2 = await canvas.getByTestId('paragraph2')
+  const paragraphRect2 = paragraph2.getBoundingClientRect()
+  expect(isInViewPort(paragraphRect2)).toBe(false)
```
broccolinisoup, line 368: Cool! 🙌🏼

colebemis:
This is great. Thank you for setting this up, @pksjce! A couple of questions:

* Do these interaction tests run during CI? Specifically, will CI fail if we accidentally break an interaction?
* When I run this locally, is this what we expect the interaction tab to look like?
CleanShot 2022-08-17 at 12 32 35@2x

pksjce:
I kinda messed up this branch badly. So I'll close this PR and open a new one.

@colebemis - Thanks!
1. Yes they run on CI. On CI, playwright runs these "storybook tests". Will mention it in the new PR.
2. The interactions for this test don't seem to be very expressive on the addon. There's also the problem that `scrollIntoView` does not come under any of the "actions" that this addon recognises. As mentioned in this https://storybook.js.org/addons/@storybook/addon-interactions at the very end, its 

> calls to userEvent.*, fireEvent, findBy*, waitFor* and expect
that get shown in the step by step debugger.

However, I have some hope that this is being improved. In storybook 7.0.0 versions, there's improvements like this https://github.com/storybookjs/storybook/pull/18555 , which help us have more expressive data in that addon. I feel that the addon is still new and not so mature yet.

PR #2226, opened 2022-08-11
Enable chromatic for all builds without approval (to see if we need to replace or upgrade)

siddharthkp:
Hi! 

Right now, to stay under the free tier usage limit, we only run chromatic after an approval step. 

I'm removing the approval step to see if we will bypass the free tier limit in a month, and ask the question if we should replace it with our own infrastructure or upgrade to a paid plan with higher limits.


joshblack (approved):
Looks great! 🚀

pksjce (approved):

PR #2227, opened 2022-08-12
Removes 'loading' prop from Segmented Control docs

mperrotti:


mperrotti:
Self-merging because this feels pointless to review

PR #2220, opened 2022-08-05
Update SegmentedControl to improve the experience for assistive technologies

mperrotti:
I got some feedback on the SegmentedControl interface guidelines PR that `role="toolbar"` didn't feel right for a segmented control (see [comment](https://github.com/primer/design/pull/276#discussion_r934791688)).

I brought it up in [this week's Accessibility Office Hours](https://github.com/github/accessibility/issues/1556), and it was suggested that we change the segmented control from `role="toolbar"` behavior to `role="list"/"listitem"` behavior.

### Screenshots

Before:
https://user-images.githubusercontent.com/2313998/180625732-901564bf-4b2e-4526-a2df-3848fb98d5b3.mp4

After:
https://user-images.githubusercontent.com/2313998/183119490-af7db406-63e5-4395-ba2f-76941691b497.mp4

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2145, opened 2022-06-24
Improves keyboard navigation for the SegmentedControl

mperrotti:
Implements keyboard navigation behavior described in the a11y design review ([comment](https://github.com/github/primer/issues/762#issuecomment-1145538220)).

### Screenshots

https://user-images.githubusercontent.com/2313998/175668068-a14a8be8-0fcf-4b3b-9962-163b8deb29d2.mp4

### Merge checklist

- [x] Added/updated tests
- [n/a] Added/updated documentation
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2228, opened 2022-08-12
Updates SegmentedControl to Alpha status

mperrotti:


docs/content/SegmentedControl.mdx
```diff
@@ -1,14 +1,15 @@
 ---
+componentId: segmentedControl
 title: SegmentedControl
-status: Draft
 description: Use a segmented control to let users select an option from a short list and immediately apply the selection
+status: Alpha
```
mperrotti, line 5: The component lifecycle label doesn't appear to be updating. Is there something else I need to do to get this showing "Alpha" instead of "Drafts"?

Screen Shot 2022-08-15 at 11 56 22 AM

siddharthkp: This might be related to staging build because I see your changes on my machine 🤔

colebemis (approved):
🎉

docs/content/SegmentedControl.mdx
```diff
@@ -1,14 +1,15 @@
 ---
+componentId: segmentedControl
```
colebemis, line 2: component ids use snake case: 
```suggestion
componentId: segmented_control
```

docs/content/SegmentedControl.mdx
siddharthkp: Deployment is fixed now, turns out it was my fault ✨ 🙃

Navigation through arrow keys still doesn't work for me, but otherwise looks perfect!

siddharthkp (approved):
Not sure if it's just me again, but navigation through arrow keys introduced in https://github.com/primer/react/pull/2145 doesn't work 🤔  

Perfect otherwise, approving in advance.

mperrotti:
@siddharthkp - it's working as expected. We removed that functionality in https://github.com/primer/react/pull/2220

PR #2233, opened 2022-08-16
Add source repo check for deploy_preview_fork

siddharthkp:
Improvement on top of https://github.com/primer/react/pull/2207 to reduce number of actions run


siddharthkp:
Self reviewing workflow change!

Issue #2231, opened 2022-08-15
`MarkdownEditor` followup work


iansan5653:
As the [`MarkdownEditor`](https://github.com/primer/react/pull/2182) component gets used in more places, we'll naturally identify some followup tasks to complete after the initial draft release. Personally I need to shift focus back to some other priorities so I can't tackle these myself right now, which is why I wanted to catalog them somewhere:

- [ ] Consumers should only be able to use refs of the exact type `Ref<MarkdownEditorHandle>` (currently any ref assignable to that type is allowed, which enables consumers to incorrectly use `Ref<HTMLTextAreaElement>` accidentally)
  - https://github.com/github/github/pull/231484#discussion_r945742831
- [ ] The component should log an error if no `MarkdownEditor.Label` child is found - this creates a `fieldset` with no `legend`
  - https://github.com/github/github/pull/231484#discussion_r945753375
- [ ] Users should be able to switch between view/edit modes with <kbd>Cmd+Shift+P</kbd>
  - https://github.com/github/memex/issues/11411
- [x] The task list icon should be updated to be consistent with the new icon used in the Rails-based editor (task list instead of checkbox)
  - Will be automatically fixed when `@primer/octicons-react` is updated
- [ ] All examples in the `MarkdownViewer` docs have `TypeError: Cannot read properties of undefined (reading '__html')`

PR #2236, opened 2022-08-16
Small improvements to `MarkdownEditor` component

iansan5653:
This has some minor followup changes to #2182:

- Adds <kbd>Ctrl+Shift+P</kbd> (<kbd>Cmd+Shift+P</kbd> on Mac) shortcut for toggling between edit & preview mode. This was a little trickier than I expected - toggling to preview mode is easy because we can add the listener to the input, but listening for the shortcut to go back to edit mode requires a global listener (because there is no visible input to focus). When we do that, we have to maintain a global stack of editors in the order they've switched to preview mode in case multiple editors are mounted at once.
- Limits the type of the `Ref` passed to `MarkdownEditor` so that consumers cannot accidentally create the ref using `useRef<HTMLTextAreaElement>`. This was possible before because `HTMLElement` was assignable to `MarkdownEditorHandle`, so I added a fake symbol 'brand' key to disallow this. It's somewhat hacky but worth it to prevent a very common mistake.
- Fixes the prop names in examples in the `MarkdownViewer` docs so the examples will render (they are erroring with `TypeError: Cannot read properties of undefined (reading '__html')`).

This fixes https://github.com/primer/react/issues/2231, except for erroring when there is no `Label` present - I tried to fix this but it's very hard to do with the current implementation of slots. I think it's not worth the extra lift right now.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


siddharthkp (approved):
Scope to drafts, pretty safe. Trust Ian on the implementation details

src/drafts/MarkdownEditor/MarkdownEditor.tsx
```diff
@@ -254,6 +276,34 @@ const MarkdownEditor = forwardRef<MarkdownEditorHandle, MarkdownEditorProps>(
...
+        const handler = (e: KeyboardEvent) => {
+          if (
+            !e.defaultPrevented &&
+            editorsInPreviewMode.at(-1) === id &&
+            isModifierKey(e) &&
+            e.shiftKey &&
+            e.key === 'p'
+          ) {
+            setView?.('edit')
+            requestAnimationFrame(() => inputRef.current?.focus())
```
iansan5653, line 292: On second thought I'm pretty sure this should be `setTimeout`
```suggestion
            setTimeout => inputRef.current?.focus())
```

iansan5653: This is what I get for trying to make this fix from my phone 🤦

PR #2239, opened 2022-08-17
Implement SplitPageLayout

colebemis:
Implements draft [SplitPageLayout](https://github.com/github/primer/issues/793) using PageLayout.

👉  [SplitPageLayout docs](https://primer-7893ff356e-13348165.drafts.github.io/SplitPageLayout)

**Note to reviewers:** The bulk of the implementation lives in [`src/SplitPageLayout/SplitPageLayout.tsx`](https://github.com/primer/react/pull/2239/files#diff-e375d41bb55ba3cec3bcc122758a2f04e2d597f84d7f6a7cb59c26af15dd73c2)

Closes https://github.com/github/primer/issues/793

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/SplitPageLayout.mdx
+++ b/docs/content/SplitPageLayout.mdx
@@ -153,6 +153,27 @@ If you need a more flexible layout component, consider using the [PageLayout](/P
 </Box>
 ```
 
+### With condensed padding
+
+```jsx live drafts
+<Box sx={{height: 320, overflowY: 'auto', border: '1px solid', borderColor: 'border.default'}}>
+  <SplitPageLayout>
+    <SplitPageLayout.Header padding="condensed">
+      <Placeholder label="Header" height={64} />
+    </SplitPageLayout.Header>
+    <SplitPageLayout.Pane padding="condensed">
+      <Placeholder label="Pane" height={120} />
+    </SplitPageLayout.Pane>
+    <SplitPageLayout.Content padding="condensed">
+      <Placeholder label="Content" height={320} />
+    </SplitPageLayout.Content>
+    <SplitPageLayout.Footer padding="condensed">
+      <Placeholder label="Footer" height={64} />
+    </SplitPageLayout.Footer>
+  </SplitPageLayout>
+</Box>
+```
+
 ### Without padding
 
 ```jsx live drafts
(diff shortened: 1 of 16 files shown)


docs/content/SplitPageLayout.mdx
```diff
@@ -3,156 +3,211 @@ title: SplitPageLayout
...
+      <Placeholder label="Pane" height={120} />
+    </SplitPageLayout.Pane>
+    <SplitPageLayout.Footer>
+      <Placeholder label="Footer" height={64} />
+    </SplitPageLayout.Footer>
+  </SplitPageLayout>
+</Box>
 ```
 
 ### Without padding
```
mperrotti, line 156: Non-blocking nitpick: it could be nice to add an example where padding is changed to one of the other values. That seems like the more common use case, but i could be wrong.

docs/content/SplitPageLayout.mdx
```diff
@@ -227,9 +300,19 @@ description: Provides structure for a split page layout, including independent s
 <PropsTable>
   <PropsTableRow
     name="position"
-    type={`ResponsiveValue<
-  'start' | 'end'
->`}
+    type={`| 'start'
```
mperrotti, line 303: Non-blocking: This is much clearer to the docs reader than `ResponsiveValue<'start' | 'end'>`, but it will be annoying to update when/if we change our breakpoints.

mperrotti (approved):

docs/content/SplitPageLayout.mdx
colebemis: Agreed. This is a temporary solution until we can think of a better way to document responsive values. Open to ideas!

src/PageLayout/PageLayout.stories.tsx
```diff
@@ -1,7 +1,6 @@
-import {Meta, Story} from '@storybook/react'
 import React from 'react'
```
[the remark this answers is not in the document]
colebemis: Yeah, I was also curious about this

docs/content/SplitPageLayout.mdx
colebemis: Done 👍

PR #2240, opened 2022-08-18
PageLayout.Pane: `stickyTop` bug fixes

colebemis:
I noticed a couple minor bugs in `stickyTop` that I missed during my initial review. This PR fixes them.

**Bug 1**

Sometimes passing a number to `stickyProp` gives you an unexpected result. This happens because the `sx` prop treats the number as an index into our [spacing scale](https://github.com/primer/react/blob/main/src/theme-preval.js#L50) instead of a pixel value.

CleanShot 2022-08-17 at 18 43 57@2x

**Bug 2**

If PageLayout is nested inside a scroll container, the height of the pane doesn't account for the stickyTop prop. 

CleanShot 2022-08-17 at 18 44 58@2x


broccolinisoup (approved):
Great catch about the sx props @colebemis !! ⚡ All is looking good to me. Thanks for fixing it!

PR #2241, opened 2022-08-18
Add interaction tests for pagelayout sticky

pksjce:
Describe your changes here.

Add a couple of interactive tests

To test this, check PageLayout/interactions.stories

Taking over from https://github.com/primer/react/pull/2224

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


pksjce:
@colebemis - This https://github.com/primer/react/runs/7892938558?check_suite_focus=true is where the tests will run on CI. Unfortunately it looks for tests in all stories.

src/PageLayout/PageLayout.stories.tsx
```diff
@@ -489,27 +489,38 @@ export const StickyPane: Story = args => (
...
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nam at enim id lorem tempus egestas a non ipsum.
-            Maecenas imperdiet ante quam, at varius lorem molestie vel. Sed at eros consequat, varius tellus et, auctor
-            felis. Donec pulvinar lacinia urna nec commodo. Phasellus at imperdiet risus. Donec sit amet massa purus.
-          </Box>
-        ))}
+        {Array.from({length: args.numParagraphsInPane}).map((_, i) => {
+          const testId = `paragraph${i}`
+          return (
+            <Box key={i} as="p" sx={{margin: 0}}>
+              <span data-testid={testId}>
```
pksjce, line 515: Just adding some test ids to grab blocks

src/PageLayout/interaction.stories.tsx
```diff
@@ -389,5 +390,41 @@ NonStickyPane.play = async ({canvasElement}: {canvasElement: HTMLElement}) => {
...
+}
+
+CustomStickyHeader.play = async ({canvasElement}: {canvasElement: HTMLElement}) => {
+  const canvas = within(canvasElement)
+  const contentToScroll = await canvas.getByTestId('content3')
+  contentToScroll.scrollIntoView()
+
+  // fireEvent alternative?
+  const storyWindow = await canvas.getByTestId('story-window')
+  await fireEvent.scroll(storyWindow, {top: 600})
```
broccolinisoup, line 419: I wanted to give it a go to more descriptive event names, but can't seem to make this work :( See the [doc](https://testing-library.com/docs/dom-testing-library/api-events) Maybe I am missing something? @pksjce

src/PageLayout/interaction.stories.tsx
```diff
@@ -1,7 +1,8 @@
 import {Meta} from '@storybook/react'
-import {StickyPane} from './PageLayout.stories'
-import {within} from '@storybook/testing-library'
+import {StickyPane, CustomStickyHeader} from './PageLayout.stories'
+import {within, fireEvent} from '@storybook/testing-library'
 import {expect} from '@storybook/jest'
+import {setTimeout} from 'timers/promises'
```
broccolinisoup: removed it in the next commit

src/PageLayout/interaction.stories.tsx
pksjce: They don't yet have a descriptive name feature. However, storybook 7.0.0 is going to support sections within interaction tests I believe!

src/PageLayout/interaction.stories.tsx
```diff
@@ -0,0 +1,429 @@
...
+  numParagraphsInContent: {
+    type: 'number',
+    defaultValue: 30
+  }
+}
+
+CustomStickyHeader.play = async ({canvasElement}: {canvasElement: HTMLElement}) => {
+  const canvas = within(canvasElement)
+  const contentToScroll = await canvas.getByTestId('content3')
+  contentToScroll.scrollIntoView()
```
pksjce, line 415: Do you need to scroll twice? The `fireEvent.scroll` below would be ideal. I would fix my tests to work on that if it works for you.

src/PageLayout/interaction.stories.tsx
broccolinisoup: Ahh cool! I'll remove this then, and we can try later when available.

src/PageLayout/interaction.stories.tsx
broccolinisoup: No it doesn't work :( This is what I mean [here ](https://github.com/primer/react/pull/2241#discussion_r951030514) Maybe I am missing something? 

Because the fireEvent.scroll didn't work I included contentToScroll. I'll give another go today and let you know if I make it work.

pksjce: Yeah! I had the same experience. It would have been so much better if fireEvent scroll worked in this case.Maybe we aren't using the scrollcontainer effectively?

colebemis (approved):
Thank you for setting this up, @pksjce ✨

src/PageLayout/interaction.stories.tsx
broccolinisoup: I tried it again but no luck 😢 I agree, I think it is about the scroll container. I tried the outer box, the page layout content as scroll containers but none of them are scrollable. (el.scrollHeight-el.clientHight is 0) and I tried giving auto overflow but that doesn't seem to work either. I'll remove this fireEvent for now and maybe down the track, it will be possible to do with user events from SB.

PR #2247, opened 2022-08-19
Improve the `slots` pattern and implementation

iansan5653:
### Summary

This PR contains a number of changes and improvements to the 'slots' pattern (`createSlots`) to fix performance issues, reduce complexity, and improve the API:

Closes https://github.com/github/primer/issues/1224

#### Remove `context` prop from `Slots`

The ability to pass custom non-slots-related contexts through the slots API is problematic:

- It is not typesafe (types are `any` going in _and_ coming out)
- It breaks memoization - the slots context and components cannot be memoized if a custom context object that changes on every render is passed
- It is a violation of single-responsibility principle - slots should manage slots and nothing else
- It increases complexity in the already-complex `createSlots` function

Because of these issues, particularly the memoization problem, I removed the custom context capabilities altogether. The missing functionality is easily solved by creating new contexts using `React.createContext`, which follows the single-responsibility principle and is typesafe.

#### Use normal React state to manage slots

Currently we are storing the slots object in a `ref` and forcing updates every time it changes. The idea behind this is that we can avoid extra renders on mount, when a number of slots are registering all at once. In reality this actually drastically hurt performance because React cannot batch these updates. By moving the store into normal state, we can simply update state and let React decide when to render.

This looks slightly different from normal `useState` because I'm using the slightly more low-level `useReducer` to clean up the code. `useReducer` also helps us avoid passing an object (which would have to be memoized) to `SlotsContext`.

Now that there is only one function being passed to `SlotsContext` (and that function is guaranteed to never change because it's a `reducer`), `SlotsContext` will never change - this means fewer extra renders!

#### Memoize the `Slot` component

Even with the context value never changing, we still see extra renders in the `Slots` components themselves. This is because React re-renders the entire subtree when state changes, and `Slots` change state when they render - causing them to render again. We can easily fix this by memoizing the `Slot` component with `memo`. Now it will not re-render unless its `children` change - that's only once per 'real' render.

#### Split apart the register & unregister effects

We are currently unnecessarily unregistering each slot before we register it. This is not needed because slots just overwrite the store entry. We do still need to unregister on dismount, but that can have a smaller dependency list - thus fewer state updates!

This actually has no practical effect now that we are using state to store the slots; previously it had a major impact because we were forcing a render on unregister _and_ on register. But now React would just batch the state updates anyway. However, I don't think we should _rely_ on React batching things under the hood - we should still avoid unnecessary updates.

#### Change to a hook-based API

Currently we generate a `Slots` provider component with `createSlots`, which provides the slots through a render function. This render function is hard to read, and more importantly it doesn't provide access to the `slots` record at the top-level. We often want to access this object to make decisions based on whether a slot exists or not, so this is a very cumbersome API:

```tsx
const {Slots, Slot} = createSlots<"One" | "Two">();
export const ItemSlot = Slot;

export const Item = ({children}) => {
  return (
    <Slots>
      {(slots) => {
        const oneExists = slots.One !== undefined;

        return (
          <>
            {oneExists ? "One Exists" : null}
            {slots.One}

            {children}
            {slots.Two}
          </>
        );
      }}
    </Slots>
  );
};
```

To improve this, I've changed the API to generate a `useSlots` hook instead. When called inside a component, that hook returns the `slots` record and a `SlotsProvider` component, which acts like a normal context provider (but with no props). In practice, this looks like this:

```tsx
const {useSlots, Slot} = createSlots<"One" | "Two">();
export const ItemSlot = Slot;

export const Item = ({children}) => {
  const {SlotsProvider, slots} = useSlots();

  const oneExists = slots.One !== undefined;

  return (
    <SlotsProvider>
      {oneExists ? "One Exists" : null}
      {slots.One}

      {children}
      {slots.Two}
    </SlotsProvider>
  );
};
```

This should have no performance impact, but I think it's much more readable and ergonomic.

### Screenshots

Here's a quick demo showing how many fewer renders we are seeing now vs before:

_Before: Screen recording of a console logging every `Slot` render and every top-level `Slots` provider render in the `MarkdownEditor` component. Everything renders three times on initial mount, then each time the value of the input changes, everything renders twice._

https://user-images.githubusercontent.com/2294248/185676730-6876e905-6bb9-4b1f-bc16-e401e3b2ddd9.mov

_After: A similar recording, after the changes have been made. Each `Slot` renders once on initial mount and once per change. The top-level provider renders twice on mount and twice per change._

https://user-images.githubusercontent.com/2294248/185678092-05efbc68-8078-44ab-8f79-aa4016316b15.mov

Of course, number of renders alone is not a good indicator of performance, but it does show that we've been able to simplify the event cycle here which should help resolve the edge cases we've seen. And at the same time, we've been able to improve the API, so it's a win-win.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


iansan5653:
Looks like some components (but not all 🤔) using slots are failing to render due to infinite update depth. Not sure what's going on there.

src/utils/create-slots.tsx
```diff
@@ -1,104 +1,72 @@
...
-import {useForceUpdate} from './use-force-update'
 import useLayoutEffect from './useIsomorphicLayoutEffect'
 
 /** createSlots is a factory that can create a
  *  typesafe Slots + Slot pair to use in a component definition
  *  For example: ActionList.Item uses createSlots to get a Slots wrapper
  *  + Slot component that is used by LeadingVisual, Description
+ *
+ * ⚠️ WARNING: Providing a custom `Context` in the same component as `useSlots` is called
+ * will result in an infinite render loop unless you memoize that context object!
```
iansan5653, line 11: This is a frustrating gotcha - if you don't memoize the context, the component will get stuck in an infinite cycle of re-renders. This is because the `Slot` has a matching `useContext`, so an update in the context will re-render the `Slot`. When the `Slot` renders, it registers itself, re-rendering the component with the `useSlots` hook. Since the context is defined in that component, the context updates again, _ad infinitum_.

When `Slots` was a component instead of the `useSlots` hook, this wasn't as big of an issue because you could put the context provider outside of the `Slots` component in the JSX and avoid the infinite renders. That was still a gotcha though, so I don't think it's worth reverting the hook change to fix this.

We could bring back a custom `context` prop on `SlotsProvider` and avoid the infinite render bug, but it would significantly hurt performance because the `SlotsContext` could no longer be effectively memoized.

I'm open to ideas on improving this. It's not a bad idea to memoize context objects anyway, but in this case it feels like engineers will run into this problem regularly and the solution is not obvious. Is there a way we can at least catch the infinite render error and show more useful information?

siddharthkp:
Hi @iansan5653! 

We discussed this in the engineering sync and heads up, it might be a while! We have decided to slow down and do some explorations before committing to a solution

siddharthkp:
Converting this into a draft, purely for logistics 😅

iansan5653:
@siddharthkp should we just let this close? I still think this change would solve a lot of headaches and fix performance issues. Is it worth the effort to review and try it out?

PR #2254, opened 2022-08-22
chore(test): update coverage to run only in ci

joshblack:
This PR removes the `coverage` option from `jest.config.js` and enables it in CI through `npm run test -- --coverage`. You can also run this command locally to get coverage information.

I opened this to see how others felt about this change 👀 The `coverage` option was making it difficult to use a watch workflow with jest (`npm test -- --watch`) and this seemed like one path to take to make that loop faster. 

Let me know if there would be a better way or if the intent is to pass in `--no-coverage` locally and I'd be happy to update/close as-needed 👍


colebemis (approved):
Much better. Thank you! ✨

PR #2257, opened 2022-08-22
refactor(build): update rollup build step

joshblack:
This is a draft PR to see what it would be like to use `rollup` for the full build process. Using rollup in this way can:

- Decrease package size on disk (and reduce install times)
- Reduce build time (somewhat)
- Deduplicate common code across modules (useful when using babel helpers)

It can also be a useful tool for highlighting when modules have circular dependencies within the project. It also has the side-effect of only production artifacts making their way into the package (no tests or stories). Types still seem to be emitted by the build process.

**Package size (on disk)**

| Folder | Before | After | % Change |
| :----- | :----- | :---- | :------- |
| `lib-esm` | 4.3M | 828K | -80% |
| `lib` | 4.6M | 884K | -81% |

This PR also marginally decreases the build time from roughly 18s to 13s, mostly due to it using the same bundle to output cjs/esm/umd instead of building once in rollup and compiling once for cjs/esm.

| Task | Before | After | % Change |
| :----- | :----- | :---- | :------- |
| `time npm run build` | 18.859s | 13.561s | -28% |

**Blockers**

This change cannot be adopted until circular dependencies are addressed

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -3,24 +3,60 @@ import commonjs from '@rollup/plugin-commonjs'
 import resolve from '@rollup/plugin-node-resolve'
 import {terser} from 'rollup-plugin-terser'
 import visualizer from 'rollup-plugin-visualizer'
+import packageJson from './package.json'
 
 const extensions = ['.js', '.jsx', '.ts', '.tsx']
-
 const formats = ['esm', 'umd']
-
-const plugins = [
-  babel({extensions, exclude: 'node_modules/**', runtimeHelpers: true}),
-  resolve({extensions}),
-  commonjs(),
-  terser(),
-  visualizer({sourcemap: true})
+const external = [
+  ...Object.keys(packageJson.peerDependencies),
+  ...Object.keys(packageJson.dependencies),
+  ...Object.keys(packageJson.devDependencies)
 ]
+
+const baseConfig = {
+  input: 'src/index.ts',
+  external: id => {
+    return external.some(pkg => {
+      return id.startsWith(pkg)
+    })
+  },
+  plugins: [
+    resolve({
+      extensions
+    }),
+    commonjs(),
+    babel({
+      extensions,
+      exclude: /node_modules/,
+      runtimeHelpers: true
+    })
+  ]
+}
 
 export default [
   {
-    input: 'src/index.ts',
+    ...baseConfig,
+    output: {
+      dir: 'lib-esm',
+      format: 'esm',
+      preserveModules: true,
+      preserveModulesRoot: 'src'
+    }
+  },
+  {
+    ...baseConfig,
+    output: {
+      dir: 'lib',
+      format: 'commonjs',
+      preserveModules: true,
+      preserveModulesRoot: 'src',
+      exports: 'named'
+    }
+  },
+  {
+    ...baseConfig,
     external: ['styled-components', 'react', 'react-dom'],
-    plugins,
+    plugins: [...baseConfig.plugins, terser(), visualizer({sourcemap: true})],
     output: formats.map(format => ({
       file: `dist/browser.${format}.js`,
       format,
--- a/script/build
+++ b/script/build
@@ -3,16 +3,10 @@
 set -e
 
 # Clean up
-rm -rf dist
-rm -rf lib
-rm -rf lib-esm
+npm run clean
 
 # Bundle
 npx cross-env NODE_ENV=production rollup -c
-
-# Compile
-npx cross-env BABEL_MODULE=commonjs babel src --out-dir lib --extensions ".js,.ts,.jsx,.tsx"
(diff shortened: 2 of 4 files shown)


script/build
```diff
@@ -3,17 +3,11 @@
 set -e
 
 # Clean up
-rm -rf dist
-rm -rf lib
-rm -rf lib-esm
+npm run clean
```
pksjce, line 6: So clean!

rollup.config.js
```diff
@@ -3,24 +3,60 @@ import commonjs from '@rollup/plugin-commonjs'
...
-const plugins = [
-  babel({extensions, exclude: 'node_modules/**', runtimeHelpers: true}),
-  resolve({extensions}),
-  commonjs(),
-  terser(),
-  visualizer({sourcemap: true})
+const external = [
+  ...Object.keys(packageJson.peerDependencies),
+  ...Object.keys(packageJson.dependencies),
+  ...Object.keys(packageJson.devDependencies)
```
pksjce, line 13: Is this how the bundle size is reduced?

lesliecdubs:
I’ll let other folks weigh in here, but thought this might be prudent to share (x-posting from Slack for posterity) --

[#web-systems](https://github-grid.enterprise.slack.com/archives/CMZ4DC9BL), our sister UI platform team over in the Eng org, historically made the call to [use rollup to build GitHub assets](https://github.com/github/web-systems/blob/0c5b4c38959b58782c41341892c6d5d9dc024f50/docs/adr/0010-use-rollup-to-build-github-assets.md?plain=1#L50). More recently, they decided to [roll back this decision and adopt webpack instead](https://github.com/github/web-systems/blob/0c5b4c38959b58782c41341892c6d5d9dc024f50/docs/adr/0014-use-webpack-for-bundling-assets.md?plain=1#L68). Perhaps there’s insight about their decision-making process in here that could be useful for us to consider, too.

pksjce (commented):
This is a really cool change! I have a few questions.

- Will the lib-esm package size decrease be felt by our consumers? Will their bundle sizes decrease? 
- What kind of tests were you able to run for these new bundled files? Can you make sure nextjs/remix are able to consume them with no issues? We've had tons of issues regarding bundling now working for these frameworks. Maybe you could test in memex too?
- Did you test for importing draft components. The subpaths were one of my pain points with esm modules.
- Also I'm a little apprehensive about making dependencies external. Maybe I don't understand the context properly.

rollup.config.js
```diff
@@ -3,24 +3,60 @@ import commonjs from '@rollup/plugin-commonjs'
...
+    return external.some(pkg => {
+      return id.startsWith(pkg)
+    })
+  },
+  plugins: [
+    resolve({
+      extensions
+    }),
+    commonjs(),
+    babel({
```
keithamus, line 28: I'd recommend avoiding babel here. We dropped babel from dotcom and all public repos a few years ago because:

- babel is quite slow. Typescript or esbuild or swc are much faster and equally as capable
- Babel tends to err on the side of compatibility which means transforming things that don't necessarily need to be transformed due to edge cases. (An example of this is object rest spread gets transformed if you specify edgehtml as a browser, even though edge supports object rest spread). This results in bigger bundle sizes than necessary.
- Configuring babel can be tricky and can result in varying compatibilities. Typescripts `target` is a little more blunt (just specifying a year to compile down to) but that's much easier to propagate across many repos.

In addition, moving from babel would be a great step to coalescing all our repositories around a central build config. As mentioned doctor uses swc, while all other repositories just use typescript. 

Here's Primer ViewComponents configuration: https://github.com/primer/view_components/blob/b0dbefb471d8262a2a45235ccaf0501260f00cd4/rollup.config.js#L15

keithamus:
> [#web-systems](https://github-grid.enterprise.slack.com/archives/CMZ4DC9BL), our sister UI platform team over in the Eng org, historically made the call to [use rollup to build GitHub assets](https://github.com/github/web-systems/blob/0c5b4c38959b58782c41341892c6d5d9dc024f50/docs/adr/0010-use-rollup-to-build-github-assets.md?plain=1#L50). More recently, they decided to [roll back this decision and adopt webpack instead](https://github.com/github/web-systems/blob/0c5b4c38959b58782c41341892c6d5d9dc024f50/docs/adr/0014-use-webpack-for-bundling-assets.md?plain=1#L68). Perhaps there’s insight about their decision-making process in here that could be useful for us to consider, too.


This makes sense given the context of dotcom which requires delivering of assets in ways which webpack supports better than rollup, but for libraries we still use rollup across the board as it provides much simpler configuration.

joshblack:
Thanks all for leaving a comment! Going to close out this exploration and will post a follow-up after addressing the circular deps issue 😄 Will also make sure to revisit the rollup setup with TypeScript vs Babel 👍

PR #2260, opened 2022-08-23
Add hooks to draft exports

iansan5653:
Very minor PR to add the hooks in `drafts/hooks/` to the exports defined in `drafts/index.ts`. I intended to do this when I introduced them but forgot.

The only hook I excluded is `useUnifiedFileSelect` because I think the use-case is fairly specific here and it's not worth making it public if we want to more easily change the implementation later. If someone requests it we can export it.


src/drafts/MarkdownEditor/index.ts
```diff
@@ -39,5 +39,5 @@ export type {MarkdownEditorProps} from './MarkdownEditor'
 export type {Emoji} from './suggestions/_useEmojiSuggestions'
 export type {Mentionable} from './suggestions/_useMentionSuggestions'
 export type {Reference} from './suggestions/_useReferenceSuggestions'
-export type {FileUploadResult} from './_useFileHandling'
+export type {FileUploadResult, FileType} from './_useFileHandling'
```
iansan5653, line 42: I added `FileType` to the `MarkdownEditor` exports because it's useful when building the `acceptedFileTypes` prop.

src/drafts/hooks/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './useCombobox'
```
iansan5653, line 1: I wasn't 100% sure if I can just `export *` vs calling out each `export type` individually, but it appears to work fine based on the `export * from './MarkdownViewer'` in the `drafts/index.ts` file.

colebemis (approved):

PR #2265, opened 2022-08-24
PageHeader: Initial implementation

pksjce:
PageHeader component's initial implementation.

Related issues: 
- https://github.com/github/primer/issues/795

Figma specs for reference :
- https://www.figma.com/file/Ee0OrXuOLXMDqUW83EnDHP/PageHeader-(FY23-Q1)?node-id=2%3A7206&t=vNgm2MP4dzYmS6lU-0 

Screen Shot 2022-11-17 at 3 52 55 pm

### Tasks
- [x] High fidelity implementation 
- [x] Make sure every composite component's visibility is manageable through the API `hidden` prop with the provided default values (Default values varies depending on the viewport)
- [x] Low fidelity implementation of the composite components' layout. (exact spacing / typography etc from Figma specs)
- [x] Storybook controls should cover the API and all props (where practical)


src/PageHeader/PageHeader.stories.tsx
```diff
@@ -0,0 +1,413 @@
...
+    },
+    LeadingVisual: {
+      ...OcticonArgType([CodeIcon, GitPullRequestIcon, PeopleIcon]),
+      table: {
+        category: 'TitleArea Slot'
+      },
+      description:
+        'Leading visualLeading visuals are optional and appear at the start of the title. They can be octicons, avatars, and other custom visuals that fit a small area.'
+    },
+    'LeadingVisual.hidden': {
```
vdepizzol, line 186: In these stories, removing these additional `LeadingVisual`, `TrailingVisual`, `TrailingAction` options requires the user to "turn on" the hiding. I wonder if we can provide a better experience.

For example, in the `SplitPageLayout` story, I've included `hasPane`, `hasHeader` toggles as the first item of each category to make it easier to parse all the functionality:

Link: https://primer.style/css/storybook/?path=/story/components-layout-beta-splitpagelayout--playground

src/PageHeader/PageHeader.stories.tsx
```diff
@@ -0,0 +1,413 @@
...
+    layout: 'fullscreen',
+    controls: {expanded: true}
+  },
+  args: {
+    'ContextArea.hidden': false,
+    ParentLink: 'Previous Page',
+    'ParentLink.hidden': false,
+    'ContextBar.hidden': true,
+    'ContextAreaAction.hidden': true,
+    'BackButton.hidden': false,
```
vdepizzol, line 36: Now that I'm seeing this live, it seems to me having this as a `leadingAction` (since it's placed alongside `trailingAction`, `leadingVisual` and `trailingVisual`, a là ActionListItem) makes more sense and is a more flexible solution.

Especially with newer designs considering other actions as at the leading position:

Screen Shot 2022-11-22 at 18 06 54

I think that should clear any confusion between `ParentLink` and `BackButton`, what do you think @broccolinisoup? 🙇

broccolinisoup: Yes, it makes a lot of sense!

src/PageHeader/PageHeader.stories.tsx
```diff
@@ -0,0 +1,413 @@
...
+            narrow: 'medium',
+            regular: 'large',
+            wide: 'large'
+          } `
+        }
+      },
+      description:
+        '`medium` is the most common page title size. Use for static titles in most situations. `large` for for user-generated content such as issues, pull requests, or discussions. `subtitle` when a PageHeader.Title is already present in the page, such as in a SplitPageLayout.'
+    },
+    LeadingVisual: {
```
broccolinisoup, line 178: @vdepizzol I wonder if a dropdown control with a few choice of Octicons provides a good experience here? Because LeadingVisual can be Avatars or other custom visuals as you mentioned on the Figma specs. Same goes for `TraliningVisual` . I added a description attribute to give some more info (see a few lines below.) but I am not sure if that would be helpful. Curious to hear your thought on that 👀

src/PageHeader/PageHeader.stories.tsx
broccolinisoup: This is something I wasn't sure either. Thanks for providing feedback on that 🙇🏼 I was looking from a perspective that slots come visible as default and consumers can manage their visibility by hiding them. I even got confused myself when writing them and I am glad you mentioned it too.

I changed the prop from being `hidden` to `visible` now and controls became `hasContextArea`, `hasParentLink`, `hasLeadingVisual` etc and I feel it is clearer now. What you think about the current structure now?

src/PageHeader/examples.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <Hidden on={['regular', 'wide']}>
+          <Button variant="primary">New</Button>
+        </Hidden>
+        <IconButton aria-label="More" icon={KebabHorizontalIcon} />
+      </PageHeader.Actions>
+    </PageHeader.TitleArea>
+  </PageHeader>
+)
+
+export const WebHooks = () => (
```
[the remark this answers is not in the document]
broccolinisoup: Ah great idea!

src/PageHeader/examples.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+  <PageHeader sx={{padding: 3}}>
+    <PageHeader.ContextArea>
+      <PageHeader.ParentLink href="http://github.com">Repository Settings</PageHeader.ParentLink>
```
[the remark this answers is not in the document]
broccolinisoup: Thank you for the link! I updated them all.

src/PageHeader/examples.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <Hidden on={['narrow']}>
+          <Button variant="primary">New Webhook</Button>
+        </Hidden>
+
+        <Hidden on={['regular', 'wide']}>
+          <Button variant="primary">New</Button>
+        </Hidden>
+      </PageHeader.Actions>
+    </PageHeader.TitleArea>
+  </PageHeader>
```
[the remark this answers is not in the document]
broccolinisoup: That is so true! I will certainly include more examples here :)

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,354 @@
...
+            medium: '20px',
+            subtitle: '20px'
+          }[currentVariant],
+          lineHeight: {
+            large: '48px',
+            medium: '32px',
+            subtitle: '32px'
+          }[currentVariant],
+          fontWeight: {
+            large: '600',
```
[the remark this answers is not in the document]
broccolinisoup: Oh, so it should be normal (400) for large style texts?

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,354 @@
...
+}
+
+// PageHeader.TitleArea: The main title area of the page. Visible on all viewports.
+// PageHeader.TitleArea Sub Components: PageHeader.BackButton, PageHeader.LeadingVisual, PageHeader.Title, PageTitle.TrailingVisual, PageHeader.TrailingActions, PageHeader.Actions
+// PageHeader.BackButton and PageHeader.TrailingAction are only visible on regular viewports therefore they come as hidden on narrow viewports and their visibility can be managed by their exposed `hidden` prop
+// ---------------------------------------------------------------------
+const TitleArea: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({children, sx = {}}) => {
+  return (
+    <Box
+      sx={merge<BetterSystemStyleObject>({display: 'flex', flexDirection: 'row', alignItems: 'center', gap: '8px'}, sx)}
```
broccolinisoup, line 146: You mentioned the actions should be aligned to top when title is multiline where cases like this. (Attaching the screenshot you shared)

Screen Shot 2022-11-22 at 18 12 38

Where do you think the rest of the items align for cases like this? For example if they have `LeadingVisual` and/or `TrailingVisual`?

src/PageHeader/examples.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+)
+
+export const WebHooks = () => (
+  <PageHeader sx={{padding: 3}}>
+    <PageHeader.ContextArea>
+      <PageHeader.ParentLink href="http://github.com">Repository Settings</PageHeader.ParentLink>
+    </PageHeader.ContextArea>
+    <PageHeader.TitleArea sx={{paddingTop: 2, paddingX: 1}}>
```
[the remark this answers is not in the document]
broccolinisoup: This was just my sketchy stories, sorry! 
I removed all of the sx prop from the subcomponents. I added `8px` gap between each main slot (ContextArea, TitleArea, DescriptionArea, NavigationArea) and gave `16px` padding around the parent `PageHeader` component for narrow viewports and `24px` for the regular and wide. Does that sound right at all?

src/PageHeader/examples.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <Hidden on={['regular', 'wide']}>
+          <Button variant="primary">New</Button>
+        </Hidden>
+      </PageHeader.Actions>
+    </PageHeader.TitleArea>
+  </PageHeader>
+)
+
+export const PullRequestPage = () => (
+  <PageHeader sx={{padding: 2, paddingTop: 3}}>
```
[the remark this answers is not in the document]
broccolinisoup: I replied above 😊

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,376 @@
...
+  children,
+  visible = onlyVisibleOnNarrowView,
+  sx = {}
+}) => {
+  const isVisible = useResponsiveValue(visible, false)
+  const contentNavStyles = {
+    display: isVisible ? 'flex' : 'none',
+    flexDirection: 'row',
+    alignItems: 'center',
+    gap: '8px',
```
joshblack, line 76: Super random, do we need to use a spacing token like from: https://primer.style/primitives/spacing#size for gaps or are those not available yet?

Separately, should the unit here be `rem` instead of `px` or do we use `px` for layouts?

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,376 @@
...
+}
+export type ParentLinkProps = PageHeaderProps & LinkProps
+
+const ParentLink = React.forwardRef<HTMLAnchorElement, ParentLinkProps>(
+  (
+    {
+      children,
+      sx = {},
+      href,
+      'aria-label': ariaLabel = `Back to ${children}`,
```
joshblack, line 101: I would have expected the content (`children`) to be the label of the `<a>`, is there a specific reason for the overriding `aria-label` in this scenario? Is it that we always want to prepend `Back to` before the text?

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,376 @@
...
+    padding: useResponsiveValue(
+      {
+        narrow: 3,
+        regular: 4,
+        wide: 4
+      },
+      false
+    )
+  }
+  return <Box sx={merge<BetterSystemStyleObject>(rootStyles, sx)}>{children}</Box>
```
joshblack, line 58: Would we ever need to support an `as` prop in order to support semantic landmarks like `<header>` or is the intent that this would not be a landmark on the page?

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// ContextAreaActions
+// ---------------------------------------------------------------------
+const ContextAreaActions: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({
+  children,
+  sx = {},
+  visible = onlyVisibleOnNarrowView
+}) => {
+  const isVisible = useResponsiveValue(visible, false)
```
joshblack, line 156: For visibility in layout, would it make sense to shift this kind of logic into CSS with breakpoints so that we don't have to do any extra work in SSR to predict the size and avoid a flash of change when JS is initialized?

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,376 @@
...
+  // Check if we need responsive values for heading is so should we update as prop's type for Heading component?
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+} & PageHeaderProps
+
+const Title: React.FC<React.PropsWithChildren<TitleProps>> = ({
+  children,
+  sx = {},
+  variant = 'medium',
+  visible = true,
+  as = 'h3'
```
joshblack, line 238: When would you say we should use the `as` prop versus something like:

```
<PageHeader.Title>
  <h3>Pull Request</h3>
</PageHeader.Title>
```

That lets them compose through `children`? Would it be that we want it as a prop because it _must_ be a heading of a specific level? I also think we should be careful about default heading levels and that it might help to make the level explicit instead of potentially providing an inappropriate default.

joshblack (commented):
Sorry for the delay @broccolinisoup! Left a couple of comments/notes, also had a question around alignment, should the underline nav align here?

image

src/PageHeader/PageHeader.tsx
broccolinisoup: This one is on the [Figma Specs ](https://www.figma.com/file/Ee0OrXuOLXMDqUW83EnDHP/PageHeader-(FY23-Q1)?node-id=2%3A7403&t=a2tLMyUcq6SHAR87-0) but we could definitely discuss its necessity. I was thinking its purpose to give more context ([example](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8)) but open to feedback as always 🙏🏼

src/PageHeader/PageHeader.tsx
broccolinisoup: That makes so much sense! The way I implemented them is very much inspired by PageLayout and considering PageHeader will be used within PageLayout (I think most of the time) so it is important that we stay consistent for the DX. Having all said that, I see @colebemis has a [TODO](https://github.com/primer/react/blob/main/src/hooks/useResponsiveValue.ts#L48) for the `useResponsiveValue` to bring SSR support

I am happy to work on that enhancement. 

Are we thinking something like that for the SSR support or something different?

```
 // for narrow viewport
  '@media (max-width: 768px)': {
    display: hidden.narrow ? 'none' : 'flex'
  },
  // for regular viewport
  '@media (min-width: 768px) and (max-width: 1440px)': {
    display: hidden.regular ? 'none' : 'flex'
  },
  // for wide viewport
  '@media (min-width: 1440px)': {
    display: hidden.wide ? 'none' : 'flex'
  },

```

We could definitely optimise and make it reusable but I just wanted to gauge the feeling before going too further?

src/PageHeader/PageHeader.tsx
broccolinisoup: I like that approach! I completely agree on being careful about the default level and considering cases where the title is actually not a title (see this [example](https://github.com/primer/react/pull/2265#discussion_r1029967215)), we could benefit from this approach. Thanks!

src/PageHeader/PageHeader.tsx
broccolinisoup: I think that makes sense. PageHeader can be a part of PageHeader (it could even be most of the time) but I agree, we should give an option where it needs to be the `header` landmark. Thanks!

broccolinisoup:
> Sorry for the delay @broccolinisoup! Left a couple of comments/notes, also had a question around alignment, should the underline nav align here?
> 
> image

The spacing is still very arbitrary here. I'll get it cleared with design too.

src/PageHeader/PageHeader.tsx
colebemis: Media queries vs the `useResponsiveValue` hook is a tradeoff between performance and developer ergonomics. In cases where we can use media queries easily, we should do that. In other cases, we can fall back to the `useResponsiveValue` hook. Does that make sense?

Re: SSR support for `useResponsiveValue` - I'm unsure how to make SRR work and keep the same API. Let me know if you have any ideas!

colebemis:
```
 // for narrow viewport
  '@media (max-width: 768px)': {
    display: hidden.narrow ? 'none' : 'flex'
  },
  // for regular viewport
  '@media (min-width: 768px) and (max-width: 1440px)': {
    display: hidden.regular ? 'none' : 'flex'
  },
  // for wide viewport
  '@media (min-width: 1440px)': {
    display: hidden.wide ? 'none' : 'flex'
  },

```

A couple notes about this code snippet:

- The `regular` range includes the `wide` range too. So the `regular` media query would be `@media (min-width: 768px)`. [Reference](https://github.com/github/primer/blob/main/adrs/2022-04-15-responsive-design-api-guidelines.md#2-update-nomenclature-for-creating-responsive-components-and-web-pages)
- This would cause unexpected results if the user only specified `narrow` and `regular` keys (not `wide`): `hidden={{regular: true, narrow: false}}`. In that code snippet, `display` would be set to `flex` in the wide viewport. But the intended outcome should be that the component is still hidden on wide viewports. Does that make sense?

src/PageHeader/PageHeader.tsx
broccolinisoup: Sorry for not answering to everything you mentioned! I use the `get` function to get the spacing from the "primitives" (from theme-preval to be precise). 

Re `rem` vs `px` I am not sure, it makes sense to use rem but I couldn't find any design guidelines around units. I am a bit confused how I could apply `rem` using the values from "theme-preval" as well.  I'll check this with design first to see how do we want the spaces to be.

src/PageHeader/examples.stories.tsx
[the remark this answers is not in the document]
broccolinisoup: @mperrotti could I please get your thoughts on that? 🙏🏼

src/PageHeader/PageHeader.stories.tsx
broccolinisoup: @mperrotti could I please get your thoughts on that? 🙏🏼

src/PageHeader/PageHeader.tsx
broccolinisoup: @mperrotti could I please get your thoughts on that? 🙏🏼

src/PageHeader/PageHeader.tsx
broccolinisoup: @mperrotti could I please get your thoughts on that? 🙏🏼

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
mperrotti: I interpreted @vdepizzol 's comment the same way.

src/PageHeader/PageHeader.stories.tsx
mperrotti: I'm good with the `hasContextArea`, `hasParentLink`, `hasLeadingVisual` etc approach

src/PageHeader/PageHeader.tsx
mperrotti: We'd have to write some custom JS to detect if the lines wrap or not. I tried `align-items: baseline`, and it seems to work well for single line and multi-line titles.

https://user-images.githubusercontent.com/2313998/206012763-9bea8126-950c-48ab-a902-c6e5bcd777d2.mp4

mperrotti: The only problem is that the `LeadingVisual` looks a little too high up with `align-items: baseline` on a single line.

I think we need a little more time figuring out which layout styles will give us the best layout for all cases. Maybe we can try playing with the line-height or the top and bottom padding of the icon container?


https://user-images.githubusercontent.com/2313998/206014747-3711f5d6-937d-4789-9447-53be17a9db51.mp4

src/PageHeader/PageHeader.stories.tsx
mperrotti: I think the Octicons work well and cover a majority of the use cases for a PageHeader's LeadingVisual

src/PageHeader/PageHeader.tsx
mperrotti: Curious to hear what a11y folks think about this. Tagging this week's a11y FR @alexislucio

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+const Description: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({children, sx = {}, hidden = false}) => {
+  const isHidden = useResponsiveValue(hidden, true)
+  return (
+    <Box
+      sx={merge<BetterSystemStyleObject>(
+        {
+          display: isHidden ? 'none' : 'flex',
+          flexDirection: 'row',
+          alignItems: 'center',
+          gap: get('space.2'),
```
mperrotti, line 346: I think we should use `get('stack.gap.condensed')` (from [UI stack tokens](https://primer.style/primitives/spacing#ui-stack))

cc @langermank

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+        <Link
+          ref={ref}
+          as={as}
+          aria-label={ariaLabel}
+          muted
+          sx={merge<BetterSystemStyleObject>(
+            {
+              display: isHidden ? 'none' : 'flex',
+              alignItems: 'center',
+              // min touch target size 44px (20 + (12x2))
```
mperrotti, line 144: I think we have tokens for touch target size. cc @langermank

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+              // TODO: rem for line height?
+              // TODO: no 12px in theme - is that okay?
```
mperrotti, line 146: If they in fact exist, we might be able to just set dimensions using the minimum touch target size tokens.

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+          lineHeight: {
+            large: '48px',
+            medium: '32px',
+            subtitle: '32px',
+          }[currentVariant],
+          fontWeight: {
+            large: '400',
+            medium: '400',
+            subtitle: '400',
+          }[currentVariant],
```
mperrotti, line 281: Let's use [typography design tokens](https://primer.style/primitives/typography) instead of hard-coded strings. We can pair if you're having a tough time figuring out which tokens to use.

src/PageHeader/PageHeader.tsx
langermank: Everything in primitives is in rem units now, but we're still testing how this behaves in dotcom– that's why we don't have any official guidance on it yet. Its very safe to use rem units though, so I would go for it! 

We're also not clear yet on the CSS strategy for PRC, so I hesitate to start integrating the new sizing tokens in. We're still making breaking changes to the names and might switch to using CSS variables next quarter in React. I would just use their value in rem units for now until we can roll out primitives to PRC.

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+  'aria-label'?: React.AriaAttributes['aria-label']
+  // TODO: check if we need any other types
+  as?: React.ElementType | 'header' | 'div'
+} & sharedPropTypes
+
+const Root: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({children, sx = {}, as = 'div'}) => {
+  const rootStyles = {
+    display: 'flex',
+    flexDirection: 'column',
+    gap: get('space.2'),
```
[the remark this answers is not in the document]
langermank: Commented above, but I think just sticking to the raw values for now is safest as we're about to make some breaking changes to the names.

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+  narrow: true,
+  regular: false,
+  wide: false,
+}
+
+// Root
+// -----------------------------------------------------------------------------
+export type PageHeaderProps = {
+  'aria-label'?: React.AriaAttributes['aria-label']
+  // TODO: check if we need any other types
```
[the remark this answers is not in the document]
broccolinisoup: I plan to take them into tasks to address in different PRs. I'll remove them today.

mperrotti:
@broccolinisoup - Katie already commented, but ignore any of my comments that suggest using any of the tokens from these two pages:
- https://primer.style/primitives/spacing
- https://primer.style/primitives/typography

There's an [ADR PR](https://github.com/github/primer/pull/1545) open that will introduce breaking changes when the names change. We can come back to these styles and use the correct tokens once they're ready.

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+  as?: React.ElementType | 'header' | 'div'
+} & sharedPropTypes
+
+const Root: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({children, sx = {}, as = 'div'}) => {
```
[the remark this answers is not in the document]
broccolinisoup: Oh great question! As far as I understand that wouldn't be valid in React 18 because they are removing implicit children type declaration in React.FC

Let me know if I missed anything! 

Resources might be helpful
-changelog: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/46691
- a blog post of one of the maintainers of DefinitelyTyped: https://solverfox.dev/writing/no-implicit-children/

src/PageHeader/PageHeader.tsx
broccolinisoup: Oh thank you! That is very helpful.

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+      },
+    }
+  } else {
+    return {display: fallback ? 'none' : 'flex'}
+  }
+}
+
+// PageHeader.ContextArea : Only visible on narrow viewports by default to provide user context of where they are at their journey. `visible` prop available
+// to manage their custom visibility but consumers should be careful if they choose to hide this on narrow viewports.
+// PageHeader.ContextArea Sub Components: PageHeader.ParentLink, PageHeader.ContextBar, PageHeader.ContextAreaActions
```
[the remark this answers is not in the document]
broccolinisoup: Sure! I'll start writing them soon!

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+    ),
+  }
+  return (
+    <Box as={as} sx={merge<BetterSystemStyleObject>(rootStyles, sx)}>
+      {children}
+    </Box>
+  )
+}
+
+function displayResponsively<T, F>(value: T, fallback: F): BetterSystemStyleObject {
```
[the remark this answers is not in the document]
broccolinisoup: Ahh, sorry I forgot to mention about it. It is an experiment to use CSS breakpoints rather than relying on using `useResponsiveHook` and it comes from this [convo] and I only used it for one sub component. (https://github.com/primer/react/pull/2265#discussion_r1036355747) I am actually thinking to tackle this in a different PR, this one is already big enough 😆

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+      [`@media screen and (max-width: 768px)`]: {
+        display: 'narrow' in responsiveValue ? (responsiveValue.narrow ? 'none' : 'flex') : fallback ? 'none' : 'flex',
+      },
+      [`@media screen and (min-width: 768px)`]: {
+        display:
+          'regular' in responsiveValue ? (responsiveValue.regular ? 'none' : 'flex') : fallback ? 'none' : 'flex',
+      },
+      [`@media screen and (min-width: 1440px)`]: {
+        display: 'wide' in responsiveValue ? (responsiveValue.wide ? 'none' : 'flex') : fallback ? 'none' : 'flex',
+      },
```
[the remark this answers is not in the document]
broccolinisoup: I think this could consider within the [suggestion](https://github.com/primer/react/pull/2265#discussion_r1041511281) as well, so keeping the hard-coded values for now. Let me know if I misunderstand anything!

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+// ContextBar
+// Generic slot for any component above the title region. Use it for custom breadcrumbs and other navigation elements instead of ParentLink.
+// ---------------------------------------------------------------------
+
+const ContextBar: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({
+  children,
+  sx = {},
+  hidden = hiddenOnRegularAndWide,
+}) => {
+  const isHidden = useResponsiveValue(hidden, false)
```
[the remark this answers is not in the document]
broccolinisoup: Agreed! I left a comment about it [here](https://github.com/primer/react/pull/2265/files#r1041551464) as well

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+  return (
+    <Box
+      sx={merge<BetterSystemStyleObject>(
+        {
+          display: isHidden ? 'none' : 'flex',
+          flexDirection: 'row',
+          alignItems: 'center',
+          gap: get('space.2'),
```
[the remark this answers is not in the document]
broccolinisoup: As per suggested [here](https://github.com/primer/react/pull/2265#discussion_r1041511281), I used "0.5rem" for all these

src/PageHeader/PageHeader.tsx
broccolinisoup: As per suggested [here](https://github.com/primer/react/pull/2265#discussion_r1041511281), I'll keep the hard-coded rem here but I am confused about rem values [here](https://primer.style/primitives/typography) for line-height. The output values seem to be like `calc(48/32)` (for large text) would that mean I would use `1.5rem` for large text line-height?

src/PageHeader/examples.stories.tsx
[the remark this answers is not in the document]
broccolinisoup: Ahh that makes perfect sense to me! Let me clarify. Not having spacing around the PageHeader root component will prevent us from having misalignment with the PageHeader siblings and you are suggesting to remove that spacing around and let consumers manage their spacing on the parent. Did I understand correctly? @mperrotti

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
broccolinisoup: I just checked and seems like the weight is 600 now for the large texts? https://primer.style/primitives/typography#styles cc @langermank @lukasoppermann

src/PageHeader/PageHeader.tsx
broccolinisoup: Ah thanks for the exploration, it is great and super helpful! Let me plan with it today and see what I'll come up with!

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,387 @@
...
+  const rootStyles = {
+    display: 'flex',
+    flexDirection: 'column',
+    gap: get('space.2'),
+    // 24px for wide and regular, 16px for narrow viewports
+    padding: useResponsiveValue(
+      {
+        narrow: 3,
+        regular: 4,
+        wide: 4,
```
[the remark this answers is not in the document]
broccolinisoup: Actually no, we don't! I just realised, thanks for pointing this out 🙌🏼

src/PageHeader/PageHeader.tsx
broccolinisoup: @mperrotti ^^

src/PageHeader/examples.stories.tsx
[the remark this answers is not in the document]
mperrotti: Yup, you understood correctly 👍

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
mperrotti: Oops - I forgot about that. Ignore what I said!

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
mperrotti: If we're going to tackle this in a separate PR, do you think we should create an issue for that work once this merges?

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
mperrotti: I think you're right - we'll wait until we have the new Primitive tokens.

src/PageHeader/PageHeader.tsx
mperrotti: No, I think we'd just use `1.5` because line-heights can be unit-less. The `rem` unit refers to the *root* font size (the font size set on `<html>`), so instead of `1.5 * {font size of element}` it would be `1.5 * {font size set on the document root}`

broccolinisoup: Ohh that makes prefect sense!! I was thinking taking the root font-size as an anchor to proportionally scale the line-height but no we should anchor the font-size of the element. Thanks for clarifying 💖

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
broccolinisoup: Agreed! https://github.com/github/primer/issues/1569 done!

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
broccolinisoup: I updated the large and medium to use `600` as per suggested [here](https://primer.style/primitives/typography#styles) let me know if I misunderstood!

src/PageHeader/PageHeader.tsx
broccolinisoup: I used only line-height to manage the min touch target size (2.75 rem). Let me know if you have any concerns!

src/PageHeader/PageHeader.tsx
broccolinisoup: I played with it and giving height to each sub-component's container `height:2rem; alignItems: 'center'` along with giving `alignItems: 'flex-start'` to the main TitleArea container seems to be looking as expected. Here are a few screenshots to show the alignment

## Single Line
Screen Shot 2022-12-07 at 1 46 35 pm

## Multi Line
Screen Shot 2022-12-07 at 1 46 15 pm

## When increased the browser's default font-size
Screen Shot 2022-12-07 at 1 46 51 pm

and here is the commit for these changes : https://github.com/primer/react/pull/2265/commits/98bdc6acd7d91fb500bdddab2814a6daf7131203

What do you think @mperrotti ? 👀

src/PageHeader/PageHeader.tsx
broccolinisoup: I'll tackle this on a different PR @joshblack , if that is okay? I created an [issue](https://github.com/github/primer/issues/1571) for it. Let me know if you have any concerns.

src/PageHeader/PageHeader.tsx
broccolinisoup: I'll tackle this is a separate PR. I created an [issue](https://github.com/github/primer/issues/1569) for this 🙌🏼

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,375 @@
...
+        <Link
+          ref={ref}
+          as={as}
+          aria-label={ariaLabel}
+          muted
+          sx={merge<BetterSystemStyleObject>(
+            {
+              display: isHidden ? 'none' : 'flex',
+              alignItems: 'center',
+              lineHeight: '2.75rem', // min touch target size 44px (16*2.75)
```
mperrotti, line 113: I don't actually think we need this. Our buttons don't have a minimum height of 44px, and they're still  considered to be accessible.

src/PageHeader/PageHeader.tsx
mperrotti: This direction seems good to me. Did you choose `2rem` based on the title font size and line height? If so, I'd store `2rem` in a variable and use that variable for setting the heights. If there was another reason you chose `2rem`, what was it?

Also - we'll need to change `2rem` to match the font size. See what happens when I set the title variant to large:
Screenshot 2022-12-07 at 7 34 52 PM

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,375 @@
...
+      as={as}
+      sx={merge<BetterSystemStyleObject>(
+        {
+          fontSize: {
+            large: '2rem',
+            medium: '1.25rem',
+            subtitle: '1.25rem',
+          }[currentVariant],
+          lineHeight: {
+            large: 1.5, // calc(48/32)
```
mperrotti, line 248: What are these `calc`s from?

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,375 @@
...
+  const isHidden = useResponsiveValue(hidden, false)
+  return (
+    <Heading
+      as={as}
+      sx={merge<BetterSystemStyleObject>(
+        {
+          fontSize: {
+            large: '2rem',
+            medium: '1.25rem',
+            subtitle: '1.25rem',
```
mperrotti, line 245: The typography looks identical for `medium` and `subtitle`. Is this intentional?

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,375 @@
...
+    display: isHidden ? 'none' : 'flex',
+    flexDirection: 'row',
+    alignItems: 'center',
+    gap: '0.5rem',
+    order: REGION_ORDER.ContextArea,
+  }
+  return <Box sx={merge<BetterSystemStyleObject>(contentNavStyles, sx)}>{children}</Box>
+}
+// adopted from React.AnchorHTMLAttributes
+type LinkProps = {
```
mperrotti, line 76: Is there a reason we can't do this instead?
```
type LinkProps = Pick<
  React.AnchorHTMLAttributes,  'download' |
  'href' |
  'hrefLang' |
  'media' |
  'ping' |
  'rel' |
  'target' |
  'type' |
  'referrerPolicy'
>
```

(non-blocking feedback)

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
mperrotti: The link to your comment wasn't working for me 🤔 

I see we're still using `useResponsiveValue`. Is this not possible to replace with a media query?

mperrotti (commented):
Looking good! I think we're sooo close!

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,375 @@
...
+export type PageHeaderProps = {
+  'aria-label'?: React.AriaAttributes['aria-label']
+  as?: React.ElementType | 'header' | 'div'
+} & sharedPropTypes
+
+const Root: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({children, sx = {}, as = 'div'}) => {
+  const rootStyles = {
+    display: 'flex',
+    flexDirection: 'column',
+    gap: '0.5rem',
```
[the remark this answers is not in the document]
broccolinisoup: Good idea, done!

src/PageHeader/PageHeader.tsx
broccolinisoup: When I remove this, its line-height becomes 1.5 (inherits from the parent) and would still be okay for accessibility, you think?

src/PageHeader/PageHeader.tsx
broccolinisoup: I think they are meant to be `calc(height/font-size)` and got them from [primitives](https://primer.style/primitives/typography). If they are rather confusing than being helpful, happy to remove them.

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
langermank: @broccolinisoup I chatted with @vdepizzol, it _should_ be 400– I'm going to update the design token to reflect that change.

src/PageHeader/PageHeader.tsx
broccolinisoup: I don't think so. It seemed a bit of a convention to me in the code base so I copied over. I updated now like below

```
type LinkProps = Pick<
  React.AnchorHTMLAttributes<HTMLAnchorElement>,
  'download' | 'href' | 'hrefLang' | 'media' | 'ping' | 'rel' | 'target' | 'type' | 'referrerPolicy'
>
```

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
broccolinisoup: Thanks for the follow up @langermank! Does that mean all (large, medium and subtitle) will have 400 for `font-weight`?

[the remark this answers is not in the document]
broccolinisoup: If so, I echo @mperrotti 's comment [here](https://github.com/primer/react/pull/2265#discussion_r1042811678) that subtitle and medium will have the same typography? 🤔

src/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,393 @@
...
+    >
+      {children}
+    </Box>
+  )
+}
+
+const MEDIUM_TITLE_HEIGHT = '2rem'
+const LARGE_TITLE_HEIGHT = '3rem'
+
+const TitleAreaContext = React.createContext<{
```
broccolinisoup, line 164: I cannot seem to reply your [comment](https://github.com/primer/react/pull/2265#discussion_r1042809668) so I'll write here.

>Did you choose 2rem based on the title font size and line height? 

Yes! 🙌🏼 

>If so, I'd store 2rem in a variable and use that variable for setting the heights

Great idea, done!

>Also - we'll need to change 2rem to match the font size

This is a great catch! Sorry I missed that. I thought about how to solve this and tried moving the `variant` prop to `TitleArea` (parent of `Title`) from `Title`. I think lifting the state helped both technically.

I leveraged using React context and easily shared the data between children. Creating a React context for the TitleArea and `variant` into the context and use this context in each child (`LeadingAction`,`LeadingAction`,`TrailingVisual`,`TrailingAction`) to manage children's height. 

I think moving `variant` to `TitleArea` feels more intuitive to me. Looking forward to hearing what you think @mperrotti! 

There is one thing that concerns me though and it is the sizes of the visuals or icon buttons. Should we increase their font-sizes too? 🤔

mperrotti: > There is one thing that concerns me though and it is the sizes of the visuals or icon buttons. Should we increase their font-sizes too? 🤔

I don't think anything should affect the button size except the button's `size

src/PageHeader/PageHeader.tsx
mperrotti: I think either way works. I wasn't aware the original version was an established convention in the codebase

src/PageHeader/PageHeader.tsx
mperrotti: I'd just add a comment noting that they're from @primer/primitives

src/PageHeader/PageHeader.tsx
mperrotti: Yup, I think that's fine 👍

mperrotti (approved):
🎉

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
mperrotti: I think the difference would be that `subtitle` is 400 and `medium` is 600. I think this is good to go if we've changed it to 400

PR #2272, opened 2022-08-25
PageHeader API

pksjce:
This PR documents `PageHeader` proposed API. The implementation PR will be https://github.com/primer/react/pull/2265 where the goal is to get the same api as documented here working inside storybooks. I will modify implementation based on feedback here. This PR needs to be merged first before implementation PR.

Link to PageHeader API in figma - https://www.figma.com/file/Ee0OrXuOLXMDqUW83EnDHP/PageHeader-(FY23-Q1)?node-id=2%3A80694
This is the proposed anatomy that seems most complete to me
image

Here are the corresponding figma shots for the three API examples. Together I think this gives us a decent look at the API.

### Default

image

### With Context Actions and Breadcrumbs
image

### With a description and local navigation
image

I'm looking for specific feedback on -
- API does not communicate that `PageHeader.ContextNav` will only be visible in narrow mode.
- The ADR on responsive design proposes a `Stack` component. I feel like my usecase needs this component. It's not developed yet. I'm planning on making it. WDYT?
- Naming of components and the depth of the component tree (is it too much?)

### DIFF
--- a/docs/content/PageHeader.mdx
+++ b/docs/content/PageHeader.mdx
@@ -25,13 +25,13 @@ import {PageHeader} from '@primer/react'
     <PageHeader.Title>Branches</PageHeader.Title>
     <PageHeader.TitleActions>
       {/* Visible only in narrow viewport*/}
-      <WhenNarrow>
+      <Stack direction={{narrow: 'inline', regular: 'none'}}>
         <Button variant="primary">New Branch</Button>
-      </WhenNarrow>
+      </Stack>
       {/* Visible only in regular viewport*/}
-      <WhenRegular>
+      <Stack direction={{narrow: 'none', regular: 'inline'}}>
         <Button variant="primary">New</Button>
-      </WhenRegular>
+      </Stack>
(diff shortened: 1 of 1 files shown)


docs/content/PageHeader.mdx
```diff
@@ -0,0 +1,114 @@
...
+      {/* Visible in all viewports */}
+      <IconButton aria-label="More actions" icon={KebabHorizontalIcon} />
+    </PageHeader.TitleActions>
+  </PageHeader.TitleArea>
+</PageHeader>
+```
+
+### With Context Actions and Breadcrumbs
+
+<PageHeader>
```
mperrotti, line 44: This example and the following examples don't have the `jsx live` codefence args. Is this intentional?

pksjce: nope! it was a mistake. Will fix this 👍

pksjce:
Closing this as I've already opened a PR for primer repo

PR #2276, opened 2022-08-25
test: update jest transformIgnorePatterns config

### LINKED ISSUE #2275: Catch ESM-only modules before publishing
This issue tracks our testing strategy in order to make sure that packages which we depend on ship support for both ESM and CommonJS.

After looking at the project initially, it seems like reverting the changes to this field will give us a strong signal: https://github.com/primer/react/blob/main/jest.config.js#L13

Specifically, this field should only include `'/node_modules/'`. Excluding other packages from this list is a good signal that this package ships an ESM-only package.

**Limitations**

The limitation of relying on this field to catch this problem is that if a code path is not tested through Jest then it will not be caught since we rely on Jest to throw an error when CommonJS is not supported from a package in `node_modules`

joshblack:
Closes #2275

This PR updates our `transformIgnorePatterns` config in Jest to not exclude modules in `node_modules`. Typically, excluding packages through this field is an indicator that the package only ships with support for ESM. Using the default value for this field allows us to catch if a package is not shipping with support for ESM before we publish.

Note: the tests for this PR are expected to fail since we currently rely on packages that ship ESM-only bundles. These should be addressed before merging this PR in.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,7 +9,10 @@ module.exports = {
     '<rootDir>/src/utils/test-helpers.tsx'
   ],
   testMatch: ['<rootDir>/(src|codemods)/**/*.test.[jt]s?(x)', '!**/*.types.test.[jt]s?(x)'],
-  transformIgnorePatterns: [
-    'node_modules/(?!@github/combobox-nav|@koddsson/textarea-caret|@github/markdown-toolbar-element)'
-  ]
+
+  // Note: this configuration option should not exclude any packages within
+  // `node_modules`. When excluding packages from `node_modules`, we have no way
+  // to test if this package is shipping with support for both ESM and CommonJS
+  // @see https://github.com/primer/react/issues/2275
+  transformIgnorePatterns: ['/node_modules/']
 }


pksjce:
Nice that it already caught one of the incompatible packages.
Seems like the fix required for tests should be in github-combonav?

joshblack:
@pksjce yes! Exactly

pksjce:
Though, I think they put those folders in the ignore list on purpose? Since we already encounter esm challenges with them. I think @iansan5653 knows about this in depth.

joshblack:
@pksjce I think you're dead on, the packages in the list currently only ship ESM modules and so they are included in the ignore pattern in order to be transformed by babel into CommonJS for Jest.

I think that removing them from this list is a quick way to check for the presence of ESM in our setup, it would also allow us to write tests like:

```js
test('CommonJS', () => {
  expect(() => {
    require('@primer/react');
  }).not.toThrow();
});
```

Since Jest wouldn't include any dependencies of `@primer/react` in the babel transform.

Another option we could take is similar to the consumer test where we have a standalone, e2e-like test that explicitly tries to `require('@primer/react')`

PR #2277, opened 2022-08-26
UnderlineNav with updated design for states and counter

broccolinisoup:
Closes [#1240](https://github.com/github/primer/issues/1240)

UnderlineNav has updated design and this PR is for the updated states (default, selected, hover and focus state) as well as adding counter as a prop.

### Screenshots 

Screen Shot 2022-09-01 at 3 13 03 pm
Screen Shot 2022-09-01 at 3 12 32 pm

### Screen Recording

https://user-images.githubusercontent.com/1446503/187836734-e7d1c516-5ab2-445c-83d4-bf20b9995f51.mp4

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/thirty-mayflies-travel.md
+++ b/.changeset/thirty-mayflies-travel.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': major
+---
+
+UnderlineNav.Link renamed to UnderlineNav.Item along with updated styles
--- a/.changeset/thirty-mayflies-travel.md
+++ b/.changeset/thirty-mayflies-travel.md
@@ -1,5 +1,5 @@
 ---
-'@primer/react': major
+'@primer/react': minor
 ---
 
 UnderlineNav.Link renamed to UnderlineNav.Item along with updated styles
--- a/src/UnderlineNav2/UnderlineNav.tsx
+++ b/src/UnderlineNav2/UnderlineNav.tsx
@@ -5,6 +5,8 @@ import {UnderlineNavContext} from './UnderlineNavContext'
 import {ActionMenu} from '../ActionMenu'
 import {ActionList} from '../ActionList'
 import {useResizeObserver, ResizeObserverEntry} from '../hooks/useResizeObserver'
+import {useFocusZone} from '../hooks/useFocusZone'
+import {FocusKeys} from '@primer/behaviors'
 
 type Overflow = 'auto' | 'menu' | 'scroll'
 type ChildWidthArray = Array<{width: number}>
@@ -84,6 +86,16 @@ export const UnderlineNav = forwardRef(
   ) => {
     const backupRef = useRef<HTMLElement>(null)
     const newRef = (forwardedRef ?? backupRef) as MutableRefObject<HTMLElement>
+
+    // This might change if we decide tab through the navigation items rather than navigationg with the arrow keys.
+    // TBD. In the meantime keeping it as a menu with the focus trap.
+    // ref: https://www.w3.org/WAI/ARIA/apg/example-index/menubar/menubar-navigation.html (Keyboard Support)
+    useFocusZone({
+      containerRef: backupRef,
+      bindKeys: FocusKeys.ArrowHorizontal | FocusKeys.HomeAndEnd,
+      focusOutBehavior: 'wrap'
+    })
+
     const styles = {
       display: 'flex',
       justifyContent: align === 'right' ? 'flex-end' : 'space-between',
@@ -144,7 +156,7 @@ export const UnderlineNav = forwardRef(
       <UnderlineNavContext.Provider
         value={{setChildrenWidth, selectedLink, setSelectedLink, afterSelect: afterSelectHandler, variant}}
       >
-        <Box as={as} sx={merge(styles, sxProp)} aria-label={label} ref={newRef}>
+        <Box tabIndex={0} as={as} sx={merge(styles, sxProp)} aria-label={label} ref={newRef}>
           <Box as="ul" sx={merge<BetterSystemStyleObject>(overflowStyles, ulStyles)}>
             {responsiveProps.items}
           </Box>
--- a/src/UnderlineNav2/UnderlineNav.tsx
+++ b/src/UnderlineNav2/UnderlineNav.tsx
@@ -87,6 +87,8 @@ export const UnderlineNav = forwardRef(
     const styles = {
       display: 'flex',
       justifyContent: align === 'right' ? 'flex-end' : 'space-between',
+      borderBottom: '1px solid',
+      borderBottomColor: 'border.muted',
       align: 'row',
       alignItems: 'center'
     }
--- a/src/UnderlineNav2/UnderlineNav.tsx
+++ b/src/UnderlineNav2/UnderlineNav.tsx
@@ -96,7 +96,8 @@ export const UnderlineNav = forwardRef(
       listStyle: 'none',
       padding: '0',
       margin: '0',
-      marginBottom: '-1px'
+      marginBottom: '-1px',
+      alignItems: 'center'
     }
 
     const [selectedLink, setSelectedLink] = useState<RefObject<HTMLElement> | undefined>(undefined)
--- a/src/UnderlineNav2/UnderlineNav.tsx
+++ b/src/UnderlineNav2/UnderlineNav.tsx
@@ -163,6 +163,7 @@ export const UnderlineNav = forwardRef(
 
           {actions.length > 0 && (
             <ActionMenu>
+              {/* set margin 0 here because safari puts extra margin around the button */}
               <ActionMenu.Button sx={{m: 0}}>More</ActionMenu.Button>
               <ActionMenu.Overlay>
                 <ActionList>
--- a/src/UnderlineNav2/UnderlineNav.tsx
+++ b/src/UnderlineNav2/UnderlineNav.tsx
@@ -87,8 +87,6 @@ export const UnderlineNav = forwardRef(
     const styles = {
       display: 'flex',
       justifyContent: align === 'right' ? 'flex-end' : 'space-between',
-      borderBottom: '1px solid',
-      borderBottomColor: 'border.muted',
       align: 'row',
       alignItems: 'center'
     }
--- a/src/UnderlineNav2/UnderlineNav.tsx
+++ b/src/UnderlineNav2/UnderlineNav.tsx
@@ -163,7 +163,7 @@ export const UnderlineNav = forwardRef(
 
           {actions.length > 0 && (
             <ActionMenu>
-              <ActionMenu.Button>More</ActionMenu.Button>
+              <ActionMenu.Button sx={{m: 0}}>More</ActionMenu.Button>
               <ActionMenu.Overlay>
                 <ActionList>
                   {actions.map((action, index) => {
--- a/src/UnderlineNav2/UnderlineNavItem.tsx
+++ b/src/UnderlineNav2/UnderlineNavItem.tsx
@@ -99,7 +99,6 @@ export const UnderlineNavItem = forwardRef(
       textAlign: 'center',
       textDecoration: 'none',
       paddingX: 1,
-      marginRight: 3,
       ...(variant === 'small' ? smallVariantLinkStyles : defaultVariantLinkStyles),
       '&:hover > div[data-component="wrapper"] ': {
         backgroundColor: 'neutral.muted'
@@ -114,7 +113,7 @@ export const UnderlineNavItem = forwardRef(
 
     const borderStyles = {
       // How to use primer primitives for space 3?
-      width: `calc(100% - 16px)`,
+      width: `calc(100% - 8px)`,
       height: 2,
       backgroundColor: 'primer.border.active'
     }
@@ -146,7 +145,7 @@ export const UnderlineNavItem = forwardRef(
       [onSelect, afterSelect, ref, setSelectedLink]
     )
     return (
-      <Box as="li" sx={{display: 'grid'}}>
+      <Box as="li" sx={{display: 'flex', flexDirection: 'column', alignItems: 'center'}}>
         <Box
           as={Component}
           href={href}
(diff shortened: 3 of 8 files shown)


src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,51 @@ export const UnderlineNavLink = forwardRef(
     const textStyles: BetterSystemStyleObject = {
       whiteSpace: 'nowrap'
     }
+
+    const wrapperStyles = {
```
broccolinisoup, line 81: I added another Box (div) that wraps icon, text and the counter and gave some padding to achieve smaller grey rectangular hover state.
Screen Shot 2022-08-29 at 3 12 39 pm

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,51 @@ export const UnderlineNavLink = forwardRef(
...
+      display: 'inline-flex',
+      paddingY: 1,
+      paddingX: 2
+    }
     const smallVariantLinkStyles = {
-      paddingY: 2,
```
broccolinisoup, line 74: reduced the paddingY as a unit comes from the wrapper

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,51 @@ export const UnderlineNavLink = forwardRef(
...
-      borderBottom: '2px solid transparent',
-      borderColor: selectedLink === ref ? 'primer.border.active' : 'transparent',
       textDecoration: 'none',
-      paddingX: 2,
+      paddingX: 1,
       marginRight: 3,
       ...(variant === 'small' ? smallVariantLinkStyles : defaultVariantLinkStyles),
-      '&:hover, &:focus': {
-        borderColor: selectedLink === ref ? 'primer.border.active' : 'neutral.muted',
+      '&:hover > div[data-component="wrapper"] ': {
```
broccolinisoup, line 103: when anchor is hover, I added styles to the wrapper.

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,51 @@ export const UnderlineNavLink = forwardRef(
...
-      paddingY: 3,
+      paddingY: 2,
       fontSize: 1
     }
 
     const linkStyles = {
       display: 'inline-flex',
       color: 'fg.default',
       textAlign: 'center',
-      borderBottom: '2px solid transparent',
```
broccolinisoup, line 86: I had to remove the border entirely because of the focus style below. (The focus class on link element)

     &:focus': {
        outlineColor: 'accent.fg',
        borderRadius: 2,
        outlineOffset: '-6px',
        transition: '0.2s ease'
      }

The outline border can only be rounded with borderRadius. However when I added the borderRadius to the focus style, as this focus class gets triggered when **the item is selected as well**, it ended up rounding the orange border bottoms too. So I removed the borders from the links and add a standalone element and styled them to look like border bottom so that they won't affect from the focus styles. Please let me know what you think or if you have a different idea.

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -129,18 +157,26 @@ export const UnderlineNavLink = forwardRef(
...
+              </Box>
+            )}
+            {counter && (
+              <Box as="span" data-component="counter" sx={counterStyles}>
+                <CounterLabel>{counter}</CounterLabel>
+              </Box>
+            )}
+          </Box>
         </Box>
+        {selectedLink === ref && <Box as="span" sx={borderStyles}></Box>}
```
broccolinisoup, line 178: Standalone element to represent the border bottom.

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,51 @@ export const UnderlineNavLink = forwardRef(
...
+        outlineColor: 'accent.fg',
+        borderRadius: 2,
+        outlineOffset: '-6px',
         transition: '0.2s ease'
       }
     }
+
+    const borderStyles = {
+      // How to use primer primitives for space 3?
+      width: `calc(100% - 16px)`,
```
broccolinisoup, line 117: How to use primer primitives for space 3?

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,51 @@ export const UnderlineNavLink = forwardRef(
...
-      '&:hover, &:focus': {
-        borderColor: selectedLink === ref ? 'primer.border.active' : 'neutral.muted',
+      '&:hover > div[data-component="wrapper"] ': {
+        backgroundColor: 'neutral.muted',
+        borderRadius: 2
+      },
+      '&:focus': {
+        outlineColor: 'accent.fg',
+        borderRadius: 2,
+        outlineOffset: '-6px',
```
broccolinisoup, line 110: I don't think we can set X and Y outline offsets ([ref](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset)) , so this is the limitation came from using native outline.  (Not super close to the design -  see the x paddings are less in the design)

### Design
Screen Shot 2022-08-29 at 3 33 57 pm

### Storybook
Screen Shot 2022-08-29 at 3 40 39 pm

src/UnderlineNav2/index.ts
```diff
@@ -1,10 +1,10 @@
 import {UnderlineNav as Nav, UnderlineNavProps} from './UnderlineNav'
-import {UnderlineNavLink, UnderlineNavLinkProps} from './UnderlineNavLink'
+import {UnderlineNavItem, UnderlineNavItemProps} from './UnderlineNavItem'
 
 const UnderlineNav = Object.assign(Nav, {
-  Link: UnderlineNavLink
+  Item: UnderlineNavItem
```
pksjce, line 5: This would be a breaking change wrt to the old UnderlineNav. We will need to mention it in the changelog.

pksjce (commented):
The focus rings work so well! 💯 

I can see two things out of place.
1. The underline from the nav seems to have gone away.
2. The selected underline seems a bit longer on both sides compared to the Item.

image

Also I'm sceptical about the name change from UnderlineNav.Link to UnderlineNav.Item. It would break existing API. Though, I agree its a change that makes full sense. Not sure how to proceed with it honestly.

broccolinisoup:
Thanks for the review @pksjce ! 

> The underline from the nav seems to have gone away.

I used [the states reference](https://www.figma.com/file/N9JSJxBe6XAIeUjkvl4Bsy/UnderlineNav-Updates-Q1?node-id=8%3A12051) from Figma and there is no underline there so I though this line must have been removed with the updated design but for all the other references in Figma, the line is there! I added it back ✨  Sorry for the confusion! 

> The selected underline seems a bit longer on both sides compared to the Item.

Yes, you are right! I might have missed that. I also realised there was way too much margin between the li items. So all together, I re-styled them and this is for the reference to compare before and after. (I updated the PR description as well with the new screenshots)

Before

Screen Shot 2022-08-29 at 3 04 21 pm

After
Screen Shot 2022-08-30 at 4 11 22 pm

And the design

Screen Shot 2022-08-30 at 4 14 24 pm

> Also I'm sceptical about the name change from UnderlineNav.Link to UnderlineNav.Item. It would break existing API. Though, I agree its a change that makes full sense. Not sure how to proceed with it honestly.

Yeah I agree. It makes sense to me too and I see other "new" components such as ActionList uses Item too so it is good to be consistent. I guess we can discuss with others as well and find the best strategy to move forward with the major release.

danielguillan:
Sorry for the confusion about the removal of the underline. 

I initially intentionally removed it so with the removal of `actions,` consumers could easily build a custom UnderlineNav + actions horizontal layout and deal with the bottom border on the parent container. I also did so to support the use case in the code clone tab where we don't use a border at all.

But now that we have a clearer idea of what needs to be supported (and not) and how we'll move forward in the mid-long term, we should bring the underline back.

danielguillan:
I also think the name change from `UnderlineNav.Link` to `UnderlineNav.Item` makes sense. Please feel free to point anyone interested in this discussion to the design API proposal: https://github.com/github/primer/pull/1239

src/UnderlineNav2/UnderlineNavItem.tsx
broccolinisoup: with the new commit this is not relevant anymore

src/UnderlineNav2/UnderlineNavItem.tsx
broccolinisoup: with the new commit this is not relevant anymore

src/UnderlineNav2/UnderlineNavItem.tsx
broccolinisoup: with the new commit this is not relevant anymore

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,73 @@ export const UnderlineNavLink = forwardRef(
...
+      borderColor: selectedLink === ref ? 'primer.border.active' : 'transparent',
       ...(variant === 'small' ? smallVariantLinkStyles : defaultVariantLinkStyles),
-      '&:hover, &:focus': {
-        borderColor: selectedLink === ref ? 'primer.border.active' : 'neutral.muted',
-        transition: '0.2s ease'
+      '&:hover > div[data-component="wrapper"] ': {
+        backgroundColor: 'neutral.muted',
+        transition: 'background .12s ease-out'
+      },
+      '&:focus': {
```
broccolinisoup, line 109: new focus implementation. As @langermank pointed out today that Safari doesn't respect borderRadius for the outlines so new focus implementation, inspired from [Primer CSS](https://github.com/primer/css/blob/main/src/support/mixins/misc.scss)

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,73 @@ export const UnderlineNavLink = forwardRef(
...
+        // where focus-visible is supported, remove the focus box-shadow
+        '&:not(:focus-visible) > div[data-component="wrapper"]': {
+          boxShadow: 'none'
+        }
+      },
+      '&:focus-visible > div[data-component="wrapper"]': {
+        boxShadow: `inset 0 0 0 2px #0969da`
+      },
+      // renders a visibly hidden "copy" of the label in bold, reserving box space for when label becomes bold on selected
+      '& span[data-content]::before': {
```
broccolinisoup, line 123: From [Primer CSS](https://github.com/primer/css/blob/main/src/navigation/underline-nav.scss#L59) 🙌🏼

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,73 @@ export const UnderlineNavLink = forwardRef(
...
+        content: 'attr(data-content)',
+        display: 'block',
+        height: 0,
+        fontWeight: '600',
+        visibility: 'hidden'
+      },
+      // selected state styles
+      '&::after': {
```
broccolinisoup, line 131: Previous element for the border bottom was creating a jumpy affect when selected as it was conditionally rendered. So to avoid that jumpiness, used pseudo class (ref: [Primer CSS](https://github.com/primer/css/blob/main/src/navigation/underline-nav.scss#L90))

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -70,30 +77,73 @@ export const UnderlineNavLink = forwardRef(
...
-        borderColor: selectedLink === ref ? 'primer.border.active' : 'neutral.muted',
-        transition: '0.2s ease'
+      '&:hover > div[data-component="wrapper"] ': {
+        backgroundColor: 'neutral.muted',
+        transition: 'background .12s ease-out'
+      },
+      '&:focus': {
+        outline: 0,
+        '& > div[data-component="wrapper"]': {
+          boxShadow: `inset 0 0 0 2px #0969da`
```
broccolinisoup, line 112: TODO: use primitives here

src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -143,14 +156,14 @@ export const UnderlineNav = forwardRef(
...
-        <Box as={as} sx={merge(styles, sxProp)} aria-label={label} ref={newRef}>
+        <Box tabIndex={0} as={as} sx={merge(styles, sxProp)} aria-label={label} ref={newRef}>
           <Box as="ul" sx={merge<BetterSystemStyleObject>(overflowStyles, ulStyles)}>
             {responsiveProps.items}
           </Box>
 
           {actions.length > 0 && (
             <ActionMenu>
-              <ActionMenu.Button>More</ActionMenu.Button>
+              <ActionMenu.Button sx={{m: 0}}>More</ActionMenu.Button>
```
broccolinisoup, line 166: to fix extra margin around the button on safari

pksjce: wow, can you add a comment here about this?

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -36,16 +38,21 @@ export type UnderlineNavLinkProps = {
    */
   leadingIcon?: React.FunctionComponent<IconProps>
   as?: React.ElementType
+  /**
+   * Counter
+   */
+  counter?: number
```
pksjce, line 44: @danielguillan - Wondering if we are going to support loading state for counters.

broccolinisoup: Just making a comment here for the types. I added the counter prop with only number type just to make the Counter component redundant. We can add another type for the loading state if we are going to support 😊

pksjce (approved):

.changeset/thirty-mayflies-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': major
```
pksjce, line 2: I don't think this should be a major change yet. We'll kind of block ourselves doing that. Once we make all the changes, we can graduate from drafts to main folder with a major change. Could you do that before merging?

src/UnderlineNav2/UnderlineNavItem.tsx
danielguillan: Yes, we are going to support loading states. We can add support incrementally in another PR if that makes things easier.

Issue #2245, opened 2022-08-19
Regression: @primer/react is depends on esm only module @github/combobox-nav


okuryu:
**Describe the bug**
This is a similar problem to #2238. When I try to use the latest @primer/react in the Next.js app, I get the following error:

```
> next build

info  - SWC minify release candidate enabled. https://nextjs.link/swcmin
info  - Linting and checking validity of types
info  - Creating an optimized production build
info  - Compiled successfully
info  - Collecting page data ...Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/rokumura/work/tests/test-20220819/my-app/node_modules/@github/combobox-nav/dist/index.js from /Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/drafts/hooks/useCombobox.js not supported.
Instead change the require of index.js in /Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/drafts/hooks/useCombobox.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/drafts/hooks/useCombobox.js:8:43)
    at Object.<anonymous> (/Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/drafts/InlineAutocomplete/_AutocompleteSuggestions.js:16:20)
    at Object.<anonymous> (/Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/drafts/InlineAutocomplete/InlineAutocomplete.js:22:55)
    at Object.<anonymous> (/Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/drafts/InlineAutocomplete/index.js:8:50)
    at Object.<anonymous> (/Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/FormControl/FormControl.js:28:50)
    at Object.<anonymous> (/Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/FormControl/index.js:13:43)
    at Object.<anonymous> (/Users/rokumura/work/tests/test-20220819/my-app/node_modules/@primer/react/lib/index.js:669:43)
    at Object.551 (/Users/rokumura/work/tests/test-20220819/my-app/.next/server/pages/_app.js:22:31)
[...]

PR #2280, opened 2022-08-27
Update `"exports"` to use the correct conditions

DJMcNab:
From the npm docs:

> `"node"` - matches for any Node.js environment. Can be a CommonJS or ES module file. In most cases explicitly calling out the Node.js platform is not necessary.

At the moment, `node` environments unconditionally use the `require` module type in `lib`. However, this module type is broken because of https://github.com/primer/react/issues/2245. This change allows using the ES modules version for node environments instead, when specified using `import`.

Note that this doesn't fix the issue causing https://github.com/primer/react/issues/2245 - importing `"@primer/react"` using require will still be broken. However, this does allow this module to work with nodejs (when using `import`), including next.js.

### Merge checklist

I'm not sure any of these are applicable, but I've left them pending further guidance.

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


package.json
```diff
@@ -6,26 +6,23 @@
   "module": "lib-esm/index.js",
   "exports": {
     ".": {
-      "node": "./lib/index.js",
       "require": "./lib/index.js",
-      "default": "./lib-esm/index.js"
+      "import": "./lib-esm/index.js",
```
okuryu, line 10: JSON doesn't allow a trailing comma.

pksjce:
I like this change. It makes the exports more specific to the environment. Happy to merge when CI passes.

pksjce (approved):

package.json
DJMcNab: Ah, the danger of trusting myself with GitHub's web editor.

Thanks, I'll get that sorted.

DJMcNab:
I think this does need to result in a release of this package.

I don't believe this package's API includes it being imported matching neither of given conditions, so I don't believe this is to be a breaking change. Although I don't think there's any harm in marking this as a breaking change to be safe. I'll be happy to update either way, but I'm also happy for someone else to add the changeset once a consensus is reached.

pksjce:
@DJMcNab - Do you think removing the "default" option bite some consumer in some way or cause a regression? All implementations will fall in either "require" or "import" right?

DJMcNab:
No, I don't believe it will cause a regression.

Any JavaScript runtime will match one of the two conditions, so it would have to be a non JavaScript consumer. But even if there were one, I doubt we support that use case. And I can't think of a single possible case where it would be needed.

pksjce:
I tried your PR on a nextjs app setup with SSR. I still get this error.
image

But its supposed to get fixed with this change right?

DJMcNab:
I was finally able to get a local development version of this working - I had been having issues with virtualisation, so couldn't actually build this locally until I got that sorted.

We appear to be running into https://github.com/vercel/next.js/issues/37419 now. I could resolve that locally by setting this module to `"type": "module"`. Alternatively, I think setting `experimental.esmExternals` in the next config to `"loose"` also fixes this (I'm not sure if this PR is required in that case?)

However, after this I ran into another issue, where webpack dislikes [the dynamism in styled-components](https://github.com/styled-components/styled-components/blob/9ab5204d18bb8329e6971e108d270a49b79c0430/packages/styled-components/src/constructors/styled.tsx#L14-L17). (this appears to be https://github.com/styled-components/styled-components/issues/3601 )

Overall, I'm a bit out of my depth here I'm afraid. Surprisingly, webpack seems to either have no escape hatch for dynamic fields on the default export, or has extremely poor discoverability for that escape hatch. It appears that setting [`usedExports`](https://webpack.js.org/configuration/optimization/#optimizationusedexports) to `false` might fix this, but at this point I'm not sure I have the will to dig into this much further.

One other thing I did find is that in a project without this PR, just importing the specific files directly works. e.g.
```typescript
import BaseStyles from "@primer/react/lib-esm/BaseStyles";
```
Note that it needs to be `lib-esm` for typescript to pick it up properly.
This is a good enough workaround for me, until either this package becomes ESM only, and/or next.js sorts itself out.
I'm not sure why this doesn't run into https://github.com/styled-components/styled-components/issues/3601

PR #2290, opened 2022-08-30
chore: narrow typescript range in consumer test

joshblack:
This PR reduces the range for typescript from `^4.4.4` to `~4.7.2` as it appears that versions >= 4.8.2 cause an issue with `@types/styled-components`. The goal of this PR is to get the consumer-test back to green and then PR the fix in afterwards to support TS versions >= 4.8.2

In order to support TS >= 4.8.2, I believe we'll need to update `@types/styled-components` to >= 5.1.26 along with some other types in the project

### DIFF
--- a/consumer-test/package.json
+++ b/consumer-test/package.json
@@ -4,6 +4,6 @@
   },
   "dependencies": {
     "@primer/react": "*",
-    "typescript": "^4.4.4"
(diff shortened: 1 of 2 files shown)


colebemis (approved):
🚀

broccolinisoup:
yaaay 🚀

PR #2301, opened 2022-09-02
Hidden component

pksjce:
The `Hidden` component will be used to hide or show content in a responsive manner. We can do this across three viewport breakpoints aka 'narrow', 'regular' and 'wide'.

It is inspired by [Braid design system](https://github.com/seek-oss/braid-design-system/tree/master/packages/braid-design-system/lib/components/Hidden)

This was built to satisfy use cases specific to `PageHeader` eg - 
Responsive title and context nav

Here, we want to show different text inside the primary `Button` based on viewport value.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


docs/content/Hidden.mdx
```diff
@@ -0,0 +1,56 @@
+---
+title: Hidden
+description: Use Hidden to responsively hide or show content in narrow, regular and wide viewports.
+source: https://github.com/primer/react/blob/main/src/Hidden/index.tsx
+status: Alpha
+componentId: hidden
+---
+
+The `Hidden` component is a `<span />` that will help you hide or show content/components in different viewports.
+**Attention:** Use Hidden only to control behaviour in a responsive manner. It does not take any `sx` props.
```
joshblack, line 10: Quick question, would there be a way to control block vs inline flow for this component? Or is the intention that it should always be inline?

colebemis: Does hidden have to render any DOM element? Could it be something like this:

```jsx
function Hidden({when, children}) {
  const isHidden = useResponsiveValue(...)
  return !isHidden ? children : null
}

File: docs/content/Hidden.mdx [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+</Hidden>
+```
+
+## Props
+
+<PropsTable>
+  <PropsTableRow
+    name="when"
+    type="string or Array"
+    description="Can be one or more values of 'narrow', 'wide', 'regular'"
```
colebemis, line 34: Wonder if a more specific definition would be helpful here:
```suggestion
    type={`
| 'narrow'
| 'regular'
| 'wide'
| Array<'narrow' | 'regular' | 'wide'>`}
  description="The viewport range(s) at which children should be hidden"
```

docs/content/Hidden.mdx
pksjce: yeah so the tossup was between do a `display:none` or not render the component at all. I like not rendering. reduces markup + no need to specify `display` value at all.

docs/content/Hidden.mdx
pksjce: Yeah, I could write the actual size values.

siddharthkp:
Hi 👋 

Should we start this component in drafts till we feel confident about our responsive story?

pksjce:
Closing this PR. 
Basically after discussion with @siddharthkp and reviewing Chakra library, I think it's better we use [styled system's responsive props](https://styled-system.com/responsive-styles) to power this hide behaviour. I will update the API docs accordingly.

pksjce:
After some more thought and trying out the alternative `sx` way of having responsive css, I'm somewhat gravitating back to the simplicity and utility of this component.
I have addressed existing feedback and moved this component to the drafts folder. I think we should test out the utility before writing off this component.

src/Hidden/Hidden.tsx
```diff
@@ -0,0 +1,25 @@
...
+import {viewportRanges} from '../hooks/useResponsiveValue'
+
+type Viewports = 'narrow' | 'regular' | 'wide'
+
+type HiddenProps = {
+  on: Array<Viewports> | Viewports
+  children: React.ReactNode
+}
+
+export const Hidden = ({on: hiddenViewports, children}: HiddenProps) => {
```
joshblack, line 12: I loved the `when` convention from the docs, it was nice to read through it and think: "Okay, this is hidden when the viewport is narrow and wide". Is there a mental scheme you'd recommend with `on`?

src/Hidden/Hidden.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const isNarrowViewport = useMedia(viewportRanges.narrow)
+  const isRegularViewport = useMedia(viewportRanges.regular)
+  const isWideViewport = useMedia(viewportRanges.wide)
```
joshblack, line 15: Do you think it'd be possible to generate one media query string from the given viewports? Would be cool to do something like:

```jsx
function mediaQuery(rangeOrRanges: Array<Viewports> | Viewports) {
  const ranges = Array.isArray(rangeOrRanges) ? rangeOrRanges : [rangeOrRanges]
  return ranges.map(range => viewportRanges[range]).join(', ')
}

export const Hidden = ({on: hiddenViewports, children}: HiddenProps) => {
  const matches = useMedia(mediaQuery(hiddenViewports))
  if (matches) {
    return null
  }
  return children
}
```

Basically any time one of the media queries from `on` is true, we know it should hidden, otherwise return the children.

src/Hidden/Hidden.tsx
joshblack, line 12: Would we need to worry about providing a default value for SSR in this exploration?

docs/content/drafts/Hidden.mdx
```diff
@@ -0,0 +1,56 @@
...
+componentId: hidden
+---
+
+The `Hidden` component is a utility component that will help you hide or show content/components in different viewports.
+**Attention:** Use Hidden only to control behaviour in a responsive manner. It does not take any `sx` props.
+
+## Example
+
+```jsx live
+<Hidden when="narrow">
```
joshblack, line 15: Based on the implementation below, will this be changing to `on`?

joshblack (commented):
Love this component! Left some comments and questions below, let me know if they make no sense 😅 haha

src/Hidden/Hidden.tsx
pksjce: I kept going back and forth with `when` and `on` in my head. Sorry for the mix-up. Vini was also leaning towards `when`. Will change to that. 
For me its just a shorter sentence to say ` hidden on narrow viewport` vs `hidden when viewport is narrow`. When I think of it, I feel like its a non-native english speaker thing. 🤣

src/Hidden/Hidden.tsx
pksjce: Yeah, its a good optimisation!

siddharthkp (approved):
Let's ship it and use it internally to get a better sense of how it feels :)

(especially because the approach I wanted to try with sx doesn't work yet)

PR #2323, opened 2022-09-12
fix(ButtonBase): memoize sx prop

joshblack:
Part of https://github.com/github/primer/issues/1191

This PR tackles some of the small wins identified by @mattcosta7 over in https://github.com/github/primer/issues/1191

Specifically, it focuses on providing stable identity for `StyledButton` and the two underlying `Box` components in `ButtonBase` by:

- Pull out `iconWrapStyles` as a constant
- Pull out `trailingIconStyles` as a constant
- Create a default value for the `sx` prop, this allows it to be passed into `sxStyles` without triggering a new object each render


src/Button/ButtonBase.tsx
```diff
@@ -6,26 +6,37 @@ import {useTheme} from '../ThemeProvider'
...
-      display: 'inline-block'
-    }
     const sxStyles = useMemo(() => {
-      return merge.all([getButtonStyles(theme), getSizeStyles(size, variant, false), getVariantStyles(variant, theme)])
-    }, [theme, size, variant])
+      return merge.all([
+        getButtonStyles(theme),
+        getSizeStyles(size, variant, false),
+        getVariantStyles(variant, theme),
+        sxProp as SxProp
```
mattcosta7, line 27: I wonder if splitting these into 2 separate useMemos would be slightly better

I expect that the first 3 calls are _mostly static_ while the sxProp, if user provided, is unlikely to be memoized.

I'm not sure, but merging the first 3 once, and the user provided sx with that once, might be a bit better?

mattcosta7:
```
const sxStyles = useMemo(() => merge.all([
        getButtonStyles(theme),
        getSizeStyles(size, variant, false),
        getVariantStyles(variant, theme)
]), [...])

const sxValue = useMemo(() => merge(sxStyles, sxProps), [sxStyles, sxProps])
```

it's _likely_ the case that when users provide sxProps, they'll often bail out of memoization here, but when they don't we should avoid re-merging these.

joshblack: Totally agreed, will split them up in this way 👍

colebemis (approved):

colebemis (approved):

PR #2324, opened 2022-09-12
feat(PageLayout): support a ref on PageLayout.Pane

joshblack:
Reference https://github.com/github/primer/issues/1284

This PR updates `PageLayout.Pane` to use `React.forwardRef` and places the `forwardRef` on the `Box` wrapping `children`

Let me know if I'm interpreting the right element from #1284, I assumed it was the `Box` surrounding the `children` of the component 👀 I also tried to type the `React.forwardRef` correctly but would love a second look at that to see if I got the type correct.


colebemis (approved):
Looks great. Thanks, @joshblack!

Can you update the prop documentation table for `PageLayout.Pane` to include this ref prop?

We have a utility component to make this easy:
```
<PropsTableRefRow
    elementType="div"
    refType="HTMLDivElement"
  />
```

Example: https://primer.style/react/Textarea#props

joshblack:
Thanks @colebemis! And of course, done ✨

PR #2325, opened 2022-09-12
Minor fixes for `SelectPanel`

iansan5653:
This is a few minor fixes for the `SelectPanel` component:

1. Change the header and footer sections to use semantic `header` and `footer` elements. These elements will now get the `HeaderAsNonLandmark` and `FooterAsNonLandmark` roles, meaning that they are semantically the header and footer of the dialog without being the header and footer of the entire page.
2. Add `aria-modal="true"` to the `SelectPanel` dialog. When open, the dialog is the only thing on the page that can be interacted with. Clicking outside causes the dialog to close. Thus, it is a `modal`.
3. Make the `SelectPanel` footer padding consistent with the header padding. This not only looks better (items align to a grid), it also saves a tiny bit of vertical space.

#### Before

Screenshot of the selectpanel before. The footer padding is much larger than the header padding.

#### After
Screenshot of the selectpanel after the change. The footer padding is consistent with the header padding.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/SelectPanel/SelectPanel.tsx
+++ b/src/SelectPanel/SelectPanel.tsx
@@ -205,7 +205,13 @@ export function SelectPanel({
       open={open}
       onOpen={onOpen}
       onClose={onCloseOverlay}
-      overlayProps={{...overlayProps, onKeyPress: overlayKeyPressHandler, role: 'dialog', 'aria-labelledby': titleId}}
+      overlayProps={{
+        ...overlayProps,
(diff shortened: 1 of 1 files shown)


iansan5653:
Closing this for now since it's minor changes and will conflict with #2338. We can revisit later.

PR #2331, opened 2022-09-14
TreeView: Implement arrow key navigation

colebemis:
## Summary

Adds support for arrow key navigation of a TreeView using `aria-activedescendant` (based on the [TreeView API spec](https://github.com/github/primer/pull/1273)). 

👉 **[Try it out](https://primer-72d2763430-13348165.drafts.github.io/storybook/?path=/story/composite-components-treeview--file-tree-without-directory-links)**

https://user-images.githubusercontent.com/4608155/190521523-1b95966e-c0b7-414b-afc6-9c22a6cce08b.mp4

## Impact

This PR will enable keyboard and screen reader users to navigate a TreeView.

## VoiceOver demo

https://user-images.githubusercontent.com/4608155/190522770-29e38d85-5515-40f7-a0c8-f46c2e8c56ee.mov

## Test cases

Here are the test cases I wrote for the TreeView keyboard interactions (based on the [ARIA spec](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/#keyboard-interaction-24)):

```
Keyboard interactions
    ArrowDown
      ✓ moves aria-activedescendant to the next visible treeitem (118 ms)
    ArrowUp
      ✓ moves aria-activedescendant to the previous visible treeitem (127 ms)
    ArrowLeft
      ✓ collapses an expanded item (47 ms)
      ✓ does nothing on a root-level collapsed item (27 ms)
      ✓ does nothing on a root-level end item (19 ms)
      ✓ moves aria-activedescendant to parent of end item (54 ms)
      ✓ moves aria-activedescendant to parent of collapsed item (77 ms)
    ArrowRight
      ✓ expands a collapsed item (42 ms)
      ✓ moves aria-activedescendant to first child of an expanded item (45 ms)
      ✓ does nothing on an end item (46 ms)
    Home
      ✓ moves aria-activedescendant to first visible item (100 ms)
    End
      ✓ moves aria-activedescendant to last visible item (141 ms)
    Enter
      ✓ calls onSelect function if provided (18 ms)
      ✓ toggles expanded state if no onSelect function is provided (72 ms)
      ✓ navigates to href if provided (72 ms)
```

## Reviewers

I'm open to feedback about any part of the implementation but specifically, I'd love confirmation that the keyboard and screen reader behavior work as expected.

## Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/TreeView/TreeView.tsx
```diff
@@ -22,63 +33,275 @@ export type TreeViewProps = {
...
+}
+
 // ----------------------------------------------------------------------------
 // TreeView.Item
 
 export type TreeViewItemProps = {
   children: React.ReactNode
-  onSelect?: (event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>) => void
+  defaultExpanded?: boolean
+  onSelect?: (event: React.MouseEvent<HTMLElement> | KeyboardEvent) => void
```
mperrotti, line 230: When we create stories and/or examples, let's not forget to include examples that show consumers that they have to move focus to the first focusable element in the main content area when they activate a link node.

(Nothing needs to change in this PR)

src/TreeView/TreeView.tsx
```diff
@@ -22,63 +33,275 @@ export type TreeViewProps = {
...
+      return 'end'
+  }
+}
+
+function getVisibleElement(element: HTMLElement, direction: 'next' | 'previous'): HTMLElement | undefined {
+  const root = element.closest('[role=tree]')
+
+  if (!root) return
+
+  const walker = document.createTreeWalker(root, NodeFilter.SHOW_ELEMENT, node => {
```
mperrotti, line 163: Woah, I had no idea there was a `createTreeWalker` method 😮

src/TreeView/TreeView.test.tsx
```diff
@@ -1,50 +1,706 @@
...
+      fireEvent.keyDown(document.activeElement || document.body, {key: 'End'})
+
+      const child3 = getByRole('treeitem', {name: 'Child 3'})
+
+      // aria-activedescendant should now be set to child 3
+      expect(root).toHaveAttribute('aria-activedescendant', child3.id)
+    })
+  })
+
+  describe('Enter', () => {
```
mperrotti, line 581: Should we add a test to make sure that keying "Enter" on an expandable `LinkItem` activates the link node?

mperrotti (commented):
This is very exciting! Everything is looking good, but we don't want <kbd>Space</kbd> to call `onSelect`. We're reserving that for when/if we have nodes that can be checked or unchecked.

For future reference, check Primer's documentation on [tree view accessibility](https://primer.style/design/components/treeview#accessibility), before the W3C's [tree view APG](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/). I worked with @jscholes and the a11y team to suss out what the APG gets "wrong" and define improved patterns.

---

I skimmed the implementation and mostly focused on testing the UI. I can look closer at the code once @ericwbailey gives his ✅ on the keyboard navigation.

mperrotti:
Is it expected that none of the "links" in [File Tree With Directory Links](https://primer-72d2763430-13348165.drafts.github.io/storybook/?path=/story/composite-components-treeview--file-tree-with-directory-links) do anything when they're clicked or you press "Enter" when focused on them?

src/TreeView/TreeView.tsx
colebemis: Good callout 👍

src/TreeView/TreeView.test.tsx
colebemis: Good idea. Added ✅

jdrush89:
@colebemis @mperrotti should loading items in the tree be focusable? Or should aria-activedescendant skip over them?

ericwbailey:
@jdrush89 We'll want only one focusable node for loading node(s), with an announcement that indicates how many loading nodes are present (if we know). More granular behavior to follow in [the second round of TreeView API work](https://github.com/github/primer/issues/1300).

src/TreeView/TreeView.tsx
```diff
@@ -22,63 +33,274 @@ export type TreeViewProps = {
...
+          }
+        }}
+      >
+        {children}
+      </UlBox>
+    </RootContext.Provider>
   )
 }
 
+// DOM utilities used for focus management
```
mperrotti, line 88: Non-blocking nitpicks: 

1. Would it make more sense to move these to a utilities file instead of keeping them in `TreeView.tsx`?
2. Should these functions have their own unit tests? Or do we think the TreeView tests are sufficient?

mperrotti (approved):
Approval assumes @ericwbailey also approves.

src/TreeView/TreeView.tsx
colebemis: > 1. Would it make more sense to move these to a utilities file instead of keeping them in TreeView.tsx?

Yeah, it might make sense to pull this out into a hook later. I'm going to keep it all in the same file for now so I can iterate more quickly

> 2. Should these functions have their own unit tests? Or do we think the TreeView tests are sufficient?

I'd like to avoid testing implementation details like this because we'd need to update lots of tests if we decide to change the underlying implementation (even if we don't change behavior). I think the "Keyboard interactions" unit tests give us solid coverage.

ericwbailey (approved):
I've just tested this in VoiceOver and NVDA and it's looking/sounding great! Amazing stuff 🎉

PR #2336, opened 2022-09-14
feat(internal): add MatchMedia component and extend useMedia hook

joshblack:
This PR is a proposal that extends the `useMedia` hook with a `MatchMedia` component to help with emulation of features for development and docs.

_Note: this is only if DevTools does not provide an appropriate alternative._

This may be helpful in documentation scenarios in order to demonstrate the capabilities of a component without a viewer of the docs having that feature turned on (like in [this scenario](https://github.com/primer/react/pull/2297#discussion_r970317953))

It also updates the `useMedia` hook to support this component and adds support for `addEventListener`, where supported.

---

Would love to hear what you all think! Just am opening this up as a suggestion/conversation starter to gauge interest and usefulness


joshblack:
FYI @colebemis @siddharthkp if you're interested!

colebemis:
Should this replace the existing `useMedia` hook or does it serve a different purpose? https://github.com/primer/react/blob/main/src/hooks/useMedia.ts

joshblack:
@colebemis ah sorry, I was looking for this under `utils` and completely missed the `hooks` folder 🤦 

You're dead on, they have the same purpose so I'll update the existing file for this PR and remove the one added for `matchMedia`!

joshblack:
@colebemis thanks again for the heads up 🙏 Just pushed the changes and updated the description

src/utils/environment.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Indicate whether current execution environment can access the DOM.
+ *
+ * @see https://github.com/facebook/fbjs/blob/4d1751311d3f67af2dcce2e40df8512a23c7b9c6/packages/fbjs/src/core/ExecutionEnvironment.js#L12
+ */
+export const canUseDOM = !!(typeof window !== 'undefined' && window.document && window.document.createElement)
```
colebemis, line 6: This will be handy 😍

src/hooks/useMedia.tsx
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * This should be used for development and documentation only in situations
+ * where devtools cannot emulate this feature
+ *
+ * @example
+ * <MatchMedia features={{ "(pointer: coarse)": true}}>
+ *   <Children />
+ * </MatchMedia>
+ */
+export function MatchMedia({children, features = defaultFeatures}: MatchMediaProps) {
```
colebemis, line 121: Nice! This will be so helpful for docs 👍

colebemis (approved):
Love it 💖

src/hooks/useMedia.tsx
broccolinisoup: OMG this is what I needed! 😍  THANK YOU SO MUCH! @joshblack

broccolinisoup (approved):
Looks awesome 🔥  Looking forward to trying out when it is merged 🚀

src/hooks/useMedia.tsx
```diff
@@ -0,0 +1,151 @@
...
+type MatchMediaProps = {
+  children: React.ReactNode
+  features?: MediaQueryFeatures
+}
+
+const defaultFeatures = {}
+
+/**
+ * Use `MatchMedia` to emulate media conditions by passing in feature
+ * queries to the `features` prop. If a component uses `useMatchMedia` with the
```
joshblack, line 109:
```suggestion
 * queries to the `features` prop. If a component uses `useMedia` with the
```

PR #2337, opened 2022-09-14
feat(ThemeProvider): add custom `script` prop to override `dangerouslySetInnerHTML`

joshblack:
Closes https://github.com/github/primer/issues/1049

This PR introduces an approach to avoid using `dangerouslySetInnerHTML` if it's supported in the SSR framework.

Specifically, it introduces a new prop `script` that allows consumers to provide a built-in `Script` component. This component can then use the built-in way for frameworks to render trusted content with `<script>` tags such that their content is not sanitized.

Note: by default any content of a `<script>` tag will be sanitized and will result in a hydration DOM mismatch error.

This approach is opt-in, so any existing usage will continue to use the `dangerouslySetInnerHTML` path. For teams looking to use `script` there should be an example in our docs (will add in if this is the approach we want to take!) that looks like the following:

```jsx
import Script from 'next/script'
import {ThemeProvider, ThemeScriptProps} from '@primer/react/ThemeProvider'

function ThemeScript(props: ThemeScriptProps) {
  return <Script {...props} strategy="beforeInteractive" />
}

export default function App({Component, pageProps}) {
  return (
    <ThemeProvider preventSSRMismatch script={ThemeScript}>
      <Component {...pageProps} />
    </ThemeProvider>
  )
}
```

**Alternatives**

Outside of providing a `script` prop for interop, we could also render an element that is not a `script` and place content within there. For example:

```
<SCThemeProvider theme={resolvedTheme}>
  {children}
  {props.preventSSRMismatch ? (
    <div id="__PRIMER_DATA__" hidden>
      {JSON.stringify({resolvedServerColorMode: resolvedColorMode})}
    </div>
  ) : null}
</SCThemeProvider>
```

This approach will not receive the same sanitization as a `<script>` would and would be accessible during hydration by the component.

I'm not quite sure what the drawbacks of this approach would be. The only discernible different to me is that this element would be rendered where the ThemeProvider is used, where-as a `Script` from next.js with `stategy="beforeInteractive"` would be placed in `<head>`.

Would love feedback on these different approaches! There is also a Next.js example in this PR that can be used to try out the different patterns 👀


colebemis (approved):
This seems like a reasonable approach 👍 

A few questions:
- Would using CSS variables for theming eliminate the need for this script entirely?
- Will the SSR approaches we're exploring in Dotcom be compatible with this approach? (This is probably a question for Web Systems)
- Should we add documentation for the `script` prop in the Theming section of the Primer React docs?

joshblack:
@colebemis love these questions 🔥 

> Should we add documentation for the script prop in the Theming section of the Primer React docs?

Definitely 👍 

> Will the SSR approaches we're exploring in Dotcom be compatible with this approach?

Would there be a good point-of-contact or channel I could reach out in to double-check? 

> Would using CSS variables for theming eliminate the need for this script entirely?

I'm not sure, I think if you need to know the theme during rendering then you'll need some kind of bridge to keep things in sync. If you can keep things CSS-only then I think you'd be 100% good to go without it, but if you ever need to change what you render depending on the theme (like illustration) I think you'll end up needing this.

The bridge can definitely take on different shapes though, probably the simplest are a class name (`.theme-<theme-name>`) or a data-attribute (`data-theme="theme-name"`). I bet you could even use a CSS Custom Property for the theme value too. Each of these would still need to be outside of the scope of the react tree in order to be read during hydration to determine the theme.

Hope I understood that question right, let me know if that makes sense or if I misunderstood anything!

siddharthkp:
Came here to say I really like the alternative 😅 because it solves multiple problems (no dangerouslySetInnerHTML and no API for framework specific scripts)

joshblack:
@siddharthkp sounds good! Is the other approach switching it from a `<script>` to something like a `<div>` and reading from it? Happy to make that change 👍

PR #2343, opened 2022-09-16
docs(PageLayout): update custom sticky header to <header>

joshblack:
This is a quick follow-up to: https://github.com/github/primer/issues/1113 which updates our custom sticky header example to use a `<header>` element.


src/PageLayout/PageLayout.stories.tsx
```diff
@@ -602,6 +602,7 @@ export const CustomStickyHeader: Story = args => (
   // a box to create a sticky top element that will be on the consumer side and outside of the PageLayout component
   <Box data-testid="story-window">
     <Box
+      as="header"
```
broccolinisoup, line 605: 🙌🏼

broccolinisoup (approved):

PR #2344, opened 2022-09-17
fix(SelectPanel): .includes repacle with .find to avoid wrong checks …

theguriev:
The problem is that the object reference may change due to external filters and therefore the .includes will not work correctly. The .find solves this problem.  

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/SelectPanel/SelectPanel.tsx
+++ b/src/SelectPanel/SelectPanel.tsx
@@ -123,7 +123,7 @@ export function SelectPanel({
       return {
         ...item,
         role: 'option',
-        selected: 'selected' in item && item.selected === undefined ? undefined : finalItemsSelected.includes(item),
+        selected: 'selected' in item && item.selected === undefined ? undefined : !!finalItemsSelected.find(el => el.id === item.id),
         onAction: (itemFromAction, event) => {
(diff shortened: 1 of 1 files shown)


lesliecdubs:
@mperrotti are you able to give this a review?

src/SelectPanel/SelectPanel.tsx
```diff
@@ -123,7 +123,7 @@ export function SelectPanel({
       return {
         ...item,
         role: 'option',
-        selected: 'selected' in item && item.selected === undefined ? undefined : finalItemsSelected.includes(item),
+        selected: 'selected' in item && item.selected === undefined ? undefined : !!finalItemsSelected.find(el => el.id === item.id),
```
mperrotti, line 126: Non-blocking nitpick: `Boolean(finalItemsSelected.find(el => el.id === item.id))` is more readable.

```suggestion
        selected: 'selected' in item && item.selected === undefined ? undefined : Boolean(finalItemsSelected.find(el => el.id === item.id)),
```

mperrotti (approved):

PR #2348, opened 2022-09-19
TreeView: Add `current` prop

colebemis:
## Summary

Adds a `current` prop to `TreeView.Item` and `TreeView.LinkItem` to indicate which item is current (e.g., matches the user's current URL).

CleanShot 2022-09-19 at 12 06 16

```jsx
<TreeView>
  <TreeView.LinkItem href="#1">Item 1</TreeView.LinkItem>
  <TreeView.LinkItem href="#2" current>Item 2</TreeView.LinkItem>
  <TreeView.LinkItem href="#3">Item 3</TreeView.LinkItem>
</TreeView>
```

## Behavior

👉 [Try it out](https://primer-2e77022cab-13348165.drafts.github.io/storybook/?path=/story/composite-components-treeview--file-tree-with-directory-links)

- The path to the `current` item in a tree will be expanded by default
- The `current` item will be expanded by default if it contains sub-items
- The initial value of `aria-activedescendant` will be the `current` item (if there is a `currrent` item).

https://user-images.githubusercontent.com/4608155/191096655-55beed28-3a0d-4267-8c5a-38747a16def4.mp4

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/TreeView/TreeView.tsx
```diff
@@ -304,6 +329,7 @@ const Item: React.FC<TreeViewItemProps> = ({defaultExpanded = false, onSelect, o
             }
           }}
           sx={{
```
mperrotti, line 331: The value of `sx` could get reallly big as we start to get more into the styling. Maybe we could consider moving the style code into it's own file or at least outside of the component function's scope.

(Non-blocking for this PR)

mperrotti (commented):
`current` is set on the Button parent node, but it looks like `aria-activedescendent` is being set to the ID of the `Button.tsx` child node.

Screen Shot 2022-09-19 at 3 35 43 PM

colebemis:
> `current` is set on the Button parent node, but it looks like `aria-activedescendent` is being set to the ID of the `Button.tsx` child node.
> 
> Screen Shot 2022-09-19 at 3 35 43 PM

Hmm, I'm not able to reproduce that: 

CleanShot 2022-09-19 at 12 51 50

mperrotti:
@colebemis - I can't reproduce that anymore either. I think I must have unintentionally clicked `Button.tsx`

src/TreeView/TreeView.tsx
```diff
@@ -36,19 +36,29 @@ export type TreeViewProps = {
...
-    if (!activeDescendant) {
-      const firstItem = document.querySelector('[role="treeitem"]')
-      if (firstItem) setActiveDescendant(firstItem.id)
+    if (rootRef.current && !activeDescendant) {
+      const currentItem = rootRef.current.querySelector('[role="treeitem"][aria-current="true"]')
+      const firstItem = rootRef.current.querySelector('[role="treeitem"]')
+
+      // If current item exists, set it as the active descendant
+      if (currentItem) {
+        setActiveDescendant(currentItem.id)
```
mperrotti, line 49: Is it weird that `aria-activedescendant` updates on the parent, but `aria-current` does not move unless you manage it with state?

It seemed strange to me at first, but then I thought about it and was ok with it.

(non-blocking)

mperrotti (approved):

src/TreeView/TreeView.tsx
colebemis: Yeah, I think of `aria-current` more as a "default value" for `aria-activedescandant` rather than state that should be kept in sync

src/TreeView/TreeView.tsx
```diff
@@ -36,19 +36,29 @@ export type TreeViewProps = {
...
+      }
+      // Otherwise, set the first item as the active descendant
```
colebemis, line 51:
```suggestion
      // If current item exists, use it as the initial value for active descendant
      if (currentItem) {
        setActiveDescendant(currentItem.id)
      }
      // Otherwise, initialize the active descendant to the first item in the tree
```

src/TreeView/TreeView.tsx
colebemis: I update the code comments to make that more clear 👍

src/TreeView/TreeView.test.tsx
```diff
@@ -59,7 +59,22 @@ describe('Markup', () => {
     expect(subtree).toBeNull()
   })
 
-  it('initializes aria-activedescendant to the first item by default', () => {
+  it('initializes aria-activedescendant to the current item by default', () => {
```
broccolinisoup, line 62: Oh this is cool!

src/TreeView/TreeView.tsx
```diff
@@ -319,6 +345,19 @@ const Item: React.FC<TreeViewItemProps> = ({defaultExpanded = false, onSelect, o
             },
             [`[role=tree][aria-activedescendant="${itemId}"]:focus-visible &`]: {
               boxShadow: (theme: Theme) => `0 0 0 2px ${theme.colors.accent.emphasis}`
+            },
+            '[role=treeitem][aria-current=true] > &': {
+              bg: 'actionListItem.default.selectedBg',
```
broccolinisoup, line 350: This primitive looks tidied to the ActionList, I haven't seen many primitives like this so far in Primer. Is it temporary or any other reasoning behind using a component scoped primitive name?

colebemis: TreeView builds on ActionList styles so we're reusing some ActionList tokens/primitives here

src/TreeView/TreeView.stories.tsx
```diff
@@ -45,37 +45,42 @@ export const FileTreeWithDirectoryLinks: Story = () => (
...
-            <TreeView.LinkItem href="#index-html">index.html</TreeView.LinkItem>
-            <TreeView.LinkItem href="#favicon-ico">favicon.ico</TreeView.LinkItem>
-          </TreeView.SubTree>
-        </TreeView.Item>
-        <TreeView.LinkItem href="#package-json">package.json</TreeView.LinkItem>
-      </TreeView>
-    </nav>
-  </Box>
-)
+
```
colebemis, line 48:
```suggestion
```

PR #2349, opened 2022-09-19
docs(PageLayout): add docs with NavList and landmarks

joshblack:
Closes https://github.com/github/primer/issues/1304

This PR updates the docs of `PageLayout.Pane` based on feedback from: https://github.com/github/primer/issues/1113#issuecomment-1249560852. It includes more information about how `PageLayout.Pane` does not provide any landmark roles by default. It also includes suggestions as to appropriate roles that could be used along with an example of usage with a component which has a landmark role built-in (`NavList`)

### DIFF
--- a/docs/content/PageLayout.mdx
+++ b/docs/content/PageLayout.mdx
@@ -202,6 +202,39 @@ Add `offsetHeader` prop to specify the height of the custom sticky header along 
 </Box>
 ```
 
+### With `NavList`
+
+`PageLayout.Pane` by itself does not specify or provide landmark roles. Using
+components that define their own landmark roles, such as `NavList`, will provide
+clear structure to the page that is not present by default in `PageLayout.Pane`.
+
+It's important to note that multiple landmark roles, such as multiple
+`<nav>` elements, should be uniquely labelled in order to clarify what the
+navigation container is used for.
+
+```jsx live
+<PageLayout>
+  <PageLayout.Header>
+    <Placeholder label="Header" height={64} />
+  </PageLayout.Header>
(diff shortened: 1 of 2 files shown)


docs/content/PageLayout.mdx
```diff
@@ -262,6 +295,20 @@ Each component may be labeled through either `aria-label` or `aria-labelledby` i
...
+consider the type of content being rendered inside of `PageLayout.Pane` and if
+it requires a landmark role. Common landmark roles include:
+
+- [`aside`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)
+- [`navigation`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)
+
+Some components, such as `NavList` may already include a landmark role. In these
+situation, these components use an appropriate landmark role and no further
+action is needed when using `PageLayout.Pane`.
+
```
broccolinisoup, line 311: This is great! 🙌🏼

docs/content/PageLayout.mdx
```diff
@@ -202,6 +202,39 @@ Add `offsetHeader` prop to specify the height of the custom sticky header along
 </Box>
 ```
 
+### With `NavList`
+
+`PageLayout.Pane` by itself does not specify or provide landmark roles. Using
+components that define their own landmark roles, such as `NavList`, will provide
```
colebemis, line 208: Can we add a link to the NavList docs here?

```suggestion
components that define their own landmark roles, such as [`NavList`](/NavList), will provide
```

colebemis (approved):
Love it! Very clear 💖

PR #2352, opened 2022-09-20
Add resizable support to the `<PageLayout>` component

japf:
Describe your changes here.

This PR adds resizable support to the `<PageLayout>` component.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


colebemis:
This is looking great, @japf! 💖 

I was poking around in the storybook and noticed that the pane width seems to "jump" when you start dragging. Any ideas why this might be happening?

https://user-images.githubusercontent.com/4608155/191313165-55ca089b-35ca-4a7f-865a-7708bf907584.mp4

 I see this PR is marked as "work in progress" so feel free to ignore if this is a known issue :)

japf:
👋 @colebemis I pushed a few fixes, feel free to have another look. Please let me know what you think is missing to mark the PR as ready for review. I guess I should update the docs?

japf:
@colebemis updated again, please have another look 😄

src/PageLayout/PageLayout.tsx
```diff
@@ -370,6 +433,8 @@ export type PageLayoutPaneProps = {
    */
   positionWhenNarrow?: 'inherit' | keyof typeof panePositions
   width?: keyof typeof paneWidths
+  canResizePane?: boolean
```
colebemis, line 436: What do you think about calling this prop `resizable`?

I think `<PageLayout.Pane resizable>` reads well.

colebemis:
@japf this is looking great! @joshblack is going to take the lead on code review. He'll be your point-of-contact for any questions you have :)

src/PageLayout/PageLayout.tsx
japf: Yes, good point. Done in https://github.com/primer/react/pull/2352/commits/56413f8de76ef2b50c20347869e2c5beb202daad

joshblack:
Looking great! 🥳 Thanks so much for taking this on 🙏 Just wanted to leave some notes/questions on behavior since I wasn't sure what was expected

1. SC seems to be generating class names per clamp value

    It seems like Styled Components might be generating class names for each value in the `clamp()` 🤔 Would it be possible to pull the value out into an inline style or CSS Custom Property to avoid generating the class names? Or is there potentially something we can do with SC?

    https://user-images.githubusercontent.com/3901764/192837578-f4f902a7-0043-4591-baee-b94b658abbd5.mov


2. What's the expectation for when the layout goes narrow?

    I wasn't sure what the intent was for when the layout goes narrow, should the resize bar effectively be hidden at that breakpoint? Just wanted to double-check

    image


3. Text selection when resizing on FF

    Super minor, but noticed that text selection was happening for FF when hitting the edges of the resize behavior 🤔 

    https://user-images.githubusercontent.com/3901764/192835796-9e77d4c2-a51f-43af-9857-b5442729fc10.mov

4. Safari on iPad

    I wasn't sure how to get this to work when testing on iPad, it seems like the resize bar is interactive but when dragging it didn't seem to work (I guess because of `onMouseDown`?) Would we need to use something like `onPointerDown` to try and capture both cases or is the intent to drop down to touch events for iPad support?

5. Making the resize focusable

    Was curious what direction we were thinking about for making this keyboard operable 👀 
 One idea I had was to give the resize bar `role="separator"` and `tabindex="0"` to make it focusable. We could then use the following attributes to communicate state + purpose as it changes:

    - `aria-label` to label the resize bar
    - `aria-valuenow` to set the number reflecting the current position
    - `aria-valuemin` and `aria-valuemax` to represent the min/max values for resize
    - `aria-orientation="vertical"` to reflect orientation

    We could use `onKeyDown` on that element then and listen to ArrowLeft/ArrowRight to adjust it by px values and Shift+ArrowLeft, Shift+ArrowRight for increments of 10


Sorry for the long list here! Just wanted to say thanks again for taking this on and let me know if you have any questions or if I'm missing anything!

davekenn:
Just a note that our new code view is getting feedback that our file tree is too narrow when searching for files (goto file).  This resizing will solve that problem for us!  Cant wait for this to get to production!

src/PageLayout/PageLayout.tsx
```diff
@@ -423,6 +488,8 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
       positionWhenNarrow = 'inherit',
       width = 'medium',
       padding = 'none',
+      resizable = false,
+      paneWidthStorageKey = 'paneWidth',
```
colebemis, line 492: Since this is a prop on the `Pane` component, including `pane` in the prop name feels redundant. How do you feel about calling this `widthStorageKey`?
```suggestion
      widthStorageKey = 'paneWidth',
```

japf: Yes, good point. Done in https://github.com/primer/react/pull/2352/commits/7cfccdbb231ccfb3c3c0d2e69e2c16b4dd525d4d

joshblack:
@japf just pushed up a small change for that styled components issue! Let me know what you think, also happy to make a PR for this kind of stuff in the future to merge into your branch if that's preferred. Just let me know!

joshblack:
@japf just pushed up some docs btw for the two props added!

joshblack (approved):

colebemis:
Looks like this change causes the gap between the pane and content to be twice as large when there's no divider displayed:

CleanShot 2022-10-19 at 10 41 02

colebemis:
👋  @japf FYI I'm going to work on getting this over the finish line today. Planning to polish some of the interactions and styling.

src/PageLayout/useHorizontalResize.tsx
```diff
@@ -0,0 +1,97 @@
...
+  storageKey?: string
+) {
+  const [isDragging, setIsDragging] = useState(false)
+  const [isResizing, setIsResizing] = useState(false)
+  const [paneWidth, setPaneWidth] = useState<undefined | number>()
+
+  useEffect(() => {
+    let hasValue = false
+    if (storageKey) {
+      const storedWidth = window.localStorage.getItem(storageKey)
```
japf, line 17: As @mattcosta7 pointed out in https://github.com/github/issues/issues/3652 we might want to use a wrapped around the `localStorage` call to avoid throwing an error for folks who disabled 3rd party tracking in their browser.

colebemis:
Update: I finished addressing @joshblack's comments. Here's a quick video demo: https://share.cleanshot.com/imPRk8

This should be ready for another round of review. cc @joshblack

src/PageLayout/useHorizontalResize.tsx
colebemis: Good call. Updated ✅

src/PageLayout/PageLayout.tsx
```diff
@@ -416,13 +518,17 @@ const paneWidths = {
   large: ['100%', null, '256px', '320px', '336px']
 }
 
+const defaultPaneWidth = 0.2 // 20% of viewport width
```
colebemis, line 521: Is 20% the right default width for the pane?

joshblack (commented):
Looks great! Noticed a couple odd behaviors when testing in Safari and Firefox, curious if you know what's going on @colebemis  🤔 

Safari

https://user-images.githubusercontent.com/3901764/198400493-2e27e917-95b4-4b97-957a-49d21006ae55.mov

Firefox

https://user-images.githubusercontent.com/3901764/198400809-3723130d-05ef-4717-86c5-8901c4d728ac.mov

colebemis:
@joshblack I think those odd behaviors are related to how we set the max width of the pane. The max width of the pane is relative to the entire viewport size (`vw`), which can lead to unexpected behavior when PageLayout is nested inside smaller containers (as seen in the docs). Do those odd behaviors occur in Firefox and Safari in the fullscreen storybook example?

colebemis:
@joshblack I'm open to new ideas about how to handle min/max pane width if you have any! Here's the spec from @vdepizzol: https://github.com/github/primer/issues/581

joshblack (approved):

PR #2353, opened 2022-09-20
move markdown-toolbar-element into unconditional render

keithamus:
This moves the import of `markdown-toolbar-element` out of a conditional, which seems to be causing SegFaults due to poor support of ESM modules in various tools like Jest. Moving the import out of the conditional allows these tools to pick up the import which resolves segfaults.

We may want to move all web component definitions to a central file; `github/github` could de-duplicate these imports as it already imports them via PVC or via direct imports.

Closes #2339, refs #2334

### Merge checklist

- [x] Added/updated tests (N/A)
- [x] Added/updated documentation (N/A)
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/drafts/MarkdownEditor/_FormattingTools.tsx
+++ b/src/drafts/MarkdownEditor/_FormattingTools.tsx
@@ -1,5 +1,4 @@
-import React, {forwardRef, useImperativeHandle, useRef} from 'react'
-import '@github/markdown-toolbar-element'
+import React, {forwardRef, useImperativeHandle, useRef, useEffect} from 'react'
 
 export type FormattingTools = {
   header: () => void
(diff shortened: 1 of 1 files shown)


src/drafts/MarkdownEditor/_FormattingTools.tsx
```diff
@@ -1,4 +1,5 @@
-import React, {forwardRef, useImperativeHandle, useRef, useEffect} from 'react'
+import React, {forwardRef, useImperativeHandle, useRef} from 'react'
+import '@github/markdown-toolbar-element'
```
joshblack, line 2: Just wanted to double-check, I think one intent behind the condition is to allow for SSR. With this change, I believe SSR would no longer work because `@github/markdown-toolbar-element` references globals like `HTMLElement` and tries to register itself using `if (!window.customElements.get('...'))` when imported.

keithamus: That's correct. I think as we look more towards SSR we can look to centralise these dependencies into a client-side only bundle.

PR #2356, opened 2022-09-20
Only run replaceChildren with htmlContainer change

nicolleromero:
Upstreams a fix applied in https://github.com/github/memex/pull/12104

Setting the dependency of `htmlContainer` to the MarkdownViewer `useEffect` prevents `replaceChildren` from running unnecessarily (prevents removing and then immediately adding the element and triggering a bug in the selector-observer library).

### Screenshots

See referenced PR for screenshots

### Merge checklist

- ~~[ ] Added/updated tests~~
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- ~~[ ] Tested in Edge~~

### DIFF
--- a/.changeset/weak-hotels-shout.md
+++ b/.changeset/weak-hotels-shout.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+MarkdownViewer bug fix: Only run `replaceChildren` when `htmlContainer` changes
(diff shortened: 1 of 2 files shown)


iansan5653 (approved):
Looks good to me!

colebemis (approved):
Thanks, @nicolleromero!

PR #2357, opened 2022-09-20
TreeView: Implement typeahead

colebemis:
## Summary

Implements typeahead search of items in a TreeView:

👉 [Try it out](https://primer-51ba89d964-13348165.drafts.github.io/storybook/?path=/story/composite-components-treeview--file-tree-with-directory-links)

https://user-images.githubusercontent.com/4608155/191384491-744194d8-423c-4767-95cf-2a6a6f6c4abd.mp4

Per @jsholes's [suggestion](https://github.com/github/primer/issues/1234#:~:text=for%20each%20treeview.-,The%20%22Keyboard%20interactions%22%20section%20indicates%20that%20%22a%2Dz%2C%20A%2DZ%2C%20all%20printable%20chars%22%20will%20be%20supported%20for%20typeahead%2C%20but%20not%20that%20users%20can%20type%20multiple%20typeahead%20characters%20within%20a%20given%20timeout.%20Nor%20is%20the%20timeout%20mentioned.,-Note%3A%20the%20recommended), users can type multiple typeahead characters within a given timeout (300ms).

## Test cases

```
Typeahead
    ✓ moves aria-activedescendant to the next item that matches the typed character (288 ms)
    ✓ does nothing if no items match the typed character (198 ms)
    ✓ supports multiple typed characters (268 ms)
    ✓ prioritizes items following the current aria-activedescendant (188 ms)
    ✓ wraps around to the beginning if no items match after the current aria-activedescendant (236 ms)
```

## What should reviewers focus on?

* Check that the typeahead behavior in the [storybook stories](https://primer-51ba89d964-13348165.drafts.github.io/storybook/?path=/story/composite-components-treeview--file-tree-with-directory-links) works as expected
* Check that the typeahead behaviors works well with screen readers
* Review the [typeahead implementation](https://github.com/primer/react/pull/2357/files#diff-9de2d8eda6e5085e8b2d961fd35014e641594a6d7f7a2a79ca80710568150b55)
* Provide any feedback or missing functionality that you've noticed

## Merge checklist

- [x] Added/updated tests
- [ ] ~~Added/updated documentation~~
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/TreeView/useTypeahead.ts
```diff
@@ -0,0 +1,97 @@
+import React from 'react'
+
+type TypeaheadOptions = {
+  containerRef: React.RefObject<HTMLElement>
+  onFocusChange: (element: Element) => void
+}
+
+export function useTypeahead({containerRef, onFocusChange}: TypeaheadOptions) {
```
mperrotti, line 8: 😍

src/TreeView/useTypeahead.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ * Wraps an array around itself at a given start index
+ *
+ * @example
+ * wrapArray(['a', 'b', 'c', 'd'], 2) // ['c', 'd', 'a', 'b']
+ */
+function wrapArray<T>(array: T[], startIndex: number) {
+  return array.map((_, index) => array[(startIndex + index) % array.length])
```
mperrotti, line 96: Very cool. I love those rare moments where you get to use a modulo.

mperrotti (commented):
This is looking great!

Just two questions before I approve:
1. Will `useTypeahead` still work if child nodes are loaded asynchronously when their parent directory is opened?
2. Any idea how we could handle typeahead functionality if tree nodes are virtualized? It would be nice if I could still jump to a matching node even if it hasn't been loaded into the DOM yet.

colebemis:
> Will useTypeahead still work if child nodes are loaded asynchronously when their parent directory is opened?

Yeah, this implementation should work correctly with asynchronously loaded items. I will verify when I implement the async functionality 👍 

> Any idea how we could handle typeahead functionality if tree nodes are virtualized? It would be nice if I could still jump to a matching node even if it hasn't been loaded into the DOM yet.

Hmm. I haven't considered virtualization yet, but I imagine it will add a fair amount of complexity to the overall implementation. Is virtualization support in-scope for TreeView v1?

mperrotti:
@colebemis - we haven't discussed whether virtualization is in scope for v1.

I think we can move forward without virtualization support for v1. When/if it comes up we'll collaborate with product engineers to support virtualization.

Sound good?

mperrotti (approved):

src/TreeView/useTypeahead.ts
```diff
@@ -0,0 +1,111 @@
...
+
+      // Ignore key presses that occur with a modifier
+      if (event.ctrlKey || event.altKey || event.metaKey) return
+
+      // Update the existing search value with the new key press
+      setSearchValue(value => value + event.key)
+
+      // Reset the timeout
+      window.clearTimeout(timeoutRef.current)
+      timeoutRef.current = window.setTimeout(() => setSearchValue(''), 300)
```
jdrush89, line 35: should this use `useSafeTimeout`?

colebemis: Ah, I didn't know about that hook. Great suggestion! Just updated it ✅

ericwbailey (approved):
This is so cool.

jdrush89 (approved):

Issue #2351, opened 2022-09-20
Button props spilling to DOM element


cbix:
### Description

The Button props `leadingIcon`, `trailingIcon`, `variant` and `size` are being passed down to the `button` or custom (polymorphic) component, leading to React warnings like
```
Warning: React does not recognize the `leadingIcon` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `leadingicon` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

This happens because these props are not destructured like the others, however they serve no purpose outside the `ButtonBase` logic:

https://github.com/primer/react/blob/9eaa37e98b8cd12a0b953f190bcbff0f26a3dd22/src/Button/ButtonBase.tsx#L11
https://github.com/primer/react/blob/9eaa37e98b8cd12a0b953f190bcbff0f26a3dd22/src/Button/ButtonBase.tsx#L20

### Steps to reproduce

With this code using react-router
```tsx
<Button variant="outline" leadingIcon={MarkGithubIcon} as={Link} to="/github">Test</Button>
```
and development mode, check the console logs, they should give the warning mentioned above.

### Version

v35.6.0

### Browser

Chrome

rezrah:
Could this be happening because those props should be transient? https://styled-components.com/docs/api#transient-props

We'd need to upgrade `styled-components` to `5.1` from `4.4` (😬) to verify this, but this issue has been happening for a while in this library.

PR #2359, opened 2022-09-21
feat(Button): support as prop, add CustomButtonProps type

joshblack:
Closes https://github.com/primer/react/issues/2351

This PR grabs the extra values from props in `ButtonBase` and spreads them onto `StyledButton`.

One thing I noticed is that the props for this custom component would not align with `ButtonProps` and so I created a `CustomButtonProps` type to help out with this. I'd love feedback on if this would be the best way to support this, I wasn't quite sure what approach to take.

Something else I noticed is that the custom component through `as` must accept `sx` so I made sure the type intersects with `SxProp`.


siddharthkp (approved):
No brainer, ship it!

PR #2361, opened 2022-09-23
[UnderlineNav] : Keep selected item always visible

broccolinisoup:
Closes [#1296](https://github.com/github/primer/issues/1296)

Storybook link: https://primer-1475f2ddc1-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav

This PR is for keeping UnderlineNav's selected link always visible. This means;

- [x] If an item selected and the window is resized or a when a page is loaded and the selected item stays out of the viewport, component always make sure to keep this item on the list and not be pulled in to the dropdown menu. It does it by interrupting the natural order of the list and appending this selected item to the list as the last item.
- [x] If an item from the dropdown menu is selected, component moves this item out of the menu, appends it to the list and pulls as many item as required into the menu to make room for the selected item. (Natural order is interrupted here as well)

### Screen Recordings

https://user-images.githubusercontent.com/1446503/193276303-b846fc80-d6bf-400c-bb07-385ccde4b16c.mov

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/UnderlineNav2/UnderlineNav.test.tsx
+++ b/src/UnderlineNav2/UnderlineNav.test.tsx
@@ -1,4 +1,4 @@
-import React, {useLayoutEffect} from 'react'
+import React from 'react'
 import '@testing-library/jest-dom/extend-expect'
 import {fireEvent, render} from '@testing-library/react'
 import {
--- a/src/UnderlineNav2/UnderlineNav.test.tsx
+++ b/src/UnderlineNav2/UnderlineNav.test.tsx
@@ -1,7 +1,16 @@
-import React from 'react'
+import React, {useLayoutEffect} from 'react'
 import '@testing-library/jest-dom/extend-expect'
 import {fireEvent, render} from '@testing-library/react'
-import {CodeIcon, EyeIcon} from '@primer/octicons-react'
+import {
+  IconProps,
+  CodeIcon,
+  IssueOpenedIcon,
+  GitPullRequestIcon,
+  CommentDiscussionIcon,
+  ProjectIcon,
+  ShieldLockIcon,
+  GraphIcon
+} from '@primer/octicons-react'
 
 import {UnderlineNav} from '.'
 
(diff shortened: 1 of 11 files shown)


src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -138,16 +143,75 @@ export const UnderlineNav = forwardRef(
...
+      const itemsLeftInList = [...responsiveProps.items].slice(0, indexToSliceAt)
+      const updatedItemList = [...itemsLeftInList, prospectiveListItem]
+      // Form the new menu items
+      const itemsToAddToMenu = [...responsiveProps.items].slice(indexToSliceAt)
+      const updatedMenuItems = [...actions]
+      // Add itemsToAddToMenu array's items to the menu at the index of the prospectiveListItem and remove 1 count of items (prospectiveListItem)
+      updatedMenuItems.splice(indexOfProspectiveListItem, 1, ...itemsToAddToMenu)
+
+      // This is to prevent calling the overflowEffect function again when the responsiveProps is being updated.
+      setIsSwapping(true)
```
broccolinisoup, line 184: ## Help needed

The callback of the `swapMenuItemWithListItem` function is updating responsiveProps in the state. Because its state changes, the UnderlineNav gets re-rendered. And when it is re-rendered, it runs the `resizeObserverCallback` which runs `overflowEffect` function that is primarily how we display items in the list and in the menu with the given children as UnderlineNav.Item. 

Setting `isSwapping` state `true` here prevents the `overflowEffect` function running when the responsiveProps is being updated and we can see the items move in the screen recording which is what we need however, it also prevents running the `overflowEffect` function when it is actually needed such as re-size or selecting a visible item because there is no where `isSwapping` gets assigned to false. 

Maybe I am so absorbed in the same code for a while but I cannot seem to figure this out. I need to prevent `overflowEffect` function from running when we are swapping items and let it run in all other cases.

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -64,28 +63,54 @@ export const UnderlineNavItem = forwardRef(
...
+          Number(getComputedStyle(icon).marginLeft.slice(0, -2))
+        : 0
+
+      setChildrenWidth({text, width: domRect.width})
+      setNoIconChildrenWidth({text, width: domRect.width - iconWidthWithMargin})
       preSelected && selectedLink === undefined && setSelectedLink(ref as RefObject<HTMLElement>)
-    }, [ref, preSelected, selectedLink, setSelectedLink, setChildrenWidth, setNoIconChildrenWidth])
+
+      if (selectedLinkText === text) {
+        setSelectedLink(ref as RefObject<HTMLElement>)
```
broccolinisoup, line 101: https://github.com/primer/react/pull/2361/files#diff-23b9142fc4791adf1608d6fa33d6452c29f358137c29ddaed0341beb8c438a32R211

src/UnderlineNav2/UnderlineNav.tsx
broccolinisoup: Help found 💖

broccolinisoup:
Hi @danielguillan 👋🏼 Keeping the selected item visible is ready for you to review 😊 This PR includes both overflow behaviour as well as the loading counter states. Feel free to review and test them as well here. 
Please see the storybook link [here](https://primer-1475f2ddc1-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav) and on the description as well and I'll update the docs next week 📝  Looking forward to hearing your thought!

src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -106,14 +105,19 @@ const overflowEffect = (
       for (const [index, child] of childArray.entries()) {
         if (index < numberOfItemsPossibleWithMoreMenu) {
           items.push(child)
+          // keeping selected item always visible.
+        } else if (child.props.selected) {
```
siddharthkp, line 109: yay ssr friendly hack 👍

danielguillan:
This looks great!

My only feedback is that we should persist the item order when selecting visible items. That is, when we swap a selected item to make it visible and the user then selects an already visible item, we should keep the current (updated) item order instead of restoring the original one. This is to prevent the visual change that happens away from the user's cursor which can be confusing.

https://user-images.githubusercontent.com/175638/193542634-46a85ecf-24e6-4a14-b138-c0b26a2b8500.mp4

In the recording above, when we click "Actions", the "Insights" item shouldn't be swapped back with "Projects". "Insights" should still be the last item. Same for when we click "Pull requests", the "Settings" item shouldn't be swapped back with "Projects".

broccolinisoup:
> This looks great!
> 
> My only feedback is that we should persist the item order when selecting visible items. That is, when we swap a selected item to make it visible and the user then selects an already visible item, we should keep the current (updated) item order instead of restoring the original one. This is to prevent the visual change that happens away from the user's cursor which can be confusing.
> 
>  Always.visible.item.swap.mp4 
> In the recording above, when we click "Actions", the "Insights" item shouldn't be swapped back with "Projects". "Insights" should still be the last item. Same for when we click "Pull requests", the "Settings" item shouldn't be swapped back with "Projects".

👋🏼 @danielguillan Thanks so much for your review! No worries, I wasn't sure about that ordering behaviour and it makes sense to reduce the layout shifts. I'll revert it now

src/UnderlineNav2/UnderlineNav.test.tsx
```diff
@@ -79,34 +113,89 @@ describe('UnderlineNav', () => {
...
+  it('respects loadingCounters prop', () => {
+    const {getByText} = render(<ResponsiveUnderlineNav loadingCounters={true} />)
+    const item = getByText('Actions').closest('a')
     const loadingCounter = item?.getElementsByTagName('span')[2]
     expect(loadingCounter?.className).toContain('LoadingCounter')
     expect(loadingCounter?.textContent).toBe('')
   })
 })
+
+describe.skip('UnderlineNav when overflow occurs on fine pointer devices', () => {
```
broccolinisoup, line 132: 🆘 Help Needed 🆘 

I cannot seem to make these tests work. Because jsdom doesn't have the ability to calculate where elements will be visually laid out as a result of CSS [ref](https://github.com/jsdom/jsdom#unimplemented-parts-of-the-web-platform) the list items width return 0 so that I cannot really test the responsive behaviour. I tried to add `Object.defineProporty` above but that doesn't seem to do the trick. I would appreciate any help on to test these responsive behaviour i.e `More` menu button visibility.

broccolinisoup: Just a context here that we are discarding writing unit test for this complex visual cases. I'll be writing interaction tests for it on a different PR 🥳

pksjce (approved):
Lets go!

PR #2362, opened 2022-09-23
Add isolate: isolation to AvatarStack

pablonete:
Prevent AvatarStack to overlap Dialog or other portal components.
AvatarStack uses `z-index` that leak to other components, specially Dialog or other portals. Adding `isolation` it won't leak.

I've updated the storybook to make visible this scenario, but if you think it's too specific, I can either split into a separate story, or just drop the storybook changes. I'm open to suggestions, it's my first time contributing here 🤓 

#### Before

image

#### After

image

### Merge checklist

- [x] Added/updated tests: updated storybook to showing this scenario
- [ ] Added/updated documentation: no need
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/AvatarStack.tsx
+++ b/src/AvatarStack.tsx
@@ -12,6 +12,7 @@ type StyledAvatarStackWrapperProps = {
 const AvatarStackWrapper = styled.span<StyledAvatarStackWrapperProps>`
   display: flex;
   position: relative;
+  isolation: isolate;
   height: 20px;
   min-width: ${props => (props.count === 1 ? '20px' : props.count === 2 ? '30px' : '38px')};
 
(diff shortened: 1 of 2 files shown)


iansan5653 (approved):
Looks great! Thanks for upstreaming this fix.

lesliecdubs:
@pablonete sorry this is still awaiting a review!

@siddharthkp it looks like you were assigned; when you're back from vacation, can you please take a look?

siddharthkp:
> @siddharthkp it looks like you were assigned; when you're back from vacation, can you please take a look?

Huh! Strange, i remember reviewing a similar change and wasn't sure about the side effects, so we wanted to test it with canary version before merging this.

PR #2366, opened 2022-09-23
fix(package): update exports to import/require

joshblack:
Update the [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) used for the entrypoints of `@primer/react` from `node`, `require`, and `default` to `require` and `import`.

In the current setup, `node` will match to CommonJS which can cause issues with frameworks like Next.js which can handle ESM directly. In the proposed setup, the CommonJS path will be brought in when the package is imported through `require()`. The ESM path will be brought in when the package is imported using `import from` or `import()`.

This PR also addresses issues in the CommonJS bundle by inlining ESM-only dependencies. This is intended as an interim step to get this bundle working again and is not intended as a long-term solution.

**Testing**

Included in this PR is a Next.js example under `examples/nextjs`. This example can be used to see the difference in the `exports` field in order to demonstrate that `@primer/react` can participate in server-side rendering.

This will not include all exports as some still may reference DOM globals and for which a separate test will need to be introduced.

To see how the export changes work, you can:

- Visit `package.json`
- Remove the `"."` export value, run `npm run develop` in `examples/nextjs` to verify that Next.js tries to use `exports` to lookup the entrypoint for `@primer/react` (it should fail to find it)
- Restore the `"."` export value, remove the `"import"` key, run `npm run develop` in `examples/nextjs` to verify that Next.js tries to lookup the `import` value (it should fail to load path ".")
- Restore the `"import"` key, run `npm run develop`, and verify that Next.js can compile (you can also remove `"require"` to double-check that Next.js is not going down that path)


devephy:
I think all these pull requests should be settled first or closed

colebemis (approved):
This seems like a great incremental step. Ship it! 🚢

PR #2372, opened 2022-09-26
Add `sshd` to codespace container

koddsson:
I want to install `sshd` on the codespace container.

### DIFF
--- a/.changeset/fast-moles-protect.md
+++ b/.changeset/fast-moles-protect.md
@@ -1,5 +1,0 @@
----
-'@primer/react': patch
----
-
-Add `sshd` to codespace container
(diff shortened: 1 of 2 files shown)


.changeset/fast-moles-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Add `sshd` to codespace container
```
colebemis, line 5: This doesn't need a changeset since it's not a consumer facing change

colebemis (approved):

.changeset/fast-moles-protect.md
koddsson: 👍🏻

Issue #1979, opened 2022-03-17
Button content grid cells are equal


dusave:
**Describe the bug**
With version 35.x, Buttons have a new content grid, containing spots for leading and trailing icons along with the content of the button. The issue is that the columns are all `1fr` by default, giving much more space to the icon cells than they need whilst compressing the content cell.

**To Reproduce**
1. Create a button with leading and/or trailing icons, along with content
2. See error

<img width="333" alt="image" src="https://user-images.githubusercontent.com/2447456/158883921-38619ad8-dd35-4c03-902d-f0cf3f3d84b1.png">

**Expected behavior**
Grid cells should be `gridTemplateColumns: "min-content 1fr min-content"`
<img width="334" alt="image" src="https://user-images.githubusercontent.com/2447456/158884004-037c9707-9002-406e-bbe9-7d134f33282d.png">

lesliecdubs:
Hey @dusave, I'm so sorry we didn't get back to you on this much sooner! Apologies 🙏🏻 

Are you able to confirm this is still a problem?

tallys:
Update, @langermank is working on resolving this right now 🎉

lesliecdubs:
Noting that the linked PR includes breaking changes, so we'd schedule them to go out with the next major release.

langermank:
Addressed here https://github.com/primer/react/pull/2792

PR #2374, opened 2022-09-26
Button bug fixes and visual updates

langermank:
**Note**: These are **breaking changes** and should be considered for the next major release

## Improvements

- Add a `block` prop for full width
- Add `alignContent` prop to align content to [center or start](https://github.com/github/primer/issues/939)
- Use `control` sizing CSS variables
- Use `height` over padding for more control over sizing
- Refactor `Button.Counter` to better align with the other visual props
- Rename visual props to better align with other Primer components

## Visual updates

Addresses a few things missing from this discussion: https://github.com/github/primer/discussions/1016

- Deprecate `outline` variant
- Slightly more condensed padding
- Improved `invisible` hover states (will update these colors when new functional colors are addressed)

Closes:
- https://github.com/primer/react/issues/1979
- https://github.com/github/primer/issues/939

### Screenshots

Please refer to the comprehensive Storybook stories- this is just a small sample of some of the changes

| before | after | description |
| -- | -- | -- |
| image | image | trailing action slot alignment |
| image | image | default padding, font-weight |
| image | image | large height |
| icon was blue | image | invisible svg color |
| image | image | trailing action locks right in full width |

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Button/ButtonFeatures.stories.tsx
```diff
@@ -0,0 +1,40 @@
+import {EyeIcon, TriangleDownIcon, HeartIcon} from '@primer/octicons-react'
+import React, {useState} from 'react'
+import {Button} from '.'
+
+export default {
+  title: 'Components/Button/Features'
+}
+
+export const Default = () => <Button>Default</Button>
```
langermank, line 9: I'm not using `Template.bind()` for feature stories as an attempt to maintain the story markup for the "Source" addon panel. This makes it easier for someone to copy/paste the example.

I will write an ADR about this format as it impacts our docs work 😄

src/Button/IconButton.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+      }
+    },
+    icon: {
+      control: {
+        type: 'select',
+        options: Object.keys(icons)
+      },
+      mapping: icons
+    },
+    ariaLabel: {control: 'text'}
```
langermank, line 36: Can someone help me figure this prop out? I guess we don't expose it as a normal prop like `ariaLabel`, the consumer is mean to supply `aria-label="hello"`.

- Do we like that API?
- If yes, how can I hook into it to surface it as a Storybook control? Its not camel case..

error message about storybook prop

src/Button/ButtonFeatures.stories.tsx
rezrah: I like this, but I do worry that we're reverting back to `storiesOf` format instead of leveraging CSF correctly in order to make an optional addon functional. If there's overarching appetite for what the addon provides, to the extent that we're willing to exchange for longer, more verbose `*.stories.tsx` files, I guess that's fine. 

Personal take. I like this explicit approach, but I'm a little nervous about future maintainability and interoperability with Storybook direction.

rezrah: An example of the interoperability is Storybook Interactions. I think we'd lose this altogether? 

Example of how it works right now 👇 
https://user-images.githubusercontent.com/13340707/195042883-a409265d-d1f0-4604-8133-bf120beab048.mov

I'll do some testing in Brand to see what happens if we move away from template binding, as there may be a way to do this by attaching the interactions config to the individual Object prototype.

langermank: Its so frustrating that Storybook doesn't support this addon with the bind syntax _that they recommend_ 😅. I found several open issues on it when I went digging.

What's weird is that the story code shown in Docs does just fine with the template syntax (see below)

image

The main reason I'm trying to drive adoption of this `Story` source code addon is to support the new Primer docs initiative. We'll be relying on Storybook preview in an iFrame to render code samples, and won't have hand written code samples anymore. Since we're driving people to Storybook, I think its somewhat critical that we provide an easy way to copy code blocks. **If its doable, I would be 100% down to try and build our own addon to accomplish this!** But I would need eng resources for that one as its a bit out of my wheelhouse.

src/Button/IconButton.stories.tsx
broccolinisoup: From what I observe, for the aria attributes we seem to follow the native convention for our API. `aria-label` over `ariaLabel` 

To make that type error go, you can use `aria-label` instead of `ariaLabel`

For argTypes
```
'aria-label': {control: 'text'}
```

For args
```
'aria-label': 'Icon button description',
```

And I think you don't need to explicitly set up the `aria-label` prop in the component anymore. 

```
<IconButton {...args} />
```

Let me know what you think 👀

src/Button/ButtonBase.tsx
```diff
@@ -4,29 +4,59 @@ import Box from '../Box'
...
-        {TrailingIcon && (
-          <Box as="span" data-component="trailingIcon" sx={{...iconWrapStyles, ml: 2}}>
-            <TrailingIcon />
+      <StyledButton
+        as={Component}
+        sx={merge(sxStyles, sxProp as SxProp)}
+        {...props}
+        ref={forwardedRef}
+        data-component={block ? 'block' : null}
+        data-size={size === 'small' ? 'small' : size === 'large' ? 'large' : undefined}
```
broccolinisoup, line 38: We don't want to have `data-size` attribute for medium sizes?

Not a blocker, just a suggestion to maybe improve readability - feel free to ignore if it is not helpful.

If we don't mind having `data-size` attribute for medium sizes maybe we can do `data-size={size}`

If we only want the attribute for `small` and `large` sizes maybe we can do `data-size={size === 'small' || size === 'large' ? size : undefined}`
 Let me know what you think 👀

src/Button/IconButton.tsx
```diff
@@ -1,24 +1,22 @@
...
-  ])
+  const sxStyles = merge.all([getBaseStyles(theme), getVariantStyles(variant, theme), sxProp as SxProp])
   return (
-    <StyledButton sx={sxStyles} {...rest} ref={forwardedRef}>
-      <Box as="span" sx={{display: 'inline-block'}}>
-        <Icon />
-      </Box>
+    <StyledButton
+      data-component="IconButton"
+      data-size={size === 'small' ? 'small' : size === 'large' ? 'large' : undefined}
```
broccolinisoup, line 14: same suggestion I mentioned above.

src/__tests__/Button.test.tsx
```diff
@@ -29,10 +29,10 @@ describe('Button', () => {
     expect(onClick).toHaveBeenCalledTimes(1)
   })
 
-  it('respects width props', () => {
-    const container = render(<Button sx={{width: 200}}>Block</Button>)
+  it('respects width prop', () => {
```
broccolinisoup, line 32: Should it be "respects block prop"?

src/Button/types.ts
```diff
@@ -41,11 +47,19 @@ export type ButtonProps = {
   /**
    * The leading icon comes before button content
    */
-  leadingIcon?: React.FunctionComponent<React.PropsWithChildren<IconProps>>
+  alignContent?: AlignContent
```
broccolinisoup, line 50: Should we add a unit test for it?

src/Button/types.ts
```diff
@@ -41,11 +47,19 @@ export type ButtonProps = {
...
+  /**
+   * The leading icon comes before button content
+   */
+  leadingVisual?: React.FunctionComponent<React.PropsWithChildren<IconProps>>
   /**
    * The trailing icon comes after button content
    */
-  trailingIcon?: React.FunctionComponent<React.PropsWithChildren<IconProps>>
+  trailingVisual?: React.FunctionComponent<React.PropsWithChildren<IconProps>>
```
broccolinisoup, line 58: Same question - should we add a unit test for it?

src/Button/IconButton.stories.tsx
langermank: Thank you! This totally worked 🎉 I didn't know I could use a string for a prop name in Storybook args!

src/Button/Button.stories.tsx
```diff
@@ -28,223 +18,64 @@ export default {
...
+        type: 'radio',
+        options: ['center', 'start']
+      }
+    },
+    block: {
+      control: {
+        type: 'boolean'
+      }
+    },
+    leadingVisual: {
```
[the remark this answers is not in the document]
langermank: Brilliant, thanks! Swapped all the stories with that.

src/Button/ButtonBase.tsx
langermank: Since medium is default, we don't _really_ need an attribute for it for CSS. I like your second suggestion though!

langermank:
Closing in favor of https://github.com/primer/react/pull/2714

PR #2379, opened 2022-09-27
Don't crash when building docs

koddsson:
I moved the `@github/markdown-toolbar-element` import into the render function like suggested in https://www.gatsbyjs.com/docs/debugging-html-builds/

Closes https://github.com/github/markdown-toolbar-element/issues/68

### Screenshots

### Merge checklist

- [ ] ~Added/updated tests~
- [ ] ~Added/updated documentation~
- [ ] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

cc/ @keithamus

### DIFF
--- a/src/drafts/MarkdownEditor/_FormattingTools.tsx
+++ b/src/drafts/MarkdownEditor/_FormattingTools.tsx
@@ -1,4 +1,4 @@
-import React, {forwardRef, useImperativeHandle, useRef, useEffect} from 'react'
+import React, {forwardRef, useImperativeHandle, useRef} from 'react'
 
 export type FormattingTools = {
   header: () => void
@@ -48,12 +48,7 @@ export const FormattingTools = forwardRef<FormattingTools, {forInputId: string}>
     reference: () => referenceRef.current?.click()
   }))
 
-  useEffect(() => {
-    // eslint-disable-next-line prettier/prettier
-    (async function () {
-      await import('@github/markdown-toolbar-element')
-    })()
-  })
+  require('@github/markdown-toolbar-element')
 
   return (
     <markdown-toolbar for={forInputId} style={{display: 'none'}}>
--- a/src/drafts/MarkdownEditor/_FormattingTools.tsx
+++ b/src/drafts/MarkdownEditor/_FormattingTools.tsx
@@ -1,5 +1,4 @@
 import React, {forwardRef, useImperativeHandle, useRef, useEffect} from 'react'
-import '@github/markdown-toolbar-element'
 
 export type FormattingTools = {
   header: () => void
@@ -50,6 +49,8 @@ export const FormattingTools = forwardRef<FormattingTools, {forInputId: string}>
     mention: () => mentionRef.current?.click(),
     reference: () => referenceRef.current?.click()
   }))
+
+  require('@github/markdown-toolbar-element')
 
   return (
     <markdown-toolbar for={forInputId} style={{display: 'none'}}>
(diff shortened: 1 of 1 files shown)


keithamus (approved):
LGTM

joshblack:
Responded on the upstream issue, let me know what you think!

Separately, would this run into the same mix-and-match issue described in: https://github.com/primer/react/pull/2339 ? Specifically:

> This mix of `require` and `import` confuses transpilers and occasionally breaks things in unexpected and hard-to-debug ways ([for example, see this Slack thread](https://github.slack.com/archives/C01L618AEP9/p1663170031622459?thread_ts=1663028544.311919&cid=C01L618AEP9)).

koddsson:
I pushed a commit that utilizes `useEffect` and dynamic imports so that we can avoid `require` calls if we want.

joshblack:
@koddsson sorry about the dynamic import stuff 😞 Seems like the same issue that @iansan5653 ran into.

For the mix-and-match, this would end up being a bundler behavior, right? If I understand right, `require()` is undefined in an ESM scope.

Also wanted to check-in with the deploy preview, it seems like the build is failing with this error:

image

koddsson:
> Also wanted to check-in with the deploy preview, it seems like the build is failing with this error:

I don't see this anywhere. Do you have a link to this UI?

joshblack:
@koddsson let me know if this works! https://github.com/primer/react/actions/runs/3144532666/jobs/5110661581

I think it's showing up for the `Deploy (fork)` workflow

src/drafts/MarkdownEditor/_FormattingTools.tsx
```diff
@@ -49,6 +48,8 @@ export const FormattingTools = forwardRef<FormattingTools, {forInputId: string}>
     reference: () => referenceRef.current?.click()
   }))
 
+  require('@github/markdown-toolbar-element')
```
mattcosta7, line 51: I don't think doing this in component scope is ideal, sine this calls require on every render (sure it's often a noop), but also runs in server environments. 

```
useEffect(() => { require....}, [])
```

would at least ensure it doesn't run in non-browser environments, however I believe that was already tried and lead to the issues with `require'ing` an esmodule.

mattcosta7: evaluating this require in a context that runs on the server likely explains the CI error @joshblack  noted 

```
error Building static HTML failed

  34 | }
  35 | const styles = new WeakMap();
> 36 | class MarkdownButtonElement extends HTMLElement {
     | ^
  37 |     constructor() {
  38 |         super();
  39 |         const apply = () => {


  WebpackError:ReferenceError: HTMLElement is not defined
  ```

Comment (koddsson) [reply src/drafts/MarkdownEditor/_FormattingTools.tsx]: Yup! Me and @joshblack paired on this earlier and I just forgot to push the commit.

**Comment by koddsson** [comment]
I annoyingly can't reproduce the deploy error locally. Any ideas? Can you reproduce it?

**Comment by koddsson** [comment]
> I annoyingly can't reproduce the deploy error locally. Any ideas? Can you reproduce it?

Seems like this was a issue with me not having push access to this repo. Seems to be running green now 👍🏻

File: src/drafts/MarkdownEditor/_FormattingTools.tsx [hunk 2]
```diff
@@ -49,6 +48,10 @@ export const FormattingTools = forwardRef<FormattingTools, {forInputId: string}>
     reference: () => referenceRef.current?.click()
   }))
 
+  useEffect(() => {
+    require('@github/markdown-toolbar-element')
+  })
```
mattcosta7, line 53:
```suggestion
  }, [])
```

we don't want to evaluate this every render, just once should be enough

PR #2387, opened 2022-09-30
refactor(MarkdownEditor): guard web component imports

joshblack:
**Approach**

This PR proposes a convention for dependencies intending to run only on the client. For components looking to use these dependencies, they can use a bare import to a file under `src/web-components`. This file should include a check to see if the current runtime supports the DOM before using a dynamic `import()` to bring in the dependency.

```ts
// src/web-components/markdown-toolbar-element.ts
import { canUseDOM } from '../utils/environment';

if (canUseDOM) {
  import('@github/markdown-toolbar-element');
}

// src/path/to/component
import '../web-components/markdown-toolbar-element';
export default function ExampleComponent() {
  // ...
}
```

Files under `src/web-components/*` are treated as side effects to make sure that they are included in the bundle even if no imports are used.

**Caveats**

Dynamic import is not supported in environments like Jest until support lands for ESM in Node.js VMs[^1]. We can use `babel-plugin-dynamic-import-node` locally to polyfill this in Node.js environments. However, downstream teams may not be using this plugin and will run into an unexpected error if we ship with this change.

As a result, we have a couple of options:

- Create Node.js specific bundles which include this plugin
- Try and feature detect to determine if we should use `import()` or `require()`. This could potentially look like:

    ```ts
    import {canUseDOM} from '../utils/environment'
    
    if (canUseDOM) {
      if (typeof require === 'function') {
        Promise.resolve().then(() => {
          require('@github/markdown-toolbar-element')
        })
      } else {
        import('@github/markdown-toolbar-element')
      }
    }
    ```

[^1]: https://jestjs.io/docs/ecmascript-modules

**Looking forward**

One concern that is not implemented is if a component would like to listen to when the module is available _before_ rendering. If this scenario comes up, we could use a convention where each module exposes a `ready()` or `register()` function that returns a `Promise` that resolves when the module is loaded (or rejects if there is an error)

### DIFF
--- a/.changeset/moody-bats-invent.md
+++ b/.changeset/moody-bats-invent.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Refactor MarkdownEditor to import client-only dependencies outside of useEffect
--- a/babel.config.js
+++ b/babel.config.js
@@ -39,7 +39,12 @@ module.exports = {
     },
     test: {
       presets: makePresets('commonjs'),
-      plugins: [...sharedPlugins, ['@babel/plugin-transform-modules-commonjs'], replacementPlugin('test')]
+      plugins: [
+        ...sharedPlugins,
+        ['@babel/plugin-transform-modules-commonjs'],
+        replacementPlugin('test'),
+        'babel-plugin-dynamic-import-node'
+      ]
     }
   }
 }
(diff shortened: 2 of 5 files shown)


joshblack:
FYI @koddsson @iansan5653 here is another approach we could take for `@github` deps that can guard rendering to only the client (or DOM-like environments in the case of tests/JSDOM)

PR #2388, opened 2022-09-30
TreeView: Support async behavior

colebemis:
### Summary

- Adds storybook stories demonstrating how to asynchronously load TreeView items (with success and error states).
- Adds a `TreeView.LoadingItem` component to render while loading TreeView items.

#### Loading items (success)

👉 [Try it out](https://primer-2653f54d2f-13348165.drafts.github.io/storybook/?path=/story/components-treeview--async-success)

https://user-images.githubusercontent.com/4608155/193333060-6942de33-a1c2-45ed-864c-064d115730dc.mp4

#### Loading items (error)

👉 [Try it out](https://primer-2653f54d2f-13348165.drafts.github.io/storybook/?path=/story/components-treeview--async-error)

https://user-images.githubusercontent.com/4608155/193328674-1af138c3-0815-4a0f-a09f-0a786b3507e8.mp4

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/TreeView/TreeView.stories.tsx
```diff
@@ -404,4 +405,148 @@ function TreeItem({
...
+
+  return (
+    <Box sx={{p: 3}}>
+      <nav aria-label="File navigation">
+        <TreeView aria-label="File navigation">
+          <TreeView.Item
+            onExpandedChange={async isExpanded => {
+              if (asyncItems.length === 0 && isExpanded) {
+                // Show loading indicator after a short delay
+                const timeout = setTimeout(() => setIsLoading(true), 500)
```
colebemis, line 429: 🙋 Is showing the loading item after a short delay an accessible pattern? cc @ericwbailey

src/TreeView/TreeView.tsx
```diff
@@ -362,6 +362,23 @@ const LinkItem: React.FC<TreeViewLinkItemProps> = ({href, onSelect, ...props}) =
   )
 }
 
+// ----------------------------------------------------------------------------
+// TreeView.LoadingItem
+
+const LoadingItem: React.FC = () => {
+  return (
+    // TODO: What aria attributes do we need to add here?
+    <Item>
```
colebemis, line 371: 🙋 What aria props should we add to the LoadingItem? cc @ericwbailey

src/TreeView/TreeView.stories.tsx
```diff
@@ -404,4 +405,148 @@ function TreeItem({
...
+            }}
+          >
+            <TreeView.LeadingVisual>
+              <TreeView.DirectoryIcon />
+            </TreeView.LeadingVisual>
+            Directory with async items
+            <TreeView.SubTree>
+              {isLoading ? <TreeView.LoadingItem /> : null}
+              {error ? (
+                <ConfirmationDialog
```
colebemis, line 516: 🙋 Is ConfirmationDialog the right kind of dialog for error messages? cc @ericwbailey

mperrotti: I think so

src/TreeView/TreeView.stories.tsx
mperrotti: Context for @ericwbailey: we do this to avoid showing loading indicators before a user might need them. When users see a lot of loading indicators, I think it makes the UI seem slower than it actually is.

src/TreeView/TreeView.tsx
```diff
@@ -362,6 +362,23 @@ const LinkItem: React.FC<TreeViewLinkItemProps> = ({href, onSelect, ...props}) =
...
+      <LeadingVisual>
+        <Spinner size="small" />
+      </LeadingVisual>
+      <Text sx={{color: 'fg.muted'}}>Loading...</Text>
+    </Item>
+  )
+}
+
+LoadingItem.displayName = 'TreeView.LoadingItem'
```
mperrotti, line 380: I like this. I'm going to start doing this instead of keeping the name of the `const`.

src/TreeView/TreeView.stories.tsx
mperrotti, line 429: 500ms felt _a little_ too long to see expanded content. Maybe we could try 250-300ms?

This is non-blocking feedback and just one person's opinion.

src/TreeView/TreeView.tsx
```diff
@@ -417,7 +434,7 @@ const LeadingVisual: React.FC<TreeViewVisualProps> = props => {
   const children = typeof props.children === 'function' ? props.children({isExpanded}) : props.children
   return (
     <Slot name="LeadingVisual">
-      <Box sx={{color: 'fg.muted'}}>{children}</Box>
+      <Box sx={{display: 'flex', color: 'fg.muted'}}>{children}</Box>
```
mperrotti, line 437: What do we need flexbox for? Is it just a vertical alignment thing? Or are we actually laying out multiple children?

mperrotti (approved):
Approved because we discussed doing the ARIA work in a follow-up PR

src/TreeView/TreeView.tsx
colebemis: When you render an SVG (Octicon) in the leading visual slot, it adds unwanted extra space unless the container is a flex container.

src/TreeView/TreeView.stories.tsx
colebemis, line 429:
```suggestion
                const timeout = setTimeout(() => setIsLoading(true), 300)
```

src/TreeView/TreeView.stories.tsx
```diff
@@ -404,4 +405,148 @@ function TreeItem({
...
+export const AsyncError: Story = args => {
+  const [isLoading, setIsLoading] = React.useState(false)
+  const [isExpanded, setIsExpanded] = React.useState(false)
+  const [asyncItems, setAsyncItems] = React.useState<string[]>([])
+  const [error, setError] = React.useState<Error | null>(null)
+
+  async function loadItems() {
+    if (asyncItems.length === 0) {
+      // Show loading indicator after a short delay
+      const timeout = setTimeout(() => setIsLoading(true), 500)
```
colebemis, line 481:
```suggestion
      const timeout = setTimeout(() => setIsLoading(true), 300)
```

src/TreeView/TreeView.stories.tsx
colebemis: Agreed. I'll reduce it to 300ms

docs/content/TreeView.mdx
```diff
@@ -181,6 +181,10 @@ Since stateful directory icons are a common use case for TreeView, we provide a
 </Box>
 ```
 
+### With asynchronously loaded items
+
+See [storybook](https://primer.style/react/storybook?path=/story/components-treeview--async-success) for examples with asynchronously loaded items.
```
colebemis, line 186:
```suggestion
See [Storybook](https://primer.style/react/storybook?path=/story/components-treeview--async-success) for examples with asynchronously loaded items.
```

src/TreeView/TreeView.tsx
mperrotti: Ok, I figured it was something like that. I've dealt with that before, and I don't understand _why_ it works 🙈 

I think it also works for `grid`

src/TreeView/TreeView.stories.tsx
ericwbailey: I think this should be okay.

src/TreeView/TreeView.tsx
ericwbailey: I'm in the process of testing this out, but I believe we'll need a combination of `aria-hidden` (to hide decorative nodes), `aria-busy` (to communicate loading state), and `aria-live` (to announce state change).

PR #2390, opened 2022-09-30
Use node 18 in all workflows

koddsson:
We shouldn't be using old or unsupported node versions in our workflows.


koddsson:
This seems to be down to Gatsby being on a really old version but we can't upgrade because doctocat peers to Gatsby 2.x 😭

PR #2398, opened 2022-10-03
Change decision for ADR-008-experimental-components

siddharthkp:
After having multiple conversations around this topic to validate our assumptions (see footnotes[^1][^2][^3]), I think we should change [our decision](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-008-experimental-components.md) to the runner up in the initial ADR

---

**tl;dr of diff: Instead of publishing components from [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared), we should create a new repository to keep experimental components that are shared between monolith and non-monolith projects (like memex, docs, even primer)**

why: We already have a federated model in place without putting a label on it. Each application has experimental components built for their purpose (github/github has [react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared), memex has a catalog of [reusable components](https://github.com/github/memex/tree/main/src/client/components)). When there is an opportunity to share a pattern between the monolith and memex, we need a place that's equally convenient for both teams to upstream and maintain long term.

---

Initial assumptions:
1. [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) is the easiest place for monolith developers to contribute to and we should treat memex as the exception, not the rule.
    - While the first part of the statement is true, I don't think we should treat memex as the exception. If it's difficult for developers in memex to upstream components here or maintain them over time, we wouldn't benefit from sharing them at all
    - Writing components to be published as a package is a different challenge, we would require checks to ensure a component does not import dotcom code or other internal dependencies into the published package. Assumptions made inside the monolith might not work for other projects
    - Monolith deployments have a different workflow than traditional open source libraries (example: primer/react) which will slow teams down
2. We would publish all components from react-shared
    - Not all components in react-shared are useful to share between projects, only components that are used across pages like repos/issues and memex are useful to share. 
    -  Components that deal with monolith specific navigation and data fetching logic aren't helpful for non-monolith projects, the added constraints from point 1 would be unnecessary to add to all components
3. Bonus: We did not see ourselves putting experimental or draft components from primer/react in monolith/react-shared, but a github/react-experimental repository outside the monolith would be a good place to host experimental components that are primer/react candidates (instead of keeping them in primer/react), that would add another team as shared maintainer of this new repository.

Sidenote: We could use better visibility into potentially reusable components that have been built inside a application (thinking deployed storybook and docs)

[^1]: https://github.com/github/react-lang/issues/68#issuecomment-1262782081
[^2]: https://github.com/github/github/discussions/235360
[^3]: https://github.com/github/planning-tracking/discussions/1429


contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -47,15 +47,15 @@ The approach that has served us well since Dec 2021 has been "drafts" (along wit
 
 1. npm package: "Experimental" components, as defined by the component lifecycle, should not be part of semantically versioned npm package for `@primer/react`.
 
-2. npm package: Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@github/react-commentbox`
+2. npm package: Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@github/experimental-react-commentbox`
 
-3. The code for experimental components should live in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) and published as individual packages as suggested in point 2.
```
siddharthkp, line 52: This is the relevant diff, split view recommended: https://github.com/primer/react/pull/2398/files?diff=split&w=0#diff-5e30a807664dafbee9f227901a2459809a2c2226e183880f8bf6e05b1f52a7ceL52

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -47,15 +47,15 @@ The approach that has served us well since Dec 2021 has been "drafts" (along wit
...
+3. The code for experimental components should live in a new repository, example: [github/react-experimental](https://github.com/github/react-experimental) and published as individual packages as suggested in point 2.
 
-   The monolith already has a place where reusable react components live. This has the lowest barrier to entry for folks working in the monolith.
```
siddharthkp, line 54: This is the relevant diff, [split view recommended](https://github.com/primer/react/pull/2398/files?diff=split&w=0#diff-5e30a807664dafbee9f227901a2459809a2c2226e183880f8bf6e05b1f52a7ceL52-L54)

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -66,10 +66,8 @@ This decision would increase the amount of work required for non-monolith projec
  
    Keeping experimental components in primer _org_ suggests that the primer _team_ would take up maintenance of these components while they are still candidates (bugs, a11y remedial, etc.). This will be a new parallel workstream for the team and with our current team size, we might not be able to give it the required attention.
 
-2. The code for experimental components should live in a new repository `github/react-shared` or `github/primer-react-candidates` instead of `primer/react-candidates`
+2. The code for experimental components should live inside the monolith in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) instead of a new repository.
```
siddharthkp, line 69: Moving previous decision into "Options considered"

dusave:
> Not all components in react-shared are useful to share between projects, only components that are used across pages like repos/issues and memex are useful to share.

I agree with this statement. My follow up to this is: Will we be sunsetting/removing `react-shared` once our solution is up? If not, I think that will be the default place that developers would contribute to due to its relative location as well as lack of checks/QA.

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -47,15 +47,15 @@ The approach that has served us well since Dec 2021 has been "drafts" (along wit
 
 1. npm package: "Experimental" components, as defined by the component lifecycle, should not be part of semantically versioned npm package for `@primer/react`.
 
-2. npm package: Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@github/react-commentbox`
+2. npm package: Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@github/experimental-react-commentbox`
```
dusave, line 50: I would still suggest that we create a new top level NPM org so we don't have a flood of new packages in `@github`. This would:
1. Make it easier to find them all
2. Reduce the duplication in the names (`@gre/commentbox` [GRE = GitHub React Experimental, just a random idea] as opposed to `@github/experimental-react-commentbox`)

aricwalker:
@shiftkey @iansan5653 when y'all get a chance, could you give this updated proposal a read?  Thanks!

lesliecdubs:
Thanks for iterating on this ADR, @siddharthkp!

**ADR decision maker:** @colebemis  
**Decision deadline:** EOD Wednesday, 12 October or earlier

Your role as decision maker is to gather feedback (including the folks already tagged into this PR to review), probe for discussion and disagreement, and then make an informed decision based on the feedback, identified risks, and trade-offs about whether to adopt the ADR or not. Your role is not to decide based on your own personally preferred approach.

shiftkey (approved):

siddharthkp:
> I agree with this statement. My follow up to this is: Will we be sunsetting/removing react-shared once our solution is up? 
> If not, I think that will be the default place that developers would contribute to due to its relative location as well as lack of checks/QA.

I see what you're hinting at here. 

But, I don't think we should sunset react-shared. The default place should be closest to the application as that's what will help the team ship the fastest. Teams like repos and issues would still benefit from react-shared without any added friction.

However, when an opportunity to share a pattern between the monolith and memex arises, there might be additional work required to upstream to the new repo `github/experimental-react-components` if the component is not built in an agnostic manner.

_I think_ what we are missing here is a shared approach to building components across teams. 🤔

dusave:
> I think what we are missing here is a shared approach to building components across teams

Yes, but it's also about a *singular* place to share across teams. If I as a developer have to check Primer, and then our shared repo, and then finally `react-shared`, I'm more likely to give up. Especially if that component _does_ end up being in `react-shared` which won't be packaged and published. 

I agree, that would be the lowest bar to entry, but I have a strong worry of platform fragmentation.

My suggestion would be that we over communicate. We say we're building a new solution and we plan to deprecate `react-shared` and have those components brought over by X date or something. Or maybe we use that as an experimentation with developers to make sure it's easy to do. Either way, I *strongly* suggest we get rid of it.

siddharthkp:
> Either way, I strongly suggest we get rid of it.

We disagree over this point, we should talk more :)

japf:
Is there any implication in terms of sharing a single Relay Store? For the Issue Viewer component that relies on Relay, we'd ideally have a way to have a single Relay Store when we use this component within Hyperlist.

cc @brasic from PR as this ADR will likely impact PRs at some point as well

contributor-docs/adrs/adr-008-experimental-components.md
```diff
@@ -47,15 +47,15 @@ The approach that has served us well since Dec 2021 has been "drafts" (along wit
...
 
-3. The code for experimental components should live in [github/github/modules/react-shared](https://github.com/github/github/tree/master/app/assets/modules/react-shared) and published as individual packages as suggested in point 2.
+3. The code for experimental components should live in a new repository, example: [github/react-experimental](https://github.com/github/react-experimental) and published as individual packages as suggested in point 2.
```
alexus37, line 52: Working on code which is part of two repos has always been a pain point for me personally. How can we make sure that when extracting code into its own repo we still have a smooth developer experience and the feeling that we are still in one repo?

siddharthkp: 1. My assumption is [npm link](https://docs.npmjs.com/cli/v8/commands/npm-link) would make this feel like one repo / one application
2. canary builds will as well
3. You'll still need to merge the component repo PR to be able to publish a new version. The goal would be to make that super quick with self review checklist and automated review in CI

siddharthkp:
> Is there any implication in terms of sharing a single Relay Store? For the Issue Viewer component that relies on Relay, we'd ideally have a way to have a single Relay Store when we use this component within Hyperlist.

@japf Till now, we've only looked at self-contained presentation components (no opinions about state or data, only internal state if needed).

I think it might be useful to share this use case on https://github.com/github/github/issues/240228 (sharing components that a dependency on the data layer)

Lukeghenco:
I wonder if this approach still makes sense in response to https://github.com/github/planning-tracking/issues/1487. If Memex is in DotCom, this approach seems like an unnecessary hurdle towards sharing components in the near future and something that should likely not exist if/when Memex is moved into DotCom. 

@siddharthkp any thoughts?

contributor-docs/adrs/adr-008-experimental-components.md
mattcosta7: I think NPM link can work, but there's often the "2 version of react" scenario, leading to a 

link react/react-dom from the monolith in separate repo
link component from separate repo in dotcom

which can get hairy. 

Not sure I have much to offer here, but it's worth considering whether we can improve that flow somehow (not sure we can, just thinking if we can, we should try to make it foolproof.  `npm link` is already finnicky at times

iansan5653:
> I wonder if this approach still makes sense in response to https://github.com/github/planning-tracking/issues/1487. If Memex is in DotCom, this approach seems like an unnecessary hurdle towards sharing components in the near future and something that should likely not exist if/when Memex is moved into DotCom.

To be clear, that's still only a pitch and likely has a very long way to go before/if it becomes reality.

I honestly think that even if that pitch were to become reality, a separate repo for experimental components would still be better. A standalone repo would:

- Force components to be self-contained, making it easier to upstream to PRC
- Allow for versioning components to make it easier to manage and ship changes
- Make it much easier to host a documentation site for all of the components
- Make it easier to host Storybook deployments (particularly for individual PR instances)

In a general sense, I think that developing standalone React components is a different task from developing github.com, and having a separate repo allows us to tailor an environment that is better suited for that task.

mattcosta7:
> * Force components to be self-contained, making it easier to upstream to PRC

I think this is the biggest plus.  It constrains the API to ensure reusability, something that isn't quite optimized for when implementing application code

> * Allow for versioning components to make it easier to manage and ship changes

I'd be curious whether we take advantage of that internally much

> * Make it much easier to host a documentation site for all of the components

That _shouldn't_ be blocked from the dotcom repo (or any repo). GitHub pages builds could be generated.

> * Make it easier to host Storybook deployments (particularly for individual PR instances)

That _shouldn't_ be blocked from the dotcom repo (or any repo). GitHub pages builds could be generated.

> In a general sense, I think that developing standalone React components is a different task from developing github.com, and having a separate repo allows us to tailor an environment that is better suited for that task. Whether or not Memex joins the main dotcom repo doesn't change that assertion, because Memex is a part of github.com, not a set of standalone components.

I agree with this, generally.  I think there is the other side of that, where the workflow of push commit in separate repo, npm i in main repo, see change didn't fix issue, repeat is often heftier than just developing both in a local environment, but the enforcement of boundaries is important (and one that often slides in less constrained environments)

`npm link` can help with that, but is often pretty clunky and hard to debug (or even tell it isn't working).

the dance of

The code in the component library needs the react version from the application, but the application needs the code from the component library means double linking (monorepos do this internally via hoisting dependencies or linking shared dependencies - since we won't have that luxury it's not so simple)


All to say, I'm not arguing against the separate repo, but I am curious if it actually adds all of the value we want to get from it. The biggest boost still is that the cycle to land/version changes is much faster and streamlined compared to trying to do that in the monolith codebase - but also, how necessary would it be with colocated code. 


I think the real thing we keep dancing around is that the separate repo that can move faster is good, and should have _all_ monolith react code in it, to maintain the quick feedback cycle and local development workflows, while colocating code that relies on each other.... that isn't antithetical to this being a separate repo _or_ moving memex into the monolith though.

mattcosta7:
> Agree with this. However, like @alexus37 in [this comment](https://github.com/primer/react/pull/2398#discussion_r988077708), I'm concerned with developer experience. I was futzing with the experimental markdown editor component trying to determine whether I needed to make a change at all, and that proved challenging. If this process is more seamless, I'm all for it. I could not for the life of me get npm link to work. 🤔

I think this need to be as simple as 

`script/setup-react-experimental` and it needs to _just work_ from then on.  for this to be successful. 

that's not a simple order, but I think the only way to make this process successful. 

Then a quick and easy way to push code that can be used to replace a version in the repo via npm i. to avoid lots of context dropping and time spent waiting for xxxx to build

jdrush89:
> My follow up to this is: Will we be sunsetting/removing react-shared once our solution is up?
@dusave I think we will still need a place to share monolith specific code that is allowed to have monolith specific dependencies. Things like the ref selector or the symbols panel make sense to share between Repos and PRs but don't necessarily make sense in a public shared repo.

dusave:
> but don't necessarily make sense in a public shared repo.

Just want to make this abundantly clear here: The proposed repo will **not** be public. It will be internal only, with the option to upstream components to Primer if it is determined to align with their vision/goals.

Further, while I agree that sharing components within the monolith definitely makes sense, I don't think that sharing components in multiple locations makes sense. IMO having a singular location that is accessible to all applications, regardless of whether they exist in or out of the monolith is key here.

jdrush89:
Who are the consumers of this new repo going to be? If it's just the monolith and memex then its purpose is pretty clear to me, it should hold components that are common to the monolith and memex. If memex was part of the monolith we wouldn't be creating a separate repo for shared components, because there is overhead in developing components outside of the repo they are used in vs importing them directly. Components that are very likely not needed outside of the monolith shouldn't have to live in this new repo. What's likely to happen anyways is that something like the RefSelector would just live under Repos in the monolith and if PRs needed it they would import it directly. It would just be less transparent that it's being shared.

Overall, in determining where to keep shared components I think the question to ask is 'what is the direction of React development at GitHub?' Will more separate repos like memex be popping up, will memex consolidate with the monolith, or will we keep the status quo? If memex were to move into the monolith, I think having everything in react-shared would make a lot more sense. We could still have this separate repo in the interim, but we should identify components that are actually needed in both and not try to put every piece of shared code into it.

dusave:
> Who are the consumers of this new repo going to be?

The consumers would be every React developer at GitHub. Right now, the only outlier from the monolith is Memex, yes. However, I have had conversations with other teams about their explorations in moving away from the monolith as well.

> If memex was part of the monolith

Again, this is a proposal. This is not going to happen anytime soon (most likely > 6 months if not longer). Who knows what the state of both this repo, React at GitHub, and the monolith will be in at that time.

> there is overhead in developing components outside of the repo they are used in vs importing them directly

I'll grant you the additional overhead in the initial contribution. After that, it should be much smoother. 

> Components that are very likely not needed outside of the monolith shouldn't have to live in this new repo.

As to _whether_ they should be, just for the sake of argument for a minute: Let's say we keep components in `react-shared`. Even in that scenario, the current system is not maintainable. Currently, there is no versioning, no checks, no documentation, and no quality bar or forward advancement. This is not ok. So _even if_ we were to utilize `react-shared`, it would be going through a refactoring and process adjustment to align it with where we want components to go going forward. Based on some of your comments (and I could be misinterpreting), it sounds like you think we should continue to allow relative path imports, no publishing, and just allow any and all components in that directory. I strongly disagree with that strategy and have outlined my reasons why in the previous iteration of this ADR.

> What's likely to happen anyways is that something like the RefSelector would just live under Repos in the monolith and if PRs needed it they would import it directly.

If that is likely to happen, then we should work with the teams to ease the friction. However, as devs in a large organization, we also must accept some ownership in maintaining a good system, which includes practicing good coding hygiene. 

---

Bottom line is, we don't know where things are going in the future wrt React at GitHub. We can only guess, strategize, and work the with teams that are working hard and making it a reality. I agree that this solution has cons, but IMHO it has many more pros that outweigh them. However, I will also admit that I believe that moving into a monorepo (similar to the monolith, but constructed in a more modular way) is the way to go, and further I would have this repo be part of that. But we don't have that today and it's unclear if that's where we are heading or not.

Currently, the monolith dev experience has some substantial friction points, both with the size and with the deploy queue. I know that this has been being worked on for awhile now and am excited to see where it goes. Once that happens, it may make more sense to migrate this back in. The beauty of this is it wouldn't change the consumers' code. They would still import it as a separate package just as before.

colebemis (approved):
As the [decision maker](https://github.com/primer/react/pull/2398#issuecomment-1266220745) for this ADR, I’ve carefully read through the comments and discussed this change with engineers working on Repos, PRs, Memex, Web Systems, and Primer.

## Decision

- We will move forward with this ADR by validating the approach with a small set of components that need to be shared between Memex and the monolith (e.g., MarkdownEditor, DatePicker). This work is already in progress ([tracking issue](https://github.com/github/github/issues/240228)).
- If Memex decides to [move into the monolith](https://github.com/github/planning-tracking/issues/1487), we will reevaluate this decision.
- Engineers can continue to build React components in their respective repositories. When we notice a need to share a component between the monolith and Memex, we will hoist the component into the github/experimental-react-components repository. In other words, we will not remove the `react-shared` directory from github/github.
- We will define clear ownership on all files in the github/experimental-react-components repository using [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners).
- We will build the toolchain (i.e., linting, testing, storybook) in github/experimental-react-components such that it can be reused in github/github, github/memex, and primer/react.

## Tradeoffs

Here are some tradeoffs we’re accepting as part of this decision:

- Introducing a separate repo for component development will create more friction in the development process. We will do our best to mitigate this tradeoff with tooling and documentation to create a low barrier to entry for contribution.
- Publishing a package for each experimental React component will create versioning challenges that don’t exist when all components are in the same repository.
- A shared repository could create confusion around ownership. We will try to reduce confusion with clear documentation and codify the ownership model with a CODEOWNERS file.

## Open questions

Here are some open questions that I’ve collected from the comments and 1:1 discussions:

- Who will be responsible for maintaining the infrastructure (e.g., bundling, publishing) of the  github/experimental-react-components repo?
- Who will be responsible for upstreaming components from github/experimental-react-components to primer/react?
- Who would be responsible for migrating the usage of experimental components to Primer components after a component is upstreamed?
- What will the release cadence be for packages in the github/experimental-react-components repo?
- Who will be responsible for publishing new releases of experimental components?
- How will people find components that _could_ be shared but aren’t yet? In other words, how will people find components to move into the github/experimental-react-components repo? 
- What would make us reevaluate the decision in this ADR?

We don't need to answer all these questions now, but we will need to address them as part of the [initiative to validate this contribution model](https://github.com/github/github/issues/240228).

Update: I've moved these questions into a [separate discussion](https://github.com/github/experimental-react-components/discussions/49). Let's continue the conversation there.

## Thank you

Huge thank you to the following people for sharing your feedback 💖

- @alexus37
- @dgrief
- @dusave
- @iansan5653
- @japf
- @jdrush89
- @jfuchs
- @keithamus
- @Lukeghenco
- @mattcosta7
- @milemons
- @rezrah
- @shiftkey
- @talum

PR #2399, opened 2022-10-03
Primitives v2: map existing space values to css variables

siddharthkp:
Replaces space px values with base css variables from [tokens-v2-private](https://github.com/github/primer/issues/889). 

[See it in action](https://primer-c8129d4358-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav--default-nav) or [see snapshots diff](https://github.com/primer/react/pull/2399/files?file-filters%5B%5D=.snap&show-viewed-files=true)

The assumption is that these css variables would be [imported by the application independently](https://github.com/github/github/blob/master/app/assets/stylesheets/bundles/design-tokens-v2-private/index.scss) behind a feature flag

### Known issues:

Negative values don't work yet. 

`left: -2` results in `-var(--base-size-8,8px)` instead of `calc(-1 * var(--base-size-8,8px))`

chromatic caught these in 3 components, [see diff →](https://www.chromatic.com/build?appId=61a90feace7802003a4d9c45&number=1705)

1. [ActionList.Item](https://github.com/primer/react/blob/main/src/ActionList/Item.tsx#L56)
2. [TreeView](https://github.com/primer/react/blob/main/src/TreeView/TreeView.tsx#L234)
3. [ActionMenu](https://github.com/primer/react/blob/main/src/ActionMenu.tsx#L89)

We need to fix this in styled-system/styled-components because applications can also use negative values with sx

### Limitations

If we want this change to be internal and backward-compatible, 
1. we can only replace values that are already defined in theme-preval today. New values like `--base-size-12`, 20, 28, 36, 44 cannot be introduced in between
2. we can only change the internal implementation, we can't change the API for space properties to be closer to primitives v2
    ```js
     marginLeft: 4 // resolves to theme.space[4] which is var(--base-size-24) and not --base-size-4
    ```

We could introduce a parallel scheme that is introduced slowly (this would need some changes in how we merge themes)

```js
  marginLeft: 'size.4' // instead of marginLeft: 4
  // or
  marginLeft: 'space.4'
```

I generally like this API better because it makes it clear that this is coming from tokens and not a px value

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,6 +1,7 @@
 import {addons} from '@storybook/addons'
 import {withPerformance} from 'storybook-addon-performance'
 import {withThemeProvider, toolbarTypes} from '../src/utils/story-helpers'
+import '@primer/primitives/tokens-v2-private/css/tokens/base/size/size.css'
 
 export const globalTypes = toolbarTypes
 export const decorators = [withThemeProvider, withPerformance]
--- a/src/NavList/__snapshots__/NavList.test.tsx.snap
+++ b/src/NavList/__snapshots__/NavList.test.tsx.snap
@@ -26,8 +26,8 @@ exports[`NavList renders a simple list 1`] = `
 .c0 {
   margin: 0;
   padding-inline-start: 0;
-  padding-top: 8px;
-  padding-bottom: 8px;
+  padding-top: var(--base-size-8,8px);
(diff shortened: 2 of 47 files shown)


lesliecdubs:
@siddharthkp did we want this to get closed? Seems like this might be work we'd want to keep 👀

PR #2405, opened 2022-10-04
Fix loss of focus bug

nicolleromero:
Update: pausing on this proposed fix as I proposed a more targeted fix in the [linked PR](https://github.com/github/memex/pull/12471 ).

This PR fixes a bug where an input can lose focus when a user is typing.

Closes https://github.com/github/issues-graph/issues/1049

### Screenshots

See https://github.com/github/memex/pull/12471 for before and after screenshots and additional details.

### Merge checklist

- [ ] Added/updated tests
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/drafts/MarkdownViewer/MarkdownViewer.tsx
+++ b/src/drafts/MarkdownViewer/MarkdownViewer.tsx
@@ -107,7 +107,10 @@ const MarkdownViewer = ({
   // If we were to inject the `...htmlContainer.children` instead of the container element itself,
   // those children elements would be moved from the `htmlContainer` to the `outputContainer`. Then if
   // other effects use `htmlContainer.querySelectorAll`, they wouldn't find any elements to affect
-  useEffect(() => outputContainerRef.current?.replaceChildren(htmlContainer), [htmlContainer])
+  useEffect(() => {
+    if (document.activeElement && outputContainerRef.current?.contains(document.activeElement)) return
+    outputContainerRef.current?.replaceChildren(htmlContainer)
+  }, [htmlContainer])
 
   return loading ? (
     <Box sx={{display: 'flex', justifyContent: 'space-around', p: 2}}>


iansan5653 (approved):
Just to add some more context here for the Primer team: we encountered an issue where live updates might come in as the user is interacting with focusable UI elements (ie, `<input>`) inside the `MarkdownViewer`. When those updates change the viewed HTML, the `MarkdownViewer` replaces all its contents with the new HTML and this clears the user's current focus. By avoiding the replacement if the user is focused inside, we can avoid clearing the focus.

I do think one improvement here might be to, instead of completely ignoring the change, _delay_ it until focus leaves the viewer container. I'm not sure if that's really necessary but it seems like it might reduce edge cases.

broccolinisoup:
> Just to add some more context here for the Primer team: we encountered an issue where live updates might come in as the user is interacting with focusable UI elements (ie, `<input>`) inside the `MarkdownViewer`. When those updates change the viewed HTML, the `MarkdownViewer` replaces all its contents with the new HTML and this clears the user's current focus. By avoiding the replacement if the user is focused inside, we can avoid clearing the focus.
> 
> I do think one improvement here might be to, instead of completely ignoring the change, _delay_ it until focus leaves the viewer container. I'm not sure if that's really necessary but it seems like it might reduce edge cases.

@iansan5653 👋🏼 Thanks for providing the context! Delaying might be a solution, sounded good to me but I also have a concern that what if the updated content might possibly have some context that might affect the content user is about to submit? Maybe an edge case of an edge case 😬

nicolleromero:
@iansan5653 @broccolinisoup I agree with the delay strategy, and we're all set with the targeted fix in Memex for the time being. It might be helpful to have this solution in our collective back pocket if we receive any similar reports where we need to suppress a live update in a similar scenario.

PR #921, opened 2020-11-19
Remove duplicate padding on Dropdown links

emplums:
For some reason, we were applying some padding to any links in `Dropdown.Item`. I've removed that!

Closes #902 

#### Before
image

#### After
image

### Merge checklist
- [ ] Added or updated TypeScript definitions (`index.d.ts`) if necessary
- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/components/blob/main/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2406, opened 2022-10-05
[UnderlineNav] Accessibility Remediations

broccolinisoup:
This PR addresses the [accessibility feedback](https://github.com/github/primer/issues/921#issuecomment-1267507562) from @ericwbailey on UnderlineNav. It also includes re-styling the arrow buttons to give more real estate for items for be selected as well as styling the focus ring of these buttons. I thought it will be good to ship it with accessibility remediations as focus rings were not in a great state to be sent to sign off review. Please let me know if anyone has any concerns with it 🙌🏼 

Storybook link: https://primer-9cc12887c1-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav--internal-responsive-nav

Accessibility Eng Review ticket: [#921](https://github.com/github/primer/issues/921) 

### Remediations

- Add aria-label to the `UnderlineNav`(nav item) on the stories and docs
- Update the arrow buttons' accessible names by making the `UnderlineNav`'s ariaLabel prop available to them
- Scroll focused item into the view
- Add `aria-disabled` attribute to the arrow buttons and conditionally set true/false when they become visible/hidden ( visibility depends on if there is any overflowing item on left/right)
- Trigger focus on the hidden arrow button after their `aria-disabled` attribute set to true. (Main motivation is to trigger a reread on the assistive technology with the `aria-disabled=true` to communicate the arrow button is no longer there)
- Update selectedLink style to ensure it is visible in Forced Colours mode
- Add `focusable=true` to icons that are `aria-hidden=true`

### Screenshots and screen recordings

https://user-images.githubusercontent.com/1446503/194796494-458b550d-287f-44c7-8ab8-e52a90a600f2.mov

#### Min touch target size

Screen Shot 2022-10-10 at 12 48 08 pm

#### Focus rings (Feedback needed)
Screen Shot 2022-10-10 at 12 48 34 pm

### Merge checklist

- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [x] Tested in Edge


src/UnderlineNav2/styles.ts
```diff
@@ -92,15 +92,23 @@ export const getLeftArrowHiddenBtn = (theme?: Theme) => ({
   opacity: 0,
   top: 0,
   bottom: 0,
-  left: 0
+  left: 0,
+  /* This is intentionally set to prevent any visual focus ring from showing up
+   * when a hidden element gets focus just to trigger assistive technology to reread */
+  '&[aria-disabled="true"]:focus': {
+    outline: 'none'
+  }
```
ericwbailey, line 100:
```suggestion
  left: 0
```

This might seem counter-intuitive at first, but we'll actually want to preserve the focus ring. The control is still interactive in this context in that it communicates being conditionally disabled. 

A user who relies on focus rings to understand what is and is not interactive would need the ring preserved here, with the button's visuals (and potential announcement) communicating conditionally-applied state.

src/UnderlineNav2/styles.ts
```diff
@@ -92,15 +92,23 @@ export const getLeftArrowHiddenBtn = (theme?: Theme) => ({
...
 export const getRightArrowHiddenBtn = (theme?: Theme) => ({
   ...getArrowBtnStyles(theme),
   opacity: 0,
   top: 0,
   bottom: 0,
-  right: 0
+  right: 0,
+  '&[aria-disabled="true"]:focus': {
+    outline: 'none'
+  }
```
ericwbailey, line 111:
```suggestion
  right: 0
```

src/UnderlineNav2/styles.ts
```diff
@@ -189,6 +197,12 @@ export const getLinkStyles = (
     bg: selectedLink === ref ? theme?.colors.primer.border.active : 'transparent',
     borderRadius: 0,
     transform: 'translate(50%, -50%)'
+  },
+  '@media (forced-colors: active)': {
+    '::after': {
+      // Support for Window Force Color Mode https://learn.microsoft.com/en-us/fluent-ui/web-components/design-system/high-contrast
+      bg: selectedLink === ref ? 'LinkText' : 'transparent'
+    }
```
ericwbailey, line 205: Love it 🎉

ericwbailey (changes requested):
I can't speak to the underlying logic, but the functionality works from an assistive technology perspective.

One small comment on preserving focus rings, otherwise I'd say this is good to go!

src/UnderlineNav2/styles.ts
broccolinisoup: Hi @ericwbailey 👋🏼 Thank you so much for your review! I recorded a quick video to reply here because needed to demonstrate something, I hope it is helpful! Let me know what you think 😊 

https://user-images.githubusercontent.com/1446503/194270894-c1e00a46-0f5d-4c46-b5e9-d4947dde9312.mp4

src/UnderlineNav2/styles.ts
broccolinisoup: Because I updated the entire styles of the arrow buttons, I cannot apply your suggestions here, sorry but I removed them in the updated styles. I'm resolving this conversation here and we can discuss my concern further in the above comment 🙂

broccolinisoup:
@danielguillan I have updated the arrow button styles and the focus ring! I would appreciate your review 🙏🏼 I might have to address your feedback if there is any on a different PR because I'm aiming to send the UnderlineNav for the sign off a11y review early next week 🥳 Let me know if you have any concerns with it 🙌🏼

danielguillan:
@broccolinisoup! Thank you for the updates! 

We'll need to ensure the arrow buttons provide a minimum touch target size of `44px`. This is where the arrow slide-in/out effect becomes needed so we don't end up with large paddings at both ends of the component. We can do this in a new PR. 

The UnderlineNav component scroll behavior arrow buttons have a 44 pixel touch target size

src/UnderlineNav2/styles.ts
ericwbailey: Oh, this is interesting. Is there a place I can test this behavior?

broccolinisoup: Yes, this PR's [storybook link](https://primer-9cc12887c1-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav--internal-responsive-nav) should work. Thank you!!

broccolinisoup: Just for providing more context here, it happens because we eventually hide the arrow button when there is no overflow on the direction that the button controls. And we add `aria-disabled=true` and re-focus on it just before hiding it. I hope it helps!

broccolinisoup:
> @broccolinisoup! Thank you for the updates!
> 
> We'll need to ensure the arrow buttons provide a minimum touch target size of `44px`. This is where the arrow slide-in/out effect becomes needed so we don't end up with large paddings at both ends of the component. We can do this in a new PR.
> 
> The UnderlineNav component scroll behavior arrow buttons have a 44 pixel touch target size

Hi @danielguillan sorry that I ignored the minimal touch target! I have pushed another commit with the slide in/out. I would appreciate your review 🙏🏼 It is working fine on my side and I like it, definitely less paddings and more space. 

Also, could I get a review on the focus rings as well? I gave enough paddingX to the button to make the minimal touch target but not sure about paddingY. It is rectangular now, so not sure if it is a desired state. Let me know what you think 😊 

Note: I did the slide in/out here because there were still other concerns holding this PR - just to let you know 😊

src/UnderlineNav2/styles.ts
broccolinisoup: 👋🏼 again @ericwbailey! I pushed some update on the buttons since my last message above and this behaviour doesn't really occur anymore which is good I believe? I added a screen recording on the description and the [storybook link](https://primer-9cc12887c1-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav--internal-responsive-nav) has the updated version. It doesn't occur anymore because we used to hide (opacity:0) the buttons with a `transition` but now we do `display:none` with a `transform:translateX(dynamicValue)`  when there is no overflow but the rest is same as in adding `aria-disable=true` and refocusing. I am sorry for pushing new changes; we didn't really have many alternatives to solve [this specific problem](https://github.com/primer/react/pull/2406#issuecomment-1271307239) 😞 Looking forward to hearing what you think!

danielguillan:
@broccolinisoup Thanks for all the updates! 🙌 

### Focus ring + touch target

I don't think we need to override the paddings of the arrow buttons. We want a small IconButton with its default size and focus ring and a safe touch target area of '44 × 44 pixels`. 

UnderlineNav arrow buttons with focus rings and safe touch area

The current wrapper `div` is already the right size, so we might be able to leverage that one for the touch target.
Arrow button wrapper div measures 44 pixels by 44 pixels

Also, while playing with the scroll behavior on mobile I've seen the focus ring flashing on the arrow buttons when they slide in and out. But I don't seem to be able to reproduce it every time…

### Component overflow hidden

We need to set the overflow of the root element to `hidden` so the arrow buttons aren't visible outside of it when sliding out.

Without `overflow: hidden` (current):
Arrow button overflows the UnderlineNav root element

With `overflow: hidden`:
Arrow button does not overflow the UnderlineNav root element

### Fade effect

Sometimes, the fade effect gradient layer becomes a solid background when scrolling and using the arrow buttons.

Solid background instead of fade effect

broccolinisoup:
> @broccolinisoup Thanks for all the updates! 🙌
> 
> ### Focus ring + touch target
> I don't think we need to override the paddings of the arrow buttons. We want a small IconButton with its default size and focus ring and a safe touch target area of '44 × 44 pixels`.
> 
> UnderlineNav arrow buttons with focus rings and safe touch area
> 
> The current wrapper `div` is already the right size, so we might be able to leverage that one for the touch target. Arrow button wrapper div measures 44 pixels by 44 pixels
> 
> Also, while playing with the scroll behavior on mobile I've seen the focus ring flashing on the arrow buttons when they slide in and out. But I don't seem to be able to reproduce it every time…
> 
> ### Component overflow hidden
> We need to set the overflow of the root element to `hidden` so the arrow buttons aren't visible outside of it when sliding out.
> 
> Without `overflow: hidden` (current): Arrow button overflows the UnderlineNav root element
> 
> With `overflow: hidden`: Arrow button does not overflow the UnderlineNav root element
> 
> ### Fade effect
> Sometimes, the fade effect gradient layer becomes a solid background when scrolling and using the arrow buttons.
> 
> Solid background instead of fade effect

Great feedback, thanks!! I'll address them in the next PR 🙌🏼

src/UnderlineNav2/styles.ts
ericwbailey: Thanks for the detail for the changes! I just re-tested it on VoiceOver on iOS and it sounds like it works as intended 🎉

ericwbailey (approved):
🚀

pksjce (approved):
Lets go! 🚀

PR #2409, opened 2022-10-05
Codespaces broken due to invalid config

mattcosta7:
No component changes here

The devcontainer was updated, but in an invalid way (json without a comma). this fixes that issue so that codespaces build properly


joshblack (approved):

koddsson (approved):
Whoops! Thanks for this!

PR #2410, opened 2022-10-05
ci(visual_testing): use workflow_run trigger for visual_testing

joshblack:
This PR updates the Visual Testing workflow in order to be safely run from PRs from branches and forks without. Specifically, it uses the `workflow_run` event to run after the CI workflow has completed and only interacts with a built artifact, `storybook-static`, that it provides to Chromatic.

The CI workflow has been updated to include a job that builds and uploads the output of `storybook` to be used by Chromatic in the Visual Testing workflow.

Reference: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

**Testing**

It seems like `workflow_run` triggers may only be tested after merged into `main` so it doesn't seem possible to test this in the PR here. As a result, I tried to pull out this pattern into a separate repository here: https://github.com/joshblack/actions-security to demonstrate this behavior. Specifically:

- Here is a PR from a user without privileges: https://github.com/joshblack/actions-security/pull/2 (dependabot)
  - Confirm that chromatic still reports to the PR in the status checks
- Here is the VRT workflow file: https://github.com/joshblack/actions-security/blob/main/.github/workflows/vrt.yml
- Here is the CI workflow file, specifically the storybook job: https://github.com/joshblack/actions-security/blob/main/.github/workflows/ci.yml#L23


.github/workflows/ci.yml
```diff
@@ -5,38 +5,139 @@ on:
...
         working-directory: examples/nextjs
         run: npm ci
-
       - name: Build examples
         working-directory: examples/nextjs
         run: npm run build
+
+      # Consumer tests
+      #
+      # `prepare` is a special case in `npm` and likes to run all the time, even
```
broccolinisoup, line 81: Interesting! Thanks for documenting here 🙌🏼

PR #2412, opened 2022-10-07
Add Storybook Storysource addon

langermank:
[Storysource](https://storybook.js.org/addons/@storybook/addon-storysource) adds a panel with a code snippet of the story "source" code for easy copy/paste. The source code view depends on the format of the story, and unfortunately isn't compatible with `template.bind()` syntax. However, if we treat "Feature" stories as code snippets and allow the Playground story to use controls, we kind of get the best of both worlds. We might consider investing in our own source code addon that is more consistent.

Closes https://github.com/github/primer/issues/1385

### Screenshots

It will generally look something like this:

Storybook addons panel showing the source panel active, which renders a code snippet of the story

I also fixed the duplicate theme decorator in some stories, which improves the docs view:

Before:
Storybook docs page with a code snippet showing base Storybook code instead of component markup

After:
Storybook docs page with a code snippet showing component markup

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,11 +1,11 @@
 module.exports = {
   stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
   addons: [
+    {name: '@storybook/addon-essentials', options: {backgrounds: false}},
+    '@storybook/addon-storysource',
+    '@storybook/addon-interactions',
     '@storybook/addon-a11y',
     '@storybook/addon-links',
-    '@storybook/addon-interactions',
-    '@storybook/addon-storysource',
-    {name: '@storybook/addon-essentials', options: {backgrounds: false}},
     'storybook-addon-performance/register',
     {name: 'storybook-addon-turbo-build', options: {optimizationLevel: 2}},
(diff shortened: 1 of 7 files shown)


.storybook/main.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
     '@storybook/addon-a11y',
     '@storybook/addon-links',
     '@storybook/addon-interactions',
+    '@storybook/addon-storysource',
```
joshblack, line 7: For the order of the panels, how would you feel about:

- Controls
- Story
- Interactions
- Accessibility
- Actions
- Performance

Also, do you think with the StorySource being added that removing the `@whitespace/storybook-addon-html` would be appropriate or is that serving a different use-case?

joshblack (approved):
🥳

.storybook/main.js
langermank: Oh I love this! I didn't think about changing the order, let me give that a go.

The `html` addon shows the source `html` markup which can be helpful for accessibility testing (or just quickly getting feedback about dom structure as you push changes). However, I think its broken in development 😦 it looked like it had a hate around it to only run in dev mode, but its not running at all for me.

PR #2419, opened 2022-10-10
feat(bundle): add exports for test-helpers

joshblack:
**Background**

Certain files from `utils` are not included in the package, specifically if they are not imported by a file through an entrypoint (like `src/utils/test-helpers.tsx`)

**Approach**

This PR adds additional entrypoints for files expected to be in the package. Currently, files under `utils` will not be included in the bundle unless they are referenced/imported by existing code. In the case of test helpers, this file is now explicitly included as it's export pattern would not fall under the existing `lib-esm/*`.


siddharthkp (approved):
Looks good! ship it :)

siddharthkp:
If you feel confident about this, let's merge this

siddharthkp:
Update: I tried to see what's the diff in `lib` and `lib-esm` from this PR, not very confident now 😅: https://github.com/primer/react/pull/2421/files

rollup.config.js
```diff
@@ -1,10 +1,35 @@
...
+
+  // "./lib-esm/*"
+  ...glob.sync(['src/*', 'src/*/index.js'], {
+    cwd: __dirname,
+    onlyFiles: true,
+    // Note: ignore theme-preval as it is handled through the theme import and
+    // specifying it as an entrypoint creates an intermediate file
+    ignore: ['src/theme-preval.js']
+  }),
+
```
joshblack, line 28:
```suggestion
```

siddharthkp (approved):
paired and tested this, let's ship it

PR #2421, opened 2022-10-11
(don't merge) See diff for #2419

siddharthkp:
- context: https://github.com/primer/react/pull/2419#issuecomment-1274738617


lib-esm/ActionList/Description.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import Box from '../Box.js';
-import '../sx.js';
+import '@styled-system/css';
```
siddharthkp, line 3: this feels like a strange replacement?

also why was `import '../sx.js'` was imported without any import specifiers??

lib-esm/Autocomplete/AutocompleteMenu.js
```diff
@@ -3,7 +3,7 @@ import { scrollIntoView } from '@primer/behaviors';
 import { ActionList } from '../deprecated/ActionList/index.js';
 import { useFocusZone } from '../hooks/useFocusZone.js';
 import Box from '../Box.js';
-import Spinner from '../Spinner.js';
+import StyledSpinner from '../Spinner.js';
```
siddharthkp, line 6: strange but okay, pretty harmless

lib-esm/Avatar.js
```diff
@@ -25,6 +25,5 @@ Avatar.defaultProps = {
   alt: '',
   square: false
 };
-var Avatar$1 = Avatar;
 
-export { Avatar$1 as default };
+export { Avatar as default };
```
siddharthkp, line 29: there's a bunch of removed `$1` which honestly i like that

lib/index.js
```diff
@@ -92,7 +92,7 @@ var merge__default = /*#__PURE__*/_interopDefaultLegacy(merge);
 exports.theme = themePreval;
 exports.themeGet = constants.get;
 exports.BaseStyles = BaseStyles;
-exports.ThemeProvider = ThemeProvider["default"];
+exports.ThemeProvider = ThemeProvider.ThemeProvider;
```
siddharthkp, line 95: as long as these 2 are the same, all good!

lib/utils/test-helpers.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
siddharthkp, line 1: nice, this is the file we want

lib/constants.js
```diff
@@ -43,8 +43,12 @@ const whiteSpace = system({
 });
 const TYPOGRAPHY = compose(styledSystem__namespace.typography, whiteSpace);
 // Border props
-compose(styledSystem__namespace.border, styledSystem__namespace.shadow);
+const BORDER = compose(styledSystem__namespace.border, styledSystem__namespace.shadow);
```
siddharthkp, line 46: compose used to called without assigning to anything, that's strange?!

also, were we not exporting border and layout?

lib-esm/constants.js
```diff
@@ -19,6 +19,8 @@ const whiteSpace = system({
 });
 const TYPOGRAPHY = compose(styledSystem.typography, whiteSpace);
 // Border props
-compose(styledSystem.border, styledSystem.shadow);
+const BORDER = compose(styledSystem.border, styledSystem.shadow);
+// Layout props
+const LAYOUT = styledSystem.layout;
 
-export { COMMON, TYPOGRAPHY, get };
+export { BORDER, COMMON, LAYOUT, TYPOGRAPHY, get };
```
siddharthkp, line 26: i'm definitely missing something because it looks like border and layout were not exported till now? is it because they were tree shaken?

lib-esm/utils/test-helpers.js
```diff
@@ -0,0 +1,13 @@
+// JSDOM doesn't mock ResizeObserver
```
siddharthkp, line 1: nice

lib-esm/constants.js
joshblack: I think so, it seemed like they aren't re-exported through `src/index.ts` and were removed based on usage in the codebase

lib/constants.js
joshblack: I think what happens when `BORDER` is removed is that it considers `compose` and if it is pure. Looking at `styled-system`, it doesn't have `sideEffects` listed so it can't assume `compose` is pure and has to leave it in case of side-effects.

lib-esm/hooks/index.js
```diff
@@ -0,0 +1,12 @@
+export { useOnOutsideClick } from './useOnOutsideClick.js';
```
siddharthkp, line 1: this is a big win! love it

lib-esm/constants.js
siddharthkp: yay it's fixed!

siddharthkp:
job done, let's ship

Issue #2423, opened 2022-10-11
`MarkdownEditor` suggestions break when querying by 3 or more characters


iansan5653:
We noticed an odd bug in Projects today (https://github.com/github/memex/issues/12620) -- when the user types more than 2 characters to filter the suggestions list, the suggested values become completely incorrect. For 2 or fewer characters, the suggestions make sense, but something seems to break after that. I was able to reproduce this by adding a few more suggestions to the `MarkdownEditor` story.

In this list, "GitHub" should be first and "mnl" shouldn't appear at all:

<img width="290" alt="Screenshot of the MarkdownEditor suggestions. The user has typed '@git' and the first suggestion is the user 'mnl'." src="https://user-images.githubusercontent.com/2294248/195124762-00458701-4676-4189-bd4a-22203d18c593.png">

iansan5653:
I've found out that this is because `fzy.js` (the fuzzy filtering library we are using) returns `Infinity` if the length of the match item is less than the length of the query. Very odd.

Issue #2422, opened 2022-10-11
`MarkdownEditor` should allow lazy-loading suggestions


iansan5653:
In order to reduce API load and improve time-to-render, we want to lazy-load suggestions in Projects by waiting until the suggestions are actually needed before making the API request (see https://github.com/github/memex/issues/12396). The `MarkdownEditor` doesn't support this at the moment, so it would like require changing the props to accept functions instead of (or in addition to) the suggestion arrays.

I am already working on this; just opening an issue to track over here.

japf:
cc @anleac as we discussed this topic today while working on https://github.com/github/issues/issues/3377

PR #2424, opened 2022-10-11
Fix `MarkdownEditor` suggestions filtering bug and allow lazy-loading suggestions

iansan5653:
1. Fixs a bug where suggestions values shorter than the filter query would _always_ appear first in the list, even if they don't match the query. This fixes https://github.com/primer/react/issues/2423
2. Adds support for lazy-loading suggestions data so it doesn't have to be ready when the editor is mounted. This was done by changing the prop types to optionally accept an async function instead of a plain old array. This way, the consumer can wait to make an API request until the `MarkdownEditor` asks for the data. Then they can cache the data for future rendering. The new prop types look like `emojiSuggestions?: Emoji[] | (() => Promise<Emoji[]>)`. This resolves https://github.com/primer/react/issues/2422
3. [minor] Improves the `MarkdownEditor` test suite to perform the mocking of `offsetHeight` on the `HTMLElement` prototype in a `beforeAll` instead of on the element instances themselves in a `layoutEffect`. This is far less buggy and removes any chance of race conditions (which I was encountering when moving to async event handlers for rendering suggestions).

### DIFF
--- a/src/drafts/MarkdownEditor/MarkdownEditor.stories.tsx
+++ b/src/drafts/MarkdownEditor/MarkdownEditor.stories.tsx
@@ -1,6 +1,6 @@
 import {DiffIcon} from '@primer/octicons-react'
 import React, {Meta} from '@storybook/react'
-import {useState} from 'react'
+import {useRef, useState} from 'react'
 import BaseStyles from '../../BaseStyles'
 import Box from '../../Box'
 import MarkdownEditor, {Emoji, Mentionable, Reference, SavedReply} from '.'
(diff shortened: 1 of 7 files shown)


Lukeghenco (approved):
The lazy loading of emojis, references and suggestions is killer. 

The bug fixes here are stuff we've logged in the pull requests app we are building too. This should resolve our same issues.

Thanks for leading the charge here.

siddharthkp (approved):
Scoped to drafts, ship it

PR #2425, opened 2022-10-12
[UnderlineNav] : Final implementation and UI fixes & docs improvement

broccolinisoup:
TLDR: This PR addresses a couple of implementation issues, UI fixes as well as some documentation improvement. 

---

Storybook link: https://primer-7f80a98c95-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav--internal-responsive-nav
Docs link: https://primer-7f80a98c95-13348165.drafts.github.io/drafts/UnderlineNav2

---

### Issues addressed in this PR:
- Render menu items as the given `as` prop to the UnderlineNav
- Wrap the `nav` with a container that `overflow: hidden` to prevent arrow buttons showing up when they are supposed to be hidden. 
- Display loading counters for all nav items.
- Add `string` type to `counter` prop to support cases like `12K`
- Remove `event.preventDefault()` from click and keypress handlers as UnderlineNav should support with `href` along with react routing configuration. When react routing is configured, it ignores the `href`
- Take the `4px` left and right padding off from the link items and add `8px` as a `gap` between the list items to align with design specs and take that gap value into account when calculating the possible number of items that can fit in the screen.
- UI fixes to align with design specs.

### Improvements in this PR:
- Add responsive behaviour example to the documentation (Scroll behaviour for the course pointer devices)
- Add React router example to the documentation
- Some docs update in general

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/docs/content/drafts/UnderlineNav2.mdx
+++ b/docs/content/drafts/UnderlineNav2.mdx
@@ -153,16 +153,23 @@ render(<Navigation />)
 ### With React Router
 
 ```jsx
-import {Link} from 'react-router-dom'
+import {Link, useNavigate} from 'react-router-dom'
 import {UnderlineNav} from '@primer/react/drafts'
 
 const Navigation = () => {
+  const navigate = useNavigate()
   return (
     <UnderlineNav aria-label="Repository">
       <UnderlineNav.Item as={Link} to="code" counter={4} selected>
         Code
       </UnderlineNav.Item>
-      <UnderlineNav.Item counter={44} as={Link} to="issues">
+      <UnderlineNav.Item
+        counter={44}
+        as={Link}
+        onSelect={() => {
+          navigate('issues')
+        }}
+      >
         Issues
       </UnderlineNav.Item>
       <UnderlineNav.Item as={Link} to="pulls">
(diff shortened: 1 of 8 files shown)


broccolinisoup:
👋🏼 @danielguillan I have implemented the feedback you have given me [here](https://github.com/primer/react/pull/2406#issuecomment-1273016569). I have a concern here for the touch target size though. I am not sure if it would be semantically correct to make the 44x44 wrapper div clickable to achieve the min touch target while keeping the button in a smaller width and height (because the focus ring should be around the button) if that makes sense. I _think_, the entire target touch area should be a button element. If that is the case, here how they will/should look. Unless I am missing a point here! 

Screen Shot 2022-10-14 at 12 40 49 pm

Rest should be fine. I would appreciate your feedback on the fixes 🙏🏼  Also just to provide a context on giving `8px` gap between items and removing the `4px` padding left and right, I had to slightly change the responsive calculation because  the items got `8px` shorter due to removing the padding and that space went into the gap. it should be working fine and should be aligned with the design specs (I hope at least.) 

Looking forward to hearing your feedback 🙌🏼

broccolinisoup:
Overflow behaviour drastically changed due to the accessibility concerns and it would have been too complex to deal with merge conflicts here. So I am closing this PR and addressed above issues separately in these PRs: 

- https://github.com/primer/react/pull/2447
- https://github.com/primer/react/pull/2448
- https://github.com/primer/react/pull/2449

PR #2428, opened 2022-10-12
Update TreeView based on a11y feedback

joshblack:
This PR addresses the following items from: https://github.com/github/primer/issues/815#issuecomment-1273922579

- [x] Call `event.preventDefault()` on arrow key events
- [x] Fix overlapping hover background and focus ring

_Note: let me know if the refactor for the arrow key behavior isn't as clear as the original, happy to revert back!_


colebemis (approved):
🚀

Issue #2430, opened 2022-10-13
Warning occurs `PageLayout` within Next.js


archlabx:
### Description

If `PageLayout` is used within Next.js, `useLayoutEffect` and `useMedia` warning occurs like https://github.com/primer/react/issues/1453.

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
    at Root (webpack-internal:///./node_modules/@primer/react/lib-esm/PageLayout/PageLayout.js:50:17)
    at Llayout (webpack-internal:///./components/Llayout.jsx:13:20)
    at exports.ThemeProvider (\node_modules\styled-components\dist\styled-components.cjs.js:1:24886)
    at ThemeProvider (webpack-internal:///./node_modules/@primer/react/lib-esm/ThemeProvider.js:35:26)
    at MyApp (webpack-internal:///./pages/_app.js:15:18)
    at StyleRegistry (\node_modules\styled-jsx\dist\index\index.js:448:36)
    at AppContainer (\node_modules\next\dist\server\render.js:298:29)
    at AppContainerWithIsomorphicFiberStructure (\node_modules\next\dist\server\render.js:327:57)  
    at div
    at Body (\node_modules\next\dist\server\render.js:614:21)
`useMedia` When server side rendering, defaultState should be defined to prevent a hydration mismatches.
```

### Steps to reproduce

1. `yarn create next-app --typescript`
2. `yarn add @primer/react styled-components`
3. Make `Layout` with `import {PageLayout} from '@primer/react'` in `Layout.tsx`
4. Wrap `<Component {...pageProps} />` around `ThemeProvider` and `Layout` in `_app.tsx`
5. `next dev` and reload page

### Version

35.11.0

PR #2432, opened 2022-10-13
fix(PageLayout): update useLayoutEffect to isomorphic effect

joshblack:
Closes https://github.com/primer/react/issues/2430

This PR updates `useStickyPaneHeight` to use the `useIsomorphicLayoutEffect` utility instead of `React.useLayoutEffect` to work in SSR.

It also updates `useResponsiveValue` to use the `fallback` value during SSR by setting the default value of `useMedia` to `false`.

Would love feedback here to see if this update matches the intent of `useResponsiveValue` or if I should find a way to expose these fallback values through the component API 👀


colebemis (approved):
Nice!

PR #2434, opened 2022-10-14
feat(TreeView): add label prop for Leading and Trailing Visual

joshblack:
This PR adds in support for describing a `TreeView.Item` with `aria-describedby` and `TreeView.LeadingItem` and `TreeView.TrailingItem`. Each visual slot takes in an optional `label` prop which can be used to describe the overall `TreeView.Item`.

Background: https://github.com/github/primer/blob/main/apis/tree-view-api.md#leading-and-trailing-visuals


src/TreeView/TreeView.tsx
```diff
@@ -429,24 +439,37 @@ function useSubTree(children: React.ReactNode) {
 
 export type TreeViewVisualProps = {
   children: React.ReactNode | ((props: {isExpanded: boolean}) => React.ReactNode)
+  // Provide an accessible name for the visual. This should provide information
+  // about what the visual indicates or represents
+  label?: string
```
colebemis, line 444: Can you add this to the props table for LeadingVisual and TrailingVisual in the docs?

.changeset/pretty-bats-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
```
colebemis, line 2: I've been marking all changes to TreeView as patch changes since it's still a "draft"

```suggestion
'@primer/react': patch
```

src/TreeView/TreeView.tsx
joshblack: Done!

docs/content/TreeView.mdx
```diff
@@ -289,6 +289,11 @@ See [Storybook](https://primer.style/react/storybook?path=/story/components-tree
     type={`| React.ReactNode
 | (props: {isExpanded: boolean}) => React.ReactNode`}
   />
+  <PropsTableRow
+    name="label"
+    type="string"
+    description="Provide an accessible name for the leading visual. This should provide information about what the visual indicates or represents"
```
colebemis, line 295: Should we mention that if a leading or trailing visual is purely decorative that you don't need to provide a label?

colebemis (approved):
Looks great! Just left one question: https://github.com/primer/react/pull/2434#discussion_r996175988

PR #2442, opened 2022-10-17
feat: enable inline for CommonJS bundle

joshblack:
This is a follow-up to: https://github.com/primer/react/pull/2404 after https://github.com/primer/.github/pull/8 was merged. It restores the ability to inline ESM-only modules in the CommonJS path.

The RC workflow should now include the `lib/node_modules` directory when publishing which should remedy the underlying issue when testing out release candidates.

> **Note**
> Dynamic `require()` in `_FormattingTools.tsx`

There is an exception to the inline behavior described above with dynamic require statements. By default the `commonjs` plugin will attempt to hoist these declarations to the top-level when `transformMixedEsModules` is `true`. As a result, these statements are left as-is.

There is one potential path forward with these calls where they are converted to dynamic `import()` statements in source. When they are built for ESM and CommonJS they are then converted to `require()` statements to work around a bug present in Jest and V8.

Another path forward is to determine the way to make these top-level imports so that we can avoid dynamic `require()` and `import()` statements altogether.


joshblack:
@broccolinisoup would it be possible for this to make it into the release this cycle? 🙏

broccolinisoup:
> @broccolinisoup would it be possible for this to make it into the release this cycle? 🙏

Sure! Is there any tests on the release checklist you suggest running again or should our CI checks be enough to confirm?

joshblack:
@broccolinisoup I think that if it works in the memex smoketest we should be good to go!

colebemis (approved):

broccolinisoup:
Confirming that the canary release works in memex 🥳

PR #2444, opened 2022-10-17
Dialog: Remove landmark elements

colebemis:
Remove landmark elements from Dialog based on accessibility feedback from @kendallgassner: https://github.com/primer/react/pull/2352#issuecomment-1261173654

> Will you remove the banner and footer landmark semantics on the dialog component. These landmarks should not be in a dialog since they are not a standalone page.


kendallgassner (approved):
✨ Thank you!

Issue #1956, opened 2022-03-14
Inaccurate type definitions for Button variant prop


rezrah:
**Describe the bug**
Type definitions for `Button` are not being exported correctly in the transpiled output.

E.g:

`variant` is being reported as a string type, rather than the `VariantType` union it should report.

This _may_ be happening with other props, but hasn't been verified.

Due to this incorrect typing, the migration from the v34 Button to v35 becomes challenging as the v34 Button also had a `variant` prop that was used for size, so the TypeScript compiler assumes the size was valid and shows the following error (rather than one related to invalid discriminated unions).

```jsx
<Button
    variant="small"
    onClick={() => onCreate(minimumStartDate, defaultDuration)}
    leadingIcon={PlusIcon}
    {...testIdProps('add-default-iteration-button')}
  >
    Add iteration
  </Button>
```

leads to this error:

`Error: Uncaught [TypeError: Cannot convert undefined or null to object]`

This can also be reproduced on StackBlitz:

**To Reproduce**
1. Go to StackBlitz https://stackblitz.com/edit/primer-react-v35?file=index.tsx
2. Hover over the variant prop on Button
![button props](https://user-images.githubusercontent.com/13340707/158157126-247898ac-e660-479d-a972-a88d0fd48397.png)

**Expected behavior**
variant should show `'default' | 'primary' | 'invisible' | 'danger' | 'outline'` 

**Additional context**
Possibly related? https://github.com/primer/react/issues/1948

lesliecdubs:
👋🏻 @pksjce we're doing some board cleanup and wanted to check in on this one. Is it on your radar / still in progress?

jclem:
This affects all props on `Button` other than the generic indexed properties for `string`, `number` and `symbol`. None of the other props are type-checked or offered up by language server assistance.

mattcosta7:
I believe https://github.com/primer/react/pull/2445 fixes this

PR #2445, opened 2022-10-17
Update `Button` props to ensure they correctly type check

mattcosta7:
Describe your changes here.

Discussed separately in slack

types for Button were being _broadened_ signficantly because the generic interface exposed by the polymorphic component could not be adequately described by the base `forwardRef` implementation.  Using the PolymorphicForwardRefComponent helper, allows the type fidelity to be maintained. 

I've added a type test to validate a few simple cases, and updated places where the types were incorrect in the codebase (now easily noted as failures to build)

One thing to consider - this might be a breaking change for consumers, even though the api _should_ have been this way already

fixes https://github.com/primer/react/issues/1956

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/ActionMenu.tsx
```diff
@@ -83,12 +83,12 @@ const MenuButton = React.forwardRef<AnchoredOverlayProps['anchorRef'], ButtonPro
         <Button
           type="button"
           trailingIcon={TriangleDownIcon}
-          sx={merge(
+          sx={merge<BetterSystemStyleObject>(
             {
               // override the margin on caret for optical alignment
               '[data-component=trailingIcon]': {marginX: -1}
             },
-            sxProp as SxProp
```
mattcosta7, line 91: this was wrong before, but worked because it was becoming any

src/Button/Button.tsx
```diff
@@ -1,16 +1,18 @@
...
-const ButtonComponent = forwardRef<HTMLButtonElement, ButtonProps>(
-  ({children, ...props}, forwardedRef): JSX.Element => {
-    return (
-      <ButtonBase ref={forwardedRef} as="button" {...props}>
-        {children}
-      </ButtonBase>
-    )
-  }
-)
+const ButtonComponent: PolymorphicForwardRefComponent<'button', ButtonProps> = forwardRef<
```
mattcosta7, line 6: the fix was to use this helper type

src/Button/types.ts
```diff
@@ -34,7 +34,7 @@ export type ButtonBaseProps = {
    */
   disabled?: boolean
 } & SxProp &
-  HTMLAttributes<HTMLButtonElement> &
+  React.ButtonHTMLAttributes<HTMLButtonElement> &
```
mattcosta7, line 37: we actually only want button attributes, not all possible attributes (I think)

src/drafts/MarkdownEditor/_ViewSwitch.tsx
```diff
@@ -23,14 +23,14 @@ export const ViewSwitch = ({selectedView, onViewSelect, onLoadPreview, disabled}
   const {label, icon, ...sharedProps} =
     selectedView === 'preview'
       ? {
-          variant: 'invisible',
+          variant: 'invisible' as const,
```
mattcosta7, line 26: VariantType isn't a string, so these need narrowing

src/Button/types.ts
mattcosta7: happy to revert this (and the similar) changes if that makes sense.  I realize this is out of scope in most places, potentially, although also maybe more accurate

src/_TextInputInnerAction.tsx
```diff
@@ -5,7 +5,7 @@ import {Button, IconButton, ButtonProps} from './Button'
 import Tooltip from './Tooltip'
 import {BetterSystemStyleObject, merge, SxProp} from './sx'
 
-type TextInputActionProps = Omit<React.HTMLProps<HTMLButtonElement>, 'aria-label' | 'size'> & {
+type TextInputActionProps = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'aria-label' | 'size'> & {
```
mattcosta7, line 8: this needed to be narrowed to avoid the dom `as` property that `HTMLProps` defines as `as: string` from being improperly considered as a value that gets passed to `Button`.

mattcosta7:
Not sure why canary release seems to fail

src/Button/Button.tsx
jfuchs: It might be nice to audit PRC for files where we aren't using PolymorphicForwardRefComponent (maybe just any file that contains `"forwardRef"` but not `"PolymorphicForwardRefComponent"`?

mattcosta7:
I tested the canary build on memex - https://github.com/github/memex/pull/12803

button now properly catches a bunch of type errors that were introduced in that codebase, without any false positives.  Feeling pretty confident here

src/__tests__/Button.types.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import {Button} from '../Button'
+
+export function shouldAcceptOnlyAChildProp() {
```
jclem, line 4: Could we also test one of the Primer props for accepting valid values and rejecting invalid values? eg `size`

mattcosta7: There's a bunch of tests - https://github.com/primer/react/blob/main/src/__tests__/Button.test.tsx that already validate that some props are accepted, but happy to add additional ones here

mattcosta7: 45d5ec3244a5ac29f53fc8c2cc2a7778857f683d

Added here, which should validate all of the `ButtonProps` an aria/dom prop and the `onClick` handler's types I believe

jclem (approved):

keithamus (approved):
LGTM

joshblack (approved):

colebemis (approved):
🚀

PR #2463, opened 2022-10-20
docs(TreeView): remove navigation from label

joshblack:
Reference: https://github.com/github/primer/issues/702

This PR renames the `File navigation` labels to `Files`.


colebemis (approved):
✨

PR #2465, opened 2022-10-20
TreeView: Add aria-atomic attribute to live region

colebemis:
Adds `aria-atomic="true"` to TreeView live region based on [accessibility feedback](https://github.com/github/primer/pull/1389#discussion_r998750245).


ericwbailey (approved):
Short, and to the point!

joshblack (approved):

PR #2466, opened 2022-10-21
[UnderlineNav2]: Introduce disclosure widget pattern

broccolinisoup:
Following up the accessibility [sign-off review](https://github.com/github/primer/issues/1112) feedback (ref: [comment](https://github.com/github/primer/issues/1112#issuecomment-1281157150)), this PR discarding the `ActionMenu` and introduces a disclosure widget.

Storybook test link : https://primer-0a924a1192-13348165.drafts.github.io/storybook/iframe.html?args=&id=components-underlinenav--internal-responsive-nav&viewMode=story

The reasons for discarding the `ActionMenu` are

1. The `ActionMenu` is a portal which means it doesn't follow the button element (that toggles the menu) in the DOM hierarchy. Considering `aria-controls` attribute's [poor support](https://a11ysupport.io/tech/aria/aria-controls_attribute), we cannot rely on it therefore, the menu has to follow the button in the DOM order. 
2. Keyboard accessibility on the `ActionMenu` is not aligning with the disclosure pattern one. I.e focus trap or only arrow key navigation.

What did we introduce? (Disclosure widget pattern)

- We have a button that toggles a menu.
- Menu immediately follows the button element in the DOM hierarchy.
- Button has `aria-controls` and its value is the id of the menu.
- Button has `aria-expanded` attribute and conditionally have true / false value.
- Navigation items (both list and menu items) are navigable with arrow keys (both horizontal and vertical) and the tab key.
- Menu doesn't trap the focus.
- Menu can be closed with clicking outside and ESC. 

### Screenshots

There shouldn't be any visual differences between the `ActionMenu` version and the disclosure widget version. 

### Merge checklist

- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/UnderlineNav2/UnderlineNav.tsx
+++ b/src/UnderlineNav2/UnderlineNav.tsx
@@ -2,16 +2,21 @@ import React, {useRef, forwardRef, useCallback, useState, MutableRefObject, RefO
 import Box from '../Box'
 import sx, {merge, BetterSystemStyleObject, SxProp} from '../sx'
 import {UnderlineNavContext} from './UnderlineNavContext'
-import {ActionMenu} from '../ActionMenu'
-import {ActionList} from '../ActionList'
 import {useResizeObserver, ResizeObserverEntry} from '../hooks/useResizeObserver'
 import CounterLabel from '../CounterLabel'
 import {useTheme} from '../ThemeProvider'
 import {ChildWidthArray, ResponsiveProps} from './types'
 
-import {moreBtnStyles, getDividerStyle, getNavStyles, ulStyles, menuItemStyles, GAP} from './styles'
+import {moreBtnStyles, getDividerStyle, getNavStyles, ulStyles, getMenuStyles, menuItemStyles, GAP} from './styles'
 import styled from 'styled-components'
 import {LoadingCounter} from './LoadingCounter'
+import {Button} from '../Button'
+import {useFocusZone} from '../hooks/useFocusZone'
+import {FocusKeys} from '@primer/behaviors'
+import {TriangleDownIcon} from '@primer/octicons-react'
+import {useOnEscapePress} from '../hooks/useOnEscapePress'
+import {useOnOutsideClick} from '../hooks/useOnOutsideClick'
+import {ActionList} from '../ActionList'
 
 export type UnderlineNavProps = {
   'aria-label'?: React.AriaAttributes['aria-label']
@@ -129,6 +134,8 @@ export const UnderlineNav = forwardRef(
     const navRef = (forwardedRef ?? backupRef) as MutableRefObject<HTMLElement>
     const listRef = useRef<HTMLUListElement>(null)
     const moreMenuRef = useRef<HTMLLIElement>(null)
+    const moreMenuBtnRef = useRef<HTMLButtonElement>(null)
+    const containerRef = React.useRef<HTMLUListElement>(null)
 
     const {theme} = useTheme()
 
@@ -194,6 +201,7 @@ export const UnderlineNav = forwardRef(
     const afterSelectHandler = (event: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>) => {
       if (!event.defaultPrevented) {
         if (typeof afterSelect === 'function') afterSelect(event)
+        closeOverlay()
       }
     }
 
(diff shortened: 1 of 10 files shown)


ericwbailey:
Carrying [this comment](https://github.com/primer/react/pull/2447#issuecomment-1285375223) over here to continue the conversation:

>We have a button that toggles a menu.
>Menu immediately follows the button element in the DOM hierarchy.

Nice!

>Button has aria-controls and its value is the id of the menu.
>Button has aria-expanded attribute and conditionally have true / false value.

Yep!

>Menu items are navigable with arrow up and arrow down as well as the tab key. (Arrow keys are not implemented in the demo)
>Menu doesn't trap the focus.

Sounds great!

>Menu can be closed with clicking outside and esc. (Not implemented in the demo)
 
I view this as a nice-to-have, but certainly not a blocker functionality-wise.

>I tried to whip something up quickly today, mainly to get something more tangible for us to discuss from. You can see the half baked prototype [here](https://primer-bebeaa10ea-13348165.drafts.github.io/storybook/iframe.html?args=&id=components-underlinenav--internal-responsive-nav&viewMode=story#).

It seems like a lot of the following is already on your radar, based on the previous comment. I'm outlining my observations of the linked prototype, just to have it written down to reference.

For the `button` element used to trigger the overflow menu:

- This might seem counterintuitive, but we'll want to remove `aria-haspopup="true"`, as `true` is aliased to `menu`. `menu` in this context has a very specific meaning, which is suggesting the presence of something along the lines of a right click menu rather than a navigation menu.
- We'll also want to remove the `tabindex="0"` declaration, as the `button` element itself is focusable without it.
- We'll need `aria-expanded="false" initially set, to communicate that it is expandable when focused.

For the menu itself:

- We'll want to remove the declaration of `role="menu` on the `ul` element, for the same reason as `aria-haspopup="true". This is a navigation-style menu, and not a context-style menu.
- For the link items, we'll also want to remove:
    - `tabindex="0"`, as the link element itself is focusable.
    - `aria-labelledby`, as the link's text string supplies its accessible name.
    - `role="menuitemradio"`, as a navigation context uses links and not menu items.
    - `aria-checked="false"`, as navigation items are not toggleable.
    - `aria-keyshortcuts`, as we'll not want keyboard shortcut commands for navigation items.

This feels to me like a lot of it is getting inherited from ActionMenu, so hopefully decoupling from that component will address things.

broccolinisoup:
@ericwbailey I apologise again for pointing out the wrong demo. [Here](https://primer-0a924a1192-13348165.drafts.github.io/storybook/iframe.html?args=&id=components-underlinenav--internal-responsive-nav&viewMode=story) is the correct one. With some update, these are the key points that will come with the disclosure pattern:

- We have a button that toggles a menu.
- Menu immediately follows the button element in the DOM hierarchy.
- Button has aria-controls and its value is the id of the menu.
- Button has aria-expanded attribute and conditionally have true / false value.
- Navigation items (both list and menu items) are navigable with arrow horizontal, arrow vertical keys and the tab key.
- Menu doesn't trap the focus.
- Menu can be closed with clicking outside and esc. (Currently focus doesn't return the menu but I think it should right?)

Thank you so much again 🙏🏼

src/UnderlineNav2/styles.ts
```diff
@@ -142,3 +145,17 @@ export const menuItemStyles = {
...
   textDecoration: 'none'
 }
+
+export const getMenuStyles = (theme?: Theme, isWidgetOpen?: boolean) => ({
+  position: 'absolute',
+  top: '90%',
+  right: '0',
+  boxShadow: '0 1px 3px rgba(0, 0, 0, 0.12), 0 1px 2px rgba(0, 0, 0, 0.24)',
+  borderRadius: '12px',
+  backgroundColor: `${theme?.colors.canvas.overlay}`,
```
siddharthkp, line 155: We probably don't need to access this from theme and can replace with the token name

```suggestion
backgroundColor: 'canvas.overlay'
```

src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -265,46 +313,53 @@ export const UnderlineNav = forwardRef(
...
+                  onClick={onAnchorClick}
+                  trailingIcon={TriangleDownIcon}
+                >
+                  More
+                </Button>
+                <ActionList
+                  selectionVariant="single"
+                  ref={containerRef}
+                  id="disclosure-widget"
+                  sx={getMenuStyles(theme, isWidgetOpen)}
```
siddharthkp, line 330: Sorry perf on my mind 😅 

I think we can make this a static object instead of using a function call


```suggestion
                  sx={menuStyles}
                  style={{ display: isWidgetOpen ? 'block' : 'none' }}
```

ericwbailey (approved):
I left a comment about utilizing a UUID for the `id`, but other than that this sounds great to me in AT. Loving the roving tabindex, too!

src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -235,6 +243,46 @@ export const UnderlineNav = forwardRef(
...
+      [isWidgetOpen]
+    )
+
+    useOnOutsideClick({onClickOutside: closeOverlay, containerRef, ignoreClickRefs: [moreMenuBtnRef]})
+    const onAnchorClick = useCallback(
+      (event: React.MouseEvent<HTMLButtonElement>) => {
+        if (event.defaultPrevented || event.button !== 0) {
+          return
+        }
+        toggleOverlay()
```
joshblack, line 282: This seems to be the only place where `toggleOverlay` is called, would it make sense to inline it from L:252 here?

```suggestion
        setIsWidgetOpen(isWidgetOpen => !isWidgetOpen)
```

joshblack (approved):
Looks awesome! Just left a minor comment and wanted to +1 the `id` suggestion from @ericwbailey on the `<ul>` for the disclosure.

Awesome work ✨

src/UnderlineNav2/styles.ts
broccolinisoup: Oh I didn't know I can just pass the token name! Thank you 🙌🏼  Is it only for colours as we import only colour primitives currently maybe? I see mix practises around, where can I learn more about this? 👀 or should I just dig in? 😅

src/UnderlineNav2/UnderlineNav.tsx
broccolinisoup: Oh of course! Keep them coming 😄  I'll push the commit from my local because I need to modify the styles.ts as well to make that function a static obj

src/UnderlineNav2/UnderlineNav.tsx
broccolinisoup: Yes, it does! 🙂 Thank you 🙌🏼 I'll do add the commit on my local as I also need to remove the toggleOverlay function ✨

src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -265,46 +313,53 @@ export const UnderlineNav = forwardRef(
...
+                  aria-expanded={isWidgetOpen}
+                  onClick={onAnchorClick}
+                  trailingIcon={TriangleDownIcon}
+                >
+                  More
+                </Button>
+                <ActionList
+                  selectionVariant="single"
+                  ref={containerRef}
+                  id="disclosure-widget"
```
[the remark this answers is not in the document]
broccolinisoup: Ahh great point, thanks! Pushed a commit ✨

src/UnderlineNav2/styles.ts
broccolinisoup: I'll do a clean up to remove the other unnecessary `theme` args in the next PR!

colebemis (approved):
Latest commit looks good 👍 Ship it 🚢

PR #2470, opened 2022-10-23
[UnderlineNav2]: Add visually hidden heading where `aria-label` is present

broccolinisoup:
Following up the accessibility [sign-off review](https://github.com/github/primer/issues/1112) feedback (ref: [comment](https://github.com/primer/react/pull/2447#issuecomment-1282755318)), this PR adds a visually hidden heading where aria-label is present

Story book link: https://primer-a937bd5bd7-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav--default-nav

### Screenshots

A screenshot of the DOM that shows a visually hidden h2 heading as a sibling to the nav landmark

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/twelve-spoons-walk.md
+++ b/.changeset/twelve-spoons-walk.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+UnderlineNav2: Render a visually hidden heading for screen readers when aria-label is present
--- a/src/UnderlineNav2/UnderlineNav.test.tsx
+++ b/src/UnderlineNav2/UnderlineNav.test.tsx
@@ -112,4 +112,10 @@ describe('UnderlineNav', () => {
     expect(loadingCounter?.className).toContain('LoadingCounter')
     expect(loadingCounter?.textContent).toBe('')
   })
+  it('renders a visually hidden h2 heading for screen readers when aria-label is present', () => {
+    const {container} = render(<ResponsiveUnderlineNav />)
+    const heading = container.getElementsByTagName('h2')[0]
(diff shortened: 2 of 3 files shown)


ericwbailey (changes requested):
One small thing. Could we append ` navigation` to the heading's string? 

The reason is a `nav` element will be announced as a navigation landmark, but the heading does not indicate it is for navigation without the appended string suppling the context.

joshblack (approved):
Changes look great on the implementation side ✨ 

Is there a good way to learn more about the technique itself? Was surprised that this would work for visually hidden content but I guess it makes sense if the user can find it and then is expected to hit tab to go to following content

ericwbailey:
@joshblack This came up in conversation with James (not tagging him here to keep his notifications down). His point was you want to have a heading present to provide quick context to people navigating via heading, which is a very common navigation technique. It's more a quality of life improvement than it is an accessibility issue, but I think a good one!

broccolinisoup:
> One small thing. Could we append ` navigation` to the heading's string?
> 
> The reason is a `nav` element will be announced as a navigation landmark, but the heading does not indicate it is for navigation without the appended string suppling the context.

Ahh makes sense! Just pushed a commit.
A screenshot of the DOM that shows a visually hidden h2 heading as a sibling to the nav landmark

ericwbailey (approved):
Woohoo! 🚀

PR #2478, opened 2022-10-24
feat(TreeView): add support for empty parent nodes

joshblack:
This PR updates `TreeView.Item` and `TreeView.SubTree` based on the feedback over in: https://github.com/github/primer/pull/1389#discussion_r998754809

### DIFF
--- a/src/TreeView/TreeView.test.tsx
+++ b/src/TreeView/TreeView.test.tsx
@@ -1,7 +1,10 @@
 import {fireEvent, render, waitFor} from '@testing-library/react'
 import React from 'react'
+import {act} from 'react-dom/test-utils'
 import {ThemeProvider} from '../ThemeProvider'
 import {SubTreeState, TreeView} from './TreeView'
+
+jest.useFakeTimers()
 
 // TODO: Move this function into a shared location
 function renderWithTheme(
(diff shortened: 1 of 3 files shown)


src/TreeView/TreeView.test.tsx
```diff
@@ -983,13 +990,15 @@ describe('Asyncronous loading', () => {
...
+    act(() => {
+      jest.runAllTimers()
     })
+
+    // First child should be focused
+    expect(firstChild).toHaveFocus()
   })
 
-  it.only('moves focus to parent item after closing error dialog', async () => {
+  it.skip('moves focus to parent item after closing error dialog', async () => {
```
colebemis, line 1001: Is this test not working? Curious why we are skipping it.

colebemis (commented):
Looks great! Could you add a couple of tests to make sure the `aria-expanded` attribute is behaving as expected in the sync and async cases?

src/TreeView/TreeView.test.tsx
joshblack: @colebemis it seems like something is going on with it in `main` 🤔 

When using `it.only` it passes, when removing `it.only` the test ends up failing. Not sure what's going on with it.

joshblack: I resolved it in the branch, I think it should be okay now!

PR #2480, opened 2022-10-25
ci: split ci workflow into distinct jobs

joshblack:
**Background**

The current CI workflow and CI job is made up of several stages, specifically: linting, testing, type checking, and building examples. This job is a required status check in order to be merged.

This PR breaks up these stages into discrete jobs to reduce the total time spent in CI. It also isolates the stages in to jobs which can independently report status for contributors to better understand any issues present in CI.

<details>
<summary>Job time information</summary>

| Workflow | Job | Time | 
| :------- | :-- | :--- |
| CI       | CI  | 7m   |
| CI       | lint | 2m  |
| CI       | test | 6m  |
| CI       | type-check | 2m  |
| CI       | examples | 2m  |

</details>

#### Changelog

**New**

**Changed**

- Update `ci.yml` workflow to use distinct jobs instead of one `ci` job, specifically: `lint`, `test`, `type-check`, `examples`
- Update `ci.yml` to include `concurrency` key to cancel any existing jobs when a new commit is pushed

**Removed**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,12 @@ on:
       - main
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  ci:
-    name: CI
+  lint:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


broccolinisoup (commented):
Super cool 🔥 Just a question, both `test` and `examples` require build so compared to previous steps, we build twice now if I understand correctly. How much does that effect the total CI time, like something that is noticeable or not?

joshblack:
Hey @broccolinisoup! 👋 

> How much does that effect the total CI time, like something that is noticeable or not?

Thankfully each job is run in parallel so there shouldn't be any impact. The overall time it takes to finish the CI run should be essentially the longest individual job, as a result (which I think is test)

broccolinisoup (approved):
🥳

PR #2484, opened 2022-10-26
Create RelativeTime component

keithamus:
This adds the `RelativeTime` component, which uses the [`<relative-time>` web component](https://github.com/github/time-elements) as a backing component.

Tracking issues: https://github.com/github/primer/issues/1258, https://github.com/github/primer/issues/1347, https://github.com/github/primer/issues/1252


package-lock.json
```diff
@@ -145,6 +147,22 @@
         "styled-components": "4.x || 5.x"
       }
     },
+    "../time-elements": {
```
siddharthkp, line 150: 👀

src/RelativeTime/RelativeTime.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+import styled from 'styled-components'
+import {RelativeTimeElement} from '@github/time-elements'
+import {createComponent} from '@lit-labs/react'
+import sx from '../sx'
+
+export const RelativeTime = styled(createComponent(React, 'relative-time', RelativeTimeElement))(sx)
```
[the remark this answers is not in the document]
josepmartins: 😮 🔥

src/RelativeTime/RelativeTime.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import styled from 'styled-components'
+import {RelativeTimeElement} from '@github/time-elements'
+import {createComponent} from '@lit-labs/react'
+import {ComponentProps} from '../utils/types'
+import sx from '../sx'
+
+const RelativeTime = styled(createComponent(React, 'relative-time', RelativeTimeElement))(sx)
```
mattcosta7, line 8: types by default here don't seem to be correct (at least when imported into another project)

src/RelativeTime/RelativeTime.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import styled from 'styled-components'
+import {RelativeTimeElement} from '@github/time-elements'
```
mattcosta7, line 3: given this import now calls 

window.customElements.define in module scope, what type of story are we providing for users who use this in react, but also use server side rendering?

src/RelativeTime/RelativeTime.tsx
mattcosta7: or, at least they don't support `Date` objects, which may be expected?

mattcosta7: https://github.com/github/memex/pull/13207#discussion_r1017906919

keithamus: Types are correct. I followed up in that thread:

> Thanks for the catch! The way the properties work, datetime can only ever be a string, but date can be set to a Date (overriding datetime).

src/RelativeTime/RelativeTime.tsx
keithamus: I think we can stub out `HTMLElement` and `window.customElement.[get/has/define]` within an SSR context fairly simply, which should make this a painless lift. I don't foresee any major consequences to that, apart from perhaps pushback from folks wanting that in the SSR runtime codebase. If there is sufficient pushback then we can resort to having a browser only bundle in Primer React, but I think that should be a last resort.

mattcosta7: For non-github consumers, that stub pattern will likely be un-ideal (and even for us - I'd be hesitant to stubbing those). 

a client bundle makes sense, although I don't love that complexity add either (especially given the tradeoffs on how that bundle codesplits (one big import with things people may never use, where lazy loading may not be ideal either for those wanting single bundles vs many small imports needing to be managed separately)

but glad we have some ideas in the works for it

keithamus: Running multi-modal JS (or "isomorphic JS" or "Universal JS") faces these problems writ large, and is something we'll need to grapple with in the broader context, not just for those two globals but others too. An engineer can trivially reach for globals like `ResizeObserver`, `EventTarget`, `performance.now`, `URL`, `fetch`, `console.log`, `crypto.subtle`. While all work in a browser most won't in NodeJS. Deno, however, provides all except `ResizeObserver`. I think this illustrates an interesting juncture where some philosophies are fixed towards making application code renderer-agnostic, but other philosophies are looking towards making run-times more similar to browsers. Node is gradually leaning more towards providing these globals too - it recently added `EventTarget`.

On the other hand, yes, client bundles make more and more sense if we consider that not every element on the page needs a client side component definition. As we start to move away from that idea within React, a natural conclusion may be what other frameworks like Rails have done: render components server side, and deliver a handful of interactive ones as dedicated client components. This moves away from multi-modal JS ("Isomorphic JS"/"Universal JS") as it demarcates that some components are designed for consumption within the server and some are not.

Both choices take us down one path, and come with varying trade-offs.

src/__tests__/RelativeTime.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+describe('RelativeTime', () => {
+  behavesAsComponent({Component: RelativeTime})
+
+  checkExports('RelativeTime', {
+    default: RelativeTime
+  })
+
+  it('should have no axe violations', async () => {
+    const {container} = HTMLRender(<RelativeTime />)
```
joshblack, line 16: What would be the best example for `relative-time` to check against with axe? Would it be something like this from the examples, potentially? 

```tsx
  it('should have no axe violations', async () => {
    const {container} = HTMLRender(<RelativeTime datetime="2014-04-01T16:30:00-08:00">April 1, 2014</RelativeTime>)
    const results = await axe(container)
    expect(results).toHaveNoViolations()
  })
```

jest.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/src/utils/test-matchers.tsx', '<rootDir>/src/utils/test-deprecations.tsx'],
   testMatch: ['<rootDir>/(src|codemods)/**/*.test.[jt]s?(x)', '!**/*.types.test.[jt]s?(x)'],
   transformIgnorePatterns: [
-    'node_modules/(?!@github/combobox-nav|@koddsson/textarea-caret|@github/markdown-toolbar-element)'
+    'node_modules/(?!@github/combobox-nav|@koddsson/textarea-caret|@github/[a-z-]+-element|@lit-labs/react)'
```
joshblack, line 10: Could we also update the Rollup config for the packages that are ESM only? This should help in the CommonJS case so people won't have to update their Jest config.

joshblack (commented):
Super cool! Just left a couple of comments/questions 👀

src/RelativeTime/RelativeTime.tsx
joshblack: I think React is definitely trying to have dedicated client and server components based on the RFC and Next.js work, but I think that client components will still require the ability to "pre-render" on the server.

This seems like something we should aim to support in order to make sure a component renders something usable when delivered to the client and can be enhanced as JavaScript loads and the component is hydrated. One thing I would like to avoid are components that noop on the server and then have to wait for hydration on the client in order to render something onto the page.

Curious what you all think about this 🤔

In the interim, for things like `HTMLElement` would a shim like what they do in lit feel appropriate? ([Link](https://github.com/lit/lit/blob/a6a5591327a90e9c7d922d1d978627794909c422/packages/reactive-element/src/reactive-element.ts#L395-L398)) Then for things like `customElements.define` would it be feasible to check to see if the runtime supports it before calling it?

keithamus:
@siddharthkp `@github/time-elements` is a bit bigger than it needs to be because it has multiple elements, and also needs to internally polyfill things like `Intl.DateTimeFormat`. A lot of that will change in a breaking release but work needs to be done in the monolith before the breaking release can be made. 

Also bear in mind for the monolith, it already loads the element and so this won't have a material impact to the monolith codebase.

mattcosta7:
> Also bear in mind for the monolith, it already loads the element and so this won't have a material impact to the monolith codebase.

https://github.com/primer/react/pull/2484/files/d3c368705bc81b0255c07b892541ce5c6e1663d6..1f1ce07895a1fb13611136bdbb0f5362d726c8cd#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R85

@keithamus do we need to open up the range of this dependency to account for the fact that it might diverge from the version in dotcom over time, to avoid both versions being bundled, potentially? I don't know how well they'll de-dupe if primer/react requires a specific version, but dotcom is on a higher version?

docs/content/RelativeTime.mdx
```diff
@@ -0,0 +1,141 @@
...
+    name="precision"
+    type="'year' | 'month' | 'day' | 'hour' | 'minute' | 'second'"
+    defaultValue="'second'"
+    description="The precision to use when displaying an elapsed time"
+  />
+  <PropsTableRow
+    name="threshold"
+    type="string"
+    defaultValue="'P30D'"
+    description="The threshold (ISO8601 duration) to display relative dates within"
```
siddharthkp, line 65: Looking at the [design guidelines for relative time](https://primer-437c43db97-26441320.drafts.github.io/components/relative-time), are there options here that we should bake into the component instead of making it configurable?

keithamus: I believe RelativeTime _is_ opinionated; it has a good set of sensible defaults and there should be very little need to customise it. However date-times are a sensitive topic that are very dependent on context (I've made a PR https://github.com/primer/design/pull/331 which goes into some explanation of the principles here).

`threshold` is configurable because sometimes we want a no threshold (i.e. `PT0S`) and sometimes a much shorter threshold (e.g. `P1D`). The default of `P30D` works in 99% of use cases. From the guidelines PR:

> Some dates are especially sensitive and always should be displayed in a precise date-time format, such as the creation or expiration of certificates and keys. When a user needs to take action before a particular time (such as an expiry date) then the precise date time should be displayed

`precision` is also configurable to allow us to elide unimportant information in certain contexts. You care how many seconds your GitHub Action is taking, less so about how long it's been since a PAT was last used.

The `format` prop supporting custom formatting is definitely an escape hatch, and it is something I'd like to deprecate (bearing in mind this component has existed for a long time)! We _do_ want to move to supporting only recommended options which is why `second`, `weekday`, etc all exist. They will allow for configuration of dates and times in a way which allows flexibility while ensuring good practices are followed.

colebemis (approved):
Looks great 🚀 

Thanks for walking me through the API decisions, @keithamus! I think all the props make sense, given the complexity of this component.

josepmartins:
I only have one concern regarding the `micro` being an option in the format list. I believe `micro` should be tied to responsive behavior, as mocked in the initial [API specs](https://github.com/github/primer/blob/main/apis/time-ago-api.md#relativetime), as an option to style any variant when there isn't enough horizontal space. 

For instance, if there's enough space in its container, the `format="elapsed"`  could use the normal `X year X month...` style by default, and only `Xm Xy ...` when on mobile or tablet viewport sizes.

As a side note, and from the accessibility review, using `micro` requires extra markup for screen readers, and doesn't allow native browser translation.

keithamus:
> I have one concern regarding the `micro` being an option in the format list. I believe `micro` should be tied to responsive behavior, as mocked in the initial [API specs](https://github.com/github/primer/blob/main/apis/relative-time.md#micro), as an option for [standalone](https://primer.style/design/components/relative-time#responsive-design) use-cases, when used in compact spaces or in a mobile environment.

Acknowledged. Let's explore some options here in a follow up, as I'd like to get this shipped today and in front of engineers, and we can extend to adding support for this later (especially as there's no active areas where responsive format changes are used).

> As a side note, and from the accessibility review, using `micro` requires extra markup for screen readers, and doesn't allow native browser translation.

Acknowledge this too. I'll follow up in the web component itself to make sure we're accessible here, as it is best placed to resolve this.

PR #2491, opened 2022-10-27
test(project): add vrt and avt tests through playwright

joshblack:
**Overview**

This PR is an exploration into using Playwright for VRT (Visual Regression Testing) and AVT (Accessibility Verification Testing).

Tests are currently located in the `e2e` directory and were generated from Storybook stories. Each test has roughly the following structure:

- Each component is a test suite containing a number of stories
  - For each story
    - Snapshot the story in each theme
    - Run axe against the story

Currently, the snapshots only run against Chromium with the default browser width. There is the ability to extend snapshots per-browser, if needed, and per-breakpoint.

**Developer Experience**

A developer can run tests with Playwright using: `npx playwright test`. They can also filter tests by test name: `npx playwright test ActionList` or by groups that are listed in the test name: `npx playwright test --grep vrt`

Reports of a test run are included both in the terminal and in an HTML report that can be opened and interacted with.

<details>
<summary>Terminal</summary>

Screen Shot 2022-10-27 at 12 31 57 PM

</details>

<details>
<summary>HTML</summary>

Screen Shot 2022-10-27 at 12 32 09 PM

Screen Shot 2022-10-27 at 12 32 19 PM
</details>

A screenshot can be made by using `expect(await page.screenshot()).toMatchSnapshot()`. A page can be checked for axe violations by writing `await expect(page).toHaveNoViolations()`

In addition, developers may use the VSCode extension to interact with Playwright: https://playwright.dev/docs/getting-started-vscode

**CI**

This experiment adds in two separate jobs: `vrt` and `avt`. Each job uses Playwright's `--grep` option to only test for the specific tests that have that tag.

In order to speed up VRT tests, we make use of a `vrt-runner` job that uses a matrix strategy to create multiple runners that each run a portion of our tests. These runners rollup into the `vrt` job described above which will report the status of each runner

**Drawbacks**

- This PR does not include screenshots of interactive states for elements
- Playwright tests can take a while to run locally (around 8min on my machine)
- Playwright tests can be flaky and require re-runs unexpectedly
- Playwright does not seem to remove unused snapshots

**Questions**

- This draft captures almost _everything_ in storybook, what stories would we prioritize snapshotting?
- Is this preferable to using/paying for Chromatic?
- What kind of interactions would we like to capture in VRT?
- Should tests be auto-generated, manually curated, or some combination of both?

**Notes**

- Test files were generated using this script: https://gist.github.com/joshblack/416f29592f9ea6c783bb93a232ca7038, we could keep using this script in the project or specify that tests are added to e2e as part of the testing workflow
- Files don't have to live in `e2e`, instead we could have a convention like `ComponentName.visual.test.ts` that lives beside the component (and is ignored by Jest)


jonrohan:
Nice 👍🏻 Maybe we could work together to set this up for view_components next time we pair?

rezrah:
> This PR does not include screenshots of interactive states for elements

Have you thought about using Storybook interactions for this? [Example of interaction](https://primer.style/brand/storybook/?path=/story/components-subdomainnavbar--search-results-visible), [Example of snaphot](https://github.com/primer/brand/blob/main/packages/react/src/SubdomainNavBar/SubdomainNavBar.visual.spec.ts-snapshots/Visual-Comparison-SubdomainNavBar-SubdomainNavBar-Search-Results-Visible-1-linux.png).

> Playwright tests can take a while to run locally (around 8min on my machine)

Have you thought about running these in an Actions runner instead? Generating the snapshots locally-only may limit future contributions and snapshot regeneration to folks in darwin environment (or worse, create duplicates). 

Maybe consider doing it in a containerised environment, which you can run locally or on the server. That'll generate filenames with the `*-linux.png` suffix instead of `*-darwin.png`. I have this working in Brand so let me know if you want a walkthrough of it.

> Should tests be auto-generated, manually curated, or some combination of both?

I recommend self-generating spec files with blanket coverage so you can just run and forget about them. Where you need granular control over some part of the config, like providing timeouts, different viewports, etc.. you can do that in the generation script. Here's [a basic example](https://github.com/primer/brand/blob/9760c5844ad1646d3026a8d77f923d6375f9d3a0/packages/e2e/scripts/playwright/playwright.generate-tests.ts#L35), but can be done in many other ways too.

langermank:
> This draft captures almost everything in storybook, what stories would we prioritize snapshotting?

I would prioritize snapshotting everything within a component's `/features` directory! And `/interactions` if it exists for the component. Working on re-formatting all of our stories will make this much easier to accomplish 😄 

> Is this preferable to using/paying for Chromatic?

I like that everything lives "in house" and can be seen in GitHub (like the actual screenshots). As someone who doesn't contribute to PRC regularly, I don't really know how to access or use Chromatic.

> Should tests be auto-generated, manually curated, or some combination of both?

I think auto-generated based on that features directory.. it might be interesting if a story author could opt in a specific story, thinking maybe a more detailed one from the `/examples` directory.

joshblack:
Thanks everyone for the feedback! 🙏 

I took the suggestions from this PR and created a smaller, more focused one for `Button` over here: https://github.com/primer/react/pull/2520

PR #2492, opened 2022-10-27
fix(TreeView): add hcm styles for aria-current in TreeView

joshblack:
This addresses the HCM feedback from: https://github.com/github/primer/issues/1114#issuecomment-1292865037

TreeView uses the `HighlightText` keyword for `aria-current` and focus styles

https://user-images.githubusercontent.com/3901764/199105960-865882e2-89dd-4e60-b1ba-9698725bb1a2.mov

### DIFF
--- a/.changeset/lemon-rabbits-move.md
+++ b/.changeset/lemon-rabbits-move.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Add support for aria-current styles in high contrast mode for TreeView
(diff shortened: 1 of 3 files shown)


joshblack:
@ericwbailey would love your feedback on what the best values would be for HCM here. Mainly have just been using keywords from: https://www.w3.org/TR/css-color-4/#css-system-colors or transparent outlines.

ericwbailey (approved):
Heck yeah!

PR #2502, opened 2022-10-31
fix(PageLayout): add maxWidth to container

joshblack:
This PR updates the container used to wrap `PageLayout.Pane` and `PageLayout.Content` to match their previous parent element which set `max-width: 100%`.

Without this property, the `PageLayout.Content` pane would extend passed the edge of the page and trigger overflow. This lead to issues as described [here](https://github.slack.com/archives/C01L618AEP9/p1666969818414279) where the `position: sticky` was triggered due to the overflow.

To reproduce this issue, you can visit the reivew lab from that Slack thread and reduce the size of the browser to where the symbol pane overlaps the code area. Adding in `max-width: 100%` to the container of the `main` and Pane addresses the issue.

### DIFF
--- a/.changeset/angry-wombats-burn.md
+++ b/.changeset/angry-wombats-burn.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Add maxWidth to container of PageLayout.{Pane, Content}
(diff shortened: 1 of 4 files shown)


jdrush89 (approved):

colebemis (approved):
🚢

PR #2504, opened 2022-10-31
feat(TreeView): add support for Backspace to collapse a folder

joshblack:
Based on feedback [here](https://github.com/github/primer/issues/1114#issuecomment-1292865037), this PR adds in support for `Backspace` to collapse a folder.

**Testing & Reviewing**

You can verify that `Backspace` works by visiting a story for TreeView and expanding a folder. Pressing `Backspace` on the folder should **not** collapse it. If a parent item is available, focus should move to it. If the element is at the root, focus should not move.

https://user-images.githubusercontent.com/3901764/199109615-89e29b11-8c7a-4cd3-b24f-6244dc42ba3f.mov


colebemis:
> When inside of a folder, pressing Backspace on an item should noop.

I think pressing `Backspace` on an item should focus the parent folder. That's how I interpreted @jscholes' description here: https://github.com/github/primer/issues/1114#issuecomment-1292397344

> So, in generic terms, Backspace should: move focus to the parent node of the currently focused node, without changing the expanded state of any nodes. If the currently focused node is a root node, it should do nothing, even if that root node is an expanded parent.

jscholes:
Correct, @colebemis.  @joshblack Backspace shouldn't collapse any nodes.  It should move focus to the parent of the focused node, or do nothing if the focused node is a root node.  But the state of nodes shouldn't change.

joshblack:
Totally makes sense! Making those changes now 👍

src/TreeView/useRovingTabIndex.ts
```diff
@@ -30,20 +30,26 @@ export function getNextFocusableElement(activeElement: HTMLElement, event: Keybo
...
 
     case 'end ArrowRight':
       // Do nothing
       return
 
+    // Focus parent element
     case 'end ArrowLeft':
-      // Focus parent element
+    case 'end Backspace':
       return getParentElement(activeElement)
```
colebemis, line 53: You could also add `Backspace` to the switch statement below this on line 50 since the behavior of `Backspace` doesn't depend on the state of the node

package.json
```diff
@@ -82,7 +82,7 @@
     "@github/combobox-nav": "^2.1.5",
     "@github/markdown-toolbar-element": "^2.1.0",
     "@github/paste-markdown": "^1.4.0",
-    "@primer/behaviors": "1.3.0",
+    "@primer/behaviors": "1.3.1-rc.8de3bf8",
```
colebemis, line 85: Just published 1.3.1 for @primer/behaviors: https://github.com/primer/behaviors/releases/tag/v1.3.1

joshblack: Thanks @colebemis!

colebemis:
Wanted to call out this note from @jscholes:

> Critical note: Backspace on Windows and Delete on macOS are not the same key, even though they occupy the same physical position on most keyboards. This functionality must only apply to Backspace, and not Delete.

In the storybook, the `Delete` key on my mac keyboard is triggering the backspace behavior. I'm not sure how we can programmatically differentiate Backspace on Windows and Delete on macOS. Any ideas @jscholes @joshblack?

jscholes:
@colebemis / @joshblack Does this only apply on Mac?  I.e. does `['Backspace', 'ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'Enter'].includes(event.key)` resolve to `true` when pressing Delete specifically on Windows?  If not, I think we should be fine; WebKit may just treat them as the same.  Not sure what we should do down the line if we actually want to support node deletion though...

jscholes:
@joshblack I think Fn+Backspace on Mac is Delete specifically.  Backspace on its own is Backspace, so the current state should be good.

.changeset/fast-shoes-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Add support for Backspace to collapse group in TreeView
```
colebemis, line 5:
```suggestion
Add support for Backspace to move focus to parent TreeView item
```

colebemis (approved):

PR #2506, opened 2022-11-01
[UnderlineNav2]: Increase the unit test coverage

broccolinisoup:
Adding unit tests to UnderlineNav

### DIFF
--- a/src/UnderlineNav2/UnderlineNav.test.tsx
+++ b/src/UnderlineNav2/UnderlineNav.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import '@testing-library/jest-dom/extend-expect'
-import {fireEvent, render} from '@testing-library/react'
+import {render} from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 import {
   IconProps,
   CodeIcon,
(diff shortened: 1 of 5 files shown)


src/UnderlineNav2/UnderlineNav.test.tsx
```diff
@@ -119,3 +146,55 @@ describe('UnderlineNav', () => {
...
+    // Press up arrow
+    fireEvent.keyDown(nextItem, arrowUp)
+    // focus should be on the previous item
+    expect(item).toHaveFocus()
+    expect(nextItem.getAttribute('tabindex')).toBe('-1')
+    expect(item.getAttribute('tabindex')).toBe('0')
+  })
+  it('should move focus to the next/previous item on the list with the tab key', () => {
+    const {getByText} = render(<ResponsiveUnderlineNav />)
+    const item = getByText('Code').closest('a') as HTMLAnchorElement
```
joshblack, line 188: When do you think it would be better to do `getByText('Code').closest('a')` versus `getByRole('link', { name: 'Code' })`

src/UnderlineNav2/UnderlineNav.test.tsx
```diff
@@ -119,3 +146,55 @@ describe('UnderlineNav', () => {
...
+    const nextItem = getByText('Issues').closest('a') as HTMLAnchorElement
+    // Focus first item
+    item.focus()
+    // Press down arrow
+    fireEvent.keyDown(item, tab)
```
joshblack, line 193: When do you think would be best to use `fireEvent` versus `user.tab()` from user events? e.g. https://testing-library.com/docs/user-event/convenience/#tab

joshblack (approved):
Looks good! Just had a couple questions around conventions, was curious what your thoughts were 👀

src/UnderlineNav2/UnderlineNav.test.tsx
broccolinisoup: I didn't really know the differences! 😿 Thanks for pointing this out! I got the habit of using fireEvent but I just read about both now and using `userEvents` makes better sense here because we want to simulate the exact behaviour of user pressing the `tab` key rather than just firing the actual keydown event. TIL that userEvents might include some other chain of events before the actual keydown fireEvent. 

To answer your question, using `fireEvent` might be a better use when probably we want to test the actually event in an isolated way or if we need a different set of chain events before firing a keydown event. For example if we want to test tab key firing without focusing it maybe

src/UnderlineNav2/UnderlineNav.test.tsx
broccolinisoup: Hmm 🤔  I would use them interchangeable but probably not the best? I can't think of any case right now where using the element.closest('a') will be the best.  Do you have an example? That said, `getByRole` is more declarative so I'll opt in using this, thank for pointing this out :)

broccolinisoup:
@joshblack Thanks so much for your review, great points! I pushed another commit, I'd appreciate if you could have another look. Thanks!

PR #2358, opened 2022-09-21
[Proposal] Organize Storybook folder structure

langermank:
This PR accompanies this [Discussion](https://github.com/github/primer/discussions/1305) (hubbers only) about Storybook/Lookbook structure. This is just a start, but with our [prospects of single page component docs](https://github.com/github/primer/issues/1285) Storybook becomes even more critical for documentation. If Storybook is organized closer to our more "official" documentation, it may be easier to navigate and parse. 

### Screenshots

See the preview deploy 😄 

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

PR #2507, opened 2022-11-01
update storybook links due to format change

broccolinisoup:
Due to the format changes on storybook ([PR](https://github.com/primer/react/pull/2358)), the links on primer.style/react should be updated. 

Reviewers can look at the deployed environment and testing some of the links randomly should be enough. I tested all links, hopefully I didn't miss anything.


langermank (approved):
Sorry about this ☹️ we might want to pull these links from Storybook's `stories.json` file instead.. but for now I'll make sure to update these docs as I work on restructuring stories!

docs/content/ActionList.mdx
```diff
@@ -3,7 +3,7 @@ componentId: action_list
 title: ActionList
 status: Beta
 source: https://github.com/primer/react/tree/main/src/ActionList
-storybook: '/react/storybook?path=/story/composite-components-actionlist'
+storybook: 'https://primer.style/react/storybook?path=/story/components-actionlist'
```
joshblack, line 6: Real quick, just wanted to ask about the switch to including `https://primer.style` versus `/` and what the pros/cons of each was 👀

broccolinisoup: I am glad you pointed out @joshblack because actually this is a mistake, now I realised 🙈 

My intention switching to `https://primer.style` was because`/` doesn't direct us to the correct link on the dev environment. See [this](https://primer-cde231ddd0-13348165.drafts.github.io/ActionList) for example and click on [Storybook](https://primer-cde231ddd0-13348165.drafts.github.io/react/storybook?path=/story/composite-components-actionlist. I have seen a couple of components using `https://primer.style` and I jumped straight into it. And it was late last night when I was testing this, so totally missed the elephant here 😅 Apparently, this fixes broken links but redirects to `primer/style` not the dev env the branch is running on! I'll remove the `https://primer.style` now. Also I thought about removing `react` from the url and this seems to fix this problem but I am not sure if I miss anything here. Like why did we have `react` in the first place in the url..

broccolinisoup:
> Sorry about this ☹️ we might want to pull these links from Storybook's `stories.json` file instead.. but for now I'll make sure to update these docs as I work on restructuring stories!

No problem! Yes, 100% agree on automating this. Is `stories.json` some sort of metadata holder for stories? I haven't used it before. Another alternative way I was thinking might be leveraging exported stories? Like, we will export and import stories into the docs, we might want to leverage the exported title and with a util function, auto generate the url maybe? I am not sure how much of it is possible though, might worth looking into when we start importing stories into docs..

langermank:
> stories.json

Check it out! https://primer.style/react/storybook/stories.json 

I think `id` makes up part of the url. In my new docs prototype, I'm constructing the url based on the new file structure I'm anticipating. But that's probably very fragile, so it would be nice to pull it from data. The thing about `stories.json` is that its not separated by component, so not super clear or intuitive. I wish Storybook would let us manipulate it!

docs/content/ActionList.mdx
broccolinisoup: Apparently we need `react` because the domain name in prod is `primer.style` not `primer.style/react`! Ah this confused me. Anyhow, I reverted back to the original version and just updated story links that caused by the format change. 

There are still 3 components using `primer.style` as opposed to dynamic domain name, I haven't touched them but asked @colebemis about it.

josepmartins:
+1 to automate this from the `stories.json` file. 

Wondering if this is something we could do in the `Doctocat` layout where we show the storybook link 🤔 : https://github.com/primer/doctocat/blob/main/theme/src/components/layout.js#L153

josepmartins (approved):

joshblack (approved):

josepmartins:
I filed a PR to generate the storybook links from the `stories.json` if you want to take a look. Needs a bit more testing to not break existing pages 👍 https://github.com/primer/doctocat/pull/505

cc @langermank @broccolinisoup

broccolinisoup:
> I filed a PR to generate the storybook links from the `stories.json` if you want to take a look. Needs a bit more testing to not break existing pages 👍 [primer/doctocat#505](https://github.com/primer/doctocat/pull/505)
> 
> cc @langermank @broccolinisoup

Oh great!! I'll check out the PR. I'd like to merge this now and we can discuss the links on the other PR ✨

PR #2510, opened 2022-11-01
Remove deprecated focus style primitives

langermank:
Describe your changes here.

Removing old/unused focus style CSS vars so we can deprecate them from `primer/primitives`

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/drafts/MarkdownEditor/_MarkdownInput.tsx
```diff
@@ -118,7 +118,7 @@ export const MarkdownInput = forwardRef<HTMLTextAreaElement, MarkdownInputProps>
             width: '100%',
             borderStyle: 'none',
             height: fullHeight ? '100%' : undefined,
-            boxShadow: isDraggedOver ? 'primer.shadow.focus' : 'none',
+            outline: isDraggedOver ? 'solid 2px color.accent.fg' : undefined,
```
langermank, line 121: Is this valid markup? I didn't see any stories for this code ☹️ should I add one?

joshblack:
@langermank random question, would there be a good way to pull out common styles (like for focus in this PR and defined in https://github.com/github/primer/issues/409) into a common module to apply to components when we need them? Similar to helper classes, mixins, or functions but with styled-components.

langermank:
> @langermank random question, would there be a good way to pull out common styles (like for focus in this PR and defined in [github/primer#409](https://github.com/github/primer/issues/409)) into a common module to apply to components when we need them? Similar to helper classes, mixins, or functions but with styled-components.

@joshblack I think ideally we would put global focus styles in [this file](https://github.com/primer/react/blob/b3a98baccc19a77d16b8c42f7873073070f24b89/src/BaseStyles.tsx) and then create some shared "mixins" so we can have custom focus style overrides for components. This is what we've done in PCSS. To be honest I'm not totally sure what the state of focus styles is in PRC, as I bet we are actually seeing focus styles in memex coming from PCSS. In that BaseStyles file, its removing `outline` globally 😦 I'd be happy to follow up and work on that, but this PR is just an attempt to remove the older focus style CSS vars 😆

langermank:
@joshblack I just stumbled upon this: https://github.com/primer/react/blob/main/src/_getGlobalFocusStyles.ts

joshblack (approved):

PR #2511, opened 2022-11-01
fix(Select): move selected from <option> to defaultValue on <select>

joshblack:
We currently set `selected` on the `<option>` within a `<select>` when `placeholder` is provided. When doing so, we receive the following error from `styled-components`:

```
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
    at option
    at select
    at Select__StyledSelect
    at span
    at _TextInputWrapper__TextInputWrapper
    at block (/Users/joshblack/gh/primer/react/src/Select.tsx:67:5)
```

This PR removes the `selected` prop from `<option>` and instead optionally specifies the `defaultValue` if `placeholder` is present.

> **Note**
> When a consumer passes in `defaultValue`, it should continue to override whatever is provided since `{...rest}` is applied _after_ `defaultValue`

### DIFF
--- a/.changeset/moody-islands-impress.md
+++ b/.changeset/moody-islands-impress.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Remove selected from <option> and add defaultValue for <select> when placeholder is provided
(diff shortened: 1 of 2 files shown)


josepmartins (approved):
🚀

PR #2512, opened 2022-11-01
test(Textarea): update controlled test to provide onChange

joshblack:
Update the `Textarea` test to include `onChange` when `value` is provided. This helps to address the following error that shows up in Jest:

```
Warning: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
    at textarea
    at Textarea__StyledTextarea
    at span
    at _TextInputWrapper__TextInputBaseWrapper
    at value (/Users/joshblack/gh/primer/react/src/Textarea.tsx:67:7)
```

### DIFF
--- a/src/__tests__/Textarea.test.tsx
+++ b/src/__tests__/Textarea.test.tsx
@@ -94,7 +94,8 @@ describe('Textarea', () => {
 
   it('renders a value in the textarea', () => {
     const mockValue = 'mock value'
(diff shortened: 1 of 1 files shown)


broccolinisoup (approved):
✨

PR #2515, opened 2022-11-01
test(Overlay): add spy for console.log() from stories

joshblack:
Update the `Overlay` tests to no longer emit `console.log()` from the stories. This approach uses `jest.spyOn` plus `mockImplementation` to prevent `console.log()` from preventing into the tests.

Since `console.log()` may be used for other parts of the test, it also has a guard in case `console.log()` is called with something other than what was expected from the story

#### Changelog

**New**

**Changed**

- Add a spy for `console.log()` to avoid logging in tests

**Removed**

- Removed `expect.extend(toHaveNoViolations)` as it should be automatically loaded in the environment


josepmartins (approved):
✨

PR #2516, opened 2022-11-01
test(ChoiceInputField): add mockImplementation to warning tests

joshblack:
Update the `ChoiceInputField` tests to use `mockImplementation` on the `console.warn()` spy. This prevents the `console.warn()` message from being emitted to the console during tests.

This change also updates each check to explicitly check for the message to make sure that no other `console.warn()` message is being emitted other than what is being tested for.


broccolinisoup (approved):
Love it 😍 I'll add this to my tests as well!

PR #2367, opened 2022-09-23
Remove non-tab elements from `TabNav` focus zone

iansan5653:
When there are focuseable elements inside `TabNav` (particularly inside the `Tab`s themselves) they should not be included in the arrow-key navigation. For example, this menu button should be focuseable via <kbd>Tab</kbd>, not via arrow keys:

Screenshot of the Projects view tab menu anchor button, focused.

This change filters out any elements without the role `tab` from the focus zone. This fixes https://github.com/primer/react/issues/2319 - for more context, please see that issue.

### Merge checklist

- [x] Added/updated tests
- ~~[ ] Added/updated documentation~~
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2527, opened 2022-11-03
test(TabNav): remove nested <a> from test

joshblack:
Remove the nested `<a>` from `TabNav.Link`. This helps to avoid the following `validateDOMNesting` error that comes from React:

```
  console.error
    Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
        at a
        at a
        at TabNav__TabNavLink
        at div
        at TabNav__TabNavTabList
        at nav
        at TabNav__TabNavNav
        at div
        at TabNav__TabNavBase
        at children (/Users/joshblack/gh/primer/react/src/TabNav.tsx:31:18)
        at div
        at Box
        at WrapperComponent (/Users/joshblack/gh/primer/react/node_modules/@wojtekmaj/enzyme-adapter-utils/src/createMountWrapper.jsx:46:26)

      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:67:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:43:5)
      at validateDOMNesting (node_modules/react-dom/cjs/react-dom.development.js:10083:7)
      at createInstance (node_modules/react-dom/cjs/react-dom.development.js:10181:5)
      at completeWork (node_modules/react-dom/cjs/react-dom.development.js:19464:28)
      at completeUnitOfWork (node_modules/react-dom/cjs/react-dom.development.js:22812:16)
      at performUnitOfWork (node_modules/react-dom/cjs/react-dom.development.js:22787:5)
```

This also updates the test to reference only the tab and no longer includes the link.


broccolinisoup (commented):
Looks like this test is written recently and seems like [a valid case](https://github.com/primer/react/pull/2367) but should it be a `link` ? In the PR description, it mentions about a case where there is a nested button inside the tab so maybe instead of removing this test case, we can use a `button` element and still make sure to cover nested buttons test?

joshblack:
@broccolinisoup I think what they end up doing in Memex is changing the container `<a>` to a div for the "active tab" and so I could try doing that for the test to keep it in place 👀

joshblack:
bump @siddharthkp @broccolinisoup let me know if you have any questions!

broccolinisoup:
> @broccolinisoup I think what they end up doing in Memex is changing the container `<a>` to a div for the "active tab" and so I could try doing that for the test to keep it in place 👀

Ah I see. Thanks for clarification! I feel like we are good without testing this case as it is very specific to their implementation and not necessarily a part of the TabNav functionality

broccolinisoup (approved):

PR #2529, opened 2022-11-03
Add tests to enforce Storybook format

colebemis:
This PR adds tests to enforce the storybook format rules described in the [Storybook ADR](https://github.com/github/primer/pull/1387). Specifically, we added a test to ensure every component story file has a "Default" story and that the "Default" story doesn't accept any args.

I also updated the TreeView stories to follow the new format rules.

For now, these storybook tests are opt-in. We'll opt-in components incrementally as we reformat them to align with the ADR.

## Example test results

```
 PASS  src/__tests__/storybook.test.tsx
  src/TreeView/TreeView.stories.tsx
    ✓ exports a Default story (2 ms)
    ✓ Default story does not use args (1 ms)
```


src/__tests__/storybook.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+  .sync('src/**/*.stories.tsx')
+  .filter(file => components.some(component => file.includes(component)))
+  .filter(file => !file.endsWith('.features.stories.tsx'))
+
+for (const storyFile of storyFiles) {
+  describe(`${storyFile}`, () => {
+    const storyFileContents = fs.readFileSync(storyFile, 'utf8')
+
+    test('exports a Default story', () => {
+      expect(storyFileContents).toMatch(/export const Default/)
```
colebemis, line 17: Is there a more robust way to check the exports of the story file? Maybe with ts-morph? cc @joshblack @siddharthkp

langermank (approved):
Thanks for pairing on this!

src/__tests__/storybook.test.tsx
joshblack: @colebemis through up a draft here of a possible approach: https://github.com/primer/react/pull/2537

Curious what you think!

PR #2520, opened 2022-11-02
test: add playwright support

joshblack:
#### Overview

Preview link for Testing docs: [Link](https://github.com/primer/react/blob/d82b9f6831a98ca2d50114db4e8430292b542a39/contributor-docs/testing.md)

**New**

- Add in `contributor-docs/testing.md` to document Playwright process
- Add `playwright.config.ts` and accompanying matchers for axe and helpers for storybook
- Add `e2e` directory with tests for Button
  - Each story has it's own test suite in `Button.test.ts`
  - Each story is tested against each theme
  - Each story runs against axe
  - The default and focus state of each story is captured as snapshots
- Add `vrt` job to `CI` workflow
- Add `avt` job to `CI` workflow

**Goals**

- Keep CI time under 10min
- Screenshot the default and focus state in each Button story
- Run axe against each Button story (Note: these will currently fail due to button contrast)
- Be able to download and view test results from CI as well as locally

**Next steps**

- Add tests for the new `UnderlineNav` to introduce layout primitives to the setup
- Include interactions for `UnderlineNav`
- Add another shard for vrt runners

PR #2532, opened 2022-11-04
[UnderlineNav]: VRT Snapshot and interaction Tests (Playwright)

broccolinisoup:
This PR is a follow up on @joshblack's [playwright PR](https://github.com/primer/react/pull/2520)
 and covers UnderlineNav related next steps.

- Introduce primer layout primitives
- Tests for states of UnderlineNav on available Primer layout breakpoints and themes (5 breakpoint and 9 themes)
- Interaction tests 


e2e/components/UnderlineNav.test.ts
```diff
@@ -0,0 +1,159 @@
+import {test, expect} from '@playwright/test'
+import {visit} from '../test-helpers/storybook'
+import {themes} from '../test-helpers/themes'
+import {viewports} from '../test-helpers/viewports'
+
+test.describe('UnderlineNav', () => {
+  Object.keys(viewports).forEach(key => {
```
joshblack, line 7: If it helps, another option could be:

```suggestion
  Object.entries(viewports).forEach(([viewport, size])=> {
```

e2e/components/UnderlineNav.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    themes.forEach(theme => {
+      test(`VRT @ ${key} (${viewports[key]}px) on ${theme} theme`, async ({page}) => {
```
joshblack, line 9: What do you think of having the test structure be something like:

- Top-level: component name
- First-level: story name
- Second-level: theme + modifiers + tags

For example:

```ts
test.describe('UnderlineNav', () => {
  test.describe('Internal Responsive Nav', () => {
    themes.forEach(theme => {
      Object.entries(viewports).forEach(([viewport, width]) => {
        test(`${theme} (${viewport}) @vrt`, async ({ page }) => {
        })
      })
    });
  });
});
```

e2e/components/UnderlineNav.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  test('Hide Icons when there is not enough space to display all list items', async ({page}) => {
+    await visit(page, {
+      id: 'components-underlinenav--internal-responsive-nav'
+    })
+
+    // Default State
+    expect(await page.screenshot()).toMatchSnapshot()
+
+    // Resize
+    await page.setViewportSize({width: viewports['primer.breakpoint.md'], height: 768})
```
joshblack, line 52: Would it be helpful to run axe at all during things like resize or when the menu is open, or not really?

joshblack (commented):
So cool! Just left a couple comments/questions, can't wait to hear what you think 👀

e2e/components/UnderlineNav.test.ts
broccolinisoup: Nice ✨

e2e/components/UnderlineNav.test.ts
broccolinisoup: Ahh I guess I didn't think that we need to cover every story to VRT test 🙈  I rather chose one story to test the themes and viewports and write possible interaction tests based on the same story but I see now that is not the best approach. Thanks for mentioning that and yes your suggested structure makes perfect sense. I'll push a commit to update them

broccolinisoup:
Closing this because we have the e2e interaction test in this [PR](https://github.com/primer/react/pull/2632)

PR #2534, opened 2022-11-04
docs: add story sort function

joshblack:
#### Changelog

**New**

**Changed**

- Update `storySort` to a function pattern to enable custom sorting of sections, in particular when nested

**Removed**

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -18,9 +18,115 @@ export const parameters = {
     removeEmptyComments: true
   },
   options: {
-    storySort: {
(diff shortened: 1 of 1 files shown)


colebemis:
Is it possible to sort the Playground story above the subcomponent Playgrounds?

CleanShot 2022-11-04 at 12 02 13

Ideal order:
- Default
- Playground
- Group Playground
- Item Playground
- Link Item Playground

joshblack:
@colebemis can definitely try!

joshblack:
@colebemis just pushed something up that should do the trick, let me know if that's what you were thinking!

langermank (approved):
Thank you! 🚀

PR #2538, opened 2022-11-05
TreeView: Preserve expanded state of nested items

colebemis:
As requested by @jsholes in his [a11y sign-off review](https://github.com/github/primer/issues/1114#issuecomment-1292397344), this PR persists the expanded state of nested tree items when their parent is collapsed/expanded.

> In some instances, the tree view seems to reset the expanded/collapsed state of descendant nodes when a parent's state is changed. 
> Expanding or collapsing a parent node must not have an impact on the state of any of its descendants. I.e. if a descendent is collapsed, it should stay collapsed even if the parent is collapsed and then expanded again, and vice versa.

### Before

https://user-images.githubusercontent.com/4608155/200477465-e17243b9-abb2-4923-a821-bb5c0e700ff8.mp4

### After

https://user-images.githubusercontent.com/4608155/200477545-49728ee4-d58e-4851-bdc2-d198488773f8.mp4

### Breaking changes

To implement this change, I had to introduce a required `id` prop for all TreeView.Items.

```diff
- <TreeView.Item>...
+ <TreeView.Item id="unique-id">...
```

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated docs
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/hooks/useControllableState.ts
```diff
@@ -43,6 +43,8 @@ export function useControllableState<T>({
   const controlled = React.useRef<boolean | null>(null)
   const stableOnChange = React.useRef(onChange)
 
+  // console.log('useState', state)
+
```
colebemis, line 47:
```suggestion
```

src/TreeView/TreeView.tsx
```diff
@@ -26,8 +26,10 @@ import {useTypeahead} from './useTypeahead'
 
 const RootContext = React.createContext<{
   announceUpdate: (message: string) => void
+  expandedStateCache: React.RefObject<Map<string, boolean>>
```
colebemis, line 29: TODO: Add a code comment here explain why this cache is necessary

src/hooks/useControllableState.ts
colebemis, line 47:
```suggestion
```

src/TreeView/TreeView.tsx
```diff
@@ -257,8 +262,15 @@ const Root: React.FC<TreeViewProps> = ({'aria-label': ariaLabel, 'aria-labelledb
     setAriaLiveMessage(message)
   }, [])
 
+  const expandedStateCache = React.useRef(new Map())
+
```
joshblack, line 266: Small suggestion to avoid creating a new Map each render:

```suggestion
  const expandedStateCache = React.useRef(null)

  if (expandedStateCache.current === null) {
    expandedStateCache.current = new Map();
  }
```

joshblack (approved):

Issue #2461, opened 2022-10-20
Update test suite to fail on `console.*` calls


joshblack:
Currently there are `console.{error,warn,log}` statements being emitted when running `npm test`. This can be confusing to contributors who may not understand if the issues are pre-existing or caused by changes that they made when making a contribution.

This issue has two goals:

- [x] Add a system by which we can fail tests that call `console.{error, warn, log}` (Note: `console.log()` should only fail in CI)
- [x] Update tests which currently fail when adding this check
  - [x] [`src/drafts/MarkdownEditor/MarkdownEditor.test.tsx`](https://github.com/primer/react/pull/2560)
  - [x] `src/__tests__/useSafeTimeout.test.tsx`
  - [x] [`src/__tests__/TabNav.test.tsx`](https://github.com/primer/react/pull/2527)
  - [x] `src/drafts/MarkdownViewer/MarkdownViewer.test.tsx`
  - [x] [`src/__tests__/Textarea.test.tsx`](https://github.com/primer/react/pull/2512)
  - [x] `src/drafts/hooks/useIgnoreKeyboardActionsWhileComposing.test.tsx`
  - [x] `src/__tests__/RadioGroup.test.tsx`
  - [x] [`src/__tests__/Select.test.tsx`](https://github.com/primer/react/pull/2511)
  - [x] [`src/__tests__/Radio.test.tsx`](https://github.com/primer/react/pull/2513)
  - [x] `src/__tests__/CheckboxGroup.test.tsx`
  - [x] `src/__tests__/CheckboxOrRadioGroup.test.tsx`
  - [x] `src/NavList/NavList.test.tsx`
  - [x] [`src/__tests__/deprecated/ChoiceInputField.test.tsx`](https://github.com/primer/react/pull/2516)
  - [x] [`src/__tests__/Autocomplete.test.tsx`](https://github.com/primer/react/pull/2545)
  - [x] `src/SegmentedControl/SegmentedControl.test.tsx`
  - [x] `src/__tests__/FormControl.test.tsx`
  - [x] [`src/__tests__/ThemeProvider.test.tsx`](https://github.com/primer/react/pull/2526)
  - [x] [Overlay](https://github.com/primer/react/pull/2515)

joshblack:
Closed by: https://github.com/primer/react/pull/2730

PR #2545, opened 2022-11-07
refactor(Autocomplete): update tests and add safeTimeout to input

joshblack:
Context https://github.com/primer/react/issues/2461

Update the tests for Autocomplete so that they no longer error out when running tests. These changes capture a variety of warnings, notably:

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()
```

```
Warning: It looks like you're using the wrong act() around your test interactions.
    Be sure to use the matching version of act() corresponding to your renderer:

    // for react-dom:
    import {act} from 'react-dom/test-utils';
    // ...
    act(() => ...);

    // for react-test-renderer:
    import TestRenderer from react-test-renderer';
    const {act} = TestRenderer;
    // ...
    act(() => ...);
```

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

```
console.error
    Error: Uncaught [Error: AutocompleteContext returned null values]
```

**Note**

In order to get the snapshot tests and testing-library tests working together, I needed to split them up into a separate test suite. Otherwise, the mismatch `act()` warning would not go away. Let me know if there is a better approach when everything is in one test suite, I couldn't figure out a way to make it work 🤔 

#### Changelog

**New**

**Changed**

- Update `AutocompleteInput` to use the `useSafeTimeout` hook instead of `setTimeout` directly
- Update `Autocomplete` tests to no longer emit errors in console

**Removed**


src/Autocomplete/AutocompleteInput.tsx
```diff
@@ -12,6 +12,7 @@ import {AutocompleteContext} from './AutocompleteContext'
 import TextInput from '../TextInput'
 import {useRefObjectAsForwardedRef} from '../hooks/useRefObjectAsForwardedRef'
 import {ComponentProps} from '../utils/types'
+import useSafeTimeout from '../hooks/useSafeTimeout'
```
mperrotti, line 15: This is a great hook. I didn't even realize we had it!

src/Autocomplete/AutocompleteInput.tsx
```diff
@@ -56,13 +58,13 @@ const AutocompleteInput = React.forwardRef(
         // HACK: wait a tick and check the focused element before hiding the autocomplete menu
         // this prevents the menu from hiding when the user is clicking an option in the Autoselect.Menu,
         // but still hides the menu when the user blurs the input by tabbing out or clicking somewhere else on the page
-        setTimeout(() => {
+        safeSetTimeout(() => {
```
mperrotti, line 61: It would be even better if we could refactor this setTimeout away entirely. `InlineAutocomplete` managed to do it with a synthetic change event: https://github.com/primer/react/blob/main/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx#L155

I don't think that should block the rest of these changes though.

mperrotti (approved):

src/__tests__/Autocomplete.test.tsx
```diff
@@ -457,11 +320,145 @@ describe('Autocomplete', () => {
...
       expect(() => HTMLRender(<Autocomplete.Menu items={mockItems} selectedItemIds={[]} />)).toThrowError(
         'AutocompleteContext returned null values'
       )
+
+      spy.mockRestore()
+    })
+  })
+})
+
+describe('snapshots', () => {
```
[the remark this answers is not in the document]
joshblack: @mperrotti yeah, for a super odd reason too 🤔 Let me know if you have any ideas!

> In order to get the snapshot tests and testing-library tests working together, I needed to split them up into a separate test suite. Otherwise, the mismatch act() warning would not go away. Let me know if there is a better approach when everything is in one test suite, I couldn't figure out a way to make it work 🤔

PR #2546, opened 2022-11-07
feat(TreeView): add support for PageUp, PageDown

joshblack:
Add in support for `PageUp` and `PageDown` on `TreeView`. The way this works is that when a user presses `PageDown` while focus is within a `TreeView` we look for the first `treeitem` following the currently active item that is not currently in the viewport. If one exists, we focus it.

`PageUp` works the same way but in reverse, look for `treeitem`'s before the current item and if one is currently not in the viewport focus it.

### DIFF
--- a/src/TreeView/TreeView.tsx
+++ b/src/TreeView/TreeView.tsx
@@ -241,9 +241,57 @@ const UlBox = styled.ul<SxProp>`
 const Root: React.FC<TreeViewProps> = ({'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledby, children}) => {
   const containerRef = React.useRef<HTMLUListElement>(null)
   const [ariaLiveMessage, setAriaLiveMessage] = React.useState('')
+  const announceUpdate = React.useCallback((message: string) => {
+    setAriaLiveMessage(message)
+  }, [])
+
+  /**
+   * Manage the active element when the user presses `PageUp` or `PageDown`
+   * while using the TreeView
+   */
+  function onKeyDown(event: React.KeyboardEvent) {
+    if (containerRef.current === null) {
+      return
+    }
+
+    // When pressing PageUp, look at treeitem's above the current active element
+    // and find the first one that is not in the viewport. If one exists, focus it
+    if (event.key === 'PageUp') {
+      const {activeElement} = document
+      const items = Array.from(containerRef.current.querySelectorAll('[role="treeitem"]'))
+        .filter(item => {
+          const position = activeElement.compareDocumentPosition(item)
+          return position & Node.DOCUMENT_POSITION_PRECEDING
+        })
+        .reverse()
+
+      for (const item of items) {
+        if (!isInViewport(item)) {
+          item.focus()
+          break
+        }
+      }
+    }
+
+    // When pressing PageDown, look at treeitem's below the current active element
+    // and find the first one that is not in the viewport. If one exists, focus it
+    if (event.key === 'PageDown') {
+      const {activeElement} = document
+      const items = Array.from(containerRef.current.querySelectorAll('[role="treeitem"]')).filter(item => {
+        const position = activeElement.compareDocumentPosition(item)
+        return position & Node.DOCUMENT_POSITION_FOLLOWING
+      })
+
+      for (const item of items) {
+        if (!isInViewport(item)) {
+          item.focus()
+          break
+        }
+      }
+    }
+  }
 
   useRovingTabIndex({containerRef})
-
   useTypeahead({
     containerRef,
     onFocusChange: element => {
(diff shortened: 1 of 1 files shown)


src/TreeView/TreeView.tsx
```diff
@@ -241,9 +241,57 @@ const UlBox = styled.ul<SxProp>`
...
+  const announceUpdate = React.useCallback((message: string) => {
+    setAriaLiveMessage(message)
+  }, [])
 
-  useRovingTabIndex({containerRef})
+  /**
+   * Manage the active element when the user presses `PageUp` or `PageDown`
+   * while using the TreeView
+   */
+  function onKeyDown(event: React.KeyboardEvent) {
```
colebemis, line 252: Can we move this code into the useRovingTabIndex() hook? Something like this:

```diff
diff --git a/src/TreeView/useRovingTabIndex.ts b/src/TreeView/useRovingTabIndex.ts
index 83d336645..9903442ac 100644
--- a/src/TreeView/useRovingTabIndex.ts
+++ b/src/TreeView/useRovingTabIndex.ts
@@ -5,7 +5,12 @@ export function useRovingTabIndex({containerRef}: {containerRef: React.RefObject
   // TODO: Initialize focus to the aria-current item if it exists
   useFocusZone({
     containerRef,
-    bindKeys: FocusKeys.ArrowVertical | FocusKeys.ArrowHorizontal | FocusKeys.HomeAndEnd | FocusKeys.Backspace,
+    bindKeys:
+      FocusKeys.ArrowVertical |
+      FocusKeys.ArrowHorizontal |
+      FocusKeys.HomeAndEnd |
+      FocusKeys.Backspace |
+      FocusKeys.PageUpDown,
     preventScroll: true,
     getNextFocusable: (direction, from, event) => {
       if (!(from instanceof HTMLElement)) return
@@ -67,6 +72,14 @@ export function getNextFocusableElement(activeElement: HTMLElement, event: Keybo
     case 'End':
       // Focus last visible element
       return getLastElement(activeElement)
+
+    case 'PageUp':
+      // TODO
+      return
+
+    case 'PageDown':
+      // TODO
+      return
   }
 }
```

colebemis (changes requested):
Looks great!

I think most of this code probably belongs in the useRovingTabIndex.ts file

src/TreeView/TreeView.tsx
joshblack: @colebemis will adding it to `bindKeys` cause `event.preventDefault()` to be called? One thing that's nice about the approach here is that `event.preventDefault()` does not need to be called so we can tag-a-long with the native `PageDown` or `PageUp` functionality.

Unfortunately if the page scroll is not being updated then the heuristic will need to be changed a bit in order to work. Let me know what you think

colebemis: Ah bummer, you're right. How do you feel about pulling this code out into a `usePageUpDown` hook?

colebemis:
I checked out the branch locally, and this is the behavior I'm seeing:

https://user-images.githubusercontent.com/4608155/200644148-17fa7f40-dcdd-42a2-aa79-c5b8194cbc40.mp4

Looks like maybe a race condition between the focus change and the default PageUp/Down behavior?

joshblack:
@colebemis seeing that too on the stress test one, let me dig into it and also check on other browsers too where `PageUp`, `PageDown` may have different scroll behavior

joshblack:
@colebemis just updated with a different approach, let me know what you think!

Quick cross-browser check:

https://user-images.githubusercontent.com/3901764/200686491-3a32e054-bd33-49fe-aac5-24bce9ebc696.mov

src/TreeView/usePageUpDown.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }, [ref, onScrollEnd])
+}
+
+/**
+ * Provides a function, `onScrollEnd`, that will call the given callback when it
+ * detects that the application is no longer scrolling.
+ */
+function useScrollEnd() {
+  // Keep tracking of the current timeout and scroll listener in case the owner
```
joshblack, line 100:
```suggestion
  // Keep track of the current timeout and scroll listener in case the owner
```

src/TreeView/usePageUpDown.ts
```diff
@@ -0,0 +1,147 @@
...
+        clearTimeout(timeoutId.current)
+      }
+      timeoutId.current = window.setTimeout(() => {
+        callback()
+        window.removeEventListener('scroll', listener)
+        activeListener.current = null
+      }, 60)
+    }
+
+    window.addEventListener('scroll', listener)
```
colebemis, line 116: Can we attach this listener to the nearest scroll container instead of `window`?

It looks like the nested scroll container story in storybook doesn't work:
https://user-images.githubusercontent.com/4608155/200848391-bc13c325-9366-45de-8757-4e5c5619c24e.mp4

colebemis: Theres a `getScrollContainer` function in PageLayout that we might be able to reuse here: https://github.com/primer/react/blob/dc1cf6da2864dd7dfa4ed0cb2c5362f9ff9dd60f/src/PageLayout/useStickyPaneHeight.ts#L109-L115

joshblack: Done! Thanks for the suggesting, pulled it out into a utils file.

joshblack:
@colebemis thanks for the feedback! Updated to try another heuristic with `keyup` plus the scroll container note you brought up 👀

colebemis (approved):
Works so well! ✨ Excellent job on the implementation. This ended up being a lot trickier than I expected.

PR #2547, opened 2022-11-07
docs: enable fastRefresh and strictMode for storybook

joshblack:
Update the storybook config to enable `React.StrictMode` in stories. This tool highlights potential problems in components, for more information check out this post: https://reactjs.org/docs/strict-mode.html#ensuring-reusable-state

You can verify that `React.StrictMode` is enabled when using "unsafe" code, for example:

Screenshot 2022-11-07 at 4 33 01 PM

And you should see an error like this in the console:

Screenshot 2022-11-07 at 4 42 22 PM

#### Changelog

**New**

**Changed**

- Update storybook config to enable `strictMode`

**Removed**

- `storybook-addon-performance`, it does not seem to support React 18 and doesn't work when `strictMode` is enabled, unfortunately 😞

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -6,7 +6,6 @@ module.exports = {
     '@storybook/addon-interactions',
     '@storybook/addon-a11y',
     '@storybook/addon-links',
-    'storybook-addon-performance/register',
     {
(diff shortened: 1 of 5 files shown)


colebemis (approved):
🚀

PR #2551, opened 2022-11-08
UnderlineNav2: Add aria-hidden and sr-only class support for descriptive counters

broccolinisoup:

> The span wrapping the counter value is adjoined to the link text for screen reader users, creating accessible names like: "Issues12". The current span should be hidden from screen reader users with aria-hidden="true", and an alternative created with class="sr-only" (or relevant off-screen/screen-reader-only class) that recreates exactly the same value but with some separation and parentheses

```
<span aria-hidden="true" class="CounterLabel-sc-13ceqbg-0 ikCrlz">12</span>
<span class="sr-only">&nbsp;(12)</span>
```

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -353,10 +353,15 @@ export const UnderlineNav = forwardRef(
                             {actionElementChildren}
 
                             {loadingCounters ? (
-                              <LoadingCounter />
+                              <Box as="span" data-component="counter">
+                                <LoadingCounter />
```
josepmartins, line 357: non-blocking and asking out of ignorance: Is it necessary to add a span/wrapper to the `LoadingCounter` or should be hidden from screen readers? 🤔

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -175,7 +176,8 @@ export const UnderlineNavItem = forwardRef(
             ) : (
               counter !== undefined && (
                 <Box as="span" data-component="counter" sx={counterStyles}>
-                  <CounterLabel>{counter}</CounterLabel>
+                  <CounterLabel aria-hidden="true">{counter}</CounterLabel>
+                  <VisuallyHidden>&nbsp;({counter})</VisuallyHidden>
```
josepmartins, line 182: non-blocking: Maybe it's just relative to the `UnderlineNav`, but I wonder if we should consider adding this behavior inside the `Counter` component

broccolinisoup: I am not really sure if I can follow you, sorry! 

Do you mean like this 
```
<CounterLabel sx={counterStyles}>
    <Box as="span" data-component="counter" aria-hidden="true">
      {counter}
    </Box>
    <VisuallyHidden>&nbsp;({counter})</VisuallyHidden>
</CounterLabel>
```

If so, the counterStyles doesn't really apply to position the counter itself rather they position the string within the counter label. 

**Wrapper inside the CounterLabel**
Screen Shot 2022-11-09 at 10 48 24 am

**wrapper outside the CounterLabel** (matches with the design)
Screen Shot 2022-11-09 at 10 48 32 am

src/UnderlineNav2/UnderlineNav.tsx
broccolinisoup: That wasn't an intentional change, sorry! Thanks for pointing this out.

src/UnderlineNav2/UnderlineNavItem.tsx
josepmartins: My bad 🙇 ! The comment was more into adding all of these wrappers as part of the `Counter` component itself. 

It seems that any other component using it, like [Buttons with counters](https://primer.style/react/Button#counter-component) might benefit from adding the `Box` wrapper with the aria-hidden and the `VisuallyHidden` element.

broccolinisoup: Ahh Okay! Yeah, that is a great idea. `CounterLabel` could definitely use some help. How about I create an issue for that and we can discuss from there? I'm going to merge this pull requests as you approved but let me know if you have any concerns 🙌🏼

josepmartins: Love it 👍 ❤️ !!

broccolinisoup: Just dropping [the issue](https://github.com/github/primer/issues/1509) here for reference 🙂

PR #2552, opened 2022-11-08
UnderlineNav2: Add aria-hidden and sr-only class support for a descriptive "More" button label

broccolinisoup:

> The label of the "More" overflow button is not descriptive, and could easily result in multiple buttons with the same name if the component is used multiple times on the same page. The heading text should be added to the accessible name of the button, e.g. "More Repository navigation items" or "More: Repository navigation", e.g. using an off-screen span inside the button and hiding the existing label from screen reader users with aria-hidden. Note that, as with the counter, the visible text of "More" must be included in the accessible name verbatim.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


ericwbailey (approved):

src/UnderlineNav2/UnderlineNav.tsx
```diff
@@ -325,7 +325,10 @@ export const UnderlineNav = forwardRef(
                   onClick={onAnchorClick}
                   trailingIcon={TriangleDownIcon}
                 >
-                  More
+                  <Box as="span" aria-hidden="true">
+                    More
+                  </Box>
+                  <VisuallyHidden as="span">{`More: ${ariaLabel} items`}</VisuallyHidden>
```
joshblack, line 331: I think it could be helpful to structure this like:

```suggestion
                  <Box as="span">
                    More <VisuallyHidden as="span">{`${ariaLabel} items`}</VisuallyHidden>
                  </Box>
                  
```

This could help to avoid issues identified in: https://www.w3.org/WAI/WCAG22/Techniques/general/G211 where the accessible and visual label differ and make sure that they at least include a common base (in this case, text: "More")

Curious what you think @broccolinisoup @ericwbailey 👀

joshblack (changes requested):
Left a comment above for a structure question, let me know what you think!

src/UnderlineNav2/UnderlineNav.tsx
broccolinisoup: 😮 I LOVE IT SO MUCH

ericwbailey: SAME

joshblack (approved):

PR #2553, opened 2022-11-08
UnderlineNav2: Only allow `Enter` and `Space` to select item

broccolinisoup:

>  For some reason, focusing any of the nav links and typing any printable character on the keyboard (including numbers) activates the link and causes it to become the current page. This should be removed.

### Merge checklist

- [ ] Added/updated tests - N/A
- [ ] Added/updated documentation - N/A
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


ericwbailey (approved):

src/UnderlineNav2/UnderlineNavItem.tsx
```diff
@@ -121,11 +121,11 @@ export const UnderlineNavItem = forwardRef(
...
     const keyPressHandler = React.useCallback(
       event => {
-        if (!event.defaultPrevented && [' ', 'Enter'].includes(event.key)) {
-          if (typeof onSelect === 'function') onSelect(event)
-          if (typeof afterSelect === 'function') afterSelect(event)
+        if ([' ', 'Enter'].includes(event.key)) {
+          if (!event.defaultPrevented && typeof onSelect === 'function') onSelect(event)
+          if (!event.defaultPrevented && typeof afterSelect === 'function') afterSelect(event)
+          setSelectedLink(ref as RefObject<HTMLElement>)
         }
```
joshblack, line 128: Just wanted to leave a potential refactor idea, shouldn't change any functionality. Feel free to dismiss if you think they aren't really helping anything 👀 

- Since there are only two members in `[' ', 'Enter']` it might make sense to just explicitly check for both of them. If there are more expected members, though, I totally would get having a collection
- Instead of doing the explicit guard for each callback, this might be able to be simplified into a single `if` branch and use the optional chaining syntax `name?.()`

```suggestion
        if (event.key === ' ' || event.key === 'Enter') {
          if (!event.defaultPrevented) {
            onSelect?.(event)
            afterSelect?.(event);
          }
          setSelectedLink(ref as RefObject<HTMLElement>)
        }
```

Just a couple of thoughts, feel free to disregard if they're not helpful!

joshblack (approved):

src/UnderlineNav2/UnderlineNavItem.tsx
broccolinisoup: I like it ✨ They are helpful!!

` onSelect?.(event)` makes sense. How do I make sure that it is a function type as well? Or checking its type isn't  necessary here at all? 🤔

broccolinisoup: Because `onSelect` is an exposed event to the API, I feel like it might worth to check the type. It is unlikely that will error out when there is type checking and proper linting rules in place on consumer side but still just to be double cautious to prevent them passing things on to `onSelect` other than functions. I'd like to keep the function type checking if that is okay with you and see how it goes 👀  

I'll apply your other suggestions on a commit from my local ✨

PR #2562, opened 2022-11-10
Automatically assign release conductor

siddharthkp:
This workflow looks at `changeset-release/main` and assigns the release conductor on schedule (on pagerduty)

Added `PAGERDUTY_API_KEY_SID` to this repo, we should eventually replace that with a key that's not dependent on me 😋 

Have tested the API, but can't test this PR on a different branch, would need to merge to main to test


joshblack (approved):
🔥

PR #2568, opened 2022-11-10
refactor(TreeView): update PageUp, PageDown to page-based movement

joshblack:
Based on feedback from: https://github.com/github/primer/issues/1114#issuecomment-1310441556, this PR updates `TreeView` to have a page-based motion when using PageUp and PageDown. Specifically, it tries to capture the following behaviors:

- When pressing Page Up or Page Down, focus should move by a consistent number of nodes
  - This number is currently based on the height of tree and the number of visible items within it
- If the number of nodes between the user's current focus and first root node is less than the paging amount, Page Up should basically act like the Home key.
- Similar to above, Page Down should act as the End key

#### Changelog

**New**

**Changed**

- Update `useRovingTabIndex` hook to support PageUp and PageDown

**Removed**

- `usePageUpDown` hook


src/TreeView/useRovingTabIndex.ts
```diff
@@ -150,3 +162,68 @@ export function getLastElement(element: HTMLElement): HTMLElement | undefined {
...
+  }
+
+  const items = Array.from(root.querySelectorAll('[role="treeitem"]')).filter((item): item is HTMLElement => {
+    return item.nodeType === Node.ELEMENT_NODE
+  })
+  const scrollContainer = getScrollContainer(root)
+  const visible = items.filter(item => {
+    return isVisible(item.firstElementChild as HTMLElement, scrollContainer)
+  })
+  const pageSize = items.indexOf(visible[visible.length - 1]) - items.indexOf(visible[0]) + 1
```
colebemis, line 179: Is there a reason we can't use `visible.length` to compute the page size?

src/TreeView/useRovingTabIndex.ts
```diff
@@ -150,3 +162,68 @@ export function getLastElement(element: HTMLElement): HTMLElement | undefined {
...
```
colebemis, line 197: We could probably pull out a `getPageSize()` helper function to reduce duplication

.changeset/soft-rockets-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Update PageUp and PageDown for TreeView to move by a consistent number of nodes
```
colebemis, line 5: Since the PageUp/Down functionality hasn't shipped in a release yet, we probably don't need to add a changeset

colebemis (approved):
Functionality feels great! 💖 Just left some polish-related comments about the code.

PR #2580, opened 2022-11-14
ci: add vrt workflow for updating snapshots

joshblack:
This PR adds in a workflow for updating snapshots when the label "visual changes approved" is present. This is useful when a contributor would like to update snapshots automatically for their changes or is unable to update them locally.

The current strategy for this workflow is to run after the `workflow_run` event. The sole job in this workflow, `update-snapshots`, only runs when the following conditions are met:

- The CI workflow has completed and was successful
- The event is from the `primer/react` repository
- The Pull Request has the label "visual changes approved"

The `workflow_run` event and the label check is inspired by: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ to try and make sure the action only runs in the appropriate context since it will can use secrets/permissions.

One thing to consider is that this workflow may not be able to trigger status checks to fire on the commit it creates. If so, we will need to figure out a way to use a PAT in order to get these status checks to fire.

**Testing & Reviewing**

Unfortunately, these types of actions are hard to test from the Pull Request stage so this will likely require merging and then we can test and correct from there.

### DIFF
--- a/.github/workflows/vrt.yml
+++ b/.github/workflows/vrt.yml
@@ -1,7 +1,7 @@
 name: vrt
 on:
   workflow_run:
-    workflows: [ci]
+    workflows: [CI]
     types:
       - completed
 
--- a/.github/workflows/vrt.yml
+++ b/.github/workflows/vrt.yml
@@ -1,0 +1,51 @@
+name: vrt
+on:
+  workflow_run:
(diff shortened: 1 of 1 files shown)


colebemis (approved):
Looks great! 🚀

Issue #2582, opened 2022-11-14
Remove `@whitespace/storybook-addon-html` from storybook


joshblack:
This issue tracks removing `@whitespace/storybook-addon-html` from storybook as it does not support React 18.

It seemed appropriate to remove this as a result, however if there is a use-case for this let me know! Any alternatives would be very much so appreciated 🙏

rezrah:
👋 @joshblack this was added to help the accessibility team view rendered output more easily in Storybook in a11y reviews. Unsure how frequently it's used, but worth checking in with @alliethu about it.

langermank:
I agree this plugin is useful for accessibility review and also general development, but I don't think it should be a blocker. Perhaps we could make our own addon to accomplish this 😄 I bet it would be simple? (famous last words)

joshblack:
Thanks so much for the context @rezrah @langermank! I appreciate it 🙏 

Do either of you know of an alternative by any chance? I'll also keep an eye on the repo as it seems like they have been making changes recently. It seems like right now they have react-dom@18 as a peer dep but not react 🤔 

Looking forward, worse-can we can definitely bring it over as a custom add-on 👍

lesliecdubs:
We reviewed at the PRC planning meeting today. No one had ideas for an alternate addon for this behavior. +1 to adding it as a custom add-on for now if the accessibility team requires it for their review.

PR #2583, opened 2022-11-14
chore: remove storybook-addon-html from storybook

joshblack:
Closes https://github.com/primer/react/issues/2582

#### Changelog

**New**

**Changed**

- Update storybook config to no longer refer to removed add-on

**Removed**

- Remove `@whitespace/storybook-addon-html` as it does not support React 18

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -14,10 +14,7 @@ module.exports = {
       options: {
         optimizationLevel: 2
       }
-    },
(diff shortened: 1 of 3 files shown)


josepmartins (approved):
LGTM! 

just heads up that @rezrah left a comment on the issue: https://github.com/primer/react/issues/2582#issuecomment-1314516040

joshblack:
Going to defer on this one in case the add-on itself gets updated (🤞)

Will revisit this closer to the switch point if we need it

PR #2586, opened 2022-11-16
test: add storybook stories test

joshblack:
Add in support for generating e2e test files. This approach uses manually entered stories within `script/generate-e2e-tests.js` to generate files. This can be extended in the future to read from `stories.json` that is generated in `storybook-static`.

This helper also uses `recast` to parse and verify the generated files are valid. This will preserve file formatting compared to parsers like babel and could also allow us to easily modify/extend files while maintaining the base behavior in the future.

Finally, this PR uses the new `update snapshots` label to demonstrate how contributors can update snapshots automatically through CI instead of updating them locally on their machine.

#### Changelog

**New**

- Add `script/generate-e2e-tests.js` to scaffold e2e test files
- Generate `ActionList.test.ts` in `e2e/components`
- Save axe results in `.playwright/axe`

**Changed**

- Rename `avt` (accessibility verification testing) to `aat` (automated accessibility testing) across files and workflows
- Update default viewport width to use primer large breakpoint

**Removed**


e2e/components/Button.test.ts
```diff
@@ -22,7 +22,7 @@ test.describe('Button', () => {
       })
     })
 
-    test('axe @avt', async ({page}) => {
+    test('axe @aat', async ({page}) => {
```
colebemis, line 25: What does AAT stand for?

joshblack: @colebemis I take it to mean "automated accessibility tests"

PR #2589, opened 2022-11-16
Assert that component story titles start with Component/

colebemis:
Ensures component Storybook stories have consistent titles


langermank (approved):
Nice!

PR #2590, opened 2022-11-17
Output component.json file for documentation purposes

colebemis:
## Context

We're moving towards a unified IA for Primer docs: https://github.com/github/primer/issues/1307

## Problem

We cannot access metadata about our React components and props from outside this repository.

## Solution

We're moving the source of truth for component prop documentation out of `.mdx` files and into `.docs.json` files that live in the same directory as the component's source code. Example:

```
src/
  Button/
    Button.tsx
    Button.docs.json
```

This allows us to generate a `generated/component.json` file that we can use access outside of this repository to generate documentation.

### Changed

- For maintainers and contributors: You can move component prop documentation from MDX files (e.g. `docs/content/TreeView.mdx`) to a JSON files (e.g. `src/TreeView/TreeView.docs.json`). In the MDX files, use the `ComponentProps` to render the prop data from `.docs.json` files:

```diff

+ import data from '../../src/TreeView/TreeView.docs.json'

## Props

- ### TreeView
- <PropsTable>...</PropsTable>
- ### TreeView.Item
- <PropsTable>...</PropsTable>

+ <ComponentProps data={data} />
```


colebemis:
Looks like actions don't run on automated commits. Is there a way around that?

.github/workflows/ci.yml
```diff
@@ -189,3 +189,19 @@ jobs:
         with:
           name: avt
           path: .playwright/report
+
+  build-components-json:
```
joshblack, line 193: Would it satisfy requirements if this was uploaded as an artifact when something is merged into `main`?

colebemis: It's nice to see the `generated/component.json` diff in the same PR that made the changes. I don't think we get that if we only run this when merging to `main`, correct?

colebemis:
Going to close this PR for now because it was mostly a proof-of-concept, and I don't want it to get stale while I'm on vacation next week. I'll revisit this when we begin implementing the new [Primer docs IA](https://github.com/github/primer/issues/1415)

broccolinisoup:
👋🏼 @colebemis we paired with @jonrohan today for the code scanning of PRC (working towards https://github.com/github/primer/issues/1519) and we started scanning the PRC in the code-stats-repo (we npm installed @primer/react into that repo) mainly because to write the script first. The plan is using the same script in any repo to scan PRC.

We followed the same approach as primer_view_components which is having a list of components name and searching the names in a given repo. For that, we needed the list of the react components and we used `generated/component.json` . We manually copied the file to see if that would work, but would love to export that json and use it in the scan script. See the [PR](https://github.com/github/code-stats-collector/pull/29) here. 

Do you think the way you wrote the build script and generate the `components.json` is likely be the way for the docs initiative? If so, we can maybe use the script you wrote to generate the `components.json` for the code scanning as well?

colebemis:
@broccolinisoup Yup! We can use this `components.json` for collecting stats. I chatted with @jonrohan about that today. I will publish a canary version of @primer/react with `components.json` so y'all can test it out.

broccolinisoup:
> @broccolinisoup Yup! We can use this `components.json` for collecting stats. I chatted with @jonrohan about that today. I will publish a canary version of @primer/react with `components.json` so y'all can test it out.

Thanks! 😍

src/Button/Button.docs.json
```diff
@@ -0,0 +1,70 @@
+{
+  "id": "button",
+  "name": "Button",
+  "status": "alpha",
+  "a11yReviewed": false,
```
andrialexandrou, line 5: is it possible we can apply a logic that indicates when the component was last reviewed? if components will be changing often, which I hope they will be, we'll want to indicate how long it's been since we've approved it. I think!

broccolinisoup: 👋🏼 @andrialexandrou, sorry for late response! We have [this issue](https://github.com/github/accessibility/issues/789) in our backlog to track of the last a11y reviewed SHA of the components. Is this something similar what you were thinking? I don't think we have discussed what kind of automation logic we could implement for[this issue](https://github.com/github/accessibility/issues/789) yet. Do you think DataDog would be more convenient for accessibility team to see the accessibility data or do you think maybe we could have this data in multiple places?

andrialexandrou: this information would be useful for everyone if it could be surfaced in the Primer docs somehow. it's okay to defer on this just didn't know latest decisions.

broccolinisoup: No worries! We will catch up about how to solve this, hopefully soon!

broccolinisoup:
Ah so happy that this is merged 💖  I'll polish up [my open PR](https://github.com/github/code-stats-collector/pull/42) that uses the generated component.json for the PRC tracking this week ✨

PR #2591, opened 2022-11-17
Add `useRefObjectAsForwardedRef` to public hooks

iansan5653:
Adds the `useRefObjectAsForwardedRef` hook to the publicly exported hooks. This is a generic and useful hook that we can use in the Memex application and elsewhere.


colebemis (approved):
🚀

joshblack:
Trying to dig into usage of this in PRC, is this kind of acting like a `useMergedRefs` type hook or is there something distinct going on here with the `useImperativeHandle`?

It seems like this kind of acts like a superset in that it doesn't have to be an HTML element where hooks like a `useMergedRefs` would expect only HTML elements?

PR #2593, opened 2022-11-17
Make resizable SplitPageLayout.Pane keyboard accessible.

### LINKED ISSUE #2522: Resizable SplitPageLayout.Pane is not keyboard accessible
### Description

The resizer on the SplitPageLayout.Pane should be focusable and able to be adjusted using the arrow keys.

### Steps to reproduce

1. Go to https://primer.style/react/SplitPageLayout
2. Try to focus the resizer in the resizable example using tab
3. See that the resizer is not in the tab order and cannot be change with the keyboard.

### Version

v35.14.0

### Browser

Chrome

radglob:
Added a `tabIndex` prop and `onKeyDown` event handler to `PageLayout.Pane` (which `SplitPageLayout.Pane` uses) to allow keyboard controls to change the pane's size. Only handles `ArrowRight` and `ArrowLeft` keys.

Also added a text input that is only visible to screen readers that allows users to change the width of the pane from the minimum percentage to the maximum percentage it can take up within its parent. This is still visible to a screen reader even on narrow views where the pane width is not resizable.

Closes #2522

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


colebemis:
This looks like a great start! Tagging @ericwbailey for accessibility guidance on how this interaction should work.

ericwbailey:
Hey @radglob, it's looking like the resize control isn't announced or operable via assistive technology (tested with VoiceOver and NVDA). 

Unfortunately, only a `tabindex` declaration on a `div` is not sufficient. The resizing control needs to have an accessible name and role for it to be understood, or be controlled via visually hidden (`sr-only`) controls that have an accessible name and role.

For example, we could tie the divider to something like a range input, which would provide an accessible name, role, and preset range of values a user could control:

``` html
<div class="sr-only">
  <label for="uuid-panel-widths">
    Panel widths
  </label>
  <input 
    id="uuid-panel-widths" 
    name="range" 
    min="0" 
    max="100" 
    value="50" 
    step="10"
    type="range" />
```

A `value` of `50` here would set it at 50% width distribution, and `step="10"` lessens the amount of keypresses a user would need to make to adjust things in either direction.

We might also want to get more explicit, to both clarify what is getting resized and lessen the amount of interaction needed:

``` html
<div class="sr-only">
  <label for="uuid-width-pane">
    Pane width
  </label>
  <input 
    id="uuid-width-pane" 
    name="telephone" 
    autocorrect="off" 
    autocomplete="none" 
    type="tel"
    value="50">

  <label for="uuid-width-content">
    Content width
  </label>
  <input 
    id="uuid-width-content" 
    name="telephone" 
    autocorrect="off" 
    autocomplete="none" 
    type="tel"
    value="50">
</div>
```

For this setup, we'd have some JS to calculate 100 minus the last submitted value and apply the result to the input not modified. So, setting the content's width to `80` would automatically set the panel's width to `20`.

I think it would also be helpful to have more descriptive names than "panel" and "content", whose jargon might not be apparent to the end user. Ideally we'd describe resizing the width of what each pane contains.

ericwbailey:
Unsure if you've seen it or not, but I have [some additional context in this comment](https://github.com/github/primer/issues/581#issuecomment-1284357318).

joshblack:
@ericwbailey that scenario you described with the slider definitely matched what my expectations were. What would you think of the experience where we:

- Use the existing `div` as the visual appearance of the rule between the pane and content
- Place the `sr-only` content for slider within it
- Use `:focus-within` styling on the `div` to give it the appearance of a focus outline

This could allow the focus styles for the interaction while providing an accessible name and making sure that the control appears in-context when focused.

Curious what you think or if there is a better alternative flow for this kind of resize case 👀

ericwbailey:
@joshblack Two thoughts!

1. I like that idea a lot, but also
2. We wound up going with a text input that uses a percentage value.

My reasoning was that a slider control might require a lot of repetitive manual keypresses to configure the pane into its desired width, while a text input would be a little less laborious. I usually try to include a text input when a slider control is used in general, so it seemed like it'd be better to just consolidate.

joshblack:
@ericwbailey sorry for the delay, and thanks for adding in context 🙏 

With this approach, once thing I'm wondering about is the intended flow for the keyboard-only use-case. It seems this approach  ends up adding two additional tab stops (the `<input>` and the submission `<button>`) but won't provide a visual indicator due to the `sr-only` styles.

ericwbailey:
@joshblack Great question! I've [asked in Slack](https://github.slack.com/archives/C01UWR2MEUB/p1670430365991479) if we have an existing treatment for this sort of scenario, otherwise I'm thinking of modeling something along the lines of LinkedIn's initial skip nav treatment:

> Video content: A menu appearing at the top of LinkedIn when tab is pressed for the first time. Navigation options for skipping to search, then main content, and then closing the menu are present, and navigated through via tab, with a visible focus state.

[Kapture 2022-12-07 at 13.42.49.webm](https://user-images.githubusercontent.com/634191/206268767-c5aa5d0d-556b-4e5f-8a7f-7f78977f94b1.webm)

I also think that given that we don't have one standard approach yet that it might make sense to revisit visually showing this later, with more consensus on how to do it. Does that sound good to you?

radglob:
@ericwbailey per the conversion in Slack, should I remove the `tabindex`-based functionality and just support the percentage input?

Also, I'm somewhat confused by the skip link example you linked. Should we show a skiplink to this input? And if so, should the input then be visible to all users, not just those using screen readers?

ericwbailey:
>@ericwbailey per the conversion in Slack, should I remove the tabindex-based functionality and just support the percentage input?

Yep, we can remove the `tabindex` portion and use the input in its place.

>Also, I'm somewhat confused by the skip link example you linked. Should we show a skiplink to this input? And if so, should the input then be visible to all users, not just those using screen readers?

Sorry about that. We don't need to provide a skiplink to the input, I was more referring to the visually hidden until focused treatment LinkedIn uses. 

Since we don't have consensus how how something like that should look or behave across Primer yet, I'm holding off on adding it here so there's not wasted effort. I'm definitely planning on providing guidance and code if and when that does happen, though, so this one is not on you 🙂

radglob:
Alright. That code path is removed.

joshblack:
@ericwbailey definitely +1 to the visible on focus treatment from LinkedIn, totally agreed 👍 

> I also think that given that we don't have one standard approach yet that it might make sense to revisit visually showing this later, with more consensus on how to do it. Does that sound good to you?

Just wanted to double-check, it seems like the current implementation could be considered an SC 2.4.7 violation since it would not be visible on focus. I feel like I'm missing something with this rule or the exception here and just wanted to check-in to better understand on my end.

The current approach seems to work for the screen reader path, but the keyboard-only path does not seem to be addressed through this approach. Is that be something that we're looking to add in through follow-up iterations?

ericwbailey:
>Just wanted to double-check, it seems like the current implementation could be considered an SC 2.4.7 violation since it would not be visible on focus. I feel like I'm missing something with this rule or the exception here and just wanted to check-in to better understand on my end.

You're definitely right on the 2.4.7 call, thank you.

I was chatting with @langermank yesterday about a visual treatment for appears on focus content, as it seems like GitHub/Primer doesn't have a standard way of displaying this sort of content yet.

>Is there a flow that I'm missing for this or are we looking to add this in through follow-up iterations?

I need to chase down the best way forward for this, as it's representing a pretty sizable change in this task's scope. I also don't think it's on @radglob to own those changes.

radglob:
@ericwbailey What should we do for next steps? This PR has been open for a while, and it sounds like there's a lot that needs to change here.

src/PageLayout/PageLayout.tsx
```diff
@@ -670,20 +723,44 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
...
+                    name="pane-width"
+                    value={widthPercent}
+                    autoCorrect="off"
+                    autoComplete="off"
+                    type="tel"
+                    onChange={event => {
+                      setWidthPercent(event.target.value)
+                    }}
+                  />
+                </label>
```
ericwbailey, line 759:
```suggestion
                <label htmlFor={`${paneId}-width-input`}>
                  Pane width
                </label>
                <p id={`${paneId}-input-hint`}>
                  Use a value between {minPercent}% and {maxPercent}%
                </p>
                  <input
                    id={`${paneId}-width-input`}
                    aria-describedby={`${paneId}-input-hint`}
                    name="pane-width"
                    inputmode="numeric"
                    pattern="[0-9]*"
                    value={widthPercent}
                    autoCorrect="off"
                    autoComplete="off"
                    type="text"
                    onChange={event => {
                      setWidthPercent(event.target.value)
                    }}
                  />
```

Slight tweaks here, and this one is on me. We're:

1. Moving the input hint to after the `label`, so the announcement order is the label, and then a clue about how to operate it.
2. Un-nesting the `input` from its corresponding `label`, for some voice control software corner cases. The programatic association is still present via the `for`/`id` pairing.
3. Replacing `type="tel"` with [this guidance from GOV.UK](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/)

ericwbailey (changes requested):
@radglob One little tweak and I think we're good to merge.

Apologies that it took some time to get back to you. The path we're going to take is to ship incrementally, with the larger issue of content on focus in Primer being brought up in https://github.com/primer/design/issues/338/.

ericwbailey (approved):
🙌

colebemis (approved):
Excellent work on this @radglob @ericwbailey! ✨

PR #2613, opened 2022-11-24
Fix assign_release_conductor.yml

siddharthkp:
- [x] Add branch check
- [x] Get pagerduty schedule instead of on-call
- [x] Add assignee
- [x] Add reviewer


siddharthkp:
Self reviewing because it's a safe scoped change :)

PR #2630, opened 2022-11-30
test(vrt): add TreeView vrt and aat tests

joshblack:
#### Changelog

**New**

- Scaffold out TreeView tests for VRT and AAT

**Changed**

**Removed**


colebemis (approved):
Love it 💖

PR #2631, opened 2022-11-30
test(e2e): add vrt and axe tests for UnderlineNav

joshblack:
#### Changelog

**New**

- Add `UnderlineNav.test.ts` for AAT and VRT tests
- Add screenshots for VRT

**Changed**

- Update generate script for e2e tests

**Removed**


broccolinisoup:
Thanks for walking me through the snapshot update @joshblack , super cool 🥳 

Rethinking about what stories to snapshot, for static stories, I feel like "Overflow Template" and "Counters Loading State" should be enough? Because Overflow Template really covers all other cases. If, for example, colour contrast fails on "With Icons", it will fail on "Overflow Template" too unless there is something broken on the "With Icon" story itself. Or am I too simplifying this? Would there be cases that could cause a regression on "With Icons" but not "Overflow Template" ? 🤔 

Another thing is "examples".  I just checked the colour contrast on the actual dotcom pull request page and both (dotcom and story) uses same colour. So we could disable some stories as you mentioned, but I am also thinking because intention of examples are how the component behaves when it is used with other Primer components (as far as I understand) and because UnderlineNav doesn't really interact with any of the components on the page, just a piece of the composition, I wonder if we really need to snapshot examples?? I am really thinking from VRT for UnderlineNav. If we need/should also worry about VRT of the all stories, yes we should include them too.

I feel like my thinking is too simplified so there are probably things I am missing here 😬 Let me know! 

And thanks again, this is super cool!

broccolinisoup (commented):
Thanks for walking me through the snapshot update @joshblack , super cool 🥳

Rethinking about what stories to snapshot, for static stories, I feel like "Overflow Template" and "Counters Loading State" should be enough? Because Overflow Template really covers all other cases. If, for example, colour contrast fails on "With Icons", it will fail on "Overflow Template" too unless there is something broken on the "With Icon" story itself. Or am I too simplifying this? Would there be cases that could cause a regression on "With Icons" but not "Overflow Template" ? 🤔

Another thing is "examples". I just checked the colour contrast on the actual dotcom pull request page and both (dotcom and story) uses same colour. So we could disable some stories as you mentioned, but I am also thinking because intention of examples are how the component behaves when it is used with other Primer components (as far as I understand) and because UnderlineNav doesn't really interact with any of the components on the page, just a piece of the composition, I wonder if we really need to snapshot examples?? I am really thinking from VRT for UnderlineNav. If we need/should also worry about VRT of the all stories, yes we should include them too.

I feel like my thinking is too simplified so there are probably things I am missing here 😬 Let me know!

And thanks again, this is super cool!

broccolinisoup (commented):
Only 3 failing now, because of the status label. Let me know how you would like to go with them, either disable or remove the examples stories from snapshotting. 🙌🏼

broccolinisoup:
> @broccolinisoup thanks for doing that! For the 3 remaining, I ended up disabling the theme check for them since it's a known issue 👍
> 
> > Rethinking about what stories to snapshot, for static stories, I feel like "Overflow Template" and "Counters Loading State" should be enough?
> 
> Sounds good! I think if you feel like they cover everything then we should be good to go 🔥

Should we remove the other stories then? I am a bit indecisive 😬 With a fresh mind now in the morning, I am thinking that although, I still _do_ think "Overflow Template" and "Counters Loading State" should be enough, as we are still trying it out what stories we want to snapshot, maybe we can merge this as is and see how that works, see if the stories that I think not necessary will catch any regression?

broccolinisoup (approved):
🚀

joshblack:
@broccolinisoup sounds good to me! Happy to try it out and see, feel free to remove any of them that you feel like are unhelpful!

PR #2635, opened 2022-12-01
fix(Radio): remove redundant aria-disabled

### LINKED ISSUE #2634: Disabling Radio FormControl in RadioGroup results in axe violation
### Description

To disable a certain `Radio` button in a `RadioGroup` of `FormControl`s, you need to set the `disabled` prop on the `FormControl` element of the radio option.  If you do, you get the following axe violation

```
helpers.ts:37 [Critical] [aria-disabled] may be used in place of native HTML [disabled] to allow tab-focus on an otherwise ignored element. Setting both attributes is contradictory.

https://www.w3.org/TR/html-aria/#docconformance-attr
```

When you inspect the element in the DOM it indeed has both `disabled` and `aria-disabled="true`" HTML attributes.

### Steps to reproduce

Render this JSX:

```
<RadioGroup name="options">
  <RadioGroup.Label>Options</RadioGroup.Label>
  <FormControl>
    <Radio value="option1" />
    <FormControl.Label>Option 1</FormControl.Label>
  </FormControl>
  <FormControl disabled>
    <Radio value="option2" />
    <FormControl.Label>Option 2</FormControl.Label>
  </FormControl>
</RadioGroup>
```

### Version

35.14.2

### Browser

Chrome

joshblack:
Closes #2634 

Removes redundant `aria-disabled` when the `<input>` is disabled.

> **Note**
> Waiting for a good way to repro this with axe so that we can make sure this doesn't regress in the future 👀

### DIFF
--- a/src/Radio.tsx
+++ b/src/Radio.tsx
@@ -76,7 +76,6 @@ const Radio = React.forwardRef<HTMLInputElement, RadioProps>(
         name={name}
         ref={ref}
         disabled={disabled}
-        aria-disabled={disabled ? 'true' : 'false'}
         checked={checked}
         aria-checked={checked ? 'true' : 'false'}
         required={required}


src/Radio.tsx
```diff
@@ -76,7 +76,6 @@ const Radio = React.forwardRef<HTMLInputElement, RadioProps>(
         name={name}
         ref={ref}
         disabled={disabled}
-        aria-disabled={disabled ? 'true' : 'false'}
```
broccolinisoup, line 79: Hey @joshblack! Sorry, I am a bit confused about why we removed `aria-disabled` . I understand that the GitHub axe rule of having both attribute are conflicting and it makes sense but [here](https://www.w3.org/TR/html-aria/#docconformance-attr) it says

> Unless otherwise stated, authors MAY use aria-* attributes in place of their HTML equivalents on HTML elements where the aria-* semantics would be expected. For example, authors MAY specify aria-disabled=true on a [button](https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element) element, while also implementing the necessary scripting to functionally disable the button, rather than the use disabled attribute.

I would expect us to remove the `disabled` and update styling accordingly? 🤔  I might be missing something, sorry but would be great to clarify this, then I'll continue to review 😊

joshblack: Hey @broccolinisoup! Great question, I definitely went with the default here to avoid having to reimplement built-in `disabled` behavior for `aria-disabled` and styles. What are you thinking would be the advantages in using `aria-disabled` over `disabled` here?

In the past, I feel like the big reason was that people wanted the control to be discoverable in order to share why it was disabled. I'm not sure if this would apply here like it would with a button but let me know if there is a scenario you're thinking of that we should make sure to include

joshblack: @broccolinisoup makes sense to me! Happy to make that change 👍 

Maybe what I can do is add this to our next major release ticket with the following changes in mind:

- `Radio` should no longer accept a `disabled` prop (this we would have to remove in the next major)
- `Radio` in the interim accepts `disabled` _or_ `aria-disabled` (but not both) and styling/behavior should apply in either scenario

Curious what you think 👀

broccolinisoup: Thanks @joshblack!

> `Radio` should no longer accept a `disabled` prop (this we would have to remove in the next major)

I really like that 🙌🏼  Can we also check it with the accessibility team to make sure if they don't suggest anything else or anything more?

>`Radio` in the interim accepts `disabled` or `aria-disabled` (but not both) and styling/behaviour should apply in either scenario

My gut feeling says we should allow `aria-disabled` not `disabled` but I recognise the effort around updating the styles to make sure the ones with `aria-disabled`  look disabled too. So maybe we can merge your PR as is now as the interim solution and make sure we have issue to address removing `disabled` in the next major realised? Does that sound good at all?

joshblack: > Can we also check it with the accessibility team to make sure if they don't suggest anything else or anything more?

Definitely! I'll reach out.

> My gut feeling says we should allow aria-disabled not disabled

Makes sense to me! I think the only catch is that we can't remove support for `disabled` without it being a breaking change. I think that's where some of the interim ideas were coming from (support either case but not both at the same time). Are you thinking of having `disabled` map to `aria-disabled` or were you wanting to remove it as part of addressing the parent issue?

broccolinisoup (approved):

PR #2640, opened 2022-12-02
Jdrush89/tree content visibility

jdrush89:
I tried implementing tree virtualization in a number of ways. First, I tried using intersection observers on the subtree contents and paging in rows as they approached the viewport. The problem with this approach is when the user scrolls quickly, they may end up paging in a large subdirectory which then takes time to render. The pop-in while scrolling was a degraded experience compared to the non-virtualized tree. The complexity added from trying to maintain typeahead and other keyboard shortcuts, and keeping the focused and current rows materialized was also extremely high and likely to be prone to bugs.
I also tried virtualizing each row's contents with intersection observers. The pop-in was much less severs, though the initial render took much longer because each row had to handle and intersection event to figure out whether to show or hide. I settled on using the ```content-visiblity: auto``` property, which delays the rendering of each row without removing it
from the DOM. Each row can pass in a containIntrinsicSize property to specify it's estimated height , then the content-visibilty property will be added as well. Doing this is very simple and won't affect accessibility in any way. In testing with large trees, the initial render time is 20% faster and any lagginess in updating hover styles seems to go away. The drawback is that this property is not supported in FF or Safari yet, though it should be aded at some point.

The main problems virtualization was trying to solve was around the initial render time, hover styles, and typeahead time. I realized the typeahead implementation was causing multiple full rerenders of the tree by keeping the current search term in state. By moving it to a ref instead, the typeahead time sped up drastically. content-visiblity helps the initial render time some. In the file tree on dotcom, we make sure the tree renders after the hpc, so its initial render time isn't extremely critical, we also only render the first 100 items of each directory in the first pass, then do the rest afterwards. This greatly speeds up the initial render and essentially accomplishes the gains virtualizing each subdirectory's contents would make.  Once SSR is in place, the initial render should be sped up drastically as well.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/TreeView/TreeView.tsx
```diff
@@ -86,7 +86,7 @@ const UlBox = styled.ul<SxProp>`
   .PRIVATE_TreeView-item {
     outline: none;
 
-    &:focus-visible > div {
+    &.focus-visible > div {
```
jdrush89, line 89: this is to fix focus not showing in safari

docs/content/TreeView.mdx
```diff
@@ -300,6 +300,11 @@ See [Storybook](https://primer.style/react/storybook?path=/story/components-tree
     type="boolean"
     description="The expanded state of the item when it is initially rendered. Use when you do not need to control the state."
   />
+  <PropsTableRow
+    name="containIntrinsicSize"
+    type="string"
+    description="The size of this item's contents. Passing this will set 'content-visiblity: auto' on the content container, delaying rendering until the item is in the viewport."
```
colebemis, line 306: Is there a downside to always setting `content-visibility: auto`? Since the consumer doesn't control the [height of TreeView.Item](https://github.com/primer/react/blob/main/src/TreeView/TreeView.tsx#L106), it feels strange to ask the consumer to pass a height value here.

src/TreeView/TreeView.tsx
colebemis: Could we do both?
```suggestion
    &:focus-visible > div,
    &.focus-visible > div  {
```

docs/content/TreeView.mdx
jdrush89: The consumer does control the height by passing in whatever contents they want. Only the min-height is set.

src/TreeView/TreeView.tsx
```diff
@@ -408,6 +419,9 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
             style={{
               // @ts-ignore CSS custom property
               '--level': level,
+              contentVisibility: containIntrinsicSize ? 'auto' : undefined,
+              // @ts-ignore CSS custom property
+              'contain-intrinsic-size': containIntrinsicSize,
```
joshblack, line 424:
```suggestion
              containIntrinsicSize,
```

Would it make sense for this to be `containIntrinsicSize` since we're setting an inline style? Otherwise I think we get the following error:

image

src/TreeView/useTypeahead.ts
```diff
@@ -18,6 +18,46 @@ export function useTypeahead({containerRef, onFocusChange}: TypeaheadOptions) {
...
+      // Don't change focus if the search value is empty
+      if (!searchValue) return
+
+      if (!containerRef.current) return
+      const container = containerRef.current
+
+      // Get focusable elements
+      const elements = Array.from(container.querySelectorAll('[role="treeitem"]'))
+        // Filter out collapsed items
+        .filter(element => !element.parentElement?.closest('[role=treeitem][aria-expanded=false]'))
```
joshblack, line 33:
```suggestion
      // Get focusable elements
      const elements = Array.from(container.querySelectorAll(':not([aria-expanded=false]) [role="treeitem"]'))
```

Would this alternative end up producing the same result as the original or would they be different sets? Just an idea if we wanted to do this in one pass

joshblack (commented):
Just left a couple of comments! In general, the changes to typeahead seem great 🔥 (thanks for doing that)

The content visibility stuff I feel less sure about, given some of the stuff you said that are coming up to help reduce render time (and lack of support in FF/Safari) do you think this is something that needs to be introduced into the component or is a nice to have?

src/TreeView/TreeView.tsx
jdrush89: updated!

src/TreeView/useTypeahead.ts
jdrush89: I don't think we need to filter at all. Rows whose parents are collapsed don't get rendered. I think they used to be at one point, but that was changed. It seems to behave correctly with the filter removed.

jdrush89:
@joshblack Thanks for taking a look! I think the content-visibility is worth using for the file tree since it's easy to add and increases perf for large trees for most users. For FF/Safari, it is just a no-op until they add support for it. If you didn't want to add it to the TreeView component in primer, we could just add css rules to the 'PRIVATE_TreeView-item-container' class in dotcom, assuming that classname doesn't change.

joshblack (approved):

PR #2652, opened 2022-12-05
Button alignment and size bug fixes

langermank:
**Note**: These are the **non breaking changes** from this [PR](https://github.com/primer/react/pull/2374) 

## Improvements

- Add a `block` prop for full width
- Add `alignContent` prop to align content to [center or start](https://github.com/github/primer/issues/939)
- Use `control` sizing CSS variable values (not using CSS vars just yet)
- Use `height` over padding for more control over sizing

## Visual updates

Addresses a few things missing from this discussion: https://github.com/github/primer/discussions/1016

- Slightly more condensed padding
- Improved `invisible` hover states (will update these colors when new functional colors are addressed)

Closes:
- https://github.com/primer/react/issues/1979
- https://github.com/github/primer/issues/939

### Screenshots

Please refer to the comprehensive Storybook stories- this is just a small sample of some of the changes

| before | after | description |
| -- | -- | -- |
| image | image | trailing action slot alignment |
| image | image | default padding, font-weight |
| image | image | large height |
| icon was blue | image | invisible svg color |
| image | image | trailing action locks right in full width |
| image | image | `invisible` buttons use default text color if visuals are present, and accent blue if no visuals. |

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2653, opened 2022-12-05
Update Button stories

langermank:
This updates Button stories to match out new formatting, but also sets up e2e tests for this upcoming PR: https://github.com/primer/react/pull/2652

This should only impact Storybook and e2e tests.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack (commented):
Quick question, do we also need to update the storybook URL in `Button.mdx` or will that end up in the same place? (Wasn't sure about the URL changes with the new stories)

langermank:
> Quick question, do we also need to update the storybook URL in `Button.mdx` or will that end up in the same place? (Wasn't sure about the URL changes with the new stories)

Ahh yuuuup! Totally forgot 😅 done! Thank you 🙌

joshblack (approved):

Issue #2411, opened 2022-10-06
The aria-keyshortcuts is wrongly attached for the ActionList.LinkItem components


jchuerva:
From https://github.com/github/repos/issues/3034

It looks like for `ActionList.LinkItem` component, the `aria-keyshortcuts` is attached the the `<a>` instead of `<li>` which is incorrect.

![image](https://user-images.githubusercontent.com/5942864/194352716-0fe7c774-1651-4470-b5f8-1326b2f1b088.png)

lesliecdubs:
Thanks for reporting this bug, @jchuerva! This looks like it'll require a slightly more complex fix so we're moving it to our backlog for now but hope to revisit soon.

siddharthkp:
Leaving here for who would work on this: After moving aria-keyshortcuts, need to make sure the link can still be opened with keyboard

lesliecdubs:
Good questions @green6erry! If you'd like an accessibility expert to weigh in on them, I'd suggest asking about this over in #accessibility on Slack or signing up for the next [Accessibility Eng+ Office Hours](https://github.com/github/accessibility/issues?q=is%3Aissue+is%3Aopen+label%3Ameeting+office+hours) for more hands-on support and advice. 

This [helpful accessibility guide](https://github.com/github/accessibility#im-looking-for-help) may come in handy for your work!

tallys:
hey @siddharthkp, checking in on this

PR #2657, opened 2022-12-06
2411 the aria keyshortcuts is wrongly attached for the actionlistlinkitem components

green6erry:
Closes [2411](https://github.com/primer/react/issues/2411)

Removes this disparity of `aria-keyshortcuts` value, which when from mnemonics, was being assigned to the `<li>` wrapper of LinkItem, whereas if it was `aria-keyshortcuts` was manually defined, it would assign to the inner `<a>` element (creating two keyboard shortcuts that looked like they were selecting the same element, but were not). This also applies the correct prop `role="menuitem"` to the `<a>` element instead of the `<li>` in Link Item.
### Screenshots

image
Previously, in this above example, pressing "u" would make the wrapper `<li>` active, and "d" would make the inner `<a>` active. Now, "u" doesn't do anything as the mnemonic keyshortcut is overridden by the user defined "d".

#### Updated

- Updated LinkItem to use either mnemonic or user defined `aria-keyshortcuts` to select the `<a>` tag within
- Updated LinkItem so that the `role="menuitem"` is being assigned to correctly selectable DOM element inner `<a>` instead of the wrapper `<li>`

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


joshblack (approved):
Looks great! 🥳

joshblack:
cc @siddharthkp would love to hear any thoughts you might have on this approach for handling the `aria-keyshortcuts` on `ActionList.LinkItem` from the parent item. It seems like this should work but wasn't sure the full scope of how this component is used.

PR #2659, opened 2022-12-06
Refine 'as' prop type in button component and fix name of aria type

kendallgassner:
Describe your changes here.

- `aria-labelby` -> `aria-labelledby` (aria-labelby is not an aria attribute)
- restrict the button 'as' prop at runtime to be either a semantic `<a>` or `<button>`

I introduced the library [react-merge-refs](https://www.npmjs.com/package/react-merge-refs) to do this but please let me know if there is a better way 😸  

### Open Questions

Should this be a patch? Since its a new warning message?

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack (commented):
Looks great! Thanks for doing this 🙏 

Just left one comment, I think it might be the convention in PRC but I'm not 100% sure 😅

kendallgassner:
@joshblack 
> Just left one comment, I think it might be the convention in PRC but I'm not 100% sure 😅

Did you submit the comment? 😅 I am not seeing it on my end.

Also, would this be a patch?

src/Button/ButtonBase.tsx
```diff
@@ -18,6 +19,7 @@ const trailingIconStyles = {
 const ButtonBase = forwardRef<HTMLElement, ButtonProps>(
   ({children, as: Component = 'button', sx: sxProp = defaultSxProp, ...props}, forwardedRef): JSX.Element => {
     const {leadingIcon: LeadingIcon, trailingIcon: TrailingIcon, variant = 'default', size = 'medium', ...rest} = props
+    const innerRef = React.useRef<HTMLElement>()
```
joshblack, line 22:
```suggestion
    const innerRef = React.useRef<HTMLElement>()
    useRefObjectAsForwardedRef(forwardedRef, innerRef)
```

Here was that convention that I was talking about! This should be available in `src/hooks/useRefObjectAsForwardedRef.ts`. This seems to show up in a couple components that both have a local and forwardedRef

joshblack:
@kendallgassner sorry about that! Just left a comment above 👍 And good question about `patch` versus `minor`, I think regardless it will be included in the next `minor` so what you have currently in the changeset should totally work.

src/Button/ButtonBase.tsx
kendallgassner: AMAZING thank you!

src/Button/ButtonBase.tsx
```diff
@@ -1,10 +1,11 @@
-import React, {ComponentPropsWithRef, forwardRef, useMemo} from 'react'
+import React, {ComponentPropsWithRef, ForwardedRef, forwardRef, useMemo} from 'react'
```
kendallgassner, line 1:
```suggestion
import React, {ComponentPropsWithRef, forwardRef, useMemo} from 'react'
```

src/Button/ButtonBase.tsx
```diff
@@ -26,8 +30,15 @@ const ButtonBase = forwardRef<HTMLElement, ButtonProps>(
       return merge(baseStyles, sxProp as SxProp)
     }, [baseStyles, sxProp])
 
+    React.useEffect(() => {
+      if (!(innerRef.current instanceof HTMLButtonElement) && !(innerRef.current instanceof HTMLAnchorElement)) {
+        // eslint-disable-next-line no-console
+        console.warn('This component should be an instanceof a semantic button or anchor')
```
joshblack, line 36: I think one think we'll need to add is an `if (__DEV__)` check to this so that it only is emitted in development. 

```suggestion
        if (__DEV__) {
          console.warn('This component should be an instanceof a semantic button or anchor')
        }
```

Hope that makes sense!

joshblack (changes requested):

joshblack (approved):
Thanks for doing this!

Issue #2536, opened 2022-11-04
React 18 Support


joshblack:
This issue is an umbrella for all issues associated with internally updating PRC to use React 18.

> **Note**
> Teams may use `@primer/react` with React 18, this issue specifically tracks internal changes and updates that need to be made.

- [ ] Docs
  - [ ] Update `react` and `react-dom` to React 18
  - [ ] Optionally, update gatsby to v4 and doctocat theme
- [ ] Storybook
  - [ ] The following dependencies cannot be updated as they have no published version supporting React 18. We'll need to decide if it's okay to remove them and if there are replacements
    - [ ] `@whitespace/storybook-addon-html@5.0.0` -> cannot be updated
    - [ ] `storybook-addon-performance@0.16.1` -> cannot be updated
- [ ] `@primer/react`
  - [ ] Dependencies
    - [ ] Update `react` and `react-dom` to React 18
    - [ ] Update the following dependencies to meet the new peer range for React 18
      - [ ] `@react-aria/ssr@3.1.0`
    - [ ] Address the following dependencies with conflicting peer dependencies
      - [ ] `peer react@"^16.13.1 || ^17.0.0" from @mdx-js/react@1.6.22`
      - [ ] `peer react@"^16.8.4 || ^17.0.0" from react-inspector@5.1.1`
      - [ ] `peer react@"^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1" from react-element-to-jsx-string@14.3.4`
      - [ ] `peer react-dom@"^0.14.8 || ^15.0.1 || ^16.0.0 || ^17.0.1" from react-element-to-jsx-string@14.3.4`
  - [ ] Testing
    - [ ] Update `@testing-library/react` to latest to use `createRoot` API
    - [ ] Remove `enzyme` and associated React 17 adapter (optionally instead unsupported 18 adapter)
    - [ ] Remove usage of `enzyme` APIs and assertions from the following test suites
      - [ ] `src/drafts/MarkdownEditor/MarkdownEditor.test.tsx`
      - [ ] `src/__tests__/TextInput.test.tsx`
      - [ ] `src/__tests__/Token.test.tsx`
      - [ ] `src/__tests__/SideNav.test.tsx`
      - [ ] `src/__tests__/Details.test.tsx`
      - [ ] `src/__tests__/UnderlineNav.test.tsx`
      - [ ] `src/__tests__/SelectMenu.test.tsx`
[...]

PR #2661, opened 2022-12-06
feat(project): update project to use React 18

joshblack:
Related: https://github.com/primer/react/issues/2536

This PR is a rollup of the following PRs to update the project to use React 18 in development:

- https://github.com/primer/react/pull/2673
- https://github.com/primer/react/pull/2674
- https://github.com/primer/react/pull/2675

The strategy used is to test both React 17 and React 18. In the codebase, we also need to check for the presence of React 18 features before using them in order to maintain backwards compatability.


script/set-react-version.js
```diff
@@ -0,0 +1,88 @@
...
+          name: '@testing-library/react',
+          version: '12.1.5',
+        },
+      ],
+    },
+  ],
+  [
+    '18',
+    {
+      devDependencies: [
```
broccolinisoup, line 34: Curios why we specifically chose these dev dependencies to list here? 🤔 Are they the only ones that are effected by the React version upgrade?

broccolinisoup (approved):
Looks good to me, just left a question.

script/set-react-version.js
joshblack: @broccolinisoup exactly, I think any other deps that we want to include should be ones that mirror support for React 17/18 otherwise we can't really ship a version of PRC that works with both versions

colebemis (approved):

PR #2662, opened 2022-12-07
PageHeader: CSS media queries for responsive sub component visibility

broccolinisoup:
Closes https://github.com/github/primer/issues/1569

Following up a https://github.com/primer/react/pull/2265#discussion_r1036355747 for shifting the responsiveness logic from JS to CSS via using media queries to make the component more SSR friendly.

Context:

Currently PageHeader uses `useResponsiveHook` to manage sub-components' visibility via hidden prop. It works fine however, this would cause an extra work for us in SSR. Instead of using the useResponsiveHook we can try managing the visibility of the subcomponents with CSS breakpoints.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/PageHeader/manageResponsiveValueWithCSS.ts
```diff
@@ -0,0 +1,53 @@
...
+ *     "display": "flex"
+ *   }
+ * }
+ */
+export function CSSManagedResponsiveValue<T, F, B, C>(
+  value: T,
+  fallback: F,
+  cssProperty: string,
+  truety: B,
+  falsy: C,
```
joshblack, line 32:
```suggestion
  truthy: B,
  falsey: C,
```

Small nit for the truth-y and false-y names. Another idea for names could be `enabled` and `disabled` for if the responsive value is enabled or disabled at that breakpoint

src/PageHeader/manageResponsiveValueWithCSS.ts
```diff
@@ -0,0 +1,53 @@
...
+): BetterSystemStyleObject {
+  if (isResponsiveValue(value)) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const responsiveValue = value as Extract<T, ResponsiveValue<any>>
+    const checkFallback = fallback ? truety : falsy
+
+    return {
```
joshblack, line 39: One thing that seems tricky about this is that it's always generating media queries even if we may not need them. For example,

```js
const hiddenOnRegularAndWide = {
  narrow: false,
  regular: true,
  wide: true,
}
```

For this object, we would only need two breakpoints, `narrow` and `regular`, since wide would already be covered by `regular`.

Curious if there would be any way to simplify this, update the default values, or if it's not worth going through that process at the moment

src/PageHeader/PageHeader.tsx
```diff
@@ -57,19 +58,17 @@ const Root: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({children, sx
...
   sx = {},
 }) => {
-  const isHidden = useResponsiveValue(hidden, false)
   const contentNavStyles = {
-    display: isHidden ? 'none' : 'flex',
     flexDirection: 'row',
     alignItems: 'center',
     gap: '0.5rem',
     order: REGION_ORDER.ContextArea,
+    ...CSSManagedResponsiveValue(hidden, false, 'display', 'none', 'flex'),
```
joshblack, line 71: Is the alternative to this helper that the author would need to write something like:

```js
const styles = {};

if (typeof hidden === 'boolean') {
  styles.display = hidden ? 'none' : 'flex';
}

if (typeof hidden === 'object') {
  if (hidden.narrow) {
    styles[breakpoints.narrow] = {
      display: hidden.narrow ? 'none' : 'flex',
    };
  }

  if (hidden.regular) {
    styles[breakpoints.regular] = {
      display: hidden.regular ? 'none' : 'flex',
    };
  }

  if (hidden.wide) {
    styles[breakpoints.wide] = {
      display: hidden.wide ? 'none' : 'flex',
    };
  }
}
```

joshblack (commented):
Love the shift over into media queries 🔥 

I feel like the biggest thing to wrap your head around in this is `CSSManagedResponsiveValue` since it does some heavy lifting to get the breakpoints right. It took me a bit to follow what exactly it needed to do but definitely see how helpful it is afterwards.

One idea to that could simplify it could be to move the fallback outside of this function and inline it into the `sx` prop directly. Then it's just responsive for adding `none` or `flex` into the corresponding breakpoints.

Either way, love what you've done!

src/PageHeader/manageResponsiveValueWithCSS.ts
broccolinisoup: Sorry! 🙈 I changed the implementation a little bit (I explained it as a response here), so we don't need these anymore.

src/PageHeader/manageResponsiveValueWithCSS.ts
broccolinisoup: This is a great feedback @joshblack! Thank you so much. I addressed this in my new implementation, would love to hear what you think 🙌🏼

src/PageHeader/PageHeader.tsx
broccolinisoup: Yes! with the previous version it was exactly this. With the new implementation, it is slightly different but still trying to lift the heavy weight of writing a bunch of condition and media queries.

broccolinisoup:
> Love the shift over into media queries 🔥
> 
> I feel like the biggest thing to wrap your head around in this is `CSSManagedResponsiveValue` since it does some heavy lifting to get the breakpoints right. It took me a bit to follow what exactly it needed to do but definitely see how helpful it is afterwards.
> 
> One idea to that could simplify it could be to move the fallback outside of this function and inline it into the `sx` prop directly. Then it's just responsive for adding `none` or `flex` into the corresponding breakpoints.
> 
> Either way, love what you've done!

Thank you 🙌🏼 I really love your idea and it also encourages consumers to make more intentional decisions I think rather than providing a value as default that might be hard to grasp at the time where it is exactly applied due to the abstraction of the `CSSManagedResponsiveValue` function. I addressed this on the new implementation. looking forward to hearing what you think!

broccolinisoup:
@joshblack I had some feedback from @colebemis as well in our 1:1 and this is why I changed the function a bit. The feedback/questions I got from Cole was

- How does this function work for responsive values that they are not boolean?
For example, `divider` styles on PageLayout, I added this in the code as [an example](https://github.com/primer/react/pull/2662/files#diff-7835f786e205175ad2625741b6c0ec3ed9e0faa3e8093abf76f1b70deb212b37R56). Mainly the function was only designed to handle boolean values before and it couldn't handle cases like this. Instead of limiting the function to only compare the given values with ternary operator (value ? truthy : falsey) we made this comparison as a param to the func to make it a bit more flexible. Looking forward to both of your feedback on this 👀 

- How would that work if there are multiple responsive values that needs to be rendered in the same component? Like what what would happen if there are multiple CSS media queries in the styles?
Cole recommended to use the `merge` function and seems to be working perfectly ✨ I intentionally left the commented out code [here](https://github.com/primer/react/pull/2662/files#diff-c6588745badb7cb5f4a18e6153b86361da10ad83805f0d12504968e0c3fbf393R66) as an example but will remove it once I find out where the best place would be to write a bit of a documentation for this function? 🤔 As Josh also mentioned, it takes a bit to get your head around the function and how to use it but I do believe its value 🙏🏼 

Looking forward to both of your feedback!

joshblack:
@broccolinisoup makes a lot of sense to me! Especially with the added flexibility around the value.

The way you structured it, it seems like there are a couple steps to the transformation of `ResponsiveValue` to an `SxProp`:

1. Normalize the responsive value (turn it from a boolean to a responsive value, if needed)

    <details>
      <summary>Code</summary>

      ```ts   
      // Normalize the given value into a ResponsiveValue shape
      function normalize(value) {
        if (typeof value === "object") {
          return value
        }
      
        if (value === true) {
          return {
            narrow: true,
            regular: true,
            wide: true
          };
        }
      
        return {
          narrow: false,
          regular: false,
          wide: false
        };
      }
      ```
    </details>

2. Map the responsive value to CSS properties (what should the "value" part of a ResponsiveValue map to in a CSS property)

    <details>
      <summary>Code</summary>

      ```ts
      // Use a `mapFn` to map the values of the ResponsiveValue object to CSS property values
      function getPropertyValues(responsiveValue, mapFn) {
        return Object.entries(responsiveValue).map(([breakpoint, value]) => {
          return [breakpoint, mapFn(value)];
        });
      }
      ```
    </details>

3. Map the result of (2) to breakpoints (how do we convert a responsive value to a CSS declaration with a property in a breakpoint)

    <details>
      <summary>Code</summary>

      ```ts
      // Get the breakpoints from the mapped ResponsiveValue object for a specific property
      function getBreakpoints(value, property) {
        const styles = {};
      
        for (const [breakpoint, value] of values) {
          styles[breakpoints[breakpoint]] = {
           [property]: value
          };
        }
      
        return styles;
      }
      ```
    </details>

I feel like the combination of these is what becomes the helper that you have made. When thinking about the names for this, it felt like `getBreakpointValues` could be a way to pair up with `ResponsiveValue` where you would call it like the signature that you have currently:

```ts
getBreakpointValues(hidden, 'display', (isHidden) => isHidden ? 'none' : 'flex');
```

**Note**: one detail I love about your implementation is that the CSS property and value live side-by-side now and it makes reading what the value should be very intuitive to me.

And under the hood it becomes the transform pipeline from above:

```ts
function getBreakpointValues(responsiveValue, property, mapFn) {
  const value = normalize(responsiveValue);
  const propertyValues = getPropertyValues(value, mapFn);
  return getBreakpoints(propertyValues, property);
}
```

src/PageHeader/manageResponsiveValueWithCSS.ts
```diff
@@ -0,0 +1,130 @@
...
+ * {
+ *  "@media screen and (max-width: 768px)": {
+ *     "height": 8
+ *   }
+ *   "@media screen and (min-width: 768px)": {
+ *     "height": 1
+ *   },
+ * }
+ */
+export function CSSManagedResponsiveValue<T>(
```
colebemis, line 72: I'm not sure about the name `CSSManagedResponsiveValue`. It seems a little long and unclear. The word "value" feels misleading because it actually returns an object with the property and the value (not just the value). How do you feel about calling it something like `responsiveCssProperty`?

src/PageHeader/manageResponsiveValueWithCSS.ts
```diff
@@ -0,0 +1,130 @@
...
+  value: T,
+  cssProperty: string,
+  comparisonFunc: (value: T) => void,
```
colebemis, line 75: A couple of notes:
- `comparisonFunc` feels like the wrong name since this function is transforming one value into another value, not comparing two values. I think `transformFn` or `mapFn` would be a more appropriate name.
- We use the return value of this function, so it's return type shouldn't be `void`. I think the correct type signature might look something like this:

```js
import * as CSS from "csstype";

function responsiveCssProperty<TInput, TOutput>(
  cssProperty: keyof CSS.Properties,
  mapFn: (value: TInput) => TOutput,
  value: TInput | ResponsiveValue<TInput>
) { ...
```

src/PageHeader/manageResponsiveValueWithCSS.ts
```diff
@@ -0,0 +1,130 @@
...
+): BetterSystemStyleObject {
+  if (isResponsiveValue(value)) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const responsiveValue = value as Extract<T, ResponsiveValue<any>>
+
+    // Build media queries with the giving cssProperty and comparisonFunc
+    const narrowMediaQuery =
+      'narrow' in responsiveValue
+        ? {
+            [`@media screen and (max-width: 768px)`]: {
```
colebemis, line 85: Can you import these media query ranges from useResponsiveValue?

https://github.com/primer/react/blob/main/src/hooks/useResponsiveValue.ts#L9-L13


```js
import {viewportRanges} from '../utils/useResponsiveValue'

...

[`@media screen and ${viewportRanges.narrow}`]: ...
```

src/PageHeader/manageResponsiveValueWithCSS.ts
```diff
@@ -0,0 +1,130 @@
...
+        ...narrowMediaQuery,
+        ...regularMediaQuery,
+      }
+    } else {
+      return {
+        ...narrowMediaQuery,
+        ...regularMediaQuery,
+        ...wideMediaQuery,
+      }
+    }
```
colebemis, line 125: How might we handle fallback/default values in this function?

For example, if a value for the `narrow` viewport is not defined in the responsive value object (`{ regular: true, wide: false }`), how do we specify what value the property should fallback too in the narrow viewport range?

Let me know if that makes sense.

src/PageHeader/manageResponsiveValueWithCSS.ts
broccolinisoup: Yes, I agree the name is unclear and too long! Josh also suggested `getBreakpointValues` and your comment about `value` makes so much sense as well. I love both of your suggestions! Maybe using `CSSProperty` is a bit misleading as well because it doesn't mention about the value?  Let me try to combine both of your feedback. 

`CSSProperty + Values = CSS Declarations`

I feel like `responsive` and `breakpoints` could be interchangeable. 
if we want to be more specific about what we return maybe we could use `getBreakpointDeclarations`?
If we want to be more consistent with the hook we could use `responsiveCSSDeclarations`?

If it hard! 😬 Let me know what you think!

src/PageHeader/manageResponsiveValueWithCSS.ts
broccolinisoup: Definitely love the `mapFn` 💖  Josh also suggested that. And sorry about the types, I completely forgot to update them. Love the way you typed them 💖

src/PageHeader/manageResponsiveValueWithCSS.ts
broccolinisoup: Ah great idea!

src/PageHeader/manageResponsiveValueWithCSS.ts
broccolinisoup: It does! And thats for commenting about that. 

I took the fallback value out of the function and made it to be declared with the CSS property within the `sx` prop inside. Josh suggested to that (see the convo [here](https://github.com/primer/react/pull/2662#issuecomment-1350383980)) and it made sense to me because it feels more imperative and an intentional choice than providing a single value.  Let me give you an example to explain it better. 

Let's take this hidden prop as an example
```
hidden={{regular:true, wide: false}}
```
If we provide the fallback value in the function it could be like this
```
CSSManagedResponsiveValue(hidden, 'display', value => {
      return value ? 'none' : 'flex'
}, false)
```
and this will resolve into

```
{
    [@media screen and (max-width: 768px)]: {
       display :'none'  // fallback
    },
    [@media screen and (min-width: 768px)]: {
       display : 'none'
    },
    [@media screen and (max-width: 768px)]: {
       display : 'flex'
    },
}
```

With the current version where we took the fallback value out of the func, now the developer (who uses the func) needs to provide an explicit CSS declaration for that fallback value. (And I am sorry that I forgot to add them 🙈 - maybe missing this led to the confusion or contributed to it. )

function
```
CSSManagedResponsiveValue(hidden, 'display', value => {
      return value ? 'none' : 'flex'
})
```
this will resolve into

```
{
    [@media screen and (min-width: 768px)]: {
       display : 'none'
    },
    [@media screen and (max-width: 768px)]: {
       display : 'flex'
    },
}
```

and in the component itself we will need to add the CSS declaration that uses the fallback value.
```
    {
      display: 'none', // `display`: is our CSS property that we are managing responsively and `none` is the fallback value.
      ......
      ...CSSManagedResponsiveValue(hidden, 'display', value => {
        return value ? 'none' : 'flex'
      }),
    },
    sx,
  )}
```

Let me know what you think and if you have any concerns 🙌🏼

src/PageHeader/manageResponsiveValueWithCSS.ts
colebemis: `getBreakpointDeclarations` sounds good to me 👍

src/PageHeader/manageResponsiveValueWithCSS.ts
colebemis: I like that! Thanks for explaining 👍

colebemis (approved):
Looks good to go after you make the function name change 👍 Nice work on this @broccolinisoup 🎉

PR #2664, opened 2022-12-07
refactor(project): update TypeScript definitions for React 18

joshblack:
Update TypeScript types for React 18. The goal for this is to support both React 18 and React 17 types.

Some of the changes included:

- `useCallback` needed the arguments to be typed
- `VFC` is deprecated in React 18
- Slot had an issue with children being a generic function
- Dependent packages on older type versions of React that are missing the `children` prop (like `react-dnd`)


joshblack:
@siddharthkp would love your feedback on how to make Slots work with generic children if you have a sec! I feel like the changes I made to get it working are not really correct and we would need to make the context itself generic, potentially.

colebemis (approved):

PR #2666, opened 2022-12-08
make button as check strip out completely in production builds

mattcosta7:
Describe your changes here.

https://github.com/primer/react/pull/2659 introduced a dev only check in an effect, but it doesn't completely strip that effect from production builds.  This removes the entire effect from the production bundle by wrapping the whole effect in the compile flag check (which violates the eslint rule of hooks, but doesn't actually break the rules since this isn't conditional, but instead just included or excluded based on build properties - it will never change between render passes)

### Screenshots

### Merge checklist

No real behavior change should be apparent here


joshblack (approved):

src/Button/ButtonBase.tsx
```diff
@@ -31,14 +31,21 @@ const ButtonBase = forwardRef<HTMLElement, ButtonProps>(
       return merge(baseStyles, sxProp as SxProp)
     }, [baseStyles, sxProp])
 
-    React.useEffect(() => {
-      if (!(innerRef.current instanceof HTMLButtonElement) && !(innerRef.current instanceof HTMLAnchorElement)) {
-        if (__DEV__) {
+    if (__DEV__) {
```
kendallgassner, line 34: thank you for the explanation!

kendallgassner (approved):

Issue #2124, opened 2022-06-14
Tooltip should use Popover API to avoid z-index issues


dgreif:
**Describe the bug**
The `Tooltip` component currently uses a `::before` pseudo-element to render its content, with a z-index of `1000001`.  While this works in a lot of scenarios, it does not match the `Portal` approach used by many of the other `Overlay`s in PRC.  In particular, tooltips can get covered up when their ancestor has created a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) (with something like `z-index: 0`).  It can also be a problem if the parent is clipped or has `overflow: hidden`.

Rather than relying on ancestors to be styled correctly, we should put the Tooltip in a `Portal`.  This way it will always be rendered with a `z-index: 1` at the page stacking context.

**To Reproduce**
See https://codesandbox.io/s/magical-tesla-4pdhth?file=/src/App.js

**Expected behavior**
Tooltip should always show up above anything with `z-index: 1`

I'm not sure if `anchoredPosition` has a 1-to-1 matching with the `direction` option, so we may need to add more options there, or else this will probably be a breaking change.  likely a breaking change anyways given the `Portal` z-index is 1 at the page level.

lesliecdubs:
Thanks @dgreif, we agree with this assertion. Moving to the backlog to address in the future.

mperrotti:
Related: a SegmentedControl with icon-only buttons (which show a tooltip when hovered) needs this work to be done to be considered accessible.

milemons:
Hello! Would it be possible to get this prioritized? 🙏  We currently can't use the tooltip component (and have been relying on `title` for most elements due to it being clipped (since we have `overflow: hidden` or `zIndex: 1` set on many items.

We also have had some issues where the tooltip position does not change automatically and sometimes the tooltip is cut off the page (ie. if there's a button on the left side of the page, the tooltip will overflow off the edge of the page)

milemons:
Related issue: https://github.com/github/repos/issues/3485
And related slack discussion for another issue: https://github.slack.com/archives/C0357UQ03KN/p1667487768081549

lesliecdubs:
Thanks all, I've moved this back to the PRC Inbox to review again on Monday. We'll look into prioritizing it in time for Repos GA, which I've been told is planned for January.

lesliecdubs:
[...]

PR #2668, opened 2022-12-08
[WIP] Use Portals to ensure tooltips are always visible regardless of neighboring stack contexts.

radglob:
Using this as a spot to flesh out what I think needs to happen for #2124.

This seems like a relatively large change, and may be a breaking change depending on what choices we make.

We want to use Portal components in the Tooltip component so we don't need to worry about the stacking contexts of neighboring content. This strategy is currently used in AnchoredOverlays. A few things are different in the case of tooltips.

1) The Tooltip component serves as a wrapper around whatever content is placed inside of it. The content of the tooltip (which is defined with the `aria-label` attribute) is rendered using the `::before` and `::after` pseudo-elements. This is different from AnchoredOverlays, which renders the anchor and the overlay as separate DOM elements.

2) The directions used by `useAnchoredPosition` (which are defined in [@primer/behaviors](https://github.com/primer/behaviors/blob/main/src/anchored-position.ts#L5)) are different from the cardinal directions used by tooltips. Tooltips also support diagonal directions, while AnchoredPosition does not.

I think we will need to update the directions in `@primer/behaviors` to support diagonal positioning. We can save ourselves a breaking change by writing a function to convert the cardinal directions used in `Tooltip` to the relative directions in `@primer/behaviors`. Alternatively, we can convert tooltips to use the AnchoredPosition directions.

Additionally, it may make sense to use DOM elements for both the tooltip and its anchor, as we only want the tooltip itself to use a Portal. The anchor should not be moved, as this can break the content flow inside of containers.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack:
Thanks for putting this together @radglob! This makes a ton of sense to me, especially some of the ways you suggested mapping existing directions to ones in `@primer/behaviors` for keeping the same API 👍 

It seems like tooltip changes looking forward are related to either:

- Positioning (for the portal use-case described here)
- Markup (to make sure the tooltips are using the correct markup instead of the existing `aria-label` solution)

In the parent issue, it seems like much of it revolves around positioning logic. Do you feel like any of the markup changes from: https://github.com/github/primer/pull/1276 would need to be introduced with this or would this be purely for the styles and the markup comes in a later addition?

radglob:
We're going to use the [Popover API](https://developer.chrome.com/docs/web-platform/popover-api/) instead of portals. I'll open a new PR.

PR #981, opened 2021-01-25
Custom Elements ADR

emplums:
Throughout the last few years folks from the Design Infrastructure, Web Systems & UI Platform teams have discussed the idea of using custom elements for behaviors in Primer React Components. The topic has been brought up several times, but we've yet to formalize our approach in an ADR that we can point to for any future questions about the idea.

This PR attempts to distill my historical knowledge of the topic and give a holistic overview of the many pros and cons for using custom elements in PRC, and ultimately why we've never moved forward with the approach.

Please feel free to add any suggestions or feedback!

Link to markdown view: https://github.com/primer/components/blob/388fed0874f2e2b3379f584768e6d1775d77e07c/contributor-docs/ADRS/custom-elements.md

PR #2672, opened 2022-12-09
Create adr-xxx-behavior-isolation.md

keithamus:
_For internal GitHub discussion on this topic, please visit and comment on [this PR](https://github.com/github/primer/pull/1588)._

Given we have now merged #2484 and demonstrated that it is both feasible and useful to use Custom Elements to drive behaviour of some Primer React Components, especially when those behaviours can be shared with Primer View Components.

This PR raises a new ADR which goes into some details about the current state of affairs. It also addresses some of the concerns originally raised in ADR 002 (https://github.com/primer/react/pull/981).


contributor-docs/adrs/adr-xxx-behavior-isolation.md
```diff
@@ -0,0 +1,137 @@
...
+
+ADR 002 describes complexity around Custom Elements and Server Side Rendering. It claims that for Custom Elements to be well supported, that Primer React would need to accumulate added complexity which would be surfaced to the user, going on to claim that while libraries may exist their usage is antithetical to Custom Element patterns.
+
+Primer React's [recent addition of the Relative Time component](https://github.com/primer/react/pull/2484) (itself a wrapper of the [`relative-time`](https://github.com/github/relative-time-element) custom element) demonstrates that a custom element can be loaded into a server side runtime such as NodeJS without causing error, by adding [2 additional lines of code to the element itself](https://github.com/github/relative-time-element/blob/4ab3f2807a02dfd687b2b7fa04553300ff87ddbe/src/relative-time-element.ts#L2-L3):
+
+```typescript
+const root = (typeof globalThis !== 'undefined' ? globalThis : window) as typeof window
+const HTMLElement = root.HTMLElement || (null as unknown as typeof window['HTMLElement'])
+```
```
dgreif, line 47: While this code is indeed compact, there a few caveats that should be considered

- This code needs to be included in _every_ custom element definition. As you state below, this could potentially be provided globally in the SSR environment, but we still need to code present in case it isn't provided globally.
- This isn't the only code needed for SSR safety - there is also [this chunk](https://github.com/github/relative-time-element/blob/main/src/relative-time-element-define.ts#L3-L14) which is responsible for registering the custom element and safely falling back to doing _nothing_ if the environment doesn't support custom elements. Compare with [this other example](https://github.com/github/clipboard-copy-element/blob/main/src/index.ts#L14-L17), which is our standard registration code.

contributor-docs/adrs/adr-xxx-behavior-isolation.md
```diff
@@ -0,0 +1,137 @@
...
+
+These lines could also be added to the global context, to avoid adding it to each element.
+
+The `RelativeTime` react component uses `@lit-labs/react`, which makes integrating the component into React painless, and requires [1 line of code for integration](https://github.com/primer/react/blob/d995bb8485fa1dfd0a3637b4851c5ac539b8416d/src/RelativeTime/RelativeTime.tsx#L8):
+
+```typescript
+const RelativeTime = styled(createComponent(React, 'relative-time', RelativeTimeElement))(sx)
```
dgreif, line 54: This is really slick!

contributor-docs/adrs/adr-xxx-behavior-isolation.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+There are no additional steps required for consumers of Primer React, and the component behaves as if it is a first-class React component, supporting the same properties and events as any regular React component. The server will render the Custom Element tag (in the case of the `RelativeTime` component, `<relative-time>`) and the client is able to assign class properties and use callbacks for events, as expected.
```
dgreif, line 57: Some of our main goals with SSR which may difficult to achieve with custom elements:

* Rendering without JavaScript
* Reducing CLS

`RelativeTime` is a perfect example of a component which we _could_ render with SSR, but will be unable to if it is wrapped in a custom element. By performing client-only rendering, we miss out on both of these SSR goals. If we instead had a `relativeTime` vanilla JS behavior (with a callback when it needed to update), we could write a very small wrapper in React which would function in both SSR and CSR environments. Additionally, that JS behavior could be consumed by the `<relative-time>` custom element for use within PVC.

contributor-docs/adrs/adr-xxx-behavior-isolation.md
```diff
@@ -0,0 +1,137 @@
...
+
+#### Styling with `styled-components`
+
+Primer React currently uses a CSS-in-JS library called [`styled-components`](https://styled-components.com/). This library accumulates objects that map to CSS properties, and during runtime it will create a hash digest to use as the class name for the component, and inject the mapped CSS into a stylesheet on the page. This way, CSS lives in JavaScript object notation within the same file as the React Component itself. In addition to this, Primer React also uses [`styled-system`](https://styled-system.com/) which allows the use of style properties directly within a React Components props. Primer React uses the `sx` prop to allow for customisation of styles.
+
+ADR 002 describes complexity around styling Custom Elements with the styled-components library. It claimed that a custom element must use another wrapping element to apply styles.
+
+Today, Primer React's RelativeTime component can be styled using the `sx` prop, just like any other Primer React component.
```
dgreif, line 67: If we start to use custom elements for behaviors which do not have a visual components (eg everything in [primer/behaviors](https://github.com/primer/behaviors/tree/main/src) today), we be adding a new Element into the DOM to wrap the targeted area. This means the wrapping custom element will become part of the flow of the document and have styles of it's own, even though it is only intended to provide _behavior_, not _visuals_. This can become a practical issues if the element being wrapped is part of a layout with flex/grid styles, in which case the wrapper needs to be styled in such a way as to take up the same amount of space as the component being wrapped would have. Is there some way to tell the browser to completely ignore an element from the styling flow?

contributor-docs/adrs/adr-xxx-behavior-isolation.md
```diff
@@ -0,0 +1,137 @@
...
+Today, Primer React includes the `RelativeTime` component. Using the `@lit-labs/react` library `RelativeTime` wraps the `relative-time` custom element and automatically maps props directly to the class instance, and converts `on*` callback props to event listeners that can then observe event announcements from the element. In this way, `RelativeTime` behaves just like if it were written in React. The utilisation of the `@lit-labs/react` library requires 1 line of code:
+
+```typescript
+const RelativeTime = styled(createComponent(React, 'relative-time', RelativeTimeElement))(sx)
+```
+
+While React 18 and below require a small library like `@lit-labs/react`, due to the [lack of support for Custom Element](https://custom-elements-everywhere.com/libraries/react/results/results.html), React 19 [will likely need no such library](https://custom-elements-everywhere.com/libraries/react-experimental/results/results.html).
+
+
+#### Organizational overhead
```
dgreif, line 114: Another concern for this section is the fact that only one version of a custom element (by tag name) can be registered on a page at a time. In dotcom, this will be a particular issue if you are soft navigating between pages using PVC and PRC, which could be using _different versions_ of the underlying custom elements. If there were a breaking change to the custom element, both PRC and PVC will need to be updated at the same time.

In contrast, if we used vanilla JS behaviors and PRC/PVC got out of sync for a short period of time, the client is able to run _both_ versions of the behavior in the same page. This obviously isn't ideal from a bundle-size perspective, but it does ensure we do not accidentally release a breaking change to our users.

dgreif (commented):
Congrats with what you have accomplished with the RelativeTime + <relative-time> setup, it really is a neat and simple wrapper! I have some thoughts, but I'll drop them on the internal version of the ADR

contributor-docs/adrs/adr-xxx-behavior-isolation.md
```diff
@@ -0,0 +1,209 @@
+# ADR XXX: Isolating behaviors through custom elements and vanilla JavaScript
```
joelhawksley, line 1: I think we need a number here?

joelhawksley: Also, perhaps we should link to this ADR from 002, specifying that 002 was superseded?

joshblack (approved):

joelhawksley (approved):

PR #2673, opened 2022-12-09
test(project): update tests for React 18

joshblack:
Update test files that were failing when using React 18.

#### Changelog

**New**

**Changed**

- Update `ThemeProvider` tests to use `@testing-library/user-event`
- Add `TextEncoder` for `useMedia` tests
- Add `act()` calls for tests for `TextInputWithTokens` and `MarkdownEditor`

**Removed**


joshblack:
Trying to figure out why `act()` violations are coming up for `MarkdownEditor.test.tsx` 🤔 

**Notes**

- `MarkdownEditor.test.tsx`
  - `savedRepliesRef.current?.openMenu()` is the call that triggers `act()` violations
  - This follows into `SavedRepliesButton` which exposes this as a ref through `useImperativeHandle`
  - This method causes a state change through `setOpen(true)`
  - `open` is passed through to `SelectPanel`, when it goes from `false` -> `true` it causes `act()` violations
  - `SelectPanel` passes `open` along to `AnchoredOverlay` which ultimately renders `FilteredActionList`
  - If `FilteredActionList` is not provided as a child to `SelectPanel`, no `act()` violations occur

colebemis (approved):

Issue #2622, opened 2022-11-28
IconButton and ActionMenu.Button types are incorrect


mattcosta7:
### Description

<img width="960" alt="Screen Shot 2022-11-28 at 12 34 10 PM" src="https://user-images.githubusercontent.com/8616962/204343530-f5e0b967-066d-44b8-96cb-49f0a0f359d9.png">

<img width="895" alt="Screen Shot 2022-11-28 at 12 33 56 PM" src="https://user-images.githubusercontent.com/8616962/204343535-377397c7-39ed-4619-bdc3-b4f2392be9ea.png">

A similar issue was fixed in - https://github.com/primer/react/pull/2445

### Steps to reproduce

Import and try to get typing info

### Version

latest

lesliecdubs:
Thanks for filing! This may be related to #2519. Moving to the backlog for now.

Issue #2519, opened 2022-11-02
Incorrect types for IconButton


colebemis:
### Description

IconButton incorrectly allows a `leadingIcon` prop to be passed to the component. 

```jsx
// IconButton doesn't accept a `leadingIcon` prop.
// This should cause a TypeScript error but it doesn't.
// Instead, it causes a runtime error.
<IconButton leadingIcon={PencilIcon} />
```

Orginally reported by @milemons in [this Slack thread](https://github.slack.com/archives/C0357UQ03KN/p1667399595359759?thread_ts=1667330297.439599&cid=C0357UQ03KN)

### Steps to reproduce

See this StackBlitz for a reproduction of the bug: https://stackblitz.com/edit/react-ts-nyunix?file=App.tsx

### Version

v35.14.0

broccolinisoup:
Thanks to @mattcosta7 for [the PR](https://github.com/primer/react/pull/2677)! I am moving this issue to `In code review` column and unassigning myself. Let me know if you have any issues with it!

broccolinisoup:
Ah not at all! @mattcosta7 I really appreciate the fixes! 🙇🏼‍♀️  I just didn't have much time to look at them

PR #2677, opened 2022-12-10
Update types for IconButton, ActionMenu.Button, TabNav.Link and internal consumers

mattcosta7:
Describe your changes here.

The types for these buttons have seemed incorrect for some time. This should correct these types and ensure they get proper autocompletion in IDEs. I'm not 100% sure if the polymorphism on AnchorMenu.Button is correct, but it's consistent with what the types purported to be previously as far as I can tell

fixes https://github.com/primer/react/issues/2622
fixes https://github.com/primer/react/issues/2519

Are there are, similar instances we should fixup while here?

I snuck in a small change more broadly, which replaces

```ts
const component = ({sx: sxProp = {}}) => {}
//with
const component = ({sx: sxProp = defaultSxProp}) => {}
```

This should avoid a new object allocation on every render of multiple components, and potentially memo/callback/effect re-evaluations related to them.  I don't expect a significant perf improvement, but this should be a small one with no negative effects

These changes lead to many incompatible typings inside primer, which I've fixed up.  I think there are likely many more typing inconsistencies in primer we should work to root out, but this should at least build as expected. 

A consideration here is "Are these fixes broad enough to be considered breaking changes" as they'll likely require consumers to update code, similar to what we've had to do here

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Button/types.ts
```diff
@@ -53,7 +47,7 @@ export type ButtonProps = {
 
 export type IconButtonProps = ButtonA11yProps & {
   icon: React.FunctionComponent<React.PropsWithChildren<IconProps>>
-} & ButtonBaseProps
+} & Omit<ButtonBaseProps, 'aria-label' | 'aria-labelledby'>
```
mattcosta7, line 50: we don't want the built in aria-label/labelled by for icon, because we explicitly override them

src/Button/types.ts
```diff
@@ -13,14 +13,9 @@ export type VariantType = 'default' | 'primary' | 'invisible' | 'danger' | 'outl
...
-/**
- * Remove styled-components polymorphic as prop, which conflicts with radix's
- */
-type StyledButtonProps = Omit<ComponentPropsWithRef<typeof StyledButton>, 'as'>
-
 type ButtonA11yProps =
-  | {'aria-label': string; 'aria-labelledby'?: never}
-  | {'aria-label'?: never; 'aria-labelledby': string}
+  | {'aria-label': string; 'aria-labelledby'?: undefined}
+  | {'aria-label'?: undefined; 'aria-labelledby': string}
```
mattcosta7, line 18: Undefined is more correct, in that

```
aria-label={undefined}
aria-labelledby="something"
```

should work

src/utils/defaultSxProp.tsx
```diff
@@ -0,0 +1,3 @@
+import {BetterSystemStyleObject} from '../sx'
+
+export const defaultSxProp: BetterSystemStyleObject = {}
```
mattcosta7, line 3: we should use this more often instead of adding inline

`sx = {}` parameters, so I extracted and shared betwen each of the button components touched in this pr

src/Button/types.ts
```diff
@@ -36,24 +31,23 @@ export type ButtonBaseProps = {
...
-  React.ButtonHTMLAttributes<HTMLButtonElement> &
-  StyledButtonProps
+  React.ButtonHTMLAttributes<HTMLButtonElement>
 
 export type ButtonProps = {
   /**
    * The leading icon comes before button content
    */
-  leadingIcon?: React.FunctionComponent<React.PropsWithChildren<IconProps>>
+  leadingIcon?: React.FunctionComponent<React.PropsWithChildren<IconProps>> | null | undefined
```
mattcosta7, line 40: should we add anything 'falsey' as options?   we internally use null and undefined

mattcosta7:
I'll test the canary build internally on monday, but I think this is probably reviewable before that

mattcosta7:
PR using this prerelease version in an internal codebase, with type fixes applied - https://github.com/github/memex/pull/13708

mattcosta7:
A few thoughts here, but from what I can tell this is the _correct_ set of typings. 

```tsx

export function ExtendedIconButton({
  ...props
}: IconButtonProps) {
  return (
    <IconButton
      {...rest}
      aria-label={expandCollapseLabel}
      aria-labelledby={undefined}
    />
  )
}
```

Any time we pass `aria-label` to an IconButton that explicitly has aria-label, we need to force aria-labelledby to be undefined manually, otherwise there's no way for the compiler to understand that both of these aren't used. 

this is a sideeffect of the typings added in https://github.com/primer/react/pull/2659

I'm not sure that we _should_ be doing this at the type level, vs dev only runtime warnings?

src/Button/Button.tsx
```diff
@@ -3,16 +3,13 @@ import {ButtonProps} from './types'
...
-  ButtonProps
->(({children, ...props}, forwardedRef): JSX.Element => {
+const ButtonComponent = forwardRef(({children, ...props}, forwardedRef): JSX.Element => {
   return (
     <ButtonBase ref={forwardedRef} as="button" {...props}>
       {children}
     </ButtonBase>
   )
-})
+}) as PolymorphicForwardRefComponent<'button', ButtonProps>
```
joshblack, line 12: Just curious, from the TS lenses what's the difference between having the type live on `ButtonComponent: TypeName` versus `value as TypeName`? 👀

src/utils/defaultSxProp.tsx
mattcosta7: we could do that.  We could also, probably just mark the default as a readonly type (which is the effect freeze has here).  

Happy to check that out to see

mattcosta7: I'd imagine any change to this object, anywhere would cause a bunch of test cases to fail, or be pretty easy to pick up in code review.  we could make one per module, instead of sharing a single default, but seems common enough where a single object made sense

src/drafts/MarkdownEditor/_SavedReplies.tsx
```diff
@@ -86,7 +86,14 @@ export const SavedRepliesButton = () => {
 
   return items ? (
     <SelectPanel
-      renderAnchor={props => <ToolbarButton {...props} icon={ReplyIcon} aria-label="Add saved reply (Ctrl + .)" />}
+      renderAnchor={props => (
+        <ToolbarButton
+          {...props}
+          icon={ReplyIcon}
+          aria-label="Add saved reply (Ctrl + .)"
+          aria-labelledby={undefined}
```
joshblack, line 94: Is the `aria-labelledby` required here due to parent `props` that's being spread or for a type-related reason?

src/drafts/MarkdownEditor/_ViewSwitch.tsx
```diff
@@ -45,7 +45,7 @@ export const ViewSwitch = ({selectedView, onViewSelect, onLoadPreview, disabled}
   return (
     <Box sx={{display: 'flex', flexDirection: 'row'}}>
       {condensed ? (
-        <IconButton {...sharedProps} disabled={disabled} icon={icon} label={label} />
+        <IconButton {...sharedProps} disabled={disabled} icon={icon} aria-label={label} />
```
joshblack, line 48: Is `label` something that ever worked for `IconButton`? Was trying to track down this type but `ButtonA11yProps` only seems to include `aria-label` and `aria-labelledby`, if I understand right?

joshblack (commented):
Thanks for doing this! 🙌 

Left a couple comments above, mostly just was curious around certain techniques or situations.

What do you feel like are the big user-facing changes with these updates to the types? Is it something manageable or would it be a larger effort?

src/drafts/MarkdownEditor/_ViewSwitch.tsx
mattcosta7: I don't think it ever worked either - or at least I didn't see it in a recent changelog

src/drafts/MarkdownEditor/_SavedReplies.tsx
mattcosta7: Yep, since we're spreading props that could have either aria-label or aria-labelled by, we need to be sure to not have both (given the pr I referenced that enforced only 1 of these exist (and only 1)

mattcosta7:
> What do you feel like are the big user-facing changes with these updates to the types? Is it something manageable or would it be a larger effort?

github/memex#13708

I think this PR should shake out a good bit of end-user changes.

Anything that previously used IconButton or ActionMenu.button in consumer code could have been incorrectly typed, and they'll need to fix it up, similar to how stories/tests were fixed in this PR. 

The impact of https://github.com/primer/react/pull/2659 also means that explicit aria-label/labelledby combos are needed anytime something spreads a prop (unless the spreading components omit one of them)

src/global.d.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line no-var
+declare var __DEV__: boolean
```
mattcosta7, line 2: declare var means we only need this once, not per file

src/utils/defaultSxProp.tsx
```diff
@@ -0,0 +1,5 @@
+import {BetterSystemStyleObject} from '../sx'
+
+export const defaultSxProp: Readonly<BetterSystemStyleObject> = __DEV__
+  ? Object.freeze<BetterSystemStyleObject>({})
+  : {}
```
mattcosta7, line 5: @joshblack froze this in dev, but typed it readonly as well, so it should help catch these easily

src/Button/types.ts
joshblack: @mattcosta7 would there be any alternatives that you'd recommend?

src/Button/types.ts
joshblack: Having `null` here definitely seems useful from the caller side. One quick question for `undefined`, would it already be included without explicitly adding it to the union type?

joshblack:
> I'm not sure that we should be doing this at the type level, vs dev only runtime warnings. I think type level checks might be too annoying, as the rest scenario is unlikely to ever pass both, but open to this being explicit as it is now (which is not wrong)

I definitely agree, the type level checks seem a little annoying for this situation given how uncommon it is. If we need this explicit check, it would be nice to move it into a separate PR/release if that will cause a non-zero set of changes downstream to try and separate that diff from most of the awesome updates that are already here 🔥 

> I think this PR should shake out a good bit of end-user changes.

Sounds great! Thanks again for doing this 🙌

mattcosta7:
> > I'm not sure that we should be doing this at the type level, vs dev only runtime warnings. I think type level checks might be too annoying, as the rest scenario is unlikely to ever pass both, but open to this being explicit as it is now (which is not wrong)
> 
> I definitely agree, the type level checks seem a little annoying for this situation given how uncommon it is. If we need this explicit check, it would be nice to move it into a separate PR/release if that will cause a non-zero set of changes downstream to try and separate that diff from most of the awesome updates that are already here 🔥

@joshblack to be clear, that type level check is already on main.  Because we were not correctly typing IconButton at all before, every usage by consumers was `React.ComponentType<any>` instead of type checking. 

If we want to back out the change, that landed on main in https://github.com/primer/react/pull/2659 I'm happy to do that here, but this PR is just making the types correct to what they are on main already

> > I think this PR should shake out a good bit of end-user changes.
> 
> Sounds great! Thanks again for doing this 🙌

mattcosta7:
This might add a bit more context. 

Before this PR - the consumer typings for IconButton look like

Screen Shot 2022-12-13 at 1 44 42 PM

Screen Shot 2022-12-13 at 1 44 15 PM

So IconButton is basically given HTML Button attributes, and anys

This correctly uses the types we've defined already (adding/removing none).  But we should probably consider whether we're typing IconButton correctly in main before shipping this

joshblack:
Thanks for the extra context @mattcosta7! I really appreciate it 🙏 

To recap, based on my understanding it seems like this part of this PR is updating the following to correct for the `never` type:

```diff
type ButtonA11yProps =
-  | {'aria-label': string; 'aria-labelledby'?: never}
+  | {'aria-label': string; 'aria-labelledby'?: undefined}
-  | {'aria-label'?: never; 'aria-labelledby': string}
+  | {'aria-label'?: undefined; 'aria-labelledby': string}
```

This unfortunately will prompt some common issues in usage for consumers, for example:

```tsx
type Props = 
  | { 'aria-label': string, 'aria-labelledby'?: undefined }
  | { 'aria-label'?: undefined, 'aria-labelledby': string }

function Test(props: Props) {
  return <button type="button" {...props}  />;
}

function Test2(props: Props) {
  return <Test aria-label="hello" {...props} />;
}
```

In this scenario, `Test2` will have a type error and would need to either:

- Explicitly set `aria-labelledby={undefined}` on `Test`
- Update the type for props to something like `Omit<Props, 'aria-labelledby'>` which may not be preferred

Ideally, consumers wouldn't have to type it this way but this seems unavoidable in this spread scenario (let me know if others have an idea of how to work around this!)

I think you had a good suggestion @mattcosta7 to potentially bring this check as a runtime check, I'm curious what you think @kendallgassner about doing this or if there are any other alternatives that should be considered?

kendallgassner:
@joshblack thanks for the tag! This confused me a bit.... why not just Omit 

type Props = 
  | { 'aria-label': string, 'aria-labelledby'?: undefined }
  | { 'aria-label'?: undefined, 'aria-labelledby': string }

function Test(props: Props) {
  return <button type="button" {...props}  />;
}

function Test2(props: Omit<Props,  'aria-labelledby' >) {
  return <Test aria-label="hello" {...props} />;
}

Are you suggesting the omit in this case wouldn't be preferred?

mattcosta7:
> @joshblack thanks for the tag! This confused me a bit.... why not just Omit
> 
> ```tsx
> type Props = 
>   | { 'aria-label': string, 'aria-labelledby'?: undefined }
>   | { 'aria-label'?: undefined, 'aria-labelledby': string }
> 
> function Test(props: Props) {
>   return <button type="button" {...props}  />;
> }
> 
> function Test2(props: Omit<Props,  'aria-labelledby' >) {
>   return <Test aria-label="hello" {...props} />;
> }
> ```
> 
> Are you suggesting the omit in this case wouldn't be preferred?

That omission needs to propagate up the types for every component that ever could pass props through from consumers.  I think we're bleeding type implementations in an unexpected way doing that. 

https://github.com/primer/react/pull/2677/files#diff-683f76e89e04e13df7826c521041ae0c2b1bcffeae6288e285791e41977a53f7R89-R96

This scenario is pretty common in consumer code (passing through button props, for instance).  But since we're passing through button props, we now need to explicitly force to undefined, or go up the entire component stack to add omits to anything that might pass props through.   It does feel we're leaking details into the type system. 

If we want to enforce that every consumer do this all the time, I think it's pretty heavy handed, compared to a runtime warning in dev that says "label properly" or something. 

https://github.com/github/memex/blob/f8a68d62a9854a1c3e7c42e840424bea9ad39a6e/src/client/components/common/icon-button-with-tooltip.tsx#L52-L108

this example wraps IconButton, but since we pass through props we need to explicitly set the opposite aria property to undefined.  it feels pretty unintuitive and clunky - when the usage is unlikely to ever have the scenario occur where both get passed.

kendallgassner:
I am not super opposed to runtime I was just hoping for some type magic. I don't totally see why we wouldn't want to omit the[ aria-labelledby in the props of the memex example.](https://github.com/github/memex/blob/f8a68d62a9854a1c3e7c42e840424bea9ad39a6e/src/client/components/common/icon-button-with-tooltip.tsx#L6). It is more work but its also more accurate.

mattcosta7:
> I am not super opposed to runtime I was just hoping for some type magic. I don't totally see why we wouldn't want to omit the[ aria-labelledby in the props of the memex example.](https://github.com/github/memex/blob/f8a68d62a9854a1c3e7c42e840424bea9ad39a6e/src/client/components/common/icon-button-with-tooltip.tsx#L6). It is more work but its also more accurate.

Agreed - totally fine to say we can force it up the tree everywhere - I just mentioned it in this PR because since we'll be typing this correctly, it's likely the scenario that will cause the most code churn

src/Button/types.ts
mattcosta7: technically yes, but not necessarily.

```
{}
{ leadingIcon: null }
{leadingIcon: () => <></>}
```

would be `{ leadingIcon?: ComponentType | null }`

whereas `{ leadingIcon?: ComponentType | null | undefined }`

includes

```
{ leadingIcon: undefined }
```

due to historical quirks, this case is the same as `?`, but there've been some rumblings about eventually making them explicit in typescript. 

that's enough to say, we could do either.   We could also add false/boolean here to account for things like && returning false?

colebemis (approved):
All these changes seem reasonable. Thanks for digging into this, @mattcosta7! 

Can you help me understand the downstream impacts of these changes? In what cases will these fixes break existing code?

mattcosta7:
> All these changes seem reasonable. Thanks for digging into this, @mattcosta7!
> 
> Can you help me understand the downstream impacts of these changes? In what cases will these fixes break existing code?

@colebemis they shouldn't break code downstream, but since consumers have only had types like these (in the codeblock below), anyone might have improperly added additional props to these components, and those will become errors. 

```ts
IconButton: ForwardRefExoticComponent<any> & RefAttributes<RefObject<HTMLElement> | undefined>>
ActionMenu.Button: ForwardRefExoticComponent<any> & RefAttributes<RefObject<HTMLElement> | undefined>>
```

Additionally, since https://github.com/primer/react/pull/2659 add the requirement for an exclusive or on aria-label or aria-labelledby to IconButton, anything spreading props that might have both of these will need to explicitly make one of them undefined or `Omit` them from types. 

Happy to chat through some more, but the stories/tests in this pr, and the pr using the canary build - https://github.com/github/memex/pull/13708 highlight areas where some churn might be incurred while we correct the typings here.

joshblack:
Also tried to do a draft PR over in dotcom to see what the splash zone would be for these type changes: https://github.com/github/github/pull/251734

mattcosta7:
> Also tried to do a draft PR over in dotcom to see what the splash zone would be for these type changes: [github/github#251734](https://github.com/github/github/pull/251734)

looks like ~35 type errors

joshblack:
Tried to classify the errors into a couple of different buckets, seems like around 60% of the errors were related to one of the following groups:

- A specific property does not exist on the type (`muted` was the most common)
- `Type '(event: MouseEvent) => void' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.`
- `Type 'ForwardRefExoticComponent<LinkProps & RefAttributes<HTMLAnchorElement>>' is not assignable to type '"a" | "button" | undefined'`
- `ElementType<any>' is not assignable to type 'FunctionComponent<PropsWithChildren<IconProps>>`

Around 9 of the issues reported were related to `aria-label` or `aria-labelledby`

joshblack (approved):
Definitely seems like the right direction forward, thanks for doing this! I would love to make sure the rollout for these changes feels smooth for people, would you be interested in pairing up at all on this for dotcom @mattcosta7? Totally get with holiday plans that this would probably be tricky haha

mattcosta7:
> Definitely seems like the right direction forward, thanks for doing this! I would love to make sure the rollout for these changes feels smooth for people, would you be interested in pairing up at all on this for dotcom @mattcosta7? Totally get with holiday plans that this would probably be tricky haha

Yeah, sure thing.  I'm around through the holiday week, so happy to help land this. 

I'll work on conflict fixes here shortly

src/Button/types.ts
```diff
@@ -37,16 +36,16 @@ export type ButtonProps = {
...
    * The trailing icon comes after button content
    */
-  trailingIcon?: React.FunctionComponent<React.PropsWithChildren<IconProps>> | null | undefined
+  trailingIcon?: React.ComponentType | null | undefined
   children: React.ReactNode
 } & ButtonBaseProps
 
 export type IconButtonProps = ButtonA11yProps & {
-  icon: React.FunctionComponent<React.PropsWithChildren<IconProps>>
+  icon: React.ComponentType
```
mattcosta7, line 48: broadens these to they work with arbitrary components, and not _only primer octicons_

src/TabNav.tsx
```diff
@@ -105,9 +106,8 @@ const TabNavLink = styled.a.attrs<StyledTabNavLinkProps>(props => ({
   }
 
   ${sx};
-`
+` as PolymorphicForwardRefComponent<'a', TabNavLinkProps>
```
mattcosta7, line 109: this is used polymorphically, but the styled-components as types conflict some with the polymorphic helper types - overriding the styled-components version gives better typing, and avoids those issues

src/__tests__/TabNav.types.test.tsx
```diff
@@ -20,3 +21,11 @@ export function shouldNotAcceptSystemProps() {
     </>
   )
 }
+
+export function shouldAcceptButtonAsProps() {
+  return <TabNav.Link as={Button} />
+}
+
+export function shouldAcceptTabNavLinkprops() {
+  return <TabNav.Link to="to something" selected as={Button} />
```
mattcosta7, line 30: validates that we can accept specific props, and as

mattcosta7:
@joshblack opened this one - https://github.com/github/github/pull/252052

Which fixes up the type errors in dotcom. We'd probably need to work with product teams to ensure the aria-labels added are good enough, but it does the heavy lifting.  

This did uncover some additional fixes necessary to improve the `as` prop in `TabNav.Link` for a case used currently in dotcom

https://github.com/primer/react/pull/2677/commits/96a6224448b283850841df032a2dfcd44359c9d1

and opens a new question related to a test.

This updates the types to support a test we have currently in this codebase.  I think the test, incorrectly, accepts an `href` attribute for a `div`.  The polymorphic as prop changes catch that. I've updated the types to support the test, but I think it might be more valid to update the test to avoid setting an href on a div.

https://github.com/primer/react/pull/2677/commits/bf1f55e24c316dad6651f1578bbed243bb8f5ceb

colebemis (approved):
Looks great. Thanks for taking this on @mattcosta7 and @joshblack! 💖

PR #2606, opened 2022-11-22
Fix issue where scroll containers nested inside of PageLayout.Pane would be hard to use on mobile.

radglob:
Setting `overflow: auto` only on larger screen sizes addresses this issue. Instead of the user needing to focus outside of an inner scroll container on mobile, all sections flow inside a single scroll container.

There's an additional storybook story (Scroll Container within Page Layout Pane) that demonstrates this.

Closes github/primer#1517

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2685, opened 2022-12-12
Setting overflow: auto without a media query to fix resizable pane bug.

### LINKED ISSUE #2684: SplitPane resizing broken in 35.15.1
### Description

In 35.15.1 we shipped https://github.com/primer/react/pull/2606 which overrides the 'overflow' to auto for the split pane view, and avoids the explict width we should be setting for resizing. 

### Steps to reproduce

Try to resize the split pane in the latest version, notice it wont resize

### Version

v35.15.1

### Browser

_No response_

radglob:
This basically reverts the work in #2606. Setting `overflow: auto` inside of a media query seemed to fix an issue where scrolling panes would be hard to use on mobile as you would have a scrolling container within another scrolling container. Changing this back to not use a media query seems to have the desired behavior without breaking resizable panes.

Closes #2684.

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/PageLayout/PageLayout.tsx
```diff
@@ -675,9 +675,7 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
               ? ['100%', null, 'clamp(var(--pane-min-width), var(--pane-width), var(--pane-max-width))']
               : paneWidths[width],
             padding: SPACING_MAP[padding],
-            [`@media screen and (min-width: ${theme.breakpoints[1]})`]: {
-              overflow: 'auto',
-            },
+            overflow: 'auto',
```
colebemis, line 678: I'm surprised this still fixes the initial issue 🤔 

I was expecting to have to something like this:
```suggestion
            overflow: [null, 'auto'],
```

mattcosta7:
While we're here, can we add any tests to avoid regressing the resizable pane in the future?

src/PageLayout/PageLayout.tsx
```diff
@@ -675,9 +675,7 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
               ? ['100%', null, 'clamp(var(--pane-min-width), var(--pane-width), var(--pane-max-width))']
               : paneWidths[width],
             padding: SPACING_MAP[padding],
-            [`@media screen and (min-width: ${theme.breakpoints[1]})`]: {
-              overflow: 'auto',
-            },
+            overflow: [null, 'auto'],
```
colebemis, line 678: Ah, I think the correct breakpoint is actually this: 
```suggestion
            overflow: [null, null, 'auto'],
```

src/PageLayout/__snapshots__/PageLayout.test.tsx.snap
```diff
@@ -69,6 +69,7 @@ exports[`PageLayout renders condensed layout 1`] = `
   --pane-max-width: calc(100vw - 511px);
   width: 100%;
   padding: 0;
+  overflow: auto;
```
colebemis, line 72: Looks like you'll need to update these snapshots

colebemis (approved):

PR #2692, opened 2022-12-14
Triggering onSelect for tree item middle click

jdrush89:
Describe your changes here.

Triggering onSelect when middle click is pressed on a tree item. Users need to support the link shortcut of middle clicking to open in a new tab. onClick does not trigger an event for middle clicks, onAuxClick has to be used instead. Consumers can check the event.button when they get onSelect to handle opening the new tab. I could see an alternative solution being that the primer tree checks for ctrl/cmd + click and middle click and fires a different selection event for these cases. I'm happy to implement that if it seems like a better solution.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/yellow-shoes-glow.md
+++ b/.changeset/yellow-shoes-glow.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Trigger onSelect when TreeView items are middle clicked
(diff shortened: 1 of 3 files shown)


colebemis (approved):
Looks good 👍 Thanks for implementing this, @jdrush89 💖

colebemis:
I don't have a mouse with a middle button, so I'm going to trust that `onAuxClick` works as you described

Issue #2595, opened 2022-11-18
Default example for `Tooltip` is inaccessible


khiga8:
The default example for the `Tooltip` doc is on a static element, which we really want to discourage. Tooltip on a static element is one of the most common accessibility issues we're seeing in dotcom, and it's usually a design problem that can't easily be fixed.

Until https://github.com/github/primer/issues/1244 is implemented, I propose we update this example to use an interactive element, emphasize that it should not be set on static elements, and link to [Tooltip alternatives.](https://primer.style/design/accessibility/tooltip-alternatives/).

lesliecdubs:
Thanks @khiga8! Dropping a link to the Tooltip docs page for reference: https://primer.style/react/Tooltip

mperrotti:
@khiga8 - what do you mean by "static element"? Do you mean that it's an element that isn't focusable?

khiga8:
@mperrotti Yep! By static, I meant non-interactive, non-focusable.

khiga8:
I keep saying static but non-interactive is probably a better term :D

khiga8:
From [Tooltip alternatives](https://primer.style/design/accessibility/tooltip-alternatives):

> Tooltips should never be set on non-interactive elements (for example, `div`, `span`, `p`), and should only ever be set on interactive elements (for example, `button`, `a`). Tooltips on non-interactive elements are not accessible to keyboard users and screen reader users.

mperrotti:
@khiga8 - is it ok to put a tooltip on a non-interactive element that has `tabindex=0`? That would at least make it focusable

khiga8:
@mperrotti No, tooltip must be on a natively interactive control. I had the same question when investigating tooltip a while back, and here's a bit more context from @jscholes on why it would not be appropriate [from an internal issue comment](https://github.com/github/accessibility/issues/331#issuecomment-952157307): 

> tooltip triggering should be limited to proper interactive, focusable controls.  If they are triggered on a `span`, `div`, `p` or whatever, making those focusable doesn't really fix any of the problems, apart from making them reachable from the keyboard.  But for users who arent sequentially navigating with Tab, the behaviour is still completely non-obvious, e.g.:


> 1. most screen reader users will never be able to trigger the announcement of the accessible description, because there is no interactive element for them to navigate directly to;
> 2. if the text is in the DOM after the textual element, there is nothing to semantically link what will just be two pieces of plain text;
[...]

PR #2695, opened 2022-12-15
Minor additional tooltip docs updates

khiga8:
Relates to: https://github.com/primer/react/issues/2595

Hi! I wanted to make a minor updates to the tooltip docs to suggest looking at our alternatives docs. Also, we should remove mention of relying on `title` attribute since this attribute is considered to not be accessible since it only appears on mouse hover. We have a linter against it in ERB  ([no-title-attribute](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-title-attribute-counter.md)) and will likely make one for React too.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/Tooltip.md
+++ b/docs/content/Tooltip.md
@@ -10,9 +10,9 @@ The Tooltip component adds a tooltip to add context to interactive elements on t
 
 Tooltips as a UI pattern should be our last resort for conveying information because it is hidden by default and often with zero or little visual indicator of its existence.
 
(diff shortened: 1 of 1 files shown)


joshblack (approved):
Thanks for doing this! 🙌 

Small aside, feel free to remove the changeset as this isn't tied to a specific release. We can add the `skip changeset` label in these kinds of scenarios for docs changes.

PR #2700, opened 2022-12-16
ci: add deploy pages experiment

joshblack:
This PR adds in an experiment for deploying multiple artifacts to GitHub Pages for deploy previews. This would replace the existing `deploy_preview.yml` and `deploy_preview_forks.yml` workflow. 

The proposed flow would be that the `CI` workflow uploads artifacts that it would like to have deployed to GitHub pages, including:

- A `storybook` preview
- A `docs` preview
- Reports from playwright for Visual Regression Testing
- Reports from playwright for Axe results

_Note: These artifacts could be extended in the future if there are other items we would like to have hosted through the GitHub Pages preview._

The proposed workflow will run once the `CI` workflow completes and will attempt to download each of these artifacts. If any individual artifact is available, it will deploy to GitHub Pages. If none are available (because the CI workflow failed), then it will not create a deployment.

This workflow uses the `workflow_run` event to deploy into the GitHub Pages environment. We only interact with static artifacts to prevent any kind of access to secrets that would run against arbitrary code.

Curious what you all think or if you have any suggestions or alternatives that could help out with this use-case 👀


joshblack:
cc @colebemis @siddharthkp @broccolinisoup curious what you all think! 👀

langermank (approved):

.github/workflows/deploy_pages.yml
```diff
@@ -0,0 +1,215 @@
...
+  checks: write
+
+jobs:
+  upload:
+    if: ${{ github.repository == 'primer/react' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create github-pages directory
+        run: mkdir github-pages
+      - name: Download storybook
```
broccolinisoup, line 23: I am not sure if I understand why we need to download storybook here 🤔  We won't deploy it to github-pages and we separately deploy it as well. I am sure I miss something here, would be great to clarify this 😊

.github/workflows/deploy_pages.yml
```diff
@@ -0,0 +1,215 @@
...
+    runs-on: ubuntu-latest
+    needs: upload
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    outputs:
```
broccolinisoup, line 179: How would the output look with these changes? Would it go to the main page of the docs as we have now?

Since you and @langermank brought the Storybook view deployment link in, I can't help but think accessing all these artefacts' links (Axe and VRT results) on the PR page 😋

broccolinisoup (commented):
I love it 💖  Just left a few questions - Thanks @joshblack ✨

.github/workflows/deploy_pages.yml
joshblack: @broccolinisoup the way I understand it, for GH Pages deploy previews we will need to download all the artifacts that should be in the preview and then place them under this `github-pages` directory which will act as the root for serving files.

My thought process was that the output of `docs` would be the default set of files served when visiting a deploy preview URL, and then we would add folders for each sub-resource like `/storybook` for Storybook, `/vrt` for Visual Regression Testing, `/axe` for Axe, etc.

Hope this makes sense, let me know if you have any other questions or if I'm misunderstanding anything here 👀

.github/workflows/deploy_pages.yml
joshblack: @broccolinisoup should be the same as it is currently! (Or at least that's my intent haha) 🤞

PR #2710, opened 2022-12-19
1518 react treeview feedback from screen reader

green6erry:
Describe your changes here.

Closes [1518](https://github.com/github/primer/issues/1518) as somewhat of a bandaid for the issue. The reason this is happening is from a larger issue with the api. For now, the aria message is now updating the status correctly based on which branch is expanded.

### Screenshots

Before:
image
`Directory with async items` is expanded, but the status reads that that the last child, `Nested Directory` is empty. This is both not true, and irrelevant to expanding `Directory with async items`

After:
image
Now when `Directory with async items` is expanded, the status accurately reads that it has loaded.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/TreeView/TreeView.tsx
```diff
@@ -402,6 +402,7 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
           aria-labelledby={labelId}
           aria-describedby={`${leadingVisualId} ${trailingVisualId}`}
           aria-level={level}
+          // aria-expanded={isSubTreeEmpty ? undefined : isExpanded}
```
colebemis, line 405:
```suggestion
```

colebemis:
> The reason this is happening is from a larger issue with the API.

Could you expand on this? What do you think the root cause of this issue is?

src/TreeView/TreeView.tsx
```diff
@@ -525,7 +533,7 @@ const SubTree: React.FC<TreeViewSubTreeProps> = ({count, state, children}) => {
 
   // Announce when content has loaded
   React.useEffect(() => {
-    if (state === 'done') {
+    if (isPending && state === 'done') {
```
colebemis, line 536: One of the problems @jsholes mentioned in the original issue was that announcements were being made with user interaction. Does this change fix that?

I wonder if we should only announce and update when the `state` is `"done"` and the item changes from not-expanded to expanded.

.changeset/rotten-hairs-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+aria status description is now accurate
```
colebemis, line 5: It's usually helpful to mention which component you're changing in the changeset
```suggestion
TreeView: aria status description is now accurate
```

src/TreeView/TreeView.tsx
green6erry: Hey @colebemis, assuming that you mean that the announcements you're referring to were the ones about the last child directory when the parent was being expanded, yes, this is a fix for that.

What you're describing sounds like the current behavior for an async folder. The status of "done" isn't mentioned unless the person expands the folder for the first time. Is this what you mean?

.changeset/rotten-hairs-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
```
colebemis, line 2:
```suggestion
'@primer/react': patch
```

I'd call this a bug fix

src/TreeView/TreeView.tsx
```diff
@@ -525,7 +532,7 @@ const SubTree: React.FC<TreeViewSubTreeProps> = ({count, state, children}) => {
...
```
colebemis, line 535: As discussed in our zoom call: Could you add a small code comment explaining why `isPending` is necessary?

colebemis (approved):
Thanks for walking me through these changes, @green6erry! Looks good to me. Just left a small suggestion about adding a code comment to explain `isPending`.

PR #2715, opened 2022-12-20
Add fixed `position` prop to Overlay

### LINKED ISSUE #2713: Add position prop to Overlay to allow better fixed position rendering
### Description

Following up from a slack thread, it would be great if we could add a `position` prop to the overlay component so we don't need to do math to change which side the overlay component renders on.

For example, [this story has the overlay rendering on the right side using some math](https://github.com/primer/react/blob/d887586133bf679a1b008f49b112cae39d764c91/src/stories/Overlay.stories.tsx#L391) which seems inefficient especially if we don't know the width of the Overlay. 

But we can achieve the same result if we add a new prop called `position` with the values `left` or `right`. From there, the user can customize the behavior of the `left`/`right` values with the `sx` prop

Today, this code does not work as one would expect — it will still render on the left of the screen

```jsx
<Overlay 
  sx={{
    position: "fixed",
    right: 0,
  }}
>
  Example Overlay
</Overlay>
```

My proposed change would allow a user to specify a side (with a default value of `left` to keep in line with existing behavior)

```jsx
<Overlay 
  position="right"
>
  Example Overlay
</Overlay>
```

### Steps to reproduce

1. Create an Overlay component
2. Try to get it to render on the right of the screen — particularly with an auto width content 

Sandbox: https://codesandbox.io/s/great-murdock-6yjjpd

### Version

v35.10.0

### Browser

Chrome

ajhenry:
This adds the following CSS properties as props to the Overlay prop to customize its position behavior. 

`position`
`top` (modified)
`left` (modified)
`right`
`bottom`

I added tests and a story to demonstrate the new behavior.

Closes #2713 

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -190,8 +190,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
     }, [anchorSide, slideAnimationDistance, slideAnimationEasing, visibility])
 
     // To be backwards compatible with the old Overlay, we need to set the left prop if x-position is not specified
-    const leftPosition: React.CSSProperties =
-      left === undefined && right === undefined ? {left: 0} : left ? {left: `${left}px`} : {left}
+  const leftPosition: React.CSSProperties = left === undefined && right === undefined ? {left: 0} : {left}
 
     return (
       <Portal containerName={portalContainerName}>
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -190,7 +190,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
     }, [anchorSide, slideAnimationDistance, slideAnimationEasing, visibility])
 
     // To be backwards compatible with the old Overlay, we need to set the left prop if x-position is not specified
-  const leftPosition: React.CSSProperties = left === undefined && right === undefined ? {left: 0} : {left}
+    const leftPosition: React.CSSProperties = left === undefined && right === undefined ? {left: 0} : {left}
 
     return (
       <Portal containerName={portalContainerName}>
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -17,6 +17,7 @@ type StyledOverlayProps = {
   maxHeight?: keyof Omit<typeof heightMap, 'auto' | 'initial'>
   visibility?: 'visible' | 'hidden'
   anchorSide?: AnchorSide
+  position?: 'left' | 'right'
 } & SxProp
 
 const heightMap = {
@@ -138,6 +139,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
       anchorSide,
       portalContainerName,
       preventFocusOnOpen,
+      position,
       ...rest
     },
     forwardedRef,
@@ -180,6 +182,16 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
       )
     }, [anchorSide, slideAnimationDistance, slideAnimationEasing, visibility])
 
+    // Calculate position based on left and position props
+    let calculatedPosition: Record<string, string> = {left: `${left || 0}px`}
+    if (position && left === undefined) {
+      if (position === 'left') {
+        calculatedPosition = {left: '0px', position: 'fixed'}
+      } else {
+        calculatedPosition = {right: '0px', position: 'fixed'}
+      }
+    }
+
     return (
       <Portal containerName={portalContainerName}>
         <StyledOverlay
@@ -191,7 +203,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
           style={
             {
               top: `${top || 0}px`,
-              left: `${left || 0}px`,
+              ...calculatedPosition,
               '--styled-overlay-visibility': visibility,
             } as React.CSSProperties
           }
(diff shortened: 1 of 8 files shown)


colebemis:
@ajhenry Thank you for opening up this PR!

How do you feel about this prop API instead:

```diff
+ position: 'absolute' | 'fixed' (default: 'absolute')
  top?: number
  left?: number
+ right?: number (default: undefined)
+ bottom?: number (default: undefined)
```

If it feels a little more consistent with CSS.

ajhenry:
@colebemis That makes a lot of sense to me!

Looking at it, it would probably make the most sense just to remove the `left` prop altogether as `position` and `left|right|top|bottom` will be added via the `sx` prop that's on the object

ajhenry:
@colebemis friendly bump to see if this is what you're looking for 😄

lesliecdubs:
Hi @ajhenry, thanks for working on this! @colebemis is out of office this week. 

Will it block you if he gets back to you on/around January 17 when he's back? If so, let me know and I can loop in another team member to review.

ajhenry:
@lesliecdubs No problem, thanks for the heads up! This isn't blocking anything so we're good

colebemis (commented):
These changes look great. That is exactly what I had in mind when I wrote my previous comment 👍 

One thing before I approve and merge the PR: Can you update the docs to match your updated code?

ajhenry:
@colebemis Updated the docs! Let me know if I need to do anything else 🚀

colebemis:
This failing test looks related to these changes:

CleanShot 2023-01-20 at 11 22 45@2x

https://github.com/primer/react/actions/runs/3970131563/jobs/6805498589#step:8:606

ajhenry:
Taking a look now

ajhenry:
So this is failing because of the way we try to parse coordinates for the `AutoSuggestions` component.

The following calculation to determine line height is returning `NaN` for specific cases.
https://github.com/primer/react/blob/4dcf658cf6155d44bac4b79556b38dcbe954f344/src/drafts/utils/character-coordinates.ts#L80-L82

When `computed.fontSize` is `normal` we fallback to `computed.fontSize` which can be blank. The problem with this is that parsing a blank string results in NaN. 

```js
parseInt("")
-> NaN
```

Should we default to 1 when fontSize is blank?

For example, when changing that util function to be the following, we no longer have a failing test
```js
// Lineheight is either a number or the string 'normal'. In that case, fall back to a
// rough guess of 1.2 based on MDN: "Desktop browsers use a default value of roughly 1.2".
const lineHeight = isNaN(parseInt(computed.lineHeight))
  ? parseInt(computed.fontSize === '' ? '1' : computed.fontSize) * 1.2
  : parseInt(computed.lineHeight)
```

src/Overlay.tsx
```diff
@@ -17,6 +17,11 @@ type StyledOverlayProps = {
   maxHeight?: keyof Omit<typeof heightMap, 'auto' | 'initial'>
   visibility?: 'visible' | 'hidden'
   anchorSide?: AnchorSide
+  position?: React.CSSProperties['position']
+  top?: React.CSSProperties['top']
+  left?: React.CSSProperties['left']
+  right?: React.CSSProperties['right']
+  bottom?: React.CSSProperties['bottom']
```
colebemis, line 24: Looks like these props aren't used in the definition of the styled component

```suggestion
```

src/Overlay.tsx
```diff
@@ -179,6 +193,10 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
       )
     }, [anchorSide, slideAnimationDistance, slideAnimationEasing, visibility])
 
+    // To be backwards compatible with the old Overlay, we need to set the left prop if x-position is not specified
+    const leftPosition: React.CSSProperties =
+      left === undefined && right === undefined ? {left: 0} : left ? {left: `${left}px`} : {left}
```
colebemis, line 198: I think this can be simplified to something like this:

```suggestion
  const leftPosition: React.CSSProperties = left === undefined && right === undefined ? {left: 0} : {left}
```

Numbers are assumed to be `px` when passed to the `style` prop.

colebemis (commented):
Did some digging and this should fix the failing test:

```diff
diff --git a/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx b/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
index d29b5e122..53e64d282 100644
--- a/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
+++ b/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
@@ -198,8 +198,8 @@ const InlineAutocomplete = ({
         inputRef={inputRef}
         onCommit={onCommit}
         onClose={onHideSuggestions}
-        top={suggestionsOffset.top}
-        left={suggestionsOffset.left}
+        top={suggestionsOffset.top || 0}
+        left={suggestionsOffset.left || 0}
         visible={suggestionsVisible}
         tabInsertsSuggestions={tabInsertsSuggestions}
       />
 ```

The `NaN` was probably an issue before this PR, but the previous implementation covered up the issue by doing `top: `${top || 0}px`.

src/Overlay.tsx
ajhenry: Perfect!

ajhenry:
I think this is ready for action @colebemis

colebemis (approved):
🥳

PR #2561, opened 2022-11-10
docs(testing): Update testing contribution documentation

broccolinisoup:
Closes [#1508](https://github.com/github/primer/issues/1508)

We don't have a proper contribution docs for our testing practises. It would be beneficial to have some sort of guideline to see the tools we use, testing structure, the general practises and such. As a new comer to the team, I had some struggles to understand our testing guidelines and @joshblack and I were chatting about it so we thought it would be helpful to have a testing documentation. This PR is intentionally not at a final the stage. It would be great if you could provide feedback and we can improve the current draft and release. 🙌🏼 

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2729, opened 2022-12-30
Hidden: Refactor the  component to use the new `getBreakpointsDeclarations` util function and add missing stories and tests

broccolinisoup:
Closes https://github.com/github/primer/issues/1664

`Hidden` component is being used in `PageHeader` and working as expected. After improving the way we manage the responsive visibility on `PageHeader` via `getBreakpointDeclarations` utility function to reduce the layout shift in SSR - I applied the same strategy to `Hidden` component. I refactored it to use the `getBreakpointDeclarations` function rather than `useResponsiveHook` as this is something can be achieved with CSS as well.

I also wrote unit tests, added docs and  stories. 

### Screenshots

No visual changes. All examples are available on storybook and docs. 

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti (commented):
Looking good! I just found one small bug in the PageHeader story.

When I resize the PageHeader story to a narrow viewport, I get an error:
Screenshot 2023-01-13 at 1 52 00 PM

src/Hidden/Hidden.tsx
```diff
@@ -9,18 +10,39 @@ export type HiddenProps = {
   children: React.ReactNode
 }
 
+function isArray(value: Array<string> | string): value is Array<string> {
+  return Array.isArray(value)
+}
+
 export const Hidden = ({on: hiddenViewports, children}: HiddenProps) => {
```
joshblack, line 17: This is unrelated to this PR, but from a naming scheme `on` for breakpoints feels like an awkward phrase to me. Typically `on` is used for actions/events (like event handlers) and `when` or `only` is used for conditions (like breakpoints). I'm curious if there would be interest in revisiting this and using terms like:

```tsx
<Hidden when="narrow" />
<Hidden above="narrow" />
<Hidden below="wide" />
<Hidden breakpoint="wide" />
<Hidden breakpoints={['narrow', 'regular']} />
```

src/Hidden/examples.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+  parameters: {
+    layout: 'fullscreen',
+    controls: {expanded: true},
+  },
+  args: {},
+}
+
+// See if there is an interest to take this into global params
+
+const PrimerViewports = {
```
joshblack, line 20: Similar to the shared note above, would be awesome if we could derive this from a shared spot to keep everything in sync (and pull from when primitives are ready as Mike mentions)

src/Hidden/features.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  title: 'Drafts/Components/Hidden/Features',
+  component: Hidden,
+}
+export default meta
+
+export const HideContent = () => (
+  <Box>
+    <Hidden on={['narrow']}> This value is shown in regular and wide viewports</Hidden>
+    <Hidden on={['regular']}> This value is shown in narrow and wide viewports</Hidden>
+    <Hidden on={['wide']}> This value is shown in narrow and regular viewports</Hidden>
```
joshblack, line 16: For these common cases it might be nice to call out that you can pass a `string` and it should work as expected.

```suggestion
    <Hidden on="narrow"> This value is shown in regular and wide viewports</Hidden>
    <Hidden on="regular"> This value is shown in narrow and wide viewports</Hidden>
    <Hidden on="'wide"> This value is shown in narrow and regular viewports</Hidden>
```

src/Hidden/features.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  </Box>
+)
+
+export const RenderContentResponsively = () => (
+  <Box>
+    <Hidden on={['narrow']}>
```
joshblack, line 22:
```suggestion
    <Hidden on="narrow">
```

src/Hidden/features.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import {Meta} from '@storybook/react'
+import React from 'react'
+import {Hidden} from './Hidden'
+import {Box, Button} from '..'
+
+const meta: Meta = {
+  title: 'Drafts/Components/Hidden/Features',
+  component: Hidden,
+}
+export default meta
```
joshblack, line 10:
```suggestion
export default {
  title: 'Drafts/Components/Hidden/Features',
  component: Hidden,
} as Meta
```

joshblack: Random, should we be using `ComponentMeta` and `ComponentMeta<typeof Hidden>` now in storybook? 🤔 Looking at the TS examples in: https://storybook.js.org/docs/react/api/csf

joshblack (changes requested):
Looking great! Thanks for taking the time to convert things over 🙏 Left some comments, let me know what you think!

src/Hidden/Hidden.test.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react'
+import '@testing-library/jest-dom/extend-expect'
+import {render} from '@testing-library/react'
+import {Hidden} from '.'
+import MatchMediaMock from 'jest-matchmedia-mock'
+import {behavesAsComponent, checkExports, renderStyles, checkStoriesForAxeViolations} from '../utils/testing'
+
+let matchmedia: MatchMediaMock
+describe('Hidden', () => {
+  beforeAll(() => {
```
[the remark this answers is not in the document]
broccolinisoup: That makes sense! I agree, easier to scan and read when there are lines between. I am not very particular about it, so happy to update! 

Would that makes sense to use an eslint rule? 👀 I found [this](https://github.com/dangreenisrael/eslint-plugin-jest-formatting)

src/Hidden/Hidden.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    matchmedia = new MatchMediaMock()
```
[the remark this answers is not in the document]
broccolinisoup: Ah sorry! Yeah totally!

src/Hidden/Hidden.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    Hidden,
+  })
+  it('renders `on` prop as expected', () => {
+    const {container} = render(
+      <Hidden on={'narrow'}>
+        <div>Hidden on narrow</div>
+      </Hidden>,
+    )
+    expect(container).toMatchSnapshot()
+  })
```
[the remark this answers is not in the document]
broccolinisoup: Oh, how come? Sorry, I didn't realise this before.

src/Hidden/Hidden.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+  it('renders the styles as expected when a single viewport value is provided as a string via `on` prop', () => {
+    const expectedStyles = {
+      [`@media screen and (min-width:768px)`]: {
```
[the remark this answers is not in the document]
broccolinisoup: Ah yes, that makes perfect sense! When I wrote this, I was thinking "they should be hard-coded so that I can catch if the breakpoints fail too". However, now you said it and I realised that I am not testing the breakpoints here 😅 Thanks for pointing this out 🙌🏻 

In terms of where to drive them from, I used these values multiple times in `PageHeader` too and I always imported them from [useResponsiveHook](https://github.com/primer/react/blob/main/src/hooks/useResponsiveValue.ts#L9) . I updated the code to imported from `useResponsiveHook` again but this file might not be the best place now to keep them.  Would [theme-preval](https://github.com/primer/react/blob/main/src/theme-preval.js) make sense at all? If no, happy to move them over `src/utils/breakpoints.ts`!

src/Hidden/examples.stories.tsx
broccolinisoup: I was actually thinking to move them over to the global storybook parameters. Because the current preview screen options are `['Small mobile', 'Large mobile', 'Tablet']` and they don't mean much (at least to me) - and what breakpoints they use are not explicit either. I think it would be more useful to use the breakpoints we use in Primer. What do you think? cc @mperrotti as well as he mention below

I pushed a [commit](https://github.com/primer/react/pull/2729/commits/41e4b635943dd90384dbbc4b16caec78d88ecf26) for this change. Let me know what you think 🙌🏻 

Story book responsive screen option list that includes values 'XSmall', 'Small', 'Medium', 'Large', 'XLarge', 'XXLarge'

src/Hidden/Hidden.test.tsx
```diff
@@ -36,7 +37,8 @@ describe('Hidden', () => {
   })
   it('renders the styles as expected when a single viewport value is provided as a string via `on` prop', () => {
     const expectedStyles = {
-      [`@media screen and (min-width:768px)`]: {
+      // `.replace` is used because renderStyles return the JSON object without a space after the column
+      [`@media screen and ${viewportRanges.regular.replace(': ', ':')}`]: {
```
broccolinisoup, line 41: `replace` 😢  Any alternatives?

src/Hidden/Hidden.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
-import {useMedia} from '../hooks/useMedia'
-import {viewportRanges} from '../hooks/useResponsiveValue'
+import {ResponsiveValue} from '../hooks/useResponsiveValue'
+import {getBreakpointDeclarations} from '../utils/getBreakpointDeclarations'
+import Box from '../Box'
 
 type Viewports = 'narrow' | 'regular' | 'wide'
```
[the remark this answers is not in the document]
broccolinisoup: Agreed! Thanks 🙌🏻

src/Hidden/Hidden.tsx
```diff
@@ -9,18 +10,39 @@ export type HiddenProps = {
...
-    show = false
-  } else if (isRegularViewport && (hiddenViewports === 'regular' || hiddenViewports.indexOf('regular') > -1)) {
-    show = false
-  } else if (isWideViewport && (hiddenViewports === 'wide' || hiddenViewports.indexOf('wide') > -1)) {
-    show = false
+  let responsiveObj: ResponsiveValue<boolean | number | string> | null = null
+  // If we receive array of viewports, we need to tranform this array into ResponsiveValue object
+
+  // For array type
+  if (isArray(hiddenViewports)) {
```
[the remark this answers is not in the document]
broccolinisoup: Love it! It is much cleaner 🙌🏻

src/Hidden/features.stories.tsx
broccolinisoup: This is a great point Josh! Feels like we should be moving onto `ComponentMeta` and `ComponentMeta<typeof Hidden>` .  Love this summary [here](https://github.com/storybookjs/storybook/discussions/16650#discussioncomment-1617480). I saw in multiple places we import the types explicity from the component. This way it should be cleaner 🙌🏻

broccolinisoup:
> Looking good! I just found one small bug in the PageHeader story.
> 
> When I resize the PageHeader story to a narrow viewport, I get an error: Screenshot 2023-01-13 at 1 52 00 PM

This is interesting, I cannot seem to replicate it. Do you mind sharing a screen recording?

src/Hidden/Hidden.tsx
broccolinisoup: Sure, I am happy to re-think and discuss about it. Just to clarify that it seems like we have [breakpoints](https://primer.style/primitives/spacing#breakpoints) and [viewport ranges](https://primer.style/primitives/spacing#viewport-ranges) and we use viewport ranges for the `on` prop.

Considering that, using `when` makes sense to me. Tagging @mperrotti and @colebemis to get their opinion too 🙌🏻

src/Hidden/Hidden.test.tsx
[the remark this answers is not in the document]
joshblack: Made a quick draft PR with some thoughts over at: https://github.com/primer/react/pull/2780

Let me know if it'd be helpful to write more or schedule a group chat for this! Definitely don't want to step on any toes if they've been helpful.

src/Hidden/Hidden.test.tsx
joshblack: @broccolinisoup could we have a `breakpoint` helper like:

```suggestion
      [breakpoint(viewportRanges.regular)]: {
```

Let me know if I'm missing anything! For context, this was the setup/structure I was thinking of from Carbon: https://github.com/carbon-design-system/carbon/blob/main/packages/layout/src/index.js#L43-L83 definitely don't mean we should copy it by any means, just was thinking about how the breakpoints were structured and the helper functions that made it simple to emit media queries in JS

src/Hidden/Hidden.test.tsx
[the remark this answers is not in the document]
joshblack: @broccolinisoup I'm down either way! Happy for it just to be a formatting thing we do if it's easier to not make it a rule.

mperrotti:
Here you go @broccolinisoup. This is from the [preview deployment](https://primer-01b2eb7ae8-13348165.drafts.github.io/storybook/?path=/story/drafts-components-pageheader-examples--pull-request-page).

https://user-images.githubusercontent.com/2313998/213017797-3f7aac79-d707-413b-b2b4-b1fc43fd420d.mp4

src/Hidden/Hidden.test.tsx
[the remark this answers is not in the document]
broccolinisoup: Ok sounds good to me too!

src/Hidden/Hidden.test.tsx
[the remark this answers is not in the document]
broccolinisoup: Thank you so much for the ADR! It makes perfect sense.

>They don't seem to add much value in our existing setup

I thought it was due to a "defect" or some misconfiguration in our current setup and for some reason the comparison doesn't work with initial snapshot vs the future version. Sorry I over thought this!

I think this ADR is a great place to land on a decision about snapshot testing strategy as it came up [before](https://github.com/primer/react/pull/2561#discussion_r1021356033). I appreciate your wrote it up 🙏🏻  I'll add follow up comments on this.

broccolinisoup:
>Here you go @broccolinisoup. This is from the [preview deployment](https://primer-01b2eb7ae8-13348165.drafts.github.io/storybook/?path=/story/drafts-components-pageheader-examples--pull-request-page).

@mperrotti Oh right!! This is on the new canvas! Sorry I missed that. Actually, this is a bug (more like an uncovered case) in UnderlineNav! Timing is a coincidence that I am synchronously working on [it](https://github.com/primer/react/pull/2770).

src/Hidden/Hidden.test.tsx
broccolinisoup: Thank you! I still needed `replace` but took your suggestion into account and came up with something similar. Here is the [commit](https://github.com/primer/react/pull/2729/commits/37fd3b854c9d4d4645ff8b7be3deb8880df4b7eb) 

If that looks good, I will replace the `mediaQueriesForViewportRanges` with the hard-coded values. Also, open to name suggestions 🙌🏻

src/Hidden/Hidden.test.tsx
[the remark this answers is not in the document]
broccolinisoup: The snapshot checking also comes from `behavesAsComponent` - I'll wait for other folks to weigh in on the [ADR PR](https://github.com/primer/react/pull/2780) before removing it from the util testing function - just to make sure everyone is on the same page 🙌🏻

broccolinisoup:
@joshblack just renamed a few things to reduce confusion around breakpoints/ranges and refactored `PageHeader` and `getBreakpointsDeclarations` function to use the media queries from a shared place (src/utils/layout.ts). ([commit](https://github.com/primer/react/pull/2729/commits/da4353b95b62ebd8bbdf791d590dbe9b3b3fd5de))

I'd like to merge this to move forward with the accessibility eng review for PageHeader - but please let me know if you have any concern - happy to address them in a following PR! - thanks so much again for your suggestions 🙏🏻

joshblack (approved):
🎉

PR #2730, opened 2023-01-02
test(setup): setup failOnConsole

joshblack:
Use `jest-fail-on-console` to fail tests when `console.*` methods are used. This helps to:

- Make sure the test suite runs without any issues
- Reduce confusion from contributors when logs are emitted when tests run

Specifically, this helps to make sure that no `console.{error,warn}` calls from React are emitted during tests.

**Updated**

- [x] TextInput.test.tsx
- [x] CheckboxOrRadioGroup.test.tsx
- [x] MarkdownEditor.test.tsx
- [x] ConfirmationDialog.test.tsx
- [x] MarkdownViewer.test.tsx
- [x] SegmentedControl.test.tsx
- [x] CheckboxGroup.test.tsx
- [x] RadioGroup.test.tsx
- [x] FormControl.test.tsx


dgreif (approved):
Looks great! I've found this adds a ton of clarity for dotcom, glad to see it added in PRC as well 🎉

src/utils/test-matchers.tsx
```diff
@@ -99,3 +100,13 @@ expect.extend({
...
+
+const failOnAllConsoleMessages = process.env.CI === 'true'
+
+failOnConsole({
+  shouldFailOnWarn: true,
+  shouldFailOnError: true,
+  shouldFailOnDebug: failOnAllConsoleMessages,
+  shouldFailOnInfo: failOnAllConsoleMessages,
+  shouldFailOnLog: failOnAllConsoleMessages,
+})
```
broccolinisoup, line 112: cool!

src/__tests__/ConfirmationDialog.test.tsx
```diff
@@ -122,14 +122,31 @@ describe('ConfirmationDialog', () => {
...
+    // environment
+    if (REACT_VERSION_LATEST) {
+      // eslint-disable-next-line jest/no-conditional-expect
+      expect(spy).toHaveBeenCalledTimes(1)
+      // eslint-disable-next-line jest/no-conditional-expect
+      expect(spy).toHaveBeenCalledWith(
+        expect.stringContaining('Warning: ReactDOM.render is no longer supported in React 18'),
+      )
+    }
+    spy.mockRestore()
```
broccolinisoup, line 150: I like this case!

src/UnderlineNav2/UnderlineNav.test.tsx
```diff
@@ -112,6 +114,14 @@ describe('UnderlineNav', () => {
...
     await user.click(item)
     expect(onSelect).toHaveBeenCalledTimes(1)
+
+    if (REACT_VERSION_LATEST) {
+      expect(spy).not.toHaveBeenCalled()
+    } else {
+      // Warning: It looks like you're using the wrong act() around your test interactions
+      expect(spy).toHaveBeenCalledTimes(2)
+    }
+    spy.mockRestore()
```
broccolinisoup, line 124: Could you tell me a little bit more about this test? Not sure if I understand it well. Thank you 🙏🏻

broccolinisoup (approved):
Looks great! 🚀

src/UnderlineNav2/UnderlineNav.test.tsx
joshblack: @broccolinisoup would you have time to pair up on this next week? Would love to walk through this case with you, it's super strange 🤔

I've encountered this with a couple components and it seems to happen where there is a mix of snapshot testing from the test helpers and react testing library. I don't think anything is wrong with the test itself and instead it's the error showing up due to this mix-and-match of `react-test-renderer` and `@testing-library/react`. 

In `@testing-library/react`, there is an "act-compat" module that seems to be picking the wrong thing and causing this error to get thrown but I've been unable to reproduce what specifically is causing it.

PR #2749, opened 2023-01-09
test(e2e): update waitForSelector for storybook

### LINKED ISSUE #2743: Update VRT workflow to properly wait for assets and page load
We're running into two flakiness situations with VRT that impacted the release this week:

- https://github.com/primer/react/actions/runs/3855032716
  - Timing issue for UnderlineNav
- https://github.com/primer/react/actions/runs/3855021520
  - The source is incorrectly a blank avatar
- https://github.com/primer/react/actions/runs/3855032716
  - The source has a spinner as a snapshot

The second two of these should be handled by `waitForSelector` and `waitForImages` but they don't seem to be having the intended effect. The first one may be related to React 18 timing but this is uncertain.

joshblack:
Closes #2743

Update the `page.waitForSelector` to include two specific checks:

> `body.sb-show-main:not(.sb-show-preparing-story)`

It seems like there is a condition that we run into where the `<body>` element has both `.sb-show-main` and `sb-show-preparing-story`. This check makes sure that we _only_ have `.sb-show-main`

> `#root > *`

This will check to make sure that content has been loaded within the story so that helpers like `waitForImages` work off of loaded content

### DIFF
--- a/e2e/test-helpers/storybook.ts
+++ b/e2e/test-helpers/storybook.ts
@@ -36,6 +36,7 @@ export async function visit(page: Page, options: Options) {
   }
 
   await page.goto(url.toString())
-  await page.waitForSelector('.sb-show-main')
+  await page.waitForSelector('body.sb-show-main:not(.sb-show-preparing-story)')
+  await page.waitForSelector('#root > *')
   await waitForImages(page)
 }


broccolinisoup (approved):
I love this PR so much! Thanks for writing a great description to explain the changes as well ✨

PR #2750, opened 2023-01-09
Showing the tooltip on focus-within

jdrush89:
Show the tooltip on focus-within. There were rules to show the tooltip on focus, but those will never get triggered since the tooltip element itself is not focusable. It should be the tooltip's child getting focus that triggers it showing.
This is required for [Mouse-Keyboard parity](https://github.com/github/accessibility-audits/issues/2966) which states that elements that show on hover should also show on focus.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/poor-melons-share.md
+++ b/.changeset/poor-melons-share.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Making onClick of ActionList.LinkItem functional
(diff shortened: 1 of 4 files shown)


iansan5653 (approved):
I 100% support this - it would fix a lot of headaches and hacky workarounds. Thanks!

joshblack (approved):

PR #2733, opened 2023-01-03
Create deprecating-components.md

broccolinisoup:
Closes https://github.com/github/primer/issues/1593

This PR is based on deprecating React components [ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-006-drafts.md) , my notes from chatting with PRC folks about the process as well as my first time deprecating a component experience. So this is a pretty much raw document that I would love to get your thoughts on 🙏🏼 

Note: I copy-pasted quite a few info from the [ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-006-drafts.md), thanks to @siddharthkp :) 

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2753, opened 2023-01-09
docs: add versioning guide

joshblack:
Add a versioning guide to `contributor-docs/versioning.md` to see what kind of consensus the team has for types of changes and what semver bumps they correspond to 👀 

Would be awesome if this could be a living document that we can refer to whenever different types of scenarios come up in order to consistently apply our understanding of semver against `@primer/react`.

### DIFF
--- a/contributor-docs/versioning.md
+++ b/contributor-docs/versioning.md
@@ -1,0 +1,119 @@
+# Versioning
+
+<!-- prettier-ignore-start -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+## Table of Contents
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+<!-- prettier-ignore-end -->
+
+## Overview
+
+The Primer team aims to follow
+[Semantic Versioning](https://semver.org/) (semver) for each of the packages
+that we ship. From semver.org, this means that:
+
+> Given a version number MAJOR.MINOR.PATCH, increment the:
+>
+> 1. **MAJOR** version when you make incompatible API changes,
+> 2. **MINOR** version when you add functionality in a backwards compatible
+>    manner, and
+> 3. **PATCH** version when you make backwards compatible bug fixes.
+>
+> _Additional labels for pre-release and build metadata are available as
+> extensions to the MAJOR.MINOR.PATCH format._
+
+As a result, whenever you see a `minor` or `patch` update for a package from the
+Primer you should feel confident that you can update without
+anything breaking in your project. For a full list of changes that the team
+commits to being backwards-compatible, view the [changes](#changes) table
+below.
+
+## Changes
+
+| Category   | Type of change                                                | semver bump |
+| :--------- | :------------------------------------------------------------ | :---------- |
+| Component  | A component is added                                          | `minor`     |
+|            | A component is deprecated                                     | `minor`     |
+|            | A component is removed                                        | `major`     |
+| Props      | A prop is added to a component                                | `minor`     |
+|            | The type of a prop is made more general                       | `minor`     |
+|            | The type of a prop is made more specific                      | `major`     |
+|            | A prop is deprecated                                          | `minor`     |
+|            | A prop is removed from a component                            | `major`     |
+|            | The element to which additional props are added to is changed |             |
+| Markup     | The DOM node that an `id` corresponds to is changed           |             |
+|            | The DOM node that an `aria-label` corresponds to is changed   |             |
+|            | The `role` of a component is changed                          |             |
+| Styles     | The `position` of the outermost element is updated            |             |
+|            | The `display` property of the outermost element is updated    |             |
+|            | A flex or grid property is updated                            |             |
+| Behavior   | Interactions are added to a component                         |             |
+|            | Interactions are removed from a component                     |             |
+| TypeScript | A type is added                                               | `minor`     |
+|            | A type is made more general                                   | `minor`     |
+|            | A type is made more specific                                  | `major`     |
+|            | A type is removed                                             | `major`     |
+| Package    | An entrypoint is added to the package                         | `minor`     |
+|            | An entrypoint is removed from a package                       | `major`     |
+
+### Examples
+
+#### A prop is added to a component
+
+semver bump: **minor**
+
+```diff
+type Props = {
+  propA: string;
++  propB: string;
+};
+
+function ExampleComponent({
+  propA,
++  propB,
+}: Props) {
+  return (
+    <>
+      <span>{propA}</span>
++      <span>{propB}</span>
+    </>
+  );
+}
+```
+
+#### An existing prop is deprecated
+
+semver bump: **minor**
+
+```diff
+type Props = {
+  propA: string;
++  /**
++   * @deprecated This prop will be removed in the next major version of
++   * `@primer/react`. Please use <replacement> instead.
++   */
+  propB: string;
+};
+
+function ExampleComponent({
+  propA,
++  propB,
+}: Props) {
+  return (
+    <>
+      <span>{propA}</span>
++      <span>{propB}</span>
+    </>
+  );
+}
+```
+
+#### The DOM node that an `id` corresponds to is changed
+
+semver bump: **major**
+
+#### The DOM node that an `aria-label` corresponds to is changed
+
+semver bump: **minor**


contributor-docs/versioning.md
```diff
@@ -0,0 +1,131 @@
...
+anything breaking in your project. For a full list of changes that the team
+commits to being backwards-compatible, view the [changes](#changes) table
+below.
+
+## Changes
+
+| Category   | Type of change                                                | semver bump |
+| :--------- | :------------------------------------------------------------ | :---------- |
+| Component  | A component is added                                          | `minor`     |
+|            | A component is deprecated                                     | `minor`     |
```
broccolinisoup, line 46: How would you feel expanding this or rewording?

Because the current practise deprecating components is major but as our development practises get mature, hopefully the deprecation process will be a minor change.

contributor-docs/versioning.md
```diff
@@ -0,0 +1,131 @@
...
+|            | A component is removed                                        | `major`     |
+| Props      | A prop is added to a component                                | `minor`     |
+|            | The type of a prop is made more general                       | `minor`     |
+|            | The type of a prop is made more specific                      | `major`     |
+|            | A prop is deprecated                                          | `minor`     |
+|            | A prop is removed from a component                            | `major`     |
+|            | The element to which additional props are added to is changed |             |
+| Markup     | The DOM node that an `id` corresponds to is changed           |             |
```
broccolinisoup, line 54: Would you say they can be major due to the possibility of consumers using them as selectors for tests for example? I remember we talked about it before - just wanted to make sure if I understand these right.

contributor-docs/versioning.md
```diff
@@ -0,0 +1,131 @@
...
+|            | The DOM node that an `aria-label` corresponds to is changed   |             |
+|            | The `role` of a component is changed                          |             |
+| Styles     | The `position` of the outermost element is updated            |             |
+|            | The `display` property of the outermost element is updated    |             |
+|            | A flex or grid property is updated                            |             |
+|            | A selector is added                                           |             |
+|            | A selector is removed                                         |             |
+|            | The specificity of a selector is raised                       |             |
+|            | The specificity of a selector is lowered                      |             |
```
broccolinisoup, line 63: These are really good!

contributor-docs/versioning.md
```diff
@@ -0,0 +1,131 @@
...
+| Behavior   | Interactions are added to a component                         |             |
+|            | Interactions are removed from a component                     |             |
```
broccolinisoup, line 65: Do you think changes related to responsiveness can go under this?

contributor-docs/versioning.md
```diff
@@ -0,0 +1,131 @@
...
+| TypeScript | A type is added                                               | `minor`     |
+|            | A type is made more general                                   | `minor`     |
+|            | A type is made more specific                                  | `major`     |
+|            | A type is removed                                             | `major`     |
+| Package    | An entrypoint is added to the package                         | `minor`     |
+|            | An entrypoint is removed from a package                       | `major`     |
+
```
broccolinisoup, line 72: Should we add a section for `Accessibility` or things related to accessibility can go under other corresponding sections?

contributor-docs/versioning.md
```diff
@@ -0,0 +1,131 @@
...
+### Examples
```
broccolinisoup, line 73: Love it 💖

broccolinisoup (commented):
Thank you so much for writing this up! I left a few comments, looking forward to hearing your thoughts 🙌🏼

contributor-docs/versioning.md
```diff
@@ -0,0 +1,131 @@
...
+> 3. **PATCH** version when you make backwards compatible bug fixes.
+>
+> _Additional labels for pre-release and build metadata are available as
+> extensions to the MAJOR.MINOR.PATCH format._
+
+As a result, whenever you see a `minor` or `patch` update for a package from the
+Primer you should feel confident that you can update without
+anything breaking in your project. For a full list of changes that the team
+commits to being backwards-compatible, view the [changes](#changes) table
+below.
```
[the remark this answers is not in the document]
joshblack: @broccolinisoup you're totally right, the wording is off. It should just be types of changes. Updating now!

contributor-docs/versioning.md
joshblack: Updated it to be `major` to better describe the current situation, if that works! Let me know 👀

contributor-docs/versioning.md
joshblack: @broccolinisoup I think that would be a great case where it probably should be if the component accepts an `id` prop

contributor-docs/versioning.md
joshblack: Ooo, I like that a ton. Great suggestion 🙌

contributor-docs/versioning.md
joshblack: @broccolinisoup makes sense to me! What kind of situations would be good to place under an `Accessibility` category?

contributor-docs/versioning.md
```diff
@@ -0,0 +1,132 @@
...
+>
+> 1. **MAJOR** version when you make incompatible API changes,
+> 2. **MINOR** version when you add functionality in a backwards compatible
+>    manner, and
+> 3. **PATCH** version when you make backwards compatible bug fixes.
+>
+> _Additional labels for pre-release and build metadata are available as
+> extensions to the MAJOR.MINOR.PATCH format._
+
+As a result, whenever you see a `minor` or `patch` update for a package from the
```
rezrah, line 35:
```diff
- ... a `minor` or `patch` update for a package from the Primer you should feel confident ...
+ ... a `minor` or `patch` update to a package from Primer you should feel confident ...
```

contributor-docs/versioning.md
```diff
@@ -0,0 +1,132 @@
...
+corresponding semver bumps, check out the [changes](#changes) table below.
+
+For a full list of releases, visit our [releases](https://github.com/primer/react/releases) page.
+
+## Changes
+
+| Category   | Type of change                                                | semver bump |
+| :--------- | :------------------------------------------------------------ | :---------- |
+| Component  | A component is added                                          | `minor`     |
+|            | A component is deprecated                                     | `major`     |
```
rezrah, line 47: This one's interesting:

1. In Primer Primitives deprecating something doesn't constitute a material breaking change in the output, but serves as a flag that something will soon be removed and that we have provided an alternative to ease migration
2. During v35 major release, I recall the team suggested in a retro that we don't make deprecations major. We had the same feedback from library users.

contributor-docs/versioning.md
```diff
@@ -0,0 +1,132 @@
...
+| Styles     | The `position` of the outermost element is updated            | `<todo>`    |
+|            | The `display` property of the outermost element is updated    | `<todo>`    |
+|            | A flex or grid property is updated                            | `<todo>`    |
+|            | A selector is added                                           | `<todo>`    |
+|            | A selector is removed                                         | `<todo>`    |
+|            | The specificity of a selector is raised                       | `<todo>`    |
+|            | The specificity of a selector is lowered                      | `<todo>`    |
+| Behavior   | Interactions are added to a component                         | `<todo>`    |
+|            | Interactions are removed from a component                     | `<todo>`    |
+| TypeScript | A type is added                                               | `minor`     |
```
rezrah, line 67: Do you have any examples of this one in mind? I think this could also be a `patch` in certain scenarios?

contributor-docs/versioning.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### An existing prop is deprecated
+
+semver bump: **minor**
+
+```diff
+type Props = {
+  propA: string;
++  /**
++   * @deprecated This prop will be removed in the next major version of
```
rezrah, line 107: ❤️ 

Wondering if these would need to be mopped up at a later date. Is there a mechanism or process you had in mind to do that already?

contributor-docs/versioning.md
joshblack: @rezrah I definitely agree, deprecations in `minor` releases feel more appropriate/align with expectations.

contributor-docs/versioning.md
joshblack: Really just was thinking of a narrow case where a type has been added to the Public API of the package.

In this scenario, I framed it as a `minor` since it would be additive and would require someone to explicitly import it in order to use it.

Let me know if there is a better way to talk about types in this kind of context, I am honestly not sure how to frame type changes in terms of semver so was just applying a `propTypes` mindset to this kind of section 👀

contributor-docs/versioning.md
joshblack: @rezrah I'd honestly love a comment on the issue for next major that is updated every week or whatever cadence with all these with a `@deprecated` comment 😂 

But generally I feel like at the very least if we keep track of deprecations in something like a next major release issue it can help for tracking what's changing, writing migration docs, etc.

contributor-docs/versioning.md
broccolinisoup: > During v35 major release, I recall the team suggested in a retro that we don't make deprecations major. We had the same feedback from library users.

I am very curious to learn more about this. It is very interesting.

We will be deprecating `UnderlineNav` in v36 and it is planned to be a major change. We re-wrote the component entirely and the public API has changed. I don't have much experience with versioning but I am super keen to improve! I am curious to know if it is still possible to refactor `UnderlineNav` and introduce as a minor change with warning the deprecation and providing the new API as an alternative 🤔  Would it worth giving a try or not really at this stage? 

[Current UnderlineNav](https://primer.style/react/UnderlineNav)
[Newly written UnderlineNav](https://primer.style/react/drafts/UnderlineNav2)

rezrah: @joshblack @broccolinisoup I've somehow found the FigJam file I was referencing in my earlier comment. You should receive a link to it in your emails. 

@broccolinisoup - my reading of your situation based on the diagram would be to issue the deprecation notice through comms (docs, changelogs, etc) and a patch (or minor) update of the libraries (for JSDOC `@deprecate` annotations). Main thing is to leave it in-place. We're just issuing notice AoT.

Then wait for the next major release (~couple of weeks) to move it into the deprecated bundle and move UnderlineNavV2 from drafts into main bundle. Eventually few months after, we can remove deprecated component altogether. 

Could this still work for you?

rezrah: @joshblack @broccolinisoup I've somehow found the FigJam file I was referencing in my earlier comment. You should receive a link to it in your emails. 

@broccolinisoup - my reading of your situation based on the diagram would be to issue the deprecation notice through comms (docs, changelogs, etc) and a patch (or minor) update of the libraries (for JSDOC `@deprecate` annotations). Main thing is to leave it in-place. We're just issuing notice AoT.

Then wait for the next major release (~couple of weeks) to move it into the deprecated bundle and move UnderlineNavV2 from drafts into main bundle. Eventually few months after, we can remove deprecated component altogether. 

Could this still work for you?

contributor-docs/versioning.md
rezrah: ah gotcha, thanks for clarifying. I think a type addition that doesn't make a material API change, or something that doesn't require the user to react to can go into a patch. A prominent new feature however, like a new component entirely - which includes new types - would constitute a minor in my mind. I agree that it's nuanced however. Perhaps we can split this up into different type edge cases.

contributor-docs/versioning.md
rezrah: Yeah, that's nice. We could use a CI workflow to grep the codebase for `@deprecated` annotations and then auto update the next major issue. Good idea 👍

contributor-docs/versioning.md
broccolinisoup: > I've somehow found the FigJam file I was referencing in my earlier comment. You should receive a link to it in your emails.

Thanks, this is brilliant  🙌🏼

> Could this still work for you?

@rezrah Thank you so much for clarifying the process! According to your explanation, we are on track. The only thing is different than the plan is updating the old UnderlineNav with the `@deprecate` status as a patch (or minor) update before the major release. Would that still be a good practise without promoting the new UnderlineNav as an alternative component (As this plan to come as a major later)? I think this is what I am trying to wrap my head around.

 Thank you so much 🙏🏼

contributor-docs/versioning.md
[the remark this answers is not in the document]
broccolinisoup: Sorry, I didn't provide an example so it made it blur. I love the way you group by the area and list the out the changes to expect. The only difference I was thinking, instead of randomly listing the changes under a specific area, we could list the patch changes first, minor second and major as the last. 

Instead of

```
| Props      | A prop is added to a component                                | `minor`     |
|            | The type of a prop is made more general                       | `minor`     |
|            | The type of a prop is made more specific                      | `major`     |
|            | A prop is deprecated                                          | `minor`     |
|            | A prop is removed from a component                            | `major`     |
```

We could maybe do

```
| Props      | A prop is added to a component                                | `minor`     |
|            | The type of a prop is made more general                       | `minor`     |
|            | A prop is deprecated                                          | `minor`     |
|            | The type of a prop is made more specific                      | `major`     |
|            | A prop is removed from a component                            | `major`     |
```

Let me know what you think 🙌🏼

contributor-docs/versioning.md
broccolinisoup: I guess things in my mind can certainly go under other categories but maybe explicitly having a section for `Accessibility` feels more inclusive? 🙂 

I was thinking;
- Exposing a mandatory `aria` attribute in the API
- Introducing a11y remediations that result in breaking changes in a current user behaviour
- Updating keyboard interactions

Does any of it sound applicable at all?

contributor-docs/versioning.md
broccolinisoup: Hi! @rezrah just wanted to resurface this, would love to hear your thoughts 😊

broccolinisoup: That sounds great! I appreciate the very detailed explanation as well 🙇🏻‍♀️  I'll capture this info in the deprecation process documentation [PR](https://github.com/primer/react/pull/2733).

PR #2754, opened 2023-01-09
ci: add size summary for package exports

joshblack:
Add a `size` job to the CI workflow that reports on the size of entrypoints by adding it to the summary of the workflow.

In the future, we could have this job report the sizes through a JSON file which can be used to diff between different versions of the package to report on size increase or decrease


broccolinisoup (commented):
I admittedly don't utilize the current size CI check as a metric at all 🙈 I think the new size summary for components would be super helpful to be on track.

- I think it would be very useful for us too to have the data of how the changeset affects the component's size (same as [here](https://github.com/github/engineering/discussions/2698))
- Would it be useful to have the same data for things like hooks/utility functions etc as well? Maybe a "splash zone" of how other components' sizes are affected by the changes of a utility function/hook etc? 👀

joshblack:
@broccolinisoup following up, I think I could get it to work but might take a bit longer than I can squeeze in over the next couple of weeks. Do you think adding it to the summary here works in the interim? Happy to revisit this when I have more time 👀

broccolinisoup (approved):
> @broccolinisoup following up, I think I could get it to work but might take a bit longer than I can squeeze in over the next couple of weeks. Do you think adding it to the summary here works in the interim? Happy to revisit this when I have more time 👀

Sorry I missed the first notification!  _I think_ it is a great start! and we can go from here when we have some free time. Let me know if I can help in any way! Do you think would it be helpful to create an issue for this to track or not really? 

Thank you so much for doing this 🙏🏻   

>For the hooks and utility functions, the hope is that it is showing everything exported from the package over in: https://github.com/primer/react/actions/runs/3877690607#summary-10535942438 Let me know if it's missing anything 👀

Fair enough! My eyes were looking for utility functions that were not exported which I just realised! - this is why I wrote the comment - but from the exported packages perspective, all looking good! 💖

Issue #2724, opened 2022-12-22
onClick of ActionList.LinkItem not getting called


jdrush89:
### Description

I tried updating dotcom to primer-upgrade-35.16 and noticed that the onClick passed to ActionList.LinkItems no longer gets called when clicking on the item.

### Steps to reproduce

```
<ActionList>
  <ActionList.LinkItem onClick={() => console.log('test')}>
    <ActionList.LeadingVisual>
      <LinkIcon />
    </ActionList.LeadingVisual>
    github/primer
  </ActionList.LinkItem>
</ActionList
```
Change the [ActionList.LinkItem example](https://primer.style/react/ActionList#with-links) to this and try clicking the item. The console message does not get printed.

### Version

v35.16

### Browser

Chrome

jdrush89:
@radglob I have a fix here https://github.com/primer/react/pull/2757

PR #2757, opened 2023-01-10
Jdrush89/fix linkitem onclick

jdrush89:
Call the onClick of ActionList.LinkItems when they are clicked. The onClick prop was allowed to be passed in, but was being overridden by ActionList.Item's clickHandler.

Closes https://github.com/primer/react/issues/2724

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/ActionList/LinkItem.tsx
```diff
@@ -39,11 +39,24 @@ export const LinkItem = React.forwardRef(({sx = {}, active, as: Component, ...pr
...
+        const clickHandler = (event: React.MouseEvent) => {
+          onClick && onClick(event)
+          props.onClick && props.onClick(event as React.MouseEvent<HTMLAnchorElement>)
+        }
+        return (
+          <Link
+            as={Component}
+            sx={merge(styles, sx as SxProp)}
+            {...rest}
+            {...props}
```
jdrush89, line 52: The props passed in by the consumer should override the ones set by default by the item. Both onClicks will be called. We could do the same with onKeyPress, but consumers shouldn't really be using onKeyPress since it's deprecated for onKeyDown.

src/__tests__/ActionList.test.tsx
```diff
@@ -155,6 +155,21 @@ describe('ActionList', () => {
...
   })
 
+  it('should call onClick for a link item', async () => {
+    const onClick = jest.fn
+    const component = HTMLRender(
+      <ActionList role="listbox">
+        <ActionList.LinkItem onClick={onClick}>Primer React</ActionList.LinkItem>
+      </ActionList>,
+    )
+    const link = await waitFor(() => component.getByRole('link'))
```
radglob, line 165: Overall, this looks great, but I don't think this test is passing based on this line. It doesn't look like the 'link' role is assigned at all. If you can button that up, I think this is good to go.

Thanks for knocking this out!

radglob (changes requested):

radglob (approved):
Great work on this! Thanks!

joshblack (approved):
Thanks for doing this! 🥳

PR #2762, opened 2023-01-11
Avoid tokens weird type checking/eliding

mattcosta7:
Describe your changes here.

Relates to https://github.com/primer/react/pull/2760

In this pr we omitted `onResize` and `onResizeCapture` as typescript was causing these to elide a mandator unknown type, which was super confusing.

I believe this was caused by the fact that these props aren't equally defined across all polymorphic types, and forwardRef swallows polymorphism, making things wonky. 

I've re-typed these components using the polymorphic helper, and added some type tests to help validate they don't regress.  I think we would be well served extending our type tests further to cover most/all components, as I'm sure others are limited in the same or similar ways

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/gentle-badgers-confess.md
+++ b/.changeset/gentle-badgers-confess.md
@@ -2,4 +2,6 @@
 '@primer/react': patch
 ---
 
-update types for tokens
+Token: Update component type to be PolymorphicForwardRefComponent.
+
+this avoids types being swallowed by forwardRef (which isn't polymorphic)
--- a/.changeset/gentle-badgers-confess.md
+++ b/.changeset/gentle-badgers-confess.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+update types for tokens
--- a/src/Token/Token.tsx
+++ b/src/Token/Token.tsx
@@ -5,10 +5,11 @@ import {defaultSxProp} from '../utils/defaultSxProp'
 import TokenBase, {defaultTokenSize, isTokenInteractive, TokenBaseProps} from './TokenBase'
 import RemoveTokenButton from './_RemoveTokenButton'
 import TokenTextContainer from './_TokenTextContainer'
+import {ForwardRefComponent as PolymorphicForwardRefComponent} from '../utils/polymorphic'
 
 // Omitting onResize and onResizeCapture because seems like React 18 types includes these menthod in the expansion but React 17 doesn't.
 // TODO: This is a temporary solution until we figure out why these methods are causing type errors.
-export interface TokenProps extends Omit<TokenBaseProps, 'onResize' | 'onResizeCapture'> {
+export interface TokenProps extends TokenBaseProps {
   /**
    * A function that renders a component before the token text
    */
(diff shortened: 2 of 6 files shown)


src/Token/TokenBase.tsx
```diff
@@ -48,16 +49,16 @@ export interface TokenBaseProps
    * Which size the token will be rendered at
    */
   size?: TokenSizeKeys
-}
 
-type TokenElements = HTMLSpanElement | HTMLButtonElement | HTMLAnchorElement
+  href?: string
```
mattcosta7, line 53: we need href defined, because `Token` uses it internally. and typescript will yell if we try to use this without adding it, since it's not, by default, allowed everywhere. 

adding it here allows its use across all of the TokenBase wrappers, which ensures it can work correctly

.changeset/gentle-badgers-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+update types for tokens
```
broccolinisoup, line 5: What do you think updating the changeset to be more descriptive? Maybe something like
>Token: Update component type to be `PolymorphicForwardRefComponent`

src/Token/Token.tsx
```diff
@@ -1,14 +1,15 @@
 import React, {forwardRef, MouseEventHandler} from 'react'
 import Box from '../Box'
-import {merge, SxProp} from '../sx'
+import {BetterSystemStyleObject, merge, SxProp} from '../sx'
```
broccolinisoup, line 3: 👍🏼

broccolinisoup (approved):
>I believe this was caused by the fact that these props aren't equally defined across all polymorphic types, and forwardRef swallows polymorphism, making things wonky.

This makes perfect sense! 

>I think we would be well served extending our type tests further to cover most/all components, as I'm sure others are limited in the same or similar ways.

Thanks for the suggestion ✨ I see there are type tests for quite a few components but the recent ones seem to lack of them. I'll chat with the team about it. 

This looks great to me! 🚀  Thanks for raising this PR 🙏🏼

.changeset/gentle-badgers-confess.md
mattcosta7: updated!

joshblack (approved):

joshblack:
> I believe this was caused by the fact that these props aren't equally defined across all polymorphic types, and forwardRef swallows polymorphism, making things wonky.

Would there be a good place to learn more about how `forwardRef` swallows polymorphism? (Or this concept more broadly). It was the first time I had seen the output of a type get inlined in such a way that information was lost and I would love to learn more about it if you have any suggestions 👀 

For the type tests, is this something that should be a part of the testing requirements for components/props added or changed in a component? It doesn't really seem like there are other alternatives to guarantee that things are showing up, curious if there are other approaches that teams use or if this is kind of our best best moving forward to catch issues like this. Would love to hear your thoughts!

mattcosta7:
> > I believe this was caused by the fact that these props aren't equally defined across all polymorphic types, and forwardRef swallows polymorphism, making things wonky.
> 
> Would there be a good place to learn more about how `forwardRef` swallows polymorphism? (Or this concept more broadly). It was the first time I had seen the output of a type get inlined in such a way that information was lost and I would love to learn more about it if you have any suggestions 👀

I don't know of one offhand, but a quick search brings up things like

https://fettblog.eu/typescript-react-generic-forward-refs/
https://twitter.com/jaredpalmer/status/1217149616082296833?lang=en

the crux of the issue is that a HOC in a static context can't return a generic that infers from an internal generic which seem related.  Which is a mouthfull but this might be more easy to grok

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0e64955901599be1e43a2e50c556d141a2b8fdd8/types/react/index.d.ts#L776-L781

the types for forwardRef don't return a component that's possible of generically defining something based on a certain prop, so it effectively 'swallows' up the as prop (which can't be inferred because it can't be defined)

> For the type tests, is this something that should be a part of the testing requirements for components/props added or changed in a component? It doesn't really seem like there are other alternatives to guarantee that things are showing up, curious if there are other approaches that teams use or if this is kind of our best best moving forward to catch issues like this. Would love to hear your thoughts!

joshblack:
@mattcosta7 makes perfect sense, thanks so much!

@broccolinisoup I'm going to merge this is as the follow-up to the TODO from: https://github.com/primer/react/pull/2760 if that makes sense, let me know if it makes the release more complicated and I'd be happy to revert/add it back in after the release 👍

PR #2763, opened 2023-01-11
PageHeader: Add visual ordering and `PageLayout` story example

broccolinisoup:
1) Add `order` CSS attribute
Just added visual ordering for the children using `order` css attribute. I am not too sure if they do any good though. As this attribute only changes visual ordering and this order is not reflected on the DOM. `PageHeader` children have to be written in the correct order to provide the same ordering both for visually and in the DOM. Slots might be an option but there are quite a few children so I don't know if this would be practical. I am open to suggestions 🙌🏼 
2) Add an example story with `PageLayout`

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/utils/testing.tsx
+++ b/src/utils/testing.tsx
@@ -247,8 +247,25 @@ export function checkStoriesForAxeViolations(name: string, storyDir?: string) {
     if (typeof Story !== 'function') return
 
     const {storyName, name: StoryFunctionName} = Story as StoryType
+
+    beforeEach(() => {
+      // IntersectionObserver isn't available in test environment
+      const mockIntersectionObserver = jest.fn()
+      mockIntersectionObserver.mockReturnValue({
+        observe: () => null,
+        unobserve: () => null,
+        disconnect: () => null,
+      })
+      window.IntersectionObserver = mockIntersectionObserver
+    })
+
     it(`story ${storyName || StoryFunctionName} should have no axe violations`, async () => {
-      const {container} = HTMLRender(<Story />)
+      const {container} = HTMLRender(
+        <ThemeProvider theme={defaultTheme}>
+          <Story />
+        </ThemeProvider>,
+      )
+
       const results = await axe(container)
       expect(results).toHaveNoViolations()
     })
--- a/.changeset/happy-snails-learn.md
+++ b/.changeset/happy-snails-learn.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+PageHeader: Add visual ordering for layout enforcement
--- a/src/PageHeader/PageHeader.tsx
+++ b/src/PageHeader/PageHeader.tsx
@@ -229,6 +229,7 @@ const TitleArea: React.FC<React.PropsWithChildren<TitleAreaProps>> = ({
           {
             display: 'flex',
             gap: '0.5rem',
+            order: REGION_ORDER.TitleArea,
             ...getBreakpointDeclarations(hidden, 'display', value => {
               return value ? 'none' : 'flex'
             }),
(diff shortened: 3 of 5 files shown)


src/utils/testing.tsx
```diff
@@ -247,8 +247,25 @@ export function checkStoriesForAxeViolations(name: string, storyDir?: string) {
...
+        observe: () => null,
+        unobserve: () => null,
+        disconnect: () => null,
+      })
+      window.IntersectionObserver = mockIntersectionObserver
+    })
+
     it(`story ${storyName || StoryFunctionName} should have no axe violations`, async () => {
-      const {container} = HTMLRender(<Story />)
+      const {container} = HTMLRender(
```
broccolinisoup, line 263: [checkStoriesForAxeViolation](https://github.com/primer/react/blob/main/src/PageHeader/PageHeader.test.tsx#L182) was failing on `PageHeader` after adding the new `WithPageLayout` story, because it wasn't able to access the theme object in `PageLayout` on lines like this 
```
[`@media screen and (min-width: ${theme.breakpoints[1]})`]
```

So I wrapped the `<Story />` with the `<ThemeProvider></ThemeProvider>` here and also needed to mock the IntersectionObserver. Tests are succeeding now. Let me know if there is a better way to do this 🙌🏼

mperrotti: Not that I know of. This seems like a reasonable solution to me.

mperrotti (commented):
I don't understand the reason behind the `order` changes. Could you explain?

---

It looks like these changes break the spacing between the action buttons

Before:
Screenshot 2023-01-13 at 1 44 37 PM

After:
Screenshot 2023-01-13 at 1 45 28 PM

broccolinisoup:
> I don't understand the reason behind the order changes. Could you explain?

@mperrotti My intention behind the ordering is to make sure the children are always rendered/displayed in the intended order even though they are "mistakenly" written in a different order. For example the `TitleArea` children should always follow the order of `LeadingAction`, `LeadingVisual`, `Title`, `TrailingAction`, `TrailingVisual` and `Actions`. Same goes for main children as well i.e `ContextArea`, `TitleArea`, `DescriptionArea`, `NavigationArea` I am not sure if this is absolutely necessary but I saw this pattern on [PageLayout](https://github.com/primer/react/blob/main/src/PageLayout/PageLayout.tsx#L16) and wanted to give it a go. However, as I mentioned on the description, I am some concerns. Just copy-pasted what I wrote in the description as well. 

>Just added visual ordering for the children using order css attribute. I am not too sure if they do any good though. As this attribute only changes visual ordering and this order is not reflected on the DOM. PageHeader children have to be written in the correct order to provide the same ordering both for visually and in the DOM. Slots might be an option but there are quite a few children so I don't know if this would be practical. I am open to suggestions 🙌🏼

mperrotti:
Thanks @broccolinisoup - I understand now.

I think this is a nice way to enforce layout/placement without relying on slots. I can't think of a cleaner way to do it.

broccolinisoup:
> Thanks @broccolinisoup - I understand now.
> 
> I think this is a nice way to enforce layout/placement without relying on slots. I can't think of a cleaner way to do it.

Thanks @mperrotti!

>It looks like these changes break the spacing between the action buttons

This seems to be an issue on the [Hidden-component-refactor](https://github.com/primer/react/tree/Hidden-component-refactor) that I branched off this PR. I updated the styling of the story on the [Hidden-component-refactor](https://github.com/primer/react/tree/Hidden-component-refactor) branch and rebased again , so it should be good  ✨

mperrotti (approved):

PR #2765, opened 2023-01-11
Remove shadow from `MarkdownEditor` `textarea`

iansan5653:
There is a weird little subtle box-shadow on the `MarkdownEditor` input:

screenshot

This small fix removes that:

screenshot


joshblack (approved):
🎉

PR #2767, opened 2023-01-11
chore(project): add format task for files

joshblack:
Add a `format` and `format:diff` task to be used to format files that may not be currently included in the `eslint` setup. Specifically, this allows us to use prettier for other file types like YML or in plain markdown files.

In addition, this adds a `format` job to the CI workflow that runs `format:diff` to see if any files are unformatted that could be formatted.

This PR also runs `npm run format` to format files caught by `format:diff`


joshblack:
bump @siddharthkp if you have a sec!

colebemis (approved):
Ship it ✨

PR #2769, opened 2023-01-12
PageHeader: Type fixes and add type tests

broccolinisoup:
Describe your changes here.

There was a type issue I realised that I was using `PageHeaderProps` as prop type for all children which is brutally wrong 😬 So I have fixed that and I added type tests.

### Screenshots

No visual changes

### Merge checklist

- [x] Added/updated tests

### DIFF
--- a/src/PageHeader/PageHeader.types.test.tsx
+++ b/src/PageHeader/PageHeader.types.test.tsx
@@ -1,0 +1,50 @@
+import React from 'react'
+import {PageHeader} from '../PageHeader'
+
+// PageHeader
+export function acceptsAsProp() {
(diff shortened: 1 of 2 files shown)


src/PageHeader/PageHeader.types.test.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import {PageHeader} from '../PageHeader'
+
+// PageHeader
```
broccolinisoup, line 4: I would love to hear if you have any suggestions to extend the type tests ✨

joshblack (approved):

PR #2770, opened 2023-01-12
UnderlineNav2: Handle the case when container is too small to render any items

### LINKED ISSUE #2752: UnderlineNav throws error when screen resizes to a point where no items can render
### Description

https://user-images.githubusercontent.com/3901764/211390398-e7cd542e-0e68-4b8e-8359-e8b0fd93c2bf.mov

_Video showing the browser being resized to a smaller size. As it gets smaller, items from the underline nav disappear from the list of items. When the space is small enough that no items can be rendered, an error is thrown_

### Steps to reproduce

1. Visit the storybook for `UnderlineNav`: https://primer.style/react/storybook/iframe.html?args=&id=drafts-components-underlinenav-features--overflow-template&viewMode=story
2. Make the window smaller until the error shows up

### Version

v35.16.0

### Browser

_No response_

broccolinisoup:
Closes #2752
### Issue
UnderlineNav v2 was throwing an error when the container is too small to render any items. Although there was no item in the list, it was still trying to swap a menu item with a list item to keep the selected one always visible and it resulted an error.

### Solution
We fixed the issue by making sure we only swap items to keep the selected one visible **when there is at least 1 element on the list**.  When the screen is too small to display one list element with the more menu button, we pull all items in the menu. 

### Caveat
When we display only the menu button, the menu items are not visually accessible. And this is something we are aware of. This is a very edge case as a browser can get that much narrow only when the dev tool is open. We will make sure we address this when working on [mobile design for UnderlineNav.](https://github.com/github/primer/issues/1100#issuecomment-1362100886)

### Screenshots

https://user-images.githubusercontent.com/1446503/215007018-b04b2ce1-0145-49f0-aead-a41bb3a7232f.mp4

_Video displays the browser is being resized to a smaller size when the dev tool is open on the right hand side. As the browser gets smaller, the list items are being pulled into an overflow menu and they disappears from the screen. When the browser size is too small to display one list item and the menu button, all items goes into the menu and the overflow menu button's text goes from "More" to "Menu". Then the "Menu" button is clicked to show the items but as there is not enough space, the menu items are visually not accessible. Then the browser is being resized to a larger size, the menu items popping out of the menu and are displayed as list items to confirm the original functionality is working as expected._

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack (approved):
🥳

danielguillan:
I think that from both design and accessibility perspectives, we might want to avoid horizontal scrolling in this scenario too. 

A solution could be be to move all items into a dropdown menu if the visible item + `more` dropdown don't fit in the available space. The dropdown could be labeled using the provided `aria-label`.

broccolinisoup:
> I think that from both design and accessibility perspectives, we might want to avoid horizontal scrolling in this scenario too.
> 
> A solution could be be to move all items into a dropdown menu if the visible item + `more` dropdown don't fit in the available space. The dropdown could be labeled using the provided `aria-label`.

Thanks @danielguillan! I was very unsure about it after discarding the scroll behaviour but I couldn't really think of any other solution while making sure to keep the selected item visible and out of the menu. Your suggestion sounds good! I also wonder how we could indicate the selected item when it is in the dropdown?

broccolinisoup:
https://user-images.githubusercontent.com/1446503/212797844-2ba214a7-9ae7-48f7-950e-d77db21b04aa.mp4

Video displays the browser is being resized to a smaller size. As the browser gets smaller, the list items are being pulled in to an overflow menu and they disappears from the screen. When the browser size is too small to display one list item and the menu button, the last is being pulled into the menu as well and the More Menu button text changes to "Items" from "More"

@danielguillan is this something close to what you were thinking? I also updated the button text to "Items" when we only display the men because "More" didn't make sense to me but it is just trying really, let me know what you think 🙌🏻  I also updated the `aria-label` accordingly

danielguillan:
@broccolinisoup yes, that's very in line with what I was thinking. 👍 

As for the label, instead of "Items", could we perhaps use the provided `aria-label`? 

Another thing I'm thinking is that in this scenario, now that the whole menu is contained in the dropdown we might want to use a `NavList` or similar approach instead of the disclosure pattern. We might need some input from accessibility folks here.

Also, as currently implemented, the dropdown on the disclosure pattern doesn't position itself taking into account the viewport boundaries, so the menu is sometimes rendered outside of the canvas.

screenshot -HKbadxyI-000166@2x

broccolinisoup:
>As for the label, instead of "Items", could we perhaps use the provided aria-label?

@danielguillan We sure could do that! Although it is a best practise and we warn consumers on the dev env when they don't provide `aria-label` but it is still an optional prop in the API. So what would be an alternative when there is no `aria-label`?

>Another thing I'm thinking is that in this scenario, now that the whole menu is contained in the dropdown we might want to use a NavList or similar approach instead of the disclosure pattern. We might need some input from accessibility folks here.

I feel this goes under the scope of designing/improving UnderlineNav for narrow viewports. The current solution resolves the bug and also, this scenario is an edge case because a browser could only get as small as in this case when only the dev tool is open. So considering all these, I think we can handle future improvements maybe outside of this bug/PR as we are cautious of delivering the component  this quarter? ([ref](https://github.com/github/primer/issues/1100#issuecomment-1362100886))

danielguillan:
> So what would be an alternative when there is no aria-label?

I don't have a strong candidate for that. Maybe something like "Navigation" or "Menu"?

> I feel this goes under the scope of designing/improving UnderlineNav for narrow viewports.

👍

broccolinisoup:
> > So what would be an alternative when there is no aria-label?
> 
> I don't have a strong candidate for that. Maybe something like "Navigation" or "Menu"?
> 
> > I feel this goes under the scope of designing/improving UnderlineNav for narrow viewports.
> 
> 👍

Hi @danielguillan when using `ariaLabel` for the menu button text there is a chance that it might overflow. See the screenshot please. I wonder if we should use a set word to display to reduce unpredicted style flaws. Possibly an overlapping word with the visually hidden content. Currently screen reader will announce the menu  as `More ${ariaLabel} items` when there are list items so I think to stay consistent with the experience we should make the menu button announced as `${ariaLabel} items` when there is only a menu button. If we want to avoid unpredicted style flaws while having an overlapping a word with the screen reader announce, we could use `Items` . This was my thinking pattern but happy to discuss and see if we can come up with something better. 

A menu button with a text of _GitHub Discussions_ but the tail of the content is obscured by the dev tool because text is overflowing

Thanks so much for your help 🙏🏻

danielguillan:
@broccolinisoup Good point! The I think we could stick with "Menu". What do you think?

broccolinisoup:
> @broccolinisoup Good point! Then we could stick with "Menu". What do you think?

Sounds good @danielguillan - I updated it with "Menu". For the visual users, they will see the button with the "Menu" text, for screen reader users, it will be announced `${ariaLabel}&nbsp;Menu`

broccolinisoup:
Hi @primer/accessibility  👋🏻  Would it be possible to get a review for our work here? [This story](https://primer-4661593960-13348165.drafts.github.io/storybook/iframe.html?args=&id=drafts-components-underlinenav-features--overflow-template&viewMode=story) should be a good one to test. I provided info on the description but please let me know if you would need further details. Thank you 🙌🏻

inkblotty:
Based on our conversation in Office Hour, this is a usable component. Some of the semantics are a bit overkill, but still accessible.

The visibility problem at 150 px width is a non-issue and does not need to be resolved from an accessibility standpoint.

I will link the recording once it's uploaded.

@broccolinisoup - If we missed testing something you needed eyes on, please let us know (`@github/accessibility-reviewers`). We're still getting our process for async Office Hours down 🙇

broccolinisoup:
> Based on our conversation in Office Hour, this is a usable component. Some of the semantics are a bit overkill, but still accessible.
> 
> The visibility problem at 150 px width is a non-issue and does not need to be resolved from an accessibility standpoint.
> 
> [Link to the recording](https://github.rewatch.com/video/xmanhqi2tjlg7vnk-accessibility-office-hours-feb-1-2023)
> 
> @broccolinisoup - If we missed testing something you needed eyes on, please let us know (`@github/accessibility-reviewers`). We're still getting our process for async Office Hours down 🙇

Thank you @inkblotty so much! I just watched the recording and this was my first time submitting an issue to accessibility office hours so there were lots I learnt from this experience and how I can provide a better context and question next time - I do appreciate your patience with me 🙏🏻  

I am so sorry that I forgot to provide the storybook link in the description and I could have explain better what exactly I needed to have an accessibility review on. You covered everything I was concerned though - thank you so much. I wasn't too worried about the 150px screen size as I mentioned it on the caveat. Although, we don't officially support this size, for users who somehow happen to end up in this size, I was still curious to understand if having only a menu button would that be good solution in terms of accessibility. Maybe it is a little bit overthinking 🙂 You answered all my questions and thank you so much again 🙇🏻‍♀️

inkblotty:
So glad to hear all your questions were answered, even though we were a bit
confused. Let us know how we can help in the future 🙂

PR #2772, opened 2023-01-13
ci(vrt): update vrt flow for updating snapshots

joshblack:
Update the `vrt` workflow into two part:

- The `vrt` workflow which runs on the `pull_request` event trigger and uploads an artifact that contains updated snapshots
- An `update-snapshots` workflow which will download the snapshots and commit them back to the branch
  - This work will also remove the appropriate label when finished running

I have a running example of these two ideas over in this test repo: https://github.com/joshblack/vrt-workflow-experiment


colebemis (approved):

.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,137 @@
+name: update-snapshots
+on:
+  workflow_run:
+    workflows: [vrt]
+    types:
+      - completed
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
joshblack, line 9:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch }}
```

.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,137 @@
...
+on:
+  workflow_run:
+    workflows: [vrt]
+    types:
+      - completed
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
```
joshblack, line 11:
```suggestion
```

PR #2784, opened 2023-01-18
Expose`useIsSSR` from react aria

theinterned:
I'd like to expose the [`useIsSSR` hook from `react-aria`](https://react-spectrum.adobe.com/react-aria/ssr.html#ssr-specific-rendering) as an export from primer.

## Why?

I was in the middle of re-implementing similar functionality in a consumer of `primer/react`'s `SSRProvider` [when I noticed that this hook already exists and could be available to me](https://github.com/github/github/pull/254652#issuecomment-1387386417).

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/six-books-brake.md
+++ b/.changeset/six-books-brake.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": minor
+---
+
+Expose the [`useIsSSR` hook from `@react-aria/ssr`](https://react-spectrum.adobe.com/react-aria/ssr.html#ssr-specific-rendering)
--- a/src/index.ts
+++ b/src/index.ts
@@ -181,6 +181,6 @@ export type {TextareaProps} from './Textarea'
 
 export {UnderlineNav as UnderlineNav2} from './UnderlineNav2'
 
-export {SSRProvider, useSSRSafeId} from './utils/ssr'
+export {SSRProvider, useSSRSafeId, useIsSSR} from './utils/ssr'
 export {default as sx, merge} from './sx'
 export type {SxProp} from './sx'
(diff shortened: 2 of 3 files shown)


theinterned:
I do have a question about this: given that React now gives us the functionality to render stable IDs on server and client, will we no longer need the `@react-aria/ssr` package? 

I suppose that if there are deprecation plans for the package, then this PR might not make sense.

joshblack:
Hey @theinterned! 👋 

Great question, I believe that we won't need `@react-aria/ssr` as a dependency anymore in React 18 for the reasons that you shared.

What we're currently doing in the project as we bridge between React 17 and 18 is to use an intermediate `useId` hook: https://github.com/primer/react/blob/main/src/hooks/useId.ts. This hook picks between `React.useId()` if in React 18 and `useSSRSafeId` if in React 17. The goal will be in the next major to ultimately remove `useSSRSafeId` in preference of `React.useId()` 👍

theinterned:
> The goal will be in the next major to ultimately remove `useSSRSafeId` in preference of `React.useId()` 👍

@joshblack okay so given this, it sounds like we _shouldn't _ start depending on more stuff (`useIsSSR`) from `@react-aria/ssr`? And instead we should close this PR?

joshblack:
@theinterned unfortunately yeah, I think so. Hope this makes sense, let me know if I'm misunderstanding anything here.

theinterned:
@joshblack nope that's perfectly fine! I'll continue with our internal implementation instead. Thanks

PR #2785, opened 2023-01-19
Limit use of `as` in Text component

TylerJDev:
[**WIP**]: This change aims to stem usage of `<Text>` in cases where the `as` prop might lead to inaccessible use, or there's already an existing Primer component for that element type. The goal is to keep the flexibility of Primer, while giving a smaller group of expected elements to utilize with the `as` prop. 

* Reduces usage of `as` in `<Text>` by only allowing a group of components that align with `<Text>` (e.g. [inline text based elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element#inline_text_semantics)).

Closes [#1695](https://github.com/github/primer/issues/1695)

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Text.tsx
```diff
@@ -1,13 +1,28 @@
...
 `
 
-export type TextProps = ComponentProps<typeof Text>
+export type TextProps = {
+  as?: typeof RestrictedAs.inlineText[number]
+} & React.ComponentProps<'span' & typeof StyledText>
+
+const Text = ({children, as = 'span', ...rest}: TextProps) => {
+  return (
+    <StyledText as={RestrictedAs.inlineText.includes(as) ? as : undefined} {...rest}>
```
colebemis, line 22: In development mode, could we also output an error in the console if the developer passed an invalid `as` value? If we're ignoring their value, we should warn them about it. @green6erry has code for this.

src/Text.tsx
```diff
@@ -1,13 +1,28 @@
+import React from 'react'
 import styled from 'styled-components'
 import {COMMON, SystemCommonProps, SystemTypographyProps, TYPOGRAPHY} from './constants'
 import sx, {SxProp} from './sx'
-import {ComponentProps} from './utils/types'
 
-const Text = styled.span<SystemTypographyProps & SystemCommonProps & SxProp>`
+const RestrictedAs = {
+  inlineText: ['span', 'p', 'b', 'i', 'em', 'small', 'strong', 'u', 'code', 'q', 's', 'label', 'div'],
+} as const
```
TylerJDev, line 8: This is mainly just as an example, since the structure might change if we did keep this solution. 

I wasn't too sure where/if we keep constants somewhere specific? I saw our `constants.ts` file, but wasn't sure if this was the proper place to put variables like this?

src/Text.tsx
```diff
@@ -1,13 +1,28 @@
+import React from 'react'
 import styled from 'styled-components'
 import {COMMON, SystemCommonProps, SystemTypographyProps, TYPOGRAPHY} from './constants'
 import sx, {SxProp} from './sx'
-import {ComponentProps} from './utils/types'
 
-const Text = styled.span<SystemTypographyProps & SystemCommonProps & SxProp>`
+const RestrictedAs = {
```
colebemis, line 6: Is there a reason you're using PascalCase for this object?

src/Text.tsx
```diff
@@ -1,13 +1,28 @@
...
   ${COMMON};
   ${sx};
 `
 
-export type TextProps = ComponentProps<typeof Text>
+export type TextProps = {
+  as?: typeof RestrictedAs.inlineText[number]
+} & React.ComponentProps<'span' & typeof StyledText>
+
+const Text = ({children, as = 'span', ...rest}: TextProps) => {
```
colebemis, line 20: To avoid a breaking change, we'll need to wrap this in a `React.forwardRef()` call.

colebemis: styled-components automatically handles ref forwarding. If we're wrapping the styled component, we'll need to handle forwarding refs manually.

src/Text.tsx
TylerJDev: Good call, I forgot that there might be confusion on why applying an element isn't working as expected.

src/Text.tsx
TylerJDev: Nope, I'll swap this out for camel.

src/Text.tsx
TylerJDev: I see, I'll adjust this, and make note of it for future cases!

src/Text.tsx
TylerJDev: Should now be added. I added a message around what elements are allowed, though I want to make sure it isn't too verbose.

src/Text.tsx
TylerJDev: Adjusted! 

I saw that we utilize `PolymorphicForwardRefComponent` so I opted to utilize that for the types - but would love a double check to make sure everything looks solid TS wise.

src/Text.tsx
```diff
@@ -1,13 +1,53 @@
...
+
+const StyledText = styled.span<SystemTypographyProps & SystemCommonProps & SxProp>`
   ${TYPOGRAPHY};
   ${COMMON};
   ${sx};
 `
 
-export type TextProps = ComponentProps<typeof Text>
+export type TextProps = {
+  as?: typeof restrictedAs[number]
```
colebemis, line 34: Does `keyof restrictedAs` work here?

src/Text.tsx
```diff
@@ -1,13 +1,53 @@
...
+} & React.ComponentProps<'span' & typeof StyledText>
+
+const Text = React.forwardRef(({children, as = 'span', ...rest}, forwardedRef) => {
+  if (__DEV__ && !restrictedAs.includes(as)) {
```
colebemis, line 38: The `if (__DEV__)` statement may need to be on its own for the compiler to remove it for production builds. cc @joshblack 

```suggestion
  if (__DEV__) {
    if (!restrictedAs.includes(as)) {
```

src/Text.tsx
```diff
@@ -1,13 +1,53 @@
+import React from 'react'
 import styled from 'styled-components'
 import {COMMON, SystemCommonProps, SystemTypographyProps, TYPOGRAPHY} from './constants'
+import {ForwardRefComponent as PolymorphicForwardRefComponent} from './utils/polymorphic'
 import sx, {SxProp} from './sx'
-import {ComponentProps} from './utils/types'
 
-const Text = styled.span<SystemTypographyProps & SystemCommonProps & SxProp>`
+const restrictedAs = [
```
colebemis, line 7: `restrictedAs` might be a confusing name for this variable because it could imply that the following values are restricted. Maybe `allowedElements` is more clear?

src/Text.tsx
```diff
@@ -1,13 +1,53 @@
...
 `
 
-export type TextProps = ComponentProps<typeof Text>
+export type TextProps = {
+  as?: typeof restrictedAs[number]
+} & React.ComponentProps<'span' & typeof StyledText>
+
+const Text = React.forwardRef(({children, as = 'span', ...rest}, forwardedRef) => {
+  if (__DEV__ && !restrictedAs.includes(as)) {
+    console.warn(
```
colebemis, line 39: You can use an eslint-disable comment to hide this lint error: https://github.com/primer/react/blob/main/src/Link.tsx#L62

src/Text.tsx
joshblack: +1 to the convention of "dev-only" blocks wrapping the code that is stripped way. In either case, I think Terser will optimize away `if (false && condition)` thankfully but I would love it if we could still wrap the block in `if (__DEV__)`

TylerJDev: Perfect, thanks y'all! I'll adjust this and make note for future cases.

src/Text.tsx
TylerJDev: I'm getting an error here when using `keyof`, but I think it's because I changed `restrictedAs` from an object to an array.

PR #2786, opened 2023-01-19
ci(vrt): update vrt to use multiple runners

joshblack:
Add the ability to update snapshots faster using multiple runners.

Currently, our update flow only uses one process to generate snapshots. As the number of snapshots and tests grows, this process will take longer and longer. In order to reduce the total amount of time, we can break up these tests into separate runners and combine all the snapshots together at the end into one commit.


joshblack:
bump @colebemis when you have a sec!

colebemis (approved):
🚀

PR #2789, opened 2023-01-20
1573 accessibility request primer react link enforce what as can be set to with a runtime validation rule 1572

green6erry:
Describe your changes here.

Closes [#1573](https://github.com/github/primer/issues/1573) (type the issue number after # if applicable; otherwise remove this line)

The user sees a warning if they implement `Link` without using an `a` or `button` element in the `as` prop.

### Screenshots

image

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Link.tsx
```diff
@@ -38,5 +41,41 @@ const Link = styled.a<StyledLinkProps>`
...
+     */
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useEffect(() => {
+      if (
+        innerRef.current &&
+        !(innerRef.current instanceof HTMLButtonElement) &&
+        !(innerRef.current instanceof HTMLAnchorElement)
+      ) {
+        // eslint-disable-next-line no-console
+        console.warn('This component should be an instanceof a semantic button or anchor')
```
colebemis, line 63: Can we be more specific about which component is causing this warning? If I just saw `This component...` in the console, it would be hard to track down. If we log the ref too, it might be easier to find the corresponding component.

```suggestion
        console.error('Error: Found `Link` component that renders an inaccessible element', innerRef.current, 'Please ensure `Link` always renders as <a> or <button>')
```

colebemis (approved):

PR #2804, opened 2023-01-23
Reimplemented tooltips using the Popover API.

### LINKED ISSUE #2124: Tooltip should use Popover API to avoid z-index issues

radglob:
Reimplemented the Tooltip component using the Popover API. This should not change anything about the tooltip component from a user's perspective.

Bring this in line with the PVC implementation, this doesn't support the `align` prop. Users should use the diagonal directions for this instead.

Since the [Popover API](https://developer.chrome.com/docs/web-platform/popover-api/) is not yet available in stable browsers, this includes a [polyfill](https://github.com/oddbird/popover-polyfill) as well as an update to `styled-components` from 4.x to 5.x to take advantage of the [shouldForwardProp]() feature. That update will block merging to main, since we would need to update all dependent applications to use 5.x. Primer currently supports either 4.x or 5.x.

Closes #2124.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


radglob:
This is going to be held up for the moment. There's a conflict in transitive dependencies (specifically some packages that storybook uses) that can't be updated without bumping a major version on Storybook.

PR #2807, opened 2023-01-24
feat(Button): add default type

### LINKED ISSUE #2806: Button component has no default type
### Description

The `Button` component does not have a default `type` attribute. In e.g. react-bootstrap, the `Button` component by default has `type` set to `button`. Without such a default, the `type` will default to `submit` within forms, yielding the following warning when using e.g. Formik:

You submitted a Formik form using a button with an unspecified `type` attribute.  Most browsers default button elements to `type="submit"`. If this is not a submit button, please add `type="button"`.

As it is normal to add multiple buttons within a form, it seems safer and more intuitive to have a default type set to `button`, and then explicitly specify which button to use for submission.

### Steps to reproduce

1. Create page with the following form:
```
<form>
<Button type="submit">Submit</Button>
<Button>Recalculate<Button>
</form>
```

2. The second button will default to `type=submit` instead of `type=button`. Expected behavior would be that the second `Button` type would default to `button`.

### Version

v35.2.0

### Browser

Firefox

joshblack:
Closes #2806 

Update the default type for `Button` to be `"button"`. By default, buttons are inferred to be of `type="submit"` which can cause some unexpected behavior, especially within forms.

#### Changelog

**New**

**Changed**

- Add `type="button"` to `Button.tsx`

**Removed**


langermank (approved):
Do you think this is a breaking change potentially?

joshblack:
@langermank I think if someone relies on inferred `type="submit"` then unfortunately yeah, it could be considered one 😞

colebemis (approved):

iansan5653 (approved):
LGTM! Regarding moving this to `ButtonBase` instead: I don't think it's necessary because `ButtonBase` is only used elsewhere in `LinkButton` (which renders as an `a` tag) and some deprecated components.

This is theoretically a breaking change but I've personally never seen anyone rely on the implicit `submit` type of a button. I'd be surprised if this is done anywhere.

joshblack:
I think we can try this out in the upcoming release and see how it lands, happy to revert it back if it becomes problematic!

mattcosta7:
I go 50/50 on this, but agreed with feeling it out

I always think html is weird here by defaulting type to submit in forms, but also that consistency with html is nice (even though it's non-intuitive)

PR #2759, opened 2023-01-10
move defaultProps to function defaults

manuelpuyol:
I'm updating all non-deprecated components that use `defaultProps` to stop using it and using function defaults instead.
As described in the issue below, `defaultProps` is going to be deprecated so it'd be nice to be clear of it before the new React version is released.

Closes https://github.com/primer/react/issues/2758

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #2808, opened 2023-01-24
Fix forwardRef regression

colebemis:
Fixes a regression in https://github.com/primer/react/pull/2759 that removed support for the `ref` prop on Avatar, Flash, and Label.


colebemis:
This doesn't need a changeset because the regression we're fixing in this PR hasn't been released yet.

joshblack (approved):

PR #2819, opened 2023-01-26
chore(project): update stories to csf3

joshblack:
[Announcement](https://storybook.js.org/blog/storybook-csf3-is-here/)

This PR runs `npx storybook@next migrate csf-2-to-3` to update stories from CSF2 to CSF3.

### DIFF
--- a/src/stories/useAnchoredPosition.stories.tsx
+++ b/src/stories/useAnchoredPosition.stories.tsx
@@ -90,52 +90,53 @@ const Anchor = styled(Box)`
   padding: ${get('space.3')};
 `
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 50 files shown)


src/stories/useAnchoredPosition.stories.tsx
```diff
@@ -254,23 +255,6 @@ const Main = styled('main')`
...
-   elements are not rendered at full height! Notice how <Main> uses
-   top, left, right, and bottom all set to 0 to achieve a full-size box.
-
-2. The positioning routine needs to know the size of the overlay before
-   calculating its position! Therefore, we use visibility: hidden to
-   prevent showing a single frame of the overlay being positioned at
-   (0, 0).
-
-*/
-
```
joshblack, line 273: Seems like this should be included again 🤔

PR #2821, opened 2023-01-26
Update Label stories

langermank:
As part of https://github.com/github/primer/issues/1488

Adding a Playground, Default and feature stories for docs.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Label/index.ts
```diff
@@ -0,0 +1 @@
+export * from './Label'
```
colebemis, line 1: This should fix the build errors
```suggestion
import Label from './Label'
export type {LabelProps} from './Label'
export default Label
```

src/Label/Label.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+import Label from './Label'
+
+export default {
+  title: 'Components/Label',
+  component: Label,
+  args: {
+    variant: 'default',
+    size: 'small',
+  },
+  argTypes: {
```
colebemis, line 13: Can we apply `args` and `argTypes` directly to the playground story?

```
Playground.args = {...}
Playground.argTypes = {...}
```

The way this is written, I think the Default story inherits these args.

langermank: Yes, thanks for the reminder!

joshblack (approved):

PR #2824, opened 2023-01-27
Check that .docs.json files have matching .stories.tsx files

colebemis:
Adds a test to ensure that every `.docs.json` file has a corresponding `.stories.tsx` file.


src/__tests__/storybook.test.tsx
```diff
@@ -73,3 +75,18 @@ describe.each(components)('%s', (_component, stories) => {
...
+// eslint-disable-next-line jest/no-identical-title
+describe.each(jsonFiles)('%s', filepath => {
+  const name = path.basename(filepath, '.docs.json')
+  if (allowlist.includes(name)) {
+    test('has a corresponding .stories.tsx file', () => {
+      const storyFilepath = path.join(ROOT_DIRECTORY, filepath.replace('.docs.json', '.stories.tsx'))
+      expect(fs.existsSync(storyFilepath)).toBe(true)
+    })
+  }
+})
```
joshblack, line 92:
```suggestion
const jsonFiles = glob.sync('src/**/*.docs.json', {
  cwd: ROOT_DIRECTORY,
}).filter((filepath) => {
  const name = path.basename(filepath, '.docs.json')
  return allowlist.includes(name);
});

// eslint-disable-next-line jest/no-identical-title
describe.each(jsonFiles)('%s', filepath => {
  test('has a corresponding .stories.tsx file', () => {
    const storyFilepath = path.join(ROOT_DIRECTORY, filepath.replace('.docs.json', '.stories.tsx'))
    expect(fs.existsSync(storyFilepath)).toBe(true)
  })
})
```

joshblack (approved):
Looks great! Just left a minor suggestion for filtering the files at the `glob` call which could avoid the conditional `test()` block if that is helpful.

langermank (approved):

PR #2848, opened 2023-01-31
Fix `useDynamicTextareaHeight` not taking into account padding

iansan5653:
The `useDynamicTextareaHeight` uses a utility function that calculates the number of pixels from the top edge of the component to the bottom of a given character. We then use that number of pixels to set the height of a `textarea` so that the height of the `textarea` always fits the text inside of it.

However, the utility function _includes_ the top padding in that height. But when we set `height` on the underlying element, we don't want to include the top padding because when `box-sizing` is `content-box`, the `height` of an element is it's _interior_ height - the height not including any padding or borders.

When I wrote the hook I intended to include a workaround for this by setting the padding to zero, calculating the height, and then resetting the padding back to what it was. But somehow the lines got switched up and I was performing the steps in the wrong order, causing the workaround to have no effect. The result of this was that `MarkdownEditor` always had slightly too much space below the bottom line of text.

This PR fixes the order of the steps, and also changes the hook to accept a `ref` instead of an element instance. This is technically a breaking change but we don't need a major bump because this is a draft hook and is not used anywhere in `github/github`.

#### Before

Notice how the input expands before it needs to:

https://user-images.githubusercontent.com/2294248/215885274-5dad916f-b5c3-46a1-9191-c5bb9af671cb.mov

#### After

Now the input fits more tightly to the text:

https://user-images.githubusercontent.com/2294248/215885448-1967d5ba-fa90-448d-bcd2-1b8cc16e8771.mov


src/drafts/hooks/useDynamicTextareaHeight.ts
```diff
@@ -1,12 +1,12 @@
...
 
 import {SxProp} from '../../sx'
 import {getCharacterCoordinates} from '../utils/character-coordinates'
 
 type UseDynamicTextareaHeightSettings = {
-  minHeightLines: number
-  maxHeightLines: number
-  element: HTMLTextAreaElement | null
+  minHeightLines?: number
+  maxHeightLines?: number
```
broccolinisoup, line 8: We seem to have default values for these. Would there be any cases where they become undefined here?

src/drafts/hooks/useDynamicTextareaHeight.ts
```diff
@@ -16,43 +16,47 @@ type UseDynamicTextareaHeightSettings = {
...
-  element,
+  elementRef,
   value,
 }: UseDynamicTextareaHeightSettings): SxProp['sx'] => {
   const [height, setHeight] = useState<string | undefined>(undefined)
   const [minHeight, setMinHeight] = useState<string | undefined>(undefined)
   const [maxHeight, setMaxHeight] = useState<string | undefined>(undefined)
 
   useLayoutEffect(() => {
+    const element = elementRef.current
```
broccolinisoup, line 36: Would it worth to check if `elementRef` is not null here or not really?

src/drafts/hooks/useDynamicTextareaHeight.ts
iansan5653: This hook is exported from drafts, so it's usable outside of the `MarkdownEditor`. In some cases consumers might not want to provide these - I just figured I'd make it more flexible.

src/drafts/hooks/useDynamicTextareaHeight.ts
iansan5653: It shouldn't ever be `null` since it's a required prop - I think ESLint will actually complain about an unnecessary check in that case

src/drafts/hooks/useDynamicTextareaHeight.ts
broccolinisoup: Ah that makes sense

broccolinisoup (approved):
Looks good to me 🚀

PR #2850, opened 2023-02-01
CounterLabel: Do not parse children to string for visually hidden content

### LINKED ISSUE #2837: CounterLabel doesn't work for non-string children
### Description

On the search results page ([e.g.](https://github.com/search?q=foo&type=repositories)) we use the `<CounterLabel>` element, which is rendering "([object Object])":
```html
<span class="_VisuallyHidden__VisuallyHidden-sc-11jhm7a-0 rTZSs">&nbsp;([object Object])</span>
```

Which is not great.

`CounterLabel` should render this correctly or be updated to only accept string children.

### Steps to reproduce

We use `CounterLabel` like this
```jsx
    <CounterLabel data-testid="resolved-count-label" sx={{height: 20, lineHeight: '20px', py: 0, px: 2}}>
      <Count isLowerBound={resultCountIsLowerBound} value={totalResults} />
    </CounterLabel>
```
But CounterLabel [has code that renders this](https://github.com/primer/react/blob/main/src/CounterLabel.tsx#LL41C7-L41C63):
```jsx
<VisuallyHidden>&nbsp;{`(${children})`}</VisuallyHidden>
```
Which doesn't work for children that aren't simple strings.

### Version

35.18.0

### Browser

Chrome, Safari, Firefox, Edge, iOS Safari

broccolinisoup:
Closes #2837 

We added a visually hidden content for `CounterLabel` to improve the accessibility. We added a non-breaking space `&nbsp;` and wrapped the counter children with parentheses to allow screen readers to pause before reading the counter but we were also parsing the children to string and it caused issues for non-string or number children. This PR addresses this problem. 

We actually do not need to parse it to string to be able to have the expected visually hidden content. So just removing it solves the issue. 

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


josepmartins (approved):
LGTM! ❇️

Issue #2814, opened 2023-01-25
`DialogV2` should have `aria-modal="true"`


iansan5653:
### Description

The `DialogV2` component enables a focus trap and background overlay by default. This makes it a `modal` dialog by definition:

> A section of content is "modal" means navigation is limited to the area itself and the background (the ancestors and siblings of the modal) is hidden. Setting `aria-modal="true"` on [`dialog`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) and [`alertdialog` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alertdialog_role) containers indicates the presence of a "modal" element to users of assistive technology

By default, items with `role="dialog"` have `aria-modal="false"`, indicating to screen readers that navigation outside the dialog is possible. But it is not, so this is misleading and can get screen readers into a state where interaction appears possible but is not.

Explicitly setting the `aria-modal="true"` prop on the `Dialog` component does _not_ work as a workaround because that prop is not forwarded to the underlying element.

### Steps to reproduce

Render a `Dialog` component from `@primer/react/drafts`

### Version

v35.17.0

### Browser

Chrome

langermank:
Wondering if we might want to close this issue in favor of https://github.com/github/primer/issues/1551 which will include using the PVC `Dialog` as a reference for markup and accessibility features.

lesliecdubs:
Thanks for filing @iansan5653, and thanks for the link @langermank!

The PVC `Dialog` component has been approved by the accessibility team and we have plans to replicate the improved markup in React as part of https://github.com/github/primer/issues/1551. Feel free to follow progress along on that issue, or reach out if you're interested in contributing the change.

iansan5653:
Would a PR be accepted for a quick fix to resolve this for now? This is a pretty significant accessibility problem for which there is no good workaround - it seems to me like it would probably be worth shipping the one-line fix for now even if this component would be replaced by that related work eventually.

lesliecdubs:
Yes, we would be happy to review a PR contribution assuming it's a quick fix! Let me know if you want me to reopen this for you to track against.

PR #2851, opened 2023-02-01
Add `aria-modal` attribute to `Dialog` (V2) element

iansan5653:
Adds the `aria-modal` attribute to the container for the `DialogV2` element. This is required because the dialog always instantiates a focus trap. The only way to interact with the underlying page is to close the dialog:

> Relevant only on `dialog` and `alertdialog` containers, setting `aria-modal="true"` tells assistive technologies to let the user know the ability to interact with, or access other content on the page requires the modal dialog to be closed or otherwise lose focus. ([MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal))

Adding `aria-modal` prevents screen readers from continuing past the dialog and starting to read the underlying content, which they will do without the attribute.

Closes https://github.com/primer/react/issues/2814


src/Dialog/Dialog.tsx
```diff
@@ -302,6 +302,7 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
             role={role}
             aria-labelledby={dialogLabelId}
             aria-describedby={dialogDescriptionId}
+            aria-modal
```
iansan5653, line 305: I considered adding a prop to allow users to override this, but since we don't allow them to override the focus zone or exit behavior, there is no valid scenario in which this dialog is not modal.

colebemis (approved):

joshblack (approved):

PR #2856, opened 2023-02-03
ADR 010: Merge drafts status into experimental

siddharthkp:
Closes https://github.com/github/primer/issues/1463

[Rendered markdown →](https://github.com/primer/react/blob/adr-10/contributor-docs/adrs/adr-010-drafts-are-experimental.md)


lesliecdubs:
Thanks for writing this up, @siddharthkp! @primer/react-reviewers please take a look and leave any feedback on this ADR. Will also add a few reviewers who have a stake in our lifecycle criteria.

This is both a technical and product decision, so I think it makes the most sense for @justinbyo to serve as the decision maker.

**ADR decision maker:** @justinbyo
**Decision deadline:** **EOD February 10, 2023** - let me know if you need more time!

broccolinisoup (commented):
Thank you so much for writing this up @siddharthkp ✨ I have a few questions 

- Would this change affect our practises in PRC, in any way, on creating and releasing components?  (Except the path is changing and we will have a bundle with a different name) 
- In the [component lifecycle docs](https://primer.style/contribute/component-lifecycle#experimental), it states that `experimental` components often are built outside of Primer - should we re-write this?

colebemis (approved):
Sounds good to me 👍

broccolinisoup:
> > Would this change affect our practises in PRC, in any way, on creating and releasing components? (Except the path is changing and we will have a bundle with a different name)
> 
> Nope, this ADR doesn't propose any changes. (I do want to create a different ADR in the future for that 😅)
> 
> > In the [component lifecycle docs](https://primer.style/contribute/component-lifecycle#experimental), it states that experimental components often are built outside of Primer - should we re-write this?
> 
> Yes! Good spot!

Lovely! Thank you 🙌🏻 

Also created https://github.com/github/primer/issues/1799  so that we don't forget 😊

broccolinisoup (approved):

justinbyo:
I very much like the consolidation proposed here. The problem statement is spot-on and I think this solution addresses it well. "Draft" is meaningless to the end consumer, and I like that we're consolidating around [the component lifecycle](https://primer.style/contribute/component-lifecycle#experimental).

@siddharthkp @broccolinisoup Can you clarify what impact this ADR will have on components labelled "experimental" currently built outside of Primer? Specifically, this statement:

> In the [component lifecycle docs](https://primer.style/contribute/component-lifecycle#experimental), it states that experimental components often are built outside of Primer - should we re-write this?

I'm concerned this might create confusion because we have "experimental" components, like DatePicker, that are in the [experimental-react-components repo](https://github.com/github/experimental-react-components). We're not planning to upstream DatePicker yet, so it will remain "experimental" and also outside the Primer lifecycle and ownership for the foreseeable future.

contributor-docs/adrs/adr-010-drafts-are-experimental.md
```diff
@@ -0,0 +1,61 @@
...
+
+We should drop the `draft` label and use `experimental` instead.
+
+While not completely accurate (UnderlineNav2 is much more polished than `experimental` maturity), the lack of confusion makes it a good trade-off.
+
+Fine-grained information about component maturity, useful for maintainers, is also communicated by the [status checklist in the component documentation](https://primer.style/react/drafts/UnderlineNav2#status).
+
+### Side effects
+
+We should create a new parallel import path for `@primer/react/experimental`.
```
joshblack, line 53: As an alternative, would love to explore other ways of indicating API stability (like through an `unstable_*` prefix through the main bundle). It would be great to limit the number of entrypoints that we ship, especially if the intent is to have them move to the main entrypoint over time.

contributor-docs/adrs/adr-010-drafts-are-experimental.md
```diff
@@ -0,0 +1,61 @@
...
+
+```diff
+// new version of component exported from /experimental
+- import { UnderlineNav } from '@primer/react/drafts'
++ import { UnderlineNav } from '@primer/react/experimental'
+```
+
+We would need to keep `@primer/react/drafts` in the short term as well for backward compatibility.
```
joshblack, line 61:
```suggestion
We would need to keep `@primer/react/drafts` in the short term as well for backwards compatibility.
```

joshblack (commented):
Definitely makes sense to consider these explorations as `experimental` 💯 

From the side-effects side, I'd separately love to explore alternatives to separate bundles to:

- Reduce the number of total entrypoints
- Provide an easier way to deprecate components, or features of a component
- Provide an easier migration process when moving between versions of a component

Specifically, right now an awkward flow is that in order to deprecate stuff we have to point teams to use a separate entrypoint for a component.

From the consumer side, I think simplifying things into "stable" and "unstable" could help to indicate the stability of an item and our commitment to maintaining its API between minor and patch versions.

contributor-docs/adrs/adr-010-drafts-are-experimental.md
broccolinisoup: I am loving this proposal! 💖 Even as a maintainer of PRC, it took me a while to wrap my head around this ~ not sure how the experience was for consumer both Hubbers and non Hubber consumers.

broccolinisoup:
>@siddharthkp @broccolinisoup Can you clarify what impact this ADR will have on components labelled "experimental" currently built outside of Primer? Specifically, this statement:
"In the [component lifecycle docs](https://primer.style/contribute/component-lifecycle#experimental), it states that experimental components often are built outside of Primer - should we re-write this?"

I'm curious to hear Sid's thoughts around this. When I wrote my question, I imagined we would be have "Primer" experimental components as well as [experimental-react-components repo](https://github.com/github/experimental-react-components) and I was proposing to update the docs to particularly reword the line of "experimental components often are built outside of Primer" which is not correct anymore and give more context of "Primer" experimental components. @justinbyo

siddharthkp:
> We're not planning to upstream DatePicker yet, so it will remain "experimental" and also outside the Primer lifecycle and ownership for the foreseeable future.

That's totally fine :)

In the docs today, we say "Proof-of-concept, often built outside of Primer". This qualification of outside/inside Primer is unnecessary because we _often_ build experimental components inside Primer as well. 😅

justinbyo:
I'm giving this a 👍 

To summarize: I think it solves a very confusing product problem for end consumers.

A couple things to keep our eye on that came up, outside the scope of this ADR:

First, now that "experimental" is a class of Primer components, we can confidently move away from calling non-Primer components experimental. Currently only two non-Primer components live in the `experimental-react-component` repo, which will most likely be deprecated in favor of the using UI monorepo for shared components. In other words, the switching cost of making "experimental" a Primer-only term is low and realistically I don't think people will be confused in practice.


Second, per @joshblack's [comment](https://github.com/primer/react/pull/2856#pullrequestreview-1288027706) – how we want to handle bundling in the future to:

> - Reduce the number of total entrypoints
> - Provide an easier way to deprecate components, or features of a component
> - Provide an easier migration process when moving between versions of a component

Third, this [issue](https://github.com/github/primer/issues/1799) to update the docs to better reflect the changes in this ADR:
- https://github.com/github/primer/issues/1799

justinbyo (approved):
👍 See [this comment](https://github.com/primer/react/pull/2856#issuecomment-1426337936) for a summary of my approval and a few considerations we need to make moving forward:

- https://github.com/primer/react/pull/2856#issuecomment-1426337936

broccolinisoup:
>we can confidently move away from calling non-Primer components experimental. Currently only two non-Primer components live in the experimental-react-component repo, which will most likely be deprecated in favor of the using UI monorepo for shared components. 

Does that mean `experimental-react-components` repo is not going to be maintained anymore?

justinbyo:
> > we can confidently move away from calling non-Primer components experimental. Currently only two non-Primer components live in the experimental-react-component repo, which will most likely be deprecated in favor of the using UI monorepo for shared components.
> 
> Does that mean `experimental-react-components` repo is not going to be maintained anymore?

There's an ADR out to make the UI monorepo the place to share non-Primer components. Once it's adopted (which is likely), we'll have to decide what to do with the `experimental-react-components` repo and we'll most likely migrate its contents to the UI monorepo. However, that decision hasn't been made yet.

- https://github.com/github/platform-ux/pull/1225

broccolinisoup:
@siddharthkp ADR seems to be approved by the decision maker @justinbyo - Should we merge it?

erinnachen:
@justinbyo @broccolinisoup This issue: https://github.com/github/platform-ux/issues/1227 tracks the plans for the `experimental-react-components` repository. I haven't moved it forward because there didn't seem urgent, but I can bump this work up in priority if there's something that needs an immediate response.

PR #2876, opened 2023-02-08
refactor(BranchName): move files to folder, add e2e tests and stories

joshblack:

Add storybook stories for `BranchName`

**Checklist**


colebemis (approved):

src/BranchName/__tests__/BranchName.test.tsx
```diff
@@ -1,10 +1,8 @@
...
-import {render, behavesAsComponent, checkExports} from '../utils/testing'
+import BranchName from '../BranchName'
+import {render, behavesAsComponent, checkExports} from '../../utils/testing'
 import {render as HTMLRender} from '@testing-library/react'
-import {axe, toHaveNoViolations} from 'jest-axe'
-
-expect.extend(toHaveNoViolations)
+import {axe} from 'jest-axe'
 
 describe('BranchName', () => {
```
broccolinisoup, line 7: Hey @joshblack sorry for dropping a comment on a merged PR. I was just confused about the new format/location of the component's tests. I see some tests are moved under `<componentName>/__tests__` which doesn't align with the [file structure ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-XXX-file-structure.md) - Curious which format to follow 👀

joshblack: @broccolinisoup maybe a convention we can follow is:

- One test file, live besides the source code
- Multiple test files, move into a `__tests__` directory

What do you think?

broccolinisoup: Either way makes sense to me. I am just happy that they are under the component folder because otherwise, I find it distracting. My only concern is the [ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-XXX-file-structure.md) suggests to keep test files besides the source code (regardless there is single file or multiple). I think we could revisit this ADR in the next react sync as it's been a while that it is accepted but not followed up with the suggested implementation - I'll add it to the agenda 😊

PR #2878, opened 2023-02-08
Update PRC ActionList implementation to have similar semantics to PVC.

radglob:
Closes https://github.com/github/accessibility-audits/issues/2939

This PR aligns the semantics (in this case, both the API and the generated markup) of `ActionList` and related components to match the implementation in PVC.

A significant change is the removal of `ActionList.Group`, which allowed sets of items to be labelled with a heading. This allowed somewhat arbitrary nesting, which we no longer want to allow.

If an `ActionList` needs a heading, users can pass the `headingProps` prop to the `ActionList`, which will then generate an `ActionList.Heading` before the `<ul>` inside the `ActionList`.

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


kendallgassner:
Hi @radglob! 

To provide more context. The [PVC ActionList was reviewed by accessibility](https://github.com/github/primer/issues/1120) already and the semantics were changed/updated to work better for users. So, we want to have semantics match to the PVC component.

langermank (commented):
Would you mind holding off on merging this until my ActionMenu stories PR merges? I think it will help us visually test. I'm noticing a visual change in how a menu like this looks, since its going from being one giant ActionList to multiple lists

before
image

your PR
image

So this _might_ mean that this is a breaking change, as the implementation might need to change.

radglob:
> Would you mind holding off on merging this until my ActionMenu stories PR merges? I think it will help us visually test. I'm noticing a visual change in how a menu like this looks, since its going from being one giant ActionList to multiple lists
> 
> before image
> 
> your PR image
> 
> So this _might_ mean that this is a breaking change, as the implementation might need to change.

Sure thing!

src/ActionMenu/ActionMenu.examples.stories.tsx
```diff
@@ -251,29 +250,33 @@ export const MultipleSections = () => {
...
-          <ActionList.Group title="Raw file content">
-            <ActionList.Item onSelect={() => alert('Workflows clicked')}>Download</ActionList.Item>
-            <ActionList.Divider />
-            <ActionList.Item onSelect={() => alert('Workflows clicked')}>Jump to line</ActionList.Item>
-            <ActionList.Item onSelect={() => alert('Workflows clicked')}>Find in file</ActionList.Item>
-            <ActionList.Divider />
-            <ActionList.Item onSelect={() => alert('Workflows clicked')}>Copy path</ActionList.Item>
-            <ActionList.Item onSelect={() => alert('Workflows clicked')}>Copy permalink</ActionList.Item>
-          </ActionList.Group>
+        <ActionList
```
langermank, line 253: This to me is this biggest breaking change. Visually it looks quite different, and the inclusion of `sx` concerns me. If it needs new styling to handle groups, we should add CSS to the component itself. Do you want to pair on this?

| before | after |
| -- | -- |
| image | image |

This is why I was hoping to see the snapshots with a diff compared to your changes here so we have a way to accurately test before/after. I think since you updated the snaps we won't see a failure here.

radglob: Yeah, it'd be good to pair on this.

radglob: Those aren't showing up for me locally (the weird overlay bubble things). Not sure what's up there.

joshblack: Super weird 🤔 @langermank I think for getting rid of them we should be able to just revert the commit that added them in from CI and then continue as normal!

src/ActionList/ActionList.stories.tsx
```diff
@@ -241,35 +241,17 @@ LinkItemPlayground.argTypes = {
...
-    type: 'string',
-  },
-}
-GroupPlayground.args = {
-  variant: 'subtle',
-  role: 'listbox',
-  title: 'Group title',
-  auxiliaryText: '',
+
+HeadingPlayground.args = {
```
langermank, line 252: I think you may need to set the `argTypes` here so we can toggle between `subtle` and `filled` variants. I wonder if you want to delete the contents of this list just to render the title, and then focus all the controls around options that the `Heading` component offers?

Right now I'm just seeing list controls
image

src/ActionList/Group.tsx
```diff
@@ -58,14 +58,14 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
...
     >
-      {title && <Header title={title} variant={variant} auxiliaryText={auxiliaryText} labelId={labelId} />}
       <GroupContext.Provider value={{selectionVariant}}>
         <Box
           as="ul"
           sx={{paddingInlineStart: 0}}
           aria-labelledby={title ? labelId : undefined}
           role={role || (listRole && 'group')}
         >
+          {title && <Header title={title} variant={variant} auxiliaryText={auxiliaryText} labelId={labelId} />}
```
langermank, line 68: Should `auxiliaryText` be `sutitle`?

src/ActionList/List.tsx
```diff
@@ -49,25 +67,34 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
...
-            showDividers,
-            role: role || listRole,
-          }}
+      <Box sx={merge(outerStyles, sxProp as SxProp)}>
+        {headingProps && <Heading {...headingProps} />}
+        <ListBox
+          sx={innerStyles}
+          role={role || listRole}
+          aria-labelledby={headingProps ? id : listLabelledBy}
+          aria-busy="true"
```
langermank, line 82: Curious about this aria-busy, why is that needed for all lists?

langermank (commented):
Once the Heading stories are updated I can take a closer look there. Otherwise left some questions/feedback!

src/ActionList/Group.tsx
radglob: This whole file should be gone, actually. I wonder if these got folded back during a merge.

src/ActionList/List.tsx
radglob: It's not needed for all lists. It fixes a violation in the Multiple Sections ActionMenu story. To my surprise, that value passes down just fine through the `ActionList` component, so I'm passing it in that story instead now.

src/ActionList/ActionList.stories.tsx
```diff
@@ -242,34 +242,53 @@ LinkItemPlayground.argTypes = {
...
 )
-GroupPlayground.argTypes = {
+
+HeadingPlayground.args = {
+  showDividers: false,
+  selectionVariant: undefined,
+  variant: 'inset',
+  headingVariant: 'subtle',
+  role: 'list',
+}
```
langermank, line 263:
```suggestion
HeadingPlayground.args = {
  variant: 'subtle',
  title: 'Group title'
}
```

src/ActionList/ActionList.stories.tsx
```diff
@@ -242,34 +242,53 @@ LinkItemPlayground.argTypes = {
...
+
+HeadingPlayground.argTypes = {
```
langermank, line 265: It wont let me suggest code :( 

We should get all this in there 
```
  variant?: 'subtle' | 'filled'
  title: string
  subtitle?: string
  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6
```

```
HeadingPlayground.argTypes = {
variant: {
    control: {
      type: 'radio',
    },
    options: ['subtle', 'filled'],
  },
title: {
    type: 'string',
  },
subtitle: {
    type: 'string',
  },
headingLevel: {
    control: {
      type: 'radio',
    },
    options: ['1', '2' etc],
  },
}
```

src/ActionList/ActionList.stories.tsx
```diff
@@ -242,34 +242,53 @@ LinkItemPlayground.argTypes = {
...
+type HeadingProps = {
+  headingVariant?: 'subtle' | 'filled'
+}
+
+export const HeadingPlayground: Story<ActionListProps & HeadingProps> = args => (
+  <ActionList headingProps={{title: 'Group title', variant: args.headingVariant}} {...args}>
+    <ActionList.Item>Copy link</ActionList.Item>
+    <ActionList.Item>Quote reply</ActionList.Item>
+    <ActionList.Item>Edit comment</ActionList.Item>
   </ActionList>
```
langermank, line 254:
```suggestion
export const HeadingPlayground: Story<ActionListProps & HeadingProps> = args => (
  <ActionList headingProps={{title: args.title, variant: args.variant, subtitle: args.subTitle, headingLevel: args.headingLevel}} {...args}>
  </ActionList>
```

If it doesn't work to just render the empty list that's fine, it just looks cleaner as this story is only focused on the heading.

radglob: That should work. And it gets around the fact that variant is overloaded.

langermank (approved):
Approving for visuals and docs 😄 it would be great if someone from eng could also approve!

src/ActionList/index.ts
```diff
@@ -18,8 +18,8 @@ export type {ActionListLeadingVisualProps, ActionListTrailingVisualProps} from '
  * Collection of list-related components.
  */
 export const ActionList = Object.assign(List, {
-  /** Collects related `Items` in an `ActionList`. */
```
siddharthkp, line 21: To keep backward compatibility, we should not remove Group from ActionList just yet. 

We can tag it as `@deprecated` with a suggestion what to use instead like we do in [PageLayout](https://github.com/primer/react/blob/main/src/PageLayout/PageLayout.tsx#L325-L339).

However, tagging a prop as `@deprecated` might fail the CI in dotcom for all the existing instances, which isn't ideal either (tagging @colebemis to confirm)

Ideally, we want this to be a 3 step process:
1. Keep both API, but recommend Heading instead of Group for new instances
2. Show a warning (not error) for old instances and refactor them overtime
3. Change the warning to error, delete Group from ActionList in next major version

I'm not sure what's the best way to only show a warning, not an error 🤔 (Good candidate for engineering sync)

siddharthkp: Bonus: because we are only deprecated a subcomponent, we don't need to keep it in `deprecated/v35/group`, we can keep it in the same file and remove it with the next major

src/ActionList/Heading.tsx
```diff
@@ -0,0 +1,74 @@
...
+  `
+
+  const Subtitle = styled.span`
+    color: ${get('colors.fg.muted')};
+    font-size: ${get('fontSizes.0')};
+    font-weight: ${get('fontWeights.normal')};
+  `
+
+  return (
+    <Box as="div" sx={merge(styles, sx as SxProp)} role="presentation" aria-hidden="true" {...props}>
```
siddharthkp, line 67: Should this still have role=presentation and aria-hidden?

ActionList.Group were only for sighted users and hidden from the screen reader, I assumed with a proper heading tag, we would want the screen reader to consume it?

siddharthkp (commented):
Left some comments about implementation detail and backward compatibility.

One change caught me off guard and could be a good starting point: 

We should get the stories reviewed for a11y (including the changes to ActionMenu and NavList)

src/theme-preval.js
```diff
@@ -1,6 +1,6 @@
 // @preval
 // This file needs to be a JavaScript file using CommonJS to be compatible with preval
-// Cache bust: 2023-02-23 12:00:00 GMT (This file is cached by our deployment tooling, update this timestamp to rebuild this file)
+// Cache bust: 2023-02-28 12:00:00 GMT (This file is cached by our deployment tooling, update this timestamp to rebuild this file)
```
siddharthkp, line 3: There are no changes to this file? Was this causing tests to fail unless you update it? 🤔

docs/content/ActionList.mdx
```diff
@@ -196,29 +196,29 @@ const SelectFields = () => {
...
-      </ActionList.Group>
-      <ActionList.Group
-        title="Hidden fields"
+      </ActionList>
+      <ActionList
         selectionVariant={
           /** selectionVariant override on Group: disable selection if there are no options */
           hiddenOptions.length ? 'multiple' : false
         }
+        headingProps={{title: 'Hidden fields'}}
```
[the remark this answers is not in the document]
radglob: That would be ideal, but this doesn't let us assign `aria-labelledby` correctly.

As far as I understand, we don't want users to add more than one heading, and we need to be able to use a generated id for the heading to label the list itself. Rendering the heading with a slot doesn't help with that, as we can't get the id out of the heading.

Also, we specifically don't want users to pass arbitrary tags to `as`, hence the `headingLevel` prop, as this can replicate inaccessible patterns.

If there's a better way to implement this with those requirements in mind, I'm happy to change this. It does feel a little clunky.

src/ActionList/Heading.tsx
radglob: That's a fair point. We can probably remove those.

docs/content/ActionList.mdx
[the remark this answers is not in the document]
siddharthkp: > and we need to be able to use a generated id for the heading to label the list itself. Rendering the heading with a slot doesn't help with that, as we can't get the id out of the heading.

Can you tell me a little more about that?

Do we want to keep the id added by the developer or can we always use a generated one?

We do the second with slots in ActionList.Item for descriptions: https://github.com/primer/react/blob/main/src/ActionList/Item.tsx#L173

> Also, we specifically don't want users to pass arbitrary tags to as, hence the headingLevel prop, as this can replicate inaccessible patterns.

We restrict the allowed values for `as` prop :) https://github.com/primer/react/blob/main/src/Heading/Heading.tsx#L10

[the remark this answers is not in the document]
radglob: I don't think we should expect developers to add an id explicitly. In this case, the id probably isn't meaningful except for ensuring that the id can be used to link via the `aria-labelledby` attribute. I don't think slots help us there since rendered components are "sealed". We can't get the id out.

Restricting the `as` values will work though. Thanks for the reference.

src/ActionList/index.ts
colebemis: > However, tagging a prop as @deprecated might fail the CI in dotcom for all the existing instances, which isn't ideal either (tagging @colebemis to confirm)

I don't think deprecated components fail CI. I'm in favor of marking this as `@deprecated` as the first non-breaking step. Then we can follow up and remove it in a major version.

colebemis (commented):
@radglob Is there a reason you went with the `headingProps` API instead of exposing `ActionList.Heading` to consumers?

```jsx
<ActionList headingProps={{title: '...'}}>
// vs
<ActionList>
  <ActionList.Heading title="...">
  ...
</ActionList>
```

radglob:
> @radglob Is there a reason you went with the `headingProps` API instead of exposing `ActionList.Heading` to consumers?
> 
> ```js
> <ActionList headingProps={{title: '...'}}>
> // vs
> <ActionList>
>   <ActionList.Heading title="...">
>   ...
> </ActionList>
> ```

This should be fixed now using slots.

docs/content/ActionList.mdx
```diff
@@ -196,29 +196,29 @@ const SelectFields = () => {
   }
 
   return (
-    <ActionList selectionVariant="multiple">
-      <ActionList.Group title="Visible fields">
+    <>
+      <ActionList selectionVariant="multiple" headingProps={{title: 'Visible fields'}}>
```
siddharthkp, line 200: Need to update this API. 

Also, I think we need to change this example's heading? Instead of `With Groups`, maybe `With Heading`?

siddharthkp, line 200: Need to update this API. 

Also, I think we need to change this example's heading? Instead of `With Groups`, maybe `With Heading`?

docs/content/ActionMenu.mdx
```diff
@@ -117,58 +117,53 @@ You can choose to have a different _anchor_ for the Menu depending on the applic
   <ActionMenu.Button>Open column menu</ActionMenu.Button>
 
   <ActionMenu.Overlay>
+    <ActionList showDividers headingProps={{title: 'Live query'}}>
```
siddharthkp, line 120: Need to update example

Ditto as ActionList, we should change the heading to `With Heading` or `divided into sections`

src/index.ts
```diff
@@ -49,7 +49,7 @@ export type {RadioProps} from './Radio'
 export {ActionList} from './ActionList'
 export type {
   ActionListProps,
-  ActionListGroupProps,
+  ActionListHeadingProps,
```
siddharthkp, line 52: ~We should probably export both for backward compatibility~

Checked dotcom, there is no usage of these types, safe to delete

.changeset/modern-coins-destroy.md
```diff
@@ -0,0 +1,38 @@
...
+  <ActionList.Group>
+    ...
+  </ActionList.Group>
+</ActionList>
+```
+
+Instead, use `ActionList`s and stack them as needed.
+```
+<div>
+  <ActionList headingProps={{ title: "Actions" }}>
```
siddharthkp, line 28: Can we update the changelog with ActionList.Heading API?

siddharthkp (commented):
Looking good, just need to update the docs in a few places

src/ActionMenu/ActionMenu.examples.stories.tsx
```diff
@@ -33,40 +33,40 @@ export const GroupsAndDescriptions = () => {
...
+              >
+                <ActionList.LeadingVisual>
+                  <MilestoneIcon />
+                </ActionList.LeadingVisual>
+                {milestone.name}
+                <ActionList.Description variant="block">Due by {milestone.due}</ActionList.Description>
+              </ActionList.Item>
+            ))}
+        </ActionList>
+        <ActionList sx={{paddingTop: 0}} selectionVariant="single">
```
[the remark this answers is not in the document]
radglob: No, this was added to keep the e2e snapshots from conflicting. It was rather annoying to get to work. I might be able to remove that and update the snapshots instead. I think I tried doing that before and it caused lots of unrelated snapshots to update as well.

src/ActionList/index.ts
```diff
@@ -18,8 +18,8 @@ export type {ActionListLeadingVisualProps, ActionListTrailingVisualProps} from '
...
-  Group,
```
siddharthkp, line 22: We want to keep this, right?

reference: https://github.com/primer/react/pull/2878/files#r1147703351

radglob: Ah, yeah! I removed this when I thought we were deprecating it by moving the path.

.changeset/modern-coins-destroy.md
siddharthkp: @radglob I think this was resolved without an update 🤔

siddharthkp (approved):
My previous comments are all marked as resolved, ship it! :shipit:

PR #2882, opened 2023-02-08
refactor(Breadcrumbs): move files to folder, add e2e tests and stories

joshblack:

Add storybook stories for `Breadcrumbs`

**Checklist**

- [ ] (Optional) Create a Playground story


colebemis (approved):

src/Breadcrumbs/__tests__/Breadcrumbs.test.tsx
```diff
@@ -1,10 +1,8 @@
 import React from 'react'
-import {Breadcrumbs, Breadcrumb} from '..'
-import {render, behavesAsComponent, checkExports} from '../utils/testing'
+import {default as Breadcrumbs, Breadcrumb} from '..'
```
broccolinisoup, line 2: This is not related this PR but we seem to use mix of default and named exports - I am not sure if there is any pattern or we do it randomly?

broccolinisoup (approved):

PR #2883, opened 2023-02-08
refactor(Flash): move files to folder, add e2e tests and stories

joshblack:

Add storybook stories for `Flash`

**Checklist**


colebemis (approved):

src/Flash/Flash.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react'
+import {Meta, Story} from '@storybook/react'
```
broccolinisoup, line 2: Should we use `MetaComponent` and `StoryComponent` here? https://github.com/storybookjs/storybook/discussions/16650#discussioncomment-1617480
Or was there any reason we used this version?

joshblack: @broccolinisoup great catch, I've just been using existing templates but I think you're dead-on that we should be using `MetaComponent` and `StoryComponent`.

cc @langermank just a heads up for above 👆, it also seems to address props like `as` and `forwardedAs` causing issues on `argTypes` 🥳

Issue #2881, opened 2023-02-08
Primer React Box component examples appear out of date


ktravers:
### Description

In the docs for the [Box component](https://primer.style/react/Box), the examples all use props that no longer appear to be available for this component.

For example, my team recently tried to use the `gridGap` prop as documented [here](https://primer.style/react/Box#grid), which resulted in the following linter failure: https://janky.githubapp.com/80188803/output.

### Steps to reproduce

1. In a codebase with Primer React important, attempt to use any of the Box component code examples as currently documented. For example:
    ```jsx
    <Box display="grid" gridTemplateColumns="1fr 1fr" gridGap={3}>
      <Box p={3} borderColor="border.default" borderWidth={1} borderStyle="solid">
        1
      </Box>
      <Box p={3} borderColor="border.default" borderWidth={1} borderStyle="solid">
        2
      </Box>
      <Box p={3} borderColor="border.default" borderWidth={1} borderStyle="solid">
        3
      </Box>
    </Box>
    ```
1. Observe `primer-react/no-system-props` error(s)

### Version

35.18.0

### Browser

Chrome, Safari, Firefox

broccolinisoup:
👋🏻 @ktravers Thanks for taking the time to file this issue! For context, these props (such as `gridGap`) are [system props](https://system-props.com/) and [we are moving away from them](https://primer.style/react/system-props) - sorry that the docs were not up-to-date and possibly misleading - we'll make sure to update them. `sx` prop can be used for styling on our components. I hope this is helpful - please let us know if you have any concerns 🙌🏻

siddharthkp:
Created a PR for review with a quick fix: https://github.com/primer/react/pull/2886

PR #2886, opened 2023-02-09
Docs: use sx instead of system props in Box examples

siddharthkp:
- Fixes https://github.com/primer/react/issues/2881


broccolinisoup (approved):
Looks great ✨ Thanks for doing this Sid!

Issue #2888, opened 2023-02-09
Remove Chromatic integration


joshblack:
The following checks are triggered by Chromatic for Visual Regression Testing:

- UI Tests: used to indicate any regressions in storybook
- Storybook Publish: used to upload the storybook from the PR to Chromatic

These both are triggered by this workflow file: https://github.com/primer/react/blob/main/.github/workflows/visual_testing.yml

Unfortunately, the current Chromatic plan means that these tests are pending once we hit our snapshot limit. They also are unable to work from forks since the secret used for the Chromatic API Key is inaccessible to these PRs.

Given our move to [Playwright for Visual Regression Testing](https://github.com/primer/react/issues/2781), this issue will disable and remove the workflow file above and disable Chromatic for the project.

PR #2891, opened 2023-02-10
Delete visual_testing.yml

joshblack:
Closes https://github.com/primer/react/issues/2888

This PR removes the workflow for our integration with Chromatic


broccolinisoup (approved):
Wohoo! 🎉

PR #2893, opened 2023-02-10
Button API alignment

langermank:
This PR includes **breaking changes** to bring the PRC Button up to the agreed upon API spec and match implementation with PVC.

- Changes `leadingIcon` and `trailingIcon` to `leadingVisual` and `trailingVisual` (a pattern introduced with ActionList)
- Removes `Button.Counter` as a child component, replacing it with a `count` prop. This change allows us to use the `trailingVisual` slot for counters, and also aligns better with all the other child elements within Button (visuals are props, not child components in this case.)
- Removes the `outline` button variant as we wish to only support `invisible` buttons.

### Screenshots

The only visual change is that the `outline` button variant has been removed.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


langermank:
If someone wants to pair with me on how to add a codemod to handle all these prop changes, that would be great!

.changeset/flat-onions-sort.md
```diff
@@ -0,0 +1,7 @@
+---
+"@primer/react": major
+---
+
+- Changes `leadingIcon` and `trailingIcon` to `leadingVisual` and `trailingVisual`
+- Removes `Button.Counter` as a child component, replacing it with a `count` prop. This change allows us to use the `trailingVisual` slot for counters.
```
broccolinisoup, line 6: Is the prop name `count` or `trailingVisualCount`? 🤔

docs/content/Button.mdx
```diff
@@ -98,17 +90,12 @@ A separate component called `IconButton` is used if the action shows only an ico
 </>
 ```
 
-### Counter component
+### Button with counter
 
-A common use case for primer is a button with a counter component which shows the child count value.
-We provide `Button.Counter` as a composite component which requires you to provide a number as child.
-The counter will match the `variant` styles of the parent button.
+To show a count value as a trailing visual inside `Button`, pass a value to the `count` prop. The counter will match the `variant` styles of the parent button.
```
broccolinisoup, line 95: Same comment as above (prop name)

joshblack (approved):
🥳

PR #2901, opened 2023-02-14
feat(project): add warning, invariant utilities and dev-expression plugin

joshblack:
This is a follow-up to the ADR: https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-012-development-warnings.md

It adds several helpers:

- `warning` for emitting warnings based on a condition
- `warn` as an alternative to `console.warn()` that is removed in production builds
- `invariant` which is used to assert states that must be true before moving on

The `warning` and `invariant` helpers are transformed by the added plugin: [`babel-plugin-deve-expression`](https://www.npmjs.com/package/babel-plugin-dev-expression). This plugin will transform calls in the following way:

```js
// Before
warning(condition, message);

// After
if (process.env.NODE_ENV !== 'production') {
  warning(condition, message);
}
```


broccolinisoup (approved):
Looks great! Should we do a follow up PR to refactor the existing `console.warn`s especially without the `__DEV__`  wraps?

joshblack:
@broccolinisoup good call, I went ahead and opened up an ADR and included that as part of the impact of this kind of convention: https://github.com/primer/react/blob/321801aca743c519bd2dd40fe0f49503f2f60a26/contributor-docs/adrs/adr-XXX-development-warnings.md

If folks feel okay about this approach then I think you're right that we should go through and make sure calls to `console.warn()` are either wrapped in a `__DEV__` block or use `warning`

joshblack:
Re-requesting your review @broccolinisoup after that ADR was merged! Let me know if the changes make sense, I also added in an `invariant()` helper to help out with some of the DataTable stuff I'm working. Hope that's okay, let me know if not and I can remove!

src/utils/__tests__/warning.test.ts
```diff
@@ -0,0 +1,30 @@
+import {warning} from '../warning'
+
+test('emits a message to console.warn() when the condition is `false`', () => {
```
broccolinisoup, line 3: Should this be 'emits a message to console.warn() when the condition is `true`' ?

`warning(true, 'test')` -> warn if condition true 🤔  Or did I confuse myself?

joshblack: You're right @broccolinisoup! Sorry about that, I updated the implementation but forgot to change the test title. I'll make that update now 👍

broccolinisoup (approved):
Looks great ✨

PR #2910, opened 2023-02-17
Add `isolation:isolate` to `ButtonGroup` container

iansan5653:
Using `z-index` in reusable components is generally a bad idea, because `z-index` is one of the most subtle side-effecting CSS rules in existence. It will inevitably break things down the road. The only way to use `z-index` internally without affecting surrounding UI is to create a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) to isolate the effects

In `ButtonGroup`, the `z-index` on the focused element _is_ necessary to be able to show the focused button's outline above the other buttons. So we can't remove it altogether. But we can definitely isolate it, and there's no reason not to - the only goal here is to make the element show above its fellow buttons - not above the entire app. The easiest way to make a new stacking context with no other effects is `isolation: isolate`.

Fixes https://github.com/github/memex/issues/8722

### DIFF
--- a/.changeset/fair-lobsters-sin.md
+++ b/.changeset/fair-lobsters-sin.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Add `isolation:isolate` to `ButtonGroup` container
--- a/src/ButtonGroup.tsx
+++ b/src/ButtonGroup.tsx
@@ -6,6 +6,7 @@ import {ComponentProps} from './utils/types'
 const ButtonGroup = styled.div`
   display: inline-flex;
   vertical-align: middle;
+  isolation: isolate;
 
   && > * {
     position: relative;
--- a/src/__tests__/deprecated/__snapshots__/Button.test.tsx.snap
+++ b/src/__tests__/deprecated/__snapshots__/Button.test.tsx.snap
@@ -301,6 +301,7 @@ exports[`ButtonGroup renders consistently 1`] = `
   display: -ms-inline-flexbox;
   display: inline-flex;
   vertical-align: middle;
+  isolation: isolate;
 }
 
 .c0.c0 > * {


joshblack (approved):

joshblack:
Loved your explanation for this @iansan5653, thanks for taking the time to write that up 🙏 

Do you know if there is a good lint rule or test that we could include to try and catch this stuff across components? Would love to make sure `z-index` usage by default also uses `isolation` unless it requires coordination in a parent context that we don't control.

iansan5653:
🤔 I think it would be pretty tricky to figure out how to lint for this. You might be able to test for it by walking up the DOM tree and checking for a new stacking context but even that would be tricky to get right. Alternatively we could lint for _all_ instances of `z-index` / `zIndex` and force a lint disable comment explaining why it is needed and how it is contained. 

There is an open [tech debt issue](https://github.com/github/memex/issues/12553) in Memex on the same topic where I outline a set of rules that we can universally follow. We can document this in a flowchart:

```mermaid
flowchart TD
    A[Does this need to render<br> over the entire application?]
    A -->|Yes| B
    A -->|No| C
    B(Use <code>Portal</code> to render into<br> the default <code>primerPortalRoot</code>)
    C[Can you rearrange the DOM to<br> control stacking?]
    C -->|Yes| D
    C -->|No| E
    D(Render the element that should<br> be on top last in the DOM)
    E(Use <code>z-index</code> and also apply <code>isolation: isolate</code><br> to the component's container)
```

---

Alternatively, maybe we can create a new Primer abstraction to help with this? Just thinking out loud here, but maybe some sort of context provider that also acts as a stacking context container, like so:

```tsx

const AvatarStackingContext = createStackingContext()

const AvatarStack = (users) => (
  <AvatarStackingContext.Container>
    {users.map((user, i) => (
      <AvatarStackingContext.Layer level={users.length - i} key={user.login}>
        <Avatar user={user} />
      </AvatarStackingContext.Layer>
    )}
  </AvatarStackingContext.Container>
)
```

It's verbose, but maybe it's worth it for the explicit management of stacking here? Maybe we can make it more intuitive with named layers, or validate the layers somehow (maybe only allow each layer number to be used once). Or maybe it's overkill and we should just use `isolation: isolate` 🤷.

Issue #2836, opened 2023-01-30
Add instructions to test React apps built with primer/react


Zenulous:
Recently I've been working on a project that utilizes this component library. We use jest and `react-test-renderer` to create tests for our frontend. There are issues when running the tests due to CSS definitions: 
![image](https://user-images.githubusercontent.com/45333650/215474243-50db72f4-580d-418f-bccc-43da41aca5d9.png)

Solutions for this include importing helpers like https://github.com/primer/react/blob/main/src/utils/test-helpers.tsx, and I think it can be valuable to refer to this in documentation as it took me quite some time to figure out how to properly mock CSS.

lesliecdubs:
👋🏻 @JoseInTheArena would this be of interest to Web Systems in regards to providing documentation about how to test React apps on the paved path?

PR #2916, opened 2023-02-20
Add guidance on testing to the `Getting Started` docs

broccolinisoup:
Closes https://github.com/github/primer/issues/1823
Closes https://github.com/primer/react/issues/2836

### Screenshots

No visual changes

### Merge checklist


docs/content/getting-started.md
```diff
@@ -134,3 +134,7 @@ Of course, customize the array based on the `@types/` packages you have installe
 ## Silencing warnings
 
 Like React, Primer React emits warnings to the JavaScript console under certain conditions, like using deprecated components or props. Similar to React, you can silence these warnings by setting the `NODE_ENV` environment variable to `production` during bundling.
+
+## Testing
+
+Testing your application with Primer React is no different than testing your application with any other React library. Depending on your test environment and the testing libraries you use, you may need polyfills. For example if you are using `jest`, it runs via Node runtime and using [JSDOM](https://github.com/jsdom/jsdom) as a DOM implementation, so you will need to mock some browser APIs. We have [helpers](https://github.com/primer/react/blob/main/src/utils/test-helpers.tsx) that you can utilize to mock some of these APIs.
```
joshblack, line 140: Would it be helpful to also  add a note/code snippet on how to include this file of test helpers in Jest or would that be better placed somewhere else in the project?

broccolinisoup: I think it is a great point but I don't think we export the file? I imagined they could copy the mocks they needed from the test-helpers but it would be ideal to import the file. Maybe I am missing something here though.

colebemis: I feel like copying and pasting those mocks is a reasonable approach if we want to avoid increasing the API surface area of Primer React.

siddharthkp: One vote for exporting test-helpers, I think copy pasting is a secret trick that we tell people after they ask us why it broke 😅

broccolinisoup: Oh sorry yes I just realised we export the file! 

Would below be accurate to document?
```
import "@primer/react/lib-esm/utils/test-helpers"; (For ESM)
import "@primer/react/lib/utils/test-helpers"; (For CommonJS)
```

broccolinisoup: Thank you! Just pushed a commit

PR #2919, opened 2023-02-21
PageHeader: Rename stories and add e2e tests

broccolinisoup:
Tracking issue: https://github.com/github/primer/issues/1488

Add storybook stories for `PageHeader`

**Checklist**


ollie-iterators:
I think you need to generate screenshots

broccolinisoup:
@langermank This is a part of [story renaming](https://github.com/github/primer/issues/1488) but because it is a layout component I wasn't sure how the `Default` story would look like 🤔 Any ideas?

joshblack (approved):

langermank:
@broccolinisoup oh this is tricky! I think providing a very basic case for `Default` makes sense. I would go with what loads in the `Playground` I think, so just the title.

PR #2920, opened 2023-02-21
Docs: Change system props to sx for another page

siddharthkp:
Found another example of using system props in documentation, replaced with sx prop


docs/content/overriding-styles.mdx
```diff
@@ -40,13 +40,10 @@ Just like [values passed to system props](https://styled-system.com/responsive-s
...
 ```jsx live
 <Box
-  borderWidth="1px"
-  borderStyle="solid"
-  borderColor="border.default"
-  borderRadius={2}
-  p={2}
   sx={{
-    bg: ['neutral.subtle', 'accent.subtle', 'success.subtle', 'attention.subtle', 'danger.subtle'],
+    display: ['flex', 'flex', 'block', 'block],
```
siddharthkp, line 44: using a slightly better example here.

joshblack (approved):

docs/content/overriding-styles.mdx
broccolinisoup, line 44:
```suggestion
    display: ['flex', 'flex', 'block', 'block'],
```
I think it is missing a single quote

siddharthkp: good eye! thank you!!

PR #2921, opened 2023-02-21
Update storybook-tests.yml

joshblack:
Use Node.js v16 instead of v18 in `storybook-tests.yml`. This should fix the timeout issue that we had starting last week.

### DIFF
--- a/.github/workflows/storybook-tests.yml
+++ b/.github/workflows/storybook-tests.yml
@@ -7,11 +7,14 @@ on:
 
 jobs:
   test:
-    timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


.github/workflows/storybook-tests.yml
```diff
@@ -7,11 +7,14 @@ on:
 
 jobs:
   test:
-    timeout-minutes: 60
+    timeout-minutes: 30
```
broccolinisoup, line 10: Thanks! - yes it was very long 😅

broccolinisoup (approved):
🚀

PR #2928, opened 2023-02-22
docs(adr): add development-warnings adr

joshblack:
ADR for Development Warnings in `@primer/react`

[Rendered markdown](https://github.com/primer/react/blob/7e423651a13836a1ecebd0198c2d939fa37f027b/contributor-docs/adrs/adr-012-development-warnings.md)

[Pull Request](https://github.com/primer/react/pull/2901)

### DIFF
--- a/contributor-docs/adrs/adr-XXX-development-warnings.md
+++ b/contributor-docs/adrs/adr-XXX-development-warnings.md
@@ -9,7 +9,7 @@ Proposed
 There are situations where we would like to provide warnings to developers who
 use `@primer/react` that something may be deprecated, unsupported, etc. Often,
 these will be emitted using `console.warn()`. When using `console.warn()` by
-itself, we run into a situation where code that is only be meant for development
+itself, we run into a situation where code that is only meant for development
 is included in production code. As a result, it would be helpful to establish
 patterns around how to provide warnings to developers in order to make sure
 that:
@@ -20,7 +20,7 @@ that:
 
 ## Decision
 
-Code that is meant for development-only warnings **must** be wrapped within a
+Code that is meant for development-only warnings or checks **must** be wrapped within a
 `__DEV__` block.
 
 ```tsx
@@ -32,7 +32,7 @@ function ExampleComponent() {
 ```
 
 Under-the-hood, the `__DEV__` block will be compiled to a `NODE_ENV` check so
-that it is stripped when `NODE_ENV` is `production`.
+that it is stripped when `NODE_ENV` is set to `'production'`.
 
 > **Note**
 > Contributors may wrap hooks within a `__DEV__` block even though hooks are not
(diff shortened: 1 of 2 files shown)


mperrotti (approved):
I like this direction! Once this merges, could we create an issue to implement the `warning()` helper?

contributor-docs/adrs/adr-XXX-development-warnings.md
```diff
@@ -0,0 +1,60 @@
...
+> Contributors may wrap hooks within a `__DEV__` block even though hooks are not
+> meant to be called conditionally. This is because the `__DEV__` check can be
+> considered constant in that it will always be true or false for the
+> environment (development or production).
+
+When a contributor would like to communicate a warning to a developer, they
+should use the `warning()` helper.
+
+```ts
+warning(condition, 'This is the message that is logged when condition is false-y')
```
colebemis, line 47: I would have guessed that the message would display if the condition is truthy. I read that code as "warn if _condition_"

If this is a convention, I'm fine to keep it. But I figured I'd mention my initial confusion.

If the name of the helper was something like `assert()`, I think falsey would make more sense.

joshblack: Totally makes sense! Happy to flip it if that's more intuitive for folks 👍 

I think the convention stems from [`invariant()`](https://github.com/facebook/fbjs/blob/51e174bead2c0bb6bb9c62d7cd2d63338bcef2d6/packages/fbjs/src/__forks__/invariant.js#L30) ultimately, and things like `assert()`, that frame things from the perspective of "this condition must be met". When it's not met, throw an error, log a warning, etc

To contextualize the two, here is the warning from `Heading` in the two styles:

```tsx
// Warn if condition is true
warning(
  innerRef.current && !(innerRef.current instanceof HTMLHeadingElement),
  'This Heading component should be an instanceof of h1-h6'
);
```

```tsx
// Warn if condition is false
warning(
  innerRef.current && innerRef.current instanceof HTMLHeadingElement,
  'This Heading component should be an instanceof of h1-h6'
);
```

joshblack:
Kind of a break-out from: https://github.com/primer/react/pull/2928#discussion_r1117264832

Would we want this to also address unrecoverable situations? Or keep it scoped to development warnings?

By unrecoverable, I think of things like FB's `invariant()` helper that allows you to state conditions that must be met in order to proceed. If they aren't met, an error is thrown with a custom message.

I feel like TS covers a lot of this but that there sometimes is validation that needs to happen if the type system cannot cover a situation.

contributor-docs/adrs/adr-XXX-development-warnings.md
broccolinisoup: Thanks Josh! Warning when the condition truthy feels more intuitive to me. That said, it is totally cool to me if we want follow the convention that it stems from.

broccolinisoup (approved):

lesliecdubs:
@joshblack so far it seems most folks are in alignment here, but if you'd like me to assign a decision-maker on this one give me a shout.

Issue #2927, opened 2023-02-22
Button style overrides for size medium break in 35.20


mattcosta7:
### Description

When applying size small or large, the sx overrides apply, howevever the color of the text loses its override when size=medium

https://user-images.githubusercontent.com/8616962/220628972-f09fcfa5-5f7b-49a7-9391-9bd9b2723627.mov

### Steps to reproduce

Add a button with sx overrides, and apply size=small vs size=medium

### Version

v35.20.0

langermank:
Thanks for filing this issue! I can reproduce this, however `medium` is the default size 🤔 so you don't ever really need to set `size="medium"`. With that in mind, I'm not sure this is worth fixing (unless its an easy fix, which I suspect it might not be.) Tagging @broccolinisoup to see if she has any thoughts!

mattcosta7:
> With that in mind, I'm not sure this is worth fixing (unless its an easy fix, which I suspect it might not be.) Tagging @broccolinisoup to see if she has any thoughts!

I'd disagree that it's not worth fixing - unless we remove `medium` as a valid size prop?  it seems like we shouldn't expose an API that breaks the styling of the feature?

broccolinisoup:
Thanks @mattcosta7 for reporting the issue! I added a quick fix and seems to be working on the story https://github.com/primer/react/pull/2929

PR #2929, opened 2023-02-23
Button: Do not add [data-size="medium"]` when size is explicitly specified as medium

broccolinisoup:
Describe your changes here.

Closes https://github.com/primer/react/issues/2927

### Context
The issue is when `size="medium"` prop is passed to `Button`, the styles provided via `sx` prop don't apply. With the way we apply `Button` styles that were provided via `sx`  is described here. 

### Issue and solution
The reason that wasn't working for the `size="medium"` is because the function is adding `[data-size="medium"]` but we don't have any styles defined with that data-attribute (Because the default size of `Button` is `medium` therefore it is styled on the base class) so it was simply disregarding everything. Simply checking if the size is not equal to `medium`solves the issue.

### Screenshots

No visual changes

### Merge checklist

- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/calm-moons-serve.md
+++ b/.changeset/calm-moons-serve.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Button: Fixes the style override issue when size is explicitly specified as medium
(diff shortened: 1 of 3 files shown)


langermank (approved):
Thank you for this! 🚀

joshblack (approved):

PR #2932, opened 2023-02-23
2841 use githubaxe config as config for axe

### LINKED ISSUE #2841: Use `@github/axe-config` as config for axe
We currently configure `axe` inside of Playwright through our `toHaveNoViolations` matcher: https://github.com/primer/react/blob/main/e2e/matchers/toHaveNoViolations.ts#L47. This issue is to update the configuration here to use the `@github/axe-github` package: https://www.npmjs.com/package/@github/axe-github. These rules can be passed to the `axe.configure()` call that currently exists.

green6erry:
This is a stop gap to close out #2841. There were issues with integrating `@axe-github` with`@playwright/test` and so, for the sake of time, energy, and priorities, @joshblack and @erinnachen gave the okay that we can just manually add the rules/checks needed for now.

### Screenshots
When looking for axe violations with Playwright, we are now using the manually added rules and checks that are generated from `@axe-github` https://www.npmjs.com/package/@github/axe-github

image

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


.changeset/spicy-donkeys-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
```
joshblack, line 2: I think you can safely remove this changeset since it's not impacting anything library related! 👍

joshblack (changes requested):
Looks great! Thanks for doing this 🙏 Just left a comment for the changeset but should be good after that 👍

joshblack (approved):
🎉

PR #2935, opened 2023-02-24
feat: add Blankslate component

joshblack:
Adds the `Blankslate` component to PRC. I didn't see this component already implemented, if there is an alternative or if this component is deprecated let me know! Figured it'd be a quick one to add if it was helpful.

> **Note**
> This is not being added to a bundle, like drafts, at the moment. It's intent is to be used for some DataTable work to demo loading and empty states. Let me know if there is a better pattern or something to use instead!

**References**

- [Interface Guidelines](https://primer.style/design/ui-patterns/empty-states)
- [Primer CSS](https://primer.style/css/components/blankslate)
- [PVC](https://primer.style/view-components/components/beta/blankslate)


src/Blankslate/Blankslate.tsx
```diff
@@ -0,0 +1,119 @@
...
+export type BlankslateProps = React.PropsWithChildren<{
+  /**
+   * Add a border around this component
+   */
+  border?: boolean
+
+  /**
+   * Constrain the maximum width of this component
+   */
+  narrow?: boolean
```
langermank, line 49: I feel like this should just be a variant with `narrow` and `spacious` but for now it makes sense to just follow the existing API!

langermank (approved):
I like those data-attributes ❤️

src/Blankslate/Blankslate.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import {BookIcon} from '@primer/octicons-react'
+import {Meta, Story} from '@storybook/react'
+import React from 'react'
+import {Blankslate} from '../Blankslate'
+import {ComponentProps} from '../utils/types'
+
+export default {
+  title: 'Drafts/Components/Blankslate',
+  component: Blankslate,
```
colebemis, line 9: Could you add `'Blankslate'` to the list of components in the storybook test? https://github.com/primer/react/blob/main/src/__tests__/storybook.test.tsx

src/Blankslate/Blankslate.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+  subcomponents: {
+    'Blankslate.Visual': Blankslate.Visual,
+    'Blankslate.Heading': Blankslate.Heading,
+    'Blankslate.Description': Blankslate.Description,
+    'Blankslate.PrimaryAction': Blankslate.PrimaryAction,
+    'Blankslate.SecondaryAction': Blankslate.SecondaryAction,
+  },
+} as Meta<typeof Blankslate>
+
+export const Default: Story<ComponentProps<typeof Blankslate>> = args => (
```
colebemis, line 19: Default stories should accept args. Maybe we can create a Playground story that accepts args.

docs/content/Blankslate.mdx
```diff
@@ -0,0 +1,123 @@
+---
+title: Blankslate
+componentId: blankslate
+---
+
+import data from '../../src/Blankslate/Blankslate.docs.json'
+
+Use `Blankslate` when there is a lack of content within a page or section. Use as placeholder to tell users why something isn't there.
```
colebemis, line 8: If Blankslate is not being exported from any bundle, we should either not include it in the docs or add a blurb explaining why it's not in the main bundle.

src/Blankslate/Blankslate.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+export type VisualProps = React.PropsWithChildren
+
+function Visual({children}: VisualProps) {
+  return <Box sx={{color: 'fg.muted', mb: 3}}>{children}</Box>
+}
+
+export type HeadingProps = React.PropsWithChildren<{
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | undefined
```
colebemis, line 72: Is `undefined` necessary here?
```suggestion
  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
```

src/Blankslate/index.tsx
```diff
@@ -0,0 +1,9 @@
+import {Blankslate as BlankslateImpl, Visual, Heading, Description, PrimaryAction, SecondaryAction} from './Blankslate'
+
+export const Blankslate = Object.assign(BlankslateImpl, {
```
colebemis, line 3: We usually combine the component and subcomponents in the `Blankslate.tsx` file? Is there a reason you moved it to the index file?

colebemis (commented):
The implementation looks good! Just left a few minor comments about docs and storybook stories.

src/Blankslate/index.tsx
joshblack: I think it was due to trying to avoid a `default` export in `Blankslate.tsx`. Updated `Blankslate.tsx` to export a default export with the assignment 👍

colebemis (approved):

PR #2937, opened 2023-02-27
TextArea: Resolve storybook axe violation that was caused by explicitly setting an id

broccolinisoup:
Describe your changes here.

No need to explicitly set an id for `TextArea` component on its story. The id is inferred from the `args`

Closes https://github.com/github/primer/issues/1863

### Screenshots

No visual changes

### Merge checklist

- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/stories/Textarea.stories.tsx
+++ b/src/stories/Textarea.stories.tsx
@@ -69,7 +69,7 @@ export const TextareaStory = (args: FormControlArgs<TextareaProps>) => {
     <Box as="form" sx={{p: 3}}>
       <FormControl {...parentArgs}>
         <FormControl.Label {...labelArgs} />
-        <Textarea id="textarea" {...args} />
+        <Textarea {...args} />
         {captionArgs.children && <FormControl.Caption {...captionArgs} />}
         {validationArgs.children && validationArgs.variant && (
           <FormControl.Validation {...validationArgs} variant={validationArgs.variant} />


src/stories/Textarea.stories.tsx
```diff
@@ -69,7 +69,7 @@ export const TextareaStory = (args: FormControlArgs<TextareaProps>) => {
     <Box as="form" sx={{p: 3}}>
       <FormControl {...parentArgs}>
         <FormControl.Label {...labelArgs} />
```
broccolinisoup, line 71: @joshblack the `<FormControl.Label />` is intended to be visually hidden in this example and I remember we talked about it in the last axe-clean call. I didn't fix it here because it is "broken" on some other form elements as well, [TextInput](https://primer.style/react/storybook/?path=/story/components-forms-textinput--default) for example. I am also not sure if we should just fix them to be actually visually hidden or would it make sense to fix it and maybe add another story with labels visible? I am just thinking loudly.

joshblack (approved):

src/stories/Textarea.stories.tsx
joshblack: Oh wild 🤔 I think you're right that being able to have stories with and without visual labels would be helpful just to demonstrate this (with the default being having a visual label)

broccolinisoup: Reported an issue for this: https://github.com/primer/react/issues/2947

PR #2948, opened 2023-02-28
Add story source code to components.json

colebemis:
This PR adds source code for storybook stories to the generated `components.json` file so that we can display that source code in the storybook embeds on primer.style/design (example: https://primer.style/design/components/tree-view/react#examples)

CleanShot 2023-02-27 at 19 26 01@2x

## Before/After

`generated/component.json`:

CleanShot 2023-02-27 at 19 27 59@2x


langermank (approved):
This is such an exciting improvement! 🚀

script/components-json/build.ts
```diff
@@ -45,5 +99,59 @@ fs.writeFileSync('generated/components.json', JSON.stringify(data, null, 2))
...
+
+      const code = prettier
+        .format(generate(func).code, {
+          parser: 'typescript',
+          singleQuote: true,
+          trailingComma: 'all',
+          semi: false,
+        })
+        .trim()
+        .replace(/;$/, '')
```
joshblack, line 145: This might cover the `;` usage at the beginning too that seems to be showing up 🤔 

```suggestion
        .replace(/;$/, '')
        .replace(/^;/, '')
```

joshblack (approved):
Woohoo! 🥳 Just left a comment for the leading `;` but looks amazing 🔥

script/components-json/build.ts
colebemis: Good catch. Done ✅

PR #2951, opened 2023-02-28
feat: add DataTable and Table to drafts

joshblack:
Closes https://github.com/github/primer/issues/1814

This PR adds `DataTable` and `Table` to the drafts bundle with the following API:

- `DataTable`
- `Table.*`
  - `Table.Container`
  - `Table.Title`
  - `Table.Subtitle`
  - `Table`
  - `Table.Head`
  - `Table.Body`
  - `Table.Row`
  - `Table.Header`
  - `Table.Cell`

The idea behind this implementation is that `DataTable` is an alternative to `Table` and can slot into that position:

```tsx
<Table.Container>
  <Table.Title id="table-title">Repositories</Table.Title>
  <DataTable
    aria-labelledby="table-title"
    // ...
  />
</Table.Container>
```

`Table` may also be used there and each part of `Table` is available under that identifier.

One risk with this approach is that `Table.*` contains both descendants of a `Table` and ancestors like `Container`, `Title`, or `Subtitle`. The risk here would be that certain terms may be overloaded, like in `Table.Header`. This would refer to a `<th>` but in the context of a layout within `Table.Container` would refer to the header region of a layout.

#### Changelog

**New**

- Add tests for `Table.*` components

**Changed**

- Update `interface` to `type`
- Split up into `DataTable.tsx` and `Table.tsx`
- Add `Table` and `DataTable` to drafts entrypoint
- Add explicit `role` tags for `th`, `tr`, `td` for usage of `display: grid` for column width options

**Removed**


.changeset/spotty-beers-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Add DataTable, Table to drafts entrypoint
```
broccolinisoup, line 5: I am wondering when will be the best time to start exporting new components in `experimental` bundle instead of `drafts`? According to [the ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-010-drafts-are-experimental.md)

src/DataTable/DataTable.features.stories.tsx
```diff
@@ -1,35 +1,13 @@
...
-  Table,
-  TableHead,
-  TableBody,
-  TableRow,
-  TableHeader,
-  TableCell,
-  TableContainer,
-  TableTitle,
-  TableSubtitle,
-} from '../DataTable'
```
colebemis, line 14: ✨

src/drafts/index.ts
```diff
@@ -1,9 +1,12 @@
...
+ * This is the place where we keep components that are not part of the public
+ * api yet (not in main bundle). We don't recommend using it in production.
  *
- *  But, they are published on npm and you can import them for experimentation/feedback.
- *  example: import {ActionList} from '@primer/react/drafts
+ * But, they are published on npm and you can import them for experimentation/feedback.
+ * example: import {ActionList} from '@primer/react/drafts
  */
+
+export {DataTable, Table} from '../DataTable'
```
colebemis, line 9: Should we export any types?

colebemis (approved):
Exciting to see this take shape!

src/drafts/index.ts
joshblack: Makes sense to me as part of the definition of done. To implement this, should I include props as exports for any exported component?

.changeset/spotty-beers-cheer.md
joshblack: Great point, reached out about this over in Slack to see where that work is. For this PR I'll add it to `drafts` and will add it to `experimental` once that gets created 👀

PR #2953, opened 2023-02-28
feat(DataTable): add support for actions and different layouts

joshblack:
Closes https://github.com/github/primer/issues/1889

Add support for the different layouts for a `DataTable` (excluding filter). This adds a couple of stories that show support for the table with both a single action and multiple actions. It also includes support for a footer but this will be implemented by [pagination](https://github.com/github/primer/issues/1890)

This PR adds two new components to accomplish this: `Table.Divider` and `Table.Actions`. It extends `Table.Container` to be a CSS Grid with different areas for the parts of the table.

#### Feedback

I would love to hear what you all think about this approach for the layout and if there are any other alternatives that I should explore here or that you suggest using!

#### Changelog

**New**

- Add `Table.Divider` for rendering a divider in a table
- Add `Table.Actions` for rendering a group of actions for a table
- Add stories for the following situations:
  - With action
  - With action only (no title)
  - With actions
  - With actions only (no title)

**Changed**

- Add support for `sx` prop to `Table.Container`
- `Table.Container` is now a CSS Grid
- Add `.Table` class name for `Table` component

**Removed**

### DIFF
--- a/src/DataTable/Table.tsx
+++ b/src/DataTable/Table.tsx
@@ -3,7 +3,7 @@ import React from 'react'
 import styled from 'styled-components'
 import Box from '../Box'
 import {get} from '../constants'
-import {SxProp} from '../sx'
+import sx, {SxProp} from '../sx'
 import {SortDirection} from './sorting'
 
 // ----------------------------------------------------------------------------
@@ -138,12 +138,6 @@ const StyledTable = styled.table<React.ComponentPropsWithoutRef<'table'>>`
     font-weight: 600;
     text-align: start;
   }
-
-  /* Spacing if table details are present */
-  .TableTitle + &,
-  .TableSubtitle + & {
-    margin-top: ${get('space.2')};
-  }
 `
 
 export type TableProps = React.ComponentPropsWithoutRef<'table'> & {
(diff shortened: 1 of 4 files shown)


joshblack:
Quick question for you @mperrotti, should the title of the table align to the top edge of the actions container? Or do we want them to be aligned vertically centered?

**Top-aligned**

Screenshot 2023-02-28 at 4 21 46 PM

**Centered**
Screenshot 2023-02-28 at 4 21 55 PM

langermank (approved):
Your CSS is real nice ✨ love the use of grid areas! Stories look nice too

colebemis (approved):
💖

mperrotti:
> should the title of the table align to the top edge of the actions container? Or do we want them to be aligned vertically centered?

Vertically centered is good. I know PageHeader does top-aligned, but that's because they have a different set of layout problems where vertically-centered is more likely to create weird whitespace.

src/DataTable/Table.tsx
```diff
@@ -275,10 +269,63 @@ function TableCell({children, scope, ...rest}: TableCellProps) {
 // ----------------------------------------------------------------------------
 // TableContainer
 // ----------------------------------------------------------------------------
+const StyledTableContainer = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  grid-template-areas:
```
mperrotti, line 275: 😍 great use of grid

mperrotti (approved):

Issue #2952, opened 2023-02-28
Support non-string content in `Token`


iansan5653:
`Token` only supports strings as content, via the `text` prop. This is inconsistent with `Label`, which is a similar component that supports full `ReactNode` as `children`.

There doesn't seem to be any reason not to support complex content here, such as formatted text. 

I propose one of these options:

- Add a `children` prop but leave the `text` prop for legacy uses
- Add a `children` prop and delete the `text` prop as a breaking change
- Allow `ReactNode` instead of `string` for the `text` prop as a way to avoid a breaking change

joshblack:
Definitely +1 to `ReactNode`, totally makes sense to have them consistent as long as there is no issues with nested markup. It doesn't seem like we'd run into that with Token since we wouldn't have a situation of nested interactive content in an interactive tag, seemingly.

broccolinisoup:
Thanks @iansan5653 for reporting this 🙌🏻 Agreed to @joshblack +1 for the `ReactNode`

iansan5653:
🤔 So it turns out it's a little more complicated than I expected. It looks like we force `string` content because we use the content to build an `aria-label` when the token is deletable, to indicate the keyboard shortcut for deletion. Ie: "Bug, press backspace or delete to remove". For example (shown the the most complex case when the token is interactive):

```html
<span aria-label="Bug, press backspace or delete to remove">
  <button>Bug</button>
  <span tabindex="-1" aria-hidden><!-- delete icon --></button>
</span>
```

A couple of options to address this:

- Add a `hidden` span with this content and use `aria-describedby` to link this hint text. This feels more ergonomic because it won't have to be read for every single token (particularly annoying when there are many tokens). And by not building a string from the content of the token, we avoid mandating that the contents are easily stringable.
- Render the deletion hint as a `VisuallyHidden` child of the element. This would not help with redundancy but would be an improvement over the current experience when the token is interactive, because it would properly label the button instead of the wrapping `span`.
- Remove the keyboard shortcut hint and make the delete button interactive (focusable and non-hidden with the correct role) instead.

I'm somewhat partial to the last one as it seems to also help keyboard users (who currently don't see the hint and also can't focus the delete button). Maybe it would be confusing to have multiple buttons called "Delete" when there is a set of tokens, but that could be solved from the parent container by using `list`/`listitem` to wrap the tokens.

iansan5653:
[...]

PR #2954, opened 2023-02-28
Allow `React.ReactNode` as `Token` content

iansan5653:
Updates the props of `Token` to accept any `ReactNode` instead of just `string` in the `text` prop. This is a non-breaking change because all strings will still work fine.

This required updating the token somewhat because we were previously generating an `aria-label` like `${text}, press backspace or delete to remove`. If `text` was, say, an object, this label would become `[object Object], press backspace or delete to remove`. So instead, we now render this as a `VisuallyHidden` text inside the token so it becomes part of the token's computed content without overriding it. This is more accessible anyway because it doesn't require placing a label on a non-interactive `span`.

Closes https://github.com/primer/react/issues/2952


joshblack (approved):
Thanks for doing this! 🙌

PR #2956, opened 2023-02-28
Adds Storybook and source code links to StateLabel docs

mperrotti:
Now that we have StateLabel stories, we can link to the Storybook.


broccolinisoup (approved):
🚀

PR #2959, opened 2023-03-01
Checkbox stories

langermank:
As part of https://github.com/github/primer/issues/1488

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  formControlArgTypesWithoutValidation,
+  getFormControlArgsByChildComponent,
+} from '../utils/form-story-helpers'
+
+const excludedControlKeys = ['required', 'value', 'validationStatus', 'sx']
+
+export default {
+  title: 'Components/Checkbox',
+  component: Checkbox,
+  parameters: {controls: {exclude: excludedControlKeys}},
```
joshblack, line 16: Nice pattern 🔥

src/CheckboxGroup/CheckboxGroup.features.stories.tsx
```diff
@@ -0,0 +1,130 @@
+import React from 'react'
+import {Checkbox, CheckboxGroup, FormControl, Box} from '..'
+
+export default {
+  title: 'Components/CheckboxGroup/Features',
```
joshblack, line 5: Do we want to include `component` and `subcomponent` keys on the feature/example stories or nah?

src/__tests__/CheckboxGroup.test.tsx
```diff
@@ -172,5 +172,4 @@ describe('CheckboxGroup', () => {
   })
 })
 
-checkStoriesForAxeViolations('CheckboxGroup/fixtures')
-checkStoriesForAxeViolations('CheckboxGroup/examples')
+checkStoriesForAxeViolations('CheckboxGroup.features', '../CheckboxGroup/')
```
joshblack, line 175:
```suggestion
```

This should be covered by the Playwright tests now 🎉

langermank: OH what! Okay, so if I add Playwright tests for specific stories I can remove this check?

src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+    </Box>
+  )
+}
+Playground.args = {
+  checked: false,
+  indeterminate: false,
+  visuallyHidden: false,
+  disabled: false,
+  labelChildren: 'Default checkbox',
+  captionChildren: 'Always unchecked unless `checked` is set to true in Storybook controls',
```
mperrotti, line 38: Non-blocking nitpick: maybe write the property keys for children like this so they don't look like prop names:
```
['label children']: 'Default checkbox'
```

This would apply to `labelChildren`, `captionChildren`, and `validationChildren`.

mperrotti (approved):

src/Checkbox/Checkbox.stories.tsx
langermank: I really like this idea, but I couldn't get it to work 🤔 maybe it needs to happen also in the form-story-helpers.. I think I would prefer to follow up on this after I finish all of the form stories

joshblack (approved):

src/Checkbox/Checkbox.stories.tsx
mperrotti: Aw man that's too bad. Yea, we can absolutely come back to this later if it ends up being a problem.

colebemis (commented):
@langermank Could you add `Checkbox` and `CheckboxGroup` to the storybook tests file? https://github.com/primer/react/blob/main/src/__tests__/storybook.test.tsx

PR #2965, opened 2023-03-01
test(PageHeader): remove axe check

joshblack:
Remove the axe checks for the stories from `PageHeader` as these should be covered by the e2e test. This also helps to address a flakiness problem we've been having with this test recently.


broccolinisoup (approved):
Thanks for doing this! ✨

PR #2966, opened 2023-03-01
Update `IssueLabelToken` colors

colebemis:
Introduces @lukasoppermann's new light mode colors for issue labels.

:sparkles: [Storybook preview](https://primer-0ed3cda262-13348165.drafts.github.io/storybook/?path=/story/components-issuelabeltoken--default-token)

Closes https://github.com/github/primer/issues/1896

### Before

https://user-images.githubusercontent.com/4608155/222282563-400f560e-f2db-47a3-b0fe-f1f483036a03.mp4

### After

https://user-images.githubusercontent.com/4608155/222282225-42e66bbd-65bd-4c9f-acac-cf705d7962f9.mp4


package.json
```diff
@@ -109,6 +109,7 @@
     "focus-visible": "^5.2.0",
     "fzy.js": "0.4.1",
     "history": "^5.0.0",
+    "hsluv": "1.0.0",
```
mperrotti, line 112: Ohh this looks nice. Is it worth adding another dependency that does color functions? We already have `color2k`.

If you think it's worth it I'm on board 👍

mperrotti (approved):
Code changes look great! There could be a potential design issue that could occur when we change this in production. If there are issue labels that are the same hue with different lightness and saturation, they will look almost identical.

https://user-images.githubusercontent.com/2313998/222288966-a2703d07-8dab-4a11-852a-01b6812e143b.mp4

package.json
lukasoppermann: color2k does not support `hsluv` as far as I am aware.

lukasoppermann:
> Code changes look great! There could be a potential design issue that could occur when we change this in production. If there are issue labels that are the same hue with different lightness and saturation, they will look almost identical.

Hey @mperrotti, very true. I mentioned this in the discussion that if we ignore saturation we will get the same colors. I don't really have a solution for this though and I am not sure there is one.

Do you see this as a problem, or are you just suggesting we need to test and inform people so that this can be addressed?

mperrotti:
Yea I know this is already an issue on dark mode. I think it's safe to move forward with these. We'll just keep an eye on how/if it feels broken in a staff ship.

package.json
mperrotti: No I don't think it does, but it still supports the kind of color manipulations we're doing here. I'm ok with keeping this dependency though. I couldn't find a bundle size, but I imagine it's tiny.

colebemis: > color2k does not support hsluv as far as I am aware.

Yup. color2k doesn't allow us to convert between hex and hsluv, so we need this new dependency.

mattcosta7:
Looks like the changes here fix an issue on light themes (other than the base light them), that would be beneficial to push and continue improving on later 

see the pages build here: https://github.com/github/github/pull/267491

Issue #2947, opened 2023-02-28
Labels of form elements on storybook have `VisuallyHidden` classes but they are not visually hidden


broccolinisoup:
## Issue
Looks like the form elements on storybook ([TextInput](https://primer.style/react/storybook/?path=/story/components-forms-textinput--default) as one example) are intended to be visually hidden (see [the code reference](https://github.com/primer/react/blob/main/src/utils/story-helpers.tsx#L334)) but they are not actually visually hidden. It came up during one of the `axe-clean` [violation fix](https://github.com/primer/react/pull/2937#discussion_r1118226584) and would be great to fix all these form elements that have the same problem. 

## Proposed solution

Removed the `VisuallyHidden` parameter from the `Form.Label` component (code [ref1](https://github.com/primer/react/blob/main/src/stories/TextInput.stories.tsx#L58), [ref2](https://github.com/primer/react/blob/main/src/utils/story-helpers.tsx#L334)) in below components' stories and make sure the Default story has a visual label and add another story to demonstrate the component with a visually hidden title.

`Autocomplete`, `InlineAutocomplete`, `Select`, `TextArea`, `TextInput`, `TextInputWithTokens`

### Steps to reproduce

1. Go to the Primer React storybook https://primer.style/react/storybook
2. Check out any of the Form Elements that were mentioned above https://primer.style/react/storybook/?path=/story/components-forms-autocomplete--add-new-item
3. Inspect and see the label element in the dom that has `VisuallyHidden` class.

### Version

v35.20.0

### Browser

Chrome, Safari, Firefox, Edge, iOS Safari

mperrotti:
I'm not understanding the problem. The `visuallyHidden` prop is meant to allow users to provide a label and hide it visually but keep it accessible to assistive technologies.

I don't think we should merge this, but if we do it would be a breaking change for people who currently use the `visuallyHidden` prop on `FormControl.Label`.

PR #2970, opened 2023-03-02
Only render `VisuallyHidden` component when the component is visually hidden

broccolinisoup:
This PR turns out to be more of a proposal than fix. 

I was looking into [this issue](https://github.com/primer/react/issues/2947) and realised that the labels are not intended to be visually hidden, I don't think at least. The `VisuallyHidden` classname in the dom is probably misleading. What is actually happening here is that apparently we have `isVisible` prop for `VisuallyHidden` component and when it is `true` we still render the content as `VisuallyHidden` but keep it visible, see [this](https://github.com/primer/react/blob/main/src/_InputLabel.tsx#L36) as an example. 

I was wondering would it reduce confusion if we only use `VisuallyHidden` component to render visually hidden content and in visible cases, we can render as usual like I pushed in this PR. Looking forward to hearing what you think

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Autocomplete/AutocompleteMenu.tsx
+++ b/src/Autocomplete/AutocompleteMenu.tsx
@@ -9,7 +9,7 @@ import Spinner from '../Spinner'
 import {useSSRSafeId} from '../utils/ssr'
 import {AutocompleteContext} from './AutocompleteContext'
 import {PlusIcon} from '@primer/octicons-react'
-import VisuallyHidden from '../_VisuallyHidden'
+import ToggleVisibility from '../_ToggleVisibility'
 
 type OnSelectedChange<T> = (item: T | T[]) => void
 type AutocompleteMenuItem = MandateProps<ItemProps, 'id'>
@@ -300,7 +300,7 @@ function AutocompleteMenu<T extends AutocompleteItemProps>(props: AutocompleteMe
   }, [selectedItemIds, setSelectedItemLength])
 
   return (
-    <VisuallyHidden isVisible={showMenu}>
+    <ToggleVisibility isVisible={showMenu}>
       {loading ? (
         <Box p={3} display="flex" justifyContent="center">
           <Spinner />
@@ -322,7 +322,7 @@ function AutocompleteMenu<T extends AutocompleteItemProps>(props: AutocompleteMe
           )}
         </div>
       )}
-    </VisuallyHidden>
+    </ToggleVisibility>
   )
 }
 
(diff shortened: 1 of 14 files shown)


mperrotti:
> the labels are not intended to be visually hidden, I don't think at least

It's intended to let consumers choose to provide a label that's only accessible to assistive technologies but isn't visually shown.

mperrotti (commented):
I don't think I'm understanding the problem with `<VisuallyHidden>`. Maybe we can pair and come up with a solution together?

joshblack:
Just wanted to share a couple other ideas for ways to use `VisuallyHidden` here that came up during the meeting!

<details>
<summary>Use directly on text label</summary>

```tsx
const InputLabel: React.FC<React.PropsWithChildren<Props>> = ({
  children,
  disabled,
  htmlFor,
  id,
  required,
  visuallyHidden,
  sx,
  as = "label",
}) => {
  const label = visuallyHidden ? (
    <VisuallyHidden>{children}</VisuallyHidden>
  ) : (
    children
  );
  return (
    <Box
      as={as}
      sx={}
      htmlFor={htmlFor}
      id={id}
      sx={{
        fontWeight: "bold",
        fontSize: 1,
        display: "block",
        color: disabled ? "fg.muted" : "fg.default",
        cursor: disabled ? "not-allowed" : "pointer",
        alignSelf: "flex-start",
        ...sx,
      }}
    >
      {required ? (
        <Box display="flex" as="span">
          <Box mr={1}>{label}</Box>
          <span aria-hidden="true">*</span>
        </Box>
      ) : (
        label
      )}
    </Box>
  );
};

```
</details>

<details>
<summary>Conditional</summary>

```tsx
const InputLabel: React.FC<React.PropsWithChildren<Props>> = ({
  children,
  disabled,
  htmlFor,
  id,
  required,
  visuallyHidden,
  sx,
  as = "label",
}) => {
  const label = (
    <Box
      as={as}
      sx={}
      htmlFor={htmlFor}
      id={id}
      sx={{
        fontWeight: "bold",
        fontSize: 1,
        display: "block",
        color: disabled ? "fg.muted" : "fg.default",
        cursor: disabled ? "not-allowed" : "pointer",
        alignSelf: "flex-start",
        ...sx,
      }}
    >
      {required ? (
        <Box display="flex" as="span">
          <Box mr={1}>{children}</Box>
          <span aria-hidden="true">*</span>
        </Box>
      ) : (
        children
      )}
    </Box>
  );

  if (visuallyHidden) {
    return <VisuallyHidden>{label}</VisuallyHidden>;
  }

  return label;
};

```
</details>

<details>
<summary>Crazy idea</summary>

Basically move `visuallyHidden` out from a prop and have people compose through `children`

```tsx
<VisuallyHidden>
  <InputLabel />
</VisuallyHidden>
```
</details>

PR #2972, opened 2023-03-02
Add classname to button for gh/gh experiment

siddharthkp:
Companion PR for https://github.com/github/primer/issues/1894

Add a className to Button component to create a reliable selector for gh/gh

To select a specific variant, use data-attribute: `.prc-Button[data-variant="primary"]`


siddharthkp:
Don't need this anymore :)

PR #2974, opened 2023-03-02
Fix ellipsis truncation in `Token`

iansan5653:
Text in `Token` is supposed to truncate with an ellipsis when the `Token` reaches 100% of the container width. But the `TokenTextContainer` has `overflow: visible`, which disables truncation. Removing that rule allows the preceding `overflow: hidden` line to activate. 

However, in inline elements `overflow` cannot be `hidden` only in the x direction, so `overflow: hidden` will also clip descenders (like the tail on a "g"). To fix this, we also need to also change the `line-height` to `normal`.

Before: 
screenshot of token showing text spilling past the end

After: 
screenshot of token showing text truncated with ellipsis (three dots)


joshblack:
@iansan5653 this might be out of scope for this PR, but from an a11y perspective would we need to provide a mechanism for someone to view the full label contents if it is truncated? Otherwise could this fall under: https://www.w3.org/TR/WCAG21/#resize-text

joshblack:
@iansan5653 thanks for taking the time to write this up! I appreciate it so much 🙏 And definitely agreed with what you shared around.

I'm not sure how much we could build-in to this component due to the different circumstances you shared, but what would you think of this as a rule of thumb:

1. If you can avoid truncation, please do that instead
2. If you need to truncate, use a tooltip
3. If the content that is being truncated is interactive, place the tooltip on the interactive element
4. If the content that is being truncated is non-interactive, use a toggletip

joshblack:
@iansan5653 definitely 👍 it seems like this is similar to some other components so I'll make a note of this under the umbrella of truncation so we can revisit it!

joshblack (approved):

PR #2976, opened 2023-03-02
test(axe): default `color-contrast` check to false

joshblack:
After talking to @langermank, it seems like it might reduce friction to default the `color-contrast` check to `false`. This will be enabled to true once the color work in primitives is complete.


langermank (approved):

broccolinisoup (approved):
LOVE THIS SO MUCH

PR #2986, opened 2023-03-06
[Looking for feedback] For ADRs, convert status into a table

siddharthkp:
We add a status to each ADR (`Draft | Approved | Adopted`), but often you'll run into code that does not follow the ADR. 

This happens because accepting an ADR is not equal to refactoring old code to match ADR guidelines, which can be confusing when referring to old code on how to do a thing.

**Proposal: We should call this difference out in the ADR until it is completely adopted.**

Examples:

// [adrs/adr-001-typescript.md](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-001-typescript.md)

| Stage    | Status |
| -------- | ------ |
| Approved | ✅     |
| Adopted  | ✅     |

// [adrs/adr-009-interaction-tests.md](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-009-interaction-tests.md)

| Stage    | Status |
| -------- | ------ |
| Approved | ✅     |
| Adopted  | 🚧     |

This doesn't answer the question of "what needs to be done" or "what is remaining". Ideally, each accepted ADR should be accompanied with an issue detailing the work that needs to be done to `Adopt` that ADR

Example:

| Stage    | Status |
| -------- | ------ |
| Approved | ✅     |
| Adopted  | 🚧   [primer/react#2986](https://github.com/primer/react/pull/2986)  |


contributor-docs/adrs/adr-002-behavior-isolation.md
```diff
@@ -2,9 +2,13 @@
...
 ## Status
 
-~~Approved 2021-02-04~~
 Superseded by [ADR 010](./adr-010-behavior-isolation.md) (2023-01-31)
 
+| Stage    | Status |
+| -------- | ------ |
+| Approved | ✅     |
+| Adopted  | ✅     |
+
```
keithamus, line 11: Presuming that these are here for at-a-glance understanding of the progress of these ADRs, and given this one is now superseded, does it make sense to add a row or otherwise turn these emojis into :warning: or something?

contributor-docs/adrs/adr-010-behavior-isolation.md
```diff
@@ -2,7 +2,10 @@
 
 ## Status
 
-Accepted
+| Stage    | Status |
+| -------- | ------ |
+| Approved | ✅     |
+| Adopted  | ❓     |
```
keithamus, line 8:
```suggestion
| Adopted  | ✅     |
```

joshblack (commented):
Definitely sounds helpful! Would love to give it a try.

contributor-docs/adrs/adr-002-behavior-isolation.md
lesliecdubs: +1 to Keith. If an ADR is superseded, I might suggest removing both "Approved" and "Adopted" statuses and introducing "Superseded" with a link to the superseding ADR.

contributor-docs/adrs/adr-004-children-as-api.md
```diff
@@ -2,7 +2,10 @@
 
 ## Status
 
-Approved 2022-05-10
+| Stage    | Status |
+| -------- | ------ |
+| Approved | ✅     |
+| Adopted  | ❌     |
```
lesliecdubs, line 8: If we have an issue for adoption, might it be useful to include a link in the `Adopted` status column? And if we don't have an issue, we probably should!

rezrah (commented):
Love the idea @siddharthkp, thanks for proposing this ✨ 

- I'm wondering if emoji's alone are suitable to express status. Sometimes status of a particular ADR can be incredibly nuanced, which is quite difficult to convey using visual imagery alone. Not sure if this is good for screen reader users either? 🤔 

  As an example, I'm not sure what ❌ or ❓means here without needing to read supplementary info. I'm also more interested in the reasons it's not ✅. 
  
  I also anticipate a lot of incoming PRC users could struggle to disambiguate these. Being verbose and explicit could be helpful in increasing adoption of these ADR's elsewhere around GitHub.
  
  Building on what @keithamus and @lesliecdubs suggested, instead of using `approved` for example, I think `active` and `superseded` may be enough to communicate status. `Approved` in GH nomenclature is a bit redundant in hindsight, as the ADR was intrinsically approved at some point or other to be in the repo.

  E.g. of what could be enough to convey status for drive-by readers
  
  **Status**: ✅ Active 
  **Implemented in**: `n/a` {link to issue/pr}

  or

  **Status**: ⚠️ Superseded by {link to adr/issue}
  **Implemented in**: {link to issue/pr}


- Whatever is decided here, we should consider hoisting the outcome of it to an ADR template in github/primer. If we can be consistent with updating the status at authoring/merge time, the ADR author can also be responsible for creating the implementation issues. [Here's an example of the template I've created for Primer Brand](https://raw.githubusercontent.com/primer/brand/main/contributor-docs/adrs/adr-template.md), which uses hidden HTML comments to encourage the author to select one of the indicated statuses. I'd love for us to have a consistent way to do this at an org level across all libraries.

contributor-docs/adrs/adr-004-children-as-api.md
siddharthkp: Strong agree!

contributor-docs/adrs/adr-010-behavior-isolation.md
siddharthkp: thank you!

siddharthkp:
> As an example, I'm not sure what ❌ or ❓means here without needing to read supplementary info. I'm also more interested in the reasons it's not ✅.

Oops sorry,❓is not a status, it's just a placeholder I used because I don't know the status of the ADR. Not going to commit that. 😅

> I think active and superseded may be enough to communicate status. Approved in GH nomenclature is a bit redundant in hindsight, as the ADR was intrinsically approved at some point or other to be in the repo.

I agree with you that Approved is redundant if the ADR is merged. We have some ADRs that are "active" but only partially implemented.

I think it's important to convey three pieces of information here:

1. Is this ADR still relevant or has it been superseded
2. Has this ADR been fully adopted (implemented) yet?
3. Where can I find more about the implementation details

I think we can assume an ADR is active by default, unless mentioned otherwise.

I like the idea of 


| Stage    | Status |
| -------- | ------ |
| Adopted | ✅     |

| Adopted  | ❌   [primer/react#2986](https://github.com/primer/react/pull/2986)  |

colebemis (commented):
My first impression is that ❌ feels like a final state. It reads as "won't do" or "failed" rather than "not started" or "in progress."

I wonder if 🚧 would be a more appropriate emoji.

siddharthkp:
> My first impression is that ❌ feels like a final state. It reads as "won't do" or "failed" rather than "not started" or "in progress." I wonder if 🚧 or 🟡 would be a more appropriate emoji since they read more as "pending."

Makes sense! Love  🚧 old school internet vibe
CaCapitolHill1114Under_Construction

lesliecdubs:
Pinging @siddharthkp, can you review outstanding comments and re-request reviews here as needed so we can get this over the line?

joshblack (approved):

PR #2994, opened 2023-03-07
ADR 013: Location of experimental react components maintained by Primer

siddharthkp:
[See rendered version](https://github.com/primer/react/blob/adr-013/contributor-docs/adrs/adr-013-primer-experimental-location.md)

### DIFF
--- a/contributor-docs/adrs/adr-013-primer-experimental-location.md
+++ b/contributor-docs/adrs/adr-013-primer-experimental-location.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-❌ Not Adopted, <issue with implementation details goes here>
+❌ Not Adopted, `<issue with implementation details goes here>`
 
 &nbsp;
 
--- a/contributor-docs/adrs/adr-013-primer-experimental-location.md
+++ b/contributor-docs/adrs/adr-013-primer-experimental-location.md
@@ -1,0 +1,63 @@
+# ADR 013: Location of experimental react components maintained by Primer
+
+## Status
+
+| Stage    | Status |
+| -------- | :----: |
+| Accepted |   ❌   |
+| Adopted  |   ❌   |
+
+## Context: Experimental components (or drafts) in primer/react
+
+As we work on new components (or rewrites of old components), we often need to start with a "experiment" and not include it in the semantically versioned (semver-ed) main bundle.
+
+[ADR-006](./adr-006-drafts.md) explains our use of "drafts", here are the relevant parts:
+
+- When we are not sure about the API of a component, we put it in a different bundle: "primer/react/drafts". [ADR 010: Merge drafts status into experimental](./adr-010-drafts-are-experimental) recommends changing the name to `experimental`
+- The contract with consumers is that you are opting into a component in which we don't have 100% confidence yet, the API and structure can change.
+
+In [ADR 008: Lowering the barrier to entry](./adr-008-experimental-components) (Sep, 2022), we decided to create room to share experimental components between memex and the monolith with a new repository ([github/experimental-react-components](https://github.com/github/experimental-react-components)). Now that [memex is part of the monolith](https://github.com/github/planning-tracking/issues/1755), that [repository has been archived](https://github.com/github/platform-ux/issues/1248) (Feb, 2023).
+
+Which leaves us with: Where should experimental components maintained by Primer live?
+
+&nbsp;
+
+### Problems with publishing experimental components in @primer/react npm package:
+
+**Incompatible with semver**: Because we haven't reached the final API or structure, we need to ship breaking changes to experimental components. In the current setup, these breaking changes are part of minor or patch releases to primer/react.
+
+On the consumer side, this makes every upgrade a potentially breaking upgrade nullifying the benefits of semver.
+
+In the case of multiple teams working on the monolith, this can mean finding breaking changes for a new experimental component in a part of the app that you're not familiar with.
+
+&nbsp;
+
+## Decision
+
+1. Experimental components, as defined by the [component lifecycle](https://primer.style/contribute/component-lifecycle#experimental), should not be part of semantically versioned npm package for `@primer/react`.
+
+2. Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@primer/react-experimental-underlinenav`.
+
+3. The source code for all experimental components should continue to live in the same repository ([primer/react](https://github.com/primer/react)).
+
+### Tradeoffs
+
+1. The responsibility of using compatible versions of experimental components and primer/react is shifted to the consumer.
+2. This would result in lots of new packages in the `@primer` scope on npm, creating some clutter.
+
+&nbsp;
+
+### Other options considered
+
+The source code for experimental components can be hosted inside the monolith in [github/github/ui/packages](https://github.com/github/github/tree/master/ui/packages).
+
+Pros:
+
+1. Easier to experiment with production use cases.
+2. Eliminates the need for versioning and publishing components.
+
+Cons:
+
+1. Harder to build and test changes that depend on companion changes in primer/react.
+2. Breaking changes have to be adopted by all teams & usage using the experimental at the same time.
+3. Harder to track and limit usage with controlled collaboration while it's still experimental or not accessible. Failed experiments become tech debt for the application.
--- a/contributor-docs/adrs/adr-013-primer-experimental-location.md
+++ b/contributor-docs/adrs/adr-013-primer-experimental-location.md
@@ -39,10 +39,11 @@ In the case of multiple teams working on the monolith, this can mean finding bre
 
 3. The source code for all experimental components should continue to live in the same repository ([primer/react](https://github.com/primer/react)).
 
-### Tradeoffs
+### Tradeoffs / Risks
 
 1. The responsibility of using compatible versions of experimental components and primer/react is shifted to the consumer.
 2. This would result in lots of new packages in the `@primer` scope on npm, creating some clutter.
+3. When, both primer/react and dotcom include an experimental component in their dependencies, it's easy to end up with 2 major versions of an experimental package. This should be okay as long as they don't conflict. (for example, PageLayout might use an experimental Stack or ActionList might use an experimental implementation of slots)
 
 &nbsp;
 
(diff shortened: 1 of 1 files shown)


joshblack (commented):
Thanks for putting this up! So exciting 🥳 

I love the idea of PRC moving into a monorepo, not just for delivering experimental packages but also to help with internal tools and architecture. I also appreciate the callout around managing expectations for semver and how it will be delivered into the monorepo.

I had some feedback and questions after going through it, thanks again for putting this up!

**Feedback**

- Consider a naming scheme like `@primer/react-{component}-{stage}` versus `@primer/react-{stage}-{component}`
  - `@primer/react` (current)
  - `@primer/react-{component}` (unused)
  - `@primer/react-{component}-{stage}` (proposed)


**Questions**

- Could you detail the steps of moving a component from experimental to being in stable? How do teams move from an experimental package to a stable package?
- Are we recommending that `@primer/*-experimental` packages are installed in specific workspaces and avoid installing them in the `primer` workspace on dotcom?
  - If they are installed in the `primer` workspace like other dependencies, I would imagine we'd run into the same problem identified where a single update breaks multiple workspaces
- Will experimental packages be released on a different schedule to `@primer/react` or will they keep the same schedule? (or publish as-needed?)
- Is there an impact on the new primer.style architecture for docs information?
- What conventions would there be for these components in storybook?
- Where would common helpers live that need to be shared between both `@primer/react` and experimental packages?

contributor-docs/adrs/adr-013-primer-experimental-location.md
```diff
@@ -0,0 +1,62 @@
...
+
+As we work on new components (or rewrites of old components), we often need to start with a "experiment" and not include it in the semantically versioned (semver-ed) main bundle.
+
+[ADR-006](./adr-006-drafts.md) explains our use of "drafts", here are the relevant parts:
+
+- When we are not sure about the API of a component, we put it in a different bundle: "primer/react/drafts". [ADR 010: Merge drafts status into experimental](./adr-010-drafts-are-experimental) recommends changing the name to `experimental`
+- The contract with consumers is that you are opting into a component in which we don't have 100% confidence yet, the API and structure can change.
+
+In [ADR 008: Lowering the barrier to entry](./adr-008-experimental-components) (Sep, 2022), we decided to create room to share experimental components between memex and the monolith with a new repository ([github/experimental-react-components](https://github.com/github/experimental-react-components)). Now that [memex is part of the monolith](https://github.com/github/planning-tracking/issues/1755), that [repository has been archived](https://github.com/github/platform-ux/issues/1248) (Feb, 2023).
```
dusave, line 18: Might want to clarify that `experimental-react-components` is the one that got archived, not `memex` 😅

contributor-docs/adrs/adr-013-primer-experimental-location.md
```diff
@@ -0,0 +1,62 @@
...
+
+In the case of multiple teams working on the monolith, this can mean finding breaking changes for a new experimental component in a part of the app that you're not familiar with.
+
+&nbsp;
+
+## Decision
+
+1. Experimental components, as defined by the [component lifecycle](https://primer.style/contribute/component-lifecycle#experimental), should not be part of semantically versioned npm package for `@primer/react`.
+
+2. Each experimental component should be independently published and versioned with semver. Breaking changes should be tagged as major versions. Example: `@primer/react-experimental-underlinenav`.
```
dusave, line 38: > Breaking changes should be tagged as major versions

_Post 1.0_. It's important to note that pre-1.0, it is totally acceptable and expected for components to have breaking changes. Once 1.0 has been reached, following breaking changes should cause a major version bump

contributor-docs/adrs/adr-013-primer-experimental-location.md
```diff
@@ -0,0 +1,62 @@
...
+
+3. The source code for all experimental components should continue to live in the same repository ([primer/react](https://github.com/primer/react)).
```
dusave, line 40: Will these experimental components be published externally as well? I _assume_ yes

contributor-docs/adrs/adr-013-primer-experimental-location.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Tradeoffs
+
+1. The responsibility of using compatible versions of experimental components and primer/react is shifted to the consumer.
```
dusave, line 44: _Kind of_. I think this is an acceptable tradeoff, but I'm curious: what happens when they reach stable?

I would love to see Primer move to a monorepo that houses all stable _and_ experimental components as individual packages, with the `@primer/react` package essentially importing and re-exporting all components. That gives us a few things:

1. It would let experimental components continue to be separate packages once they become stable, thereby eliminating the need to update usages in code (i.e. change imports from the individual package to the bundled package)
2. It would allow for all the code to live in one place, making it easier to share and re-use components
3. It would allow for consumers to install/import _only what they need_. While tree-shaking covers us today, this is still a nice to have to be able to have dependencies on a smaller subset of components, in the event of incompatibility, breaking change churn, and bugs that might pop up.

contributor-docs/adrs/adr-013-primer-experimental-location.md
```diff
@@ -0,0 +1,62 @@
...
+2. This would result in lots of new packages in the `@primer` scope on npm, creating some clutter.
+
+&nbsp;
+
+### Other options considered
+
+The source code for experimental components can be hosted inside the monolith in [github/github/ui/packages](https://github.com/github/github/tree/master/ui/packages).
```
dusave, line 51: This is where things currently reside for some components (e.g. QueryBuilder). Would the plan be to migrate those out?

dusave (commented):
👀 This is exciting!

contributor-docs/adrs/adr-013-primer-experimental-location.md
siddharthkp: I can see that working as well. You can assume every release can be a breaking one, so major/minor might not be as informative or even as common.

If we follow pre-1.0 versions, I guess the ceremonious 1.0 would be a move to stable primer/react.

contributor-docs/adrs/adr-013-primer-experimental-location.md
siddharthkp: Yep! Just like primer/react

contributor-docs/adrs/adr-013-primer-experimental-location.md
siddharthkp: I like the idea in theory, but I really don't like the maintenance overhead it comes with of matching compatible versions both in the library and application.

wrote an unpublished rant about this earlier: https://www.sid.st/unpolished/component-versioning

contributor-docs/adrs/adr-013-primer-experimental-location.md
siddharthkp: I don't know enough about QueryBuilder🤔  Is this a component maintained by primer developers?

dusave: heh that is the question. The _Web Component_ is currently in `app/components/primer/experimental`, while the React wrapper is in `app/assets/modules/react-shared`. So... partially? 😛

contributor-docs/adrs/adr-013-primer-experimental-location.md
dusave: That write up is a good summary, but I think it also misses a key benefit: If I am only consuming `Button`, I don't care about upgrades to other components, nor do I want to have to deal with the potential of some side effect causing `Button` to behave in an unexpected way. Publishing individual packages solves that.

Also, wrt Composite components (e.g. your Button/Dialog example), versioning is a bit more nuanced. If the Dialog implements the new button, _does_ it require a major version bump? I would argue no, unless the API changes. If the implementation details are hidden away from the user, a minor or patch version could be used.

However on the flip side, there is still the larger dependency management question/mess. If I upgrade `Dialog`, but want to keep my old usage of `Button`, now I have to manage two versions of the component. Not _that_ bad, but would be compounded by the addition of more components.

lesliecdubs:
Thanks for putting this together, @siddharthkp!

ADR decision maker: @colebemis 
Decision deadline: EOD Friday, 7 April or sooner

colebemis (approved):
After discussing with @siddharthkp and @dgreif (Web Systems), this seems like the right path forward ✅ 

A few notes from my discussions:
- Is there a downside to having a circular dependency between `@primer/react` and `@primer/react-*-experimental` packages? If so, can we set up a linter to prevent creating a circular dependency?
- FYI from @dgreif: github/github prevents having multiple versions of the same package, so it won't be possible to install multiple versions of experimental components in dotcom.
- Implementing this ADR will require us to set up monorepo tooling in primer/react. github/github uses `nx`.

@lesliecdubs Should we create an issue to track the implementation of this ADR?

colebemis:
@joshblack 

> Is there an impact on the new primer.style architecture for docs information?

I imagine we'd mark these components as "experimental" in the React documentation page, but their experimental status won't affect the top-level navigation.

lesliecdubs:
> @lesliecdubs Should we create an issue to track the implementation of this ADR?

@colebemis yes please! Let's add that issue to the "Inbox" on the Primer Roadmap (GitHub staff only) to be prioritized, ideally with a "size" label to give a rough estimate of complexity for the work.

lesliecdubs:
Pinging for an update on this one. @colebemis would you be willing to open an issue to track the implementation of this ADR [as discussed](https://github.com/primer/react/pull/2994#issuecomment-1509529150) and then we can try to get this merged?

siddharthkp:
Keeping this open for a few more days as we decide if we want to change our strategy for v36

lesliecdubs:
> Keeping this open for a few more days as we decide if we want to change our strategy for v36

Bumping @siddharthkp, did we have a chance to validate we're happy with this strategy with regards to v36?

siddharthkp:
> Bumping @siddharthkp, did we have a chance to validate we're happy with this strategy with regards to v36?

We decided not to mix this with v36 (it's already too complex) and hold implementation (and decision) until after v36!

lesliecdubs:
Copy that. If we don't come back around to this in Q3, it might be good to get it on the official docket to address in Q4.

lesliecdubs:
Not stale

dusave:
Has this conversation gone as far as it's going to? I don't feel we ever really came to a resolution on this.

siddharthkp:
Re-opened and converted to draft!

PR #3004, opened 2023-03-09
Support custom emojis & declarative `loading` state in `MarkdownEditor` suggestions

iansan5653:
1. Adds support for custom emojis (like :octocat:, which has no Unicode representation) by allowing emoji suggestions to take the shape of `{name: string, url: string}` where `url` is an image.
2. Adds support for a declarative loading state for Markdown editor suggestions by providing the string `"loading"` as the suggestions prop. We already supported asynchronous suggestions via an async function, but this imperative strategy isn't idiomatic React and is incompatible with `react-query`, which we'd like to use in Projects for caching data more effectively. My bad for originally writing it this way 😬. I'd really like to completely change it but I want to avoid breaking changes for now and this works just fine.

This change is completely non-breaking and won't require any additional QA if we can get it into this release (as evidenced by the Storybook stories which didn't require any changes).


src/drafts/MarkdownEditor/MarkdownEditor.test.tsx
```diff
@@ -866,13 +866,14 @@ describe('MarkdownEditor', () => {
...
-    const emojis: Emoji[] = [
+    const emojis = [
       {name: '+1', character: '👍'},
       {name: '-1', character: '👎'},
       {name: 'heart', character: '❤️'},
       {name: 'wave', character: '👋'},
       {name: 'raised_hands', character: '🙌'},
-    ]
+      {name: 'octocat', url: 'https://github.githubassets.com/images/icons/emoji/octocat.png'},
+    ] satisfies Emoji[]
```
mperrotti: Ohh I've read about `satisfies` but haven't actually used it yet. Nice.

mperrotti (approved):

joshblack (approved):

joshblack:
@iansan5653 sorry about the Jest issue! just opened up: https://github.com/primer/react/pull/3005 to avoid this in the future for TypeScript 4.9 features like `satisfies` 😎

PR #3007, opened 2023-03-09
refactor(DataTable): update initial sort state logic

joshblack:
Closes https://github.com/github/primer/issues/1854

This PR refactors the initial sort logic to both validate and test the different situations that we can find our selves in with `data`, `columns`, `initialSortColumn`, and `initialSortDirection`.

It also adds several validation warnings for these scenarios in order to make it easier to write and debug code related to sorting.

We are covering specifically the following scenarios with this work:

| Column `sortBy` | `initialSortColumn` | `initialSortDirection` | Outcome |
| :-------------- | :------------------ | :--------------------- | :------ |
| Set | Set to Column with `sortBy` | Not set | Sort by column in `ASC` order |
| Set | Set to Column with `sortBy` | Set to `ASC` | Sort by column in `ASC` order |
| Set | Not set | Set to `ASC` | Sort by first sortable column in `ASC` order |
| Any | Set to `Column` that does not exist | Not set | Unable to find column, warn about value |
| Not set | Set to `Column` without `sortBy` | Not set | No column is sortable, warn about no column with `sortBy` set |
| Not set | Not set | Set to `AC` | No column is sortable, warn about no column with `sortBy` set |

### Testing & Reviewing

- Double check the cases above to see if I missed any
- Make sure I covered all these cases with tests 😅 
- Double-check that no behavior has regressed in storybook demoes

### DIFF
--- a/src/DataTable/useTable.ts
+++ b/src/DataTable/useTable.ts
@@ -7,7 +7,7 @@ import {ObjectPathValue} from './utils'
 interface TableConfig<Data extends UniqueRow> {
   columns: Array<Column<Data>>
   data: Array<Data>
-  initialSortColumn?: string
+  initialSortColumn?: string | number
   initialSortDirection?: Exclude<SortDirection, 'NONE'>
 }
 
@@ -39,7 +39,7 @@ interface Cell<Data extends UniqueRow> {
   rowHeader: boolean
 }
 
-type ColumnSortState = {id: string; direction: Exclude<SortDirection, 'NONE'>} | null
+type ColumnSortState = {id: string | number; direction: Exclude<SortDirection, 'NONE'>} | null
 
 export function useTable<Data extends UniqueRow>({
   columns,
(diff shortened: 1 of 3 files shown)


src/DataTable/useTable.ts
```diff
@@ -207,6 +172,66 @@ export function useTable<Data extends UniqueRow>({
...
+  initialSortColumn?: string | number,
+  initialSortDirection?: Exclude<SortDirection, 'NONE'>,
+) {
+  if (initialSortColumn !== undefined) {
+    const column = columns.find(column => {
+      return column.id === initialSortColumn || column.field === initialSortColumn
+    })
+
+    if (column === undefined) {
+      if (__DEV__) {
```
colebemis, line 186: Is this a good place to use the `warning` utility you created?

colebemis (approved):
Logic makes sense ✅ 
Storybook stories still seem to work as expected ✅ 
Ship it 🚀

src/DataTable/useTable.ts
joshblack: @colebemis exactly 👍

PR #3012, opened 2023-03-10
Fix pagelayout stories axe violations

radglob:
Adds `tabIndex`, `role="region"` and requires either an `aria-label` or `aria-labelledby` on the `PageLayout.Pane` component when overflow is detected. Overflow is determined by checking if the `scrollHeight` of the pane is greater than the `clientHeight`.

Closes https://github.com/github/primer/issues/1865.

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


joshblack:
Thanks for putting this up @radglob! Just wanted to double-check on requirements for this, I believe that for this we would need the following:

- The `PageLayout.Pane` component needs to be aware of overflow
- When overflow is detected, the `PageLayout.Pane` component will:
  - Add the `tabIndex` attribute[^1]
  - Add `role="region"`[^1]
  - Add `aria-label` with an accessible label or `aria-labelledby` to an appropriate element[^1]

A big challenge for this component seems to be getting that `aria-label` or `aria-labelledby` to be a valid that's relevant since it only accepts `children`.

Is that you're understanding as well for this issue @ericwbailey @TylerJDev? Or is there anything missing?

Let me know if that makes sense @radglob or if I'm overthinking this 🤔 

[^1]: https://www.tpgi.com/short-note-on-improving-usability-of-scrollable-regions/)

TylerJDev:
@joshblack - Yup, that is my understanding as well. I'm guessing we'll need to make the accessible name a required prop for `pane`? If so, we might need to emphasize on why that's needed so that developers give it a good name.

@radglob - Let me know if you want to pair on this issue! Josh listed all the items we'd need to make sure this is accessible, but if you have any questions on the accessible bits let me know! :grin:

joshblack (changes requested):
Looks great! I think only big thing is to make sure that one story gets the `aria-label` to stop it from error'ing out. I left a couple other comments too, let me know what you think!

From a package release perspective, how do you feel about emitting a `warning` instead of an `error` for the missing `aria-label` or `aria-labelledby`? This could prompt people to add the label without potentially causing existing code to fail. Just wanted to get your thoughts on this 👀

joshblack (approved):
Looks great!

PR #3024, opened 2023-03-14
Add `Box` stories

simurai:

Add storybook stories for `Box`

**Checklist**

- [ ] ~~(Optional) Create visual regression tests for the component~~
    - I left it out since the `Box` component is only "visible" when used with `sx` props.


simurai:
> Create visual regression tests for the component

Is this beneficial to add? I left that out since the `Box` component is only used with `sx` props.

colebemis (approved):
🚀

PR #3027, opened 2023-03-14
Export useResizeObserver from @primer/react

talum:
Exporting the `useResizeObserver` hook from @primer/react so it can be used outside of Primer!

Closes https://github.com/github/primer/issues/1908

### Screenshots

n/a

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/nasty-hats-smell.md
+++ b/.changeset/nasty-hats-smell.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Export `useResizeObserver` hook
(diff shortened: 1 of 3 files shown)


joshblack:
Hey @talum! 👋 Thanks so much for taking the time to make a PR for this 🙏 

As a rule of thumb, when should we expect things be exported from Primer for hooks like this versus be packages in dotcom? (e.g.: https://gh.io/AAjq3a7) I wasn't sure and just wanted to double-check 👀

talum:
@joshblack Great question, and I don't know about the answer! I'm following guidance from this comment: https://github.com/github/primer/issues/1908#issuecomment-1467375326 

I would have thought it better to export this from Primer rather than implement a different version in dotcom. Interestingly, I don't even remember finding that in dotcom when I was searching. I definitely could use that hook in dotcom instead of exporting this one. What do y'all think?

colebemis (approved):
Thank you! ✨

joshblack:
@talum totally! Makes sense 👍

joshblack (approved):

PR #3029, opened 2023-03-14
More a11y improvements for Pagination

mperrotti:
- Instead of making the current page an `<em>` tag, keep it as a clickable link
- Change the tone the screenreader uses when reading the last number before we start skipping numbers. This was done by appending "..." to the aria-labels of those pages.

### Screenshots

There should be no visual changes

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Pagination/model.tsx
```diff
@@ -65,6 +65,12 @@ export function buildPaginationModel(
     for (let idx = 0; idx < sorted.length; idx++) {
       const num = sorted[idx]
       const selected = num === currentPage
+      const last = sorted[idx - 1]
+      const next = sorted[idx + 1]
+      const lastDelta = num - last
+      const nextDelta = num - next
+      const preceedsBreak = nextDelta !== -1
```
joshblack, line 72: Spelling nit:

```suggestion
      const precedesBreak = nextDelta !== -1
```

joshblack (approved):
Looks great! Definitely noticed the difference in pitch with NVDA 👍

src/Pagination/model.tsx
mperrotti: So embarrassng 🙈

PR #3030, opened 2023-03-14
feat(DataTable): add support for overflow behavior

joshblack:
Closes https://github.com/github/primer/issues/1835

Add support for overflow to `Table`. This will create a `role="region"` with a `tabindex` set so that when the table is scrollable someone may interact with it (reference axe rule: https://dequeuniversity.com/rules/axe/4.0/scrollable-region-focusable)

This also adds in an e2e test for this behavior along with screenshots for the various states

### Testing & Reviewing

- Visit the `With overflow` story in storybook
- Verify that the `<table>` element is wrapped in a `div` with `role="region"`
  - This region should be focusable
- Visit another story in the storybook
- Resize the viewport until overflow occurs
- Verify that the `<table>` element is wrapped in a `div` with `role="region"`, as above
- Resize the viewport until overflow no longer occurs
- Verify that the `<table>` element is no longer wrapped in a `div`


mperrotti:
I only want the table to scroll so the table's header and actions are always visible. I think it would also be a violation of [SC 1.4.10 (Reflow)](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html) if anything besides the table scrolls.

Let me know if you want to pair on this :)

joshblack:
@mperrotti makes sense! Do you know how to do this with a native `<table>` or would we need to add a wrapping node around it for the overflow?

mperrotti:
> Do you know how to do this with a native <table> or would we need to add a wrapping node around it for the overflow?

I'm not totally sure - we probably need to add a wrapping node.

joshblack:
@mperrotti sounds good! Just refactored and added a scrollable region wrapper under `Table`, let me know if this was moreso what you were looking for 👀

colebemis (approved):
Love it 💖

PR #3032, opened 2023-03-15
Tooltip: Update the component markup for a11y remediations & refactor the existing styles towards more static styling

broccolinisoup:
Closes https://github.com/github/primer/issues/1909
Closes https://github.com/github/primer/issues/1907
Closes https://github.com/github/primer/issues/1918

Storybook features: https://primer-7bad896c1f-13348165.drafts.github.io/storybook/?path=/story/components-tooltip-features--tooltip-description-type-tooltip 
Docs: https://primer-7bad896c1f-13348165.drafts.github.io/Tooltip

This PR refactors the current `Tooltip` component to update the markup and styles to match with the expected accessibility requirements. 

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/Tooltip2.tsx
```diff
@@ -0,0 +1,141 @@
+import React, {Children, useEffect, useRef} from 'react'
+import Box from './Box'
+import {BetterSystemStyleObject, merge, SxProp} from './sx'
+import {useId} from './hooks/useId'
+import {isFocusable} from '@primer/behaviors/utils'
```
broccolinisoup, line 5: @joshblack I found a little gem 🎀  This seems to be working as expected but let me know what you think!

src/Tooltip2.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+export type Tooltip2Props = {
+  direction?: 'n' | 'ne' | 'e' | 'se' | 's' | 'sw' | 'w' | 'nw'
+  text?: string
+  noDelay?: boolean
+  align?: 'left' | 'right'
+  wrap?: boolean
+  type?: 'label' | 'description'
```
broccolinisoup, line 13: `type` is a new prop that I added and defaulted to `label` to prevent braking the existing usage.

src/Tooltip2.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const tooltipRef = useRef<HTMLDivElement>(null)
+  const child = Children.only(children) // make sure there is only one child
+
+  // we need this check for every render
+  useEffect(() => {
+    if (tooltipRef.current) {
+      const childNode = tooltipRef.current.children[0] // For now, I assume it has one node but that is not true
+      if (!isFocusable(childNode as HTMLElement)) {
+        throw new Error('Tooltip2: The child element must be focusable')
+      }
```
broccolinisoup, line 111: @TylerJDev 
>Would it be possible to enhance this through TypeScript, or log a warning on dev to ensure the trigger element (child) is an interactive element?

I added a runtime check - not sure if that would be possible with the type check. I chat this with @joshblack as well. Let me know what you think about that!

TylerJDev: Oh wow, I love this! So basically we're checking the element that's passed to `Tooltip2`, and seeing if it's focusable with the `isFocusable` util? 

This is perfect, and I love how slick it is! Let's definitely run with this :grin:

broccolinisoup: >Oh wow, I love this! So basically we're checking the element that's passed to Tooltip2, and seeing if it's focusable with the isFocusable util?

Yes 🙌🏻

src/Tooltip2.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+export type TriggerPropsType = {
+  'aria-describedby'?: string
+  'aria-labelledby'?: string
+  'aria-label'?: string
+}
+
+const TOOLTIP_ARROW_EDGE_OFFSET = '16'
+
+const tooltipClasses = ({direction, noDelay, align, wrap}: Omit<Tooltip2Props, 'type' | 'text' | 'children'>) => ({
```
broccolinisoup, line 24: The styles are all copy paste from the current Tooltip

src/Tooltip2.tsx
```diff
@@ -0,0 +1,257 @@
...
+    'aria-labelledby': type === 'label' ? id : undefined,
+  }
+  // Only need tooltip role if the tooltip is a description for supplementary information
+  const role = type === 'description' ? 'tooltip' : undefined
+  // aria-hidden true only if the tooltip is a label type
+  const ariaHidden = type === 'label' ? true : undefined
+
+  return (
+    <>
+      <style>
```
broccolinisoup, line 232: I need feedback on this 👀 I couldn't find a way to declare animations without using `keyframes` from `styled-components` so I ended up doing this.

src/Tooltip2.tsx
joshblack: @broccolinisoup for `type` here, I bet we could have the default be `type = 'label'` as that seems to be the semantics from:

> Attention: we use aria-label for tooltip contents. However, aria-label replaces the text content of an element in screen readers, so only use Tooltip on elements with no existing text content.
> [Docs](https://primer.style/react/Tooltip)

Would this help with releasing this in a minor release? We could then make this required, if appropriate, in the next major.

Let me know if that makes sense! 

---

Side-note, the link for usage points to PVC. Is there a query link that you were sharing? Would love to check it out!

src/Tooltip2.tsx
```diff
@@ -0,0 +1,292 @@
...
+  const child = Children.only(children) // make sure there is only one child
+  const [open, setOpen] = useState(false)
+
+  // we need this check for every render
+  useEffect(() => {
+    if (tooltipRef.current) {
+      const childNode = tooltipRef.current.children[0] // For now, I assume it has one node but that is not true
+      if (!isFocusable(childNode as HTMLElement)) {
+        throw new Error('Tooltip2: The child element must be focusable')
+      }
```
joshblack, line 227: Couple of ideas for the guard block:

```suggestion
  if (__DEV__) {
    useEffect(() => {
      if (tooltipRef.current) {
         // For now, I assume it has one node but that is not true
        const childNode = tooltipRef.current.children[0]
        invariant(
          isFocusable(childNode as HTMLElement),
          'The `Tooltip2` component expects a single React element that contains interactive content. Consider using a `<button>` or equivalent interactive element instead.'
        );
      }
    });
  }
```

joshblack: Question: would it be possible to add a `ref` to the trigger element in `cloneElement`?

src/Tooltip2.tsx
```diff
@@ -0,0 +1,292 @@
...
+          },
+          // onMouseLeave: () => {
+          //   setOpen(false)
+          // },
+          onKeyDown: (e: React.KeyboardEvent) => {
+            if (open && e.key === 'Escape') {
+              e.stopPropagation()
+              setOpen(false)
+            }
+          },
```
joshblack, line 282: One idea, we should make sure to compose any handlers we had with what the `child` may have.

In other words, if someone writes:

```js
<Tooltip>
  <Button onKeyDown={() => { /* ... */ }} />
</Tooltip>
```

We should make sure not to override their `onKeyDown`

src/stories/Tooltip.stories.tsx
```diff
@@ -22,10 +23,179 @@ export default {
...
       <IconButton icon={SearchIcon} aria-label="Search" />
     </Tooltip>
   </Box>
 )
+
+// As a label for an IconButton
+export const Tooltip2LabelTypeTooltip = () => (
+  <Box sx={{p: 5}}>
+    <Tooltip2>
+      <IconButton icon={SearchIcon} aria-label="Search" />
```
joshblack, line 38: It'd be great if we could move the `aria-label` up a level and not support inferring it from the child. Is this something that currently exists in the component that we're needing to support still?

src/Tooltip2.tsx
```diff
@@ -0,0 +1,292 @@
...
+    content: '""',
+    width: 0,
+    height: 0,
+    color: 'neutral.onEmphasis',
+    pointerEvents: 'none',
+    borderStyle: 'solid',
+    borderWidth: '6px',
+    borderColor: 'transparent',
+    opacity: 0,
+    ...(open && {
```
joshblack, line 52: As a style refactor, what would you think about using data attributes for some of these states? This could avoid regenerating the objects for `styled-components` when props change as we could represent it like this statically:

```tsx
const StyledTooltip = styled.div`
  position: relative;
  display: inline-block;

  /* ... */

  [data-direction="n"], [data-direction="ne"], [data-direction="nw"] {
    /* ... */
  }
`
```

So that props will update `data-*` attributes but the styles won't have to be rebuilt each time.

src/stories/Tooltip.stories.tsx
broccolinisoup: I responded [here](https://github.com/primer/react/pull/3032#discussion_r1142806481) 🙂 Thanks again for pointing this out! 

Question - For this type of usage

```
<Tooltip2 aria-label="Search tooltip">
    <IconButton icon={SearchIcon} aria-label="Search button itself" />
</Tooltip2>
```

With the current remediations, the button will end up having both `aria-label` and `aria-labelledby` attribute and I have [this line of code](https://github.com/primer/react/pull/3032/files#diff-3516f4ecd704760b75cc577257c4c3e36f6830d590bdf4c37364d4edde78547dR234) to remove the `aria-label` from the trigger child as it will have `aria-labelledby` but I don't know this is a good practise? 

I tested it on VoiceOver and it seems like when both `aria-label` and `aria-labelledby` are present, it will disregard the `aria-label`. So I am a bit not sure what would be the best practise here 🤔

src/Tooltip2.tsx
broccolinisoup: oh true! 🙈 I'll do that

src/Tooltip2.tsx
broccolinisoup: >Couple of ideas for the guard block

Oh I love that 💖  I forgot we have this utils now - thank you! (I applied suggestion in the local to avoid linting issues)

>Question: would it be possible to add a ref to the trigger element in cloneElement?

I guess we could, where would we use it?

src/Tooltip2.tsx
broccolinisoup: I was trying to remove the styled components completely considering that we are looking into moving away from it but without having an alternative, it doesn't make much sense. I found a couple of possible ways when I searched on google but it doesn't worth the effort, I don't think. I'll revert this back to use `styled-components` for now.

src/Tooltip2.tsx
broccolinisoup: @joshblack I refactor the styles to use data attributes for `directions` prop and `open` state. I wasn't sure about `wrap` and `align` props - would that be too many data-attributes if I so the same thing for them? 🤔 cc @langermank would love your opinion on that too ☺️

src/stories/Tooltip.stories.tsx
```diff
@@ -22,10 +23,185 @@ export default {
...
+export const Tooltip2LabelTypeTooltip = () => (
+  <Box sx={{p: 5}}>
+    <Tooltip2 aria-label="Search tooltip">
+      <IconButton icon={SearchIcon} aria-label="Search button itself" />
+    </Tooltip2>
+  </Box>
+)
+
+// As a label for an IconButton
+export const Tooltip2CustomIcon = () => (
```
langermank, line 44: For some reason this is getting a little cut off. I'm not sure if you worked on viewport boundaries at all, but I would assume it should always reposition itself to not break the viewport

image

src/Tooltip2.tsx
langermank: Hmm good question.. it might be nice to use data att for everything for the sake of consistency, and I wouldn't worry about having too many in there 😄

broccolinisoup: Sounds good to me! 😄 I updated the code!

src/stories/Tooltip.stories.tsx
broccolinisoup: As far as I understand, this would be achieved with the popover api ([issue](https://github.com/primer/react/issues/2124) for ref) and unfortunately this tooltip is still uses `z-index`. This PR is mainly remediating the existing Tooltip a11y issues and I thought it would be a good opportunity to update the styles as well for saving us time in the near future when migrating them over to .css

broccolinisoup: @langermank I will see how much effort needs for this to make it with the current structure. If that is okay, I'd like to tackle this separately, independent from this PR.

broccolinisoup:
ping @joshblack - once you get a chance, could I please get another review? I believe all feedback and remaining remediations are addressed. Thank you 🙌🏻

broccolinisoup:
@TylerJDev 👋🏻 I believe I finalised all of the remediations. Could I please get your review when you get a chance please? 🙇🏻‍♀️ 
Here is the [storybook link](https://primer-7bad896c1f-13348165.drafts.github.io/storybook/?path=/story/components-tooltip-default--tooltip-2-custom-icon) to see different type of tooltips and styles. Looking forward to hearing your thoughts 🙌🏻

src/Tooltip2.tsx
```diff
@@ -0,0 +1,402 @@
...
+    >
+      {React.cloneElement(child as React.ReactElement<TriggerPropsType>, {
+        ...{
+          // if it is a type description, we use tooltip to describe the trigger
+          'aria-describedby': type === 'description' ? id : undefined,
+          // If it is a type description, we should keep the aria label if it exists, otherwise we remove it because we will use aria-labelledby
+          'aria-label': type === 'description' ? (children as React.ReactElement).props['aria-label'] : undefined,
+          //   If it is a label type, we use tooltip to label the trigger
+          'aria-labelledby': type === 'label' ? id : undefined,
+        },
```
joshblack, line 379: I think we should be able to flatten this one level:

```suggestion
        // if it is a type description, we use tooltip to describe the trigger
        'aria-describedby': type === 'description' ? id : undefined,
        // If it is a type description, we should keep the aria label if it exists, otherwise we remove it because we will use aria-labelledby
        'aria-label': type === 'description' ? (children as React.ReactElement).props['aria-label'] : undefined,
        //   If it is a label type, we use tooltip to label the trigger
        'aria-labelledby': type === 'label' ? id : undefined,
```

src/Tooltip2.tsx
```diff
@@ -0,0 +1,402 @@
...
+        data-wrap={wrap}
+        data-delay={noDelay}
+        {...rest}
+        // Only need tooltip role if the tooltip is a description for supplementary information
+        role={type === 'description' ? 'tooltip' : undefined}
+        // stop AT from announcing the tooltip twice when it is a label type because it will be announced with "aria-labelledby"
+        aria-hidden={type === 'label' ? true : undefined}
+        id={id}
+      >
+        {text || label}
```
joshblack, line 396: This is such a boring suggestion but great usage for `??` here 😂 

```suggestion
        {text ?? label}
```

joshblack (changes requested):
The style refactor with the data attributes is _amazing_. What a great change 🔥 I think this looks fantastic, went over it with the checklist below and just a had a couple of questions/comments on the PR!

**Checklist**

- [x] Visible on hover
- [x] Visible on focus
- [x] Dismissible with <kbd>Esc</kbd>
- [x] Supports delay
- [x] Supports label and description
- [x] Browsers
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] Edge

**Questions**

- Should the text of a tooltip be user selectable?
- Should `aria-hidden` be true when it is a description so that the tooltip is not navigable by a virtual cursor?

src/Tooltip2.tsx
```diff
@@ -0,0 +1,402 @@
...
+        data-direction={direction}
+        data-state={open ? 'open' : undefined}
+        data-align={align}
+        data-wrap={wrap}
+        data-delay={noDelay}
+        {...rest}
+        // Only need tooltip role if the tooltip is a description for supplementary information
+        role={type === 'description' ? 'tooltip' : undefined}
+        // stop AT from announcing the tooltip twice when it is a label type because it will be announced with "aria-labelledby"
+        aria-hidden={type === 'label' ? true : undefined}
```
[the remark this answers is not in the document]
TylerJDev: This is a good question! When I was looking into this, there seems to be mixed opinions on using `aria-hidden` when the tooltip is used as a description. I mainly defaulted to not having `aria-hidden` on the tooltip when it's used as a description based on an [internal conversation](https://github.slack.com/archives/C0FSWLQ0Y/p1659132347317399) around it. As there might be an intention for users to navigate to it, but only when it is type `description`.

For `role="tooltip"` not being applied for the `label` type, it seems that when the tooltip is the accName, then it isn't technically a tooltip anymore. With the added `aria-hidden` for `label` types, the user most likely wouldn't know it was a tooltip too. I'm mainly using this article around tooltips as reference https://sarahmhigley.com/writing/tooltips-in-wcag-21/. 

The support for the role is definitely lacking :sob:.

TylerJDev (commented):
This looks great @broccolinisoup! Definitely a huge improvement over our previous tooltip 🥳. 

I only have 2 comments, that I want to know your opinion of first. 

* I was wondering the feasibility around dismissing the `Tooltip` on hover? For example, if the user did not trigger the `Tooltip` with focus, but did with hover - would it be possible to to implement functionality to close the `Tooltip` when they press the "ESC" key?
  * We have similar functionality in the PVC `Tooltip`, but I know we handle events differently in PRC, and the potential need to handle this at a global level might not be as wanted here. Curious to know your thoughts about this, as I don't want to add too much additional work.
* Could we add something for the `Tooltip` in High Contrast Mode (HCM)? I noticed in the review that our outline of the `Tooltip` is missing when in this mode. Could we add something like the following to keep the outline? 

Example: 
```
@media (forced-colors: active) {
  outline: 1px solid transparent;
}
```

The above should give us an outline in HCM. Here's a way to [emulate HCM on Chrome](https://developer.chrome.com/docs/devtools/rendering/emulate-css/#emulate-css-media-feature-forced-colors) too, to see what it could look like, without using Windows.

Thanks so much for your work on this, I can't wait to have this replace our current `Tooltip`! :tada:

src/Tooltip2.tsx
broccolinisoup: Ahh good catch!

src/Tooltip2.tsx
```diff
@@ -0,0 +1,402 @@
...
+        triggerEvtHandlers.onBlur()
+      },
+      onFocus: () => {
+        _onFocus && _onFocus()
+        triggerEvtHandlers.onFocus()
+      },
+      onMouseEnter: () => {
+        _onMouseEnter && _onMouseEnter()
+        triggerEvtHandlers.onMouseEnter()
+      },
```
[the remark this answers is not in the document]
broccolinisoup: This is a great point! I guess I have a bit of a habit to take logic into functions most of the time; but in this case there are few jumps, you are right. It makes it hard to follow through and doesn't bring much value into readability when I pull out the composing logic into a function. I really love your suggestion. Thanks 🙌🏻

src/Tooltip2.tsx
```diff
@@ -0,0 +1,402 @@
...
+    }
+  }
+
+  function onKeyDown(e: React.KeyboardEvent) {
+    if (open && e.key === 'Escape') {
+      e.stopPropagation()
```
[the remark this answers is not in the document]
broccolinisoup: Ooo that is a good point but I don't know the answer! Do you have any cases/examples where we need to preventDefault you think?

src/Tooltip2.tsx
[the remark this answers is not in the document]
broccolinisoup: Thanks @TylerJDev for your reply! Thanks for providing the [internal conversation](https://github.slack.com/archives/C0FSWLQ0Y/p1659132347317399) link as well, it was very helpful! 

@joshblack Do you have any concern/question with any of it? Thank you so much for asking this great question, it broaden my understanding on the tooltip a11y guidelines 💖

broccolinisoup:
@TylerJDev Thank _you_ for the great review!

Re, closing the tooltip with `ESC` when mouse is over the trigger element, sorry I missed that in this issue! I am not so sure how to tackle this as you mentioned we need a global event listener and I am not sure if that would be possible to do that through the component 🤔 I imagine that it would the responsibility of the application that uses `Tooltip` but I might be wrong here! I'll check with the PVC team to see how they tackle this. Thanks for catching that! 

Re, the HCM, sure! I'll implement that.

src/Tooltip2.tsx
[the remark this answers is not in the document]
broccolinisoup: Found [extra context](https://github.com/primer/view_components/pull/1307) for this

src/Tooltip2.tsx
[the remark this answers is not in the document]
joshblack: I bet `stopPropagation()` should work for the dialog case so should be good to go and we can always address it later with a use-case if one comes up!

TylerJDev:
@broccolinisoup - I definitely understand what you mean around handling global events. I noticed one of our custom hooks, and I was wondering if `useOnEscapePress` could potentially be used here? I'm not too sure the specifics around the hook, but I'm wondering if this _might_ be applicable for this scenario?

Another idea is for the hook to be imported by the developer themselves to add the "escape" functionality, as this would mean it wouldn't be the components responsibility. 

Other than that, the component is looking great! I think pending this discussion, we should be good to go! Thanks again for all your work on this!

src/Tooltip2.tsx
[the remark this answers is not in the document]
joshblack: Having a hard time following the internal link 😅 Is there a specific message link for this guidance?

I think something that is odd without `aria-hidden` is that the description is announced twice, once when you focus the element and then again when moving the virtual cursor focus and it goes to the description. That seems like something that would not be expected in this kind of context for a tooltip.

src/stories/Tooltip.stories.tsx
joshblack: @broccolinisoup the `IconButton` is definitely a tough situation. Ideally, I think `IconButton` should render a `Tooltip` over the underlying `Button` so that this behavior is built-in by default. In other words, one would write:

```tsx
<IconButton icon={SearchIcon} aria-label="Search button itself" />
```

And this would include the tooltip by default.

I personally worry about `child.props` since someone could add a wrapper component in between and still retain the incorrect usage 😞 For example,

```jsx
<Tooltip2 aria-label="Search tooltip">
  <CustomIconButton />
</Tooltip2>

function CustomIconButton() {
  return <IconButton icon={SearchIcon} aria-label="Search button itself" />
}
```

I think if we can require the `child` to accept a `ref`, then we should be able to warn for either usage (basically is any child labelled) using that `ref`. e.g.

```tsx
function Tooltip({ children }) {
  const child = React.Children.only(children);
  const ref = React.useRef(null);

  // ...

  if (__DEV__) {
    useEffect(() => {
      // Check contents of `ref` for label usage
    });
  }

  // ...
  {React.cloneElement(child, { ref, /* ... */ })}
  // ...
}
```

broccolinisoup: 🔥 I love this so much. This is a great point!

broccolinisoup:
Note that the snapshots are failing on main. It should be good once we resolve the issue on main

TylerJDev (approved):
Amazing work @broccolinisoup! 🤩

This tooltip will be a huge improvement over our current one in terms of accessibility. As far as my comments, you've addressed all of them! Thanks so much for all the work on this! 🎉

src/Tooltip2.tsx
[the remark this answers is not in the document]
joshblack: @TylerJDev sounds good! Thanks so much for taking the time to explain, I appreciate it 🙏

generated/components.json
```diff
@@ -4652,6 +4614,59 @@
...
+          "name": "align",
+          "type": "'left' | 'right'"
+        },
+        {
+          "name": "aria-label",
+          "type": "string",
+          "description": "Should be utilised for label type tooltips and it is going to be used to label the tooltip trigger."
+        },
+        {
+          "name": "direction",
```
edersonlucas, line 4639: What do you think about adding the default value `"defaultValue": "n"` to the component documentation to make the default direction explicit?

generated/components.json
```diff
@@ -4652,6 +4614,59 @@
...
+          "description": "When set to `true`, tooltip appears without any delay."
+        },
+        {
+          "name": "text",
+          "type": "string",
+          "description": "Should be utilised for description type tooltips and it is going to be used to describe the tooltip trigger."
+        },
+        {
+          "name": "type",
+          "type": "'label' | 'description'",
```
edersonlucas, line 4655: Same comment above. Add `"defaultValue": "label"`.

edersonlucas (changes requested):

src/Tooltip/Tooltip.docs.json
```diff
@@ -9,6 +9,11 @@
...
       "type": "'left' | 'right'"
     },
+    {
+      "name": "aria-label",
+      "type": "string",
+      "description": "Should be utilised for label type tooltips and it is going to be used to label the tooltip trigger."
+    },
     {
       "name": "direction",
       "type": "'n' | 'ne' | 'e' | 'se' | 's' | 'sw' | 'w' | 'nw'",
```
edersonlucas, line 19: What do you think about adding the default value `"defaultValue": "n"` to the component documentation to make the default direction explicit?

src/Tooltip/Tooltip.docs.json
```diff
@@ -20,9 +25,14 @@
...
     {
-      "name": "aria-label",
+      "name": "text",
       "type": "string",
-      "description": "Text used in `aria-label` (for accessibility)"
+      "description": "Should be utilised for description type tooltips and it is going to be used to describe the tooltip trigger."
+    },
+    {
+      "name": "type",
+      "type": "'label' | 'description'",
```
edersonlucas, line 34: Same comment above. Add `"defaultValue": "label"`.

src/Tooltip/Tooltip.docs.json
broccolinisoup: Ah great catch! Thank you 🙇🏻‍♀️

PR #3036, opened 2023-03-15
Use SSR-compatible slot implementation in PageLayout

colebemis:
## Problem

`PageLayout.Header` and `PageLayout.Footer` don't render server-side because PageLayout uses an SSR-incompatible approach to implement its slots API ([issue](https://github.com/primer/react/issues/1690)). 

https://user-images.githubusercontent.com/4608155/225428127-024c35b5-f01a-4a6e-a256-aeabf3dc714f.mov

Notice that the layout shifts because the header appears after the page loads.

[Slack discussion for context](https://github.slack.com/archives/C01L618AEP9/p1678833964361389)

## Solution

This PR updates PageLayout to use an SSR-compatible slots implementation.

## Tradeoff

In this new implementation, `PageLayout.Header` and `PageLayout.Footer` must be direct children of `PageLayout`. This feels like a reasonable tradeoff because `PageLayout.Header` and `PageLayout.Footer` aren't used often, and when they are used, they're almost always a direct child of `PageLayout`.

- Dotcom uses `PageLayout.Header` in 3 places. In 2 of 3 usages, `PageLayout.Header` is a direct child of `PageLayout`. In the 3rd case, `PageLayout.Header` can easily be refactored to be direct child of `PageLayout`.
- Dotcom doesn't use `PageLayout.Footer`.

## Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/hooks/unstable_useSlots.ts
```diff
@@ -0,0 +1,65 @@
...
+  [Property in keyof Type]: React.ReactElement
+}
+
+/**
+ * Extract components from `children` so we can render them in different places,
+ * allowing us to implement components with SSR-compatible slot APIs.
+ * Note: We can only extract direct children, not nested ones.
+ */
+// eslint-disable-next-line camelcase
+export function unstable_useSlots<T extends SlotConfig>(
```
joshblack, line 15: Since this is internal-only, how would you feel about calling this `useSlots` instead of `unstable_useSlots`?

```suggestion
export function useSlots<T extends SlotConfig>(
```

We could then save `unstable_*` for exported members that are exempted from semver for the `npm` package.

src/hooks/unstable_useSlots.ts
```diff
@@ -0,0 +1,65 @@
...
+  children: React.ReactNode,
+  config: T,
+): [Partial<SlotElements<T>>, React.ReactNode[]] {
+  // Object mapping slot names to their elements
+  const slots: Partial<SlotElements<T>> = mapValues(config, () => undefined)
```
joshblack, line 20: Since this is a hot path in render, do you think it would help to reduce the object transforms in `mapValues` and `getKeyByValue`? Like for keys, reduce, find, etc. Or would the different not really matter, ultimately?

One quick idea was to generate arrays for keys and value of `config` and use them to index each other:

```tsx
const slots = {};
const keys = Object.keys(config);
const values = Object.values(config);
const rest = [];

React.Children.forEach(children, child => {
  if (!React.isValidElement(child)) {
    rest.push(child);
    return;
  }

  const index = values.findIndex(value => {
    return child.type === value;
  });

  if (index === -1) {
    rest.push(child);
    return;
  }

  const slotKey = keys[index];
  if (slots[slotKey]) {
    return;
  }

  slots[slotKey] = child;
});
```

Another idea could be to use `Object.entries()` along with `find()` if that would read better.

src/hooks/unstable_useSlots.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (!React.isValidElement(child)) {
+      rest.push(child)
+      return
+    }
+
+    // Check if the child is a slot
+    const slotKey = getKeyByValue(config, child.type)
+
+    if (slotKey && slots[slotKey]) {
+      // If slot is already filled, ignore duplicates
```
joshblack, line 36: Would it be helpful here to emit a warning about a duplicate slot or would that just add noise to the console?

joshblack (approved):
Looks great 🔥 Just left some comments for some ideas while going through the code, feel free to ignore if they're not helpful 👍

src/hooks/unstable_useSlots.ts
colebemis: Done 👍

src/hooks/unstable_useSlots.ts
colebemis: Love it! This is wayy faster:

CleanShot 2023-03-15 at 18 24 57@2x

src/hooks/unstable_useSlots.ts
colebemis: Great idea! Done ✅

siddharthkp:
Hi! 

Just leaving it here in case we need it: In case this change isn't backward compatible with instances of PageLayout in gh/gh, we can unblock this team by putting this in drafts/experimental first.

colebemis:
@prashkan verified that this fixes SSR issues with PageLayout in gh/gh ✅ [Slack](https://github.slack.com/archives/C01L618AEP9/p1678984426954829?thread_ts=1678833964.361389&cid=C01L618AEP9)

PR #3040, opened 2023-03-15
Revert `Dialog` and `Dialog(draft)` folder refactor that causes import issues

broccolinisoup:
`Dialog` and `Dialog(draft)` causes import issues - so reverting back  for now to unblock folks using canary builds - slack ref: https://github.slack.com/archives/C01L618AEP9/p1678918940134579?thread_ts=1678833964.361389&cid=C01L618AEP9


colebemis (approved):
Thanks for this quick fix ⚡

PR #3046, opened 2023-03-16
docs(adr): add internal modules ADR

joshblack:
Adds an ADR for managing Internal Modules in Primer React.

[Rendered Markdown](https://github.com/primer/react/blob/fb15555b1ab174a2445b128fbb8999bf54794837/contributor-docs/adrs/adr-014-internal-modules.md)

### DIFF
--- a/contributor-docs/adrs/adr-014-internal-modules.md
+++ b/contributor-docs/adrs/adr-014-internal-modules.md
@@ -1,0 +1,62 @@
+# ADR 014: Internal Modules
+
+## Status
+
+Proposed
+
+## Context
+
+Currently all files live under the `src` directory. In the `npm` package for `@primer/react`, we specify the following export pattern:
+
+```json5
+{
+  "exports": {
+    // ...
+    "./lib-esm/*": {
+      "import": [
+        // ...
+      ],
+      "require": [
+        // ...
+      ]
+    }
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


joshblack:
cc @primer/react-reviewers would love to get your opinion when you have a sec! 👀 The tl;dr is that this is a convention for where to put modules within Primer React that we don't intent to include as a part of the public API of the package.

contributor-docs/adrs/adr-014-internal-modules.md
```diff
@@ -0,0 +1,62 @@
...
+    "./feature/internal/*": null
+  }
+}
+```
+
+
+### Impact
+
+- Update the `"exports"` field in `package.json` to exclude `./lib-esm/internal` from usage
+- (In a major release) Move internal-only files into internal folder
```
broccolinisoup, line 61: How would that affect the usages of these internal files? It makes total sense to do in a major release but I am wondering would we provide any alternatives or make exceptions to exclude some files if they are in used but intended to be internal? I hope it makes sense!

joshblack: Hey @broccolinisoup! 👋 

Great point, I'm glad you brought it up. If something is meant to be internal but it has usage in dotcom then we should definitely continue to export it from the bundle, in my opinion. I think at that point it stops being internal and is something that teams have come to rely on (albeit accidentally) and clearly has value for them and for us to support.

Curious what you think 👀

broccolinisoup: Yeah, I agree!

colebemis (approved):
Seems reasonable 👍

PR #3047, opened 2023-03-17
Add `ProgressBar` stories

simurai:

Add storybook stories for `ProgressBar`

**Checklist**


src/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export default {
+  title: 'Components/ProgressBar',
+  component: ProgressBar,
+} as Meta<typeof ProgressBar>
+
+export const Default = () => <ProgressBar />
+
+export const Playground: ComponentStory<typeof ProgressBar> = args => (
+  <ProgressBar {...args} sx={args.inline ? {width: '100px'} : {}} />
```
simurai, line 13: When using `inline` [the docs mention](https://primer.style/react/ProgressBar#inline) to also set a `width`. Which I tried adding here. It works, but having the `width` set automatically could suggest that it's part of the component and people might get confused when not also reading the docs.

Is this a big concern? A better approach? Or should we remove the `inline` option again from the Playground to avoid confusion?

simurai:
@joshblack Sorry for pinging you again 😇, but I can't get the e2e tests to pass. E.g. `vrt/upload-snapshots` [seems to time out](https://github.com/primer/react/actions/runs/4486872950/jobs/7889752020?pr=3047#step:8:2338)? 🤔

simurai:
@joshblack Thanks, again. 🙇

🤦  Ouch, I see https://github.com/primer/react/pull/3047/commits/a39cd0aafd5b38223e0813bfacc95b982b878264.

joshblack (approved):

src/ProgressBar/ProgressBar.stories.tsx
simurai: Ok, I'll merge it as is. In the future we might could add a `min-width` so that manually setting a `width` is not necessarily needed.

PR #3055, opened 2023-03-20
feat(DataTable): add Table.Skeleton

joshblack:
Batch: https://github.com/github/primer/issues/1892

This is currently a visual an API implementation of `Table.Skeleton`. The behavior is not finalized when it comes to how to indicate loading content. In the meantime, the cells themselves contain "Loading" as text content. Most of the styles for the Table cell skeleton items are lifted from TreeView.

One caveat to this API, it is currently bound to `Column` as it seemed natural to toggle between the two with an existing set of `columns`. It may be useful in the future to permit less restrictive types (like `Array<string>`) for usage.

### Testing & Reviewing

- Verify that the "With Loading" example is available on the storybook and included in the docs.json for DataTable
- Verify that the styles look as expected for the Table Skeleton
- Verify that `Table.Skeleton` is defined and its props are exported
- Verify that a test exists for rendering a `Table.Skeleton`
  - Note: I omitted a test for `rows` since it's a visual only change and I did not want to test structural changes in enzyme


colebemis (approved):
Looks great! 🚀 

- [x] Verify that the "With Loading" example is available on the storybook and included in the docs.json for DataTable
- [x] Verify that the styles look as expected for the Table Skeleton
- [x] Verify that Table.Skeleton is defined and its props are exported
- [x] Verify that a test exists for rendering a Table.Skeleton

joshblack:
bump @mperrotti when you have a sec to review!

src/DataTable/Table.tsx
```diff
@@ -141,6 +152,66 @@ const StyledTable = styled.table<React.ComponentPropsWithoutRef<'table'>>`
...
+  .TableCellSkeleton {
+    padding: 0;
+  }
+
+  .TableCellSkeletonItems {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .TableCellSkeletonItem {
```
mperrotti, line 165: If we see this style of "loading" placeholder come up again, we should consider creating some kind of "SkeletonLoader" component that can be reused. We already use something identical in TreeView.

mperrotti (approved):
I'm so sorry I didn't review this sooner! I have no idea how it slipped by me.

PR #3056, opened 2023-03-21
Remove incorrect focus styles on ActionList items

langermank:
I don't think the first item in `ActionMenu` should be focused at all by default, but this at least fixes the design bug with it. Focus styles utilize the accent blue outline, while the background is reserved for hover and active states.

Closes # https://github.com/github/primer/issues/1455

### Screenshots

Before:
image

After:
image

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -92,13 +92,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
           backgroundColor: `actionListItem.${variant}.hoverBg`,
           color: getVariantStyles(variant, disabled).hoverColor,
         },
-        ':focus:not([data-focus-visible-added]), > a:focus:not([data-focus-visible-added])': {
-          backgroundColor: `actionListItem.${variant}.selectedBg`,
-          color: getVariantStyles(variant, disabled).hoverColor,
-          outline: 'none',
-        },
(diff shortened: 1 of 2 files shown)


broccolinisoup (commented):
This change makes sense to me! I am thinking if there would be any cases that cause an issue because we don't have an visual indicator for the focus state for mouse users? 🤔 
I can't think of anything though but just to double check.

langermank:
> This change makes sense to me! I am thinking if there would be any cases that cause an issue because we don't have an visual indicator for the focus state for mouse users? 🤔 I can't think of anything though but just to double check.

Great question! That is not a requirement to my knowledge. That's why the `focus-visible` property is nice, as it allows us to only show a focus state if a keyboard is being used to navigate.

broccolinisoup:
> > This change makes sense to me! I am thinking if there would be any cases that cause an issue because we don't have an visual indicator for the focus state for mouse users? 🤔 I can't think of anything though but just to double check.
> 
> Great question! That is not a requirement to my knowledge. That's why the `focus-visible` property is nice, as it allows us to only show a focus state if a keyboard is being used to navigate.

Sounds good to me! 🚀

broccolinisoup (approved):

PR #3058, opened 2023-03-21
test(e2e): remove DataTable snapshots

joshblack:
Didn't realize this before committing 😅 The stories for `DataTable` currently use a _relative time_. As a result, each day the snapshots will be invalid. Removing these for now until the stories are refactored to use constant dates 👍


langermank (approved):
Sorry to laugh but this is hilarious 😂

PR #3065, opened 2023-03-22
Spike: Button with css

siddharthkp:
Going to test this with dotcom

Bundled css as into one file, needs to be imported explicitly

```js
import {Button} from '@primer/react/lib-esm/drafts/Button2'
import '@primer/react/lib-esm/components.css'
```


.storybook/preview.js
```diff
@@ -2,6 +2,9 @@ import {addons} from '@storybook/addons'
 import {withThemeProvider, withSurroundingElements, toolbarTypes} from '../src/utils/story-helpers'
 import {PrimerBreakpoints} from '../src/utils/layout'
 
+import '@primer/css/dist/primitives.css'
+import '@primer/css/dist/color-modes.css'
```
siddharthkp, line 6: note: global dependency on primitives!

generated/components.json
```diff
@@ -1580,16 +1580,6 @@
           "type": "| 'small'\n| 'medium'\n| 'large'",
           "defaultValue": "'medium'"
         },
-        {
```
siddharthkp, line 1583: unintentional change!

patches/rollup-plugin-lib-style+1.2.10.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/rollup-plugin-lib-style/lib/index.js b/node_modules/rollup-plugin-lib-style/lib/index.js
+index 6a3823c..a055bb6 100644
+--- a/node_modules/rollup-plugin-lib-style/lib/index.js
++++ b/node_modules/rollup-plugin-lib-style/lib/index.js
+@@ -96,7 +96,7 @@ const defaultLoaders = [
+   },
+ ];
+ 
+-const replaceMagicPath = (fileContent) => fileContent.replace(MAGIC_PATH_REGEX, ".");
++const replaceMagicPath = (fileContent) => fileContent.replace(MAGIC_PATH_REGEX, "..");
```
siddharthkp, line 10: realised the plugin is a bit buggy when nested, so patching the package temporarily. 

will replace this with a fork of the plugin

src/Button2/Button.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import styled from 'styled-components'
+import sx, {SxProp} from '../sx'
+import {ForwardRefComponent as PolymorphicForwardRefComponent} from '../utils/polymorphic'
+import classNames from './Button.module.css'
```
siddharthkp, line 5: import classNames from css

package.json
```diff
@@ -104,7 +104,7 @@
     "@types/styled-system": "^5.1.12",
     "@types/styled-system__css": "^5.0.16",
     "@types/styled-system__theme-get": "^5.0.1",
-    "classnames": "^2.3.1",
+    "classnames": "2.3.2",
```
siddharthkp, line 107: note for self: booo, should not be locked

package.json
```diff
@@ -128,6 +128,7 @@
     "@github/markdownlint-github": "^0.3.0",
     "@github/prettier-config": "0.0.6",
     "@playwright/test": "1.32.0",
+    "@primer/css": "20.8.3",
```
siddharthkp, line 131: note for self: booo, should not be locked

src/Button/Button.stories.tsx
```diff
@@ -57,6 +57,14 @@ export default {
   },
 } as Meta<typeof Button>
 
-export const Playground: Story<typeof Button> = args => <Button {...args}>Default</Button>
+export const Playground: Story<typeof Button & {trailingVisualCount: number}> = args => {
```
siddharthkp, line 60: Fixed the playground to support trailing count attribute in playground

src/CounterLabel/CounterLabel.tsx
```diff
@@ -5,6 +5,7 @@ import VisuallyHidden from '../_VisuallyHidden'
 
 export type CounterLabelProps = {
   scheme?: 'primary' | 'secondary'
+  className?: string
```
siddharthkp, line 8: note for self: check if this should be in the current PR

tsconfig.json
```diff
@@ -14,7 +14,16 @@
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
-    "typeRoots": ["./node_modules/@types", "./@types"]
+    "typeRoots": ["./node_modules/@types", "./@types"],
+    "plugins": [{"name": "typescript-plugin-css-modules"}]
   },
-  "include": ["@types/**/*.ts", "src/**/*.js", "src/**/*.ts", "src/**/*.tsx", "script/**/*.ts", "e2e/**/*.ts", "playwright.config.ts"]
+  "include": [
```
siddharthkp, line 20: note for self: booo, unrelated change?

src/CounterLabel/CounterLabel.tsx
siddharthkp: checked, it is

src/drafts/Button2/Button.module.css
```diff
@@ -0,0 +1,203 @@
+.button {
```
langermank, line 1: Is there any particular reason to rewrite this CSS instead of using what we already have from PVC? https://github.com/primer/view_components/blob/main/app/components/primer/beta/button.pcss

I think its harder to read this file with the custom CSS vars and overrides. I also think we should continue to use BEM for this test, and then if we move forward with CSS Modules write an ADR about how we want to name classes.

siddharthkp:
This PR is busted 😢, going to nuke it and start over from `main` 

@langermank, saw your comment, will copy it on the new PR!

PR #3068, opened 2023-03-23
Add horizontal alignment to column configuration

mperrotti:
Closes https://github.com/github/primer/issues/1959

### Screenshots

Kapture 2023-03-23 at 10 25 54

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/DataTable/Table.tsx
```diff
@@ -55,9 +56,22 @@ const StyledTable = styled.table<React.ComponentPropsWithoutRef<'table'>>`
...
   }
 
+  .TableHeader[data-cell-align='end'],
+  .TableCell[data-cell-align='end'] {
+    text-align: end;
+    display: flex;
+    justify-content: flex-end;
+  }
+
+  .TableHeader[data-cell-align='end'] > button {
```
joshblack, line 70: Could we add a class to the table sort button instead of using an element selector?

joshblack (approved):
Looks great! Just left a comment for the CSS Selector.

Also wanted to ask if we wanted a shared `Alignment` or `BlockAlignment` type after seeing the `ToggleSwitch` update. I think seeing `CellAlignment` there was unexpected since I didn't think of it as a cell but totally get wanting alignment values to be consistent across the library/design system 👍

src/DataTable/Table.tsx
mperrotti: Sure!

joshblack (approved):
🥳

PR #3069, opened 2023-03-23
Add repro for ActionList + Trailing button bug

siddharthkp:
before: 

image

can be overwritten with sx, but does it look fixed?

```jsx
<ActionList.TrailingVisual sx={{height: 'fit-content'}}>
  <IconButton icon={StarIcon} aria-label="star this repo" />
</ActionList.TrailingVisual>
```

image


colebemis:
For context, we originally found this use-case in `app/assets/modules/blackbird-monolith/components/FacetsPane/FacetOption.tsx`

https://github.com/github/github/pull/266085/files#diff-303c455563b0652272aac3f242cb5e743e902f3d971cd2f0d095caebf1458378

PR #3071, opened 2023-03-23
Add a component to render placeholder text in empty cells

mperrotti:
Adds a helper component for rendering placeholder text to explain why a cell has no content.

Closes https://github.com/github/primer/issues/1958

### Screenshots

Screenshot 2023-03-23 at 2 13 32 PM

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
I also considered adding this as a property on the `columns` object, but that locks the consumer into a single placeholder message for all empty cells in the column. I went with a helper component incase they have different text they want to render depending on _why_ there is no content.

Reviewers, which direction do you prefer?

src/DataTable/Table.tsx
```diff
@@ -318,6 +319,10 @@ function TableCell({children, scope, ...rest}: TableCellProps) {
   )
 }
 
+function TableCellPlaceholder({children}: TableCellProps) {
```
joshblack, line 322: `TabelCellProps` was surprising as the type for `TableCellPlaceholder` since it wouldn't be rendering a `<td>` and instead would be inside of a `<td>`. If this is the case, would it help to use a standalone type here that just has `children` defined on it?

```suggestion
type TableCellPlaceholderProps = React.ChildrenWithProps<{}>;

function TableCellPlaceholder({children}: TableCellPlaceholderProps) {
```

joshblack (changes requested):
Love the story and examples! Just left one comment on the type signature for `TableCellPlaceholder`, let me know what you think 👀

src/DataTable/Table.tsx
mperrotti: Great point! Silly mistake on my end.

joshblack (approved):
Looks great! Sorry again about mixing up `ChildrenWithProps` with `PropsWithChildren` 😅 haha

colebemis (approved):

PR #3080, opened 2023-03-27
ADR: Styling with css

siddharthkp:
[See rendered version →](https://github.com/primer/react/blob/adr-css/contributor-docs/adrs/adr-016-css.md)


contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+| Approved | ✅                            |
+| Adopted  | 🚧 _TODO: add tracking issue_ |
+
+## Context
+
+Reasons for change: Performance, utilise css variables
+
+The main reason for changing our styling architecture is to remove performance issues for applications:
+
+Relevant parts from [primer/discussions#1928](https://github.com/github/primer/discussions/1928#reasons):
```
broccolinisoup, line 16:
```suggestion
Relevant parts from [primer/discussions#1928](https://github.com/github/primer/discussions/1928#reasons)(Hubbers only link):
```

contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+
+&nbsp;
+
+### Other options considered
+
+1. Continue to author styles in typescript, compile them to .css files before shipping
+
+   We benefit from type-safety while authoring components, compiled to .css files with a babel plugin before shipping.
+
+   We are not not chosing this option to have more control on the output by keeping the authored and shipped code similar.
```
broccolinisoup, line 103:
```suggestion
   We are not choosing this option to have more control on the output by keeping the authored and shipped code similar.
```

Or is that meant to be double negative? 👀

contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+
+3. Keep styled-components underneath for supporting sx prop.
+
+   Keep backward compatibility, we might change this in the future
+
+Code example:
+
+```jsx
+import * as css from './ActionList.css'
+import classnames from 'classnames'
```
joshblack, line 75: Unrelated, would there be any interest in moving from `classnames` to `clsx`?

contributor-docs/adrs/adr-016-css.md
siddharthkp: oh lol, my bad! That is not meant to be a double negative 😅

contributor-docs/adrs/adr-016-css.md
siddharthkp: Thank you!

contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+   ii. Application benchmarks: opting out of IconButton for PR diff view (>500 files, [gh/gh#243269](https://github.com/github/github/pull/243269))
+
+   dynamic styles with sx prop: 400ms
+   dynamic styles with .css files: 165ms (60% faster)
+
+The other reasons for change are to utilise css variables for theming and improve the guardrails of consuming primer/primitives
+
+## Decisions
+
+1. Manually refactor styles to css files + css modules
```
joshblack, line 59: Would the scope of this ADR include techniques for:

- Sharing between CSS files
- Managing dependencies between CSS files

Here are some situations that I was thinking of:

- I have a helper class that I want to make sure is available to a component (like `sr-only`)
- I have a helper mixin (in the sass sense) to reset a button in order to customize styling
- I have a helper mixin or function to help with managing breakpoints

contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+
+   Use css variables from primer/primitives for inexpensive theming support, especially with server side rendering
```
joshblack, line 61: What's the expectation for using CSS Custom Properties along with stuff like breakpoints? From what I remember, syntax like the following would not work:

```css
@media (max-width: var(--some-breakpoint)) {
  // ...
}
```

Since the `@media` query is not scoped under a `:root`. Would we use helper mixins/functions/etc?

joshblack (commented):
Looking great! Just had a couple of questions around scope of the ADR but overall just wanted to say that I'm excited for the direction and really appreciate how much time you spent into preparing this and for documenting and sharing your work along the way 🔥

contributor-docs/adrs/adr-016-css.md
siddharthkp: Oh yep for sure, this is just a placeholder

contributor-docs/adrs/adr-016-css.md
siddharthkp: Haven't included any of these. Hoping to define those as we go.

> ### Out of scope
> 
> Methodologies and guidelines on how to author CSS are not part of this ADR/decision. We will define those as we move components to css and find repeating patterns.

https://github.com/primer/react/blob/adr-css/contributor-docs/adrs/adr-016-css.md#out-of-scope

contributor-docs/adrs/adr-016-css.md
langermank: We ship CSS vars to work with the custom media plugin 🎉

colebemis (approved):
Thanks for all the thought you've put into this decision, @siddharthkp! Excited to move forward with this 🎉

contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+
+2. Use css modules with hash
+
+   By importing styles in the component, we can create an explicit dependency between the two. This enables applications using primer to have control over bundling and delivery (they can be split by page or bundled as a common chunk for all react pages to share)
+
+3. Keep styled-components underneath for supporting sx prop.
+
+   Keep backward compatibility, we might change this in the future
```
mperrotti, line 69: I strongly advocate for dropping support for sx/styled-components

mperrotti (approved):

contributor-docs/adrs/adr-016-css.md
joshblack: @siddharthkp these could be worth bringing up at this point as it would influence the technical decision (in this case CSS Modules) as the different approaches have pros/cons when it comes to share and re-use.

Or should I read this ADR as more about styling with CSS (as opposed to styled-components) and that decisions like CSS Modules are not a part of the scope of this ADR?

contributor-docs/adrs/adr-016-css.md
joshblack: @siddharthkp what would the build step for this end up looking like? Would it be at publish/build time for the package or are we expecting end users to include this as part of their build toolchain?

@langermank could you share an example of how the CSS Custom Properties from primitives would interop with the custom media plugin?

joshblack:
Quick question, is there a prototype that demonstrates what the package input and output would look like after accepting this ADR? In particular, would be curious about the build step and what files end up getting published and what a consumer build setup would look like.

Was trying to put something together while reading through the ADR and I was struggling with how to:

- Emit JavaScript files
- Make sure the JavaScript files include the import to the CSS Module (or should this be to a CSS file?)
- Find a way to process the CSS Module through something like Postcss without removing the import

contributor-docs/adrs/adr-016-css.md
langermank: I have this running in a remix app with global CSS (not CSS modules). I have a watcher that compiles the PostCSS to CSS.

Media query looks like this
```
@custom-media --md-n-above (width >= 768px);
```

PostCSS file
```
@media (--md-n-above) {
  .AppLayout {
    --sidepane-width: var(--size-280);
    --header-height: var(--size-64);
    --topbar-visibility: visible;
  }
}
```

Built CSS file

```
@media (min-width: 768px) {
  .AppLayout {
    --sidepane-width: var(--size-280);
    --header-height: var(--size-64);
    --topbar-visibility: visible;
  }
}
```

contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+import * as css from './ActionList.css'
+import classnames from 'classnames'
+
+const Button = ({className, sx, ...props}}) => {
+  return (
+    <Box
+      as="button"
+      sx={sx}
+      // it's important to pass both internal and props.className to the element
+      className={classnames(css.ActionList, className)}
```
langermank, line 83: Im not sure if this is out of scope, but we might want to document this decision to offer `className` as a prop instead of `sx` moving forward

contributor-docs/adrs/adr-016-css.md
joshblack: Thanks @langermank! Appreciate it 🙏

contributor-docs/adrs/adr-016-css.md
siddharthkp: > Or should I read this ADR as more about styling with CSS (as opposed to styled-components) and that decisions like CSS Modules are not a part of the scope of this ADR?

That was my intention, but now I think it makes sense to include css modules as part of the decision

contributor-docs/adrs/adr-016-css.md
siddharthkp: We'd need to keep supporting both sx and className for the short term in order to not break dotcom. But, after we migrate off sx prop in dotcom, we'll be able to drop sx and styled-components completely!

contributor-docs/adrs/adr-016-css.md
```diff
@@ -0,0 +1,103 @@
...
+
+The other reasons for change are to utilise css variables for theming and improve the guardrails of consuming primer/primitives
+
+## Decisions
+
+1. Manually refactor styles to css files + css modules
+
+   Use css variables from primer/primitives for inexpensive theming support, especially with server side rendering
+
+2. Use css modules with hash
```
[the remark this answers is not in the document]
siddharthkp: That's fair. 

I don't think we have considered alternatives to css modules. It was mostly a choice between should we use css modules or not.

Are there alternatives that you would suggest should be added here?

siddharthkp:
@joshblack, yep! I have a spike here with 2 possible options in the description: https://github.com/primer/react/pull/3065

We'd probably start with option 1 and shift to option 2, when needed in the future

contributor-docs/adrs/adr-016-css.md
kayla-glick: If it helps the discussion at all, I'm a big fan of what Bootstrap does with their breakpoints
https://getbootstrap.com/docs/5.0/layout/breakpoints/

joshblack:
@siddharthkp thanks so much! Taking a look 👀 

For the single CSS file approach, would there be a concern about bundle size? Specifically if you're only using a handful of components but you're bringing in the styles for all of them. Or are things implemented in dotcom such that this file would be shared across services?

Another question (sorry for going on and on lol), how would we handle things like "resets" or "globals" that we want to share or dedupe? For example, if we had a common reset or common CSS like from primitives for setting themes.

siddharthkp:
> For the single CSS file approach, would there be a concern about bundle size? Specifically if you're only using a handful of components but you're bringing in the styles for all of them. Or are things implemented in dotcom such that this file would be shared across services?

I think it definitely could be a concern once we reach the stage where components that are not common across pages take up a significant % in the shared bundle. That's the point where splitting a big cached bundle would be useful. (we'd still have to test that with dotcom though, because there are a bunch of big files that get downloaded so splitting might just make it slower 😅) 

From our side, we would shift from shipping a single bundle to one global file + multiple component files which can be bundled by the application's config based on usage across pages. 

I think we should start with a single bundle and test splitting it once it has a good amount of components (10-15)

> Another question (sorry for going on and on lol)

nah nah, keep em coming please!

> how would we handle things like "resets" or "globals" that we want to share or dedupe? For example, if we had a common reset or common CSS like from primitives for setting themes

(will need to test more when we get to that stage) I think even when we ship one file per component, we will have a _global_ file that needs to loaded on all pages. I don't have a clear approach on how we would author or generate that shared file, leaving that for a future ADR that comes from usage.

joshblack:
> I think we should start with a single bundle and test splitting it once it has a good amount of components (10-15)

From the consumer side, would this kind of look like this in the beginning:

```js
// Some entrypoint file
import '@primer/react/css/styles.css';

// Use components like always
import { Button } from '@primer/react';
```

And then if we want to split things by-component:

```js
// Some entrypoint file

// Base styles
import '@primer/react/css/base.css';

// Component styles
import '@primer/react/css/button.css';

// Use components like always
import { Button } from '@primer/react';
```

Or would it be that importing `Button` automatically bring in the css file?

> I think even when we ship one file per component, we will have a global file that needs to loaded on all pages.

Agreed, generally I think the "shared" modules end up being:

- Reset styles
- Base styles, e.g. for fonts or motion or for CSS Custom Properties on `:root`

This kind of implicit dependency/ordering seems hard to express without some kind of module system which is why this was coming up in case we need some kind of processing on top of just plain css files (e.g. if we want to use imports from postcss or [sass modules](https://css-tricks.com/introducing-sass-modules/))

broccolinisoup (approved):
This is looking amazing 🔥 Thank you so much for all the effort you put into this work, it is appreciated so much 💖  I am very excited for the direction we are going to 🚀

lesliecdubs:
Bump @siddharthkp as this is going stale. Let me know if this ADR warrants an assigned decision-maker; otherwise, let's work on resolving the open comments and moving toward merge.

siddharthkp:
Yep, it's on my todo list!

PR #3087, opened 2023-03-28
3003 autocomplete when in dialog intercepts escape keypresses and click outside

### LINKED ISSUE #3003: AutoComplete, when in dialog, intercepts escape keypresses and click outside
### Description

see https://github.com/primer/react/issues/2995 for some setup

When Autocomplete is used inside a dialog, `Escape` to close the dialog fails - seemingly because the AutoComplete overlay intercepts these keypresses before they get tot he containing dialog - even when the autocomplete is not currently focuses (so the overlay itself is not visible)

### Steps to reproduce

Autocomplete in a dialog, rendered in a deferred way to ensure that it orders ontop of the dialog content.

Attempt to close the dialog with escape - note the dialog is not closeable, and the handler is never called

### Version

v35.20.0

### Browser

_No response_

green6erry:
Describe your changes here.

Closes #3003 

### Screenshots

If the Dialog component includes an Autocomplete in the body, you can still close the dialog by pressing the `escape` key.
### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Autocomplete/AutocompleteOverlay.tsx
```diff
@@ -49,6 +49,8 @@ function AutocompleteOverlay({
     return null
   }
 
+  if (!showMenu) return null
```
siddharthkp, line 52: Hi! 

Given that Overlay is part of the [public API](https://primer.style/react/Overlay/) and it accepts a `visibility` prop, should the logic of attaching listeners based on visibility be inside Overlay instead?

green6erry: Hey @siddharthkp, good question! I had originally tried to fix this bug through the `Overlay`, but essentially, `Overlay` doesn't always have a dropdown menu, but `Autocomplete` does. To implement this fix for the case where an  `Autocomplete` is contained in an `Overlay` (where the dropdown menu needs to behave as it already does) we would need to pass along a ref up to the parent, or create a context to do this, amongst other things. After spending some time going down that rabbit hole, your single line fix seemed to make a lot more sense! It does make sense to use the `showMenu` variable that already exists instead of trying to recreate it elsewhere.

Also, `visibility` in `Overlay` still needs to as it already does. We're using that every time in a `useLayoutEffect` hook, which is to make the animation work properly. The `AutoComplete` dropdown menu shouldn't effect that, it just needs to be the focus of the key event when it's being interacted with.

src/stories/Autocomplete.stories.tsx
```diff
@@ -740,6 +752,54 @@ export const InOverlayWithCustomScrollContainerRef = (args: FormControlArgs<Auto
   )
 }
 
+export const NestedInsideOverlay = (args: FormControlArgs<AutocompleteArgs>) => {
```
mperrotti, line 755: I'm not able to move focus between the inputs using the `Tab` key.

src/stories/Dialog.stories.tsx
```diff
@@ -1,9 +1,57 @@
...
+  {text: 'six', id: 6},
+  {text: 'seven', id: 7},
+  {text: 'twenty', id: 20},
+  {text: 'twentyone', id: 21},
+]
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AutocompleteItemProps<T = Record<string, any>> = MandateProps<ItemProps, 'id'> & {metadata?: T}
+
+const LabelledAutocomplete = <T extends AutocompleteItemProps>({
```
mperrotti, line 27: Just a heads up, I have a slightly related PR that adds a Storybook example for rendering an Autocomplete in a Dialog: https://github.com/primer/react/pull/3374

Story in preview deployment: https://primer-59381bd591-13348165.drafts.github.io/storybook/?path=/story/components-forms-autocomplete--in-a-dialog

src/Flash/__tests__/__snapshots__/Flash.test.tsx.snap
```diff
@@ -3,13 +3,13 @@
 exports[`Flash renders consistently 1`] = `
 .c0 {
   position: relative;
-  color: #1F2328;
+  color: #24292f;
```
mperrotti, line 6: I'm confused by all of these snapshot updates with different token values. It doesn't look like you changed anything that would cause this. Am I missing something? 🤔

src/stories/Autocomplete.stories.tsx
```diff
@@ -740,6 +752,54 @@ export const InOverlayWithCustomScrollContainerRef = (args: FormControlArgs<Auto
...
+      width="large"
+      height="small"
+      side="inside-top"
+      renderAnchor={props => <Button {...props}>open overlay</Button>}
+      // overlayProps={{
+      //   onEscape: () => {
+      //     alert('parent escape')
+      //     setIsOverlayOpen(false)
+      //   },
+      // }}
```
mperrotti, line 775: Is this intentionally commented out?

lesliecdubs:
@green6erry bump on this one for when you're back in office. Let's work on getting this closed out please!

src/Flash/__tests__/__snapshots__/Flash.test.tsx.snap
green6erry: I am too. Looking into it.

src/stories/Autocomplete.stories.tsx
green6erry: Yes, but I'll remove it.

src/stories/Autocomplete.stories.tsx
green6erry: Okay, good to know.

siddharthkp:
@green6erry Feel free to re-request review when you're ready :)

src/Autocomplete/AutocompleteOverlay.tsx
```diff
@@ -58,14 +59,17 @@ function AutocompleteOverlay({
...
         overflow: 'auto',
       }}
       {...overlayProps}
     >
       {children}
     </Overlay>
+  ) : (
+    // HACK: This ensures AutocompleteMenu is still mounted when closing the menu and all of the hooks inside of it are still called.
+    // A better way to do this would be to move the hooks to AutocompleteOverlay or somewhere that won't get unmounted.
+    <VisuallyHidden>{children}</VisuallyHidden>
```
siddharthkp, line 72: Comparing the a11y tree both production and this branch, in production (left) the listbox is not in the a11y tree until the autocomplete is opened. While, in this branch, it's in the a11y tree even when the autocomplete is closed.

because Overlay uses the css property `visibility`, [it removes the contents from the accessibility tree](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns), which VisuallyHidden does not. 

in production (left) the listbox is not in the a11y tree until the autocomplete is opened. While, in this branch, it's in the a11y tree even when the autocomplete is closed


If we can rid of the hack, that would of course be perfect. But, if not, we probably would need to make it aria-hidden as well.

src/hooks/useDialog.ts
```diff
@@ -103,19 +104,23 @@ function useDialog({
...
     },
-    [handleTab, onDismiss],
+    [handleTab],
   )
 
   const getDialogProps = () => {
     return {onKeyDown}
   }
 
+  useOnEscapePress(
```
siddharthkp, line 116: nice! ✨

siddharthkp (changes requested):
I think this is great work.

1. Found a potential accessibility concern.

2. question (not blocker for this PR, something to consider after): What is the difference between Dialog/useDialog and Overlay/useOverlay?

Issue #3098, opened 2023-03-30
`IssueLabelToken` has extremely low contrast in high contrast color modes


iansan5653:
### Description

The `IssueLabelToken` component looks fine in most color modes, but is broken in high contrast modes. 

For example, the dark gray token from the Storybook looks like this in light mode:

<img width="123" alt="screenshot of dark gray issue label with black text" src="https://user-images.githubusercontent.com/2294248/228925390-498cd71e-e29f-48b0-a355-9afb7c584c62.png">

But in high contrast mode it looks disabled:

<img width="123" alt="screenshot of light gray issue label with light gray text" src="https://user-images.githubusercontent.com/2294248/228925507-d89b7c36-3a62-4f0d-9ccf-40cb05199079.png">

Here is another example, live (!) in Projects:

<img width="268" alt="screenshot of blue and purple labels in Projects appearing washed out and faded" src="https://user-images.githubusercontent.com/2294248/228925658-e6ab70fe-7de0-4a6b-a622-982b6ce87569.png">

### Steps to reproduce

Go to [this story link](https://primer.style/react/storybook/?path=/story/components-token-features--default-issue-label-token&globals=colorScheme:light_high_contrast)

### Version

v35.23.0

### Browser

Chrome

PR #3104, opened 2023-03-31
Fix `IssueLabelToken` treating alternative light schemes as dark

iansan5653:
The `IssueLabelToken` conditionally picks colors based on whether `colorScheme === 'light'`. This strategy assumes that there exist only two color schemes, when in reality there are many.

~~The fix for this is to use `resolvedColorMode`, which has four possible values: `dark`, `night`, `light`, and `day` (it's not clear what the difference between `dark` and `night` or `light` and `day` are, but this is what the value is typed as). So instead, to accurately determine whether the color scheme is a light one, we can see if `resolvedColorMode` is `light` or `day`.~~

Update: instead of using `resolvedColorMode` we decided to use `resolvedColorTheme` and see if it `startsWith` the expected color mode. This is less robust to new theme name formats but somewhat safer because there is no guarantee that the color mode actually matches the theme in real life, and the theme is what really determines the colors that are rendered.

Alternatively, we could instead call `colorScheme.includes('light')`, but I avoided this solution because it feels fragile (not robust to the possibility of future themes that are named differently.

Closes https://github.com/primer/react/issues/3098

Before:

IssueLabelToken in all themes, looking correct in 'light' and in all dark themes but wrong in alternative light themes

After:

IssueLabelToken looking correct in all light and dark themes


colebemis (approved):
Thanks for the quick fix ⚡

broccolinisoup (approved):
Thanks! 🚀  Just missing a changeset.

Issue #3072, opened 2023-03-23
Interactive `IssueLabelToken` does not render any hover effects


iansan5653:
### Description

The `Token` component gets a darker background and a box shadow when it is interactive and hovered over. The `IssueLabelToken` is a variant of this component that should, to the extent possible, match the `Token` component and work well with it. However, `IssueLabelToken` does not show any hover states when it is interactive, making it look broken next to `Token`.

I can see why it would be difficult to show a different background color on hover because the label has a hardcoded user-controlled background color, but we should at least show the box shadow on hover to make it match the `Token`.

> 🎬 Video alt text: A screen capture showing a mouse pointer hovering over items in a set of tokens. Each token shows a mild shadow and a darker background on hover except for the tokens representing labels, which only show indicate interactivity by showing a 'click' mouse pointer.

https://user-images.githubusercontent.com/2294248/227335647-b21b18b7-b4bb-4a6e-b3a6-c059cb9d391e.mov

### Steps to reproduce

Render an interactive `IssueLabelToken`, ie by adding `as="button"` or `as="a"`:

```tsx
<IssueLabelToken text="good first issue" fillColor="#0366d6" as="button" />
```

### Version

v35.22.0

### Browser

Chrome

tallys:
tagging in @lukasoppermann for visibility

langermank:
I created a quick prototype for this one as a way for me to experiment with the new `color-mix()` CSS function 😄 https://github.com/primer/react/pull/3112

I think it needs fine tuning around hue and saturation, as it can easily look muddy by just adding white or black. Unassigning myself for now– maybe @lukasoppermann and I can pair on it a bit.

lesliecdubs:
👋🏻 @lukasoppermann I'm assigning this your way based on the discussion above, but please give me or @tallys a shout if you're not sure how to fit this in.

lukasoppermann:
> In project board cards, the tokens on the cards can be clicked to filter the view. For example, clicking the bug label would filter the view by label:bug.

Okay, yeah, that makes sense. I'll look into @langermank work.

iansan5653:
> However this only covers Primer React, would that be enough?

:+1: I only work in React apps so I can't speak for Rails/VC.

PR #3112, opened 2023-04-03
Add hover state to `IssueLabelToken`

langermank:
The most elegant solution for adding state to user generated colors is _probably_ to use the upcoming `color-mix()` function. It doesn't have full browser support yet, so I added a fallback using `background-image` with a `linear-gradient`. 

This is just a draft/prototype as I think it needs some fine tuning. 

Closes https://github.com/primer/react/issues/3072

### Screenshots

https://user-images.githubusercontent.com/18661030/229602006-bfb482e4-4758-4375-a0e9-576022eb10e7.mp4

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


lukasoppermann:
@langermank I think the way it was set up does not work. Due to the gradient overlaying the bg, the `color-mix` did not have a great effect.

I replaced this now with an overlay for light and a hsla color change for dark mode. That should fix it for now. It will never be as good as with predefined colors that have pre-defined hover colors though.

colebemis (approved):
Looks great in Storybook 🚀

PR #3124, opened 2023-04-05
replace createRef with useRef

mattcosta7:
Describe your changes here.

`createRef` is going to create a new ref object on every render, which will cause a new object to break any memoization guarantees, as well as re-evaluate hook dependencies, potentially more than desired/necessary

There's no good reason to call `createRef` in a functional component, where `useRef` hooks should be used instead, which maintain object equality on re-renders (by re-using the same object as the initial render)

### Screenshots

No visual change, should result in _minor_ perf improvements, and avoid potential issues later

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


examples/nextjs/package-lock.json
```diff
@@ -18,7 +18,7 @@
     },
     "../..": {
       "name": "@primer/react",
-      "version": "35.23.0",
+      "version": "35.24.0",
```
mattcosta7, line 21: curious why this wasn't checked in already?

docs/package-lock.json
```diff
@@ -4848,36 +4848,6 @@
       "resolved": "https://registry.npmjs.org/axobject-query/-/axobject-query-2.2.0.tgz",
       "integrity": "sha512-Td525n+iPOOyUQIeBfcASuG6uJsDOITl7Mds5gFyerkWiX7qhUTdYUBlSgNMyVqtSJqwpt1kXGLdUt6SykLMRA=="
     },
-    "node_modules/babel-eslint": {
```
mattcosta7, line 4851: this is just from running npm i, curious why it's changed?

joshblack (approved):
Thanks for doing this!

examples/nextjs/package-lock.json
joshblack: I think this might be a downside of not having the project setup with workspaces as the changeset release action is not aware of this workspace when `@primer/react` is updated

PR #3129, opened 2023-04-06
refactor(project): remove @testing-library/react-hooks in favor of @testing-library/react

joshblack:
Noticed over in: https://github.com/primer/react/pull/3077 that later versions of `@testing-library/react-hooks` do not support React 18 🤔 

When looking into this, it seems that the recommendation is to instead use `renderHook` from `@testing-library/react` directly: https://testing-library.com/docs/react-testing-library/api/#renderhook

This PR updates the call sites where we use `@testing-library/react-hooks` and also adds in an eslint rule to suggest importing from `@testing-library/react` instead, for example:

```bash
  4:9  error  'renderHook' import from '@testing-library/react-hooks' is restricted. Please use `renderHook` from `@testing-library/react` instead  no-restricted-imports
```

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,6 @@
+lib-esm
+lib
+dist
+.next
+generated
+node_modules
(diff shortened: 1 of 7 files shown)


joshblack:
Ah, seems like this is not a thing in the previous major version of `@testing-library/react` which is for React 17 😞 

Seems like a good one to table until we have React 18 support

PR #3130, opened 2023-04-06
test(e2e): update ActionMenu e2e test to remove custom contrast rules

joshblack:
This test suite has been flaky since the changes to primitives due to the custom contrast rule and the borders that trigger them.

This PR updates the custom contrast rules to use what is present in the newer versions of these e2e tests so that contrast checks can be enabled globally once primitive updates are shipped.


siddharthkp (approved):
Nice!

PR #3133, opened 2023-04-06
chore(project): update project to use Node.js v18 LTS

joshblack:
Update project from Node.js v16 LTS to Node.js v18 LTS.

> **Warning**
> This change would require maintainers to update to Node.js v18 for development

#### Changelog

**New**

**Changed**

- Update `.nvmrc` file
- Update workflows to use `node-version: 18`
- Update `devcontainer` base image

**Removed**


.github/workflows/docs_migration.yml
```diff
@@ -7,7 +7,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16.x
+          node-version: 18
```
colebemis, line 10: Just curious, are `18.x` and `18` equivalent?

colebemis (approved):
🚀

.github/workflows/docs_migration.yml
joshblack: @colebemis I believe so, will definitely keep an eye on it to make sure it's using the latest 18 / behaving as we would expect 18.x to behave 👀 

I'm also down to use that syntax (`18.x`) if that's easier to understand 👍 The project was using a mix so I went with what was most commonly used.

Another option is that we can use `lts` tags, e.g. `lts/Hydrogen` would correspond to the corresponding v18 LTS version.

PR #3151, opened 2023-04-11
refactor(SubNav): move files, create storybook

tbenning:

For reference, since the components are similar I took a look at: 
- https://github.com/primer/react/pull/2871

**Checklist**
- [X] Ensure your chosen component has a component directory in the src directory
- [X] Ensure the component has a root story file in the component directory
- [X] Ensure that the root story file exports a Default story
- [X] (Optional) Create a Playground story
- [X] Move all other stories for the component into a .features.stories.tsx file
- [X] Opt the component into storybook tests
- [ ] (Optional) Create visual regression tests for the component


tbenning:
Not 100% sure how to write e2e tests, so left it since it was optional 👍

src/SubNav/index.ts
```diff
@@ -0,0 +1 @@
+export {default, SubNavProps, SubNavLinkProps} from './SubNav'
```
colebemis, line 1:
```suggestion
export {default, SubNavProps, SubNavLinkProps, SubNavLinksProps} from './SubNav'
```

colebemis (approved):
🚀

PR #3154, opened 2023-04-11
PageLayout.Content should not use main landmark by default.

radglob:
`PageLayout.Content` will no longer render as `main` by default. Instead, end users can pass a specific `as` prop to set the rendered output.

Closes https://github.com/github/primer/issues/1586

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge


src/PageLayout/PageLayout.test.tsx
```diff
@@ -123,7 +123,9 @@ describe('PageLayout', () => {
       <ThemeProvider>
         <PageLayout>
           <PageLayout.Header aria-label="header">Header</PageLayout.Header>
-          <PageLayout.Content aria-label="content">Content</PageLayout.Content>
+          <PageLayout.Content as="main" aria-label="content">
```
colebemis, line 126: Instead of adding the `as` prop to PageLayout.Content, I'd prefer to nest the <main> element inside the content:

```suggestion
          <PageLayout.Content>
            <main aria-label="content>
```

The `as` prop API is often too permissive and creates additional type complexity. Can you think of any downsides to nesting the <main> tag inside the content component?

radglob: I don't think it really matters in the context of this test. I think this will require a change to the assertions though.

src/PageLayout/PageLayout.tsx
```diff
@@ -397,6 +397,7 @@ export type PageLayoutContentProps = {
   width?: keyof typeof contentWidths
   padding?: keyof typeof SPACING_MAP
   hidden?: boolean | ResponsiveValue<boolean>
+  as?: React.ElementType
```
colebemis, line 400: Sorry, I should have been more clear in my [previous comment](https://github.com/primer/react/pull/3154#discussion_r1167318494). I don't think the `PageLayout.Content` component should accept an `as` prop.

```suggestion
```

.changeset/eleven-humans-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+PageLayout.Content no longer renders as `main` by default. Instead, developers must specify this using the `as` prop, which defaults to `div`.
```
TylerJDev, line 5:
```suggestion
PageLayout.Content no longer renders as `main` by default. Instead, developers may add a `main` landmark within `Pagelayout.Content` themselves.
```

src/PageLayout/PageLayout.tsx
```diff
@@ -397,6 +397,7 @@ export type PageLayoutContentProps = {
   width?: keyof typeof contentWidths
   padding?: keyof typeof SPACING_MAP
   hidden?: boolean | ResponsiveValue<boolean>
+  role?: string
```
TylerJDev, line 400: I'm wondering, is this to keep backwards compatibility with existing implementations out there? We could go all in on making the developer implicitly add the `<main>` within `PageLayout.Content` and not allow usage of `role`, but I'm curious on your thoughts here.

TylerJDev (changes requested):
Very light feedback. Let me know your thoughts! 

Pending your comment on `role` usage, we'll need to adjust the documentation in `PageLayout.mdx` for `PageLayout.Content`.

TylerJDev (approved):
Left a comment for `role` usage, but I think this is good to go 🔥! There's some wording we might have to change in the docs around `main` usage now - but this can be a follow up PR later as it isn't a blocker.

Issue #3158, opened 2023-04-12
Update types for Tooltip in `main`


joshblack:
We had a couple of type errors show up in dotcom after the tooltip updates were brought in:

- Property 'className' does not exist on type 'ControlledTooltipProps'.
- Property 'role' does not exist on type 'IntrinsicAttributes & { direction?: "s" | "w" | ...
- Property 'className' does not exist on type 'IntrinsicAttributes & { direction?: "s" | "w" | ...
- Property 'id' does not exist on type 'IntrinsicAttributes & Omit<TooltipProps, "children"> & { open?: boolean | undefined;  ...
- Object literal may only specify known properties, and 'className' does not exist in type 'PortalTooltipProps'.
- Property 'className' does not exist on type 'IntrinsicAttributes & { direction?: "s" | "w" | ...
- Property 'style' does not exist on type 'IntrinsicAttributes & Omit<TooltipProps, "children"> & { open?: boolean | undefined;  ...
- Property 'id' does not exist on type 'PortalTooltipProps'.

I believe this is due to `TooltipProps` no longer intersecting with the styled component (specifically `& ComponentProps<typeof TooltipBase>`)

PR #3159, opened 2023-04-12
fix(Tooltip): update TypeScript types for TooltipProps

joshblack:
Closes https://github.com/primer/react/issues/3158

#### Changelog

**New**

- Add type test for "additional props" like `id`, `style`, etc

**Changed**

- Rename `TooltipEL` to `StyledTooltip` (super cosmetic change, let me know if I should revert it back!)
- Remove explicit generic from the styled tooltip
- Update `TooltipProps` to intersect with `ComponentProps<typeof StyledTooltip>`

**Removed**

### DIFF
--- a/.changeset/dirty-files-talk.md
+++ b/.changeset/dirty-files-talk.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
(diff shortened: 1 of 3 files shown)


broccolinisoup (approved):
🚀

PR #3160, opened 2023-04-12
fix(PageLayout): update Pane to warn instead of error

joshblack:
Closes https://github.com/github/primer/issues/2038

This PR proposes an alternative to the `invariant()` used in `PageLayout.Pane` when overflow is encountered in the component.

Instead, we will provide a `warning()` (note: this is stripped in production) that will detail what has happened and what a developer can do to address the warning.

In the future, we can set this as an `invariant()` either at runtime or in the type system but this may provide a smoother upgrade path for this behavior.

Would love any thoughts/feedback on this approach!


broccolinisoup (approved):
This approach makes sense to me! Especially after reading the issue and I agree it would be a smoother experience for the consumers to start with warning. I also think that this approach would be really good for other components as well when we introduce strict checks on accessibility props i.e. `Tooltip` .  Would it be fair to generalise that we can introduce strict checks with warning in minor releases and move to use `invariant` in major releases?

Thanks for fixing this! ✨

joshblack:
> Would it be fair to generalise that we can introduce strict checks with warning in minor releases and move to use invariant in major releases?

That makes total sense to me, I would love to do that.

PR #3169, opened 2023-04-13
Tooltip: Revert tooltip remediations until resolving breaking changes at dotcom

broccolinisoup:
## **TL;DR**

`Tooltip` component remediations and refactor causes breaking changes at dotcom and we are reverting the changes back until we resolve the issues. 

## Context

Some of the mis-usages of `Tooltip` either causes breaking changes or requires collaboration with the feature teams (potentially design input) to find a way to convey the tooltip information in other ways because some tooltips are used on non-interactive elements and we should be correcting them. [Here (Hubbers only link)](https://github.com/github/github/blob/4a9b90021d97cbffe0e3906b6b20bccab8dfe31f/app/assets/modules/react-code-view/components/blob/Banners/Bidi.tsx#L77) is a good example of a breaking change and a usage that requires collaboration to be corrected.

[In this usage (Hubbers only link)](https://github.com/github/github/blob/4a9b90021d97cbffe0e3906b6b20bccab8dfe31f/app/assets/modules/react-code-view/components/blob/Banners/Bidi.tsx#L77), folks use `StyledOcticon` as a trigger element for `Tooltip`.  The reason it breaks is that `StyledOcticons` are not using `React.forwardRef()` and in the new implementation we are trying to give `ref`  to the trigger element to do some checks for accessibility. See [error log (Hubbers only link)](https://github.com/github/github/actions/runs/4683388108/jobs/8298414829#step:7:4690) and we have [an issue filed](https://github.com/primer/octicons/issues/910) for our icons to use `React.forwardRef()` as well. The plan is to revert the PR back and take this issue into the design office hours to see how we are going to address the breaking changes.


broccolinisoup:
cc @TylerJDev

joshblack (approved):
Sorry that it ended up not working out, such a bummer 😞

colebemis (approved):

PR #3173, opened 2023-04-14
Use SSR-compatible slot implementation in ActionList/NavList

colebemis:
Swaps out the underlying slots implementation for `ActionList`/`NavList`.

See https://github.com/primer/react/issues/1690 for more context.

### DIFF
--- a/.changeset/lemon-berries-run.md
+++ b/.changeset/lemon-berries-run.md
@@ -1,0 +1,7 @@
+---
+"@primer/react": patch
+---
+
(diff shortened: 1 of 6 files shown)


joshblack (approved):
The refactor makes a ton of sense to me 👍 Cool to see things shift up a level and it's awesome how similar the implementation ends up being.

The only thing that came up was if we needed to do anything for the value of the context provider but it seems to have the same amount of re-renders as before for state updates so should be good and could be a source of performance optimization in the future if we ever need it.

PR #3186, opened 2023-04-17
feat(project): add subpath pattern to package.json to restrict internal imports

joshblack:
This PR implements the "Internal Modules" ADR so that anything imported from `./lib-esm/internal` will not compile.

#### Changelog

**New**

- Add `src/internal/hooks`, move `useOverflow` into this folder, and update improts

**Changed**

- Add `./lib-esm/internal/*` pattern to `package.json` to restrict imports from an internal folder

**Removed**

- Extra version of `useOverflow`


colebemis (approved):
🎉

PR #3196, opened 2023-04-19
Fix `useDynamicTextareaHeight` initial render with slots

iansan5653:
Sometimes, `useDynamicTextareaHeight` cannot determine what the height of the textarea should be on initial render. This can be seen in Projects in the single-select editor dialog, which is contained inside a `Dialog` v2 component:

https://user-images.githubusercontent.com/2294248/233174756-b0c18ceb-990f-463a-9e7a-0420b2a6ce12.mov

Notice how the description field initially has extra height, but then collapses down to the correct height after the user starts typing. This is jarring and annoying.

It appears that this is caused by an interaction with the 'slots' pattern where the input is initially hidden on first render, then becomes visible after the layout effect has already run. The layout effect does not rerun when the slots resolve, causing the height to be incorrectly calculated on first render.

We could update the effect to re-run on _every_ render, but because it sets state this would cause a cyclical infinite render loop. Alternatively, we could change the effect to a regular effect (non-layout) which would cause it to always be correct, but this CSS update in an effect would cause distracting flashes when not using the slots pattern.

So this is a hacky workaround to resolve issues with slots, which is itself a hacky workaround: we run a regular effect in addition to a layout effect on first render, then only layout effects from then on.


josepmartins:
Het @iansan5653, there are some linting errors with unnecessary dependencies and the UseEffect declaration. Can you take a look at those? 🙏

iansan5653:
Should be good now!

josepmartins (approved):
LGMT! I adding @rezrah as a reviewer as well 🙏

colebemis:
@iansan5653 Is this still necessary with the new implementation of slots? https://github.com/primer/react/blob/main/src/hooks/useSlots.ts

iansan5653:
Probably not - let's keep an eye on it after we update dotcom and close this out if it resolves itself.

iansan5653:
@colebemis it looks like we haven't shipped slots updates for `Dialog` (v2) yet? Maybe this would still be useful to merge to fix the bug there for now - then we can revert it once the old slots is entirely gone.

joshblack:
bump @colebemis wanted to check in if this was something we still need to include in the release or if it's already been addressed by the slots updates mentioned above 👀

iansan5653:
😕  Actually I think I was wrong - doesn't look like `DialogV2` uses slots. The only slots I can see here are `FormControl` which already got the update and should be live in dotcom. So maybe we do still need to merge this fix - maybe I was wrong about it being caused by slots.

colebemis (approved):
Thanks for your patience! These changes look good to me 👍

src/drafts/hooks/useDynamicTextareaHeight.ts
```diff
@@ -56,7 +56,16 @@ export const useDynamicTextareaHeight = ({
...
     if (maxHeightLines !== undefined) setMaxHeight(`calc(${maxHeightLines} * ${lineHeight})`)
     // `value` is an unnecessary dependency but it enables us to recalculate as the user types
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [minHeightLines, maxHeightLines, value, elementRef])
 
+  useLayoutEffect(refreshHeight, [refreshHeight])
+
+  // With Slots, initial render of the component is delayed and so the initial layout effect can occur
+  // before the target element has actually been calculated in the DOM. But if we only use regular effects,
+  // there will be a visible flash on initial render when not using slots
```
colebemis, line 66: Should we update this comment to not mention slots since the bug wasn't caused by slots?

PR #3199, opened 2023-04-19
feat(DataTable): add support for pagination

joshblack:
Batch https://github.com/github/primer/issues/1890

Add support for pagination within a DataTable. This adds in a `Table.Pagination` component along with a storybook story to demonstrate the behavior, presentation, and semantics of pagination alongside a table. This PR does not include support for loading or error states; these will be added in a follow-up PR.

Loom walkthrough of the behavior and changes here: https://www.loom.com/share/b2b67e2102074c55a86b2617860639fc

#### Changelog

**New**

- Add `Table.Pagination` and corresponding storybook story
- Add internal components
  - `ButtonReset`
  - `LiveRegion`, `LiveRegionOutlet`, and `Message`
  - `VisuallyHidden`


src/DataTable/Pagination.tsx
```diff
@@ -0,0 +1,478 @@
...
+    pageIndex,
+    pageStart,
+    pageEnd,
+    pageCount,
+    hasPreviousPage,
+    hasNextPage,
+    selectPage,
+    selectNextPage,
+    selectPreviousPage,
+  } = usePagination({
```
colebemis, line 156: Love how this simplifies the Pagination component ❤️

src/internal/components/LiveRegion.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+function useLiveRegion() {
+  const context = React.useContext(LiveRegionContext)
+  if (!context) {
+    throw new Error('useLiveRegion() must be used within a <LiveRegion>')
+  }
+  return context
+}
+
+function LiveRegion({children}: React.PropsWithChildren) {
```
colebemis, line 19: Excited about LiveRegion! Can't wait to use this in TreeView

src/internal/components/LiveRegion.tsx
```diff
@@ -0,0 +1,65 @@
...
+function LiveRegionOutlet() {
+  const liveRegion = useLiveRegion()
+  return (
+    <VisuallyHidden role="status" aria-live="polite" aria-atomic={true}>
+      {liveRegion.message}
+    </VisuallyHidden>
+  )
+}
+
+function Message({value}: {value: string}) {
```
colebemis, line 40: Is this a declarative API for making announcements?

colebemis (approved):
Nice work on this, @joshblack! Excited for async behavior next ✨

src/internal/components/LiveRegion.tsx
joshblack: @colebemis exactly, still trying to see if it works the way we anticipate 😅 Thankfully everything announces as-intended right now but a little worried about the declarative API translating to side-effects with announcements.

mperrotti: I love it!!

src/DataTable/Pagination.tsx
```diff
@@ -0,0 +1,478 @@
+import {ChevronLeftIcon, ChevronRightIcon} from '@primer/octicons-react'
```
mperrotti, line 1: I'm a little worried about keeping this[^1] in sync with our exported [Pagination](https://primer.style/react/Pagination) component.

Do you think it makes sense to do either of these?
- Introduce a new pagination component for use-cases like these
- Do a follow up PR where we update our existing Pagination component to have some of the features you've added here

[^1]: I'm only talking about the page numbers and prev/next buttons. I don't think we have other needs for the "1-10 of 1000" bit yet.

src/DataTable/Pagination.tsx
```diff
@@ -0,0 +1,478 @@
...
+                const page = offsetStartIndex + i
+                return (
+                  <Step key={i}>
+                    <Page
+                      active={pageIndex === page}
+                      onClick={() => {
+                        selectPage(page)
+                      }}
+                    >
+                      {page + 1}
```
mperrotti, line 238: Something we do in the existing Pagination component is to append a visually hidden "..." after a page number so that screen readers change their tone when reading it.

https://github.com/primer/react/blob/main/src/Pagination/model.tsx#L182-L185

src/internal/components/VisuallyHidden.tsx
```diff
@@ -0,0 +1,25 @@
+import styled from 'styled-components'
```
mperrotti, line 1: This is almost a duplicate of our existing internal [VisuallyHidden component](https://github.com/primer/react/blob/5a33122406f77a1eb186a198584fe5b8db77ccde/src/_VisuallyHidden.tsx). I like how this one's selector specifies not to apply this to focused/active elements.

Could we just update the existing component instead of introducing a new one?

mperrotti, line 1: Non-blocking nitpick: instead of putting "internal" components in the "internal" directory, I've been using the convention of prepending a `_` to the private code's file name.

I'm happy with either direction, but I think we should be consistent about it.

Could we do a follow-up to handle "internal" component files consistently?

mperrotti (approved):
This is great! I had a few comments, but nothing worth blocking the PR over.

src/DataTable/Pagination.tsx
joshblack: I definitely think there is something there where we could merge the two components, in particular to share appearance and behavior. It'd be great if we could figure out how to allow customization of the inner pages to allow for the different navigation use-cases (button vs a). I have no idea how to expose this kind of stuff to an end user without coming up with some kind of naming scheme between the two though 🤔

src/DataTable/Pagination.tsx
joshblack: Great point, let me make sure that gets added in!

src/internal/components/VisuallyHidden.tsx
joshblack: @mperrotti I can definitely take a look to see what the splash zone would be! `_VisuallyHidden` is reference in ~16 files from what I can tell and only a handful use the `isVisible` prop so refactoring shouldn't be too bad.

joshblack: Definitely! For background, the `src/internal` bit comes from the internal modules ADR: https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-016-internal-modules.md to help prevent things from being imported unexpectedly. 

I made https://github.com/primer/react/issues/3230 to try and capture what we could look into moving over, let me know if I missed anything! 👀

src/DataTable/Pagination.tsx
joshblack: Just added, thanks again for bringing this up!

src/internal/components/VisuallyHidden.tsx
mperrotti: Ahh - I totally forgot about that ADR. Thanks!

src/DataTable/Pagination.tsx
mperrotti: Should we create an issue to figure this out? Want to discuss it during the next Primer patterns working session?

joshblack: Definitely! Made one over at: https://github.com/github/primer/issues/2164 feel free to add to it if I missed anything 👀

PR #3207, opened 2023-04-21
Refactor MarkdownViewer to make it SSR-friendly

colebemis:
## Problem

`MarkdownViewer` cannot be rendered on the server

## Solution

This PR refactors `MarkdownView` to be SSR-friendly:

- Renders HTML with `dangerouslySetInnerHTML` instead of setting `innerHTML` inside an effect.
- Replaces `useLayoutEffect` with `useEffect`. `useLayoutEffect` throws an error on server-side.

Addresses https://github.com/github/primer/issues/2090

### DIFF
--- a/src/drafts/MarkdownViewer/MarkdownViewer.tsx
+++ b/src/drafts/MarkdownViewer/MarkdownViewer.tsx
@@ -64,6 +64,8 @@ const MarkdownViewer = ({
   onLinkClick,
   openLinksInNewTab = false,
 }: MarkdownViewerProps) => {
+  // We're using state to store the HTML container because we want the value
+  // to re-run effects when it changes
   const [htmlContainer, setHtmlContainer] = React.useState<HTMLElement>()
   const htmlContainerRef = React.useCallback((node: HTMLElement | null) => {
     if (!node) return
--- a/src/drafts/MarkdownViewer/MarkdownViewer.tsx
+++ b/src/drafts/MarkdownViewer/MarkdownViewer.tsx
@@ -1,5 +1,5 @@
-import React, {DOMAttributes, useCallback, useEffect, useRef, useState} from 'react'
-import {Spinner, Box} from '../..'
+import React, {DOMAttributes, useCallback} from 'react'
+import {Box, Spinner} from '../..'
 
 import {useLinkInterception} from './_useLinkInterception'
 import {useListInteraction} from './_useListInteraction'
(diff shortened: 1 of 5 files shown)


src/drafts/MarkdownViewer/MarkdownViewer.tsx
```diff
@@ -70,25 +64,24 @@ const MarkdownViewer = ({
...
+    if (!node) return
+    setHtmlContainer(node)
+  }, [])
 
   const onChange = useCallback(
     async (value: string) => {
       try {
         await externalOnChange?.(value)
       } catch (error) {
-        setHtmlContainer(createRenderedContainer(dangerousRenderedHTML.__html))
```
iansan5653, line 88: This reset is important because the state of the DOM of the container updates optimistically. For example:

- User checks a checkbox. The box is now checked
- `externalOnChange` is triggered, allowing the parent to handle the event
- The parent handler errors (ie, API request fails)
- Here, we need to reset the HTML container so the checkbox is unchecked

This is why we were creating the HTML container manually, so we could easily reset it.

src/drafts/MarkdownViewer/MarkdownViewer.tsx
```diff
@@ -104,20 +97,16 @@ const MarkdownViewer = ({
...
       <Spinner aria-label="Loading content..." />
     </Box>
   ) : (
     <Box
-      ref={outputContainerRef}
+      ref={htmlContainerRef}
       className="markdown-body"
       sx={{fontSize: 1, maxWidth: '100%', '& > div > :last-child': {mb: 0}}}
+      dangerouslySetInnerHTML={dangerousRenderedHTML}
     />
```
iansan5653, line 110: 🤔 I don't think this will work, because when the HTML changes we need to re-run the effects to reprocess the rendered DOM. But React will reuse the container in this case, so the `outputContainerRef` will not update when the `dangerouslySetInnerHTML` does

src/drafts/MarkdownViewer/_useListInteraction.ts
```diff
@@ -32,7 +32,8 @@ export const useListInteraction = ({
   // Storing the value in a ref allows not using the markdown value as a depdency of
   // onToggleItem, which would mean we'd have to re-bind the event handlers on every change
   const markdownRef = useRef(markdownValue)
-  useLayoutEffect(() => {
+
+  useEffect(() => {
     markdownRef.current = markdownValue
   }, [markdownValue])
```
iansan5653, line 38: I think it's dangerous to update refs in regular effects because they can get out of date. Can we replace this with `useTrackingRef`? Pretty sure that uses `useIsomorphicLayoutEffect` under the hood

colebemis: Looks like `useTrackingRef` just uses `useLayoutEffect` not `useIsomorphicLayoutEffect`. I swapped this `useEffect` for `useIsomorphicLayoutEffect` 👍

src/drafts/MarkdownViewer/MarkdownViewer.tsx
colebemis: Updated 👍

PR #3227, opened 2023-04-26
Storybook & build setup for css

siddharthkp:
- Follow up for https://github.com/primer/react/pull/3226
- Subset of https://github.com/primer/react/pull/3065


src/drafts/CSSComponent/component.module.css
```diff
@@ -0,0 +1,6 @@
+.component {
+  background-color: tomato;
+  color: white;
+  width: fit-content;
+  padding: 0.625rem 0.75rem;
```
siddharthkp, line 5: not using primer/primitives yet, will create another follow up PR for that! (it has complexities related to theming)

src/drafts/CSSComponent/index.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+import merge from 'classnames'
+import classNames from './component.module.css'
+
+export const Component: React.FC<React.HTMLProps<HTMLDivElement>> = ({className, ...props}) => {
```
colebemis, line 5: What purpose does this component serve?

siddharthkp: No purpose to the user, I added this as a temporary component to show in storybook and to build with rollup

colebemis: Isn't that what Button2 was for?

src/drafts/CSSComponent/component.module.css
langermank: Happy to pair or collab async on this!

src/drafts/CSSComponent/index.tsx
siddharthkp: Yep, that PR became too messy to maintain, so I wanted to set up a generic dev environment, before getting into Button specifics in a new PR

.storybook/main.js
```diff
@@ -15,6 +15,10 @@ module.exports = {
         optimizationLevel: 2,
       },
     },
+    {
+      name: 'storybook-css-modules',
```
joshblack, line 19: Would we eventually move to https://www.npmjs.com/package/@storybook/addon-styling or is there something specific about `storybook-css-modules` that we're relying on 👀

rollup.config.js
```diff
@@ -111,6 +112,11 @@ const baseConfig = {
     resolve({
       extensions,
     }),
+    postcss({
+      extract: 'components.css',
+      autoModules: false,
+      modules: {generateScopedName: 'prc_[local]-[hash:base64:5]'},
+    }),
```
joshblack, line 119: Is it possible with the plugin to emit outside of the target directory? e.g. instead of emitting both to `lib-esm/components.css` and `lib/components.css` could we emit to `css/components.css`?

joshblack (approved):
🥳 Just left a couple of questions, excited to see this get started!

.storybook/main.js
siddharthkp: We should be able to move to the storybook addon once we upgrade to storybook v7

rollup.config.js
siddharthkp: I think it should be possible. I'll try it :)

PR #3238, opened 2023-05-01
Theming for css with primitives

siddharthkp:
vertical panels of different themes, inside which a text string adapts its foreground and border color based on the theme

There are a couple things to note in this implementation:

1. temporary dev dependency on primer/css
   
   With [primer/primitives@8](https://primer.style/primitives/storybook/?path=/docs/migration-v8-guide--docs#theme-css-imports) just around the corner, we are using primitives v8 in the component with a fallback on primitives v7
   
    For v7, css variables for primitives are shipped in primer/css, which is why we have a devDependency on primer/css to test in storybook. This is only till primitives v8 are shipped in dotcom (~ 1-2 weeks), we can remove primer/css from our setup after that. 
   
2. data attributes added to BaseStyles
   color primitives change value color modes, this works by setting data attributes on the root of your application. [See documentation](https://primer.style/css/support/theming#set-a-theme).
   
   With primer/react, this should be taken care by `ThemeProvider`, but because [ThemeProvider does not render any wrapper elements](https://github.com/primer/react/blob/main/src/ThemeProvider.tsx#L119), I have chosen to [add them to BaseStyles](https://github.com/primer/react/pull/3238/files#diff-5cad94b1fb0ac48b138379a37927067430301335853ae33503a9a64a626fec16R64-R66) instead, which is used along with ThemeProvider ([including in dotcom](https://github.com/github/github/blob/master/ui/packages/react-core/BaseProviders.tsx#L26))
   
   This includes mapping the additional color modes for primer/react to primer/primitives (day → light, night → dark). We should look into deprecating these in the future!


.storybook/preview.js
```diff
@@ -2,6 +2,9 @@ import {addons} from '@storybook/addons'
 import {withThemeProvider, withSurroundingElements, toolbarTypes} from '../src/utils/story-helpers'
 import {PrimerBreakpoints} from '../src/utils/layout'
 
+import '@primer/css/dist/primitives.css'
```
langermank, line 5: I'm a little concerned about add a dep for PCSS even though I know this is experimental. I think it makes more sense to just go ahead and start using the new tokens. The size tokens (imported here through primitives.css) will be rolled out to dotcom this week or next (like removing the feature flag) and the colors will start to be tested soon. Since this is a test, I think just go ahead and use the new color tokens!

Here are the imports (but from dist of course) https://github.com/primer/primitives/blob/main/docs/storybook/.storybook/preview.css

I'll be cutting a new release probably this week

src/BaseStyles.tsx
```diff
@@ -38,10 +38,33 @@ export type BaseStylesProps = ComponentProps<typeof Base>
...
-  const {colorScheme} = useTheme()
+  const {colorScheme, colorMode, dayScheme, nightScheme} = useTheme()
+
+  /**
+   * We need to map valid primer/react color modes onto valid color modes for primer/primitives
+   * valid color modes for primer/primitives: auto | light | dark
+   * valid color modes for primer/primer: auto | day | night | light | dark
+   */
+  type colorModesForPrimitives = 'auto' | 'light' | 'dark'
+  const primerColorModeToPrimitiveColorMode: {[key in ColorModeWithAuto]: colorModesForPrimitives} = {
```
langermank, line 49: Im a little confused about "day" and "night". Can we standardize on just using light/dark/auto? https://github.com/primer/primitives/blob/main/docs/storybook/.storybook/preview.js#L35

siddharthkp: Would love to 😭 but we're stuck with an extra day and night for backward compatibility. Created [an issue](https://github.com/primer/react/issues/3246) for next major release

siddharthkp:
@langermank Need to confirm if this is a breaking change or a bug,

In primer/css/color-modes.css, `[data-color-mode="light"][data-light-theme="dark"]` is a valid combination (same for all dark themes)

[see code on unpkg](https://unpkg.com/browse/@primer/css@21.0.1/dist/color-modes.css#:~:text=%5Bdata%2Dcolor%2Dmode%3Dlight%5D%5Bdata%2Dlight%2Dtheme%3Ddark%5D)

image

but in `primer/primitives/tokens-next-private/css/`, this isn't true:

screenshot:
[see code on unpkg](https://unpkg.com/browse/@primer/primitives@7.11.7/tokens-next-private/css/functional/themes/dark.css)

screenshot:
image

.storybook/preview.js
siddharthkp: Made v8 the default now with a fallback on v7. We can remove it as soon as v8 is rolled out in dotcom :)

joshblack:
@siddharthkp quick question, in the dark themes it seems like the border tokens are not being defined. Just wanted to check-in to see if this was intended with this setup or not 👀 

Screenshot 2023-05-08 at 10 16 53 AM

It seems like potentially for these combos the custom properties are not being set at the theme provider level

Screenshot 2023-05-08 at 10 20 42 AM

siddharthkp:
@joshblack Yep! Noted that in https://github.com/primer/react/pull/3238#issuecomment-1534825375

Seems like a bug (or a intentional breaking change) in primitives, talking to @langermank about it.

lukasoppermann:
>In primer/css/color-modes.css, [data-color-mode="light"][data-light-theme="dark"] is a valid combination (same for all dark themes)

@siddharthkp this has been removed, as we figured that there is no way to get this combination in the wild. A few people checked it, but maybe we missed something. If we did, how can you get to this combination?

siddharthkp:
> @siddharthkp this has been removed, as we figured that there is no way to get this combination in the wild. 

That's perfect! Happy to drop support for it :)

siddharthkp:
For storybook, setting colorMode based on compatible scheme now. Looks much better:

image

PR #3240, opened 2023-05-01
[v36 Release] Update minimum version of react and react-dom to v18

joshblack:
Batch https://github.com/github/primer/issues/2078

This PR updates the minimum version of `react` and `react-dom` to v18 for the v36 release. It also removes our CI setup for testing against React 17 and related code that branched between React 17 and React 18

> **Note**
> The required CI checks for `next-major` will need to be updated since we no longer will have a matrix pattern for test and type-check


radglob (approved):
:shipit:

src/UnderlineNav2/UnderlineNav.test.tsx
```diff
@@ -115,12 +113,7 @@ describe('UnderlineNav', () => {
     await user.click(item)
     expect(onSelect).toHaveBeenCalledTimes(1)
 
-    if (REACT_VERSION_LATEST) {
-      expect(spy).not.toHaveBeenCalled()
-    } else {
-      // Warning: It looks like you're using the wrong act() around your test interactions
-      expect(spy).toHaveBeenCalledTimes(2)
-    }
+    expect(spy).not.toHaveBeenCalled()
```
broccolinisoup, line 116: Can we remove the spy as well or do we need to make sure and test that the warning doesn't render? If the latter, when would be a good time to remove it all together? 👀 

I am not against to keeping it btw, just trying to understand 😅

broccolinisoup (approved):
Looks great! 🥳 Just asked a question about keeping the console.warning spy.

src/UnderlineNav2/UnderlineNav.test.tsx
joshblack: @broccolinisoup I think you're totally right that we can just remove it! 👍

PR #3242, opened 2023-05-01
refactor(project): remove defaultProps from components

joshblack:
Closes https://github.com/github/primer/issues/2075


colebemis (approved):
Looks good 👍 We should test the canary version in dotcom before merging to make sure the new `as` prop type doesn't break anything.

PR #3243, opened 2023-05-01
refactor(project): remove theme-preval.js

joshblack:
The `theme-preval.js` file is used alongside `babel-plugin-preval` to generate an object containing all the tokens and values from `@primer/primitives`. For an example of its output, see: https://unpkg.com/browse/@primer/react@35.25.1/lib-esm/theme-preval.js

This approach introduces some limitations, namely that:

- It couples our project to `babel` and makes it difficult to switch to alternative bundlers (like TypeScript, esbuild, swc, etc)
- It mixes CommonJS with ESM in a way that makes it difficult to use tools like api extractor to analyze our declaration files
- It generates the largest file in our bundle at ~149kB uncompressed, 13.8kB gzip

The experiment here looks into removing our dependency on preval so that:

- We can explore other bundlers in the future
- We can simplify our bundler and babel setup by removing `babel-plugin-preval` along with transforms we use for mapping the emitted CommonJS to ESM
- We can rely on tools like api extractor to extract types instead of manually preparing them for documenting and for CI tooling

From a user perspective, this change should not be noticeable. Even though `@primer/primitives` is currently emitting CommonJS, the size difference should be minimal as we were including the package contents in the emitted `theme-preval.js` file


joshblack:
cc @colebemis would love your feedback here! 👀

colebemis (approved):
Since we're not using `polished`, I think removing `theme-preval.js` is a good idea 👍 This will make all our lives a lot easier. No more cache issues! 🎉

PR #3218, opened 2023-04-24
Release tracking

primer-css:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   [#3346](https://github.com/primer/react/pull/3346) [`63049234`](https://github.com/primer/react/commit/63049234a31b124775a3b6b67d6113cf6d9d0cb6) Thanks [@siddharthkp](https://github.com/siddharthkp)! - ActionMenu: Calling `event.preventDefault` inside `onSelect` of `ActionList.Item` will prevent the default behavior of closing the menu

-   [#3276](https://github.com/primer/react/pull/3276) [`8abf2688`](https://github.com/primer/react/commit/8abf2688952ff0e69ba843ee6ccfc5d5574abb76) Thanks [@joshblack](https://github.com/joshblack)! - Add experimental Table.ErrorDialog component

-   [#3244](https://github.com/primer/react/pull/3244) [`42a78c11`](https://github.com/primer/react/commit/42a78c116fbffda44d85cbbfb93d24f0785fde83) Thanks [@Bestra](https://github.com/Bestra)! - Add pullQueued state to StateLabel

-   [#3229](https://github.com/primer/react/pull/3229) [`c0cbdd08`](https://github.com/primer/react/commit/c0cbdd0806809236e681cab535b4a74ecdfcc14a) Thanks [@colebemis](https://github.com/colebemis)! - Add `minWidth prop to `PageLayout.Pane`and`SplitPageLayout.Pane\`

-   [#3343](https://github.com/primer/react/pull/3343) [`786013e3`](https://github.com/primer/react/commit/786013e3f44947da4fa8ecebfcb24012d93a501c) Thanks [@mperrotti](https://github.com/mperrotti)! - Allows up to 4 levels of nesting in the NavList component.

-   [#3199](https://github.com/primer/react/pull/3199) [`bfd9e0c8`](https://github.com/primer/react/commit/bfd9e0c8ee3b6b335e04c5c1d4ca82cf23760877) Thanks [@joshblack](https://github.com/joshblack)! - Add support for Pagination in DataTable

[...]

PR #3147, opened 2023-04-10
Release tracking (next major)

primer-css:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to next-major, this PR will be updated.

# Releases
### Major Changes

-   [#3569](https://github.com/primer/react/pull/3569) [`22434e2d`](https://github.com/primer/react/commit/22434e2daa2998e3735c3a6a996b709066c5465e) Thanks [@joshblack](https://github.com/joshblack)! - Update styled-components and related dependencies to v5.x

    

-   [#3542](https://github.com/primer/react/pull/3542) [`8f7186af`](https://github.com/primer/react/commit/8f7186afcfa131c14add762cf49be7abc1b6cc0d) Thanks [@broccolinisoup](https://github.com/broccolinisoup)! - Remove components from deprecated

    -   `BorderBox`, `ChoiceFieldset`, `Flex`, `Grid`, `Position`, `Dropdown`, `FormGroup`, `SelectMenu`, `InputField`, `Label`

    

-   [#3260](https://github.com/primer/react/pull/3260) [`52c8d220`](https://github.com/primer/react/commit/52c8d220c68143d15b40f5c58a87b9ccb0307a71) Thanks [@broccolinisoup](https://github.com/broccolinisoup)! - Remove components from draft bundle

    -   Move UnderlineNav2 (draft) to the main bundle

    ```diff
    - import {UnderlineNav} from '@primer/react/drafts'
    + import {UnderlineNav} from '@primer/react'
    ```

    ....

    -   Remove TreeView from drafts

    ```diff
    - import {TreeView} from '@primer/react/drafts'
    + import {TreeView} from '@primer/react'
    ```

    -   Remove SegmentedControl from drafts

    ```diff
    - import {SegmentedControl} from '@primer/react/drafts'
    + import {SegmentedControl} from '@primer/react'
    ```

    -   Remove NavList from drafts

    ```diff
    - import {NavList} from '@primer/react/drafts'
    + import {NavList} from '@primer/react'
    ```

    -   Remove SplitPageLayout from drafts

    ```diff
    - import {SplitPageLayout} from '@primer/react/drafts'
    + import {SplitPageLayout} from '@primer/react'
    ```

    

[...]

PR #3245, opened 2023-05-03
Revert "Update PRC ActionList implementation to have similar semantics to PVC. (#2878)"

broccolinisoup:
This reverts commit 87883c3fd91a5d2809268aa0dd22b87b53ba5309 ([PR](https://github.com/primer/react/pull/2878))

Because this is a major change, the PR should have been created and merged against the `next-major` branch which then would end up in the [v36 major release tracking PR](https://github.com/primer/react/pull/3147). I am reverting this commit in the main so we have [a clean minor release tracking PR](https://github.com/primer/react/pull/3218).
After we merge this PR into main, we should do;

- Cherry pick `87883c3fd91a5d2809268aa0dd22b87b53ba5309` and resolve conflicts
- Push up a new PR and make sure mark the target branch as `next-major`

After that, I think we can merge the PR as it has already been approved. 

Also, If there is any other way to solve this, please let me know or feel free to do so. This was the only one I came up with!


joshblack (approved):
Definitely makes sense with the major change in there, seems like we should have a way to either ship it in a stable behind a flag or namespace or include it with the next major release. Is there a clear preference in either case? Would this be something helpful to go over in our weekly sync or async in Slack? 👀

broccolinisoup:
Turns out the changes are meant to be minor but the changeset was incorrect. So closing this PR for now! https://github.com/primer/react/pull/3248

[Slack reference (GitHub stuff only)](https://github.slack.com/archives/C02NUUQ9C30/p1683074556249929)

PR #3247, opened 2023-05-03
Refactor FilteredActionList to address a11y violations and use new ActionList.

radglob:
The `FilteredActionList` (and by extension `SelectPanel`) use the updated `ActionList` component.

Added a `hideSelection` prop that allows us to hide the `Selection` component that usually renders when an ActionList has a `selectionVariant` defined. The dropdown in the `MarkdownEditor`'s saved replies did not show this previously, so this was added to keep the appearance the same.

Closes https://github.com/github/primer/issues/1861

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge


lesliecdubs:
@siddharthkp can you please give this a review in the next day or so? It's the last item remaining before we can close out [this batch](https://github.com/github/primer/issues/1860).

siddharthkp:
Hi!

Looks like the example in the docs is not rendering for this branch: https://primer-353202c920-13348165.drafts.github.io/SelectPanel#example

left: production, right: branch
image

siddharthkp:
@radglob Do we only need `hideSelection` for drafts/MarkdownEditor? Or are there other use cases as well?

If it's just the experimental `drafts/MarkdownEditor/SavedReplies`, I would discourage adding `hideSelection` prop to the public API for `SelectPanel` and `ActionList.Item` because it changes the visual without changing the semantics. We should not open up the API to potential misuse.

**Long term solution:**

It looks like `SavedReplies` uses `SelectPanel` but then overrides the "select" part of it 😓, this component should probably not use a SelectPanel and instead use the building blocks of SelectPanel instead (AnchoredOverlay + FilteredActionList), especially given [upcoming a11y remediations required to SelectPanel](https://github.com/github/primer/issues/2087))

**Short term solution:**

We need to pass `selectionVariant=undefined` to the underlying `ActionList`, this can be done in 2 ways

1. In `SavedReplies`, illegally and secretly pass `selectionVariant` to `SelectPanel` without making it part of the public API. [Here's a diff](https://github.com/primer/react/commit/9cd782844e1984122163805e1aa6e84b7215dbec) on what that could look like.
2. Use `ActionListContainerContext` to pass `selectionVariant` down to children without adding props to the public API. This context was built to [pass data like `SelectionVariant`](https://github.com/primer/react/blob/main/src/ActionList/ActionListContainerContext.tsx#L9) down to ActionList sub components from container components like [ActionMenu](https://github.com/primer/react/blob/main/src/ActionMenu.tsx#L121-L131) and potentially SelectPanel, FilteredActionList.

src/drafts/MarkdownEditor/_SavedReplies.tsx
```diff
@@ -54,7 +54,7 @@ export const SavedRepliesButton = () => {
       (reply, i): Item => ({
         text: reply.name,
         description: reply.content,
-        descriptionVariant: 'block',
+        descriptionVariant: 'inline',
```
siddharthkp, line 57: Checked on storybook, `inline` doesn't render inline 🤔 but as a block anyway?

image

src/SelectPanel/SelectPanel.tsx
```diff
@@ -61,6 +60,7 @@ export function SelectPanel({
   textInputProps,
   overlayProps,
   sx,
+  renderFn,
```
siddharthkp, line 63: Noticed this is a new addition to the API, but was not able to understand from the stories/examples when it should be used 🤔 

Can you explain this prop?

siddharthkp (changes requested):
Left a few questions and comments

siddharthkp (changes requested):
Left a few questions and comments

src/drafts/MarkdownEditor/_SavedReplies.tsx
radglob: No, this is inline rendering (which uses Truncate). If you set it to block, the descriptions will take up as much space as they need.
Screenshot 2023-05-11 at 10 21 51 AM

radglob: Yeah, that's what I ended up doing, but just in the case of `_SavedReplies`, since that's the only place that behavior was always occurring.

siddharthkp (commented):
Left a couple comments!

siddharthkp (commented):
Left a couple comments!

src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -63,7 +61,6 @@ export const SingleSelectStory = () => {
         selected={selected}
         onSelectedChange={setSelected}
         onFilterChange={setFilter}
-        showItemDividers={true}
```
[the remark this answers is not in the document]
radglob: Ah, good catch! I noticed that `showItemDividers` wasn't used anywhere, but I didn't realize there was something to transition to.

src/SelectPanel/SelectPanel.tsx
radglob: I removed this as a prop and just have a function defined in FilteredActionList that handles this. If we want to expose this as a customizable prop later, we can, but it currently covers every use case I saw.

colebemis:
@radglob Does this PR entirely close https://github.com/github/primer/issues/1861? It seems like it only address part of the issue:
- The ActionList used in this component is deprecated and may need to be updated

I don't see code changes to address this items:

- The aria-input-field-name rule corresponds to the item with role="listbox" which has no accessible label
- This component seems to follow a combobox interaction pattern and requires a visible label for the input. The combobox pattern may address aria-input-field-name by providing aria-labelledby on the listbox pointed towards the input
- The item with role="listbox" must have items with [role="option"](https://w3c.github.io/aria/#listbox)
- This component may require a visual label which would mean design needs to be included in this issue

src/SelectPanel/SelectPanel.tsx
siddharthkp: That's perfect! thanks!

src/SelectPanel/SelectPanel.features.stories.tsx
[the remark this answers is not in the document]
siddharthkp: > I think we should add showDividers to SelectPanel for consistency while keeping backward compatibility for the old prop showItemDividers

Reminder: We need to keep backward compatibility for `showItemDividers` until next major release. It's used in a few places: https://musical-adventure-wlr3n3k.pages.github.io/?component=selectpanel&n=selectpanel&na=selectpanel&nam=selectpanel&name=selectpanel&attribute=showitemdividers

siddharthkp (approved):
Final comments: 

- backward compatibility for `showItemDividers`
- checking if there are no other backward breaking changes

Approving in advance

src/SelectPanel/SelectPanel.features.stories.tsx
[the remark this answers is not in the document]
radglob: Do we need to keep that option in all of the stories, or just on the component?

radglob:
> @radglob Does this PR entirely close [github/primer#1861](https://github.com/github/primer/issues/1861)? It seems like it only address part of the issue:
> 
> * The ActionList used in this component is deprecated and may need to be updated
> 
> I don't see code changes to address this items:
> 
> * The aria-input-field-name rule corresponds to the item with role="listbox" which has no accessible label
> * This component seems to follow a combobox interaction pattern and requires a visible label for the input. The combobox pattern may address aria-input-field-name by providing aria-labelledby on the listbox pointed towards the input
> * The item with role="listbox" must have items with [role="option"](https://w3c.github.io/aria/#listbox)
> * This component may require a visual label which would mean design needs to be included in this issue

Great point. I think I got mired in other issues and need to circle back to those. Should be easy enough to fix now that technical concerns are out of the way.

radglob:
> > @radglob Does this PR entirely close [github/primer#1861](https://github.com/github/primer/issues/1861)? It seems like it only address part of the issue:
> > 
> > * The ActionList used in this component is deprecated and may need to be updated
> > 
> > I don't see code changes to address this items:
> > 
> > * The aria-input-field-name rule corresponds to the item with role="listbox" which has no accessible label
> > * This component seems to follow a combobox interaction pattern and requires a visible label for the input. The combobox pattern may address aria-input-field-name by providing aria-labelledby on the listbox pointed towards the input
> > * The item with role="listbox" must have items with [role="option"](https://w3c.github.io/aria/#listbox)
> > * This component may require a visual label which would mean design needs to be included in this issue
> 
> Great point. I think I got mired in other issues and need to circle back to those. Should be easy enough to fix now that technical concerns are out of the way.

These are actually fixed. I think I was looking at a stale local storybook instance.

PR #3252, opened 2023-05-05
ci: update pages check in deploy_production workflow

joshblack:
An extra step in the `deploy_production` workflow has the following error message:

The terminal output of a GitHub Actions workflow. There are several steps being shown with empty output due to an error message from the usage of the gh cli

Unfortunately, this output is used in the outputs of the guard. Specifically, in this check:

```yml
      should_deploy: ${{ steps.changeset-count.outputs.change_count == 0 && steps.has-pages.outputs.pages == 1 }}
```

As a result of this usage, our deploy guard is never truth-y and therefore storybook and an updated docs site does not get deployed.

This PR updates this step to use the built-in `GITHUB_TOKEN` secret for the `gh` cli check and adds a manual `workflow_dispatch` trigger to help with debugging in the future.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


langermank (approved):
🚀

PR #3253, opened 2023-05-05
ci: update trigger and guard for production workflow

joshblack:
This updates the `deploy_production.yml` workflow to simplify the guard step and reduce the number of times it runs.

Currently, this workflow will run on push to main and will check if there are no changesets available. If there are none, and GitHub Pages is configured, the guard will pass and the changes will deploy.

This PR makes the following changes to this flow:

* Instead of triggering on a push to `main`, this workflow instead is called when a tag is pushed
  * This happens as a part of our release workflow and will cause the deploy to match the latest tagged release
  * Pre-release tags are ignored from this trigger
* With the change above, the `guard` job is no longer needed
* Instead of checking to see if GitHub Pages are supported, this action now will only run on the `primer/react` repository
* It moves the `permissions` block from the outermost scope to the deploy job in case another job is added to this workflow

At a high-level, the flow becomes:

* A tag is pushed for a release
* The `deploy_production.yml` workflow is triggered
* A GitHub Pages deployment is created

Alternatively, one can use the `workflow_dispatch` to trigger this workflow. This is used in two scenarios:

* Deploy from a specific `tag` (preferred)
* Deploy from `main` (not preferred)

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -2,56 +2,25 @@ name: Deploy
 
 on:
   push:
-    branches:
-      - main
+    tags:
+      # Matches tags that have the shape `vX.Y.Z`. Reference:
+      # https://help.github.com/en/articles/workflow-syntax-for-github-actions#onpushpull_requesttagsbranches
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+      # Ignore tags that use a preid after `vX.Y.Z`, for example: vX.Y.Z-alpha.0
+      # https://help.github.com/en/articles/workflow-syntax-for-github-actions#example-using-positive-and-negative-patterns
+      - '!v[0-9]+.[0-9]+.[0-9]+-*'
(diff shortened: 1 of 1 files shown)


siddharthkp (approved):
I don't understand the change, but trust you to know what you're doing 😄 

Just to make sure, primer.style/react would still always point to the last stable released version, example: `v35.25.1` and not what's merged on main since the last release. Correct?

joshblack:
Great question @siddharthkp, the answer is 100% yes that it will always point to the last stable released version 👍 

The biggest change in this diff is that this workflow no longer runs on every push to `main`. Instead, it just runs whenever we push a tag (or run the workflow manually). This means that there wouldn't be a chance that a publish could occur from `main` and we don't have to do any of the changeset counting to see if we should deploy or not. Instead, we know we need to deploy if a tag is published.

Hope that makes sense, let me know if this flow makes sense in the broader context of deploys or if I'm missing anything here 👀

PR #3264, opened 2023-05-08
Add truncation features to the LabelGroup component

mperrotti:
These changes add the following truncation features to the LabelGroup component:
- truncate LabelGroup children after a static number of children (for example, truncate after the 5th label)
- truncate LabelGroup children to fit in the width of the parent

Consumers can choose to show the full list of labels inline or in an Overlay.

Closes https://github.com/github/primer/issues/2006

### Screenshots

Truncate after `n` children and show the full list inline

_Note: this screen recording is slightly out of date. Instead of rendering a "—" icon to collapse tokens, we now show a text button that says "Show less"_

https://user-images.githubusercontent.com/2313998/236933521-763b2f07-2841-4d3f-aa02-3dff5b9a1ade.mp4

Truncate to fit parent width and show the full list in an overlay

https://user-images.githubusercontent.com/2313998/236933896-6343d3da-7eec-4678-86b1-33989d90e996.mp4

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/__tests__/LabelGroup.test.tsx
```diff
@@ -1,28 +1,209 @@
...
+    expect(fourthLabel).not.toBeVisible()
+
+    await user.click(expandButton)
+
+    expect(fourthLabel).toBeVisible()
+  })
+
+  // TODO: test this some other way
+  // NOTE: It's not possible to test for side-effects of IntersectionObserver because it's a Jest mock
+  it.skip('should focus the first focusable token inline that was previously hidden when overflowStyle="inline"', async () => {
```
mperrotti, line 158: Help wanted: How might we test side-effects of `IntersectionObserver` when it has to be mocked in Jest? Is this a good case for Playwright?

src/LabelGroup/LabelGroup.tsx
```diff
@@ -1,17 +1,365 @@
...
+    setIsOverflowShown(false)
+
+    if (visibleChildCount && typeof visibleChildCount === 'number') {
+      hideChildrenAfterIndex(visibleChildCount)
+    }
+
+    // We need to manually re-focus the collapse button if we're not showing the full
+    // list in an overlay.
+    // TODO: get rid of this hack
+    setTimeout(() => {
```
mperrotti, line 204: Help wanted: I used this `setTimeout` because we need to wait until state has been updated and the expand button is back in the DOM.

This feels sloppy/hacky, but I haven't been able to figure out a better way. @siddharthkp - we paired on this last week.

src/__tests__/LabelGroup.test.tsx
```diff
@@ -1,28 +1,209 @@
...
+    await user.click(getByText('+2'))
+    await waitFor(() => getByLabelText('Show less'))
+
+    expect(document.activeElement).toEqual(getByLabelText('Show less'))
+
+    // Collapse
+    await user.click(getByLabelText('Show less'))
+    await waitFor(() => getByText('+2'))
+
+    expect(document.activeElement?.textContent).toContain('+2')
```
mperrotti, line 205: Help wanted: I had to write this weird assertion because `expect(document.activeElement).toEqual(getByText('+2'))` wasn't finding the element using `getByText('+2')`.

src/__tests__/LabelGroup.test.tsx
broccolinisoup: It is tricky! I used Playwright to test it!

broccolinisoup: It is tricky! I used Playwright to test it!

src/LabelGroup/LabelGroup.tsx
```diff
@@ -1,17 +1,365 @@
...
+      size="small"
+      aria-label="Show less"
+      variant="invisible"
+    />
+  ) : hiddenItemIds.length ? (
+    <Button ref={expandButtonRef} variant="invisible" size="small" onClick={showAllTokensInline}>
+      <VisuallyHidden>Show all</VisuallyHidden>
+      <span aria-hidden="true">+{hiddenItemIds.length}</span>
+    </Button>
+  ) : null
```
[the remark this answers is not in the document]
mperrotti: If we do it that way, a "0" will be rendered in place of the button when it's not there.

src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -101,6 +101,8 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
   focusZoneSettings,
   side = 'outside-bottom',
   align = 'start',
+  alignmentOffset,
```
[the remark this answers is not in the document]
mperrotti: In `@primer/behaviors`: https://github.com/primer/behaviors/blob/main/src/anchored-position.ts#LL235C22-L235C22

src/__tests__/LabelGroup.test.tsx
mperrotti: Thanks @broccolinisoup! I'll give that a try.

mperrotti: Added a test. It passes locally, let's hope it does in CI too 🤞

src/LabelGroup/LabelGroup.tsx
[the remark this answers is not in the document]
mperrotti: I guess we could do `(!hiddenItems.length || null) && (`, but that looks a little weird.

colebemis:
This close button looks a little squished on the "Truncate After Five" storybook example: 
CleanShot 2023-05-09 at 17 31 12@2x

https://primer-018a39f5ea-13348165.drafts.github.io/storybook/?path=/story/components-labelgroup-features--truncate-after-five

colebemis:
Design feedback: This dash icon isn't super clear to me. I could see some people confusing that button to mean "remove a token" instead of "hide some tokens." I wonder if a text button would be more clear. 
CleanShot 2023-05-09 at 17 32 33@2x

src/LabelGroup/LabelGroup.tsx
```diff
@@ -1,17 +1,365 @@
...
+    y: 0,
+    top: 0,
+    left: 0,
+    bottom: 0,
+    toJSON: () => undefined,
+  })
+
+  const {theme} = useTheme()
+
+  const overlayPaddingPx = parseInt(get('space.2')(theme), 10)
```
colebemis, line 146: FYI, this code may break when we switch to using CSS variables. Shouldn't be a problem now though.

src/LabelGroup/LabelGroup.tsx
```diff
@@ -1,17 +1,365 @@
...
+      data-overflow={overflowStyle === 'inline' && isOverflowShown ? 'inline' : undefined}
+      sx={sxProp}
+    >
+      {React.Children.map(children, (child, index) => {
+        if (hiddenItemIds.includes(index.toString())) {
+          // hidden children
+          return (
+            <div
+              // data-index is used as an identifier we can use in the IntersectionObserver
+              data-index={index}
```
colebemis, line 307: I wonder if using index as an identifier could create bugs if the labels change order (example: if tokens can be dragged and dropped).

mperrotti:
Just a heads-up: there was some feedback during a Primer patterns working session that we should also look into using ActionMenu to show truncated labels as list items. If we decide to move forward with that, this feature could be added after this merges.

broccolinisoup (commented):
We had a pairing session with @joshblack today and reviewed your PR 😊 I left some comments on the code and have a few questions about a11y as well. 

- I am curious to know if there are downsides of using an Overlay here in terms of accessibility, mainly because it is a portal. We used disclosure pattern for UnderlineNav and this seems like a very similar case and as far as I understand portals are okay for like window type modal where the keyboard focus is explicitly set on the portal itself and trapped but things like this, might cause some accessibility issues. For example, we can interact with the interactive tokens using the tab key which is expected, however when the overlay is opened, tab key doesn't work anymore and I needed to switch to using down and up arrow key to move around and it traps the focus as well. Are they intentional or downsides of using an Overlay?
- Another thing we discussed with Josh was that would that make sense to have the labels as a list? Like using `ul` and `li` mainly help screen reader users to hear the number of labels in the container. Just an idea - curious to hear your thoughts.

The component overall looks super cool! - I know from my experience that it is a very tricky behaviour 😅 Looking forward to hearing your thoughts on the comments. 

@joshblack feel free to comment/add if I misinterpreted anything!

src/LabelGroup/LabelGroup.tsx
```diff
@@ -1,17 +1,367 @@
...
+
+  React.useEffect(() => {
+    // If we're not truncating, we don't need to run this useEffect.
+    if (!visibleChildCount) {
+      return
+    }
+
+    if (visibleChildCount === 'auto') {
+      // Instatiates the IntersectionObserver to track when children fit in the container.
+      const observer = new IntersectionObserver(
```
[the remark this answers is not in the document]
mperrotti: I wanted to avoid having to measure each label for performance reasons. Each time you measure a DOM node, the browser does a reflow.

src/LabelGroup/LabelGroup.tsx
```diff
@@ -1,17 +1,367 @@
...
+  ) : (
+    <StyledLabelGroupContainer data-overflow="inline" sx={sxProp}>
+      {children}
+    </StyledLabelGroupContainer>
+  )
+}
+
+LabelGroup.displayName = 'LabelGroup'
+
+LabelGroup.defaultProps = {
```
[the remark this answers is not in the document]
mperrotti: Ah good to know! Will update.

broccolinisoup (commented):
Hi Mike! I mentioned about his comment on my previous review but not sure if it was lost within others.

I am particularly curious to know if there are downsides of using an Overlay here in terms of accessibility, mainly because it is a portal. We used disclosure pattern for UnderlineNav and this seems like a very similar case and as far as I understand portals are okay for like window type modal where the keyboard focus is explicitly set on the portal itself and trapped but things like this, might cause some accessibility issues. For example, we can interact with the interactive tokens using the tab key which is expected, however when the overlay is opened, tab key doesn't work anymore and I needed to switch to using down and up arrow key to move around and it traps the focus as well. Are they intentional or downsides of using an Overlay?

mperrotti:
@broccolinisoup - I don't think there are downsides to using an Overlay or Portal for this case. The focus will be moved inside the Overlay once it's opened, but you would use the tab key to keep moving focus within the non-modal dialog, not the arrow keys.

I showed this to @ericwbailey before I opened the PR and followed his guidance on keyboard navigation.

broccolinisoup:
>The focus will be moved inside the Overlay once it's opened, but you would use the tab key to keep moving focus within the non-modal dialog, not the arrow keys.

@mperrotti sorry this is taking so long but I cannot use tab keys when the overlay is opened, only arrow keys make me move through the items. Please see the video below and let me know if I am misunderstanding anything here.

https://github.com/primer/react/assets/1446503/f740ddf6-0e71-4bc5-a5e6-7b9cbe8eb149

broccolinisoup (commented):
Hello @mperrotti 👋🏻  I would love to clarify these two things please before moving further:

- https://github.com/primer/react/pull/#issuecomment-1628506204
- https://github.com/primer/react/pull/3264#discussion_r1218782181

Let me know if I am missing anything!

mperrotti:
ok, fixed it!

src/LabelGroup/LabelGroup.tsx
```diff
@@ -1,17 +1,367 @@
...
+  showAllTokensInline,
+}) =>
+  isOverflowShown ? (
+    <Button ref={collapseButtonRef} onClick={collapseInlineExpandedChildren} size="small" variant="invisible">
+      Show less
+    </Button>
+  ) : hiddenItemIds.length ? (
+    <Button ref={expandButtonRef} variant="invisible" size="small" onClick={showAllTokensInline}>
+      <VisuallyHidden>Show all</VisuallyHidden>
+      <span aria-hidden="true">+{hiddenItemIds.length}</span>
```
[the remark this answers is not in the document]
ericwbailey: Yes, especially if the `span` has `aria-hidden="true"` declared on it.

broccolinisoup (commented):
Thanks @mperrotti for fixing the arrow keys!! It works with the tab keys now! I have found a couple of issues while playing with it. Let me know what you think of them! 

When I resize the container when the items are open, dynamic rendering gets a bit confused.

https://github.com/primer/react/assets/1446503/e8c6bf5e-3fc1-4bfd-a3ea-efc2e2b93e3d

When we display the second row, I see we are flex wrapping but the first line jumps a bit and creates a layout shift. 

https://github.com/primer/react/assets/1446503/83b73562-0975-475c-a1c4-a4acebb0e7c5

mperrotti:
Hmm...those issues look tricky. I'll try and get them fixed though 👍

mperrotti:
Fixed those too!

mperrotti:
@broccolinisoup - could you give this another look when you get a chance?

broccolinisoup (approved):
Thanks so much @mperrotti 

I tested it again with the issues I mentioned before in mind and all working beautifully ✨

Tested with a mouse and keyboard
- Non interactive tokens with auto truncate/expand inline
- Non interactive tokens with auto truncate/expand overlay
- Interactive tokens with auto truncate/expand inline
- Interactive tokens with auto truncate/expand overlay

On Chrome, Safari, Edge and Firefox. 

It looks great to me overall. Just tiny styling and focus issue I just discovered but I don't think it is a blocker for this PR. Let me know what you think.

**Focus ring on Edge (latest)**

Label Group interactive tokens story open on edge browser and the third token is focused and the focus ring is black

**Initial focus is wrong on Firefox**

Initial focus on Firefox jumps to the element that is just before the `+X` in this story (components-labelgroup-features--truncate-auto-expand-inline-with-interactive-tokens) I can't add a video sorry - something is wrong with my Kap but let me know if you can't replicate it. 

Otherwise, I love these changes and thanks for your patience with my while I was being nitpicky 😁

ericwbailey:
I am unable to reproduce this on https://primer-018a39f5ea-13348165.drafts.github.io/storybook/?path=/story/components-labelgroup-features--truncate-auto-with-interactive-tokens.

Firefox 115.0.2 on macOS Ventura 13.5 in a private window. 

One thing I'm wondering: is focus following last click placement? That can affect things.

mperrotti:
Interesting - I'm also on Firefox 115.0.2, but macOS Ventura 13.4.1.

I'm going to play with it a little bit more, but I think it might be safe to merge as-is.

mperrotti:
I'm not able to reproduce when I load the story in it's own window. I think it's safe to merge 🙂

PR #3267, opened 2023-05-09
Add back check for GroupContext in ActionList.Selection to fix issues introduced to github/github

radglob:
While refactoring the ActionList component, there was some back and forth about how to handle ActionList.Group. Initially, it was going to be deleted, but we ultimately decided to mark it as deprecated to not introduce a breaking change. I forgot to add back this code, so existing usages of ActionList.Group were breaking in subtle ways.

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge

### DIFF
--- a/src/ActionList/Selection.tsx
+++ b/src/ActionList/Selection.tsx
@@ -1,9 +1,9 @@
 import React from 'react'
 import {CheckIcon} from '@primer/octicons-react'
-import {ListContext, ActionListProps} from './List'
+import {ListContext} from './List'
 import {ActionListItemProps} from './shared'
 import {LeadingVisualContainer} from './Visuals'
-import {ActionListGroupProps, GroupContext} from './Group'
+import {GroupContext} from './Group'
 
 type SelectionProps = Pick<ActionListItemProps, 'selected'>
 export const Selection: React.FC<React.PropsWithChildren<SelectionProps>> = ({selected}) => {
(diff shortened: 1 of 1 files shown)


src/ActionList/Selection.tsx
```diff
@@ -3,14 +3,18 @@ import {CheckIcon} from '@primer/octicons-react'
...
 export const Selection: React.FC<React.PropsWithChildren<SelectionProps>> = ({selected}) => {
+  const {selectionVariant: groupSelectionVariant} = React.useContext(GroupContext)
   const {selectionVariant: listSelectionVariant} = React.useContext(ListContext)
 
   /** selectionVariant in Group can override the selectionVariant in List root */
   /** fallback to selectionVariant from container menu if any (ActionMenu, SelectPanel ) */
-  const selectionVariant: ActionListProps['selectionVariant'] = listSelectionVariant
+  let selectionVariant: ActionListProps['selectionVariant'] | ActionListGroupProps['selectionVariant']
+  if (typeof groupSelectionVariant !== 'undefined') selectionVariant = groupSelectionVariant
+  else selectionVariant = listSelectionVariant
```
joshblack, line 17:
```suggestion
  const selectionVariant = groupSelectionVariant ?? listSelectionVariant;
```

Small refactor idea for this line, hope the types line up the same 🤔 When looking at the type here, it seems to match up with the union that exists, e.g. `false | 'single' | 'multiple' | undefined`

joshblack (changes requested):

src/ActionList/Selection.tsx
radglob: Fine with me. The above is basically what was in place before, but that's a bit more succinct!

PR #3269, opened 2023-05-09
fix(ActionList): update to read from group context if selectionVarian…

joshblack:
This is a follow-up PR to https://github.com/primer/react/pull/3267

It updates `ActionList.Item` to read from `ActionList.Group`'s `selectionVariant` value, if it exists. This currently addresses an issue downstream where a page is using `ActionList.Group` with `selectionVariant="single"` and the output is a `menuitem` instead of `menuitemradio`.


radglob (approved):
:shipit:

broccolinisoup (approved):
🚀

PR #3278, opened 2023-05-12
Add PostCSS color var fallback for upcoming CSS work

langermank:
Configures the same PostCSS plugin we'll be using in PVC to test new v8 color vars. This plugin adds a pre-determined color fallback on build, so no need to manually add a fallback to CSS.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


rollup.config.js
```diff
@@ -116,6 +118,13 @@ const baseConfig = {
       extract: 'components.css',
       autoModules: false,
       modules: {generateScopedName: 'prc_[local]-[hash:base64:5]'},
+      plugins: [
+        postcssCustomPropertiesFallback({
+          importFrom: {
+            customProperties: importedJSONFromPrimitives,
+          },
+        }),
+      ],
```
joshblack, line 127: Are these fallbacks for a "default" theme? Would it be possible to have these fallbacks be dynamic for whatever the product specifies as the default theme? e.g. in Sass this might look like:

```scss
// Exporting tokens in package
@use 'theme';

$bgColor-default: var(--bgColor-default, theme.get('bgColor-default'));

// Product usage
@use '@primer/styles/scss/themes';
@use '@primer/styles/scss/theme' with (
  $theme: themes.$dark,
);
@use '@primer/styles/scss/tokens' as *;

.my-selector {
  background: $bgColor-default;
}
```

rollup.config.js
```diff
@@ -7,6 +7,8 @@ import glob from 'fast-glob'
 import {visualizer} from 'rollup-plugin-visualizer'
 import postcss from 'rollup-plugin-postcss'
 import packageJson from './package.json'
+import postcssCustomPropertiesFallback from 'postcss-custom-properties-fallback'
+const importedJSONFromPrimitives = require('./node_modules/@primer/primitives/tokens-next-private/fallbacks/color-fallbacks.json')
```
joshblack, line 11: Can we import directly from the package or is that not a supported path?

```suggestion
const importedJSONFromPrimitives = require('@primer/primitives/tokens-next-private/fallbacks/color-fallbacks.json')
```

This only comes up as we are looking into how to structure this as a set of workspaces and a package may not be installed as a sibling to this build config.

joshblack (commented):
Left a couple of questions 👀

rollup.config.js
langermank: Its passing a CSS var as a fallback, so it does respect theme. In this case it will look like `var(--bgColor-default, var(--color-bg-default)`

rollup.config.js
langermank: Probably yes! Let me try it 😄

siddharthkp (approved):
Updated source and lockfile version :)

PR #3284, opened 2023-05-15
Refactor(ActionList): ActionList.Item should render content as a button if parent is not interactive.

radglob:
Continuing to align the ActionList semantics in PRC to match PVC. The ActionList.Item content should render an interactive button. Exceptions are when the outermost element of the ActionList.Item is an interactive element. Examples of this include NavList (where ActionList.Items can be render `as` a button) and in ActionMenu, where the top-level element is made interactive with the `menuitem` role.

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge


radglob (commented):
We did something very similar to this in https://github.com/primer/react/pull/3247. There are now a few known places where we don't want Selection components to show and they ordinarily would given our current design. We may want to refactor this in the future since it isn't a one-off issue.

src/ActionList/Item.tsx
```diff
@@ -192,17 +206,41 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
+                borderStyle: 'none',
+                backgroundColor: 'transparent',
+                cursor: 'inherit',
+                fontSize: 'inherit',
+                color: getVariantStyles(variant, disabled).color,
+                lineHeight: '20px',
+              }}
+              // @ts-ignore `as` prop may be passed to ActionList.Item, even if it isn't defined in ActionListItemProps.
+              // If this item is inside an ActionMenu, don't render an interactive button.
+              as={isTopLevelInteractive() ? 'div' : 'button'}
```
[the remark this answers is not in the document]
radglob: https://primer-124f090a4d-13348165.drafts.github.io/storybook/?path=/story/components-actionlist--link-item-playground

Any `ActionList.LinkItem` is going to have an internal div instead of a button.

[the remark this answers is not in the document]
radglob: `ActionMenu` is another example where we render a `div`. Each `<li>` has `role=menuitem`, which makes them interactive. If we had a `<button>` as the child, it breaks accessibility as we would have twice as many selectable elements. And the `<li>`s must be so since they're the direct children of a list element.

https://primer-124f090a4d-13348165.drafts.github.io/storybook/?path=/story/components-actionmenu--default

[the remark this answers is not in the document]
kendallgassner: Thank you for the examples I think I was assuming this element would become the listitem or the link -- not be the child.

lesliecdubs:
@siddharthkp would you be willing to give this a review 👀 ?

src/ActionList/Item.tsx
```diff
@@ -84,12 +84,21 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       },
     }
 
+    const isTopLevelInteractive = () =>
```
mperrotti, line 87: Is it worth checking for a `tabIndex`?

src/ActionList/Item.tsx
```diff
@@ -199,17 +212,41 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
     const containerProps = _PrivateItemWrapper ? {role: role || itemRole ? 'none' : undefined} : menuItemProps
 
-    const wrapperProps = _PrivateItemWrapper ? menuItemProps : {}
+    const wrapperProps = _PrivateItemWrapper
+      ? menuItemProps
+      : {
+          sx: {
+            display: 'flex',
+            paddingX: isTopLevelInteractive() ? 0 : 2,
+            paddingY: isTopLevelInteractive() ? 0 : '6px',
```
mperrotti, line 221: Non-blocking nitpick: we could avoid repeating ourselves if we do this instead

```suggestion
            paddingX: styles.paddingX,
            paddingY: styles.paddingY,
```

If we don't do this, I'd recommend adding a comment about '6px' being a custom value.

src/ActionList/Item.tsx
```diff
@@ -199,17 +212,41 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
+            flexGrow: 1,
+          },
+        }
 
     return (
       <ItemContext.Provider value={{variant, disabled, inlineDescriptionId, blockDescriptionId}}>
         <LiBox ref={forwardedRef} sx={merge<BetterSystemStyleObject>(styles, sxProp)} {...containerProps} {...props}>
+          {/* @ts-ignore onClick prop is only passed when _PrivateItemWrapper is set by ActionList.LinkItem. */}
```
mperrotti, line 229: Can we scope this `@ts-ignore` to the specific error? I forget if that's even possible like it is with linters...

src/ActionList/Item.tsx
mperrotti, line 87: Should we use this to throw an error (or a warning) if somebody tries to render an interactive element as a child of an ActionList.Item who's top level is interactive?

src/UnderlineNav2/UnderlineNav2.features.stories.tsx
```diff
@@ -76,7 +76,11 @@ const items: {navigation: string; icon: React.FC<IconProps>; counter?: number |
 export const OverflowTemplate = ({initialSelectedIndex = 1}: {initialSelectedIndex?: number}) => {
   const [selectedIndex, setSelectedIndex] = React.useState<number | null>(initialSelectedIndex)
   return (
-    <UnderlineNav aria-label="Repository">
+    <UnderlineNav
+      aria-label="Repository"
+      // @ts-ignore UnderlineNav does not take selectionVariant prop, but we need to pass it to the underlying ActionList so it doesn't show Selections.
```
mperrotti, line 81: Would also be nice to scope this `ts-ignore` if possible.

mperrotti (approved):

src/ActionList/Item.tsx
radglob: It won't hurt. It's not strictly speaking an expected prop, but there are a few places in dotcom that pass it in.

src/ActionList/Item.tsx
radglob: As far as I can tell, that isn't an option. It's not like eslint where you can specify.

src/ActionList/Item.tsx
radglob: I don't think so. There isn't a good way to check the parent component without passing a ref. And in general, it leads to a weird layout, so I don't think it should be much of an actual issue. Here's an example.

```
<ActionList>
  <ActionList.Item as="button">
    <ActionList.Item as="button">Testing</ActionList.Item>
  </ActionList.Item>
</ActionList>
```

And it ends up looking like this.
Screenshot 2023-06-06 at 10 00 30 AM

PR #3288, opened 2023-05-16
Updated Accessibility on ActionMenu Docs

JoshBowdenConcepts:
Describe your changes here.

Closes https://github.com/github/primer/issues/2195

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/__tests__/ActionMenu.test.tsx
+++ b/src/__tests__/ActionMenu.test.tsx
@@ -138,7 +138,9 @@ describe('ActionMenu', () => {
       </ThemeProvider>,
     )
 
-    const button = component.getAllByRole('button')[0]
+    const button = component.getByRole('button', {
+      name: 'Group by Stage',
+    })
 
     const user = userEvent.setup()
     await user.click(button)
--- a/src/__tests__/ActionMenu.test.tsx
+++ b/src/__tests__/ActionMenu.test.tsx
@@ -137,7 +137,8 @@ describe('ActionMenu', () => {
         <MixedSelection />
       </ThemeProvider>,
     )
-    const button = component.getByLabelText('Group by')
+
+	const button = component.getAllByRole('button')[0]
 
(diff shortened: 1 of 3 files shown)


src/__tests__/ActionMenu.test.tsx
```diff
@@ -137,7 +137,8 @@ describe('ActionMenu', () => {
         <MixedSelection />
       </ThemeProvider>,
     )
-    const button = component.getByLabelText('Group by')
+
+    const button = component.getAllByRole('button')[0]
```
joshblack, line 141: Is it easy to find the label text so we can keep using `getByLabelText` or not really? Would be awesome to keep `getByLabelText` here unless we are explicitly checking the ordering of the buttons. Hope that makes sense, let me know if I'm misunderstanding anything 👀

JoshBowdenConcepts: I haven't found a way to get the Button by the label text via this testing suite which is why I went to this more generic approach. You are correct that it would be better to retrieve this in another way. I am just not sure how to yet. Do you have any ideas? I have tried:

- `component.getByLabelText('Group by Stage')`

joshblack: @JoshBowdenConcepts I bet this would work:

```suggestion
    const button = component.getByRole('button', {
      name: 'Group by Stage',
    })
```

Sadly I think `getByLabelText` only works with `<label>` so we need to either `getByText` or `getByRole` with the `name` matcher if we want to also check for the button semantics.

JoshBowdenConcepts: This is really cool I have never used a specifier like that! I believe this works I will merge it in!

joshblack (approved):

PR #3296, opened 2023-05-18
refactor(project): move to npm workspaces experiment

joshblack:
Update project to use npm workspaces

#### Changelog

**New**

- Add top-level workspace which lists examples and packages as children workspaces

**Changed**

- Move primer/react under packages/react
- Update ESLint, TypeScript, Storybook, Jest, custom scripts, and more to new structure

**Removed**


package.json
```diff
@@ -1,142 +1,46 @@
...
-    "history": "^5.0.0",
-    "react-intersection-observer": "9.4.3",
-    "styled-system": "^5.1.5"
+    "test:type-check": "tsc --noEmit"
   },
   "devDependencies": {
-    "@actions/core": "1.10.0",
-    "@babel/cli": "7.21.0",
-    "@babel/core": "7.21.0",
+    "@babel/core": "7.21.8",
```
broccolinisoup, line 36: Do we keep the dependencies that are going to be used across packages on the root package.json?

examples/nextjs/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "@primer/octicons-react": "^18.2.0",
-    "@primer/react": "file:../../",
+    "@primer/react": "^35.21.1",
```
broccolinisoup, line 12: I am not sure if I am reading this right so let me know if I misunderstand this change. 

I rely on so much building the react locally and immediately make it available to the nextjs application so that I can debug and resolve issues as I go. As far as I understand keeping a stable version here won't allow this? If I am reading this right, is there a way to keep this practicality and convenience?

broccolinisoup (commented):
Thank you so much for this PR @joshblack !!

I pulled the branch and removed all node_modules (root and under examples) and then run `npm i` it worked without an issue and I am able to run the storybook 🎉 

I also went into examples/nextjs and tried to run the app with `npm run develop` but it throws an error and complaining not finding the `@primer/react` and I can confirm that they are not under the nextjs/node_modules folder.

```
..react/examples/nextjs$ npm run develop

> example-nextjs@0.0.0 develop
> next

ready - started server on 0.0.0.0:3000, url: http://localhost:3000
warn  - You have enabled experimental feature (esmExternals) in next.config.js.
warn  - Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use at your own risk.

error - ./src/components/Layout.js:3:0
Module not found: Can't resolve '@primer/react'
  1 | import React from 'react'
  2 | import Navigation from './Navigation'
> 3 | import {Box} from '@primer/react'
  4 |
  5 | export default function Layout({children}) {
  6 |   return (

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./src/pages/_app.js
```
Any idea what might be going on here?

Thanks so much again, excited to try out this npm workspaces 🤩

package.json
joshblack: @broccolinisoup that would definitely be the rule of thumb I'd love to follow. Basically any workspace should list the dependencies that it needs in order to run function. That can be in a child workspace like `@primer/react` or can be at the top-level if we have scripts there that depend on those dependencies.

joshblack:
@broccolinisoup I think you'll need to run `npm run build` for the Next.js example to pick up on the primer/react package. By default, it will try and resolve the `exports` in that `package.json` but they don't exist when first pulling down the repo.

package.json
```diff
@@ -1,142 +1,46 @@
...
     "lint:fix": "npm run lint -- --fix",
-    "lint:md": "markdownlint-cli2 \"**/*.{md,mdx}\" \"!.github\" \"!.changeset\" \"!node_modules\" \"!CHANGELOG.md\"",
+    "lint:md": "markdownlint-cli2 \"**/*.{md,mdx}\" \"!.github\" \"!.changeset\" \"!**/node_modules/**\" \"!**/CHANGELOG.md\"",
+    "prepare": "husky install",
+    "release": "npm run build && changeset publish",
+    "setup": "./script/setup",
+    "size": "size-limit",
+    "start": "concurrently npm:start:*",
+    "start:docs": "cd docs && npm run develop",
+    "start:storybook": "STORYBOOK=true start-storybook -p 6006",
```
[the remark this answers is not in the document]
joshblack: @broccolinisoup Here are two formats that I think of for targeting workspace commands:

- Specify the workspace directly with `npm`: `npm start -w example-nextjs` (this also works with multiple `-w` flags)
- Use a tool like `nx` to filter workspaces

Do you think either of these options could work? 👀

package.json
broccolinisoup: Oh right right! Make sense.

package.json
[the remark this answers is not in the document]
broccolinisoup: Yeah! I would prefer the first one with just `npm`

examples/nextjs/package.json
broccolinisoup: >This version here will automatically resolve to the packages/react folder because node_modules/@primer/react points to that workspace.

🤯  It might be just me but I wouldn't assume that it will resolve. The version number misled me. Thanks for explaining that.

broccolinisoup:
Oh why did we close this PR? 😢

PR #3312, opened 2023-05-22
SelectPanel: Add filter input label and description

colebemis:
[Storybook preview](https://primer-373fec8c4e-13348165.drafts.github.io/storybook/iframe.html?args=&id=components-selectpanel--default&viewMode=story)

- Adds optional `inputLabel` prop that will set the `aria-label` of the filter input. By default, `inputLabel` is the same as the placeholder text.
- Makes the `placeholderText` prop optional.
- Adds an ARIA description to the filter input.
- Displays a search icon in the filter input.

Part of https://github.com/github/primer/issues/2087

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/SelectPanel/SelectPanel.docs.json
+++ b/src/SelectPanel/SelectPanel.docs.json
@@ -19,6 +19,12 @@
       "description": ""
     },
     {
+      "name": "inputLabel",
+      "type": "string",
+      "defaultValue": "Same as placeholderText",
+      "description": "The aria-label for the filter input"
+    }
+    {
       "name": "overlayProps",
       "type": "Partial<OverlayProps>",
       "defaultValue": "",
--- a/src/SelectPanel/SelectPanel.docs.json
+++ b/src/SelectPanel/SelectPanel.docs.json
@@ -23,7 +23,7 @@
       "type": "string",
       "defaultValue": "Same as placeholderText",
       "description": "The aria-label for the filter input"
-    }
+    },
     {
       "name": "overlayProps",
       "type": "Partial<OverlayProps>",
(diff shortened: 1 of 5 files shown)


src/SelectPanel/SelectPanel.docs.json
```diff
@@ -18,6 +18,12 @@
       "defaultValue": "",
       "description": ""
     },
+    {
+      "name": "inputLabel",
+      "type": "string",
+      "defaultValue": "Same as placeholderText",
+      "description": "The aria-label for the filter input"
+    }
```
colebemis, line 26:
```suggestion
    },
```

src/SelectPanel/SelectPanel.tsx
```diff
@@ -141,9 +146,11 @@ export function SelectPanel({
     return {
       sx: {m: 2},
       contrast: true,
+      leadingVisual: SearchIcon,
+      'aria-label': inputLabel,
```
colebemis, line 150: :accessibility: A11y question: Should we use a visually hidden label instead of aria-label for the filter input?

TylerJDev: I'd say it doesn't matter as much as they both will announce the same. I'd say one benefit that you could potentially gain with a visually hidden label and `aria-labelledby` is that it _might_ translate better with automated translation tools versus an `aria-label`, but I believe support has gotten better since I last checked. Other than that, I think using `aria-label` is fine.

src/SelectPanel/SelectPanel.tsx
```diff
@@ -53,6 +56,8 @@ export function SelectPanel({
   renderAnchor = props => <DropdownButton {...props} />,
   anchorRef: externalAnchorRef,
   placeholder,
```
TylerJDev, line 58: Question, is this needed with `placeholderText` being available?

TylerJDev (approved):
This looks good! If we ever opted for a visual label, we could forgo the `placeholder`, but for now I think this is great!

src/SelectPanel/SelectPanel.tsx
colebemis: The naming is confusing here. It looks like `placeholder` is what gets displayed when there are no results for a filter query. `placeholderText` is the placeholder for the filter input 😅

TylerJDev: Oh I see 😆, then yeah we should be good!

PR #3317, opened 2023-05-23
Accessibility engineering review for Link component.

radglob:
Restricting the `as` prop in Link components. Leaving forcing underlining on links in text to more holistic changes.

Closes https://github.com/github/primer/issues/1881

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge


docs/content/Link.mdx
```diff
@@ -10,10 +10,6 @@ import data from '../../src/Link/Link.docs.json'
 
 The Link component styles anchor tags with default hyperlink color cues and hover text decoration. `Link` is used for destinations, or moving from one page to another.
 
-In special cases where you'd like a `<button>` styled like a `Link`, use `<Link as='button'>`. Make sure to provide a click handler with `onClick`.
-
-**Important:** When using the `as` prop, be sure to always render an accessible element type, like `a`, `button`, `input`, or `summary`.
-
```
TylerJDev, line 16: With this change, could we update the [props table](https://primer-208e588ead-13348165.drafts.github.io/Link#link) in the docs and remove the `as` prop entirely? I believe this is in `Link.docs.json`. We might also need to do the same with `LinkItem` since we're removing usage of `as` there too.

TylerJDev (commented):
Just a few comments. Everything looks great overall!

src/Link/Link.tsx
```diff
@@ -23,55 +23,27 @@ const hoverColor = system({
...
-    user-select: none;
-    background-color: transparent;
-    border: 0;
-    appearance: none;
-  }
   ${sx};
 `
 
-const Link = forwardRef(({as: Component = 'a', ...props}, forwardedRef) => {
+const Link = forwardRef(({as, ...props}, forwardedRef) => {
```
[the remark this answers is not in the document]
radglob: Yeah, we don't want to break usage immediately, instead showing a clear path forward for any code that is now breaking convention. If we just removed support for `as` altogether, I think we would break dotcom (and maybe some other stuff) in small ways.

docs/content/Link.mdx
radglob: Good call.

lesliecdubs:
Hey @radglob, is this one still awaiting review? Let's re-request from Tyler if needed and I can help nudge other reviewers if that's what you're waiting on.

docs/content/Link.mdx
```diff
@@ -10,10 +10,6 @@ import data from '../../src/Link/Link.docs.json'
 
 The Link component styles anchor tags with default hyperlink color cues and hover text decoration. `Link` is used for destinations, or moving from one page to another.
 
-In special cases where you'd like a `<button>` styled like a `Link`, use `<Link as='button'>`. Make sure to provide a click handler with `onClick`.
-
-**Important:** When using the `as` prop, be sure to always render an accessible element type, like `a`, `button`, `input`, or `summary`.
```
langermank, line 15: Is this a breaking change, removing `as`?

src/BaseStyles.tsx
```diff
@@ -26,6 +26,10 @@ const GlobalStyle = createGlobalStyle<{colorScheme?: 'light' | 'dark'}>`
   details-dialog:focus:not(:focus-visible):not(.focus-visible) {
     outline: none;
   }
+
+  p a[class*=StyledLink] {
```
langermank, line 30: Is there a way you could test this in some of our React apps? I like the idea of it, but I feel like this could easily cause unexpected visual changes. Underlined in a body of text probably won't ruffle any feathers, but there's probably some cases where its not a body of text and more like one line of text as part of a UI.

langermank (commented):
I think we should add a `link` variant to `Button` as we have in PVC to accommodate this change. I know its not ideal but it was mentioned in the audit issue and seems like a decent stop-gap before disallowing that behavior entirely.

src/BaseStyles.tsx
mperrotti: I agree with @langermank , but also have no idea where to start. @colebemis - maybe you can help?

lesliecdubs:
@TylerJDev could you help out with a review on this one? Thanks!

src/BaseStyles.tsx
TylerJDev: I think if this is a blocker then we could wait until we receive more guidance on `Link` and underline usage across Primer. I know this is the case for the PVC `Link` accessibility fixes, so if this is required here too then I think it's okay to come back to this.

If we wanted to test this we'd probably have to query up the usage of PRC `Link` across GitHub and find the instances where this component is a descendant of a paragraph element, though I would love to know if there are better ways to go about this. From a very quick look, this seems like an uncommon pairing, so there might not be too much friction, but that's only from a small sample size.

TylerJDev (approved):
Looks good to me! I think there's some [pending discussion](https://github.com/primer/react/pull/3317#pullrequestreview-1463875996) that'll need to be addressed first, but other than that I think that this is great!

src/BaseStyles.tsx
lesliecdubs: Agree with @TylerJDev that this may be better solved more holistically through upcoming work to determine underline styles for links in https://github.com/github/primer/issues/2099.

radglob: Rolled back the changes around styling links in text. This way, we can merge in the `as` restriction, and let the more holistic work handle this issue.

src/Button/types.ts
```diff
@@ -8,7 +8,7 @@ export const StyledButton = styled.button<SxProp>`
   ${sx};
 `
 
-export type VariantType = 'default' | 'primary' | 'invisible' | 'danger' | 'outline'
+export type VariantType = 'default' | 'primary' | 'invisible' | 'danger' | 'outline' | 'link'
```
mperrotti, line 11: Is there any visual difference between `'invisible'` and `'link'`?

langermank: @mperrotti yes, it has so height/padding and "link like" hover styles (underline). Its a button pretending to be a link, while `invisible` has buttony sizing, hover states, and is sometimes a muted color.

langermank (approved):

PR #3320, opened 2023-05-24
Remove aria-hidden=true from spans with required asterisk

adrianababakanian:
Describe your changes here.

Per accessibility feedback, removes `aria-hidden="true"` from the `span`s that add the `*` to `FormControl`s marked with the `required` attribute.

It's worth noting that the audit feedback was specific to an `ActionMenu` form control, which is relevant because it uses a `button` rather than an actual form element like `select` and hence can't actually have a required attribute. I wonder if we want to be more selective about which `<span>*</span>`s we want to remove `aria-hidden="true"` from, or if it's acceptable to remove it universally as currently proposed here?

### Screenshots

Before:
Screenshot 2023-05-24 at 1 03 33 PM

After:
Screenshot 2023-05-24 at 1 02 56 PM

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/pink-beds-fetch.md
+++ b/.changeset/pink-beds-fetch.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Remove `aria-hidden=true` from `span`s with required asterisk
--- a/src/_InputLabel.tsx
+++ b/src/_InputLabel.tsx
@@ -55,7 +55,7 @@ const InputLabel: React.FC<React.PropsWithChildren<Props>> = ({
       {required ? (
         <Box display="flex" as="span">
           <Box mr={1}>{children}</Box>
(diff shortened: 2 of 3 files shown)


TylerJDev (approved):
Love it! Thanks for fixing this! ✨ 

> I wonder if we want to be more selective about which <span>*</span>s we want to remove aria-hidden="true" from, or if it's acceptable to remove it universally as currently proposed here?

I think it's perfectly fine doing it universally here. I think in all cases we don't want the asterisk to be hidden.

colebemis (approved):
Thank you ✨

PR #3342, opened 2023-05-30
Update release schedule to run every Monday

joshblack:
Updates our Release Schedule workflow to run every Monday at 13:30 UTC. This should automate this workflow instead of it needing a manual dispatch every week.

For the timing, it seemed like our Pagerduty schedule switches over at 8:30am CT so I converted that time to UTC for the cron schedule. Hope it makes sense 😅


broccolinisoup (approved):
Changes makes a lot of sense! 🚀

PR #3347, opened 2023-05-31
Add type-safety for classnames in vs code

siddharthkp:
Added type-safety for classnames with [typescript-plugin-css-modules](https://github.com/mrmckeb/typescript-plugin-css-modules) in the editor.

Unfortunately, tsc doesn't load plugins yet, so we don't get typechecks for classnames in CI and are stuck with declaring loose definitions in `global.d.ts` for now.

screenshot of code with typescript error for a misspelt classname

---

Note: Make sure your editor is using the workspace version of typescript to use the plugin. [See typescript docs](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript)

screenshot of editor which shows the current typescript version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -166,6 +166,7 @@
         "ts-node": "10.9.1",
         "ts-toolbelt": "9.6.0",
         "typescript": "4.9.5",
+        "typescript-plugin-css-modules": "5.0.1",
         "unist-util-find": "1.0.2",
         "unist-util-find-before": "3.0.1",
         "unist-util-flat-filter": "2.0.0",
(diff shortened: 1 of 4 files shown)


joshblack (approved):
Very cool! 🥳 For the type error, do we need something like: https://www.npmjs.com/package/typescript-plugin-css-modules#user-content-custom-definitions or is what you had before in `global.d.ts` good to use? 👀

siddharthkp:
> Very cool! 🥳 For the type error, do we need something like: https://www.npmjs.com/package/typescript-plugin-css-modules#user-content-custom-definitions or is what you had before in global.d.ts good to use? 👀

We do :( 

I was hoping we wouldn't only to realise [tsc doesn't load plugins](https://github.com/mrmckeb/typescript-plugin-css-modules/issues/57#issuecomment-544832759)

PR #3349, opened 2023-05-31
Revert "Refactor FilteredActionList to address a11y violations and use new ActionList."

joshblack:
Reverts primer/react#3247

When going through the release flow this week, this change seemed to not be compatible with teams that are still using the deprecated `ItemProps` from `ActionList`. Some of the changes that were detected included:

- The `id` type being narrowed from `number | string` to `string`
- The `leadingVisual` and `trailingVisual` types being changed from `React.FC` to `JSX.Element` and `string`, respectively
- `onSelectionChange` type change
- `selected` type change

It seems like the best next steps for this work would be to figure out how to add a compatibility layer between old and new and make sure we have test coverage to support this for `SelectPanel` and anything else that incidentally depends on `ItemProps`


broccolinisoup (approved):

broccolinisoup:
Such a bummer that we have to revert this but it makes sense. I approved mainly to get into the release tracking branch to see how the integration tests are going to look like after reverting this.

PR #3352, opened 2023-05-31
replace `@changesets/cli` with `primer-changesets-cli`

gr2m:
This is part of an effort to provide a notification system to users of Primer releases.

To test the new `changeset` CLI

- Checkout this pull request
- `npm install`
- `npx changeset`

Before this will work, we will need to update https://github.com/primer/.github/blob/main/.github/workflows/release.yml. Turns out the `changesets/action` requires `@changesets/cli` to be installed ... so we will need to install it explicitly like it's done in https://github.com/primer/view_components/pull/2038/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34

### Screenshots

https://github.com/primer/react/assets/39992/bf795016-570d-4462-9cde-e8e423e3d964

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack:
One quick question, how will this impact the generated changelogs that show up under the tags for releases? For example: https://github.com/primer/react/releases/tag/v35.25.1

Will each change end up including the "Changed components" section that is being added to the changeset?

gr2m:
> Will each change end up including the "Changed components" section that is being added to the changeset?

Yes, that's the idea. Then consumers can use https://github.com/gr2m/primer-release-filter-action to filter out the changelogs to only the parts they are interested in.

gr2m:
> @gr2m would it be possible for the components changed to be metadata that is hidden from things like the release changelog? This would be so that it could still be present and used in tooling but wouldn't impact the existing format that the changesets action generates. Just wanted to ask and see how do-able that seemed to you 👀

I don't know how. Do you think the changelogs might become to noisy?

Maybe there could be a way to add hidden information to release notes using HTML comments, similar like what is commonly done with issue bodies and comments. But that would require even more changes to the existing flow.

I would suggest we go with the changes as they are and iterate if necessary? Worst case we can roll back the changes and try again.

Probably the proper way to solve this would be to have separate releases for each component, either with a monorepo setup or with separate repositories (like we do in @octokit), but that would be a bigger undergoing and has its own trade offs.

joshblack:
> I don't know how. Do you think the changelogs might become to noisy?

@gr2m that was my first thought with the update. I'm not sure how folk would ready it but let me bring this up to the other maintainers to get their take on it! It'd be great if we could make these changes hidden/transparent while we're trying things out in order to avoid changes to existing processes.

gr2m:
> It'd be great if we could make these changes hidden/transparent while we're trying things out in order to avoid changes to existing processes.

We cannot hide it, it's either we go ahead with it or we don't at this point.

lesliecdubs:
@gr2m I'd like to check in with the PRC maintainers one more time on this. We'll aim to get back to you by Monday at the latest.

gr2m:
I'm happy to join that discussion and answer all questions that might be left @lesliecdubs

lesliecdubs:
@gr2m thanks for joining the [PRC maintainers sync on Monday](https://docs.google.com/document/d/1XROq7mBumWcKlSg1_Acj3FXZrhDN1oD4ekL3xv7Hrlw/edit#heading=h.dg4ntq94fx81) to discuss! Following that and a subsequent discussion I had with @joshblack, we are good to continue forward 👍🏻 

As discussed live, expectations for next steps would be:
- Let's "trial" the [release notification setup](https://github.com/github/primer/issues/2318) with one team (e.g. Memex) to start to get feedback on usefulness. If there is desire to expand further, let's discuss before encouraging broader subscription.
- Let's treat this trial as [an experiment](https://www.reforge.com/blog/good-experiment-bad-experiment) by documenting the hypothesis (potentially pulled from [the pitch](https://github.com/github/planning-tracking/issues/1977)), defining success, gathering feedback from the subscribing team, and setting a deadline to determine experiment outcome. This will help us ensure the [trade-offs for maintainers](https://github.com/github/primer/issues/2318#issuecomment-1597441633) serve a real user need and are worth investing in long-term.

Let me know if you have any questions or thoughts on any of the above!

---

@primer/react-reviewers could someone in the reviewers group please give this PR a review, with the understanding that we'll treat this as an experiment as described above?

colebemis:
Have we tried adding the metadata as a comment in the changesets markdown file as a way to hide it from the rendered release notes?

```
<!-- ... -->
```

gr2m:
> Have we tried adding the metadata as a comment in the changesets markdown file as a way to hide it from the rendered release notes?
> 
> ```
> 
> ```

No but I can try that out

gr2m:
added CI: https://github.com/gr2m/primer-release-changesets-validator-action

gr2m:
This should be ready to be merged

gr2m:
I resolved the remaining failing test. Btw Node 17 is no longer maintained since June 1st. You might want to test in 19 and 20 instead

gr2m:
Hmm I'm getting all kind of errors now that are unrelated to this pull request, I'll see if I can get around it, but you should do a manual pull request and re-create the lock file, I think you'll have to update snapshots

gr2m:
This should be ready to be merged, for real this time 🙏🏼

joshblack (approved):
🥳

PR #3364, opened 2023-06-05
Fix warning on ReactDOM.render by using React 18 client API.

### LINKED ISSUE #2990: `useConfirm` calls deprecated `ReactDOM.render`, triggering warning in React 18
`useConfirm` uses `ReactDOM.render` to inject the confirmation dialog into the HTML. 

https://github.com/primer/react/blob/72f0ec67a92185572ee75c1fcd7a686b6c23b2e5/src/Dialog/ConfirmationDialog.tsx#L161-L168

This triggers a warning in the console when in development mode:

> Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot

While the warning states that the whole app will behave like it's running React 17, this isn't really the case - only the dialog will. This is because the call to `render` creates a whole new React app instance which only contains the dialog. However, this still isn't great and we definitely should resolve this.

---

The obvious solution is to just use `createRoot` instead, but it seems like there must be a more elegant alternative here than to create an entirely new React application. What if we could come up with a way to extract the `Portal` component into a `usePortal` hook, allowing rendering arbitary JSX into the default portal from any hook? Something like this:

```tsx
function useConfirm() {
  const renderIntoPortal = usePortal()

  return () => new Promise(_resolve => {
    const resolve = (result) => {
      _resolve(result)
      renderIntoPortal(null)
    }
    renderIntoPortal(
      <ConfirmationDialog onClose={resolve(false)}>
        <Button onClick={resolve(true)}>Confirm</Button>
      </ConfirmationDialog>
    )
  })
}
```

I'm not sure if this is actually technically possible though - maybe I'll try and come up with a POC this week.

radglob:
Updated `ConfirmationDialog` to use React 18's `createRoot` API instead of `ReactDOM.render`.

Closes #2990.

We may want to look into using the Portal API to avoid rendering a separate React app in the future.

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge


joshblack:
@radglob could you target the `next-major` branch for this when you get the chance? Sadly the `react-dom/client` stuff won't work with v17 so we have to have a hard change to v18 for it 😞

radglob:
> @radglob could you target the `next-major` branch for this when you get the chance? Sadly the `react-dom/client` stuff won't work with v17 so we have to have a hard change to v18 for it 😞

Ah, yeah, sure thing.

radglob:
Need to target `next-major`, will do so in a new branch.

PR #3387, opened 2023-06-07
Refactor Autocomplete to use the newest ActionList

mperrotti:
The Autocomplete API will remain largely the same, but it no longer uses the deprecated ActionList.

The only API differences are:
- `aria-labelledby` is now required on `Autocomplete.Menu`
- The `items` passed to `Autcomplete.Menu` no longer accept numbers for IDs, only strings

I was able to avoid any other breaking changes 🙂 

Closes https://github.com/github/primer/issues/523

### Screenshots

No visual changes

### Merge checklist

- [x] Added/updated tests
- [idk] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
I broke something at some point, but I'm not sure what. I'll post when I fix it. 😅

mperrotti:
Fixed!

siddharthkp (commented):
#### API changes: 

> - `aria-labelledby` is now required on `Autocomplete.Menu`

makes total sense!

> - The `items` passed to `Autcomplete.Menu` no longer accept numbers for IDs, only strings

would it help if we support both numbers and string?

Do we know if there are any other breaking changes based on the usage in dotcom? Can use this new action to test that faster now, thanks to @broccolinisoup 🙌  https://github.com/github/github/pull/277983

#### Implementation details

I trust you with the changes, is there anything specific that you want review/feedback on?

mperrotti:
@siddharthkp 

> Is there anything specific that you want review/feedback on?

Nothing in particular. I just want to confirm that these changes are at least as good as the previous `Autocomplete` implementation. If everything looks good, then we can merge and get closer to shipping v36.

---

> would it help if we support both numbers and string?

I don't think so. It's more consistent with our other component APIs to just accept a string for an ID.

---

> Do we know if there are any other breaking changes based on the usage in dotcom?

Not that I can think of.

siddharthkp (approved):
Feels pretty safe honestly. Ship it!

(can be done in a new PR after merging this) It might be helpful to add visual regression tests for Autocomplete in `main` so that we have baseline to make sure this change has only intentional visual changes

Issue #2773, opened 2023-01-13
Pagination enhancement: support responsive values for the `showPages` prop


mperrotti:
Currently, the "[previous/next pagination](https://primer.style/react/Pagination#previousnext-pagination)" variant is boolean. It would be useful if we could choose to use that variant for specific viewport ranges.

This came up while working on the [DataTable](https://primer.style/design/components/data-table). We'd like to hide the page numbers on smaller viewports.

Example:
<img width="664" alt="Screenshot 2023-01-13 at 3 47 09 PM" src="https://user-images.githubusercontent.com/2313998/212416223-3b76a58c-0b1d-4d18-9d5c-403a8da47e4f.png">

PR #3390, opened 2023-06-07
Make the `showPages` prop on our Pagination components responsive

mperrotti:
The `showPages` prop on both Pagination components can now accept a responsive value.

Closes https://github.com/primer/react/issues/2773

### Screenshots

https://github.com/primer/react/assets/2313998/1942a266-0c39-45d3-a60f-8ef07eaa8f53

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
@joshblack - instead of building it in, I wanted to give users the option to change the value for different viewport sizes. For example, I might be using pagination in a narrow sidebar on wide screens, so I'd want to set `showPages={{wide: false}}`

joshblack:
@mperrotti would this be a good case for container queries where we could make that decision based on the container size? The ideal here seems to be that we should show the pages if there is space and hide them if there is not and that could be a way of baking that in without having to customize it / tie it to media queries if the width isn't really tied to the viewport width.

mperrotti:
@joshblack - I agree that this could be a great use for container queries. Do you think I should refactor before we merge this?

joshblack:
@mperrotti if we end up using container queries, would it be that `showPages` would still be a `prop` or would this be something that is handled automatically? (only show pages if there is space)

I think if, with container queries, this behavior becomes automatic that'd be a huge win but if it's just a different way to implement `showPages` then I totally would get thinking of this as an implementation detail.

mperrotti:
> I think if, with container queries, this behavior becomes automatic that'd be a huge win

Great idea! I think it makes sense to just make this behavior automatic.

mperrotti:
@joshblack - `styled-components` won't support container queries until v6 🤬

https://github.com/styled-components/styled-components/issues/3911

Should we:
1. hold off on this until we upgrade to styled-components v6
2. merge as-is and make a breaking API change later if we still think it's better to automatically hide the page numbers for containers of a certain size
3. merge as-is and plan to keep this new `showPages` API
4. something else

There are trade-offs to all of those options, but I think **option 2** is looking the best to me right now. If you have any other ideas about how we might work around this, let me know!

src/DataTable/Pagination.tsx
```diff
@@ -202,7 +238,7 @@ export function Pagination({
             </Button>
           </Step>
           {pageCount > 0 ? (
-            <Step>
+            <Step hiddenViewportRanges={getViewportRangesToHidePages()}>
```
joshblack, line 241: Is the `hiddenViewportRanges` prop something that can be placed on `TablePaginationSteps`? Would be great to calculate this once and place it on the outermost node and then target it e.g.

```css
  @media ${viewportRanges.narrow} {
    .TablePaginationSteps[data-hidden-viewport-ranges*='narrow'] > *:not(:first-child):not(:last-child) {
      display: none;
    }

    .TablePaginationSteps[data-hidden-viewport-ranges*='narrow'] > *:first-child {
      margin-inline-end: 0;
    }
  }
```

If not, I think this is something we should calculate once instead of as a function and just pass its value to each `Step` instead of messing around with `useCallback()` and calling it for each step.

joshblack (commented):
With the stuff from container queries figured moving forward with option 2 works like you suggested would work! Just had one question on the implementation, curious what you think 👀

src/DataTable/Pagination.tsx
mperrotti: Ahh good catch! Ok, I'll refactor 👍

langermank (approved):

joshblack (approved):
🥳

PR #3392, opened 2023-06-08
[WIP/Prototype]: NewToken component

lukasoppermann:
Describe your changes here.

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
We have a major release coming up. Instead of adding a new token component, we could just modify the existing one.

lukasoppermann:
> We have a major release coming up. Instead of adding a new token component, we could just modify the existing one.

yes, that's the plan. I just want to keep the old stuff around for now, while I am prototyping. 

The goal is to merge this with the current tokens in one way or another.

maximedegreve:
I've noticed that in dark mode the contrast isn't enough for background in some instances. 
For example the purple here.

lukasoppermann:
Not stale

src/Token/NewToken.tsx
```diff
@@ -0,0 +1,177 @@
+import {CSSObject} from '@styled-system/css'
```
langermank, line 1: I think you should move this to the `drafts` directory and call it `IssueLabelToken` since that seems to be the component this would replace

langermank: https://github.com/primer/react/tree/main/src/drafts

src/Token/componentTokens.css
```diff
@@ -0,0 +1,127 @@
+/**/
+[data-color-mode='light'][data-light-theme^='light'],
```
langermank, line 2: Let's just add this to the new Primitives build, I'm not following why you want to avoid doing that? Don't worry about adding it to the old data set, just add it to the new build. I think we can prioritize dotcom for this component since its incredibly product specific, which has access to Primitives CSS variables.

We're working on adding that import to React's Storybook too, it's just slightly complicated. I can try and prioritize that if I have a better sense of your timeline with this PR.

lukasoppermann: I'd love this. The only issue I am having is that we don't know when this is shipped. I am trying to remove interdependencies, so if I add it to v8 I need v8 to be shipped to use it right?

PR #3393, opened 2023-06-09
UnderlineNav2: Remove `selectionVariant` prop from stories and refactor accordingly

broccolinisoup:
Reverting some changes back on `UnderlineNav` that was introduced https://github.com/primer/react/pull/2878

We need to pass `selectionVariant="single"` on  [the ActionList in UnderlineNav](https://github.com/primer/react/blob/main/src/UnderlineNav2/UnderlineNav.tsx#L369) because we call `onSelect` function. If we don't set it, it throws an error when an item is selected ([ActionList ref](https://github.com/primer/react/blob/main/src/ActionList/Selection.tsx#L18))

To be able to hide the `CheckIcon` we can use styling. This is how it was originally done but I took this as an opportunity and update the styling to check a data component rather than relying on the node order in the dom.  

### Screenshots

No visual changes

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Visuals.tsx
+++ b/src/ActionList/Visuals.tsx
@@ -9,6 +9,7 @@ type VisualProps = SxProp & React.HTMLAttributes<HTMLSpanElement>
 export const LeadingVisualContainer: React.FC<React.PropsWithChildren<VisualProps>> = ({sx = {}, ...props}) => {
   return (
     <Box
+      data-component="ActionList.LeadingVisual"
       as="span"
       sx={merge(
         {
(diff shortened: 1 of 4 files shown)


joshblack (approved):

mperrotti (approved):

broccolinisoup:
Because [the PR](https://github.com/primer/react/pull/2878) that introduced the problem that we are addressing here, [is reverted](https://github.com/primer/react/pull/3408) I am closing this PR, and it is already been fixed in the [new draft PR](https://github.com/primer/react/pull/3436) that re-introduces the changes.

Issue #1298, opened 2021-06-11
Update `styled-components` dependencies


dgreif:
I recently updated most of the base-level dependencies (https://github.com/primer/components/pull/1292).  One that I elected to skip was `styled-components`, which has a [major release with big perf improvements](https://medium.com/styled-components/announcing-styled-components-v5-beast-mode-389747abd987).  I tried updating, but most of the snapshots needed to be updated and there were other issues with the tests.  I don't think the changes are actually significantly breaking, but this bump should be handled separate from other dependency bumps so that it gets the proper attention it needs.

moatorres:
Hey @dgreif, 

First of all, thank you for making Primer so awesome! Primer ❤️ 

I've been experimenting with `@primer/components` for the past few days and I've found a weird bug (possibly related to `ThemeProvider` component) that creates unexpected behaviors (screenshot attached). 

In this particular case it's messing with how the styles are applied when switching between color modes. The `selected` `SideNav.Link` and the `color` of all items are "freezing", as well as all `:hover` styles not being applied they should.

### TL;DR

I strongly advise against updating `styled-components` to anything above `5.1.1`. I'd also suggest to change primer's `peerDependencies` as below.

```js
"peerDependencies": {
    "react": "^17.0.0",
    "react-dom": "^17.0.0",
    "styled-components": "4.x || 5.1.x" // or even "^5.1.x"
 }
```

<details><summary>Steps to reproduce</summary>

1. Update `styled-components` in `package.json`
```js
"devDependencies": {
   ...
  "styled-components": "5.2.0" // or any version above
}
```
2. Run `npm run playroom:start`
3. Add the snippet below (from [primer.style](https://primer.style/components/SideNav) examples)
```jsx
<ThemeProvider colorMode="night">
  <SideNav bordered maxWidth={360} aria-label="Main" m={4}>
    <SideNav.Link href="#account">
      <Text>Account</Text>
    </SideNav.Link>
    <SideNav.Link href="#home" selected>
      <Text>Profile</Text>
    </SideNav.Link>
    <SideNav.Link href="#emails">
      <Text>Emails</Text>
    </SideNav.Link>
    <SideNav.Link href="#notifications">
      <Text>Notifications</Text>
    </SideNav.Link>
  </SideNav>
</ThemeProvider>
```
4. Change `colorMode="night"` to `colorMode="day"`
[...]

PR #3397, opened 2023-06-09
Upgrade styled-components to v5

mperrotti:
### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/__tests__/__snapshots__/Box.test.tsx.snap
```diff
@@ -49,6 +49,18 @@ exports[`Box renders margin 3`] = `
...
+
+}
+
+@media screen and (min-width:768px) {
+
+}
+
+@media screen and (min-width:1012px) {
+
+}
```
mperrotti, line 62: Not sure what's causing these to get output

.changeset/proud-rules-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
```
joshblack, line 2:
```suggestion
'@primer/react': major
```

joshblack (commented):
Looks great to me 🥳 What would the best next steps for this be? Would it be going through and doing a visual inspection to make sure nothing was impacted?

.changeset/proud-rules-develop.md
mperrotti: @joshblack - I didn't realize this was a breaking change. Should I switch the branch to `next-major` and convert from a "draft" PR?

mperrotti:
@joshblack - I already went through a bunch of stuff in Storybook, but maybe we could visually review in dotcom

src/__tests__/__snapshots__/Box.test.tsx.snap
broccolinisoup: Seems like this is a know issue both [for us](https://github.com/primer/react/issues/1298) and for [the community](https://github.com/styled-components/jest-styled-components/issues/276) 

That said since we are moving away from jest snapshots ([ADR reference](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-011-snapshot-tests.md#decision)), if this is only snapshot issue (seems like it) we might maybe want to start deleting these snapshots? At least we can start with Box maybe? And if needed we can create a test in Playwright. What do you think?

mperrotti:
> if this is only snapshot issue (seems like it) we might maybe want to start deleting these snapshots? At least we can start with Box maybe? And if needed we can create a test in Playwright. What do you think?

@broccolinisoup - I think we can just delete that snapshot

joshblack (approved):
🥳

broccolinisoup:
> > if this is only snapshot issue (seems like it) we might maybe want to start deleting these snapshots? At least we can start with Box maybe? And if needed we can create a test in Playwright. What do you think?
> 
> @broccolinisoup - I think we can just delete that snapshot

I think so too!

broccolinisoup (approved):
🥳

PR #3398, opened 2023-06-09
chore(docs): update docs to React 18

joshblack:
With our move to React 18 in v36, this PR updates the docs site to use React 18. Currently, this work is targeting `main` and this could then be merged into `next-major`. However, if folks feel more confident putting it in `next-major` directly just let me know! Happy to change the base branch 👀 

### Testing & Reviewing

- This makes changes to React in our docs site so I'd appreciate any and all help testing out the deploy preview for the docs page here to make sure no regressions occur 🙏


radglob:
Glad it was only a few small changes. I think it probably makes sense to just merge this into `next-major`. I don't think there's any pressing reason for it to go into `main`, and there aren't compatibility issues there as far as I'm aware.

PR #2108, opened 2022-06-01
Basic SegmentedControl functionality

mperrotti:
Component tracking issue: https://github.com/github/primer/issues/701
Component design issue: https://github.com/github/primer/issues/761
Component React implementation issue: https://github.com/github/primer/issues/834

Implements basic `SegmentedControl` functionality. More PRs will be opened to address the `TODO` comments in these files and add the remaining functionality from the design.

[A11y design feedback](https://github.com/github/primer/issues/762) is still being worked through.

### Screenshots

Screen Shot 2022-06-01 at 6 02 16 PM
Screen Shot 2022-06-01 at 6 02 30 PM
Screen Shot 2022-06-01 at 6 02 45 PM

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3410, opened 2023-06-13
Convert @primer/react to ESM, don't generate CommonJS bundle.

radglob:
Changed some configuration files to use `.cjs` extension, others to use `.mjs` extension, and remove settings to produce CommonJS bundle.

There are likely a lot more changes we could make (ex. only using `import` instead of `require` in some places) but this seems to produce an ESM bundle and continues to work correctly in a toy project.

Closes [#2108](https://github.com/github/primer/issues/2108).

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


package.json
joshblack, line 1: Could we include a `"type": "module"` in `package.json` in addition to the bundle changes below?

package.json
```diff
@@ -6,29 +6,21 @@
...
+      "import": "./lib-esm/drafts/index.js"
     },
     "./deprecated": {
-      "import": "./lib-esm/deprecated/index.js",
-      "require": "./lib/deprecated/index.js"
+      "import": "./lib-esm/deprecated/index.js"
     },
     "./lib-esm/*": {
       "import": [
         "./lib-esm/*.js",
```
joshblack, line 22: I think if we add `type: module` for the `package.json` that we can remove the `"import"` for each of these and just set the value of each entrypoint to the path. Let me know if this doesn't work!

rollup.config.mjs
```diff
@@ -56,13 +56,6 @@ const input = new Set([
 ])
 
 const extensions = ['.js', '.jsx', '.ts', '.tsx']
-const ESM_ONLY = new Set([
-  '@github/combobox-nav',
-  '@github/markdown-toolbar-element',
-  '@github/paste-markdown',
-  '@github/relative-time-element',
-  '@lit-labs/react',
-])
```
joshblack, line 65: 🥳

joshblack (changes requested):
Looking great! Just left a comment for the `package.json` file itself 👀

joshblack:
@radglob just wanted to double check, I'm still not seeing `"type": "module"` in `package.json` and was curious if it's on my end or if it still needs to be added 👀

joshblack:
Closing this out due to the technical limitations mentioned in: https://github.com/primer/react/issues/3565 😞 Will make sure to use this as the base for any follow-up ESM work 🤞

PR #3416, opened 2023-06-15
Removes `label-has-for` from `ESLint` config

TylerJDev:
We seem to be utilizing the `jsx-a11y` rule `label-has-for`, even though this rule is [both deprecated and disabled](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-for.md) in `jsx-a11y`. This rule is still being applied because we have it set up under "rules" within our ESLint config.

The rule that replaces `label-has-for` ([`label-has-associated-control`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md)) will still catch any issues that `label-has-for` would, and require less usage of eslint-disable.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti (approved):

joshblack (approved):
🙌

PR #3423, opened 2023-06-19
Avoid calling `getCharacterCoordinates` when `fullHeight` is set in the MarkdownEditor

japf:
* fix https://github.com/github/primer/issues/2355

This fixes a performance issue where on each keystroke `getCharacterCoordinates` is called in the markdown editor. 
When `fullHeight` is set to true, we can skip this call and save some cpu cycles 🚀 

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/drafts/hooks/useDynamicTextareaHeight.ts
+++ b/src/drafts/hooks/useDynamicTextareaHeight.ts
@@ -4,7 +4,7 @@ import {SxProp} from '../../sx'
 import {getCharacterCoordinates} from '../utils/character-coordinates'
 
 type UseDynamicTextareaHeightSettings = {
-  fullHeight?: boolean
+  disabled?: boolean
   minHeightLines?: number
   maxHeightLines?: number
   elementRef: RefObject<HTMLTextAreaElement | null>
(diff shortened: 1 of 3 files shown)


src/drafts/hooks/useDynamicTextareaHeight.ts
```diff
@@ -4,6 +4,7 @@ import {SxProp} from '../../sx'
 import {getCharacterCoordinates} from '../utils/character-coordinates'
 
 type UseDynamicTextareaHeightSettings = {
+  fullHeight?: boolean
```
iansan5653, line 7: Maybe we should just call this `disabled` or something generic like that? This hook isn't specific to the Markdown editor - it can be used to size any textarea - but `fullHeight` is.

japf: Good point, done in https://github.com/primer/react/pull/3423/commits/718e28ef46e4f9382a1de612c2750be9df51ac41

mperrotti (commented):
@japf - could you add a Storybook example for this behavior in `src/drafts/MarkdownEditor/MarkdownEditor.stories.tsx`?

joshblack (approved):

PR #3427, opened 2023-06-19
Fix mode switcher

lukasoppermann:
**Summary**
Update the way we add the `data-color-mode` attribute. It was wrong before so that the `data-color-mode` attribute did not update when the user changed the theme from the theme switcher.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


langermank (approved):
Testing in a codespace, makes sense! Just need to fix those lint errors before merging

Issue #3430, opened 2023-06-20
Primer leading/trailingVisual and Icons cannot be lazy, memo, forwardRef


mattcosta7:
### Description

related potentially, but not exactly - https://github.com/primer/react/issues/2457

This would not be necessary if we passed rendered elements as props instead of passing components (as the linked PR above suggests)

Primer defines some props that accept _function components_ however they don't accept things like forwardRef components, memoized components or lazy components (at least)

Where does this show up?

Places like this one - https://github.com/primer/react/blob/21fa73cde4ff0172a95e0aca19550e5dd554c2a0/src/TextInput/TextInput.tsx#L137

this _should_ render anything that's component-like, and not just function components, but it only handles function components. 

something like this using [react-is](https://www.npmjs.com/package/react-is) might be what we want here, with type updates to match

```tsx
import {isValidElementType} from 'react-is'

 {isValidElementType(LeadingVisual) ? <LeadingVisual /> : LeadingVisual} 
```

When using primer/octicons-react version 19+, these components break because icons are all `forwardRef` components in that version

Attempting to use v19+ in `github/github` also trips these failures - https://github.com/github/github/pull/276931

### Steps to reproduce

Attempt to pass a memoized, lazy or ref forwarding component to a leadingVisual

### Version

latest

broccolinisoup:
Hello @mattcosta7!👋🏻 Thanks for raising this issue. [Bumping the octicon version](https://github.com/primer/react/pull/3266) was one of the [reverted PRs](https://github.com/primer/react/pull/3408) and I did a very poor job on updating these function type checkings on the bump PR 🙈 I already have a PR to re-introducing the bump and function type checks, https://github.com/primer/react/pull/3428, now I am using the `is-react` library as you suggested. It is still wip but I probably will reach out again due to type updates, I am struggling a bit with them.

mattcosta7:
@broccolinisoup something like this should do it? https://github.com/primer/react/pull/3437

Issue #3429, opened 2023-06-20
Re-add the PRs that are reverted from `main` due to introducing breaking changes to github.com on the v35.26.0 release


broccolinisoup:
Due to the recent deploy freeze at dotcom, we were unable to release primer/react on a regular schedule and this impacted the scope of the release candidate, therefore made it very challenging for us to address failing tests at dotcom. As a team, we made a decision to revert the PRs that we are introducing breaking changes to be able to get the current release out. [Here](https://github.com/primer/react/pull/3408) is the PR that reverts the PRs that introducing breaking changes. 

This issue focuses on adding them back to `main` to be released in one of the upcoming releases.

If you are assigned to this issue, it is high likely that we had to revert your PR(s) (apologise again ❤️ ) but now it is time to add back in. 

### Create a (draft) PR in primer/react

Create a new (draft) PR in https://github.com/primer/react with your reverted changes. You can cherry-pick the reverted commit on top of `main` or if you have a different idea/way of doing it, feel free to do it.  If you decide to go with cherry-pick, you can find the commit ID to cherry-pick in your original PR, towards the bottom. Example: "Merged via the queue into main with commit [9532977](https://github.com/primer/react/commit/95329773db772b4113368cd2774025c0486bf356) [on May 16](https://github.com/primer/react/pull/3266#event-9252410375)" Sorry if you already know about this stuff, I just wanted to make sure we don't block anyone in case they are not familiar with reverting/re-adding commits.

For example, I created one for my reverted octicon changes https://github.com/primer/react/pull/3428

### Canary version of `@primer/react` at PR time

We release a canary version of primer react, everytime when there is a new PR created as well as the PR receives new commits. If you scroll to the CI checks of your newly created draft PR, you will see one job that publishes these canary packages. ie.e "Published @primer/react — 0.0.0-xyz" - You can use this 0.0.0-xyz version to test your changes at github.com

### Dotcom development environment setup

Open up your dotcom development environment. If this is your first time, please refer to [the hub documentation](https://thehub.github.com/epd/engineering/products-and-services/dotcom/development-environment/). If you have any problems setting up your dotcom environment, please ask help on [these slack channel.](https://thehub.github.com/epd/engineering/products-and-services/dotcom/development-environment/#getting-help)

### Installing the canary version at dotcom

[...]

PR #3428, opened 2023-06-20
Bump @primer/react-octicons to v19.0.1 and update `Octicons` to use `React.forweardRef` (re-introducing reverted PRs)

broccolinisoup:
Describe your changes here.

Closes # (type the issue number after # if applicable; otherwise remove this line)

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3437, opened 2023-06-21
use react-is for checking valid element type, update primer/octicons-react

mattcosta7:
This change allows for trailing/leading components to properly accept React component types other than function components (`memo`, `lazy`, `forwardRef`, etc).  

This also bumps the version of primer/octicons-react which could be done separately, but why not just land this together.

This change is backward compat, since function components are still allowed, but properly allows for more types to be considered

Closes https://github.com/primer/react/issues/3430

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/TextInput/TextInput.tsx
```diff
@@ -24,11 +25,11 @@ export type TextInputNonPassthroughProps = {
   /**
    * A visual that renders inside the input before the typing area
    */
-  leadingVisual?: string | React.ComponentType<React.PropsWithChildren<{className?: string}>>
+  leadingVisual?: React.ElementType | React.ReactNode
```
mattcosta7, line 28: we don't pass props internally, so we don't need to actually care what the props the element can take are

src/__tests__/TextInput.test.tsx
```diff
@@ -57,10 +57,56 @@ describe('TextInput', () => {
 
   it('renders leadingVisual', () => {
     expect(render(<TextInput name="search" placeholder={'Search'} leadingVisual={SearchIcon} />)).toMatchSnapshot()
+    expect(render(<TextInput name="search" placeholder={'Search'} leadingVisual={<SearchIcon />} />)).toMatchSnapshot()
```
mattcosta7, line 60: just a bunch of tests that validate rendered content given a bunch of different element types

src/TextInputWithTokens/TextInputWithTokens.tsx
```diff
@@ -296,7 +297,7 @@ function TextInputWithTokensInnerComponent<TokenComponentType extends AnyReactCo
         visualPosition="leading"
         showLoadingIndicator={showLeadingLoadingIndicator}
       >
-        {typeof LeadingVisual === 'function' ? <LeadingVisual /> : LeadingVisual}
+        {typeof LeadingVisual !== 'string' && isValidElementType(LeadingVisual) ? <LeadingVisual /> : LeadingVisual}
```
mattcosta7, line 300: we have to check for `string` explicitly here, because things like `'div'` is a valid element type, but we don't want to try and render an empty div, but instead just pass through that text. 


We may want to consider whether accepting raw element types makes sense longterm (I don't think there's a good reason to accept a raw Component instead of just React.ReactNode).

```tsx
<Component leadingVisual={SearchIcon} /> // imo this shouldn't be allowed

<Component leadingVisual={<SearchIcon />} /> // imo this should be the way
```

but that might be a later breaking change at this point?

package.json
```diff
@@ -99,12 +99,13 @@
     "@github/relative-time-element": "^4.1.2",
     "@lit-labs/react": "1.1.1",
     "@primer/behaviors": "1.3.4",
-    "@primer/octicons-react": "18.3.0",
+    "@primer/octicons-react": "^19.3.0",
```
mattcosta7, line 102: I moved from an exact dep, to avoid version conflicts in dotcom, which this would cause eventually otherwise (if dotcom installed, for instance, 19.4.0).

Should octicons-react just be a peer dependency and not a direct dep?

mattcosta7:
this will superceed https://github.com/primer/react/pull/3428 which I should have looked for last night before starting this work, and only saw today - however this PR already fixes some issues with that one, so may be easier to just push through

joshblack (approved):

package.json
```diff
@@ -118,6 +119,7 @@
     "lodash.isempty": "4.4.0",
     "lodash.isobject": "3.0.2",
     "react-intersection-observer": "9.4.3",
+    "react-is": "18.2.0",
```
mattcosta7, line 122: going to loosen this and types so they can be managed better in dotcom

src/TextInputWithTokens/TextInputWithTokens.tsx
broccolinisoup: Oh great point! Thanks for the note as well, that was one of my questions until I came to that point!

➕ 1  for using ReactNodes instead of raw components

broccolinisoup (approved):
Thank you so much for this PR! and thanks for the notes as well, super helpful.

broccolinisoup:
I closed my PR and we can continue with this. The only thing is that because the original PR is reverted, we should individually test this at dotcom as [a part of the re-introducing strategy](https://github.com/primer/react/issues/3429). Are you happy to do that @mattcosta7 or do you prefer for me to do it? Either way works for me 🙂

src/TextInput/TextInput.tsx
mattcosta7: ooo interesting.  I didn't realize ReactNode included that, and i'm suprised it does (since ElementType isn't renderable as children directly).  I think i'd prefer this since it's a bit more explicit that this could be a Component, but could also be renderable content?

mattcosta7:
> I closed my PR and we can continue with this. The only thing is that because the original PR is reverted, we should individually test this at dotcom as [a part of the re-introducing strategy](https://github.com/primer/react/issues/3429). Are you happy to do that @mattcosta7 or do you prefer for me to do it? Either way works for me 🙂

Opened: https://github.com/github/github/pull/277602

mattcosta7:
VRT failures seem related to the actual icon changing slightly - not sure how to update those 

Screenshot 2023-06-22 at 10 38 46

broccolinisoup:
> VRT failures seem related to the actual icon changing slightly - not sure how to update those
> 
> comparison of icons in vrt output

I added `update snapshot` label, hopefully that should take care of it 😊

broccolinisoup:
>Opened: https://github.com/github/github/pull/277602
CI looks good there - let me know whehter I should merge this or y'all want to later?

Yay! Once we get the vrt green, feel free to merge it. Thanks again 🎉

PR #3445, opened 2023-06-23
update lockfile

gr2m:
Follow up to https://github.com/primer/react/pull/3352#issuecomment-1603484127

Updating the lock file results in all kinds of errors. I just want to bring it to your attention as it potentially means that in-range version updates break some of the `@prime/react` components, so users that update their lockfiles have new installs will be affected.


broccolinisoup (commented):
Hi @gr2m 👋🏻 What kind of errors are you getting? Could you elaborate on this please?
And where do you get these errors, while running which command etc?

I reinstalled the modules and run tests and storybook, all seems to be fine on my local.

gr2m:
The errors you see in the CI.

> I reinstalled the modules and run tests and storybook, all seems to be fine on my local.

Maybe the problem is resolved then, let's close this PR

gr2m:
@broccolinisoup I checked locally, the problem does persist for me.

Try this

```
git checkout main
rm -rf node_modules package-lock.json
npm install
npm test
```

PR #3462, opened 2023-06-29
docs: add versioning guide

joshblack:
This is a follow-up to: https://github.com/primer/react/pull/2753 with a narrowed scope of changes that I believe we have consensus on with respect to semver 👀 Figured it'd be useful to get this document into `contributor-docs` and add to it over time as we run into things in our release process.

Would love to know what you all think of this format and if there is anything that should be added, edited, or if there is a better approach to documenting these kinds of conventions so that we can refer to them in reviews and in our release process.

### DIFF
--- a/contributor-docs/versioning.md
+++ b/contributor-docs/versioning.md
@@ -56,7 +56,7 @@ For a full list of releases, visit our [releases](https://github.com/primer/reac
 
 semver bump: **minor**
 
-When a type is broadeded, it is now possible for that type to include both the
+When a type is broadened, it is now possible for that type to include both the
 existing value and additional values. As a result, this maintains backwards
 compatability while providing new functionality to the library.
 
--- a/contributor-docs/versioning.md
+++ b/contributor-docs/versioning.md
@@ -1,0 +1,93 @@
+# Versioning
+
+<!-- prettier-ignore-start -->
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+## Table of Contents
+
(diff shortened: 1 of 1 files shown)


contributor-docs/versioning.md
```diff
@@ -0,0 +1,93 @@
...
+|           | A prop is deprecated                                                | `minor`     |
+|           | A prop is removed                                                   | `major`     |
+
+## Reference
+
+### The type of a prop is broadened
+
+semver bump: **minor**
+
+When a type is broadeded, it is now possible for that type to include both the
```
lesliecdubs, line 59: > broadeded

Typo?

joshblack: Hah, great catch 😂

contributor-docs/versioning.md
```diff
@@ -0,0 +1,93 @@
...
+## Changes
+
+| Category  | Type of change                                                      | semver bump |
+| :-------- | :------------------------------------------------------------------ | :---------- |
+| Component | A component is added                                                | `minor`     |
+|           | A component is deprecated                                           | `minor`     |
+|           | A component is removed                                              | `major`     |
+| Props     | A prop is added                                                     | `minor`     |
+|           | [The type of a prop is broadened](#the-type-of-a-prop-is-broadened) | `minor`     |
+|           | [The type of a prop is narrowed](#the-type-of-a-prop-is-narrowed)   | `major`     |
```
broccolinisoup, line 49: I love these so much! 💖

broccolinisoup (approved):
🔥 

How can we make sure this doc becomes a living document as we discover cases/examples in release rotations or in general really? 

Would it be helpful to add this to the release conductor checklist like an optional one maybe but encouraged? 🤔

PR #3463, opened 2023-06-29
chore: add release-blocker command

joshblack:
This PR adds a custom slash command, `/release-blocker`, that can be used on Release Tracking issues to provide incident details for something that is preventing a release. This looks to mirror some of our recent guidance that has been updated and provide a structured way to review incidents in releases over time.

Would love to hear your feedback on the format here! Generally it has three areas:

- Describe the issue
- List the PRs that cause the issue
- Confirm necessary steps to revert and re-open corresponding issues

### DIFF
--- a/.github/commands/release-blocker.yml
+++ b/.github/commands/release-blocker.yml
@@ -2,6 +2,8 @@
 trigger: release-blocker
 title: Release Blocker
 description: Describe the issue(s) preventing a release from being published
+surfaces:
+  - issue
 steps:
   - type: form
     style: embedded
@@ -11,6 +13,21 @@ steps:
           id: blocker-description
           label: Description
           placeholder: Describe the issue that prevents the release from being published.
+
+      - type: dropdown
+        attributes:
+          id: blocker-type
+          label: Type
+          options:
+            - 'Component or type removed'
+            - 'Type incompatability'
+            - 'Other'
+
+      - type: input
+        id: other
+        attributes:
+          label: Other
+          description: 'Provide a custom blocker type if other was used above'
 
       - type: textarea
         attributes:
--- a/.github/commands/release-blocker.yml
+++ b/.github/commands/release-blocker.yml
@@ -1,0 +1,29 @@
+---
+trigger: release-blocker
+title: Release Blocker
+description: Describe the issue(s) preventing a release from being published
+steps:
+  - type: form
+    style: embedded
+    body:
+      - type: textarea
+        attributes:
+          id: blocker-description
+          label: Description
+          placeholder: Describe the issue that prevents the release from being published.
+
+      - type: textarea
+        attributes:
+          id: pull-requests-to-revert
+          label: Pull Requests
+          placeholder: Provide one or more Pull Requests that need to be reverted in order to publish the release.
+
+      - type: checkboxes
+        id: confirm
+        attributes:
+          label: Please confirm the following
+          options:
+            - label: I have re-opened any issues that were closed due to the reverted Pull Requests
+              required: true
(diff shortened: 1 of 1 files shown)


.github/commands/release-blocker.yml
```diff
@@ -0,0 +1,46 @@
...
+        attributes:
+          id: blocker-description
+          label: Description
+          placeholder: Describe the issue that prevents the release from being published.
+
+      - type: dropdown
+        attributes:
+          id: blocker-type
+          label: Type
+          options:
```
broccolinisoup, line 21: Ooo I love this 🔥

.github/commands/release-blocker.yml
```diff
@@ -0,0 +1,46 @@
...
+            - 'Component or type removed'
+            - 'Type incompatability'
+            - 'Other'
+
+      - type: input
+        id: other
+        attributes:
+          label: Other
+          description: 'Provide a custom blocker type if other was used above'
```
broccolinisoup, line 30: Oo I love this too! 🔥 So what is the game plan to keep this list up-to-date? 🤔 

Are we looking into adding the new blocker types when they were specified in "other" and update the list here or do you have something different in your mind?

.github/commands/release-blocker.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - type: textarea
+        attributes:
+          id: pull-requests-to-revert
+          label: Pull Requests
+          placeholder: Provide one or more Pull Requests that need to be reverted in order to publish the release.
```
broccolinisoup, line 36: I wonder if we can do release-blocker per PR?

For example just looking into the yesterday's issue (https://github.com/primer/react/issues/3455) Same PR has 2 blocking issues with different blocker types. 

If we can do per PR, we can maybe add `multiple: true` to the dropdown type attributes over [here](https://github.com/primer/react/pull/3463/files#diff-c17559b938b4bf025a7c87bff8120911af87d11349a27721609263b788c2f31cR18)?

broccolinisoup (commented):
Love this! Left a couple of questions/comments - Looking forward to hearing what you think 🙌🏻 

Also, do you have anything in your mind how we move this data over from issues to the spreadsheet to look at the data more holistically? I'd be happy to do manually after a month or so maybe but curious to hear if you have anything else in your mind.

Thanks again, this is great ✨

.github/commands/release-blocker.yml
joshblack: Ooo good call @broccolinisoup, will definitely add that in 👍

.github/commands/release-blocker.yml
joshblack: @broccolinisoup honestly nothing specific came to mind, super open to ideas here. Figured it could be a manual list that we update as things come up and if we find ourselves using this command a lot then we could try and keep track of what gets put into the "other" field and use that data to go off of (but I honestly have no idea what's best here haha)

broccolinisoup: That sounds good to me! We can try to see what works best really. I am perfectly happy to start manual 💪🏻

.github/commands/release-blocker.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - type: checkboxes
+        id: confirm
+        attributes:
+          label: Please confirm the following
+          options:
+            - label: I have re-opened any issues that were closed due to the reverted Pull Requests
```
siddharthkp, line 46: ooh smart!

siddharthkp (approved):
Very nice!

Let's ship it and iterate based on how it feels

PR #3456, opened 2023-06-27
Fix type errors in next-major

mperrotti:
Fixes type errors we started getting recently in branches off of `next-major`.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3464, opened 2023-06-29
CI TEST - Fix type errors in next-major

mperrotti:
This is the same as https://github.com/primer/react/pull/3456

I just want to see if opening a new PR will get CI to pass.


mperrotti:
Didn't work :(

PR #3466, opened 2023-06-30
Allow customization of AvatarStack hover behavior and size

mperrotti:
- Allows consumers to disable the behavior where the stack expands to show all avatars on hover
- Allows consumers to customize the size for the avatars in the stack. This can be done two ways:
  - pass a `size` prop to `AvatarStack`
  - pass a `size` prop to the `Avatar`s that are direct children of `AvatarStack`
- Does not allow an `AvatarStack` to render it's children at different sizes. The smallest size will always be picked.

Closes https://github.com/github/primer/issues/2366

### Screenshots

https://github.com/primer/react/assets/2313998/c52bf0d7-90a7-48ac-b7c4-e6edb382b6d4

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/AvatarStack/AvatarStack.tsx
```diff
@@ -4,22 +4,53 @@ import styled from 'styled-components'
 import {get} from '../constants'
 import Box from '../Box'
 import sx, {SxProp} from '../sx'
+import {AvatarProps, DEFAULT_AVATAR_SIZE} from '../Avatar/Avatar'
 
 type StyledAvatarStackWrapperProps = {
   count?: number
 } & SxProp
 
+const findSmallestNumber = (numbers: number[]): number => {
```
broccolinisoup, line 13: Would `Math.min(...numbers)` work here? 🤔  Just an idea - not a blocker at all.

broccolinisoup (approved):
Love it! 💖

maximedegreve (approved):

maximedegreve (commented):
This is great @mperrotti 🎉

One thing that's missing in this PR is support for breakpoints.

Otherwise developers will have to get around this issue using:

```javascript
<Box sx={{display: ['block', 'block', 'none']}}>
   <AvatarStack size={26}/>
</Box>
<Box sx={{display: ['none', 'none', 'block']}}>
   <AvatarStack size={20}/>
</Box>
``` 

Instead of:
```javascript
<AvatarStack size={[20,20,24]}/>
``` 

It also makes working with HTML grids hard because the Box is never removed from the DOM and will create an additional grid cell.

src/AvatarStack/AvatarStack.tsx
```diff
@@ -127,21 +163,42 @@ const transformChildren = (children: React.ReactNode) => {
 
 export type AvatarStackProps = {
   alignRight?: boolean
+  disableExpand?: boolean
```
siddharthkp, line 166: What do you think about `expand: boolean` where default is set to true?

src/AvatarStack/AvatarStack.tsx
mperrotti: Great suggestion! Soooo much easier.

src/AvatarStack/AvatarStack.tsx
mperrotti: It feels strange to have to explicitly pass `propName={false}` because boolean props are normally evaluated as `false` if the prop is excluded.

If we didn't set `expand` to `true` by default, then `<AvatarStack />` would be the same as `<AvatarStack expand={false} />`

mperrotti:
@broccolinisoup and @maximedegreve - I have your suggestions implemented.

This should be merge-able once I:
- Update tests and stories
- Fix an annoying TypeScript error when trying to pass `getBreakpointDeclarations()` to `merge()` (@broccolinisoup - maybe we can pair on this?)
- Other misc code cleanup
- Comment any complicated code I couldn't simplify

src/AvatarStack/AvatarStack.tsx
siddharthkp: > It feels strange to have to explicitly pass propName={false}

true true!

> because boolean props are normally evaluated as false if the prop is excluded

not sure about that tho 😅 

```jsx
<AvatarStack/> 
<AvatarStack expand />
<AvatarStack expand={true} />

<AvatarStack expand={false}/>
```

buuuut, happy to go with what you feel strongly about

maximedegreve:
@mperrotti Oh no!

mperrotti:
Haha thanks for sharing that @maximedegreve - I'll be fixing that while I make my other changes today.

mperrotti:
ok @maximedegreve and @broccolinisoup - this _should_ be ready to merge once CI passes 🎉

maximedegreve:
LGTM 🚀💪

dusave:
Broader question, not a blocker for this PR: Are we adding any sort of `prefers-reduced-motion` rule in the animation definitions of components to disable them for the user? This could be supplemented with the explicit prop (the consumer might want it disabled for _all_ users), but for accessibility reasons, adding the css check could make it much easier to work with.

src/Avatar/Avatar.tsx
```diff
@@ -31,18 +28,31 @@ const StyledAvatar = styled.img.attrs<StyledAvatarProps>(props => ({
...
 export type AvatarProps = ComponentProps<typeof StyledAvatar>
 
 const Avatar = React.forwardRef<HTMLImageElement, AvatarProps>(function Avatar(
-  {alt = '', size = 20, square = false, ...rest},
+  {alt = '', size = DEFAULT_AVATAR_SIZE, square = false, sx: sxProp = defaultSxProp, ...rest},
   ref,
 ) {
-  return <StyledAvatar ref={ref} alt={alt} size={size} square={square} {...rest} />
+  const avatarSx = isResponsiveValue(size)
+    ? merge(
```
broccolinisoup, line 46: Sorry I missed your ping about the type stuff.

I think we could also do 

```
merge<BetterSystemStyleObject>(...)
```

instead of typing each input `as`?

Also totally optional maybe for readability 

```jsx
const avatarSx = isResponsiveValue(size)
    ? getBreakpointDeclarations(
        size,
        '--avatar-size' as keyof React.CSSProperties,
        value => `${value || DEFAULT_AVATAR_SIZE}px`,
      )
    : {'--avatar-size': `${size}px`}
```

and then in the `StyledAvatar`

```
sx={merge<BetterSystemStyleObject>(avatarSx, sxProp)}
```

Just a suggestion - however you prefer.

broccolinisoup (approved):
Looks great! Just left a comment about the type stuff but feel free to ignore if it is not helpful. 

🚀 🚀 🚀

mperrotti:
@dusave 
>  Are we adding any sort of prefers-reduced-motion rule in the animation definitions of components to disable them users automatically? 

We do use `prefers-reduced-motion` rules in some components, but not this one. I'm going to check with a11y design to see if it makes sense for this small transition.

src/Avatar/Avatar.tsx
mperrotti: I _think_ I remember there being an performance issue with passing `merge()` directly to to `sx`, so I'm going to keep it as-is. Definitely going to use a generic instead of `as` though. Great suggestion!

broccolinisoup: Ohh I didn't know about the performance impact on using merge in sx! Thanks for letting me know!

Issue #3461, opened 2023-06-29
[ActionList] showDividers bug


josepmartins:
### Description

When using the `showDividers` property in an `ActionList` component the first line doesn't show up. I was able to reproduce it and then opened the devtools and it magically rendered the divider.

https://github.com/primer/react/assets/912236/4fe97930-3157-472a-97f9-f73457f9e80a

### Steps to reproduce

1. Add `showDividers` in an `ActionList` with 2 or more items
2. Diplay the menu and see how the first line is hidden
3. Open devtools or resize the window, the first line is displayed

### Version

v35.26.0 

### Browser

Chrome

siddharthkp:
[From @maraisr](https://github.slack.com/archives/C01L618AEP9/p1688110431448639?thread_ts=1687740736.384229&cid=C01L618AEP9), this happens in Safari and Firefox as well, so it's not browser rendering related

siddharthkp:
Update: We disable the divider when we expect a focus ring there instead.  [source code](https://github.com/primer/react/blob/f577190f97da8e427074b558ec03c4fb80182fb2/src/ActionList/Item.tsx#L148-L158)

In this example, we're expecting a focus ring for some reason when there isn't one 🤔

siddharthkp:
Update: This was a missing piece of https://github.com/primer/react/pull/3056 :)

PR #3476, opened 2023-07-03
Fix ActionMenu focused divider

siddharthkp:
- Fixes https://github.com/primer/react/issues/3461

Context:

- When you open ActionMenu with mouse click, we focus the first item by default (this matched a11y guidance at the time)
- In the past, we treated both focus and focus-visible the same and added a focus ring on the focused item
- To avoid both a divider and focus ring at the same time, we have some logic to hide divider if a surrounding item has focus or focus-visible
- In https://github.com/primer/react/pull/3056, we removed the focus ring on focus (keeping it only on focus-visible)
- Without matching divider logic, we end up with the divider hidden on focus even when there is no ring (reported in https://github.com/primer/react/issues/3461)

This pull request fixes that by removing divider logic for focus (while keeping it for focus-visible)


broccolinisoup (approved):
Looks great to me 🚀 
Also, thanks so much for the context! I love how you wrote it like a story 🤗

PR #3477, opened 2023-07-03
chore(deps): update @types/node to v18

joshblack:
Update `@types/node` to v18 since the project currently uses Node v18 LTS


siddharthkp (approved):
If typecheck is passing, I think there's no reason not to!

PR #3449, opened 2023-06-25
Export Blankslate and BlankslateProps

echarles:
Simple export of the `Blankslate` component following the way other components are exported.

Closes #3448

### Screenshots

N/A

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3478, opened 2023-07-03
docs: update titles and order for storybook hierarchy

joshblack:
Update the order and hierarchy of the storybook to match the following:

- Components
- Behaviors
- Hooks
- Deprecated
- Drafts
- Private

The notable changes for this are:

- Deprecated components -> Deprecated 
- Private components -> Private

Along with the order change. Curious to see what folks think for this change and if the ordering/hierarchy makes sense, this is a follow-up to move Blankslate out of Drafts and into Private and thought some of the other clean up could help to 👀 Let me know what you think!


siddharthkp (approved):
Looks right to me!

src/Blankslate/Blankslate.features.stories.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react'
 import {Blankslate} from '../Blankslate'
 
 export default {
-  title: 'Drafts/Components/Blankslate/Features',
+  title: 'Private/Components/Blankslate/Features',
```
broccolinisoup, line 6: We discussed this in the planning session and seems like it is useful for folks out there (https://github.com/primer/react/pull/3449) do you have any concerns to keep it as draft/experimental?

siddharthkp: I think we should only export this as experimental if we (primer) want to work on standardising or work on a11y for this component, otherwise it should be created in dotcom/ui/packages?

broccolinisoup: >I think we should only export this as experimental if we (primer) want to work on standardising or work on a11y for this component, otherwise it should be created in dotcom/ui/packages?

For this particular request came from a non-Hubber and I love the idea of exporting components as experimental because it helps/saves time some folks out there. 

>I want to avoid the scenario where folks create a feature request issue or a pull request implementing a feature, and then we don't have the design or a11y context to approve/reject it 😅

Fair enough! I think if we see this is becoming a pattern then we can set some boundaries around it?

siddharthkp: Sounds like a plan :shipit:

joshblack: @broccolinisoup I'll make the update here to move back to `Drafts`! Will re-request once that's done 👀

broccolinisoup (approved):
Thank you! 🚀

PR #3479, opened 2023-07-03
refactor: update components.json step to run at build time

joshblack:
I noticed when looking at the `merge_group` status checks that the status check for generating `components.json` was failing consistently due to the stage where it commits the file back to the upstream pull request. This PR updates the flow for `generated/components.json` so that:

- The CI check still runs the build command but _does not_ commit it back to the PR
  - This means that we will still get the errors if the command fails but not if it succeeds even if the PR is from a contributor or is in a merge group
- The `generated/components.json` file itself is now in `.gitignore` and is considered a build artifact
- The `script/build` command now includes the build command for `components.json` so that it is still included in the package artifact

Curious how you all feel about this change 👀 At the end of the day, wanted to reduce the number of red ❌'s that we see from this check while still including it to make sure things don't fail


siddharthkp (commented):
Love that this will mean more green ticks

I don't think I have ever intentionally reviewed this file in a pull request, only the component specific ones 🤔 

Am I wrong in assuming this is only an artefact, not a source file?

.gitignore
```diff
@@ -12,3 +12,4 @@ stats.html
 .env
 storybook-static
 .tool-versions
+/generated
```
broccolinisoup, line 15: Love this!

script/build
```diff
@@ -16,3 +16,6 @@ npx copyfiles -u 1 "./lib/**/*.d.ts" ./lib-esm
 
 # Copy css
 npx copyfiles --flat ./lib/components.css ./css
+
+# Build components.json file for documentation
+npx ts-node script/components-json/build
```
broccolinisoup, line 21: I am just not sure if I understand the reasoning behind of including the command here? I would appreciate if you could clarify, thank you 🙏🏻

broccolinisoup (approved):
I love this change. I believe this will also help with the other two things I think of:
- Prevent manual editing the file
- No need to fetch your branch after the first commit (I always forget that we have this commit generated and I try pushing to origin and it says my origin has changes that I don't have) It is a small thing but still :)

joshblack:
> Am I right in assuming this is only an artefact, not a source file?

Yes! Exactly. The source files, as I understand it, should only be the component json files that we write next to a component which are then used to generate the file at the top-level.

script/build
joshblack: @broccolinisoup this command should mirror what is being run currently in the CI job that commits back to the Pull Request with a generated `components.json` file (I hope 🤞 lol). With this change, we are no longer committing this file to the repo and so need to generate it before the package is published so that it's available on npm. Hope that makes sense 👀

gr2m:
@joshblack this change broke `npx changeset`

```
$ npx changeset
🦋  Loading components information for @primer/react from /code/primer/react/generated/components.json
🦋  error Error: Cannot find module '/code/primer/react/generated/components.json'
```

The `generated/components.json` needs to be present in order for the CLI to offer a choice of valid component names. We could load the components remotely if needed, but that won't work if a new component is added as part of a pull request

PR #3467, opened 2023-07-01
docs: fix textarea props description

ValbertMartins:
Describe your changes here.
 • fix wrong description to resize property of textarea component in docs
 • remove duplicated ref property of textarea in docs

### Merge checklist

- [ ] Added/updated tests
- [X] Added/updated documentation
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3480, opened 2023-07-03
test(e2e): add retry count in ci

joshblack:
Our e2e tests can be flaky at times and often just running them again is enough to get them going green ✅. This PR updates our Playwright config to retry twice in CI so avoid this flakiness issue. If tests continue to fail inconsistently then we know that a deeper issue is at play here 👀 

[Retry Docs](https://playwright.dev/docs/api/class-testconfig#test-config-retries)

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,8 +1,8 @@
 import path from 'node:path'
-import {Config} from '@playwright/test'
+import {defineConfig} from '@playwright/test'
 import './e2e/matchers/toHaveNoViolations'
 
-const config: Config = {
+export default defineConfig({
   // https://playwright.dev/docs/api/class-testconfig#test-config-test-dir
   testDir: path.join(__dirname, 'e2e'),
   testIgnore: ['**/matchers/**', '**/test-helpers/**'],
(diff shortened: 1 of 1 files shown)


siddharthkp (approved):
Fair enough!

siddharthkp:
Do we know which tests tend to be flaky to write better tests?

generated/components.json
```diff
@@ -4370,7 +4370,7 @@
           "name": "resize",
           "type": "'both' | 'horizontal' | 'vertical' | 'none'",
           "defaultValue": "'both'",
-          "description": "Changes background color to a higher contrast color"
+          "description": "Changes the resize behavior"
```
broccolinisoup, line 4373: Oo generating components.json doesn't run on forked PRs? 🤔 Because this was fixed in this [PR](https://github.com/primer/react/pull/3467)

broccolinisoup (approved):

generated/components.json
broccolinisoup: Oh I think it is not to do with the forked repo PRs, I guess it is related to https://github.com/primer/react/pull/3479?

joshblack: @broccolinisoup I think you're right that the generate command will fail on PRs from forks since they don't have access to the correct secrets in order to push back to the origin branch 😞 (since it runs in the CI job)

joshblack:
@siddharthkp I can't seem to find any rhyme or reason to them when looking at them 😞 they tend to be related specifically to resolving the storybook URL versus anything else. We have caught some animation-related flakiness previously which has been great for test writing practices.

PR #3482, opened 2023-07-04
feat: Introduces `<Toast />`

maraisr:
GitHub is wanting to introduce a `<Toast />` Primer React component, to help with letting the user know about something.

This PR here simply introduces the underlying UI component, one that an end-user will not consume — hence the lack of re-export.

The intended api:

```tsx
import {ToastOutlet, Toast} from '@primer/react'

function MyProduct() {
  const [error, setError] = useState(null)

  return (
    <React.Fragment>
      <button onClick={() => setError('my mistake')}>Whoops</button>
      {error && <Toast variant="danger">{error}</Toast>} /* 👈 toast is an element */
    </React.Fragment>
  )
}

function App() {
  return (
    <React.Fragment>
      <MyProduct />
      <ToastOutlet /> /* 👈 where we will put the items */
    </React.Fragment>
  )
}
```

Since this element is rendered in the tree, we can update them over time like any stateful component, like "Closing issue 6/10" with a loading variant. Also, during SSR this can also be leveraged in place of `<Flash />`, as we can snapshot those toast messages mounted during SSR and display them to the user once we go to render.

Lots of benefits on this solution.

PoC on direction 👉 https://codesandbox.io/s/competent-khorana-27chtg?file=/src/App.js

> Will iterate now into building out the behaviours. Will do that against this branch, so it can all land in a single merge. Just raised to get some early eyes.

### Screenshots

an example of a default toast message

### Merge checklist

- ~[ ] Added/updated tests~
- ~[ ] Added/updated documentation~
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/experimental/Toast/ToastUI.tsx
+++ b/src/experimental/Toast/ToastUI.tsx
@@ -1,7 +1,10 @@
 import React from 'react'
 import {CheckIcon, AlertIcon, StopIcon, InfoIcon, XIcon} from '@primer/octicons-react'
 
-import {Spinner, Box, IconButton} from '../..'
+import {Spinner, Text, IconButton} from '../..'
+
+import classNames from './ToastUI.module.css'
+import {useLayoutEffect, useRef, useState} from 'react'
 
 const variants = ['default', 'success', 'attention', 'danger', 'loading'] as const
 
@@ -16,7 +19,6 @@ const icon: Record<ToastVariants, React.ReactNode> = {
 export const ToastUI = ({
   dismissible = false,
   variant = 'default',
-  role = 'status',
   dismissLabel = 'Dismiss this message',
   children,
 }: React.PropsWithChildren<ToastUIProps>) => {
@@ -28,85 +30,53 @@ export const ToastUI = ({
     )
   }
 
+  const textRef = useRef<HTMLDivElement>(null)
+  const [height, setHeight] = useState<string>('0px')
+
+  useLayoutEffect(() => {
+    if (textRef.current) {
+      setHeight(window.getComputedStyle(textRef.current).lineHeight)
+    }
+  }, [])
+
+  const body = React.createElement(
+    typeof children === 'string' ? Text : 'div',
+    {
+      role: 'main',
+      className: classNames.ToastUI__text,
+      ref: textRef,
+    },
+    children,
+  )
+
   return (
-    <Box role={role} sx={{display: 'flex', justifyContent: 'flex-start'}}>
-      <Box
-        sx={{display: 'inline-flex', maxWidth: '450px', borderRadius: 2, boxShadow: 'shadow.large', overflow: 'hidden'}}
-      >
-        <Box
-          sx={{
-            display: 'flex',
-            justifyItems: 'center',
-            alignItems: 'center',
-            borderTopLeftRadius: 2,
-            borderBottomLeftRadius: 2,
-            border: '1px solid',
-            borderColor: 'border.default',
-            borderInlineEnd: 'none',
-            p: 3,
-            color: 'fg.onEmphasis',
-            '&[data-variant="default"]': {
-              backgroundColor: 'accent.emphasis',
-              borderColor: 'accent.emphasis',
-            },
-            '&[data-variant="success"]': {
-              backgroundColor: 'success.emphasis',
-              borderColor: 'success.emphasis',
-            },
-            '&[data-variant="attention"]': {
-              backgroundColor: 'attention.emphasis',
-              borderColor: 'attention.emphasis',
-            },
-            '&[data-variant="danger"]': {
-              backgroundColor: 'danger.emphasis',
-              borderColor: 'danger.emphasis',
-            },
-            '&[data-variant="loading"]': {
-              backgroundColor: 'neutral.emphasis',
-              borderColor: 'neutral.emphasis',
-            },
-          }}
-          role="presentation"
-          data-variant={variant}
-        >
+    <div role="alert" className={classNames.ToastUI} data-dismissable={dismissible}>
+      <div role="presentation" className={classNames.ToastUI__wrapper} style={{height: height}}>
+        <div className={classNames.ToastUI__Icon} data-variant={variant}>
           {icon[variant]}
-        </Box>
-        <Box
-          sx={{
-            display: 'grid',
-            gridTemplateColumns: dismissible ? 'auto 1fr' : '1fr',
-            alignItems: 'center',
-            gap: 3,
-            justifyContent: 'space-between',
-            borderBottomRightRadius: 2,
-            borderTopRightRadius: 2,
-            border: '1px solid',
-            borderColor: 'border.default',
-            borderInlineStart: 'none',
-            px: 3,
-            paddingRight: dismissible ? 2 : 3,
-          }}
-        >
-          <Box
+        </div>
+      </div>
+      <div>{body}</div>
+      {dismissible && (
+        <div className={classNames.ToastUI__wrapper} style={{height: height}}>
+          <IconButton
+            aria-label={dismissLabel}
+            icon={XIcon}
+            variant="invisible"
             sx={{
-              /* we margin to allow for margin collapsing when user-land uses as the children <Text as="p" /> */
-              marginY: 3,
+              color: 'var(--overlay-inverse-fgColor-default)',
+              '&:hover': {
+                backgroundColor: 'var(--button-invisible-bgColor-hover)',
+              },
+              '&:active': {
+                backgroundColor: 'var(--button-invisible-bgColor-active)',
+              },
             }}
-          >
-            {children}
-          </Box>
-          {dismissible && (
-            <IconButton
-              aria-label={dismissLabel}
-              icon={XIcon}
-              variant="invisible"
-              sx={{color: 'fg.default'}}
-              size="small"
-            ></IconButton>
-          )}
-        </Box>
-      </Box>
-    </Box>
+            size="small"
+          />
+        </div>
+      )}
+    </div>
   )
 }
 ToastUI.displayName = 'Toast'
@@ -124,17 +94,6 @@ export type ToastUIProps = {
    * @default false
    */
   dismissible?: boolean | undefined
-  /**
-   * Sets the role attribute of the toast, which is used by assistive technologies to communicate status to users.
-   *
-   * - "status" will not be read out of assistive technologies immediately, and won't interupt what the user is currently doing.
-   * - "alert" will be read out immediately, and will interupt what the user is currently doing. Because of this "alert" should be use sparingly.
-   *
-   * @see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role
-   *
-   * @default status
-   */
-  role?: 'status' | 'alert' | undefined
 
   /**
    * The `aria-label` to use for the close button.
--- a/src/experimental/Toast/ToastUI.tsx
+++ b/src/experimental/Toast/ToastUI.tsx
@@ -1,0 +1,147 @@
+import React from 'react'
+import {CheckIcon, AlertIcon, StopIcon, InfoIcon, XIcon} from '@primer/octicons-react'
+
+import {Spinner, Box, IconButton} from '../..'
+
+const variants = ['default', 'success', 'attention', 'danger', 'loading'] as const
+
+const icon: Record<ToastVariants, React.ReactNode> = {
+  default: <InfoIcon size="small" />,
+  success: <CheckIcon size="small" />,
+  attention: <AlertIcon size="small" />,
+  danger: <StopIcon size="small" />,
+  loading: <Spinner size="small" />,
+}
+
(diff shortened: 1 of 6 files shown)


src/experimental/Toast/ToastUI.tsx
```diff
@@ -0,0 +1,148 @@
...
+    console.assert(
+      variants.includes(variant),
+      `The variant ${variant} is not supported. Use one of ${variants.join(', ')}`,
+    )
+  }
+
+  return (
+    <Box
+      sx={{display: 'inline-flex', borderRadius: 2, maxWidth: '450px', boxShadow: 'shadow.large', overflow: 'hidden'}}
+      ref={ref}
```
maraisr, line 71: will remove this ref for now actually. Initially thought i'd require it for animations, like fading them in and out. But can always add this in later once i figure out how that'll work.

src/experimental/Toast/Toast.features.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+      )
+    },
+  ],
+  component: ToastUI,
+}
+export default meta
+
+export const Default = () => (
+  <>
+    <ToastUI variant="default">Message</ToastUI>
```
maraisr, line 26: This will swap out to the `<Toast />` with a trigger button once that's built.

Not sure if I should include a story for the internal ui component 🤔 if so, maybe this becomes `ToastUI.stories.tsx`?

inkblotty:
Hi all - Toasts are listed as a [High Risk Pattern for accessibility](https://github.com/github/engineering/discussions/3313). Please bring this pattern to an [upcoming Office Hour with the Accessibility team](https://github.com/github/accessibility/issues?q=is%3Aissue+is%3Aopen+office+hours+label%3Aa11y-eng-office-hours).

src/experimental/Toast/ToastUI.tsx
```diff
@@ -0,0 +1,145 @@
...
+            gap: 3,
+            justifyContent: 'space-between',
+            borderBottomRightRadius: 2,
+            borderTopRightRadius: 2,
+            border: '1px solid',
+            borderColor: 'border.default',
+            borderInlineStart: 'none',
+            px: 3,
+          }}
+        >
```
langermank, line 87:
```suggestion
        <Box
          sx={{
            display: 'flex',
            alignItems: 'center',
            gap: 3,
            justifyContent: 'space-between',
            borderBottomRightRadius: 2,
            borderTopRightRadius: 2,
            border: '1px solid',
            borderColor: 'border.default',
            borderInlineStart: 'none',
            px: 3,
            paddingRight: dismissible ? 2 : 3,
          }}
        >
```

If `dismissible` is true with the icon button, the padding right should be 8px for visual alignment.

src/experimental/Toast/ToastUI.tsx
```diff
@@ -0,0 +1,145 @@
...
+          <Box
+            sx={{
+              /* we margin to allow for margin collapsing when user-land uses as the children <Text as="p" /> */
+              marginBlock: 3,
+            }}
+          >
+            {children}
+          </Box>
+          {dismissable && (
```
langermank, line 96: Just a heads up I think its `dismissible` with an `i` (so needs to be changed everywhere) 😄

src/experimental/Toast/ToastUI.tsx
```diff
@@ -0,0 +1,145 @@
...
+            border: '1px solid',
+            borderColor: 'border.default',
+            borderInlineStart: 'none',
+            px: 3,
+          }}
+        >
+          <Box
+            sx={{
+              /* we margin to allow for margin collapsing when user-land uses as the children <Text as="p" /> */
+              marginBlock: 3,
```
langermank, line 91: A few things here:

- The `3` isn't mapping to a space variable (16px) I think because styled-components doesn't understand `marginBlock`. So its currently rendering 3px.
- I understand why you added this and it makes sense to have 16px of margin for the content, _but_ it looks like our Text component doesn't have any kind of browser reset. So by default it already has 1em of padding 😬 soooo.. maybe you could reset that in the scope of this component?

src/experimental/Toast/ToastUI.tsx
```diff
@@ -0,0 +1,145 @@
...
+    console.assert(
+      variants.includes(variant),
+      `The variant ${variant} is not supported. Use one of ${variants.join(', ')}`,
+    )
+  }
+
+  return (
+    <Box role={role} sx={{display: 'flex', justifyContent: 'flex-start'}}>
+      <Box
+        sx={{display: 'inline-flex', maxWidth: '450px', borderRadius: 2, boxShadow: 'shadow.large', overflow: 'hidden'}}
```
langermank, line 34:
```suggestion
        sx={{display: 'inline-flex', maxWidth: 'min(100vw, 450px)', borderRadius: 2, boxShadow: 'shadow.large', overflow: 'hidden'}}
```

I _think_ we might want to have all toasts be a min width of 450px on larger screens even if they have a small amount of content. I'm imagining if multiple toasts appear, we would want them to all have the same width in a stack. We might want to discuss this further in a pattern working group session as well with other designers. What do you think?

I'm pretty sure the CSS I suggested will give in a min width but also let it scale down for mobile 🤞

src/experimental/Toast/ToastUI.tsx
```diff
@@ -0,0 +1,145 @@
...
+              backgroundColor: 'success.fg',
+              borderColor: 'success.muted',
+            },
+            '&[data-variant="attention"]': {
+              backgroundColor: 'attention.fg',
+              borderColor: 'attention.muted',
+            },
+            '&[data-variant="danger"]': {
+              backgroundColor: 'danger.fg',
+              borderColor: 'danger.muted',
```
langermank, line 62:
```suggestion
            '&[data-variant="default"]': {
              backgroundColor: 'accent.emphasis',
              borderColor: 'accent.emphasis',
            },
            '&[data-variant="success"]': {
              backgroundColor: 'success.emphasis',
              borderColor: 'success.emphasis',
            },
            '&[data-variant="attention"]': {
              backgroundColor: 'attention.emphasis',
              borderColor: 'attention.emphasis',
            },
            '&[data-variant="danger"]': {
              backgroundColor: 'danger.emphasis',
              borderColor: 'danger.emphasis',
```

I'm making this suggestion without testing, but I think we want `emphasis` and we also want the background and border color to match.

Its easier to tell that the background color is incorrect if you check in the high contrast modes. I'm curious to see how the emphasis color looks in dark high contrast, as right now the white doesn't look correct over the lighter yellow shade.

toast component in dark high contrast mode

src/experimental/Toast/Toast.features.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+const meta: Meta = {
+  title: 'Components/Toast/Features',
+  decorators: [
+    Story => {
+      return (
+        <ThemeProvider>
+          <BaseStyles>
+            <Box sx={{display: 'inline-flex', flexDirection: 'column', gap: 3}}>{Story()}</Box>
+          </BaseStyles>
+        </ThemeProvider>
```
langermank, line 16:
```suggestion
            <Box sx={{display: 'flex', flexDirection: 'column', gap: 3}}>{Story()}</Box>
```

Should we consider some kind of ToastContainer component that will handle layout?

src/experimental/Toast/ToastUI.tsx
```diff
@@ -0,0 +1,145 @@
...
+            display: 'flex',
+            justifyItems: 'center',
+            alignItems: 'center',
+            borderTopLeftRadius: 2,
+            borderBottomLeftRadius: 2,
+            border: '1px solid',
+            borderColor: 'border.default',
+            borderInlineEnd: 'none',
+            p: 3,
+            color: 'white',
```
langermank, line 47:
```suggestion
            color: 'fg.onEmphasis',
```

langermank (commented):
Hello, I reviewed this with @broccolinisoup! We left some feedback. A few more things:

- If you wouldn't mind, could you add a Playground story? [This ADR](https://github.com/github/primer/blob/main/adrs/2022-10-07-storybook-lookbook-story-format.md) goes into detail about how to add stories to support our docs. 
- It would be good to attend accessibility design office hours to go over this API if possible 🙌 
- Have you considered animations yet? The default story wasn't working so I couldn't see what happens on open/close. We actually do have some [old Toast CSS](https://github.com/primer/css/blob/9204701416b547255a1ebaeb2bd8a20056ffb8d4/src/toasts/toasts.scss) in Primer CSS that includes animation styles. You could start with that!

src/experimental/Toast/ToastUI.tsx
langermank: I'd love to see if @maximedegreve has an opinion here! Should the toast width respond to the side of the contents, or should it have a min width so they stack evenly? I have seen it done both ways and I don't have a strong opinion.

maraisr: Hi, yeah thanks! What im wanting to achieve here is a maxWidth, hence the property. But assigned to the minimum value between 100vw and 450px. I guess this could have been articulated through `width: 100%; max-width: 45px` as well.

src/experimental/Toast/Toast.features.stories.tsx
maraisr: in the future yes, but not today! So far this quite simply just has the individual toast ui component. The container will be managed by the up-and-coming `ToastOutlet` component, 🚧

src/experimental/Toast/ToastUI.tsx
maraisr: great catch! I think @siddharthkp mentioned I could try my hand at css-modules, do you know how to get theme variables in those files?

src/experimental/Toast/ToastUI.tsx
maraisr: awesome find, 🤦🏻

src/experimental/Toast/ToastUI.tsx
langermank: 🤩 If you wanted to move towards a CSS file, you would access CSS variables instead which are a bit different! That can all be found in the [Primer Primitives Storybook](https://primer.style/primitives/storybook/?path=/story/size-base--base), but this is all very new so I don't have a ton of guidance on how to use specific properties. I'm happy to help though so if you give it a go, ping me again on this PR.

maraisr:
> Have you considered animations yet? The default story wasn't working so I couldn't see what happens on open/close. We actually do have some [old Toast CSS](https://github.com/primer/css/blob/9204701416b547255a1ebaeb2bd8a20056ffb8d4/src/toasts/toasts.scss) in Primer CSS that includes animation styles. You could start with that!

Hi yeah this is 100% just the ui for the toast component itself. We will have animations, yes. There's a concept ill also be introducing here called a `StackedList`, which will stack multiple elements on top of each other, and animate those in elements in.

It'll all come, and the work right now here is just the ui itself.

src/experimental/Toast/ToastUI.tsx
maraisr: Oh awesome! Wasn't aware that the new primitives was accessible in primer-react just yet. That's fantastic. Will give it a go!

maraisr:
cc @broccolinisoup @langermank I have changed gears for this PR to solely introduce the ToastUI internal component for us to iterate on.

About animations, yeah I will address that in a second moment. When the toast get's unmounted we'd want to first play the removal animation before removing the DOM.

And with the proposed API, toasts can be dismissed — so if a user dismisses a toast in the middle of other's, we'd need to also slide down the other toasts above it. So prefer not to touch animations in this phase of the PR.

src/experimental/Toast/ToastUI.tsx
maximedegreve: Haha you're playing with my emotions! I quickly mocked a few scenarios up and I'm leaning to not use a max-width on desktop breakpoints. On mobile breakpoints I would always set it to `100vw`.

Those are hot takes though I gave this only 10 minutes of thoughts and maybe there are more arguments than the ones described below.

Future

langermank: @maximedegreve my gut reaction is that the Toasts that respond to the content size look more natural compared to having a fixed width, so I agree with you! Thanks for those mockups ❤️ 

I don't know if we've designed for where they will appear (center top or bottom left) so that might change things if its not centered, but I think we should start with this design and go from there.

broccolinisoup:
I removed myself from the reviewers for now as the PR has been inactive for a while but please feel free to tag me again when you push more commits 🤗

maraisr:
Momentum will pick back up this week 🙌

maraisr:
Closing this, as there has been a decision made to NOT contribute this into Primer at this time.

echarles:
> Closing this, as there has been a decision made to NOT contribute this into Primer at this time.

Curious, will this be re-considered in the future?

langermank:
@echarles we are currently considering our options around messaging type components, and will have more clarity to share next year! Stay tuned.

PR #3488, opened 2023-07-05
chore: remove release blocker command

joshblack:
This PR removes the release blocker command as this format is unsupported at the moment.


broccolinisoup (approved):
Bummer 😢

PR #3491, opened 2023-07-06
Fix legacy task list position calculation

nicolleromero:
This PR fixes a bug where the combination of a tasklist + legacy task list in the side panel would cause a miscalculation of the checkbox position and result in the wrong checkbox being updated. Preventing checkboxes within code blocks from being included in the position calculation prevents this bug and other potential bugs (e.g., where a user includes the `- [ ]` checkbox notation within a code block).

Note: this fix was tested and works in both the previous and new issue viewer in Memex. 

- Closes https://github.com/github/issues/issues/6850

#### Before

https://github.com/primer/react/assets/25596676/76c5c8e7-639a-4049-855d-a7f66845589b

#### After

https://github.com/primer/react/assets/25596676/91521149-dcd7-4a7f-b979-286bf705e02b

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


JelloBagel (approved):
Thanks for fixing this bug! ✨ 

Looks good to me. Should we add a test for this?

src/drafts/MarkdownViewer/_useListInteraction.ts
```diff
@@ -4,6 +4,8 @@ import {ListItem, listItemToString, parseListItem} from '../MarkdownEditor/_useL
 
 type TaskListItem = ListItem & {taskBox: '[ ]' | '[x]'}
 
+const isCodeBlockDelimiter = (line: string) => line.trimStart().startsWith('```')
```
iansan5653, line 7: The [spec](https://github.github.com/gfm/#fenced-code-blocks) is a little looser than this, so there will be some types of code blocks that don't get caught.

They can be delimited by tildes instead of backticks:

```md
~~~
example
~~~
```

They can also have more than three backticks:

`````md
````
example
````
`````

In this case, you'd need to keep track of how long the starting delimiter is so that you can ensure the ending delimiter is the same length.

Comment (nicolleromero) [reply src/drafts/MarkdownViewer/_useListInteraction.ts:7]: Great point! I think for this specific bug fix (preventing the buggy interaction between legacy task lists and tasklists), the backtick delimiter does the trick, but I'm open to making the fix more robust (checking for tildes) to cover more edge cases.

**Comment by nicolleromero** [comment]
> Thanks for fixing this bug! ✨
> 
> Looks good to me. Should we add a test for this?

@JelloBagel Test added. :)

**Comment by joshblack** [comment]
FYI @siddharthkp for release next week this would be great to get in!

File: src/drafts/MarkdownViewer/_useListInteraction.ts [hunk 2]
```diff
@@ -4,6 +4,17 @@ import {ListItem, listItemToString, parseListItem} from '../MarkdownEditor/_useL
 
 type TaskListItem = ListItem & {taskBox: '[ ]' | '[x]'}
 
+// Make check for code fences more robust per spec: https://github.github.com/gfm/#fenced-code-blocks
+const parseCodeFenceBegin = (line: string) => {
+  const match = line.match(/^ {0,3}(`{3,}|~{3,})[^`]*$/)
```
iansan5653, line 9:
```suggestion
  const match = line.match(/^ *(`{3,}|~{3,})[^`]*$/)
```

Looks like the spec says that an opening fence can be indented an indefinite number of spaces, but the closing fence can only be indented up to 3.

src/drafts/MarkdownViewer/_useListInteraction.ts
```diff
@@ -4,6 +4,17 @@ import {ListItem, listItemToString, parseListItem} from '../MarkdownEditor/_useL
...
+  return match ? match[1] : null
+}
+
+const isCodeFenceEnd = (line: string, fence: string) => {
+  const match = line.match(new RegExp(`^ {0,3}${fence}${fence[0]}* *$`))
```
iansan5653, line 14:
```suggestion
  const match = line.match(new RegExp(`^ {0,3}${fence} *$`))
```
I think the closing fence needs to have exactly the same number of delimiters as the opening fence. For example, this is valid:

`````md
```one
````
we are still in block `one` here
```
`````

File: src/drafts/MarkdownViewer/_useListInteraction.ts [hunk 4]
```diff
@@ -4,6 +4,17 @@ import {ListItem, listItemToString, parseListItem} from '../MarkdownEditor/_useL
...
+  return match !== null
```
iansan5653, line 15: Nit: we could return `regex.test(line)` instead of `line.match(regex) !== null`. I think there might be a small performance benefit to that.

src/drafts/MarkdownViewer/_useListInteraction.ts
nicolleromero: The spec states that both an opening and closing fence can be indented no more than three spaces:

> A [fenced code block](https://github.github.com/gfm/#fenced-code-block) begins with a code fence, indented no more than three spaces.

> The closing code fence may be indented up to three spaces

I also tested and confirmed this is the case.

iansan5653: Huh, the spec has an example with more than three: https://github.github.com/gfm/#example-104

But I also tested and confirmed, this doesn't work. So maybe this is a problem with our own pipeline. Guess this is all good then :+1:

iansan5653: Side note, maybe we also should filter out lines starting with more than 3 spaces?

Ie, this is not a checklist item, but it's also not in a code block:

    - [ ] example

This came from this markdown:

```md
    - [ ] example
```

src/drafts/MarkdownViewer/_useListInteraction.ts
nicolleromero: Updated :)

src/drafts/MarkdownViewer/_useListInteraction.ts
iansan5653: Oh wow, you are totally right 👍

iansan5653 (approved):

siddharthkp (approved):

PR #3495, opened 2023-07-07
chore: leverage `clsx` for our conditional class names

maraisr:
Let's move to `clsx` as outlined in the css ADR. A lighter [more performant](https://github.com/lukeed/clsx/tree/master/bench) conditional class names library.

This will make authoring our styles with css-modules an absolute breeze.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/contributor-docs/adrs/adr-017-css.md
+++ b/contributor-docs/adrs/adr-017-css.md
@@ -73,8 +73,8 @@ The other reasons for change are to utilise css variables for theming and improv
 Code example:
 
 ```jsx
-import classNames from './ActionList.module.css'
 import clsx from 'clsx'
+import classNames from './ActionList.module.css'
 
 const Button = ({className, sx, ...props}}) => {
   return (
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "@types/styled-system": "^5.1.12",
         "@types/styled-system__css": "^5.0.16",
         "@types/styled-system__theme-get": "^5.0.1",
-        "classnames": "^2.3.1",
+        "clsx": "^1.2.1",
         "color2k": "^2.0.0",
         "deepmerge": "^4.2.2",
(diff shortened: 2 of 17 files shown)


contributor-docs/adrs/adr-017-css.md
```diff
@@ -73,16 +73,16 @@ The other reasons for change are to utilise css variables for theming and improv
 Code example:
 
 ```jsx
-import classNames from './ActionList.module.css'
 import clsx from 'clsx'
+import styles from './ActionList.module.css'
```
maraisr, line 77: @siddharthkp we may need to look at using [this](https://gist.github.com/developit/3436591071d6b8f12dc87ca53c20dfa6) to process this file.

In which case the import will look more like:

```ts
import * as styles './ActionList.module.css'
```

which we can build a quick eslint rule for as well.

siddharthkp: oh very cool! 

> import * as styles './ActionList.module.css'

as long as this is compatible with rollup, storybook and gatsby, we're good!

(Our gatsby docs throw a warning right now if you add a component with css to drafts/index 😢, it's something we have to figure out)

siddharthkp, line 77: pretty silly nitpick but, can we call the variable `classNames` instead of styles because it contains classNames not the actual styles

maraisr: yeah all good! probably just me. too many "classNames", that it's hard to quickly scan a file and know which are imported styles, which are className's in JSX and which are forwarded className props. 

but not offended one way or another! i'll rename 🎉

siddharthkp (approved):
Hurray!

PR #3496, opened 2023-07-07
chore(deps): update deps to use caret range for versions

joshblack:
Update the dependencies listed in `dependencies` to use a `^` range. This allows us as maintainers to specify the minimal satisfying version of a dependency and allow multiple versions to exist for downstream users.

This PR also adds these types of changes to our versioning docs.


langermank (approved):

broccolinisoup (approved):
Looks great! Thank you!

PR #3504, opened 2023-07-10
Add fixtures for #3044

siddharthkp:
Add fixtures for https://github.com/primer/react/issues/3044

Issue #3044, opened 2023-03-16
ActionList.Item using wrong colors and get invisible in high-contrast theme


pablonete:
### Description

We're using a `<ActionList.Item variant="danger" ... />` and it renders wrong in high-contrast themes (both dark and light) but not on any other theme:
![image](https://user-images.githubusercontent.com/579705/225661046-4826f9d9-c9ee-4cad-a1e6-5ec542074ab0.png)
Also wrong on hover:
![image](https://user-images.githubusercontent.com/579705/225661111-218b1f91-6900-480b-9982-7fbbedc13909.png)

Interestingly, text color gets fixed on resize (e.g. just opening DevTools) while icon color stays always wrong.

https://github.com/github/github/blob/8d9083549de626e8b3fa9667a2de09840f39642f/app/assets/modules/custom-properties/components/DefinitionsTable.tsx#L74

### Steps to reproduce

I'll go with the Monolith Storybook where it's easier:
1. `npm run storybook` in a `github/github` codespace
2. Once the page opens `:6006`, click Color scheme and use `light_high_contrast`
3. Click on "CustomPropertiesSchemaPage"
4. Click on the ellipsis button at the right to show the actions menu

Expected: label and icon readable
Actual: icon visible and label transparent, the opposite on mouse hover

Some repro videos on [slack](https://github.slack.com/archives/C01L618AEP9/p1678892029899589).

### Version

v35.21.0

### Browser

Chrome, Firefox

langermank:
I've got a fix for this in Rails, but looking at the [React code for ActionList](https://github.com/primer/react/blob/b55d8232ce9d990cc91862fff2f6f15ff2fdb65d/src/ActionList/shared.ts#L75) I'm not sure how to achieve the same thing. We need to target the `hover` state.

PR #3505, opened 2023-07-10
ActionList: Fix icon and description hover color for danger variant

siddharthkp:
- Fixes half of https://github.com/primer/react/issues/3044
- Merge base PR first: https://github.com/primer/react/pull/3504

Before:

leading icon and background have the same color, which makes the icon disappear, description does not have enough contrast with the background

After:

both icon and description have the same contrast as the title, making them easy to see


langermank (approved):

joshblack (approved):

siddharthkp (commented):
Love to see a clean diff :feels-good:

PR #3506, opened 2023-07-10
ci: update changesets action to build generated/components.json

joshblack:
Update the changesets validate job to build the project before running the validator action so that the `generated/components.json` file is available.

### DIFF
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -10,4 +10,15 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'npm'
+      - name: Install dependencies
+        run: npm ci
+      # This step generates the generated/components.json file expected by the
+      # validator action
+      - name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/changesets.yml
```diff
@@ -10,4 +10,15 @@ jobs:
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'npm'
+      - name: Install dependencies
+        run: npm ci
+      # This step generates the generated/components.json file expected by the
+      # validator action
+      - name: Build
+        run: npm run build
```
siddharthkp, line 23: Do we need the whole build or can we run the subset `script/components-json/build`?

joshblack: @siddharthkp I think we could get away with running a minimal build, let me know if that's preferred and I can update the command here!

broccolinisoup (approved):
Thanks for fixing that! Looks good to me, I liked Sid's suggestion around only running the component json build but up to you!

.github/workflows/changesets.yml
siddharthkp: I think so, but totally not a blocker, can follow up in a new PR!

PR #3507, opened 2023-07-10
ci: update ci workflow to remove app token requirement

joshblack:
Update the `build-components-json` job to no longer require an app token. This step is no longer needed as we do not need to commit a generated `components.json` file back to a PR. Instead, this step should validate and report errors if the format is incorrect.

Removing this requirement for the app token means that this step will not fail on PRs from forks or PRs from apps like dependabot.


broccolinisoup (approved):
Nice!

PR #3513, opened 2023-07-12
build(deps): bump `primer-changesets-cli` to 2.2.1

gr2m:
With this pull request, running `npx changeset` will run `npm run build:components.json` before trying to load component names from `generated/components.json`

screenshot of terminal showing the new behavior of npx changeset


gr2m:
> Just to confirm, this wouldn't run in the validate action right because we run it once already

This doesn't run any action, just a the `build:components.json` script locally on the users machine

siddharthkp (approved):

PR #3515, opened 2023-07-13
Address ActionList.Group heading semantics

broccolinisoup:
As a part of https://github.com/github/accessibility-audits/issues/2939

Remediation accessibility issues in `ActionList.Group`  heading semantics (follow [PVC semantics](https://primer.style/design/components/action-list/rails/alpha#heading) as it is been reviewed by accessibility)

### React Code

```jsx
<ActionList selectionVariant="multiple">
  <ActionList.Group title="Visible fields">
    <ActionList.Item>Option 1</ActionList.Item>
    <ActionList.Item>Option 2</ActionList.Item>
  </ActionList.Group>
  <ActionList.Group title="Hidden fields">
    <ActionList.Item>Option 3</ActionList.Item>
    <ActionList.Item>Option 4</ActionList.Item>
  </ActionList.Group>
</ActionList>
```

Only difference in this PR is the heading.

- Removing `role="presentation" ` and `aria-hidden="true"` and rendering the heading item as a heading not span.

### Semantics before
```html
<div>  // outer wrapper
  <ul>
    <li> // group 1 (Visible Fields)
      <div role="presentation" aria-hidden="true"> 👈🏻  This changed
	  <span id=":ro7:">Visible Fields</span> 👈🏻  This changed 
      </div>
      <ul aria-labelledby=":ro7:">....action list items....</ul>
    </li> 
    <li></li> // group 2 (Hidden Fields) - identical to group 1
  </ul>
</div>
```

### Semantics after
```html
<div>  // outer wrapper
  <ul>
    <li> // group 1 (Visible Fields)
      <div> 👈🏻  New update
	  <h3 id=":ro7:">Visible Fields</h3> 👈🏻  New update
      </div>
      <ul aria-labelledby=":ro7:">....action list items....</ul>
    </li> 
    <li></li> // group 2 (Hidden Fields) - identical to group 1
  </ul>
</div>
```

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Group.tsx
+++ b/src/ActionList/Group.tsx
@@ -75,22 +75,26 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
 
 export type HeaderProps = Pick<ActionListGroupProps, 'variant' | 'title' | 'auxiliaryText'> & {
   labelId: string
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
 }
 
 /**
- * Displays the name and description of a `Group`.
+ * Displays the title and description of a `Group`.
  *
- * For visual presentation only. It's hidden from screen readers.
  */
-const Header: React.FC<React.PropsWithChildren<HeaderProps>> = ({variant, title, auxiliaryText, labelId, ...props}) => {
+const Header: React.FC<React.PropsWithChildren<HeaderProps>> = ({
+  variant,
+  title,
+  auxiliaryText,
+  labelId,
+  as = 'h3',
+  ...props
+}) => {
   const {variant: listVariant} = React.useContext(ListContext)
 
-  const styles = {
+  const wrapperStyles = {
     paddingY: '6px',
     paddingX: listVariant === 'full' ? 2 : 3,
-    fontSize: 0,
-    fontWeight: 'bold',
-    color: 'fg.muted',
     ...(variant === 'filled' && {
       backgroundColor: 'canvas.subtle',
       marginX: 0,
@@ -100,10 +104,13 @@ const Header: React.FC<React.PropsWithChildren<HeaderProps>> = ({variant, title,
       borderColor: 'neutral.muted',
     }),
   }
+  const headerStyles = {marginBottom: 0, marginTop: 0, fontSize: 0, fontWeight: 'bold', color: 'fg.muted'}
 
   return (
-    <Box sx={styles} role="presentation" aria-hidden="true" {...props}>
-      <span id={labelId}>{title}</span>
+    <Box sx={wrapperStyles} {...props}>
+      <Box as={as} id={labelId} sx={headerStyles}>
+        {title}
+      </Box>
       {auxiliaryText && <span>{auxiliaryText}</span>}
     </Box>
   )
--- a/.changeset/forty-trees-explain.md
+++ b/.changeset/forty-trees-explain.md
@@ -1,0 +1,7 @@
+---
+'@primer/react': patch
+---
+
+ActionList group heading accessibility remediations
+
+<!-- Changed components: _none_ -->
--- a/src/NavList/__snapshots__/NavList.test.tsx.snap
+++ b/src/NavList/__snapshots__/NavList.test.tsx.snap
@@ -362,7 +362,7 @@ exports[`NavList renders a simple list 1`] = `
 `;
 
 exports[`NavList renders with groups 1`] = `
-.c8 {
+.c9 {
   display: -webkit-box;
   display: -webkit-flex;
   display: -ms-flexbox;
@@ -377,7 +377,7 @@ exports[`NavList renders with groups 1`] = `
   min-width: 0;
 }
 
-.c9 {
+.c10 {
   -webkit-box-flex: 1;
   -webkit-flex-grow: 1;
   -ms-flex-positive: 1;
@@ -410,12 +410,17 @@ exports[`NavList renders with groups 1`] = `
   padding-bottom: 6px;
   padding-left: 16px;
   padding-right: 16px;
+}
+
+.c4 {
+  margin-bottom: 0;
+  margin-top: 0;
   font-size: 12px;
   font-weight: 600;
   color: #656d76;
 }
 
-.c4 {
+.c5 {
   padding-inline-start: 0;
 }
 
@@ -426,7 +431,7 @@ exports[`NavList renders with groups 1`] = `
   padding-bottom: 8px;
 }
 
-.c6 {
+.c7 {
   position: relative;
   display: -webkit-box;
   display: -webkit-flex;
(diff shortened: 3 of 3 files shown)


src/ActionList/Group.tsx
```diff
@@ -75,22 +75,26 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
...
  *
- * For visual presentation only. It's hidden from screen readers.
  */
-const Header: React.FC<React.PropsWithChildren<HeaderProps>> = ({variant, title, auxiliaryText, labelId, ...props}) => {
+const Header: React.FC<React.PropsWithChildren<HeaderProps>> = ({
+  variant,
+  title,
+  auxiliaryText,
+  labelId,
+  as = 'h3',
```
siddharthkp, line 90: Ooof lots of considerations for this prop. I'm sorry, I missed the nuance of this in our call!


---

#### Default heading without customisation

In this PR, Header here accepts `as`, but the Group does not pass anything to it, so it always uses the default `h3`. Am I right in assuming that is intentional and we want to ship with a default h3 first, with the ability to customise it in a future PR?

Should there be a default value for Header or should we always ask? The heading html element is based on context, so it's hard to pick a default for it. In https://github.com/primer/eslint-plugin-primer-react/pull/66, we are throwing a warning/error if `as` isn't given to encourage a deliberate choice (👋 @TylerJDev).

Are we trading one set of a11y bugs (missing heading) with another? (incorrect heading level/order). That definitely can be an acceptable tradeoff we make that helps us *incrementally* improve accessibility.

---

#### Semantics for ActionList heading inside ActionMenu?

Finally, do these heading semantics apply when ActionList.Group is used inside a role="menu" like ActionMenu which is only visible when open?

---

#### Alternate strategy: Opt in

If we don't want to make that trade-off, we could make this feature opt-in and incrementally reduce number of missing heading tags one ActionList.Group instance at a time. 

So, a heading html element is only used when there is an explicit decision passed to the component (API to be decided), otherwise it's treated as presentational with aria-hidden.
   
API choices: 

1. `ActionList.Group title="Assignees" headingLevel="h2"`

2. `ActionList.Header` component that works with or without Group. Right now, if you want a title, you have to either bring your own `Heading` outside the `ActionList` or have to use `ActionList.Group` even if there is only one group

Suggestion:

_I'm not super happy with the API here, it's something we can continue to workshop_

```tsx
<ActionList> // list is labelled by the heading
  <ActionList.Header as="h3">Reviewers</ActionList.Header> 
  <ActionList.Item></ActionList.Item>
  <ActionList.Item></ActionList.Item>
  <ActionList.Item></ActionList.Item>
</ActionList>
```

```tsx
<ActionList aria-label="Reviewers"> // list needs explicit label if there is no list level Heading

  <ActionList.Group>   // group is labelled by the group heading
    <ActionList.GroupHeader as="h3">Recommended</ActionList.GroupHeader>
    <ActionList.Item></ActionList.Item>
    <ActionList.Item></ActionList.Item>
    <ActionList.Item></ActionList.Item>
  </ActionList.Group>
  
  <ActionList.Group>
    <ActionList.GroupHeader as="h3">Everyone else</ActionList.GroupHeader>
    <ActionList.Item></ActionList.Item>
    <ActionList.Item></ActionList.Item>
    <ActionList.Item></ActionList.Item>
  </ActionList.Group>
</ActionList>
```

broccolinisoup: >In this PR, Header here accepts as, but the Group does not pass anything to it, so it always uses the default h3. Am I right in assuming that is intentional and we want to ship with a default h3 first, with the ability to customise it in a future PR?

Sorry, this is purely my ignorance. 😭  `ActionList.Group` should definitely accept a prop that determines the heading level.

>Should there be a default value for Header or should we always ask?

This is a good question and we went through similar discussion on PageHeader ([issue reference](https://github.com/github/primer/issues/1571)) and decided to give a default value and limiting the `as` value to only from `h1` to `h6` and looking at the original semantic improvements PR, [same logic applies](https://github.com/primer/react/pull/2878/files#diff-eb8442fcde2f586579af39ae9d1a8d93431ce0c708fc834976c2abd9f7790093R13) there as well.

>We could chose to render a heading html element only when there is an explicit heading tag passed to the component (API to be decided), otherwise it's treated as presentational with aria-hidden.

I would love to hear what @TylerJDev or @kendallgassner thinks about it!

>Do the default heading semantics also apply when ActionList is used inside a role="menu" like ActionMenu or a Dialog, which is only visible when opened?

Great question! I am clueless here. cc @TylerJDev @kendallgassner 

I think it would be great to answer these questions first maybe before exploring the API options?

kendallgassner: If we introduce `ActionList.Heading` I would want to request that the component throws an errors if an `ActionList.Heading` is not added to a group -- would this be possible.

Also, to provide some background, these changes were requested after looking at how ActionList was used on GitHub's search results navigaion.

TylerJDev: > Do the default heading semantics also apply when ActionList is used inside a role="menu" like ActionMenu or a Dialog, which is only visible when opened?

As far as with `ActionMenu`, it doesn't seem to matter as much if they are exposed or not. I do not have much context on the original decisions, but here are my thoughts: 

* Focus navigation takes control of the user's arrow keys when entering the menu, meaning that the heading would be skipped while using arrow keys, and only navigable if the user explicitly moves to the heading.
* "Heading" contents are already announced when entering a group.
* The parent of the "Heading" (`<li>`) has a role of `none` meaning that it's ignored as an item of the menu.

The difference between `heading` usage here versus `ActionList`, is that the `Heading` will always be visible on the page when using `ActionList`, whereas in `ActionMenu`, you'll only be able to navigate to the heading when the menu is open, which isn't as helpful if you're navigating the page by headings.

Overall, I don't think it matters as much, but the benefits you gain are less than the ones in `ActionList`. I'd love any additional thoughts on this though!

One question is, does `ActionList` know when it's inside of an `ActionMenu`?

siddharthkp: for @TylerJDev

> For ActionList, it's definitely harder to make that guess, as it it could easily be an h2 or an h3 depending on the current heading hierarchy on the page. As a short term fix, I do think that setting a default heading wont hurt as much, as long as we allow developers to change it.

got it, good note!

> I realize making ActionList.Heading on its own would be a breaking change

Introducing a new API is fine by itself because it doesn't break anything on production, and would only apply to new instances. We would still need to decide what to do for existing instances though, do we set a default h3 or do we leave them as is for now

> One question is, does ActionList know when it's inside of an ActionMenu?

It does :)

siddharthkp: For @kendallgassner,

>  I would want to request that the component throws an errors if an ActionList.Heading is not added to a group -- would this be possible

Really depends on how we want to release/adopt this in dotcom.

Eventually, yes. But, throwing errors immediately is tricky because that would mean all the existing instances of groups on production would start throwing an error 😢. 

We could release this in 3 steps:
1. Add ActionList.Heading with no errors (backward compatible)
2. Add either title or ActionList.Heading to all instances of groups that are missing one (~30 of them in dotcom)
3. Add error for missing heading so that there are no new errors

kendallgassner: The three step process works for me! If we go this route can we create an issue to track this effort?

broccolinisoup:
@siddharthkp - No worries! It is actually a topic today for our call 😊 I'd like to chat with you about the next steps

PR #3517, opened 2023-07-13
Add aria attributes to the progress bar

kendallgassner:
### What

During passion week we got to play around with the ProgressBar component and we noticed that this component was missing some aria-attributes. I wanted to create a PR proposing adding these attributes to the component to improve accessibility.

I also added a prop `tabIndex` so the progressBar could be focused if a feature wanted to move focus to the component -- we found this to be essential when working on fileUpload guidance. 

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/ProgressBar/ProgressBar.tsx
```diff
@@ -39,20 +40,24 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
+    if (children && progress) {
+      throw new Error('You should pass `progress` or children, not both.')
+    }
+
+    const ariaAttributes = {
+      'aria-valuenow': typeof progress === 'number' ? Math.round(progress) : undefined,
+      'aria-valuetext': typeof progress === 'string' ? progress : undefined,
+      'aria-valuemin': 0,
+      'aria-valuemax': 100,
+      'aria-busy': progress !== 100,
```
broccolinisoup, line 54: Should we consider the string type here as well?

.changeset/plenty-mangos-sniff.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/react': patch
+---
+
+Adding aria-attributes and role to the ProgressBar component
+
+<!-- Changed components: _none_ -->
```
broccolinisoup, line 7:
```suggestion
<!-- Changed components: ProgressBar -->
```

src/ProgressBar/ProgressBar.tsx
broccolinisoup, line 1: We also have ProgressBar with multiple items like [here](https://primer-b1adc9747d-13348165.drafts.github.io/storybook/?path=/story/components-progressbar-features--multiple-items) and we don't get aria attributes  in these I realised. There is no usage of `ProgressBar.Item` though at dotcom so not sure if it would worth the effort. What do you think??

broccolinisoup (commented):
Thanks for this PR, it is a great improvement! 🙏🏻 I left some comments. Thanks for your time!

src/ProgressBar/ProgressBar.tsx
kendallgassner: @broccolinisoup Thanks for bringing this up. I ended up changing aria attributes a bit since t looks like this component really only works if progress is a number or a string of a number like ("66" or "100") Because the string is really just a number using `aria-valuetext` isn't actually necessary. 

To reduce confusion maybe we should only allow a number passed in?

src/ProgressBar/ProgressBar.tsx
kendallgassner: Good catch!!

To deal with multiple I now throw an error if children are passed in without `aria-valuenow` or `'aria-valuetext`:
   ```
    throw new Error(
      'Use the `aria-valuenow` or `aria-valuetext` so screen reader users can determine the `progress`.',
    )
    ```
    
    What do you think?

kendallgassner:
Note:

Because of `role: progressbar` AXE will break if no accessible name is provided (i.e.`aria-label` or `aria-labelledby`). I added an aria-label to all the examples but wonder if I should enforce it at a prop level or throw an error like we do witht he `children & progress`

src/ProgressBar/ProgressBar.tsx
broccolinisoup: Oh I see! I have a different take on this but I am not 100% sure, let me know what you think!

Seems like we are handling these aria attributes internally and I think it is a tad inconsistent to ask this from consumers when there are `ProgressBar.Item` children.

I am thinking maybe we would want to calculate the total progress internally and update the aria attributes as we do for the single progress item? Something along those lines (not a complete code)

```jsx
if (children) {
      let totalProgress = 0
      React.Children.toArray(children).forEach(child => {
        if (React.isValidElement(child) && child.props.progress) {
          totalProgress += parseInt(child.props.progress)
        }
      })
    }
```

And use that `totalProgress` to calculate the value of aria attributes. 

cc @siddharthkp @joshblack I would love their opinion before we decide anything concrete.

src/ProgressBar/ProgressBar.tsx
broccolinisoup: From the first look at the usages at dotcom, in practicality it seems fine but technically narrowing types are considered as `major` change. ([reference](https://github.com/primer/react/blob/main/contributor-docs/versioning.md#:~:text=minor-,The%20type%20of%20a%20prop%20is%20narrowed,-major)) And I don't know if there is any historical reasons to include `string` type in the first place. Sorry @kendallgassner we are dealing with a bit of an old component here I guess and lots to improve - I am trying to keep my review only specific to your changes but seems like some overlaps are inevitable 😬 I'll post internally as well to bring some more reviewers to get more context on this!

src/ProgressBar/ProgressBar.tsx
siddharthkp: Hi!

We could definitely loop through the children if the usage is predictable. However...

I couldn't find any [usage of ProgressBar](https://gh.io/primer-query?name=progressbar) with multiple items, so we might be able to slide it under the rug and deprecate it safely :)

src/ProgressBar/ProgressBar.tsx
```diff
@@ -23,6 +23,7 @@ const sizeMap = {
 type StyledProgressContainerProps = {
   inline?: boolean
   barSize?: keyof typeof sizeMap
+  tabIndex?: number
```
siddharthkp, line 26: Not sure if you need to add this explicitly 🤔 

If you pass tabIndex to ProgressBar, it passes it already passes it down to the html element


```jsx
<Progress tabIndex={0} />
```

TylerJDev:
> wonder if I should enforce it at a prop level or throw an error like we do with the children & progress

I think this would be a good idea to add as a prop so it's surfaced better, especially in the documentation! I'm assuming that if it's a required prop then each implementation would need to be remediated to include a label. But we'd want to include the label either way as AXE will surely throw an error if one isn't included 😕.

I do think that it would be beneficial to add it as a required prop, as the usage of the component seems to be limited across dotcom (for now). But would love additional thoughts on this @siddharthkp @broccolinisoup as y'all have more knowledge of the implications here.

src/ProgressBar/ProgressBar.tsx
kendallgassner: Amazing! I'll try this out.

src/ProgressBar/ProgressBar.tsx
```diff
@@ -39,20 +40,37 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
+    ) {
+      throw new Error(
+        'Use the `aria-valuenow` or `aria-valuetext` so screen reader users can determine the `progress`.',
+      )
+    }
+
+    const progressAsNumber = typeof progress === 'string' ? parseInt(progress) : progress
+
+    const ariaAttributes = {
+      'aria-valuenow': progressAsNumber ? Math.round(progressAsNumber) : undefined,
```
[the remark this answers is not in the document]
kendallgassner: Oh wait... I need it in the case that the value is a `number` not a string

src/ProgressBar/ProgressBar.tsx
kendallgassner: It looks like I need the type added:

A screenshot of a ProgressBar storybook story. This story shows the tabIndex with a red underline indicating a typeerror.


Is there a better type to use? Maybe `React.HTMLProps` ?

src/ProgressBar/ProgressBar.tsx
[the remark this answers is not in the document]
joshblack: @siddharthkp one note with `parseInt`, I believe it will infer the base if no `radix` is provided based on: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#radix

> If 0 or not provided, the radix will be inferred based on string's value. Be careful — this does not always default to 10!

src/ProgressBar/ProgressBar.tsx
joshblack: Maybe this could be an option?

```tsx
type StyledProgressContainerProps = {
  inline?: boolean
  barSize?: keyof typeof sizeMap
} & WidthProps &
  SxProp &
  React.HTMLAttributes<HTMLSpanElement>
```

joshblack (commented):
Thanks so much for making this PR! 🥳 Just left a couple of comments/questions

src/ProgressBar/ProgressBar.tsx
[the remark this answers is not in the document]
siddharthkp: Oh wow, good note!

src/ProgressBar/ProgressBar.tsx
```diff
@@ -39,20 +40,37 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
+    }
+
+    if (
+      children &&
+      typeof (rest as React.AriaAttributes)['aria-valuenow'] === 'undefined' &&
+      typeof (rest as React.AriaAttributes)['aria-valuetext'] === 'undefined'
+    ) {
+      throw new Error(
+        'Use the `aria-valuenow` or `aria-valuetext` so screen reader users can determine the `progress`.',
+      )
```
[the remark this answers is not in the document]
siddharthkp: +1 to warning

src/ProgressBar/ProgressBar.tsx
```diff
@@ -37,22 +38,37 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
-    throw new Error('You should pass `progress` or children, not both.')
-  }
-
-  return (
-    <ProgressContainer barSize={barSize} {...rest}>
-      {children ?? <Item progress={progress} sx={{backgroundColor: bg}} />}
-    </ProgressContainer>
-  )
-}
+export type ProgressBarProps = React.HTMLAttributes<HTMLSpanElement> & {bg?: string} & StyledProgressContainerProps &
```
kendallgassner, line 41: @joshblack I added `React.HTMLAttributes<HTMLSpanElement>` here!

joshblack (approved):
Thanks for doing this! 🥳 

I appreciated your comment about the `aria-label` above btw, I think our ideal for this stuff should be to move it into the type system but I think that would be considered a breaking change 😞 The approach here seems like a great incremental step and I'll make sure to capture this over in: https://github.com/github/primer/issues/1715 for v37!

lesliecdubs:
FYI, I pulled [this comment](https://github.com/primer/react/pull/3517#issuecomment-1636402127) out into an additional action item: https://github.com/primer/react/issues/3605

PR #3518, opened 2023-07-13
Let storybook be aware of postcss for `.module.css` files

maraisr:
Storybook's `addont-styling` will autowire webpack for us to leverage css modules within our stories, and component files. Pathing the way to start using those as opposed to styled-components. We also prefer this addon now over `storybook-css-modules`.

Huge props to @siddharthkp for starting this work over in https://github.com/primer/react/pull/3405

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


langermank (approved):
Approving but would love for @siddharthkp to also approve before merging :)

postcss.config.js
```diff
@@ -0,0 +1,18 @@
+/** @type {import('postcss-load-config').Config} */
+module.exports = {
+  plugins: [
+    require('postcss-preset-env')({
+      stage: 2, // https://preset-env.cssdb.org/features/#stage-2
+      features: {
+        'nesting-rules': {noIsPseudoSelector: true},
+        'focus-visible-pseudo-class': false,
+        'logical-properties-and-values': false,
```
broccolinisoup, line 9: Curious to hear why they are false 👀

broccolinisoup (commented):
This is very exciting! Thanks for this PR. It looks good to me but I am only watching the changes for now as I don't have much context and experience on this area as much as Sid or Katie has :)

postcss.config.js
maraisr: Took inspiration from https://github.com/primer/view_components/blob/main/postcss.config.js possibly @langermank can provide more insights

maraisr: I see @simurai/@jonrohan worked on this file and found some comments here: https://github.com/primer/view_components/pull/1638/files#r1032151244

Interesting stuff.

jonrohan: In some of our cases, the output was putting more CSS than we needed for the browsers we supported. I think the focus-visible-pseudo-class was also adding a `.focus` class which was meant to be paired with JS. So it was on, until disabled.

siddharthkp (approved):
Looks good! Ship it

PR #3521, opened 2023-07-14
ci: update vrt-runner to use docker image

joshblack:
Update `@playwright/test` and its usage in Docker and in CI. In CI, this removes the step where we install browsers and instead uses a docker image. This change helps to:

- Remove the install browsers step which can take up to 2min to complete
- In the happy path, this docker image should be cached and downloaded quickly
- When no cache is available, this step should mirror the performance of the typical browser install step

This change also updates the docker image used locally so that we can replicate tests in CI and in development.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,8 +161,6 @@ jobs:
         run: npm ci
       - name: Build storybook
         run: npx storybook build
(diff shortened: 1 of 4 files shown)


siddharthkp (approved):
Trust you with this!

Issue #3531, opened 2023-07-17
Tooltip is broken


maximedegreve:
### Description

The tooltip doesn't work and the example is broken as well.

<img width="942" alt="" src="https://github.com/primer/react/assets/980622/cab3c13e-9026-4ab8-a569-fd29c56e16c6">

### Steps to reproduce

1. Go to https://primer.style/react/Tooltip
2. Try out the example

### Version

Latest

### Browser

Safari

lesliecdubs:
@broccolinisoup could you take a quick look at this as you have been closest to ongoing work on `Tooltip` lately? 🙇🏻‍♀️  

If it seems like a more involved fix is needed (>1 hour of investigation), please leave any notes in this issue and we can decide how to prioritize from there.

PR #3534, opened 2023-07-18
Add `display: inline-block` to tooltip container

broccolinisoup:
Tooltip positioning is broken and particularly too bad on safari.

Closes https://github.com/primer/react/issues/3531

Docs permalink to test: https://primer-2fd91fe70f-13348165.drafts.github.io/Tooltip

#### Before
Screenshot of the tooltip component on safari. The button is hovered and the tooltip element itself is positioned on the button element itself and obscuring the content of the button and funny enough the tooltip's caret is positioned far right on the top of the button

#### After
Screenshot of the tooltip component on safari. The button is hovered and the tooltip element and its caret is correctly positioned to display the tooltip on the north direction which is right at the top of the triggering button

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/Tooltip.tsx
```diff
@@ -7,6 +7,7 @@ import {ComponentProps} from './utils/types'
 
 const TooltipBase = styled.span<SxProp>`
   position: relative;
+  display: inline-block;
```
broccolinisoup, line 10: We have `display:inline-block` on the pseudo elements of the tooltip ([ref](https://github.com/primer/react/blob/main/src/Tooltip.tsx#L63)) and I wonder if it was working before and got broken after some changes maybe on css? 🤷🏻‍♀️

siddharthkp: I feel a bit worried that we don't know how/when this broke, because I'm not sure if this will break something for dotcom 😅

Seems like the example in the docs has been broken since 35.2.x 🤔 https://primer-d4a57fc365-13348165.drafts.github.io/Tooltip

joshblack: One idea for our test suite could be to add in an e2e test for this and snapshot the story with the button focused. Definitely would not help with upstream testing, though, if that's what we're looking for.

broccolinisoup: I think e2e test is a great idea - at least we can catch thing in our side. I'll add that. No idea what to do catch it at dotcom though. @langermank brought up VRT tests at dotcom topic a few weeks ago to the eng sync and this might be something we can rely on in the (hopefully near) future!

PR #3536, opened 2023-07-18
Fix `ActionListItem` border-radius on hover for `full` list variant

langermank:
When ActionList is set to full variant, the border-radius is removed on items. I think this was either an oversight or we've since gone in another direction, but we don't have any designs (to my knowledge) where we want a full bleed on items.

### Screenshots

| Before | After |
| -- | -- |
| Action list item hover state before update | -- |

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -93,7 +93,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       lineHeight: TEXT_ROW_HEIGHT,
       minHeight: 5,
       marginX: listVariant === 'inset' ? 2 : 0,
-      borderRadius: listVariant === 'inset' ? 2 : 0,
+      borderRadius: 2,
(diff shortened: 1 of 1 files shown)


siddharthkp:
As far as I remember, this was intentional! 

The idea was that if ActionList is in a container and we want it to be flush with the container, it should have no margin and no border-radius. 

I'm not sure if this is still helpful today. But a good place to check would be to look at instances that use the full variant: https://musical-adventure-wlr3n3k.pages.github.io/?name=%22ActionList%22&attribute=%22variant%22

siddharthkp (changes requested):
Left a comment about checking usage before making this change

langermank:
Closing in favor of https://github.com/primer/react/pull/3556

PR #3389, opened 2023-06-07
Deprecate position prop for PageLayout.Pane, update stories and tests.

radglob:
Deprecates the `position` prop for `PageLayout.Pane`. Partially addresses https://github.com/github/primer/issues/1199 in conjunction with https://github.com/primer/react/pull/3386.

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated documentation
- [X] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3545, opened 2023-07-21
fix(PageLayout): Remove `warning` for the deprecated `position` prop

broccolinisoup:
[deprecating PageLayout.Pane PR](https://github.com/primer/react/pull/3389) introduces a warning then the prop is used and although this is a good practise, it is not dotcom-friendly. Because it fails dotcom tests since warnings are not expected.

For the time being, we are settling for removing the warnings and only going with the TS `deprecated` notice. This PR removes the warning and adds the TS `deprecated` notice.

For more context, please see the [slack thread here (GitHub stuff only)](https://github.slack.com/archives/C02NUUQ9C30/p1689797280576349)

### Screenshots

No visual change

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/PageLayout/PageLayout.tsx
```diff
@@ -563,11 +566,16 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
...
-    if (responsivePosition !== undefined) {
-      // eslint-disable-next-line no-console
-      console.warn(
-        'The `position` prop will be removed on the next major version. You should order your markup as you want it to render instead.',
-      )
+    if (__DEV__ && process.env.NODE_ENV !== 'test') {
+      // We don't want these warnings to show up on tests because it fails the tests (at dotcom) due to not extecting a warning.
+      // Practically, this is not a conditional hook, it is just making sure this hook runs only on DEV not PROD.
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      React.useEffect(() => {
```
siddharthkp, line 573: Do we still need to wrap it in a useEffect if it's  in `__DEV__ && process.env.NODE_ENV !== 'test'`?

src/SplitPageLayout/SplitPageLayout.test.tsx
```diff
@@ -17,7 +17,8 @@ describe('SplitPageLayout', () => {
   })
 
   it('renders default layout', () => {
-    const spy = jest.spyOn(console, 'warn').mockImplementation(() => {})
+    // Disabling them for now because we are not expecting them in the tests.
+    // const spy = jest.spyOn(console, 'warn').mockImplementation(() => {})
```
siddharthkp, line 21: Oooh yeah, that's not nice. Can we create set a variable in environment just like `__DEV__` called `__PRIMER_TEST__` (or process.env.primer) so that we can keep testing this in primer's test suite but not in dotcom's?

src/PageLayout/PageLayout.tsx
broccolinisoup: Hmmm, I imagine keeping the implementation same in prod and dev would help identify issues easier in general but it might not be needed. What do _you_ think?

The reason I added the useEffect to render the warning only when the dependency `responsivePosition` is changed rather than every render but happy to remove.

joshblack (approved):
+1 for TS `@deprecated`, I'd vote for just removing the warning in console personally and only use `@deprecated` till we find a better pattern for warnings here.

Here's the snippets we've been using for the changelog if it helps for messaging! (Also let me know if they're incorrect 😅)

```
  /**
   * @deprecated Use source order instead of relying on the `position` prop
   *
   * Before:
   * ```
   * <PageLayout>
   *   <PageLayout.Content />
   *   <PageLayout.Pane position="start" />
   * </PageLayout>
   * ```
   *
   * After:
   * ```
   * <PageLayout>
   *   <PageLayout.Pane />
   *   <PageLayout.Content />
   * </PageLayout>
   * ```
   */
```

src/SplitPageLayout/SplitPageLayout.test.tsx
broccolinisoup: Just for the records, as we settled for the TS `@deprecated` annotation for the time being, we are gonna worry about the warnings later 😄

PR #3547, opened 2023-07-21
Prevents body scroll when Dialog is open

mperrotti:
Sets `body` to `overflow: hidden` when a Dialog is mounted.

Closes https://github.com/github/primer/issues/2354

### How to test:

Add this line to a story:
```
<div style={{height: '120vh', width: 1}}>make it scroll</div>
```

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


mperrotti:
I'm not sure if this solution is SSR-compatible or not. I tried doing this with CSS using `overscroll-behavior`, but it didn't work.

mperrotti:
One caveat: this probably won't work on any scrollable elements within the body.

src/Dialog/Dialog.tsx
```diff
@@ -287,6 +287,21 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
...
+    // that there is already a modal open. In that case, we should bail
+    // so we don't re-enable scroll after the second dialog is closed.
+    if (document.body.style.overflow === 'hidden') {
+      return
+    }
+
+    document.body.style.overflow = 'hidden'
+
+    return () => {
+      document.body.style.overflow = ''
```
siddharthkp, line 301: Not blocking: should we keep a track of what it was set to before and reset to that, just in case it was scroll or clip instead of the default visible?

PR #3560, opened 2023-07-25
Fix source links  

josepmartins:
Fix missing and broken source links

Thread: https://github.slack.com/archives/GACAW0NPM/p1690295324163529


broccolinisoup (approved):
Oh no 😨 I think since we are moving components into folders, we didn't update the path accordingly! Thank you so much for fixing them! 🙏🏻

Issue #3554, opened 2023-07-24
SSR `FormControl` htmlFor attribute mismatch


GlebIrovich:
### Description

We are trying to [enable](https://github.com/issues/assigned) the SSR for our react app, however in dev mode I see a lot of errors related to the `htmlFor` mismatch between server and client.

![ htmlFor attribute mismatch stack trace screenshot](https://github.com/primer/react/assets/33176414/5e87df81-cabd-46bd-bf7c-5b6026e15068)

Following the suggestions in this [thread](https://github.slack.com/archives/C01L618AEP9/p1689860062470749), I manually added ids to every form control component. 

This resulted in a new issue: color mode mismatch:

```
react-dom.development.js:86 Warning: Prop `data-color-mode` did not match. Server: "dark" Client: "auto"
    at div
    at O (http://127.0.0.1/webpack/vendors-node_modules_primer_octicons-react_dist_index_esm_js-node_modules_primer_react_lib-es-fb382e-f86ce5f82dfb.js:16846:19291)
    at BaseStyles (http://127.0.0.1/webpack/vendors-node_modules_primer_octicons-react_dist_index_esm_js-node_modules_primer_react_lib-es-fb382e-f86ce5f82dfb.js:11952:5)
    at OptionalBaseStyles (http://127.0.0.1/webpack/ui_packages_react-core_deferred-registry_ts-ui_packages_react-core_AppContextProvider_tsx-ui_-36d03f-a38db5a39ee6.js:381:31)
    at Fe (http://127.0.0.1/webpack/vendors-node_modules_primer_octicons-react_dist_index_esm_js-node_modules_primer_react_lib-es-fb382e-f86ce5f82dfb.js:16846:16840)
    at ThemeProvider (http://127.0.0.1/webpack/vendors-node_modules_primer_octicons-react_dist_index_esm_js-node_modules_primer_react_lib-es-fb382e-f86ce5f82dfb.js:12149:3)
    at IntlProvider (http://127.0.0.1/webpack/ui_packages_react-core_deferred-registry_ts-ui_packages_react-core_AppContextProvider_tsx-ui_-36d03f-a38db5a39ee6.js:205:25)
    at RenderPhaseProvider (http://127.0.0.1/webpack/ui_packages_react-core_deferred-registry_ts-ui_packages_react-core_AppContextProvider_tsx-ui_-36d03f-a38db5a39ee6.js:855:32)
[...]

PR #3561, opened 2023-07-25
chore(deps): update @react-aria/ssr to ^3.5.0

joshblack:
Addresses the first half of: https://github.com/primer/react/issues/3554

When using components like `FormControl` which use `useSSRSafeId` under the hood, users may run into a hydration error when running in strict mode. This was due to the double-render behavior of StrictMode which would cause the ids to mismatch between the server, where things would only render once, and client.

This was addressed upstream by: https://github.com/adobe/react-spectrum/pull/3980 and should be available starting in `3.5.0`. This PR updates our dependency from `^3.1.0` to `^3.5.0` so that the minimum version of `@react-aria/ssr` includes this fix.

> **Note**
> In newer versions of this dependency, `SSRProvider` will warn as it is no longer needed in React 18


siddharthkp (approved):
Very very nice!

PR #3563, opened 2023-07-25
Add support for custom widths to PageLayout.Pane.

### LINKED ISSUE #3419: Allow custom widths for `PageLayout.Pane`
`PageLayout.Pane` only allows three values for the `width` prop: `"small" | "medium" | "large"`. The real default values of these keywords can be found here:

https://github.com/primer/react/blob/84993e7a557534047955f895e8a7889c0e6ce969/src/PageLayout/PageLayout.tsx#L526

These are fine when the pane is used for a vertical menu like it usually is, but there are other use-cases as well. For example, you might want to have a details/preview pane that takes up a third or more of the screen width by default. This should ideally be _much_ wider than `320px`. In this case, it would be very useful to be able to set a custom width.

For a resizable pane, we also need to be able to set minimum and maximum widths:

```tsx
width={{
  min: '400px',
  default: '550px',
  max: '700px'
}}
```

radglob:
Adds option to pass custom widths to the `width` prop of `PageLayout.Pane` to support specific widths outside those defined in the component.

Closes #3419

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge


src/PageLayout/PageLayout.tsx
```diff
@@ -466,6 +466,25 @@ Content.displayName = 'PageLayout.Content'
 // ----------------------------------------------------------------------------
 // PageLayout.Pane
 
+type Measurement = `${number}px`
```
joshblack, line 469: Is it possible that teams may want to include other units like `%` or `rem` or is `px` kind of what's expected here?

joshblack (approved):
Looks great! Just had a question about units but that's really it 👀

src/PageLayout/PageLayout.tsx
radglob: I could see that possibility. One of my concerns regarding having more options is it would complicate converting to numbers. The `updatePaneWidth` helper currently expects a number. I don't know a great way to handle that going from percentages, rems, ems, etc.

PR #3566, opened 2023-07-27
Added AnchoredOverlay story

green6erry:

Add storybook stories for AnchoredOverlay

Checklist

- [x]  Ensure your chosen component has a component directory in the src directory
- [x]  Ensure the component has a root story file in the component directory
- [x]  Ensure that the root story file exports a Default story
- [x]  (Optional) Create a Playground story
- [x]  Move all other stories for the component into a .features.stories.tsx file
- [x]  Opt the component into storybook tests
- [ ]  (Optional) Create visual regression tests for the component


green6erry:
Hey @mperrotti, heads up that this is ready for review!

src/AnchoredOverlay/AnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+    Story => {
+      return (
+        <ThemeProvider>
+          <BaseStyles>
+            <Story />
+          </BaseStyles>
+        </ThemeProvider>
+      )
+    },
+  ],
```
joshblack, line 63:
```suggestion
```

We should be good to remove these decorators as they're loaded in by default. Let me know if that causes any problems so!

src/AnchoredOverlay/AnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+      } else {
+        onClose()
+      }
+    },
+    [open, onOpen, onClose],
+  )
+
+  return (
+    <>
+      {args['With Anchor?'] ? (
```
joshblack, line 106: I bet we could get away with not having `With Anchor` in the playground and just making it an explicit feature story instead.

src/AnchoredOverlay/AnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <AnchoredOverlay
+          open={open}
+          onOpen={() => setOpen(true)}
+          onClose={() => setOpen(false)}
```
joshblack, line 110: It could be helpful to pull in `action` from `@storybook/addon-actions` to see when these events show up. Maybe something like:

```suggestion
          onOpen={() => {
            setOpen(true)
            action('onOpen')()
          }}
          onClose={() => {
            setOpen(false)
            action('onClose')()
          }}
```

src/AnchoredOverlay/AnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  return (
+    <AnchoredOverlay
+      open={open}
+      onOpen={() => setOpen(true)}
+      onClose={() => setOpen(false)}
+      renderAnchor={props => <Button {...props}>Button</Button>}
+    >
+      <Box width="100%" height="100%" backgroundColor="thistle" display="flex" flexDirection="column">
+        <img src={`//placekitten.com/200/300`} alt="kitten" />
```
joshblack, line 77: This made me laugh lol. Should we use a more generic placeholder instead of kittens as we make this more visible? The kitten part feels very dev-only / for testing instead of as documentation.

src/AnchoredOverlay/AnchoredOverlay.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+      }
+    },
+    [open, onOpen, onClose],
+  )
+
+  return (
+    <>
+      {args['With Anchor?'] ? (
+        <AnchoredOverlay
+          open={open}
```
joshblack, line 108: In the playground, it seems like `open` is automatically added to the list of controls but doesn't have an impact since we control it within the story. Is it something that we want to expose as a control or should we explicitly hide it?

joshblack (changes requested):
Thanks for porting this over to the new format! Left a couple of comments, let me know what you think 👀

src/AnchoredOverlay/AnchoredOverlay.stories.tsx
green6erry: Hey Josh, it looks like I came up with this, but it was actually from [9/1/21](https://github.com/primer/react/commit/83b888f0b325f9370ede9c5d7ac8f70702d6e8ea). Since it's been here for 2 years, it's your call if it's too dev-only. What's your stance?

mperrotti: I think we could do something like a fake "hover card" to make it feel more realistic.

Profile hover card

mperrotti (approved):

src/AnchoredOverlay/AnchoredOverlay.stories.tsx
joshblack: @green6erry makes sense! Totally get what you mean, those existing stories have been around for a while and they're just showing up here since you're moving stuff around.

I'm not sure what the expectations are for those stories under "behaviors", like Anchored Overlay. When looking at them, they look to be dev-only in that they're testing scenarios instead of serving as a reference for how to use the component.

I think if there is a quick way to simulate something like the hover card that'd be great but if not then we can always bring this in and change later 👍

joshblack (approved):

PR #3567, opened 2023-07-27
Revert "Upgrade styled-components to v5"

joshblack:
Reverts primer/react#3397

This PR was merged into the `main` branch and should be merged into the `next-major` branch 👍

PR #3569, opened 2023-07-28
chore(deps): update styled-components and related dependencies to v5

joshblack:
This is capturing the work from: https://github.com/primer/react/pull/3397 after the revert in: https://github.com/primer/react/pull/3567


broccolinisoup (approved):
Oh I didn't realised https://github.com/primer/react/pull/3397 was merging into `main`! The changeset was major so I missed that. Anyway, thanks for bringing the PR in again Josh!

PR #3579, opened 2023-07-31
Add aria-selected value to ActionList.Item.

radglob:
By adding `aria-selected` to the `<li role="option">` tags, NVDA and other screen readers will now announce that an option is "not selected" when `aria-selected=false`.

From what I've read, it is not the normal function of NVDA or JAWS to announce that an option is selected. Silence is the default behavior. https://stackoverflow.com/questions/71468332/select2-li-aria-selected-true-is-not-read-as-selected-by-screen-reader

Without `aria-selected`, state will not be announced for any option.

Closes https://github.com/github/accessibility-audits/issues/4484.

### Screenshots

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge


broccolinisoup:
FYI: I was just running tests with [the new primer action](https://github.com/github/github/blob/main/.github/workflows/primer-react-pr-test.yml) and used this PR to test its integration with github/github and all checks are passing 🤩 https://github.com/github/github/pull/283677

owenniblock:
I noticed which looking at this issue during a pairing session that the example: https://primer-9eeb77613b-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-examples--async-list-with-spinner isn't adding the `selected` attribute to the `option` element (the code is there but for some reason the HTML output doesn't include `selected`. We didn't investigate whether this was a code error in the example or an issue with the component but it might be worth checking why this is happening while we're looking at this aspect of the code.

owenniblock:
@radglob you are right! Sorry for the confusion. I'm slightly confused by the HTML specs not saying that [options are valid inside a listbox but only valid inside select, datalist or optgroup elements](https://html.spec.whatwg.org/multipage/form-elements.html#the-option-element) and yet the [WAI ARIA listbox examples behave as we plan to do stuff here](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/).

owenniblock:
[Link to code line I was referencing that tries to add selected](https://github.com/primer/react/blob/709024ff45a68c73ed6736e72d9db22b0a25cd33/src/ActionList/ActionList.examples.stories.tsx#L201)

radglob:
@owenniblock I think we still need that. In this case, `selected` is not intended to be the HTML attribute. It's just a prop here that we use to communicate whether or not an item is selected. We then use that value to populate the `aria-selected` attribute in the correct place, render the `ActionList.Selection` component properly, etc.

owenniblock:
@radglob I was just about to post to say sorry 😂 I totally misread what that code was going (I just went over the code in `Item.tsx`). Looks good to me, I even worked out why the aria and HTML specs were confusing me.

owenniblock (approved):

src/ActionList/Item.tsx
```diff
@@ -207,6 +207,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
           ref={forwardedRef}
           sx={merge<BetterSystemStyleObject>(styles, sxProp)}
           data-variant={variant === 'danger' ? variant : undefined}
+          {...(containerProps.role === 'option' && {'aria-selected': selected})}
```
joshblack, line 210: Small suggestion for clarity:

```suggestion
          aria-selected={containerProps.role === 'option' ? selected : undefined}
```

This format could make it more clear what we're setting and when along with avoiding the object spread / spread of the false-y value if `role` is not `option`

joshblack (changes requested):
Looking great! Just left a small comment for clarity, also wanted to ask if you could add a test case for this scenario? Would help a ton to make sure we don't regress in the future 🙏

src/ActionList/Item.tsx
radglob: Ah, yeah, that's better. I explicitly don't want `aria-selected` to get passed at all if the role isn't `option`.

joshblack (approved):

PR #3581, opened 2023-08-01
Add ActionList.Heading component

broccolinisoup:
As a part of https://github.com/github/accessibility-audits/issues/2939, we are adding `ActionList.Heading` component that has been proposed and discussed [here](https://github.com/github/primer/pull/2484) Most of the content below is from the [API PR](https://github.com/github/primer/pull/2484) - Thanks to @siddharthkp 🙌🏻  

**DESIRED SEMANTICS**

```html
<h3 id="heading-id" className="sx-generated">Issue actions</h2>
<ul role="list" aria-labelledby="heading-id">
  <li>
    <span><!-- leading visual --></span>
    <span>Lock conversation</span>
  </li>
  <li>
    <span><!-- leading visual --></span>
    <span>Unpin issue</span>
    <span><!-- description --></span>
  </li>
</ul> 
```

**CURRENT API**
```jsx
<Heading 
  as="h3"
  id="heading-id" 
  sx={{fontSize: 0, color:'fg.muted', paddingX: 3}}
>
  Issue actions
</Heading>

<ActionList role="list" aria-labelledby="heading-id">
  <ActionList.Item>
    {/* leading visual */}
    Lock conversation
  </ActionList.Item>
  <ActionList.Item>
    {/* leading visual */}
    Unpin issue
    {/* description */}
   </ActionList.Item>
</ActionList>
```

**NEW API**

```jsx
<ActionList role="list">
 <ActionList.Heading as="h3">Issue actions</ActionList.Heading>
 <ActionList.Item>
    {/* leading visual */}
    Lock conversation
 </ActionList.Item>
 <ActionList.Item>
    {/* leading visual */}
    Unpin issue
    {/* description */}
 </ActionList.Item>
</ActionList>
```

**Problems we are trying to solve:**
- ✅ Easy to miss, to not add heading at all (Missing heading) 
- ✅ Need to bring your own id + aria-labelledby (Heading but not connected)
- ✅ Need to bring your own heading styles (Inconistent heading styles)
- ✅ Need to bring your own padding to align with ActionList's default padding which depends on variant prop on ActionList (Inconsistent space between heading and list)

**Features/extensions that we are introducing to support:**
- ✅ Add arbitrary attributes to underlying element like id or data-test-id
- ✅ Make the heading visually hidden but readable by screen readers (see [Primer][React] ActionList semantics  accessibility-audits#2939 (comment) thread for context)
- ✅ Customize heading styles (not recommended, but for exceptions)

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.changeset/slimy-starfishes-agree.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/react': minor
+---
+
+ActionList: Add ActionList.Heading component
+
+<!-- Changed components: _none_ -->
```
broccolinisoup, line 7:
```suggestion
<!-- Changed components: ActionList -->
```

src/ActionList/Heading.tsx
```diff
@@ -0,0 +1,42 @@
...
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+} & SxProp
+
+export const Heading = forwardRef(({as: Component = 'h3', children, sx = defaultSxProp, ...props}, forwardedRef) => {
+  const innerRef = React.useRef<HTMLHeadingElement>(null)
+  useRefObjectAsForwardedRef(forwardedRef, innerRef)
+
+  const {headingId: headingId} = React.useContext(ListContext)
+
+  const styles = {
```
broccolinisoup, line 19: I copied the default group heading styles but this is probably not true? Is there a place I can refer to for the general actionlist heading styles?

src/ActionList/ActionList.features.stories.tsx
```diff
@@ -41,6 +42,109 @@ export const SimpleList = () => (
...
+    <ActionList.LinkItem href="https://github.com/primer/react/network/members">
+      <ActionList.LeadingVisual>
+        <RepoForkedIcon />
+      </ActionList.LeadingVisual>
+      <strong>225</strong> forks
+    </ActionList.LinkItem>
+  </ActionList>
+)
+
+export const WithHeadingInMenu = () => (
```
broccolinisoup, line 81: **Semantics**

The html structure is matching with the usage without a menu. Do we want anything behave differently when the actionlist is in a menu?

**Styles**

The default styles I provided is a bit broken due to not having a left and top space - should this be on consumer's side?? ([storybook reference](https://primer-c4f9c0d65c-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--with-heading-in-menu))

src/ActionList/Heading.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {useRefObjectAsForwardedRef} from '../hooks'
+import {ForwardRefComponent as PolymorphicForwardRefComponent} from '../utils/polymorphic'
+import {default as HeadingComponent} from '../Heading'
+import {ListContext} from './List'
+
+export type ActionListHeadingProps = {
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+} & SxProp
+
+export const Heading = forwardRef(({as: Component = 'h3', children, sx = defaultSxProp, ...props}, forwardedRef) => {
```
broccolinisoup, line 13: Does `role="list"` vs `role="listbox"` matter for `ActionList.Heading` implementation? 

From what I understand on [this comment](https://github.com/github/primer/pull/2484#issuecomment-1658368122), it only impacts the group headings as `ActionList.Heading` is rendered outside of the list anyway?

src/ActionList/ActionList.features.stories.tsx
siddharthkp: > The default styles I provided is a bit broken due to not having a left and top space - should this be on consumer's side?

Nah, I definitely thing we should handle this ourselves

src/ActionList/ActionList.features.stories.tsx
```diff
@@ -41,6 +42,109 @@ export const SimpleList = () => (
   </ActionList>
 )
 
+export const WithHeading = () => (
```
siddharthkp, line 45: I think we should add this heading to docs and all stories in another PR after this, to show that this is the recommended usage.

src/ActionList/Heading.tsx
siddharthkp: > it only impacts the group headings as ActionList.Heading is rendered outside of the list anyway?

yep, that is correct :)

src/ActionList/List.tsx
```diff
@@ -42,32 +47,42 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
       paddingY: variant === 'inset' ? 2 : 0,
     }
 
+    const [slots, childrenWithoutSlots] = useSlots(props.children, {
+      heading: Heading,
+    })
```
siddharthkp, line 52: Would this already work with eslint-plugin-primer-react or do we need to update that as well?

siddharthkp (commented):
Overall, looks like solid work! 

There are some tiny considerations here and there :)

src/ActionList/Heading.tsx
broccolinisoup: >Note: there is a full variant in ActionList with different padding that needs to be considered here

Great point I added this!

>I was expected [Figma: primer web](https://www.figma.com/file/GCvY3Qv8czRgZgvl1dG6lp/Primer-Web?type=design&node-id=6-24956&mode=design&t=295TfMTZyn1RluMk-0) to have the heading but it doens't 🤔

Yeah! I chatted with @langermank and she said she can help us with the design questions 🙌🏻 I'll collect them in one comment in this PR to make it easier.

src/ActionList/ActionList.features.stories.tsx
broccolinisoup: Oh that is right! All should have a heading - Not sure what I was thinking when I added a story with the heading! 😆  I can do them in this PR as well - to keep everything contained in one PR because they wouldn't be complex to review anyway.

src/ActionList/List.tsx
broccolinisoup: I think we would need to make some update on the eslint plugin. 

- We need to add `ActionList: ['ActionList.Header'],` in [this object](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/rules/direct-slot-children.js#L5)

Though, this rule will only prevent folks from using `ActionList.Heading` outside of `ActionList`, this won't warn if the `ActionList` is missing `ActionList.Heading` which I believe this is something we need as well? Can we do this with the type system at all? 🤔 Or should we look into adding another eslint rule for that too?

Also we can add `ActionList.Heading` to the [a11y-explicit-heading rule](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/rules/a11y-explicit-heading.js#L7) to restrict the values passed to `as`

src/ActionList/Heading.tsx
broccolinisoup: Ooo great point! I'll ask Kendall

src/ActionList/Heading.tsx
broccolinisoup: @langermank Here are the questions:

1) I copied the list heading styles from PVC but wanted to make sure if you have any feedback
	- inset variant (default):  [story](https://primer-c4f9c0d65c-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--with-heading)
	- full variant: [story](https://primer-c4f9c0d65c-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--with-heading-in-full-variant)
2) How are the heading styles look like when it is used in a menu? Do you have any feedback for this case? [story](https://primer-c4f9c0d65c-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--with-heading-in-menu)
3) We have [subtle](https://primer-c4f9c0d65c-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--group-with-subtle-title) and [filled](https://primer-c4f9c0d65c-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--group-with-filled-title) variants for group headings. Do we want the same thing for the list headings? PVC doesn't seem to have them.

Thank you! 🙌🏻

src/ActionList/Heading.tsx
broccolinisoup: We get this point clarified: 
> Semantic `<menu>`s don't have a place for headers within them, they should be aria-labelledby the menu button's name. 

by Chelsea ([reference](https://github.com/github/accessibility-audits/issues/2939#issuecomment-1675375647))

src/ActionList/List.tsx
broccolinisoup: Considering action lists in menus shouldn't have a heading ([reference](https://github.com/primer/react/pull/3581#discussion_r1312461027)), maybe we can enforece this with the eslint rule as well?

siddharthkp (approved):
Looks great to me. Left a couple non-blocking comments, approving in advance

src/ActionList/List.tsx
siddharthkp: > Also we can add ActionList.Heading to the [a11y-explicit-heading rule](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/rules/a11y-explicit-heading.js#L7) to restrict the values passed to as

Because ActionList.Heading is a new component, we can also use typescript to restrict allowed values for `as` :)

siddharthkp: > Considering action lists in menus shouldn't have a heading (https://github.com/primer/react/pull/3581#discussion_r1312461027), maybe we can enforece this with the eslint rule as well?

Good point. We could add an eslint rule for this, however, I'm not sure how effectively we would be able to catch incorrect usage given ActionList does not have to be a direct child of ActionMenu, it can be abstracted in a custom `UserList` or `BranchList` component that uses ActionList internally.

We also have the option of a dev environment runtime warning. Because [we know when an ActionList is contained inside ActionMenu](https://github.com/primer/react/blob/main/src/ActionList/Item.tsx#L64), we could use that to throw a warning

src/ActionList/Heading.tsx
```diff
@@ -5,54 +5,38 @@ import {useRefObjectAsForwardedRef} from '../hooks'
...
 
 export type ActionListHeadingProps = {
   as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
-  /**
-   * Style variations. Usage is discretionary.
-   *
-   * - `"filled"` - Superimposed on a background, offset from nearby content
-   * - `"subtle"` - Relatively less offset from nearby content
-   */
-  variant?: 'subtle' | 'filled'
```
siddharthkp: Yep! Agree with the decision of removing variant

src/ActionList/List.tsx
broccolinisoup: >Because ActionList.Heading is a new component, we can also use typescript to restrict allowed values for as :)

That is so true! I missed that 😅  which is [already restricted](https://github.com/primer/react/blob/5a316b5f8bd61915e8c23f22edd71e749072741d/src/ActionList/Heading.tsx#L11) anyway! 

>We also have the option of a dev environment runtime warning. Because [we know when an ActionList is contained inside ActionMenu](https://github.com/primer/react/blob/main/src/ActionList/Item.tsx#L64), we could use that to throw a warning

I love this! Thanks for the context.  I made it `invariant` to throw an error. Because first, it is a new component and it won't break anything. Second, when there is a heading the list will be labelled by the heading and completely disregard the button name. This could potentially be misleading. 

All that said, if you think that is unnecessary and very little chance of creating a misleading experience, happy to make it warning as well 😊

src/ActionList/Heading.tsx
```diff
@@ -0,0 +1,44 @@
...
+import {ListContext} from './List'
+import VisuallyHidden from '../_VisuallyHidden'
+
+export type ActionListHeadingProps = {
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+  visuallyHidden?: boolean
+} & SxProp
+
+export const Heading = forwardRef(
+  ({as: Component = 'h3', children, sx = defaultSxProp, visuallyHidden = false, ...props}, forwardedRef) => {
```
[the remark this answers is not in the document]
broccolinisoup: Ohh this is great! Yes I agree we should make it required!

broccolinisoup:
Integration PR is all green 🎉 https://github.com/github/github/pull/290302 merging the PR

PR #3582, opened 2023-08-01
Add `loading` prop for `Button` and `IconButton`

langermank:
Adds a `loading` state to both `Button` and `IconButton`. The markup follows the recommended path forward from accessibility design, and mirrors the implementation from the `Export CSV` button as part of GH org security coverage pages.

**Implementation details**

- We use a visually hidden label instead of changing the text of the button to handle text announcements better, and to not change the size of the button.
- `aria-disabled` instead of `disabled` to maintain focus management

**New props:**
`loading`: boolean
`loadingAnnouncement`: default to `Loading` or provide a custom message to be announced on ATs

Closes https://github.com/github/primer/issues/2680

### Screenshots

https://github.com/primer/react/assets/18661030/426caee6-4cb6-478f-b5f8-a84ba26e356e

https://github.com/primer/react/assets/18661030/67d49170-974b-4ed3-8e1b-022a6feabfca

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack:
This comment ended up being way too long, so sorry about that 🤦‍♂️ Let me know if it'd be helpful to bring these questions elsewhere in a different format.

---

@langermank was trying to catch up through the issue thread, let me know if I understand the use-case right 👀 

So we have scenarios, like export, where we need a button to:

- Communicate some initial state/action (like Export) that will trigger an asynchronous action (like downloading it)
- When someone activates the button, we need to:
  - Communicate that the action is happening
  - Mark the button as busy (aria-disabled) to prevent further actions from occurring
- When the action is complete, we meed to:
  - Communicate that the action is complete
  - Restore the button to an initial state (or success state if the text changes)
- If the action fails, we need to:
  - Communicate that the action has failed 
  - In some cases, focus should be restored to an element on the page

In some of the threads, it seemed like a reference for this was a multi function button kind of like: https://adrianroselli.com/2021/01/multi-function-button.html

Hope I understood that right, let me know if I missed anything 🤔 When communicating success/failure, it seems like this doesn't necessarily have to be through the live region itself but can be through some other indicator? Like a banner or some equivalent alert.

---

Couple of observations/questions when testing it out:

1. One thing I noticed is that I couldn't get VO to announce the loading state when activating with VO+Space 🤔 Only when using enter or space.

1. For `aria-busy`, I'm having a hard time understanding what it's being used for in this context. It seems like the live region will only ever be used to communicating the loading message but `aria-busy="true"` would flag that AT may ignore changes to content owned by that element until `aria-busy="false"` which is not happening in this example. 

    This may be similar to the observations in: https://github.com/github/accessibility/issues/4039#issuecomment-1690689155 if I'm reading it correctly. If we're looking to include this attribute, would it be appropriate on the `<button>` itself versus the live region?

1. When `loading` is true, should we prevent `onClick` events from firing? This would mirror the "it's disabled behavior" even though it's `aria-disabled` and not `disabled`.

1. Would it be helpful to have this as a standalone component? It seems like there are some patterns which we would like to enforce when using this component in order to make sure the different states are communicated.

1. One of the takeaways in the discussion made the following recommendation:

    > I recommend implementing a proper loading indicator, and keeping state out of buttons as much as possible.

    When are we deciding to do loading states in a button versus a loading indicator? Is the intent for this pattern to apply in the other scenarios brought up in the loading button proposal?

joshblack:
Also had a chance to record on VO on iOS and noticed something weird with the live region, seems like you can navigate to it with the virtual cursor somehow 🤔 

https://github.com/primer/react/assets/3901764/d38dc952-d208-4b18-a532-656d8eeb7392

It also seems to have the same challenge as VO on macOS where the live region is not announced. I also couldn't seem to get the `aria-describedby` message appended to the accessible name so it just announced the button as dimmed but not loading 🤔 Super strange.

src/Button/Button.features.stories.tsx
```diff
@@ -87,3 +87,28 @@ export const Small = () => <Button size="small">Default</Button>
...
+
+  return (
+    <Button loading={isLoading} onClick={handleClick} leadingIcon={DownloadIcon}>
+      Export
+    </Button>
+  )
+}
+
+export const LoadingCustomMessage = () => (
+  <Button loading loadingMessage="This is a custom loading message">
```
broccolinisoup, line 111: Is it an sr-only message?

src/Button/ButtonBase.tsx
```diff
@@ -62,40 +66,60 @@ const ButtonBase = forwardRef(
...
+          ref={innerRef}
+          data-block={block ? 'block' : null}
+          data-size={size === 'small' || size === 'large' ? size : undefined}
+          data-no-visuals={!LeadingVisual && !TrailingVisual && !TrailingAction ? true : undefined}
+          aria-disabled={loading ? true : undefined}
+          aria-describedby={loading ? 'loading-message' : undefined}
+        >
+          {Icon ? (
+            loading ? (
+              <Spinner size="small" />
```
broccolinisoup, line 83: Is the spinner size always small or should it depend on the size of the button? 🤔

src/Button/ButtonBase.tsx
```diff
@@ -62,40 +66,60 @@ const ButtonBase = forwardRef(
...
+        <StyledButton
+          as={Component}
+          sx={sxStyles}
+          {...rest}
+          ref={innerRef}
+          data-block={block ? 'block' : null}
+          data-size={size === 'small' || size === 'large' ? size : undefined}
+          data-no-visuals={!LeadingVisual && !TrailingVisual && !TrailingAction ? true : undefined}
+          aria-disabled={loading ? true : undefined}
+          aria-describedby={loading ? 'loading-message' : undefined}
```
broccolinisoup, line 79: Should this be?
```suggestion
          aria-describedby={loading ? loadingMessage : undefined}
```

src/Button/ButtonBase.tsx
langermank: Yeah, I think it looks good in the large button as well. I had the same thought 😆 I'll check again

src/Button/ButtonBase.tsx
langermank: I'm pretty sure `aria-describedby` is meant to reference the `id` of the message/string, not actually contain the string.

src/Button/Button.features.stories.tsx
broccolinisoup: Hmm good question. I think it is a good idea to indicate it in the name because I searched the code to see where we display the message and when we add this to the prop data it might not be obvious that it is sr only.

`hiddenLoadingMessage ` sounds like an option. `visuallyHiddenLoadingMessage` is too long. Is `sr-only` wording out-dated? 🤔 Could it be `srOnlyLoadingMessage`?

src/Button/ButtonBase.tsx
broccolinisoup: Ohhh sorry yes 100% 😄

src/Button/Button.features.stories.tsx
siddharthkp: Hi! 

The prop should ideally hint at what purpose it serves, but I think it doesn't have to describe implementation details. 

If `loadingMessage` feels misguiding as the message is not shown in the button, how do you feel about something along the lines of `loadingAnnouncement`? It signals a live region for screen readers without getting into what it doesn't do (visuallyHidden / srOnly)

src/Button/ButtonBase.tsx
```diff
@@ -21,6 +23,8 @@ const ButtonBase = forwardRef(
       size = 'medium',
       alignContent = 'center',
       block = false,
+      loading = false,
+      loadingMessage = 'Loading',
```
siddharthkp, line 27: Hi! 

Here's the new place to add props to the documentation: https://github.com/primer/react/blob/main/src/Button/Button.docs.json

siddharthkp:
The GIFs are so good! It looks so good how the spinner replaces the leadingVisual! 💯 

Should the spinner size change based on the size of the button? (the leadingVisual doesn't)

src/Button/Button.features.stories.tsx
langermank: Love that @siddharthkp! Thank you 😄

broccolinisoup: Love it too 😍

src/Button/ButtonBase.tsx
siddharthkp: Bonus of keeping the same size, it matches the size of leadingVisual for this satisfying interaction: 

https://github.com/primer/react/assets/18661030/426caee6-4cb6-478f-b5f8-a84ba26e356e

mperrotti:
@joshblack - thanks for your thorough review and focus on a11y! I've tried to respond to your feedback the best I could. @langermank - take a look at my responses and let me know if I got anything wrong or if you have additional feedback.

---

The assumptions in first part of [your comment](https://github.com/primer/react/pull/3582#issuecomment-1761942801) all sound correct to me.

---

> One thing I noticed is that I couldn't get VO to announce the loading state when activating with VO+Space 🤔 Only when using enter or space.

I have no idea why this happens, but it sounds like a VO bug. I'll check with a11y folks.

---

> For aria-busy, I'm having a hard time understanding what it's being used for in this context.

I'm also not sure we should be using `aria-busy` here - definitely not on the live region, but _maybe_ on the button? I'm going to take it out before sharing with a11y unless anybody objects.

---

> When loading is true, should we prevent onClick events from firing?

I think this is the correct thing to do. I'll confirm with a11y folks.

---

> Would it be helpful to have this as a standalone component?

I'm not sure we need this to be a standalone component, but I could be convinced with some use cases.

---

> When are we deciding to do loading states in a button versus a loading indicator?

I don't understand the difference between a "loading state in a button" and showing a loading indicator in a button. If you could clear this up for me, I can try and answer your questions.

---

> Also had a chance to record on VO on iOS and noticed something weird with the live region, seems like you can navigate to it with the virtual cursor somehow 🤔

> It also seems to have the same challenge as VO on macOS where the live region is not announced. I also couldn't seem to get the aria-describedby message appended to the accessible name so it just announced the button as dimmed but not loading 🤔 Super strange.

I'm not sure what's going on here. Seems like more VO bugs, but I'll check with a11y.

mperrotti:
If we decide that it's ok to remove click handlers when `loading` is passed, should we also remove click handlers if `aria-disabled` is passed manually? I vote yes. What do other reviewers think?

mperrotti:
I reverted my changes to the draft `Button` (aka `Button2`) since we talked about removing it until we're ready to switch to CSS modules.

If we change our minds, it should be easy to bring those changes back.

langermank:
@mperrotti @joshblack re: markup questions around aria-busy

I added it based on [this feedback](https://github.com/github/accessibility/issues/4039#issuecomment-1664511576). However, you can see there was some issues with it based on [this comment](https://github.com/github/accessibility/issues/4039#issuecomment-1690689155) as well. So it's not very clear what markup we are meant to use, and I think it will just need to go through another audit.

ericwbailey:
👋 Accessibility design FR here! I think axe is generating a false positive for the loading state color contrast evaluation. 

Given that the button would have both `aria-disabled="true"` applied to it an have click events also conditionally removed: I would consider it an [inactive control](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html#inactive-controls) and therefore technically exempt from contrast requirements.

mperrotti:
@ericwbailey - we haven't conditionally removed the `onClick` handlers just yet. I'm waiting for feedback on [this comment](https://github.com/primer/react/pull/3582#issuecomment-1830722517)

pksjce (commented):
Hi, I was able to get the button into this state if both leading and trailing visuals are present. Is this intentional?
image

mperrotti:
@pksjce - any idea why VRT and AAT tests are failing?

joshblack:
@mperrotti just pushed up a commit that changed the story id for the custom announcement that I think should do the trick! Will keep an eye on it to see if it works 👀

joshblack:
@mperrotti wanted to ask real quick what's the ideal end-to-end flow when using this button? Should there be some kind of announcement of success or failure that comes after the action is completed? If possible, it'd be great to have a story that shows an ideal flow/sequence.

mperrotti:
@joshblack - If the button is just an action like "Export CSV", we'll need an announcement. If the button is used to submit a form, users should follow the [interface guidelines for form validation](https://primer.style/ui-patterns/forms/overview#validation-on-submit).

I'm not sure if we need a success announcement, but I'm almost certain we need one for an error. cc @ichelsea, this week's a11y design FR

ichelsea (changes requested):
### Announcing and success states
@mperrotti hi! We actually just were talking about this in our toast/banner work today: [office hour notes](https://github.com/github/accessibility/issues/5251#issuecomment-1854745265).

This is going to be really context specific, but essentially the job to be done is that users are aware of what's happening on a page so that they can take their next steps. Success states usually don't need to be conveyed if there is some other indication of a success (toggle staying in a "on" position, creation of item taking you to the new page, etc.). 

However in the form of something that has the potential to take some time (like an export finishing), the successful completion should be announced and also conveyed somewhere, like a banner or whatever is decided to be the method of announcing that visually.

### Contrast
One other thing to note that I'm noticing here is that our loading states still have the disabled styling in which the text doesn't meet contrast. **The text and graphic indicators need to meet contrast**. One of the original reasons why I brought up the need for loading state is because we were using disabled styling and we needed to be able to see the button content. I synced up with @ericwbailey on this one to confirm. 
- WCAG states that controls not available for user interaction do not need to meet contrast.
- While this control is not available for user interaction, it _does_ conveys a status which would be equivalent to text conveying information on a page, and is important for understanding what's happening at any given moment (especially if an action is taking longer than expected). The original proposal included text outside of the button conveying this information which didn't work for various reasons, but deciding to put the text in the button makes the disabled styling rule void.  
  > The intent of this Success Criterion is to provide enough contrast between text and its background so that it can be read by people with moderately low vision 
  
  > Text that is decorative and conveys no information is excluded.

#### Summary
Regardless of the information living inside of an inactive control, it is considered information conveying a status to a user and it needs to meet contrast requirements.

langermank:
> The original proposal included text outside of the button conveying this information which didn't work for various reasons, but deciding to put the text in the button makes the disabled styling rule void.

@ichelsea there is no **visual** text change happening in a loading button. For buttons that have no visible label and no leading visuals, the content of the button is replaced with a spinner. The loading message is visually hidden. Similarly, for an icon only button the icon is replaced with a spinner and the loading message is visually hidden.

You can see this in action here: https://primer-1bba60ffcd-13348165.drafts.github.io/storybook/?path=/story/components-button-features--loading-with-leading-visual

Since the visible label text is not relevant to the loading state, does this change your opinion?

ichelsea:
> Since the visible label text is not relevant to the loading state, does this change your opinion?

@langermank It doesn't change. The loading spinner would need to make sure it passes graphical contrast still, but then it's possible that someone distracted leaves and then brings their focus back to the page who has low vision (or is outside) and has a cognitive disability (or brain fog due to a slew of reasons) and may not remember what text belongs with the loading state or what they were doing. 

WCAG woefully underrepresents cognitive disabilities in its current edition. 

Is there a strong case for making this text unreadable in this state?

langermank:
> Is there a strong case for making this text unreadable in this state?

@ichelsea simply put, I'm looking at the mechanics of the button– so given that it's `aria-disabled`, it makes sense to me that it looks `disabled`. I think there have been discussions recently about exclusively using `aria-disabled`, which would make the loading button a bit of an outlier. I'm fine with making the loading button retain the normal/rest state design.

mperrotti:
@ichelsea and @joshblack - I've addressed your feedback

ichelsea:
I do not know how to unrequest myself from a review I pressed the wrong button, apologies!

mperrotti:
@ichelsea - you can approve instead. I don't think it will let me merge until one of the people in the `react-reviewers` group approves 🙂

ichelsea (approved):

src/Button/ButtonBase.tsx
```diff
@@ -60,44 +71,82 @@ const ButtonBase = forwardRef(
...
-        )}
-      </StyledButton>
+            </>
+          )}
+        </StyledButton>
+        <VisuallyHidden>
+          <span aria-live="polite" id={loadingAnnouncementID}>
+            {loading && loadingAnnouncement}
+          </span>
+        </VisuallyHidden>
```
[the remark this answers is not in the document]
mperrotti: I ended up going to `1. Wrap these in a <div>`. This was causing `ButtonGroup` styles to not render correctly.

mperrotti:
The failing VRTs don't actually seem to be broken when I preview locally. Going to update snapshots...

Kapture 2023-12-18 at 17 53 40

mperrotti:
I don't know _how_, but it looks like this PR somehow fixed the issue where Dialog buttons weren't getting the right focus styles 😱

pksjce:
Hey @mperrotti - The loading state on a block button looks funky. Do we fix in a new PR or in this one?

Uploading Screen Recording 2023-12-21 at 6.04.47 pm.mov…

src/Button/Button.docs.json
```diff
@@ -18,15 +18,9 @@
...
-      "defaultValue": "'default'",
-      "description": "Change the visual style of the button."
-    },
-    {
-      "name": "size",
-      "type": "'small'\n| 'medium'\n| 'large'",
-      "defaultValue": "'medium'"
+      "name": "inactive",
+      "type": "boolean",
+      "description": "Whether the button looks visually disabled, but can still accept all the same interactions as an enabled button.\n This is intended to be used when a system error such as an outage prevents the button from performing its usual action.\n Inactive styles are slightly different from disabled styles because inactive buttons need to have an accessible color contrast ratio. This is because inactive buttons can have tooltips or perform an action such as opening a dialog explaining why it's inactive.\n If both `disabled` and `inactive` are true, `disabled` takes precedence."
```
pksjce, line 23: Is this because `main` has not been merged properly?

src/Button/Button.docs.json
```diff
@@ -39,6 +33,22 @@
...
     },
+    {
+      "name": "loading",
+      "type": "boolean",
+      "description": "When true, the button is in a loading state."
+    },
+    {
+      "name": "loadingAnnouncement",
+      "type": "string",
+      "description": "The content to announce to screen readers when loading."
```
pksjce, line 44:
```suggestion
      "description": "The content to announce to screen readers when loading. This requires `loading` prop to be true"
```

pksjce (approved):
LGTM 👍 
Two concerns
1. Bug with block Button that I've commented above.
2. Do you think `ConditionalWrapper` needs any more thinking as discussed in https://github.com/primer/react/pull/4095#issuecomment-1865233969 cc - @joshblack

langermank:
Moving this back to draft. I'm going to start a discussion for the team around aria-live region support potentially being a separate entity _outside_ of individual components, which would help us solve for the conditional wrapper issue!

mperrotti:
@pksjce - I think we need to give some more thought to how we're handling the ConditionalWrapper stuff. Here's the discussion @langermank mentioned: https://github.com/github/primer/discussions/2942

packages/react/src/Button/ButtonBase.tsx
```diff
@@ -64,45 +84,65 @@ const ButtonBase = forwardRef(
...
+          sx={sxStyles}
+          {...rest}
+          ref={innerRef}
+          data-block={block ? 'block' : null}
+          data-inactive={inactive ? true : undefined}
+          data-loading={Boolean(loading)}
+          data-no-visuals={!LeadingVisual && !TrailingVisual && !TrailingAction ? true : undefined}
+          data-size={size === 'small' || size === 'large' ? size : undefined}
+          aria-disabled={loading ? true : undefined}
+          aria-describedby={[loadingAnnouncementID, ariaDescribedBy]
```
mperrotti, line 99: Is it safe to only render `loadingAnnouncementID` when `loading === true`? Or do we need `loadingAnnouncementID` to be in `aria-describedby` on initial render for screen readers to pick this up?

jonrohan (approved):

mperrotti (approved):
~This has multiple approvals from other folks. Just adding my approval so I can merge.~

Merging isn't blocked because of reviews, it's blocked because `main` is locked right now. Will merge once it's unlocked.

PR #3586, opened 2023-08-02
Show underline on Link when hovered or focused.

radglob:
Resolves https://github.com/github/accessibility-audits/issues/4502 (and likely other issues).

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated documentation
- [X] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3598, opened 2023-08-07
Brings back #3586 (Show underline on Link when hovered or focused)

siddharthkp:
- Brings back https://github.com/primer/react/pull/3586
- Had to revert it to make way for a hotfix in release. primer/react#3590

Hi @radglob, sorry I need to cut a hotfix release and didn't want to push this change along with it.

Side note: This feels like a scary change for a patch release? Have we tested this in dotcom for instances where a Link component is used but it does not look like a blue text at all.


lesliecdubs:
@siddharthkp is this still valid? Note that `@radglob`'s contract has ended so we should take ownership of making this call if needed.

siddharthkp:
Superseded by https://github.com/primer/react/pull/3946

Issue #3604, opened 2023-08-08
Remove `MarkdownEditor` from `@primer/react`


iansan5653:
In order to more easily maintain and iterate on the `MarkdownEditor` component, we'd like to pull it back into GitHub's internal codebase. As this component never left `drafts`, we should be able to do this without a breaking change.

The tricky part here is going to be figuring out what to do with the internal dependencies of the component. There are two options, and the correct choice may be different for each dependency:

1. Remove it from `@primer/react` and pull it into the internal codebase along with the `MarkdownEditor`
2. Expose it publicly from `@primer/react` so it can be used in the internal codebase and in other places

I've taken a stab at going through the dependency tree and trying to figure out what to do with each thing. Ignoring things that are internal to the component or already public:

- Internal hooks:
  - [x] `useResizeObserver`: ~~Let's just expose this publicly by exporting it from `hooks/index.ts`. I think this is useful and low-risk.~~ Already exposed publicly.
  - [x] `useSlots`: I think we've previously avoided making this public because it was very unstable, but we seem to have settled on a pattern that works. This would be really useful to have access to in other codebases for making reusable components - let's expose this ~~in `hooks/index.ts` as well.~~ from drafts.
- Internal components:
  - [x] `VisuallyHidden`: This component is private in `@primer/react`, but is duplicated already in the Memex codebase. ~~I think we should go ahead and just make it public so we can remove the duplication. Maybe we can expose it from `@primer/react/drafts`?~~ We will duplicate this into the internal codebase.
  - [x] `InputLabel` (required by `MarkdownEditor.Label`): This is private in Primer and used in the editor to reuse the styling. I don't think it makes sense to expose this so there's not really a good path here except to just duplicate the styles.
- Draft components:
  - [x] `MarkdownViewer`: Migrate this along with `MarkdownEditor`. This is easy - it has no subdependencies that aren't public.
  - [x] `InlineAutocomplete`: Migrate
    - [x] This hooks in to `FormControl`, so we need to make that API public as well
- Utils:
  - [x] `character-coordinates`: Migrate, moving internal to `InlineAutocomplete`
- Publicly exposed draft hooks: these are public and used fairly often in other codebases. ~~I think we should just leave them behind and keep them publicly exposed in `@primer/react/drafts/hooks`:~~ We will migrate these as well:
  - [x] `useCombobox`
  - [x] `useDynamicTextareaHeight`
  - [x] `useIgnoreKeyboardActionsWhileComposing`
[...]

PR #3609, opened 2023-08-09
Expose `useSlots` from `drafts`

iansan5653:
Per https://github.com/primer/react/issues/3604, we'd like to eventually move the `MarkdownEditor` component back into the internal GitHub codebase. This component depends on the slots pattern, which is internal to this library. So as a prerequisite for moving the component, we need to expose the `useSlots` hook.

This will also be very useful for building other composable components.

Given the volatile nature of this hook, I've decided to expose it from `drafts` so it can still be changed and experimented with:

```ts
import {useSlots} from '@primer/react/drafts'
```


.changeset/thin-laws-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Expose `useSlots` from `drafts/hooks`
```
siddharthkp, line 5:
```suggestion
Expose `useSlots` from `drafts/hooks`

<!-- Changed components: None -->
```

siddharthkp: We have a new validation here! But, this should fix it :)

.changeset/thin-laws-beam.md
```diff
@@ -0,0 +1,7 @@
...
+
+<!-- Changed components: None -->
```
siddharthkp, line 7:
```suggestion
<!-- Changed components: _none_ -->
```

Got that special keyword wrong, this should be correct.

siddharthkp (approved):
Looks good now!

PR #3614, opened 2023-08-09
Automatically reposition `InlineAutocomplete` suggestions depending on available space

iansan5653:
`InlineAutocomplete` currently naively always places suggestions below the trigger character. This can cause the list to overflow the bottom edge of the screen.

The simple (and only slightly less naive) solution is to just put the suggestions on top if there's not enough space below. This is not an attempt to solve for every situation - the list can still overflow to the left, right, or top (if there's not enough vertical space above). And it won't automatically reposition on scroll. But it still solves for the vast majority of cases at very minimal performance cost. 

Here's a [Storybook demo](https://primer-4e8afe284c-13348165.drafts.github.io/storybook/?path=/story/components-forms-inlineautocomplete-features--auto-positioning) that places the textarea at the bottom of the screen.

https://github.com/primer/react/assets/2294248/57a0bea1-3e90-434e-8337-df12c69838e4

### DIFF
--- a/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
+++ b/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
@@ -108,7 +108,6 @@ const InlineAutocomplete = ({
           (getSelectionStart(inputRef.current) ?? 0) - showEventRef.current.query.length,
         )
       : {top: 0, left: 0, height: 0}
-  const suggestionsOffset = {top: triggerCharCoords.top + triggerCharCoords.height, left: triggerCharCoords.left}
 
   // User can blur while suggestions are visible with shift+tab
   const onBlur: React.FocusEventHandler<TextInputElement> = () => {
@@ -198,8 +197,7 @@ const InlineAutocomplete = ({
         inputRef={inputRef}
         onCommit={onCommit}
         onClose={onHideSuggestions}
-        top={suggestionsOffset.top || 0}
-        left={suggestionsOffset.left || 0}
+        triggerCharCoords={triggerCharCoords}
         visible={suggestionsVisible}
         tabInsertsSuggestions={tabInsertsSuggestions}
       />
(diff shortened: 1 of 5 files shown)


paxos:
This is great :+1:

I think it would be even better if we could specify an anchor point on top of that. In the demo video you posted, you see the position twitching around if there is only one auto complete item (that does it).
By specifying an anchor direction, we could avoid that (e.g. things would float always above the cursor).

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -52,6 +59,18 @@ export type InlineAutocompleteProps = {
...
+  /**
+   * Control which side of the insertion point the suggestions list appears on by default. This
+   * should almost always be `"below"` because it typically provides a better user experience
+   * (the most-relevant suggestions will appear closest to the text). However, if the input
+   * is always near the bottom of the screen (ie, a chat composition form), it may be better to
+   * display the suggestions above the input.
+   *
+   * In either case, if there is not enough room to display the suggestions in the default direction,
+   * the suggestions will appear in the other direction.
+   * @default "belo"
```
joshblack, line 71:
```suggestion
   * @default "below"
```

joshblack (approved):

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -52,6 +59,18 @@ export type InlineAutocompleteProps = {
...
+   */
+  suggestionsPlacement?: SuggestionsPlacement
```
siddharthkp, line 73: (Not blocking suggestion) Should we lock to the suggestion if it's given? For example, If we start at "above", stick to above even when the list becomes smaller and there is now enough space below. Creates a more consistent experience.

iansan5653: I think that works for if the list becomes shorter while it's open, but not if the list becomes larger. If we try to distinguish between those cases, the logic becomes pretty challenging and I don't know if it's worth it considering how uncommon that will actually be in practice.

I tested in VSCode just now and the suggestions list can flip as you type. I've personally never noticed that before, so I think that this is probably not a very common case.

PR #3619, opened 2023-08-10
fix(sx): support Primer colors for `border*Color` properties

### LINKED ISSUE #3618: TypeScript: Primer colors are not included in typeahead for `border*Color` properties in `sx`
### Description

Current 

Screenshot of Box component code

Expected

Screenshot of Box component code

The fix is straight forward (last famous words ...), I'll send a pull request in a minute

### Steps to reproduce

See above

### Version

v35.25.1, and in `main`

### Browser

_No response_

gr2m:
Closes #3618

### Screenshots

Before

Screenshot of Box component code

After

Screenshot of Box component code

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/angry-boats-wash.md
+++ b/.changeset/angry-boats-wash.md
@@ -1,0 +1,7 @@
+---
+'@primer/react': patch
+---
+
+Include `border*Color` properties in sx to support named Primer colors
+
+<!-- Changed components: _none_ -->
--- a/src/sx.ts
+++ b/src/sx.ts
@@ -1,10 +1,12 @@
 import css, {SystemCssProperties, SystemStyleObject} from '@styled-system/css'
 import {ThemeColorPaths, ThemeShadowPaths} from './theme'
-import {ColorProps, ShadowProps} from 'styled-system'
+import {ColorProps, BorderColorProps, ShadowProps} from 'styled-system'
 import merge from 'deepmerge'
 
 export type BetterCssProperties = {
   [K in keyof SystemCssProperties]: K extends keyof ColorProps
+    ? ThemeColorPaths | SystemCssProperties[K]
+    : K extends keyof BorderColorProps
     ? ThemeColorPaths | SystemCssProperties[K]
     : K extends keyof ShadowProps
     ? ThemeShadowPaths | SystemCssProperties[K]


.changeset/angry-boats-wash.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/react': patch
+---
+
+Include `border*Color` properties in sx to support named Primer colors
+
+<!-- Changed components: _none_ -->
```
gr2m, line 7: In this case I think `_all_` might have been the better option. If you agree and see that happening frequently let me know and I'll add the option to the changesets CLI

broccolinisoup: I had a few PRs that I thought "all" would make more sense but some were a bit vague 🤔 Similar to this case like it does affect almost all components but it doesn't really feel a component change to me maybe more of a system change that can be communicated with a different keyword? Curious to hear what you think!

src/sx.ts
```diff
@@ -1,10 +1,12 @@
 import css, {SystemCssProperties, SystemStyleObject} from '@styled-system/css'
 import {ThemeColorPaths, ThemeShadowPaths} from './theme'
-import {ColorProps, ShadowProps} from 'styled-system'
+import {ColorProps, BorderColorProps, ShadowProps} from 'styled-system'
 import merge from 'deepmerge'
 
 export type BetterCssProperties = {
   [K in keyof SystemCssProperties]: K extends keyof ColorProps
+    ? ThemeColorPaths | SystemCssProperties[K]
+    : K extends keyof BorderColorProps
```
mattcosta7, line 9: @gr2m I know this is late, but I wonder if there's some tests we can use to ensure these don't regress, or to ensure they're fully expressive of options that work, and don't include options that don't work?

gr2m: Also I feel if we added tests, we would only test behavior of styled-system?

mattcosta7: there are some other type-tests in the repo.  I think since we're exposing a type that used in many places, it would be good to have some way to validate it, and ensure it's inclusive of all possible options.

I don't think tsd is used in the repo, but might be valuable to add something like it for type only assertions?

right now I think most type tests just ensure functions compile correctly - there's a bunch of files with `.types.test` suffixes where we do type only tests

gr2m: okay cheers I'll send a follow up pull request, but can't say when I get to it

PR #3621, opened 2023-08-11
Make `FormControl` externally extensible (via HOC)

iansan5653:
### Motivation

`FormControl` supports an 'auto-wiring' functionality where it automatically applies certain props to the wrapped input in order to make accessibility easier. For example, it automatically applies an `id` prop (if none is provided) to link the wrapped input to the `Label`.

At present, this is only supported for a limited subset of components, defined internally:

https://github.com/primer/react/blob/65b888c278bb3130fef449163e787a475c971b38/src/FormControl/FormControl.tsx#L58-L67

Consumers of the library cannot add to this list, so it's impossible for consumers to create their own components that support integration with `FormControl`.

As I work on moving the `InlineAutocomplete` component out of this repository, we either need to make this functionality externally extensible or we have to break the `InlineAutocomplete` component and require it to be manually wired up whenever used. I think the first one is much more desirable as it's generally useful for all component authoring.

### Solution

To solve this, I added a new unique `Symbol` that is internal to `@primer/react`. Component authors must explicitly attach this `Symbol` to their components by wrapping the component in the `FormControl.autoWirable` higher-order component (HOC). This is the only way a component can be registered. The `FormControl` component will now simply look for the presence of this symbol on the `type` of each child element.

Example:

```tsx
const RedTextInput = FormControl.autoWirable(
  (props) => <TextInput sx={{backgroundColor: 'red'}} {...props} />
)
```

This HOC is fully compatible with `memo` and `forwardRef`.

As described in the documentation comments, wrapping a component in this HOC is a promise that the component will forward the desired props to the underlying input. This cannot be enforced by types or even runtime checks - there's no way to know that the underlying component is actually forwarding all the props. This is why it must be an explicit, well-documented decision.

The HOC does have some type-safety: it won't accept components that are _incompatible_ with the forwarded props. For example, a type error will be raised if a component has the prop `{required: number}` because `number` is not compatible with the `boolean` type that is forwarded.


src/FormControl/FormControl.tsx
```diff
@@ -197,25 +182,16 @@ const FormControl = React.forwardRef<HTMLDivElement, FormControlProps>(
...
-                    disabled,
-                    validationStatus,
-                    ['aria-describedby']: [validationMessageId, captionId].filter(Boolean).join(' '),
-                  },
-                  InputComponent.props,
-                ),
-              )}
-            {childrenWithoutSlots.filter(
-              child =>
-                React.isValidElement(child) &&
```
mattcosta7, line 216: similar question re safety here for consumers, but maybe not a big deal

src/FormControl/auto-wirable.ts
```diff
@@ -0,0 +1,41 @@
...
+
+type AutoWirableElement = ReactElement<FormControlForwardedProps, AutoWirableComponent>
+
+/**
+ * Mark a component to indicate that it supports `FormControl` autowiring by forwarding the props in
+ * `FormControlAutowireComponentProps` to an underlying form element or component. This is useful when
+ * wrapping/extending Primer form controls to make them easier to use.
+ */
+// use explicit return type to hide the `supportsAutoWiring` symbol from the public API
+export function autoWirable<P extends FormControlForwardedProps>(component: ComponentType<P>): ComponentType<P> {
```
mattcosta7: 🚀

src/Select.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import styled from 'styled-components'
 import TextInputWrapper, {StyledWrapperProps} from './internal/components/TextInputWrapper'
+import FormControl from './FormControl/FormControl'
```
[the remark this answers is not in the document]
iansan5653: I think we definitely prefer `FormControl.autowirable` and that's the only way it's accessible via the public API. There are just two internal components (radio & checkbox) that are still referenced inside the body of `FormControl` so I had to avoid the circular reference.

I might dig into why those two still need to be explicitly referenced though -- maybe that can be abstracted as well.

joshblack (commented):
Love how easy this makes it to participate in the FormControl component 🔥 Seems to make it dead simple to just call `FormControl.autowirable` on your component and you're good to go.

In terms of approach, could you share more about the pros/cons of this approach versus others we could take? For example, if we exposed something like a `useFormControl` hook that one could call to get access to the props for the input? 

For context, one concern I have (generally) is with strict Parent > Child relationships and I'd personally love to explore how we could support more flexibility when supporting the `children` prop. However, totally understand if that's just not something we really could ever explore here due to existing limitations or functionality we have to support.

iansan5653:
Here's the hook-based approach: https://github.com/primer/react/pull/3632

Honestly, I say we close this PR and go with that approach. It's much cleaner code and arguably just as nice of an API. Great suggestion @joshblack

joshblack:
Wow! Thanks so much for opening up that exploration for it, can't wait to check it out 👀 What you shared as next steps sound great to me @iansan5653, I'll make sure to checkout the PR tomorrow if that works timing-wise!

PR #3630, opened 2023-08-14
test: update playwright/test to 1.37

joshblack:
Update `@playwright/test` to [v1.37.0](https://github.com/microsoft/playwright/releases/tag/v1.37.0). This release includes support for [merging reports](https://playwright.dev/docs/test-sharding#merge-reports-cli) which allows us to combine the reports of all of our runners into a single artifact that we upload. As a result, when reviewing tests one can view the single artifact instead of downloading a specific runner artifact.

#### Changed

- Update `@playwright/test` to v1.37.0
- Update playwright config to emit blob reports in CI
- Update CI workflow for vrt-runner, aat-runner to upload blob reports
- Update the corresponding rollup job to merge blob reports from the runners
- Update the aat runner to run using the playwright docker image for consistent snapshots between local and CI

#### Removed


siddharthkp (approved):
LGTM! Thanks for regularly keeping these up to date!

PR #3632, opened 2023-08-14
Make `FormControl` externally extensible (via hook)

iansan5653:
Followup from discussion in https://github.com/primer/react/pull/3621

### Motivation

`FormControl` supports an 'auto-wiring' functionality where it automatically applies certain props to the wrapped input in order to make accessibility easier. For example, it automatically applies an `id` prop (if none is provided) to link the wrapped input to the `Label`.

At present, this is only supported for a limited subset of components, defined internally:

https://github.com/primer/react/blob/65b888c278bb3130fef449163e787a475c971b38/src/FormControl/FormControl.tsx#L58-L67

Consumers of the library cannot add to this list, so it's impossible for consumers to create their own components that support integration with `FormControl`.

As I work on moving the `InlineAutocomplete` component out of this repository, we either need to make this functionality externally extensible or we have to break the `InlineAutocomplete` component and require it to be manually wired up whenever used. I think the first one is much more desirable as it's generally useful for all component authoring.

### Solution

I exposed a new hook on the `FormControl` API: `FormControl.useForwardedProps`. This hook is given the external `props` of the component and handles merging them together and building the `aria-describedby` attribute. 

I honestly think this is much cleaner a far preferable solution vs the one proposed in https://github.com/primer/react/pull/3621. There are few downsides, and it's much more idiomatic React. We can avoid using the `Children` API and the relationship between `FormControl` and child feels much more natural and robust.

A huge advantage of this approach is that we don't need to forward props down through children - since it's context-based, supported components can be deeply nested.

The only downsides I can think of are:

1. It's a little less readable to have to wrap the `props` in a hook. This is pretty minor though and it's not something most users will have to care about at all
2. We have no way of preventing the `id` from getting assigned to multiple components. We really want to error if two components in one `FormControl` are both calling the hook, but I can't think of a clean way to do that. However, this is a pretty rare edge case and I don't think most users will actually try rendering multiple components in a single `FormControl`


src/FormControl/FormControl.tsx
```diff
@@ -55,49 +43,17 @@ const FormControl = React.forwardRef<HTMLDivElement, FormControlProps>(
...
-      }
-      if (inputProps?.required) {
-        // eslint-disable-next-line no-console
-        console.warn(
-          `instead of passing the 'required' prop directly to the input component, it should be passed to the parent component, <FormControl>`,
-        )
-      }
-    }
+    const isChoiceInput = childrenWithoutSlots.some(
+      el => React.isValidElement(el) && (el.type === Checkbox || el.type === Radio),
```
iansan5653, line 55: There's no good way to detect if the wrapped component is a `Checkbox` or `Radio` without referencing them directly. I don't like this since it means we still have to have this backwards data flow from children to parent, but I don't have an elegant solution to that problem.

src/FormControl/FormControl.tsx
```diff
@@ -137,30 +93,12 @@ const FormControl = React.forwardRef<HTMLDivElement, FormControlProps>(
...
-                    disabled,
-                    ['aria-describedby']: captionId as string,
-                  },
-                )}
-              {childrenWithoutSlots.filter(
-                child =>
-                  React.isValidElement(child) &&
-                  ![Checkbox, Radio].some(inputComponent => child.type === inputComponent),
-              )}
-            </Box>
```
iansan5653, line 163: This change also fixes an apparently previously overlooked bug where `horizontal`-layout `FormControl` would not forward the `validationStatus` or `required` props

src/FormControl/_FormControlContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (externalProps.required) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `instead of passing the 'required' prop directly to the input component, it should be passed to the parent component, <FormControl>`,
+    )
+  }
+
+  return {
+    disabled: context.disabled,
+    id: context.id,
```
iansan5653, line 53: One downside of this is that if two components inside the same `FormControl` call this hook, they could have the same ID. But that's an invalid structure anyway - `FormControl` is singular so this shouldn't be a problem.

src/FormControl/FormControl.tsx
```diff
@@ -232,4 +152,5 @@ export default Object.assign(FormControl, {
   Label: FormControlLabel,
   LeadingVisual: FormControlLeadingVisual,
   Validation: FormControlValidation,
+  useForwardedProps: useFormControlForwardedProps,
```
iansan5653, line 155: It's debatable whether this should even be publicly exposed with this approach:


On one hand, this is almost never necessary. Prior to this change, wrapping a Primer form input in another component (ie, if you want to make a `NumericTextInput` component) would have broken the `FormControl` wiring. But with context, this doesn't happen. As long as the underlying component is a Primer form input, it doesn't matter how deeply nested it is. If we don't expose this hook, we can push users to always build around Primer components and not HTML from scratch.

On the other hand, publicly exposing it allows consumers to build custom form inputs from plain HTML. Maybe that's not something we want to prevent. For example, they could build a new `RangeInput` component, which Primer doesn't have yet.

src/FormControl/_FormControlContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+      `instead of passing the 'disabled' prop directly to the input component, it should be passed to the parent component, <FormControl>`,
+    )
+  }
+
+  if (externalProps.required) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `instead of passing the 'required' prop directly to the input component, it should be passed to the parent component, <FormControl>`,
+    )
+  }
```
iansan5653, line 49: I wonder if these warnings should be suppressible (or just removed altogether). I kept them to be consistent with current behavior, but I do think there are valid extensibility cases where you might not want these warnings.

For example, a `DatePicker` might have a `TextInput` as its anchor. The `DatePicker` may want to control exactly how these forwarded props are handled, since a datepicker is a much more complicated component than just a text input. So the `DatePicker` would override the `TextInput` forwarded props, causing the warning to occur because the `TextInput` is still calling this hook.

One possible way to suppress this would be for the `DatePicker` to wrap its `TextInput` in a `FormControlContext.Provider value={null}`. But then we'd have to expose `FormControlContext` and I don't think we want to do that.

src/FormControl/_FormControlContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Make any component compatible with `FormControl`'s automatic wiring up of accessibility attributes & validation by
+ * reading the props from this hook and handling them / assigning them to the underlying form control. If used outside
+ * of `FormControl`, this hook has no effect.
+ *
+ * @param externalProps The external props passed to this component. If provided, these props will be merged with the
+ * `FormControl` props, with external props taking priority. This is also used for validating the external props,
+ * logging warnings to the console if there are conflicts.
+ */
+export function useFormControlForwardedProps<P extends FormControlForwardedProps>(externalProps: P): P {
```
iansan5653, line 26: Do we like this naming? "Forwarded props" vs "external props"? It reminds me of "forwarded refs" which also come from the parent.

.changeset/sweet-turtles-applaud.md
iansan5653, line 1: There are many whitespace-only changes in this PR -- I definitely recommend viewing with [whitespace suppressed](https://github.com/primer/react/pull/3632/files?diff=split&w=1).

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -79,7 +79,6 @@ const InlineAutocomplete = ({
   sx,
   children,
   tabInsertsSuggestions = false,
-  // Forward accessibility props so it works with FormControl
```
iansan5653, line 82: The context-based approach means `InlineAutocomplete` and `Autocomplete` no longer have to do anything special at all to work with `FormControl`, because they both wrap `TextInput`/`Textarea` which call this hook and get the context themselves. 

We still forward the rest props for `InlineAutocomplete` to avoid a breaking change (it could be useful in other situations).

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -79,7 +79,6 @@ const InlineAutocomplete = ({
...
   ...forwardProps
```
iansan5653, line 82: Forwarding isn't necessary to make it work with `FormControl` anymore since the underlying `TextInput` or `Textarea` will handle it

src/Autocomplete/AutocompleteInput.tsx
```diff
@@ -166,7 +166,6 @@ const AutocompleteInput = React.forwardRef(
         aria-haspopup="listbox"
         aria-owns={`${id}-listbox`}
         autoComplete="off"
-        id={id}
```
iansan5653, line 169: No ID is needed here as it will come from the `FormControl`

src/FormControl/FormControl.tsx
iansan5653: I feel more strongly about exposing this after looking into the ongoing work around `Filter` and `QueryBuilder` - more complex inputs that definitely could benefit from being able to directly access this info.

src/FormControl/_FormControlCaption.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react'
 import InputCaption from '../internal/components/InputCaption'
 import {SxProp} from '../sx'
-import {FormControlContext} from './FormControl'
+import {FormControlContext} from './_FormControlContext'
 
 const FormControlCaption: React.FC<React.PropsWithChildren<{id?: string} & SxProp>> = ({children, sx, id}) => {
-  const {captionId, disabled} = React.useContext(FormControlContext)
+  const {captionId, disabled} = React.useContext(FormControlContext) ?? {}
```
iansan5653, line 7: The default value for this context is now `null` so that `useForwardedProps` can handle things differently depending on whether or not the context is available

src/FormControl/FormControl.tsx
joshblack: Is the slot info something that could be passed along in the context value, potentially? Or does that not really work?

joshblack (commented):
Thanks for taking the time to put this together @iansan5653! I definitely love this approach compared to the HOC, I also kicked off an [integration PR](https://github.com/github/github/pull/285016) over on dotcom to see what kind of impact this change might have.

I feel like the main feedback that I wanted to sync up on was around usage of the hook itself and the context value.

I'm not sure if others feel this way, but it'd be great if the only way components could interact with the context value was through a hook (instead of `useContext()`). This could include the optional fallback value, as well, so that calls didn't need to `useContext() ?? {}`.

I'm also curious if you feel like a hook like `useFormControl` could work as a more strictly defined function so that all the props didn't need to be forwarded along. In other words, something like:

```tsx
function ExampleComponent({ id, required, disabled, ...rest }) {
  const formControl = useFormControl({
    id,
    required,
    disabled,
  });

  return <input {...formControl} />;
}
```

Where the props that we're "auto wiring" are the inputs and the return result basically merges them with the context value and warns, where appropriate. Also, if we feel like we need flexibility in the return of the hook an alternative for the return value could be:

```tsx
const { formControlProps } = useFormControl({
  // ...
});
```

src/FormControl/_FormControlContext.tsx
joshblack: I love the example that you brought up, I added a comment on our v37 release issue to try and capture this: https://github.com/github/primer/issues/1715#issuecomment-1682938944 as this is likely something that will stick around for v35 and v36.

Feel free to edit/add anything to the comment if you feel like there are other examples or situations that really demonstrate why this change is important 👀

src/FormControl/FormControl.tsx
joshblack: I'm definitely a fan of making this public, depending on how stable this ends up being maybe it makes sense in the drafts/experimental entrypoint until it's further validated?

I think something that was surprising initially was seeing this on the component namespace. Could hooks be something that is just accessed from the entrypoint instead of tying it to a specific component? e.g. `import { useFormControl } from '@primer/react';` or is the namespace helpful here in the same way that it helps with sub-components?

iansan5653: We definitely could expose the hook directly - I just figured the namespace was an obvious place to look for it. I am good with either one

iansan5653:
>  it'd be great if the only way components could interact with the context value was through a hook (instead of useContext()). This could include the optional fallback value, as well, so that calls didn't need to useContext() ?? {}.

Currently it's only internal components that use `useContext` (the context is not exposed publicly) and I only left it that way to be avoid making too many changes to existing code. But I definitely can make that change :+1:

> I'm also curious if you feel like a hook like useFormControl could work as a more strictly defined function so that all the props didn't need to be forwarded along.

I went back and forth on this but eventually landed on this approach for two reasons:

1. To ensure consistent handling of external props. We want to avoid have a mix where some components override external props and others let external props take priority (over `FormControl` props). This approach with the hook doesn't _force_ that, but it does strongly encourage it.
2. To support the `console` warnings when users do override the props when wrapping in a `FormControl`. The only way to do that is to know what props are incoming.

I think (1) isn't critical. If we're willing to drop the warnings (which I think are dubious anyway for reasons noted in other comments) then I would definitely agree with changing the hook to accept no arguments and only return the forwarded props. IMO, overall that's more intuitive.

src/FormControl/FormControl.tsx
iansan5653: I just tried this and honestly I don't think it makes it any cleaner. We still have to look through the children to figure out if it's a choice input, so all that changes is that we pass the slots info down. I think since these checks don't reference `props`, it makes sense to leave them here for now.

src/FormControl/_FormControlContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * `FormControl` props, with external props taking priority.
+ */
+export function useFormControlForwardedProps<P extends FormControlForwardedProps>(externalProps: P): P
+/**
+ * Make any component compatible with `FormControl`'s automatic wiring up of accessibility attributes & validation by
+ * reading the props from this hook and handling them / assigning them to the underlying form control. If used outside
+ * of `FormControl`, this hook has no effect.
+ */
+export function useFormControlForwardedProps(): FormControlForwardedProps
+export function useFormControlForwardedProps(externalProps: FormControlForwardedProps = {}) {
```
iansan5653, line 39: This type of generic signature + optional param is impossible to express without overloads. If you just do `useFormControlForwardedProps<P extends ForwardedProps>(externalProps?: P): P`, TypeScript will error because the function could potentially get called like `useForwardedProps<VerySpecificInterface>()` which would incorrectly have a return type of `VerySpecificInterface`

src/FormControl/FormControl.tsx
mperrotti: This was not a bug, this was intentional. You can't require an individual `Radio` or `Checkbox`, and they can't have their own validation status.

Instead, you handle those on the `CheckboxGroup` or `RadioGroup` level.

See the following documentation for more info:
- https://primer.style/design/ui-patterns/forms/overview#validation-message
- https://primer.style/design/ui-patterns/forms/overview#validation-message:~:text=An%20individual%20checkbox%20or%20radio%20button%20cannot%20be%20marked%20as%20required.

src/FormControl/_FormControlContext.tsx
mperrotti: I like this naming. It's clear and feels familiar.

mperrotti:
It would be helpful if we could add Storybook stories showing some examples of how consumers might use `useFormControlForwardedProps`.

src/FormControl/FormControl.tsx
mperrotti: I slightly prefer exposing the hook directly. Usually the properties on a component "object" are other components.

mperrotti (commented):
This looks so cool! My only feedback so far is:

- Add `useFormControlForwardedProps` examples to Storybook
- Add `useFormControlForwardedProps` tests
- Expose `useFormControlForwardedProps` directly (non-blocking feedback)

iansan5653:
> * Add `useFormControlForwardedProps` examples to Storybook

I see a few hooks in the Storybook, but they are all visual. I have a hard time seeing how I might make a story for this hook since it does not have a visual effect. Do you have anything in mind?

mperrotti:
@iansan5653 - it's ok if there's no visual effect. I just think it would help Primer consumers discover that hook and see an example of how it could be used. I'd recommend adding something like this in the "Hooks" section of Storybook:

```jsx
export default {
  title: 'Hooks/useFormControlForwardedProps',
  decorators: [
    Story => {
      return (
        <ThemeProvider>
          <BaseStyles>{Story()}</BaseStyles>
        </ThemeProvider>
      )
    },
  ],
  argTypes: {
    disabled: {
      type: 'boolean',
      defaultValue: false,
    },
    id: {
      type: 'string',
      defaultValue: 'some-id',
    },
    placeholder: {
      type: 'string',
      defaultValue: 'Placeholder',
    },
    required: {
      type: 'boolean',
      defaultValue: false,
    },
    validationStatus: {
      defaultValue: undefined,
      options: ['error', 'success', 'warning', undefined],
      control: {type: 'radio'},
    },
    // whatever else you might want to pass to useFormControlForwardedProps
  },
} as Meta

export const UseFormControlForwardedProps = args => {
  const {
    disabled,
    id,
    required,
    validationStatus,
    ['aria-describedby']: ariaDescribedBy,
    ...externalProps
  } = useFormControlForwardedProps(args)

  return (
    <>
      <label htmlFor={id}>Label {required ? '*' : undefined}</label>
      <input
        id={id}
        disabled={disabled}
        required={required}
        aria-invalid={validationStatus === 'invalid'}
        aria-describedby={ariaDescribedBy}
        {...externalProps}
      />
      <div id={ariaDescribedBy}>Some caption</div>
    </>
  )
}
```

or if you _just_ want to demonstrate what gets returned by the forwarded props, you could just render the result into a table or something:
```jsx
export const UseFormControlForwardedProps = args => {
  const forwardedProps = useFormControlForwardedProps(args)

  return (
    <table>
      <thead>
        <tr>
          <th>Forwarded prop</th>
          <th>Value</th>
        </tr>
      </thead>
      <tbody>
        {Object.entries(forwardedProps).map(([key, value]) => (
          <tr key={key}>
            <td>{key}</td>
            <td>{value}</td>
          </tr>
        ))}
      </tbody>
    </table>
  )
}
```

iansan5653:
@mperrotti Done -- added a story, tests, and directly exposed the hook

mperrotti (approved):

PR #3636, opened 2023-08-15
SelectPanel anchors should have aria-haspopup="dialog"

### LINKED ISSUE #3629: `SelectPanel` anchors should have `aria-haspopup="listbox"`
### Description

The anchor for [`SelectPanel`](https://primer.style/react/SelectPanel) is rendered with `aria-haspopup="true"` by default. The `true` value for this HTML attribute is equivalent to `menu`, implying that the related popup will have the `menu` role.

`SelectPanel`s have the `listbox` role, not `menu`, so the value of `aria-haspopup` here should be `listbox`.

### Steps to reproduce

Render a `SelectPanel` without a `renderAnchor` prop defined.

### Version

v35.27.1

### Browser

Chrome

electron97:
### Summary

- Currently the anchors of SelectPanel have `aria-haspopup="true"` which is incorrect.
- Since the value was set to `aria-haspopup="true"` by default in [AnchoredOverlay.tsx](https://github.com/primer/react/blob/main/src/AnchoredOverlay/AnchoredOverlay.tsx#L171).
- To overwrite that we need to add `aria-haspopup="listbox"` in `SelectPanel.stories.tsx` renderAnchor to fix the issue.

Closes #3629

### Screenshots

| Before Fix | After Fix|
|---|---|
|Screenshot 2023-08-25 at 10 35 36 PM|Screenshot 2023-08-25 at 9 52 19 PM|

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


electron97:
> Thanks so much for taking the time to open this PR! Just wanted to share this as a heads up, we're sync'ing up with the accessibility team here at GitHub to double-check the intended behavior for the upstream issue. We should be meeting on Friday this week and we'll figure out next steps from there. Hope this makes sense! Let me know if you have any questions.

Hey @joshblack ,

Thank you so much for taking your time and taking a look at the PR. This gives me motivation to contribute more to `primer/react`. Also I have seen your comment yesterday on the issue where you mentioned there will be discussion first regarding this which also made me think Did I understand completely why `aria-haspopup` equal to `listbox` and not `true`? I have worked on few accessibility issues myself in the past and then I started reading about [it](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup) and I learned what `aria-haspopup` attribute is actually used for. Thanks again for taking your time and giving me the heads up will keep on contributing to `primer/react`.

Thanks!

src/SelectPanel/SelectPanel.stories.tsx
```diff
@@ -54,7 +54,12 @@ export const Default = () => {
         title="Select labels"
         subtitle="Use labels to organize issues and pull requests"
         renderAnchor={({children, 'aria-labelledby': ariaLabelledBy, ...anchorProps}) => (
-          <Button trailingAction={TriangleDownIcon} aria-labelledby={` ${ariaLabelledBy}`} {...anchorProps}>
+          <Button
+            trailingAction={TriangleDownIcon}
+            aria-labelledby={` ${ariaLabelledBy}`}
+            {...anchorProps}
+            aria-haspopup="listbox"
+          >
```
joshblack, line 62:
```suggestion
          <Button
            trailingAction={TriangleDownIcon}
            aria-labelledby={` ${ariaLabelledBy}`}
            aria-haspopup="dialog"
            {...anchorProps}
          >
```

joshblack (changes requested):
Thanks so much for your patience! Requesting a small change to `dialog` based on the feedback in the issue. Let me know if you have any questions!

src/SelectPanel/SelectPanel.stories.tsx
electron97: Hey @joshblack,

The order which you suggested will not work

```
aria-haspopup="dialog"
{...anchorProps}
```
because first it will take the value of aria-haspopup to `dialog` then it will overwrite the value with `true` from [anchorProps](https://github.com/primer/react/blob/main/src/AnchoredOverlay/AnchoredOverlay.tsx#L171). So to fix that we need first pass `anchorProps` and then `dialog`.

```
{...anchorProps}
aria-haspopup="dialog"
```

I have updated the screenshots in the PR description with the current changes feel free to take a look.

Also I found something while testing the changes locally that the same issue still exists in [SelectPanel.features.stories.tsx](https://github.com/primer/react/blob/main/src/SelectPanel/SelectPanel.features.stories.tsx#L55).  Let me know should we fix that in this PR or open a complete new issue for that.

Thanks!

electron97:
hey @joshblack ✋ ,
It's been two months since this issue was raised so wanted to check with you once with the latest update.
Is this component now a legacy component since there is `SelectPanel2` and that's why we are not working on fixing the a11y issue or do you think the solution that I came up with is not the best solution for this issue?
Thanks!

joshblack:
Hey @electron97! So sorry about the delay, re-checking this and I think the changes look great 👍 

For SelectPanel, there has been new work on the V2 version but I think this change totally makes sense as-is. Thanks so much for making this PR!

joshblack (approved):

PR #3646, opened 2023-08-18
refactor(eslint): add pattern for importing from utils/ssr and move to useId

joshblack:
Refactor usage of `useSSRSafeId` to `useId` and add an eslint rule to catch imports from `utils/ssr`

#### Changed

- Update eslint config to use a pattern for a restricted import rule for utils/ssr
- AutocompleteMenu now uses useId
- FormControl now uses useId
- CheckboxOrRadioGroup now uses useId

#### Removed


broccolinisoup (approved):
Thank you 🙏🏻

PR #3647, opened 2023-08-18
`<InlineAutocomplete>`: Add `onSelectSuggestion()` callback to `<InlineAutocomplete />` to allow execution of additional code after an autocomplete item was selected.

paxos:
Adds `onSelectSuggestion()` callback to `<InlineAutocomplete />` to allow execution of additional code after an autocomplete item was selected.

### Screenshots

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


iansan5653 (commented):
I think this is a good idea and will unlock some useful new design patterns. I just want to highlight a few things to make sure we consider them, as I'll be OOO after tomorrow and won't be able to participate in any further discussion around this PR.

---

First, the pattern here around autocompletion is explicitly to insert text. I think it's probably OK to perform some side effect when a suggestion is selected as long as it's made clear what will happen, but we still must insert the suggested text or we will break expectations.

I want to point this out because a consumer might, for example, be tempted to set the `value` of a suggestion to an empty string and _only_ perform a side effect on selection. This would be unexpected for all users but would especially break the experience for screen reader users as the plain-text `value` of the suggestion is required to generate hidden assistive text. For example, this hidden hint text is announced when the list is open:

accessibility tree for inlineautocomplete showing a hint text '3 autocomplete suggestions available; 'monalisa' is highlighted. Press Enter to insert.

If the `value` of that suggestion was an empty string or didn't correspond to the visual representation of the suggestion, the hint would become nonsense.

---

Second, the way text is inserted into the input is more complicated than just updating the state of the input. If you just update the state of the input (outside of an `onChange` handler, which is special-cased by React) you lose the current cursor position (shifting it to the end of the input) and the undo stack will be cleared. 

So to allow insertion of text in places other than the end of the input, and to allow users to undo that insertion via <kbd>Cmd+Z</kbd>, the component uses [this `useSyntheticChange` hook](https://github.com/primer/react/blob/a0e846aeda87ee24ff3f9242c70511225e0bfd6a/src/drafts/hooks/useSyntheticChange.ts#L91) to simulate the user actually typing in the input ([via `execCommand`](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)).

I think there may be a temptation to use this event handler to update the value of the input to replace / change the inserted text, which would likely update state directly and be problematic.  

---

Neither of these are direct issues with this feature; more with how it might be used. I wonder if we can come up with a way to prevent misuse via the API? 

One option to solve the second concern might be to allow returning a `string` (ie, `string | undefined` with `undefined` leaving default behavior) to support customizing what text is inserted. Then the insertion would happen under the hood so it would still use `useSyntheticChange`. However, this conflicts with the first concern because the `value` would differ from the actual inserted text. Maybe that's OK as long as the `value` is still set to something useful -- I don't think anything in the hint text asserts that the inserted text will _exactly_ match the `value`. I think such a case might be doable but would definitely need extensive accessibility testing.

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -26,6 +26,11 @@ export type InlineAutocompleteProps = {
    * `suggestions` prop accordingly.
    */
   onShowSuggestions: (event: ShowSuggestionsEvent) => void
+
+  /** Called when a suggestion is selected.
+   */
+  onSelectSuggestion?: (suggestion: string) => void
```
iansan5653, line 32: To keep consistent with the rest of the props, maybe we should define a `SelectSuggestionsEvent` type and pass that instead of just the suggestion string. It could also include the trigger & the query text. This could give more useful data to the handler. For example:

```tsx
type SelectSuggestionsEvent = ShowSuggestionsEvent & {
  /** The suggestion that was selected. */
  suggestion: Suggestion
}
```

paxos:
@iansan5653 Please note that I removed the code that allowed overwriting of the inserted text. This PR now only adds a callback to notify a consumer what autocomplete item was selected. This should resolve most of the accessibility concerns you've mentioned :)

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
paxos: Done. Please check commit 5d1b9fadbecb7c3f7deb64cb812cecfa0999886b

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -26,6 +27,11 @@ export type InlineAutocompleteProps = {
    * `suggestions` prop accordingly.
    */
   onShowSuggestions: (event: ShowSuggestionsEvent) => void
+
+  /** Called when a suggestion is selected.
+   */
```
iansan5653, line 32:
```suggestion
  /**
   * Called when a suggestion is selected.
   *
   * @note This should be used only for performing side effects, not for modifying
   * the inserted text. Do not call `setState` in this handler or the user's cursor
   * position / undo history could be lost.
   */
```

iansan5653 (approved):
🚀 Looks great! Just a minor documentation note to call out the concerns around setting state.

BTW I will be out of office for the next 8 days, so please don't wait for my reapproval after accepting this change.

paxos:
@iansan5653 accepted your suggestion, thank you :)

iansan5653 (approved):
fast 😄

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -26,6 +27,16 @@ export type InlineAutocompleteProps = {
...
   onShowSuggestions: (event: ShowSuggestionsEvent) => void
+
+  /**
+   * Called when a suggestion is selected.
+   *
+   * @note This should be used only for performing side effects, not for modifying
+   * the inserted text. Do not call `setState` in this handler or the user's cursor
+   * position / undo history could be lost.
+   */
+  onSelectSuggestion?: (event: SelectSuggestionsEvent) => void
```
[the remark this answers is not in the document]
paxos: Updated the file - was not able to confirm the change is correct.

paxos:
Docs updated!

src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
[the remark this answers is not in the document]
paxos: This is done and CI is green.

mperrotti (approved):

PR #3660, opened 2023-08-23
chore: update gitignore file groups and add pattern for Playwright

joshblack:
With the update to Playwright, we currently have a `blob-report` folder that is generated to rollup all the test results into one location. In workflows that commit the results of running playwright with the update flag, this folder is now being committed.

This PR updates the `.gitignore` file to include a pattern for `blob-report` so this result is no longer included. It also re-groups some of the sections with comments to indicate what the groupings correspond to with respect to tools, processes, etc.


.gitignore
joshblack, line 1: Test

joshblack, line 1: Test

broccolinisoup (approved):
Looks great! Thanks for categorising them as well ✨

PR #3661, opened 2023-08-23
fix(Button): make children prop optional

joshblack:
Closes https://github.com/github/primer/issues/2601

Update our base `ButtonProps` so that `children` is optional. This allows for the following migration pattern to be supported for v36:

```tsx
// Before
<Button leadingIcon={ChatIcon}>
  <Button.Counter>{5}</Button.Counter>
</Button>

// After
<Button leadingAction={ChatIcon} count={5} />
```


langermank (approved):

broccolinisoup (approved):
🚀

PR #3663, opened 2023-08-24
ci(vrt): update snapshots workflow to use create-github-app-token

joshblack:
Update our `update-snapshots` workflow to use the new `actions-create-github-app-token` action 🥳 Hopefully this is a drop-in replacement for the existing action 🤞

### DIFF
--- a/.github/workflows/update-snapshots.yml
+++ b/.github/workflows/update-snapshots.yml
@@ -62,7 +62,7 @@ jobs:
     steps:
       - name: Generate token
         id: generate_token
-        uses: tibdex/github-app-token@v1
+        uses: actions/create-github-app-token@v1
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.PRIVATE_KEY }}


gr2m (approved):

camertron (approved):

.github/workflows/update-snapshots.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     steps:
       - name: Generate token
         id: generate_token
-        uses: tibdex/github-app-token@v1
+        uses: actions/create-github-app-token@v1
```
broccolinisoup, line 65: Is this doing a similar job to @camertron's action? https://github.com/camertron/github-app-installation-auth-action

broccolinisoup (approved):

.github/workflows/update-snapshots.yml
camertron: Awesome, yeah we should def migrate to the official action 👍

PR #3623, opened 2023-08-13
Fix: Docs for Blankslate Component

electron97:
Closes #3620 

## Summary

- Currently when visiting the React docs for [Blankslate](https://primer.style/design/components/blankslate/react) the import is incorrect.
- Also the storybook stories do not link correctly.
- Added stories to `Blankslate.docs.json` so that it appears correctly on the docs site.
- Also added `Blankslate.mdx` file to fix incorrect import.

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3670, opened 2023-08-26
Default stories for Page header do not link correctly.

### LINKED ISSUE #3669: Default stories for Page header do not link correctly.
### Description

When visiting the React docs for Page header at: https://primer.style/design/components/page-header/react

- The storybook stories do not link correctly. (the default story is incorrect)
- We should make sure that the docs for this component include the correct storybook paths.
 - Add stories to `PageHeader.docs.json` so that they appear correctly on the docs site.

### Steps to reproduce

1. Go to https://primer.style/design/components/page-header/react.
2. Check the default storybook stories.

### Version

v35.28.0

### Browser

Chrome

electron97:
- Currently the default storybook stories for [pageheader](https://primer.style/design/components/page-header/react) do not link correctly.
- Added stories in `PageHeader.docs.json` so that it link the storybook stories correctly.
- Changes in docs so skipping changeset.

Closes #3669

### DIFF
--- a/src/PageHeader/PageHeader.docs.json
+++ b/src/PageHeader/PageHeader.docs.json
@@ -3,7 +3,11 @@
   "name": "PageHeader",
   "status": "draft",
   "a11yReviewed": false,
-  "stories": [],
+  "stories": [
+    {
+      "id": "drafts-components-pageheader--default"
+    }
+  ],
   "props": [
     {
       "name": "aria-label",


joshblack (approved):

joshblack:
Thanks for tackling this @electron97!

joshblack:
Hmm, based on the validation it seems like the `stories` field here needs to be an array or stories from the `PageHeader.features.stories.tsx` file 🤔 

cc @siddharthkp it seems like this file is marked `draft` as expected, do you happen to know why the default story is not working then?

broccolinisoup:
@joshblack 👋🏻 

I was reviewing [this PR](https://github.com/primer/react/pull/3623) and the tests failed and I think it is because of the same reason you mentioned above that it is not a feature story id. 

From what I understand, we generate [default story id](https://github.com/primer/react/blob/main/script/components-json/build.ts#L41-L42) for each component and it doesn't take component status into account if I am reading this right 

I wonder if [such fix](https://github.com/primer/react/pull/3682) would resolve this? 🤔

joshblack:
@broccolinisoup ooo I think it would! Let me know when that PR is good to go, it will help out a ton 🥳

broccolinisoup:
https://github.com/primer/react/pull/3670 fixes the default story id for components that are not in the main bundle (i.e. drafts, deprecated etc) With https://github.com/primer/react/pull/3670 is out, [PageHeader docs](https://primer.style/design/components/page-header/react) are now rendered correctly 🎉 I'm closing this PR but please let me know if there is any concern 🙌🏻

PR #3677, opened 2023-08-28
Use release_with_app shared workflow

camertron:
After a long and tedious battle with Actions, I finally got a shared version of the `release.yml` action to work with GitHub Apps.


joshblack (approved):
🤞

PR #3447, opened 2023-06-23
[Bug] Leading & Trailing Visual Colors for Button

jesskuo4:
Describe your changes here.
- Changed the Leading and Trailing Visual Icon Colors for the Button Component to `fg.muted`
- Change ButtonCounter font size to 0 (12px) 

Closes https://github.com/github/primer/issues/2373

### Screenshots
| Before | After | 
| -- | -- |
| Before image of trailing/leading icons | After image of trailing/leading icons
| Before image of counter with wrong size | After image of counter with corrected size |

Please provide before/after screenshots for any visual changes

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated documentation
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3683, opened 2023-08-29
Button: Make button visuals overridable by sx

broccolinisoup:
As described in the issue, a regression introduced in https://github.com/primer/react/pull/3447 where the icon colors became not overridable by the sx prop in Button.

This issue originally goes back to introducing data-attributes selectors to Button (https://github.com/primer/react/pull/2792) and it came with some challenges with overriding styles with sx. At the time, we wrote [a custom CSS selector](https://github.com/primer/react/blob/changeset-release/main/src/Button/Button.tsx#L65) to take data attributes into account while applying the sx styles. However, the new icon styles that are introduced in https://github.com/primer/react/pull/3447, are not included in the custom selectors.

Note: We won't have to use this css selectors when we move to css modules ✨ but for now, it seems to be doing what we want.

All that said, I am always open to any kinds of feedback 🙏🏻

Closes https://github.com/github/primer/issues/2576

#### Before

Close issue button with its icon and the icon color is muted grey

#### After

Close issue button with its icon and the icon color is purple

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Button/Button.tsx
```diff
@@ -66,21 +66,31 @@ export function generateCustomSxProp(
...
+  // We need to make sure we add the data attributes to the base css class (& -> &[data-attributename="value"]]) otherwise sx styles are applied due to data attribute has more specificity than class selector
   const cssSelector = `&${size}${block}${noVisuals}` // &[data-size="small"][data-block="block"][data-no-visuals="true"]
+  cssSelectorList.push(cssSelector)
+
+  // Possible data attributes for children of button component: data-component="leadingVisual", data-component="trailingVisual", data-component="trailingAction"
+  // When sx is used in the parent button component, these styles aren't applied to its children (leadingIcon, trailingIcon, trailingAction)
+  // Because sx only applies to the base selector which is  "&", but we want it to be applied to '& [data-component="leadingVisual"]'
+  if (props.leadingIcon) cssSelectorList.push(`& [data-component="leadingVisual"]`)
+  if (props.trailingIcon) cssSelectorList.push(`& [data-component="trailingVisual"]`)
+  if (props.trailingAction) cssSelectorList.push(`& [data-component="trailingAction"]`)
```
broccolinisoup, line 83: These lines are repetitive and I am happy to take into a loop but thought keeping this way makes them more readable. Let me know if you have any feedback!

joshblack (commented):
@broccolinisoup could we also add a test for this to make sure we catch any regressions in the future in this area? 👀

src/__tests__/Button.test.tsx
```diff
@@ -89,4 +89,51 @@ describe('Button', () => {
     const button = container.getByRole('button')
     expect(button).toMatchSnapshot()
   })
+  it('overrides the styles declared with data attributes with sx prop', () => {
```
broccolinisoup, line 92: @joshblack I added tests to target specifically the css attributes that are overridden with the data attributes. Let me know what you think about the tests and if you have any recommendation 🙌🏻

src/__tests__/Button.test.tsx
```diff
@@ -89,4 +89,51 @@ describe('Button', () => {
...
+    expect(smallBtn).toHaveStyle('font-size: 32px')
+    const largeBtn = container.getByRole('button', {name: 'Large button'})
+    // [data-size="large"] sets height to 40px in the styles, it should be overrideable with sx prop
+    expect(largeBtn).toHaveStyle('height: 20px')
+    const blockBtn = container.getByRole('button', {name: 'Block button'})
+    // [data-block="block"] sets width to 100% in the styles, it should be overrideable with sx prop
+    expect(blockBtn).toHaveStyle('width: 80%')
+    const noVisualsBtn = container.getByRole('button', {name: 'Invisible button'})
+    // [data-no-visuals] sets color to accent.fg for invisible , it should be overrideable with sx prop
+    expect(noVisualsBtn).toHaveStyle('color: firebrick')
```
broccolinisoup, line 126: This test passes even when I comment out [custom sx generater](https://github.com/primer/react/blob/main/src/Button/Button.tsx#L15) which is very odd. Because the same button

```jsx
<Button variant="invisible" sx={{color: 'firebrick'}}>
    Invisible button
</Button>
```

fails to update the color to firebrick when  [custom sx generater](https://github.com/primer/react/blob/main/src/Button/Button.tsx#L15) is commented out. I don't know what is really different in test env than storybook but I added the storybook to compansate this awkward test. Let me know if you have any concerns!

broccolinisoup:
Hey @joshblack that totally works for the `color`  🎉  I wasn't very happy how the `generateCustomSxProp` function got more complex and unreadable but couldn't think of a different way. I appreciate this so much 🙏🏻  I'll definetely update the implementation just a few things clarify around this:

- I was thinking if we can apply the custom css prop solution you suggested to the rest of the styles that uses data attributes to simplify the custom sx prop function. However, as far as I understand, this might be a bit challenging for the rest of the styles that uses data attribute. 

For example to make sure sx styles are applied properly to the `size="small"` button that uses [data-attribute for setting styles](https://github.com/primer/react/blob/main/src/Button/styles.ts#L249-L253), seems like we need to create css var for each attribute that we are overriding i.e. `font-size`, `padding` etc. Is this what you are reading as well or do I miss something here?

- In terms of the jest tests, I was using `toHaveStyle` for expect but seems like that function has some issue with the css variables (some related open issues I found: [issue1](https://github.com/testing-library/jest-dom/issues/280), [issue2](https://github.com/testing-library/jest-dom/issues/322)) - Do you have any recommendation to test these styles in jest? I am also happy to add test stories for each case for VRT. Let me know what you think!

Thanks again, I appreciate your input 🙏🏻

joshblack (approved):
I'd totally be down to move forward with this if you are 👀

broccolinisoup:
@joshblack All looking good! Checks are green and tested in review lab to see the button that is originally reported as an issue

https://github.com/github/github/pull/291667
Screenshot 2023-10-11 at 12 42 45 pm

PR #3691, opened 2023-08-30
fix(ActionList): place id on item with role

joshblack:
Context: https://github.slack.com/archives/C01L618AEP9/p1693422238328979

The `id` on `ActionList.Item` should be placed on the item with the corresponding ARIA role so that a consumer can use that value to manage attributes like `aria-activedescendant` appropriately.

> **Warning**
> This change could be considered a breaking change if we consider the `id` placement as something that should be stable across minor versions.


dusave (approved):

mperrotti (approved):
CI should pass after updating snapshots.

joshblack:
@dusave checking upstream, it doesn't seem like this change introduces a regression so it could be safe to land. Wanted to double-check with you to see how you felt about this change.

I'll also ask the team to see if we're okay with this kind of change in a current minor release 👀

dusave (approved):
🚀 Awesome!

PR #3698, opened 2023-09-05
NavList add defaultOpen to NavList.Item 

gracepark:
Closes https://github.com/github/primer/issues/2512

This adds the `defaultOpen` prop to NavList.Item. We needed this because in the Docs sidebar needs the ability to have a NavList.SubNav default to open even without an `aria-current` in its nested levels like so:

Item 1 is default to open because it has the defaultOpen prop so that sub item 1 is viewable

Another pro is that this essentially resolves depending on client-side rendering of our sidebar.

A few notes about my changes:
- I named the prop `defaultOpen` to follow a similar pattern to TreeView's `defaultExpanded`. We're wondering if it'd be a little more accurate to name it something like `initiallyOpen`, `defaultToOpen` or `openOnStart`. 
- I wanted to catch the possiblity of setting a `defaultOpen` on a NavList.Item without a NavList.SubNav. I was thinking it could throw an error instead, but I noticed below other checks like depth were using `console.error`, so I did the same.

Thank you for the review, and please let me know if there are any questions or concerns!

### Screenshots

No real visual difference unless a `defaultOpen` is added to a NavList.Item with a NavList.SubNav. That is shown above in the changes I described. 

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/NavList/NavList.stories.tsx
+++ b/src/NavList/NavList.stories.tsx
@@ -51,11 +51,10 @@ export const WithNestedSubItems: Story = () => (
   <PageLayout>
     <PageLayout.Pane position="start">
       <NavList>
-        <NavList.Item defaultOpen={true} href="#">Item 1
+        <NavList.Item defaultOpen={true} href="#">
+          Item 1
           <NavList.SubNav>
-            <NavList.Item href="#">
-              Sub item 1
-            </NavList.Item>
+            <NavList.Item href="#">Sub item 1</NavList.Item>
           </NavList.SubNav>
         </NavList.Item>
         <NavList.Item href="#">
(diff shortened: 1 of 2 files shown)


siddharthkp:
> I named the prop defaultOpen to follow a similar pattern to TreeView's defaultExpanded. We're wondering if it'd be a little more accurate to name it something like initiallyOpen, defaultToOpen or openOnStart

You're completely right, but because the [HTML API has keys like `defaultValue` and `defaultChecked`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#defaultvalue), it's pretty common to see the same in the React ecosystem ([For example in these docs](https://react.dev/reference/react-dom/components/input#:~:text=type%3D%22radio%22%20inputs.-,defaultValue,-%3A%20A%20string.%20Specifies))

To keep the API familiar, we have kept the same convention for props that set the initial value in primer/react as well

siddharthkp:
- Merged for v35.29.0 https://github.com/primer/react/pull/3679

PR #3708, opened 2023-09-05
Revert "chore(deps): bump @lit-labs/react from 1.1.1 to 2.0.1 (#3652)"

joshblack:
This reverts commit 440829c0a8d450c20075b5e8da32b3ed1e359271.

This change unfortunately impacts anyone who uses the `lib` entrypoint as we compile the `@lit-labs/react` package into a CommonJS format which collapses the `exports` field in that package's `package.json` file.

This file is important because it specifies which entrypoints to use for the browser versus node. When that information is lost, it will attempt to run the code intended for the browser in a Node.js environment which causes errors like `useLayoutEffect`.

I'm unsure how to best address this so wanted to roll back and then come back with a plan for this work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@github/markdown-toolbar-element": "^2.1.0",
         "@github/paste-markdown": "^1.4.0",
         "@github/relative-time-element": "^4.1.2",
-        "@lit-labs/react": "^2.0.1",
+        "@lit-labs/react": "^1.1.1",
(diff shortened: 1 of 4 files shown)


mperrotti:
> I'm unsure how to best address this so wanted to roll back and then come back with a plan for this work.

Do we have an issue for this work yet?

mperrotti (approved):

joshblack:
@mperrotti opened one over here: https://github.com/primer/react/issues/3710

Issue #3712, opened 2023-09-06
Support controlling `ActionMenu` position (anchor side)


iansan5653:
`ActionMenu` always renders below its anchor element, and this cannot be overridden. This would be very useful to control, particularly to be able to position submenus to the right of the anchor:

![Screenshot of ActionMenu with a nested ActionMenu inside. The submenu appears below the menu item that opened it.](https://github.com/primer/react/assets/2294248/7cb9b86f-b7d3-4765-9b81-a8ecfbdf3115)

In the above screenshot, we'd like to render the menu to the right by default, as recommended by the [Primer design guidelines](https://primer.style/design/components/action-menu#multi-level).

`ActionMenu.Overlay` does forward unused props down to the underlying `Overlay` component, and `Overlay` does accept an `anchorSide` prop. This would appear to allow custom positioning, but this prop only controls the transition direction on open -- not the position. 

A simple solution would be to add a `side` prop and forward it to the `AnchoredOverlay` that powers the menu.

PR #3713, opened 2023-09-06
Deduplicate `ActionMenu.tsx`

iansan5653:
While looking into https://github.com/primer/react/issues/3712, I noticed that `ActionMenu.tsx` is duplicated. It exists once in `src/` and once in `src/ActionMenu`. The `src/ActionMenu.tsx` file is the one being exported because it takes priority over `src/ActionMenu/index.ts`, but the `src/ActionMenu/ActionMenu.tsx` file is the one being tested and rendered in Storybook. This strikes me as particularly risky as the two could diverge over time (they already have in that the `aria-labelledby` prop was only added to the file in `src/`). 

This appears to have been a leftover from some rearranging of files.

This PR cleans up the duplication by merging the two files and deleting the one in `src/`.

- Closes https://github.com/primer/react/issues/2992

PR #3714, opened 2023-09-06
Add support for controlling which side `ActionMenu` renders on (via a `side` prop on `ActionMenu.Overlay`)

iansan5653:
Adds a `side` prop to `ActionMenu.Overlay` so the consumer can control which side of the anchor the `ActionMenu` renders on. This prop is forwarded verbatim to ` Also adds a story and updates docs.

Closes https://github.com/primer/react/issues/3712

### Screenshots
screenshot of action menu appearing on right-hand side of anchor button

### Merge checklist

- ~~[ ] Added/updated tests~~
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- Stacks on:
    - [ ] https://github.com/primer/react/pull/3713


siddharthkp (approved):
Looks good! Perfect use case for the prop.

Not blocking: Can we also add a story with nested menus to keep in examples of the ideal use case?

iansan5653:
> Can we also add a story with nested menus to keep in examples of the ideal use case?

Unfortunately the nested menu case requires a bit more wiring up than that because of keyboard navigation & roles. We are internally working on a `NestedActionMenu` component which I think we should probably eventually upstream to Primer.

PR #3718, opened 2023-09-07
chore(examples): update Next.js to use same react, react-dom as @primer/react

joshblack:
Unfortunately, our Next.js example is failing due to multiple versions of React being used. This leads to a rules of hook violation and means that the example no longer runs.

This PR introduces a temporary solution to reference `react` and `react-dom` from the top-level `node_modules` folder which is shared with `@primer/react`. In the future, this should be supported through NPM workspaces instead of linking to `node_modules` directly.


broccolinisoup (approved):
Love how you solved this 🔥 This is so helpful! Thank you so much 🙏🏻

PR #3719, opened 2023-09-07
feat(npm): move deprecated components in v36 to deprecated

joshblack:
Part of: https://github.com/github/primer/issues/2628

Add components deprecated for v35 to deprecated entrypoint in v35 so that we can migrate teams before the v36 release.


siddharthkp (approved):
Looks good!

PR #3720, opened 2023-09-07
Support link underline preference

mperrotti:
This was initiated by our need for underlined links in blocks of text, and supports dotcom's new user setting to render links with underlines. The link underline setting is currently being tested on dotcom behind a feature flag.

A new setting has been added to the Storybook toolbar that allows us to preview what a component would look like when the user has specified that they prefer link underlines.

Closes https://github.com/github/primer/issues/2536

#### User who turned link underlines on in their GitHub settings
Screenshot 2023-09-14 at 5 45 07 PM

### Additional context about link underline settings

Issue: https://github.com/github/releases/issues/3340
Dotcom PR: https://github.com/github/github/pull/278701

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
There are cases where we want links in a single like of text to be underlined. Converting this to a draft until I update this PR to address those cases.

Examples:
CleanShot 2023-09-08 at 16 00 14@2x
CleanShot 2023-09-08 at 16 01 01@2x
Screenshot 2023-09-08 at 10 05 19 AM

mperrotti:
Just ignore this PR until I mark it as "Ready for review".

We have more changes to make after talking with @ericwbailey about the feature-flagged user preferences for link underlines.

src/drafts/Prose/Prose.tsx
```diff
@@ -0,0 +1,159 @@
+import React from 'react'
```
mperrotti, line 1: I'm questioning whether PRC even needs this component. It's very rare that we need to render a large block of prose in the apps that use Primer.

Reviewers - do you think we should add a Prose component?

joshblack: @mperrotti I'd be down to ship with the underline changes that you have with the custom property stuff and then see if we still need something like this in the future 👀 Like you said, it seems like underline behavior is more of a one-off thing.

joshblack:
Just wanted to check-in on this @mperrotti to see what you thought of: https://github.com/primer/react/pull/3720#discussion_r1330528288 and how you wanted me to move forward with reviewing this PR 👀

src/drafts/Prose/Prose.tsx
mperrotti: Thanks @joshblack - I totally missed your comment last week. I'll take out the Prose stuff.

.playwright/snapshots/components/Breadcrumbs.test.ts-snapshots/Breadcrumbs-Default-light-forcedUnderlines-linux.png
joshblack, line 1: @mperrotti just wanted to double-check the snapshot here, it seems like we should be seeing underlines in this snapshot, right?

mperrotti: Yes, we should. I'll investigate...

mperrotti: It looks correct in the deployment preview. Any idea what might be causing the snapshots to be wrong?

joshblack: @mperrotti let me take a look and see 🤔 something might be off on the VRT side.

joshblack (approved):
I _think_ that last commit should do the trick once the screenshots update 🤞 Looks great!

PR #3723, opened 2023-09-08
Ssr compat portal

### LINKED ISSUE #3721: Portal is not ssr compat - accesses `document` during render
### Description

When using `InlineAutocomplete` we inherit usage of `<Portal />` rendering on page load. 

I think the issue occurs because `<Portal>` calls `document.createElement` and `createPortal` with a ref to an element that can't be created. 

https://github.com/primer/react/blob/500e529df47ae5def4bf5cedc55fc099251c25a1/src/Portal/Portal.tsx#L68

we should avoid these apis in render in portal (which creates a div on every render) and rely on other mechanisms to sync it

In addition to this, it renders an _Autocomplette which uses `useCombobox` which calls `useLayoutEffect` resulting in a call to `console.error` on the server - https://github.com/primer/react/blob/500e529df47ae5def4bf5cedc55fc099251c25a1/src/drafts/hooks/useCombobox.ts#L141-L157

### Steps to reproduce

Atempt to SSR an app that uses `InlineAutocomplete`

### Version

latest

### Browser

_No response_

mattcosta7:
Describe your changes here.

Closes #3721 

Makes Portal SSR safe and avoids creating throwaway elements on every render

### Screenshots

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/rich-jars-run.md
+++ b/.changeset/rich-jars-run.md
@@ -1,0 +1,7 @@
+---
+'@primer/react': patch
+---
+
+make portal compatible with ssr
+
+<!-- Changed components: Portal -->
--- a/src/Portal/Portal.tsx
+++ b/src/Portal/Portal.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useRef, useState} from 'react'
 import {createPortal} from 'react-dom'
 import useLayoutEffect from '../utils/useIsomorphicLayoutEffect'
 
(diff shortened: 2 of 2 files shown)


.changeset/rich-jars-run.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/react': patch
```
mattcosta7, line 2: this could be seen as more than a patch, but I don't believe any code relying on portal should break, as it should only _start_ not throwing in ssr, and otherwise be consistent with the current impl

src/Portal/Portal.tsx
```diff
@@ -65,16 +65,33 @@ export const Portal: React.FC<React.PropsWithChildren<PortalProps>> = ({
   onMount,
   containerName: _containerName,
 }) => {
-  const hostElement = document.createElement('div')
+  const [element, setElement] = useState<HTMLDivElement | null>(() => {
+    // If we're in SSR, return null so that nothing is rendered
+    if (typeof document === 'undefined') return null
+    return document.createElement('div')
```
mattcosta7, line 71: when we aren't in SSR, we'll create this element immediately, and use it - avoiding the need to render null at all, but when we do SSR, we'll render null.

this could create a server mismatch, in which case we might prefer always defaulting to null - not quite sure how we'd validate that currently

examples/nextjs/src/components/Layout.js
```diff
@@ -10,6 +10,8 @@ export default function Layout({children}) {
     <>
       <Navigation />
       <main>{children}</main>
+      <div id="__primerPortalRoot__" />
+      <Portal>Portal content</Portal>
```
mattcosta7, line 14: validated this works

src/Portal/Portal.tsx
mattcosta7: Was hoping this worked, but in tsting it's not quite there.

PR #3733, opened 2023-09-11
test: add experiment for getting all public exports

joshblack:
This is an experiment that shows how we would expand our "public API" tests to include _all_ public exports from our package. Currently, we only support that which is discoverable through a native Node.js import. With this new approach, we're able to get all exports including TypeScript types.

The heuristics used could be updated a bit to be a little smarter (especially when deduping `default` exports or inferring value types through re-exported module) but I thought this was a good demonstration of how much info we could get through building up an internal graph of modules.

## Feedback

Would love to know what you all think! Let me know what you think of this approach for getting all exported members and if there are any other alternatives that you'd recommend investigating.

Also would love to know how you see this in your workflow. If you see a change to this file, in particular a removal, what would you expect to happen? Would it help to have a CI workflow that warns if the public API is changed in a way that is incompatible with the corresponding changeset (if one exists)?


script/get-exports.mjs
```diff
@@ -0,0 +1,453 @@
+import {existsSync} from 'node:fs'
```
siddharthkp, line 1: Sorry, I'm a bit confused, where would we like to call this script from? after `./script/build` (npm run build)?

joshblack: @siddharthkp whoops! This was my test script, removed in the latest commit since it's in the test file itself 👍

joshblack:
bump @siddharthkp @broccolinisoup when you get a sec 👀

siddharthkp (approved):
LGTM!

Took me a while to understand, but now I get it 😇

PR #3675, opened 2023-08-28
Remove `main` landmark from PageLayout.Content.

radglob:
Reintroducing these changes which were introduced in https://github.com/primer/react/pull/3154, but were reverted.

Closes https://github.com/github/primer/issues/1586

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated documentation
- [X] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3737, opened 2023-09-12
fix(PageLayout): restore main landmark

joshblack:
Follow-up to: https://github.com/primer/react/pull/3675

Currently, we have teams in a couple of situations where either:

- A team makes use of a `main` landmark and therefore runs into the duplicate landmark violation
- A team uses the default `main` landmark from `PageLayout.Content`

When removing the `main` landmark, teams in the latter are impacted. In this PR, we introduce an `as` prop to allow for teams in the first example to change the semantics of this component to `as="div"` as they already have a `main` defined.

## Rollout

This is an incremental rollout strategy if our ideal here is to remove `main` as a landmark from `PageLayout.Content`. With this path, we could swap teams to use `as="div"` before removing the `main` support and also dropping `as="div"`

### DIFF
--- a/.changeset/friendly-shirts-protect.md
+++ b/.changeset/friendly-shirts-protect.md
@@ -1,0 +1,7 @@
+---
+'@primer/react': patch
+---
+
+Restore default main landmark in PageLayout.Content
+
+<!-- Changed components: PageLayout, SplitPageLayout -->
--- a/src/PageLayout/PageLayout.test.tsx
+++ b/src/PageLayout/PageLayout.test.tsx
@@ -23,9 +23,7 @@ describe('PageLayout', () => {
       <ThemeProvider>
         <PageLayout>
           <PageLayout.Header>Header</PageLayout.Header>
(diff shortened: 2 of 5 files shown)


siddharthkp (approved):
Oof tough spot! But really good rollout plan!

siddharthkp:
> With this path, we could swap teams to use as="div" before removing the main support and also dropping as="div"

(not for this PR, but for after) How do you feel about this slight change:

Right now, the default element is `main`. If we want to change the default to `div`, folks that do not have duplicate `main`, would then need to wrap `PageLayout.Content` in a `<main>`

How do you feel about supporting the `as` prop on `PageLayout.Content` with just 2 options: `main | div`. 

So, changing the default from `main` to `div` would still be a breaking change, but it can be accompanied by a codemod that adds `as="main"` for instances that do not have `as="div"` already.

joshblack:
@siddharthkp I like it, works for me!

PR #3743, opened 2023-09-13
docs(versioning): update docs to include landmark roles

joshblack:
Add a section to our versioning docs to include landmark roles as they relate to semver.

## Feedback

Would love to hear what you all think about this section and if it accurately captures our semver policy 👀 Let me know if there is anything missing or anything that needs to be changed!

### DIFF
--- a/contributor-docs/versioning.md
+++ b/contributor-docs/versioning.md
@@ -11,6 +11,8 @@
   - [The type of a prop is broadened](#the-type-of-a-prop-is-broadened)
   - [The type of a prop is narrowed](#the-type-of-a-prop-is-narrowed)
   - [The `display` property used for the container of `children` is changed](#the-display-property-used-for-the-container-of-children-is-changed)
+  - [A component includes a landmark role](#a-component-includes-a-landmark-role)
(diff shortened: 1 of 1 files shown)


broccolinisoup (approved):
I love this 🔥 My only feedback/question is that should we elaborate what we mean by "potentially major" under the semver section? 👀

TylerJDev (approved):
Looks great! ✨

Off-topic, but it might be worth a discussion in the future to see if we should steer clear from having default landmarks all together in components and always give that choice to the consumer? This doesn't seem to be the norm in newer components, but I do see the friction that could be caused through updates such as with `PageLayout` and others. I really like the approach done in the fix https://github.com/primer/react/pull/3737 and I wonder if it could be applied elsewhere 🤔

iansan5653 (approved):
Awesome, thank you!

joshblack:
@TylerJDev totally agreed, I think you're dead-on that we probably shouldn't be baking in default landmark roles. Let me see where would a good place be to capture that so that we consider it in future component work 👀 (or let me know if you have any places in mind where it may be good to capture this!)

PR #3744, opened 2023-09-14
fix(ButtonBase): remove leading, trailing props from being spread onto base butotn

joshblack:
Addresses a current release blocker where leading, trailing props were not being spread onto the base button causing an "unknown DOM attribute" warning from React.

#### Changed

- Add `leadingIcon`, `leadingVisual`, `trailingIcon`, and `trailingVisual` to destructure list for ButtonBase props

#### Removed


siddharthkp (approved):
perfect!

PR #3763, opened 2023-09-25
SelectPanel2: Add option to use external anchor

siddharthkp:
There are a bunch of use cases where you'd need to open a SelectPanel with an external anchor

For example, here's an example from memex where the anchor is a row cell:

clicking a row cell opens a SelectPanel to switch milestones

&nbsp;

Another example, here's a SelectPanel that opens from an action in ActionMenu:

https://github.com/primer/react/assets/1863771/78daf150-52f1-44c8-9853-116e08a8da19

### DIFF
--- a/src/drafts/SelectPanel2/SelectPanel.stories.tsx
+++ b/src/drafts/SelectPanel2/SelectPanel.stories.tsx
@@ -660,31 +660,23 @@ export const FExternalAnchor = () => {
 }
 
 export const GOpenFromMenu = () => {
-  const initialSelectedLabels = data.issue.labelIds // mock initial state: has selected labels
-  const [selectedLabelIds, setSelectedLabelIds] = React.useState<string[]>(initialSelectedLabels)
-
   /* Selection */
-  const onLabelSelect = (labelId: string) => {
-    if (!selectedLabelIds.includes(labelId)) setSelectedLabelIds([...selectedLabelIds, labelId])
-    else setSelectedLabelIds(selectedLabelIds.filter(id => id !== labelId))
+
+  const [selectedEvents, setSelectedEvents] = React.useState<string[]>([])
+
+  const onEventSelect = (event: string) => {
+    if (!selectedEvents.includes(event)) setSelectedEvents([...selectedEvents, event])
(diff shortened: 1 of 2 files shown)


src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -77,13 +77,14 @@ const SelectPanel = props => {
   return (
     <>
       <AnchoredOverlay
+        // @ts-ignore todo
```
siddharthkp, line 80: adding a ts-ignore temporarily until I declare types for props properly in a follow up PR :)

src/drafts/SelectPanel2/SelectPanel.stories.tsx
```diff
@@ -587,6 +587,168 @@ export const EMinimal = () => {
...
+  const itemsToShow = data.labels.sort(sortingFn)
+
+  const anchorRef = React.useRef<HTMLButtonElement>(null)
+  const [open, setOpen] = React.useState(false)
+
+  return (
+    <>
+      <h1>With External Anchor</h1>
+      <p>
+        To use an external anchor, pass an `anchorRef` to `SelectPanel`. You would also need to control the `open` state
```
broccolinisoup, line 624: >You would also need to control the `open` state

Curious if we can handle the open state out of the box? Like we do in the ActionMenu.Button or AnchionMenu.Anchor. 

I am sure you have thought of this but curious to understand the reasoning 😊

src/drafts/SelectPanel2/SelectPanel.stories.tsx
```diff
@@ -587,6 +587,168 @@ export const EMinimal = () => {
...
+              <ActionList.Description variant="block">{label.description}</ActionList.Description>
+            </ActionList.Item>
+          ))}
+        </SelectPanel.ActionList>
+      </SelectPanel>
+    </>
+  )
+}
+
+export const GOpenFromMenu = () => {
```
broccolinisoup, line 660: Such a cool use case and story 🔥

broccolinisoup (approved):

PR #3769, opened 2023-09-26
ci: update release-schedule workflow with dynamic schedule

### LINKED ISSUE #3711: Update Release Schedule workflow to account for new release cadence
Currently, our [release schedule](https://github.com/primer/react/blob/main/.github/workflows/release-schedule.yml) workflow runs every week and assumes a weekly cadence. Our new release cadence is every two weeks and the workflow should be updated to reflect that.

This may require some investigation into the PagerDuty in order to get the exact schedule as it may be difficult to infer using JS directly. There is also a challenge in that it is difficult to write a CRON schedule that alternates weeks (as far as I know) so this workflow would need to be resilient to running potentially more often than necessary (every two weeks)

joshblack:
Closes #3711

Our release schedule workflow currently runs every week and will open and close issues regardless of the settings of the current schedule that we have in PagerDuty. This PR adds in a composite action, pagerduty, that will allow us to get the current and previous schedule from the PagerDuty API and uses it in the release-schedule workflow to accurately open and close issues related to this schedule

#### New

- Add `pagerduty` action under `.github/actions/pagerduty` which outputs schedule information

#### Changed

- Update `release-schedule` workflow to use new `pagerduty` action and correctly open/close release tracking issues for a given schedule

### Rollout strategy

None

### Testing & Reviewing

This is something that likely needs to be tested with `workflow_dispatch` to see if it's working as-intended. Ran it over in: https://github.com/primer/react/actions/runs/6316114089 with the new "dry" flag to see how it would run. Thankfully it picked up on the fact that we have an existing release issue but it was one that it couldn't identify because of how the ids are changing (now they just use the schedule id)


siddharthkp (approved):
Good stuff! Let's try it out!

PR #3758, opened 2023-09-22
SelectPanel2: Add default Header and Footer, Replace SelectPanel.Heading

siddharthkp:
Related context:
- https://github.com/github/primer/pull/2490
- https://github.com/github/primer/issues/2396

&nbsp;

#### Part 1/2: Add default Header and Footer when not provided

This minimal example would render the fallback header and footer:

```jsx
<SelectPanel title="Select labels" onSubmit={onSubmit}>
  <SelectPanel.Button>Assign label</SelectPanel.Button>
  
  <SelectPanel.ActionList>
    {labels.map(label => (
      <ActionList.Item
        onSelect={() => onSelect(label.id)}
        selected={selectedLabelIds.includes(label.id)}
      >
        {label.name}
      </ActionList.Item>
    ))}
  </SelectPanel.ActionList>
</SelectPanel>
```

Minimal SelectPanel still renders a Header and Footer

&nbsp;

#### Part 2/2: Heading element is mandatory and not customisable

Related context:
- https://github.com/github/primer/issues/2578 (It's not, it should always render h1)

```jsx
// before:
<SelectPanel>
  <SelectPanel.Header>
    <SelectPanel.Heading>Select labels</SelectPanel.Heading>
  </SelectPanel.Header>
   ...
</SelectPanel>


// after:
<SelectPanel title="Select labels">
   ...
</SelectPanel>
```

PR #3770, opened 2023-09-26
SelectPanel2: 3 tiny bug fixes

siddharthkp:
### Fix 1

In #3758, I forgot that we would need to add Suspense support to `useSlots` first before making footer optional. Default header is fine because it can't be wrapped in suspense, but footer totally can be.

Haven't tried to update useSlots yet, will explore later.

| Before | After |
|--------|--------|
| Select panel with two footers! | Select panel with one footer | 

### Fix 2 

The menu was not closing on it's own because I made a silly mistake. We will be able to catch these silly things once I write some tests 😅 

```diff
- if (props.open === 'undefined') setInternalOpen(false)
+ if (props.open === undefined) setInternalOpen(false)
```

### Fix 3

Added a warning to the ActionMenu story because it's implementation will most definitely change once we implement SelectPanel as a modal

SelectPanel story with a banner that says This implementation will most likely change. [See decision log for more details.](https://github.com/github/primer/discussions/2614#discussioncomment-6879407)


joshblack:
> we would need to add Suspense support to useSlots first before making footer optional

Just wanted to ask, is it possible for `useSlots` to be Suspense-compatible? I would have thought that `React.Children` and `Suspense` would be incompatible and was curious how this will work 👀

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -55,13 +55,13 @@ const SelectPanel = props => {
...
   const onInternalClose = () => {
-    if (props.open === 'undefined') setInternalOpen(false)
+    if (props.open === undefined) setInternalOpen(false)
     if (typeof props.onCancel === 'function') props.onCancel()
   }
   // @ts-ignore todo
   const onInternalSubmit = event => {
     event.preventDefault()
-    if (props.open === 'undefined') setInternalOpen(false)
+    if (props.open === undefined) setInternalOpen(false)
```
broccolinisoup, line 64: I think you fixed it in the other PR, no? https://github.com/primer/react/pull/3770

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -105,7 +105,7 @@ const SelectPanel = props => {
             {slots.header || <SelectPanel.Header />}
             {childrenInBody}
             {/* render default footer as fallback */}
-            {slots.footer || <SelectPanel.Footer />}
+            {slots.footer}
```
broccolinisoup, line 108: >I forgot that we would need to add Suspense support to useSlots first before making footer optional. 

This is probably my ignorance but what is the reason of wrapping the footer in Suspense? 🤔

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: This is that PR :)

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: If we need to fetch data before rendering the list, we'd like the footer to also be hidden until then:


https://github.com/primer/react/assets/1863771/8ce5704a-8162-439a-a105-4ee4c55ce89e

src/drafts/SelectPanel2/SelectPanel.tsx
broccolinisoup: >This is that PR :)
🙈

src/drafts/SelectPanel2/SelectPanel.tsx
broccolinisoup: Ahh that makes sense!!

broccolinisoup (approved):

PR #3773, opened 2023-09-26
Update snapshot tests because of changed avatar

siddharthkp:
I'm sorry, I changed my avatar and now a bunch of visual tests are failing 😭 

We should probably hard code the urls, this is a bit silly. But happy to just unblock for now.


siddharthkp:
@joshblack @broccolinisoup FYI!

joshblack (approved):

PR #3776, opened 2023-09-27
Update lockfiles

iansan5653:
Updates `package-lock.json` files so they aren't always changed when creating a new Codespace. Codespaces run `npm install` on create, so the lockfiles are out of date and I don't want to commit them while working on other things.


joshblack (approved):

joshblack:
@iansan5653 oh weird, totally see what you're saying 🤔

PR #3783, opened 2023-10-02
SelectPanel2: Instant selection variant

siddharthkp:
Context:
- https://github.com/github/primer/issues/2715
- https://github.com/github/primer/issues/2396

```jsx
<SelectPanel selectionVariant="instant">
  ...
</SelectPanel>
```

1.  renders ActionList with `selectionVariant=single`
2. does not render primary footer actions
3. Selecting an item calls `onSubmit` and closes the menu instantly

https://github.com/primer/react/assets/1863771/ad2477f3-9a31-43ad-acc2-434a65cd26e5


src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -244,7 +254,15 @@ const SelectPanelSearchInput = props => {
...
 
 const SelectPanelFooter = ({...props}) => {
-  const {onCancel} = React.useContext(SelectPanelContext)
+  const {onCancel, selectionVariant} = React.useContext(SelectPanelContext)
+
+  const hidePrimaryActions = selectionVariant === 'instant'
+
+  if (hidePrimaryActions && !props.children) {
+    // nothing to render
+    // todo: we can inform them the developer footer will render nothing
```
siddharthkp, line 263: not throwing validation errors yet. Will create a PR for all validations errors together

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -256,23 +274,26 @@ const SelectPanelFooter = ({...props}) => {
...
+      {hidePrimaryActions ? null : (
+        <Box sx={{display: 'flex', gap: 2}}>
+          <Button size="small" type="button" onClick={() => onCancel()}>
+            Cancel
+          </Button>
+          <Button size="small" type="submit" variant="primary">
+            Save
+          </Button>
+        </Box>
+      )}
```
broccolinisoup, line 288: Totally optional - I usually find the syntax below more readable but feel free to ignore if it doesn't resonate with you 🤗 
```suggestion
      {!hidePrimaryActions && (
        <Box sx={{display: 'flex', gap: 2}}>
          <Button size="small" type="button" onClick={() => onCancel()}>
            Cancel
          </Button>
          <Button size="small" type="submit" variant="primary">
            Save
          </Button>
        </Box>
      )}
```

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -127,7 +135,9 @@ const SelectPanel = props => {
                   container: 'SelectPanel',
                   listRole: 'listbox',
                   selectionAttribute: 'aria-selected',
-                  selectionVariant: props.selectionVariant || 'multiple',
+                  selectionVariant:
```
broccolinisoup, line 138: Can we do 

```jsx
selectionVariant: props.selectionVariant === 'instant' ? 'single' : props.selectionVariant,
```
since we have `multiple` as the default value specified above? 🤔

broccolinisoup (approved):
Looks good to me ⭐ Just left two comments/questions

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Good tip! We definitely can

```suggestion
                  selectionVariant: props.selectionVariant === 'instant' ? 'single' : props.selectionVariant,
```

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: > but feel free to ignore if it doesn't resonate with you 🤗

Firstly, thank you! I appreciate that!

because I feel the opposite 😅  

I find double negatives like `don't hide = show` confusing. There's something really clean about `hide = return null`

broccolinisoup: 😅 Fair enough!!

pksjce (commented):
While I'm able to see the story for Instant selection variant, the other stories for `SelectPanel2` seem to be erroring out
Screenshot 2023-10-20 at 9 10 41 pm

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -256,23 +273,26 @@ const SelectPanelFooter = ({...props}) => {
...
+      )}
     </Box>
   )
 }
 SelectPanel.Footer = SelectPanelFooter
 
 // @ts-ignore todo
 SelectPanel.SecondaryButton = props => {
-  return <Button {...props} size="small" type="button" />
+  return <Button {...props} size="small" type="button" block />
```
pksjce, line 295: Just wondering why this `Button` needs to be block type?

siddharthkp:
> While I'm able to see the story for Instant selection variant, the other stories for SelectPanel2 seem to be erroring out

Thanks @pksjce for catching that! Fixed now!

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Good question! 

So, for most cases of SelectPanel you'd see a Save and Cancel button on the right, but then also an optional secondary button on the left:

select panel with save, cancel and a secondary action in the footer

But, in the case of instant variant, there is no Save and Cancel, so we'd like the secondary button to take the full width of the footer:

Without block:

select panel with just a secondary action in the footer, awkwardly sitting in the left corner

With block:

select panel with just a secondary action in the footer, taking the full width of the footer

Now because the footer has 2 divs inside of it - left container and right container, it's always safe for the secondary button to take full width (`block`) of the left container.

image

pksjce: I see! thanks for the explainer!

PR #3784, opened 2023-10-02
Revert "Add aria-selected value to ActionList.Item."

radglob:
This reverts commit `242c634`.

There was some disagreement about how ActionList items should announce if they're selected or not. There's a functional difference between an item being "selected" (the user is currently focused on it) and "checked" (the option was previously chosen and is active). We want to communicate the former, but the `aria-selected` attribute is the incorrect way to do this.

This should not impact anything major in dotcom, as both changes were internal to the component, and did not require any changes by consumers.

### DIFF
--- a/.changeset/wise-boxes-peel.md
+++ b/.changeset/wise-boxes-peel.md
@@ -3,3 +3,5 @@
 ---
 
 Revert "Add aria-selected value to ActionList.Item."
+
+<!-- Changed components: ActionList -->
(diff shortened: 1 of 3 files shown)


lesliecdubs:
@joshblack could you please give this a review when you have a moment and, if all looks good to you and tests pass, merge on behalf of `@radglob`? Thank you 🙏🏻

joshblack (approved):

PR #3787, opened 2023-10-04
Move legacy `primer/primitives` into Primer React

langermank:
Part of https://github.com/github/primer/issues/2054

As we gear up for Primitives v8 (next major) I'm looking for ways we can stopgap the breaking change to provide more time for Primer libraries to upgrade. This PR copies the `dist` that Primer React depends on directly into Primer React which allows us to remove it from Primitives. The only thing truly referenced from Primitives are colors, while most other values were hand written in the theme file. 

Goals:
- Don't break anything
- No major refactor, just a stopgap

#### New

- Adds the colors `dist` from `primer/primitives` and removes dependency 

#### Changed

- Moved typography into the theme file directly and removed the unused stuff
- Converted to es modules

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/legacy-theme/js/colors/light.js
```diff
@@ -0,0 +1,602 @@
+const colors = {
```
joshblack, line 1: Could these files be `.ts` with the variable set `as const`? This could mean that the files are checked with TypeScript and we don't have to check-in the generated definition files.

langermank: Totally! Thanks, will make that change 😄

langermank: @joshblack took a stab at converting it all to JS, can you give it a look? I don't _totally_ know what I'm doing.

joshblack (changes requested):
Just left a small comment for the declaration file 👀

langermank:
@joshblack these are the "legacy" tokens, so I don't expect us to continue updating these at all unless we have a critical fix to make. New tokens are only being added to the new build as CSS vars, so we need to document a dependency on `@primer/primitives` for folks who use Primer React outside of GitHub. I see this as part of the work to move towards CSS variables, but I don't mind holding off on merging this if we think moving to CSS vars _first_ makes more sense.

src/legacy-theme/js/colors/dark.ts
```diff
@@ -0,0 +1,602 @@
+const colors = {
```
siddharthkp, line 1: the most pedantic nitpick, sorry!: the path should be `legacy-theme/ts/`

```diff
- src/legacy-theme/js/colors/dark.ts
+ src/legacy-theme/ts/colors/dark.ts
                    ↑ 

```

src/theme.ts
```diff
@@ -1,8 +1,8 @@
 import {KeyPaths} from './utils/types/KeyPaths'
-import primitives from '@primer/primitives'
+import {colors} from './legacy-theme/js/colors'
 import {partitionColors, fontStack, omitScale} from './utils/theme'
 
-const {lineHeight: lineHeights} = primitives.typography.normal
+// const {lineHeight: lineHeights} = primitives.typography
```
siddharthkp, line 5: kinda strange that we were only using lineHeight and colors from primitives. Makes it easier for this PR I guess 😅

siddharthkp:
What does this change mean for the [primer/primitives dependency in package.json](https://github.com/primer/react/blob/main/package.json#L105)?

src/legacy-theme/js/colors/dark.ts
langermank: Fixed ✅

langermank:
@siddharthkp I'd like to merge this PR and open a separate PR that introduces the CSS variables as part of the output of these TS vars. That way, the diff will show if there are any errors with the raw value.

src/theme.ts
```diff
@@ -1,8 +1,8 @@
 import {KeyPaths} from './utils/types/KeyPaths'
-import primitives from '@primer/primitives'
+import {colors} from './legacy-theme/ts/colors'
 import {partitionColors, fontStack, omitScale} from './utils/theme'
 
-const {lineHeight: lineHeights} = primitives.typography.normal
+// const {lineHeight: lineHeights} = primitives.typography
```
joshblack, line 5:
```suggestion
```

PR #3802, opened 2023-10-10
Remove option large from ButtonGroup story

lukasoppermann:
According to the docs `large` is not a valid option.

#### Changed

Removed `large` option from storybook property

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack (approved):

joshblack:
Question for the team, is there anything we can do at a type-level or at runtime to catch this kind of issue? 🤔 cc @primer/react-reviewers

lukasoppermann:
> Question for the team, is there anything we can do at a type-level or at runtime to catch this kind of issue? 🤔 cc https://github.com/orgs/primer/teams/react-reviewers

Couldn't we technically add a `size` property to buttonGroup instead of just passing the props on?

joshblack:
@lukasoppermann I think that could definitely be an option 💯  Right now it seems like folks just pass the `size` option to `Button` directly. This approach would need some plumbing to set things up (e.g. Button reads from context for it) but is super do-able.

Another pattern might be to iterate through children and provide the `size` prop through `cloneElement` but this path feels discouraged these days 😞 ([context](https://react.dev/reference/react/Children))

PR #3805, opened 2023-10-10
chore(deps): update eslint-plugin-github to v4.10

joshblack:
Background https://github.com/github/primer/issues/2757

This is for the "All repos are using latest accessibility linting configs." criteria and updates primer/react to use eslint-plugin-github 4.9.2 or greater (4.10 in our case)

#### Changed

- Update eslint-plugin-github
- Update `svg` elements to include `aria-hidden`
- Remove unnecessary `title` element
- Add lint ignore for `VisuallyHidden` element that seems to be an intentional choice from the a11y review

### Rollout strategy

None

### Testing & Reviewing

- Verify the lint job passes as expected
- Verify that changes do not regress any behavior or semantics


src/Select/Select.tsx
```diff
@@ -50,7 +50,14 @@ const StyledSelect = styled.select`
 `
 
 const ArrowIndicatorSVG: React.FC<React.PropsWithChildren<{className?: string}>> = ({className}) => (
-  <svg width="16" height="16" fill="currentColor" xmlns="http://www.w3.org/2000/svg" className={className}>
+  <svg
```
broccolinisoup, line 53: Not related to the PR changes - Interesting to see we use SVG here not Octicons. I wonder if I miss any context here 🤔

broccolinisoup (approved):

src/Select/Select.tsx
joshblack: Same 🤔 I'll post a message about it in Slack to see what folks say.

PR #3808, opened 2023-10-10
Updates ActionList async filter example to be accessible

mperrotti:
Related issue: https://github.com/github/accessibility-audits/issues/4487

- Adds filter announcement
- Adds a `<label>` for the text input and removes the `placeholder` text

Screenshot 2023-10-10 at 5 39 04 PM

### Changelog

This is just a documentation update, no changes need to be released.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


pksjce (approved):

src/ActionList/ActionList.examples.stories.tsx
```diff
@@ -187,10 +189,20 @@ export function AsyncListWithSpinner(): JSX.Element {
...
+      </FormControl>
+      <VisuallyHidden
+        isVisible={results.length === 0}
+        aria-live="polite"
+        aria-atomic="true"
+        as={Text}
+        sx={{display: 'block', fontSize: 1, m: 2}}
+      >
+        {results.length === 0 ? 'No branches match that query' : `${results.length} branches match that query`}
+      </VisuallyHidden>
```
joshblack, line 204: For live region support, we have an internal `LiveRegion` component that we could try to consolidate on if it's helpful. 

Right now it's not a standalone component (meaning it needs two components) but we could totally add a standalone variant to it that's just like `<LiveRegion>The message</LiveRegion>` if it would be helpful for any live region based a11y fixes 👀

mperrotti: Thanks for reminding me we had this! I just had to update `LiveRegionOutlet` so we could visually show the `No branches match that query` message.

src/internal/components/VisuallyHidden.tsx
joshblack, line 1: I think we should keep the original behavior here as this file was created intentionally to differ from the other behavior so that `VisuallyHidden` has no ability to not be hidden.

Instead, the intent was that if you need something to be VisuallyHidden you wrap it in the component. If not, you don't wrap it in VisuallyHidden.

Hope that makes sense!

src/ActionList/ActionList.examples.stories.tsx
```diff
@@ -187,10 +189,19 @@ export function AsyncListWithSpinner(): JSX.Element {
...
+        <TextInput onChange={filter} block />
+      </FormControl>
+      <LiveRegion>
+        <LiveRegionOutlet isVisible={results.length === 0} />
+        {results.length === 0 ? (
+          <Text sx={{display: 'block', fontSize: 1, m: 2}}>No branches match that query</Text>
+        ) : (
+          `${results.length} branches match that query`
+        )}
+      </LiveRegion>
```
joshblack, line 203: Ah, sorry I confused this earlier where the message here _would_ be visible (I totally misread the `VisuallyHidden` helper 🤦‍♂️)

The LiveRegion module currently is intended to mirror the helper that exists upstream in dotcom in that it's a visually hidden part of the page. Specifically, things passed as `children` to the `Message` component are analogous to the "announce" helper.

The important part about this is because, for certain live regions types, they must _always_ be in the DOM as dynamically adding them can cause certain Screen Readers to not announce as expected 😞 

For visible content that also is considered a live region, it would be worth having helpers for the status/alert roles (or just practices to follow). In this case, it could be:

```suggestion
      <div role="status">
        {results.length === 0
          ? <Text sx={{display: 'block', fontSize: 1, m: 2}}>No branches match that query</Text>
          : <VisuallyHidden>`${results.length} branches match that query`</VisuallyHidden>
      </div>
```

Would need to test to double-check, but I think that should satisfy the same thing as `aria-live="polite" aria-atomic="true"` and would keep the live region persistent in the DOM. Curious what you think @mperrotti and @ericwbailey 👀

joshblack (commented):
Left a couple of comments, just wanted to sorry about the confusion on my end!

src/internal/components/VisuallyHidden.tsx
mperrotti: So should we plan to remove the older `VisuallyHidden` component? I can open a separate PR for that.

mperrotti:
Thanks @joshblack ! I _think_ I addressed your feedback and removed my changes to `LiveRegion` and `VisuallyHidden`.

joshblack (approved):
Looks great!

PR #3812, opened 2023-10-12
Adds 'inactive' state to buttons

mperrotti:
Relates to https://github.com/github/primer/issues/2676

This came from the work we did on graceful degradation patterns: https://primer.style/design/ui-patterns/degraded-experiences#indicating-a-button-is-non-functional-without-disabling-it

Mock of inactive button states

#### New

Adds `inactive` prop to buttons

#### Changed

No changes

#### Removed

No removals

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Just confirm that the "inactive" button visually looks the same as a "disabled" button.

I've already reviewed this with a11y designers.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Button/ButtonBase.tsx
```diff
@@ -21,6 +21,7 @@ const ButtonBase = forwardRef(
       size = 'medium',
       alignContent = 'center',
       block = false,
+      inactive,
```
langermank, line 24: Curious, what if we just continue to use `disabled` as a prop but change it to use `aria-disabled` instead of `disabled` across the board? 

Is this primarily used for the loading state? So it looked disabled but it can be focused.. can it also still be clicked and follow through with the on click action?

mperrotti: I think there are still rare cases where we want a real `disabled` button.

The loading state is one way we can use this pattern, but it's mostly to support "degraded" buttons: https://primer.style/design/ui-patterns/degraded-experiences#indicating-a-button-is-non-functional-without-disabling-it

mperrotti:
The a11y tests are failing because the inactive button doesn't meet the contrast requirements.

However, I don't think this needs to meet the contrast requirements because it's treated like a disabled control. Is there a way to explicitly skip the color contrast check for this test?

joshblack:
@mperrotti if we're intending for inactive buttons to convey information/receive interactions, would it fall under this scenario for color contrast even if it's considered disabled? https://github.com/github/primer/discussions/652#discussioncomment-7084814

Also wanted to ask, if it's still allowing for interaction is `aria-disabled` appropriate? If I understand right, `aria-disabled="true"` would indicate that the control is not operable

mperrotti:
@joshblack - when I brought this idea to the a11y team last quarter, they recommended `aria-disabled`. The experience would be pretty much the same for AT and keyboard users - the main difference being the `aria-disabled` attribute.

I double-checked my approach with @ericwbailey this afternoon, and he thinks it's ok if "inactive" buttons don't meet the contrast minimum as long as there's some other indication on the page (such as a global banner) that informs the users that things are broken.

Also, I'm not sure if this falls into the same category as the loading state button because the button itself isn't conveying any information. If somebody with low vision hits a submit button and it "disappears" for them when it goes into the loading state, they might think they broke something.

src/Button/styles.ts
```diff
@@ -232,6 +232,9 @@ export const getBaseStyles = (theme?: Theme) => ({
   '&:active': {
     transition: 'none',
   },
+  '&[aria-disabled]': {
+    cursor: 'auto',
+  },
```
pksjce, line 237: Does this need to be added for the new Button as well?

src/Button/ButtonBase.tsx
```diff
@@ -70,6 +71,7 @@ const ButtonBase = forwardRef(
         data-block={block ? 'block' : null}
         data-size={size === 'small' || size === 'large' ? size : undefined}
         data-no-visuals={!LeadingVisual && !TrailingVisual && !TrailingAction ? true : undefined}
+        aria-disabled={inactive}
```
pksjce, line 74: I think the property `aria-disabled` needs to be removed entirely if the `inactive` value is `false`

src/Button/styles.ts
```diff
@@ -7,15 +7,15 @@ export const getVariantStyles = (variant: VariantType = 'default', theme?: Theme
       color: 'btn.text',
       backgroundColor: 'btn.bg',
       boxShadow: `${theme?.shadows.btn.shadow}, ${theme?.shadows.btn.insetShadow}`,
-      '&:hover:not([disabled])': {
+      '&:hover:not([disabled]):not([aria-disabled])': {
```
pksjce, line 10: Or these selectors should be `[aria-disabled="true"]`?

pksjce (commented):
`aria-disabled` can itself be a state or it can have value `true/false`. If we choose to add the value, then we must use style selectors that include the value. If we choose to remove the attribute when not inactive, then the current style selectors will suffice.

This can be seen in the storybooks esp `Playground` where all variants of the `Button` are inactive.

src/Button/ButtonBase.tsx
langermank:
```suggestion
        aria-disabled={inactive ? true : undefined}
```

Agree with PK here

joshblack:
@mperrotti just wanted to clarify this part from the guidance:

> An inert button has the same styles as a disabled button, but it still responds to user input.
>
> An inert button is the accessible option because buttons may not be disabled if they respond to user input.

What I had assumed when reading this was that it still responds to user input, like in the case of the dialog example, and that this would mean that the control is operable. Am I reading this correctly? I get that operable may be a funny word to use here since it's not performing the original action (creating an issue), but I understand it as operable in the WCAG sense where a user is still able to activate the control.

mperrotti:
@joshblack - Yes, that's true. I gave that context when reviewing this with a11y folks, and they still gave me the same feedback.

joshblack:
@mperrotti is there a good place to visit to learn more? Definitely surprising on my end and would love to better understand what's going on 👍

src/Button/ButtonBase.tsx
siddharthkp: > Curious, what if we just continue to use disabled as a prop but change it to use aria-disabled instead of disabled across the board?

I like the general idea of that, but it should come with handling click events with a useful message, which I'm not confident would happen :(

> I think there are still rare cases where we want a real disabled button.

Also fair!

siddharthkp:
> Adds an 'inactive' state to buttons. An inactive button looks disabled and has aria-disabled, but it can still be clicked and focused. This was added to support buttons that are broken due to availability issues, but can't be removed from the page.

100% agree with the intent and direction here.

I feel `inactive` doesn't convey enough context, how do you feel about calling the button `visuallyDisabled` in line with `visuallyHidden`?

The idea being you should not add `disabled` to button, instead you should add `visuallyDisabled` and handle click events with a helpful message.

mperrotti:
@joshblack - maybe we can get this on next week's a11y office hours schedule. I already have 2 topics for today's session 😅 

If you think this is a good idea, I can create the a11y review request issue.

mperrotti:
> I feel `inactive` doesn't convey enough context, how do you feel about calling the prop `visuallyDisabled` in line with `visuallyHidden` which has a similar purpose?

I floated `visuallyDisabled` as a possible name in last week's Primer Patterns working session, but the group liked `inactive` better. I personally think both are effective, and I'd be fine with using `visuallyDisabled` if other reviewers agree.

mperrotti:
I've been iterating on the design based on feedback from @joshblack and the a11y team. I _think_ we have a path forward, but I'm waiting for some more feedback.

mperrotti:
Snapshots (VRT and Jest) will need to be updated once we upgrade @primer/primitives after https://github.com/primer/primitives/pull/774 merges

src/Button/styles.ts
mperrotti, line 10: The a11y team told us not to pass `aria-disabled` for inactive buttons since they still accept interactions such as focus or click.

However, I chose to keep the styles for `disabled` and `aria-disabled` the same to align with PVC and PCSS.

mperrotti:
Contrast check will pass once I upgrade Primitives to the release created by this PR: https://github.com/primer/primitives/pull/776

src/Button/styles.ts
```diff
@@ -291,6 +295,11 @@ export const getButtonStyles = (theme?: Theme) => {
     '&[data-block="block"]': {
       width: '100%',
     },
+    '&[data-inactive]:not([disabled]):not([aria-disabled])': {
+      backgroundColor: `var(--button-inactive-bgColor, ${theme?.colors.btn.inactive.bg})`,
+      border: 0,
```
pksjce, line 300: So having a `border:0` affects the width of the button. The disabled button however has the border.

pksjce (commented):
Looks like the primitives update fixes all the color contrast issues (except dark-dimmed). 👍 

Two questions
- Inactive buttons in different variants all look the same. Is that intentional?
- Inactive buttons dont have a border. This makes them narrower than the other variants and disabled button too.

mperrotti:
> Inactive buttons in different variants all look the same. Is that intentional?

Yes, this is intentional.

---

> Inactive buttons dont have a border. This makes them narrower than the other variants and disabled button too.

Good catch - I'll add a border that's the same color as the background.

mperrotti:
@pksjce - I just checked the inactive button, and it looks like the height is still the same. I think we should keep it as-is.

pksjce (approved):
Lets merge it then 🚀

pksjce (changes requested):
Had more thoughts on this PR and I still am unsure of the lack of variants for inactive button. 
Could you point me to the place where this decision was made?

In the guidelines here, I find that inactive buttons can have tooltips and dialogs as well. Could you build some examples for that?

Most importantly, voiceover simply announces these as a regular button. The screenreader user does not receive any indication about the "inactiveness" of the button. Without this information, I feel the SR user will be confused on what the button exists for. cc - @ericwbailey 

Would love to have a call on actual real world use cases of the inactive button.

ericwbailey:
>Most importantly, voiceover simply announces these as a regular button. The screenreader user does not receive any indication about the "inactiveness" of the button. Without this information, I feel the SR user will be confused on what the button exists for. cc - @ericwbailey

These buttons aren't disabled, they don't perform the behavior a user expects instead. The goal in an inactive state is to communicate that behavior is conditionally restricted due to a platform outage. 

We couldn't trigger modals, tooltips, etc. if we disabled these controls. The presence of `aria-disabled` would infer that they are disabled, and also importantly, inactive. 

This is a weird situation in that they are circumstantially disabled at a very high-level by the platform outage. But since we're going for a graceful degradation approach the specific UI itself still functions in that it is repurposed to communicate this higher-level status.

What we're trying to do is maintain an equivalency in experience for what all users encounter. Here, disabling the controls for just assistive technology users would mean they don't have access to the same information non assistive technology users would experience. Consider a low vision user who also uses a screen reader: using `aria-disabled` would create a mismatch between what they may see and what is being announced.

I'm also banking on a very prominent status banner helping to communicate this, in that it will help to supply additional context.

pksjce:
Hi, so during my investigation I added `aria-disabled` to the `Button` component and found a couple of styling issues
- `aria-disabled='false'` and `aria-disabled='true'` are treated the same. aria-disabled='false' should not add disabled styles to the button


Image


- Invisible button does not respect `aria-disabled` values at all


Image

mperrotti:
`aria-disabled` were leftover from an earlier iteration of the design, but now I've removed them.

langermank (approved):

pksjce (approved):

PR #3815, opened 2023-10-13
Update hover styles for ActionList item

### LINKED ISSUE #2479: ActionMenu does not have hover styles in Dark high contrast theme
### Description

https://user-images.githubusercontent.com/1863771/197777432-8afc5d79-a5ad-43b8-a3ec-00f92897e5b3.mov

### Steps to reproduce

1. Go to https://primer.style/react/ActionMenu
2. Change theme to Dark high contrast
3. Hover on menu items

### Version

v35.12.0

### Browser

Chrome

pksjce:
Closes #2479 

In dark high contrast theme, the `ActionMenu` shows no hover color since the hover color and the menu overlay color are the same in this particular theme. However, if we refer to the `view_components` implementation, we see that they have a hover shadow as well which does the job pretty well in this theme. I've added the same shadow styles to `ActionListItem`. This looks great as the fix for our bug without regressing during other cases.

### Before

https://github.com/primer/react/assets/417268/a3c339cb-3365-4083-837f-58d4feccdf66

### After

https://github.com/primer/react/assets/417268/33f9189c-6b39-4bf9-9e93-f44e0178853d

### Note

This fixes Light High contrast mode ActionItems as well.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -125,7 +125,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         ':hover:not([aria-disabled])': {
           backgroundColor: `actionListItem.${variant}.hoverBg`,
           color: getVariantStyles(variant, disabled).hoverColor,
-          boxShadow: `inset 0 0 0 max(1px, 0.0625rem) ${theme?.colors.border.muted}`,
+          boxShadow: `inset 0 0 0 max(1px, 0.0625rem) ${theme?.colors.actionListItem.default.activeBorder}`,
         },
         '&:focus-visible, > a:focus-visible': {
           outline: 'none',
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -125,6 +125,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         ':hover:not([aria-disabled])': {
           backgroundColor: `actionListItem.${variant}.hoverBg`,
           color: getVariantStyles(variant, disabled).hoverColor,
+          boxShadow: `inset 0 0 0 max(1px, 0.0625rem) ${theme?.colors.border.muted}` ,
         },
         '&:focus-visible, > a:focus-visible': {
           outline: 'none',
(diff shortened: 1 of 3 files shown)


src/ActionList/Item.tsx
```diff
@@ -125,6 +125,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         ':hover:not([aria-disabled])': {
           backgroundColor: `actionListItem.${variant}.hoverBg`,
           color: getVariantStyles(variant, disabled).hoverColor,
+          boxShadow: `inset 0 0 0 max(1px, 0.0625rem) ${theme?.colors.border.muted}`,
```
langermank, line 128:
```suggestion
          boxShadow: `inset 0 0 0 max(1px, 0.0625rem) ${theme?. actionListItem.default. activeBorder}`,
```

I'm not sure if that syntax is right, maybe its:

```
boxShadow: `inset 0 0 0 max(1px, 0.0625rem) actionListItem.default. activeBorder`,
```

Regardless, this should use the ActionList specific color token which you can also see in the Primer View Components version 😄

langermank (changes requested):
Nice! Just one comment about the color token.

joshblack (approved):
LGTM after the token change! 🥳

src/ActionList/Item.tsx
joshblack: I think with the `sx` prop it needs to either be _just_ the token or uses that theme from context, potentially? So with the new token it would be:

```suggestion
          boxShadow: `inset 0 0 0 max(1px, 0.0625rem) ${theme?.colors.actionListItem.default.activeBorder}`,
```

pksjce: Me and @broccolinisoup went back and forth with the colors and got confused! Thanks for this input. I have updated the PR!

broccolinisoup (approved):
Looks really good! 🚀 

Note: I am not sure if it is just me but the after video in the PR description, doesn't render for me. In case you would like to update for future references 😊

langermank (approved):
Looks perfect! Thanks for making that change 😄

PR #3816, opened 2023-10-13
PageHeader: Address dom order issues (screen reader experience feedback from sign-off)

broccolinisoup:
As a part of addressing accessibility sign-off review comments https://github.com/github/primer/issues/1115#issuecomment-1499501472, this PR updates the stories to ensure the order of the elements are the following;

1. `PageHeader.Title`
2. `ContextArea` (all elements)
3. `PageHeader.LeadingAction`
4. `PageHeader.TrailingAction`
5. `PageHeader.Actions`

instead of

1. `ContextArea` (all elements)
2. `PageHeader.LeadingAction`
3. `PageHeader.Title`
4. `PageHeader.TrailingAction`
5. `PageHeader.Actions`

The motivation behind is to make sure the actions on the page that are displayed above or before the main heading (Context area actions and leading action) should come after the main title (`PageHeader.Title`). With this way, we make sure screen reader users who navigate through heading menus don't miss any actions. 

While we updated the dom order in the stories and this is how we will recommend to order the elements, we made sure everything, visually stays the same. (Leveraging CSS grid layout)

[sign-off review comment reference](https://github.com/github/primer/issues/1115#issuecomment-1499501472:~:text=Solution%3A%20move%20the,a%20design%20change.
)
#### Changed

- Layout of the Root element from flex to grid
- The order of PageHeader's sub components in the stories.

### Rollout strategy

This is technically a breaking change but PageHeader is still a draft component, so we are good to realise this as a patch. There is one instance at github/github to be updated and I'll list the commit here to be included in the release PR. 

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation (After I get 👍🏻 on the solution and the direction, I'll update the docs)
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/PageHeader/PageHeader.tsx
```diff
@@ -58,13 +58,24 @@ export type PageHeaderProps = {
 
 const Root: React.FC<React.PropsWithChildren<PageHeaderProps>> = ({children, sx = {}, as = 'div'}) => {
   const rootStyles = {
-    display: 'flex',
-    flexDirection: 'column',
+    display: 'grid',
+    // We have max 4 columns.
+    gridTemplateColumns: 'auto auto auto 1fr',
+    // First column is for context area and takes the full row. Context area items are displayed with flexbox inside.
```
mperrotti, line 64: I'm a little confused. The context area is a row, not a column. Is there something else that goes in the first column?

src/PageHeader/PageHeader.tsx
```diff
@@ -58,13 +58,24 @@ export type PageHeaderProps = {
...
+    // Second column is for leading and trailing action, leading and trailing visuals and the title.
+    // Third column is for navigation and takes the full row.
+    // Fourth column is for description and takes the full row.
```
mperrotti, line 67: These are also rows, not columns.

src/PageHeader/PageHeader.tsx
mperrotti: Since you switched to using grid-template-areas, I don't think we need these comment blocks. The code documents itself! 😄

src/PageHeader/PageHeader.tsx
```diff
@@ -58,13 +58,24 @@ export type PageHeaderProps = {
...
+    gridTemplateAreas: `
+      'context-area context-area context-area context-area'
+      'leading-action title-area trailing-action actions'
+      'navigation navigation navigation navigation'
+      'description description description description'
+    `,
     // TODO: We used hard-coded values for the spacing and font size in this component. Update them to use new design tokens when they are ready to use.
     gap: '0.5rem',
```
mperrotti, line 75:
```suggestion
    gap: 'var(--stack-gap-condensed, 0.5rem)',
```

This helps us prepare for using the Primitives v8 when we finally get there.

mperrotti (commented):
I love the switch to `grid-template-areas`!

mperrotti:
The grid gaps are creating some spacing issues. Maybe instead of using `grid-gap`, we could just put padding on the elements in the grid.

Screenshot 2023-11-22 at 2 54 23 PM

If you feel `grid-gap` is a better way to go, we could conditionally change how many columns the title area spans depending on what elements are present:
- If there's a leading visual and trailing action, keep it as-is
- If there's a leading visual and no trailing action, span from column 2 to 4
- If there's no leading visual or trailing action, span from column 1 to 4

This solution could get kind of complicated though...

src/PageHeader/PageHeader.tsx
broccolinisoup: Yeah you are right. It rather confuses I guess 😄 I'll remove them

src/PageHeader/PageHeader.tsx
broccolinisoup: I am removing them all now

broccolinisoup:
>The grid gaps are creating some spacing issues. 

@mperrotti  I saw the gaps in the VRT tests but didn't realised it was that tricky 😭 Thanks for bringing it to my attention!

>Maybe instead of using grid-gap, we could just put padding on the elements in the grid.

I am a bit hesitant on this; mainly because I think we want these gaps to be around the cells not around the items. Also I feel like your second suggestion aligns more with the overall direction in the implementation. We have grid template areas and since we don't know what sub components might be used in a given PageHeader usage, creating the areas dynamically makes more sense to me, though it is quite complex, I agree 😬 I pushed a super WIP draft PR to explore your second suggestion. https://github.com/primer/react/pull/3970 I'll tag you there when it is less WIP 😄

src/PageHeader/PageHeader.examples.stories.tsx
```diff
@@ -136,6 +143,65 @@ export const PullRequestPage = () => (
   </Box>
 )
 
+export const IssuesPage = () => (
```
broccolinisoup, line 146: This is an example to show how the focus order breaks when the heading has an interactive element inside. I plan to remove this story before merging. It is here for the code review. 

Note: I'll address warning (or even throwing an error with invariant) the cases where the heading includes an interactive element in a follow up PR.

src/PageHeader/PageHeader.examples.stories.tsx
```diff
@@ -209,60 +282,61 @@ export const FilesPage = () => (
           </ActionMenu>
         </PageHeader.ContextAreaActions>
       </PageHeader.ContextArea>
-      <PageHeader.TitleArea>
+      {/* This is a hack for now. Breadcrumbs are not leading visuals but for now we are displaying them under until we get the layout right. */}
```
broccolinisoup, line 285: @mperrotti How do you think we should handle the files page case where there is a breadcrumb component before the main title? I remember we briefly chatted about not to have the Breabcrumb under `PageHeader.LeadingAction` component which makes sense to me however, we haven't discussed it further. Would adding a special `PageHeader.Breadbrumb`  sub component be an option? I am curious what you think!

src/PageHeader/PageHeader.examples.stories.tsx
```diff
@@ -354,6 +428,7 @@ export const WithPageLayout = () => {
         <Box sx={{display: 'flex', flexDirection: 'column', gap: 3}}>
           <Box>
             <Text sx={{fontSize: 0, fontWeight: 'bold', display: 'block', color: 'fg.muted'}}>Assignees</Text>
+
```
broccolinisoup, line 431:
```suggestion
```

src/PageHeader/PageHeader.tsx
```diff
@@ -56,19 +57,90 @@ export type PageHeaderProps = {
...
+
+      for (const item of rootRef.current?.children || []) {
+        observer.observe(item)
+      }
+
+      return () => observer.disconnect()
+    }, [rootRef])
+
+    React.useEffect(() => {
+      // reset the grid template area strings when visibility is changed.
```
broccolinisoup, line 110: Is there a better/cleaner way to do this? 🤔

src/PageHeader/PageHeader.tsx
```diff
@@ -56,19 +57,90 @@ export type PageHeaderProps = {
...
+    const rootStyles = {
+      display: 'grid',
+      // We have 4 columns.
+      gridTemplateColumns: 'auto auto auto 1fr',
+      // We are rendering the template areas dynamically based on the children's visibility.
+      gridTemplateAreas,
+      gap: 'var(--stack-gap-condensed, 0.5rem)',
+    }
+
+    const [visibility, setVisibility] = React.useState({
```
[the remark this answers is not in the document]
broccolinisoup: This needs to be based on the components' visibility rather than checking if the component exists. Because component can exist but not be visible in a given viewport. (I.e. ContextArea can exist in the code and renders in the DOM in all viewport sizes but only visible in narrow viewports.)

src/PageHeader/PageHeader.examples.stories.tsx
mperrotti: Can you remind me why we can't continue to render it in `PageHeader.TitleArea`? That's what we do today. If we change the functionality of `PageHeader.TitleArea`, this could be a breaking change.

I think `PageHeader.Breadcrumb` is much better than rendering it in `PageHeader.LeadingAction`. We'll have to give this some more thought to make sure there's nothing else that should appear in the breadcrumb spot.

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
mperrotti: I think we should try and come up with a way to do this using only CSS. I'd hate to introduce a bunch of re-renders caused by state changes.

src/PageHeader/PageHeader.tsx
mperrotti: I'm sure we can figure something out 🙂

src/PageHeader/PageHeader.tsx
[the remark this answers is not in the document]
broccolinisoup: I know 😭 I wasn't able to think of a way to do it with CSS but let me re-think of any new ideas today. If not, we can work on it tomorrow in our pairing session 🙌🏻

src/PageHeader/PageHeader.tsx
```diff
@@ -219,16 +279,18 @@ const TitleArea: React.FC<React.PropsWithChildren<TitleAreaProps>> = ({
   return (
     <TitleAreaContext.Provider value={{titleVariant: currentVariant, titleAreaHeight: height}}>
```
broccolinisoup, line 280: One thing I realised when a few jest tests failed is that since`PageHeader.LeadingAction` and `PageHeader.TrailingAction` are not listed in the PageHeader.TitleArea (please see the storybook diff) anymore, they don't have access to `PageHeader.TitleArea`s context :/ So when the variant is large, it doesn't have any impact on the leading and trailing action. One solution I can think of is lifting the state up to the PageHeader component, we can rename the prop as well `titleVariant` but let me know if you have any suggestions 🙌🏻

pksjce: Does it make sense to suggest to have a contextprovider only at pageheader level? Seems a bit granular to have it for a section.

broccolinisoup: I think it is an option. When we lift the context provider to the parent component level, the props needs to be passed on that level too though, right? 

Instead of

```
<PageHeader>
    <PageHeader.TitleArea variant="large">
        <PageHeader.Title>This is the main title </PageHeader.Title>
    </PageHeader.TitleArea>
</PageHeader>
```

we will need to have 

```
<PageHeader variant="large">
    <PageHeader.TitleArea >
        <PageHeader.Title>This is the main title </PageHeader.Title>
    </PageHeader.TitleArea>
</PageHeader>
```

Because the variant is only relevant to the title area, this is why we kept the context in a granular level and only in the titlearea. 

If moving the context provider higher to `PageHeader` level and still making the prop on the `PageHeader.TitleArea` component is possible, let me know!

If we need to move the props to the parent level, we could rename the prop name as well maybe like `titleVariant`  or something?

mperrotti: I think we can do this just with CSS and remove the React Context altogether. We can pass `data-size-variant={variant}` on the parent, then pass `data-component="ComponentName"` attribute to `PageHeader.LeadingAction`, `PageHeader.TrailingAction`, and `PageHeader.Title`  do something like this:

```
const StyledRoot = styled(Box)`
&[data-size-variant="small"] {

  [data-component="LeadingAction"] {
    /* small leading action variant styles go here */
  }

  [data-component="TrailingAction"] {
    /* small trailing action variant styles go here */
  }

  [data-component="Title"] {
    /* small title variant styles go here */
  }

}

/* repeat the above for every size variant with special styles */

`
```

We can pair on this if you want

mperrotti: If `PageHeader.TitleArea` will always be a direct child of `PageHeader`, I think we can do this just with CSS and remove the React Context altogether.

- First, we need to grab the value of the `variant` prop from the `PageHeader.TitleArea`
- Then, we can pass `data-size-variant={variant}` on to the `Root` element.
- Then pass `data-component="ComponentName"` attribute to `PageHeader.LeadingAction`, `PageHeader.TrailingAction`, and `PageHeader.Title`  do something like this:

```
const StyledRoot = styled(Box)`
&[data-size-variant="small"] {

  [data-component="LeadingAction"] {
    /* small leading action variant styles go here */
  }

  [data-component="TrailingAction"] {
    /* small trailing action variant styles go here */
  }

  [data-component="Title"] {
    /* small title variant styles go here */
  }

}

/* repeat the above for every size variant with special styles */
`
```

We can pair on this if you want. I tried to think of a way to handle all of this in CSS without needing to know the variant, but I couldn't figure it out.

broccolinisoup: I love this! 💖  I gave this a go and it made me realise that we actually don't need `variant` value in leading/trailing visuals/actions (I _think_). Because we use the `variant` value to set a height for the leading/trailing visuals/actions (2rem for `medium` and `subtitle` variant and 3rem for `large` variant) and it is only because of needing to align these actions and visuals with the title when it has variant font-size due to the `variant` prop. Setting `align-items: 'center'` in the parent page header grid resolved the alignment issue. I still use the data attributes as you suggested here to set the styles for the `Title` component and it is a great idea rather than dynamically rendering styles when the prop is changed. 

You can see the changes in [this commit](https://github.com/primer/react/pull/3816/commits/a6e2335ae4df9a65ccc3c32c425cd05351b8b4bf) and the [playground story](https://primer-402ed3ebc8-13348165.drafts.github.io/storybook/?path=/story/drafts-components-pageheader--playground) is a good one to test. Let me know what you think and if I misunderstand anything!

mperrotti: > Setting align-items: 'center' in the parent page header grid resolved the alignment issue.

@broccolinisoup - setting `align-items: center` won't work for multiple lines of text. The leading/trailing visuals will be aligned to the vertical center of all lines of text, not just the first line of text.

broccolinisoup: Oo I totally missed that the leading/trailing visuals need to be aligned with the first line of text 🤦‍♀️ Sorry, yeah that won't work. I'll also update the VRT tests to include these cases.

src/PageHeader/PageHeader.examples.stories.tsx
```diff
@@ -209,60 +217,63 @@ export const FilesPage = () => (
           </ActionMenu>
         </PageHeader.ContextAreaActions>
       </PageHeader.ContextArea>
-      <PageHeader.TitleArea>
+      {/* This is a hack for now. Breadcrumbs are not leading visuals but for now we are displaying them as leading action until we find a better solution. */}
```
mperrotti, line 220: What about a `PageHeader.Breadcrumb` child?

src/PageHeader/PageHeader.examples.stories.tsx
```diff
@@ -145,8 +146,15 @@ PullRequestPageOnNarrowViewport.parameters = setViewportParamToNarrow
 export const FilesPage = () => (
   <Box sx={{padding: 3}}>
     <PageHeader>
+      <PageHeader.TitleArea>
+        <Text sx={{color: 'rgb(101, 109, 118)'}}>/</Text>
+        <PageHeader.Title as="h1" sx={{fontSize: 1}}>
```
mperrotti, line 151: This looks too large. I think it should match the size of the breadcrumb items.

Screenshot 2023-12-13 at 6 11 06 PM

src/PageHeader/PageHeader.examples.stories.tsx
```diff
@@ -67,36 +69,35 @@ WebhooksOnNarrowViewport.parameters = setViewportParamToNarrow
...
+            <ActionMenu.Overlay width="small">
+              <ActionList>
+                <ActionList.Item onSelect={() => alert('Edit button action')}>Edit</ActionList.Item>
+                <ActionList.Item onSelect={() => alert('Code button action')}>Code</ActionList.Item>
+              </ActionList>
+            </ActionMenu.Overlay>
+          </ActionMenu>
+        </Hidden>
+        <Hidden when={['narrow']}>
+          <Box sx={{display: 'flex'}}>
```
mperrotti, line 95: We need a gap here to add space between the two buttons

Screenshot 2023-12-13 at 6 13 38 PM

src/PageHeader/PageHeader.tsx
```diff
@@ -246,19 +315,19 @@ const LeadingAction: React.FC<React.PropsWithChildren<ChildrenPropTypes>> = ({
   sx = {},
   hidden = hiddenOnNarrow,
 }) => {
-  const {titleAreaHeight} = React.useContext(TitleAreaContext)
-
   return (
     <Box
+      data-component="PH_LeadingAction"
```
mperrotti, line 320: Where do we use the `data-component` attribute?

src/PageHeader/PageHeader.tsx
```diff
@@ -246,19 +315,19 @@ const LeadingAction: React.FC<React.PropsWithChildren<ChildrenPropTypes>> = ({
...
       sx={merge<BetterSystemStyleObject>(
         {
+          gridRow: GRID_ROW_ORDER.LeadingAction,
```
mperrotti, line 323: I think we need to set a height to match the height of a single line of text. `align-items: center` aligns the leading and trailing actions to the vertical center of the text block, not the vertical center of the first line of text.

[Before:](https://primer.style/react/storybook/?path=/story/drafts-components-pageheader-features--with-leading-and-trailing-actions)
Screenshot 2023-12-13 at 6 20 25 PM

[After:](https://primer-402ed3ebc8-13348165.drafts.github.io/storybook/?path=/story/drafts-components-pageheader-features--with-leading-and-trailing-actions)
Screenshot 2023-12-13 at 6 19 41 PM

mperrotti (changes requested):
The

src/PageHeader/PageHeader.tsx
broccolinisoup: I kept these data-components because it is very hard to debug what div is what sub component in the dev tool but now I am actually using them to set height.

src/PageHeader/PageHeader.examples.stories.tsx
broccolinisoup: done!

src/PageHeader/PageHeader.tsx
broccolinisoup: I pushed a commit to fix this and used data-attributes as you recommended [here](https://github.com/primer/react/pull/3816#discussion_r1420804245). 

The challenge with this approach is the CSS specificity over sx styles. 

For example below, I tried to override the font-size of the title but it doesn't apply because the data-component selector has higher specificity. 

```jsx
<PageHeader.Title as="h1" sx={{fontSize: '14px'}}>
   PageHeader component initial layout explorations extra long pull request title
</PageHeader.Title>
```
pageheader title has 14px but the size isn't reflected and in the screenshot it shows the data attribute selector has more specificity than the generated sx class.

We had the same problem with the button styles and used an [approach](https://github.com/primer/react/blob/71baff332b9da6911047d9e05cb83ee031b35a74/src/Button/Button.tsx#L68) to modify the sx styles selectors to have the same data attributes. How do you think about applying the same approach here or do you have a different idea? (You can see how that works in [the dev stories](https://primer.style/react/storybook/?path=/story/components-button-devonly--test-sx-prop))

mperrotti: I think we can use the same approach as the buttons.

broccolinisoup: Actually found out that our [other sx override approach](https://github.com/primer/react/blob/71baff332b9da6911047d9e05cb83ee031b35a74/src/Button/Button.tsx#L11) works better here 🙌🏻  Pushed a commit, seems like they can be overridden now ✨

src/PageHeader/PageHeader.examples.stories.tsx
broccolinisoup: It sounds good to me! I'd like to tackle this in a separate PR (possibly in an API design PR) if that is okay because there is no use case at dotcom for now so it is not a blocker for us ✨

src/PageHeader/PageHeader.examples.stories.tsx
broccolinisoup: with solving the sx prop issue, now this is resolves too ✨

src/PageHeader/PageHeader.features.stories.tsx
```diff
@@ -113,32 +92,34 @@ export const WithActions = () => (
...
-    </PageHeader.TitleArea>
-    <PageHeader.Description>
-      <Text sx={{fontSize: 1, color: 'fg.muted'}}>
-        <Link href="https://github.com/broccolinisoup" sx={{fontWeight: 'bold'}}>
-          broccolinisoup
-        </Link>{' '}
-        created this branch 5 days ago · 14 commits · updated today
-      </Text>
-    </PageHeader.Description>
+    <PageHeader>
```
broccolinisoup, line 110: snapshot changes for this story is expected because not wrapping it with PageHeader is not a correct use case and also caused a lack of space between the grid rows.

mperrotti (approved):
Approved assuming we'll have a follow-up PR to get rid of the hack where we render a breadcrumb in the `LeadingAction` slot.

src/PageHeader/PageHeader.tsx
```diff
@@ -56,19 +62,103 @@ export type PageHeaderProps = {
...
+      gridTemplateColumns: 'auto auto auto 1fr',
+      gridTemplateAreas: `
+      'context-area context-area context-area context-area'
+      'leading-action title-area trailing-action actions'
+      'description description description description'
+      'navigation navigation navigation navigation'
+    `,
+      //  --custom-height is a custom property (passed by sx) that can be used to override the set height.
+      // We don't want these values to be overriden but still want to allow consumers to override them if needed.
+      '&[data-size-variant="large"]': {
```
mperrotti, line 79: Non-blocking nitpick. I don't think we need to repeat this style for every size. I think we can safely assume that `--customHeight` will be defined and won't need a fallback.

broccolinisoup: I think I see what you mean. When the custom height is defined, the size won't matter anymore. 

However, if I don't repeat 
```
{
  height: `var(--custom-height, ${LARGE_TITLE_HEIGHT})`,
},
```

for every size, and only define once like 

```
'[data-component="PH_LeadingAction"], [data-component="PH_TrailingAction"],[data-component="PH_Actions"], [data-component="PH_LeadingVisual"], [data-component="PH_TrailingVisual"]':
{
    height: 'var(--custom-height)',
},
```

it will have the lower css specificity against the styles have the `data-size-variant` attribute therefore the custom height won't be applied. Let me know if I misunderstand you or if you have any recommendation to keep the css specificity higher witout repeating for every size.

mperrotti:
@broccolinisoup - why does this have the "skip changeset" label? Wouldn't we want to cut a release with these changes?

broccolinisoup:
>@broccolinisoup - why does this have the "skip changeset" label? Wouldn't we want to cut a release with these changes?

Yes, I usually add `skip label` to the big PRs in the beginning to avoid the CI job failing. After changes are clear, I write a changeset that reflect the changes. Sorry I didn't communicate about it. Considering this is approved, I'll add a changeset now. For your info, because it requires updates at dotcom, I'll hold off merging the PR until I am back from holidays (Jan 11th)

broccolinisoup:
closing this due to snapshot merge conflicts and opened another PR https://github.com/primer/react/pull/4358

PR #3819, opened 2023-10-13
SelectPanel(draft): move selected items to top on reopen

gr2m:
Work in progress. It's something that @dusave and I tried to figure out while working on a new internal component. Our understanding is that while items should not be re-sorted while the user interacts with the list, the selected items should be moved to the top when the panel is closed and opened again. But we couldn't figure out how to do it.

Dustin suggested to add an `onOpen` callback but I couldn't get it to work even with that, but I might have missed something?

/cc @siddharthkp

Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/drafts/SelectPanel2/SelectPanel.tsx
+++ b/src/drafts/SelectPanel2/SelectPanel.tsx
@@ -52,6 +52,10 @@ const SelectPanel = props => {
   // sync open state
   React.useEffect(() => setInternalOpen(props.open), [props.open])
 
+  const onInternalOpen = () => {
+    if (props.open === undefined) setInternalOpen(true)
+    if (typeof props.onOpen === 'function') props.onOpen()
+  }
   const onInternalClose = () => {
     if (props.open === undefined) setInternalOpen(false)
     if (typeof props.onCancel === 'function') props.onCancel()
@@ -79,7 +83,7 @@ const SelectPanel = props => {
         anchorRef={anchorRef}
         renderAnchor={renderAnchor}
         open={internalOpen}
-        onOpen={() => setInternalOpen(true)}
+        onOpen={onInternalOpen}
         onClose={onInternalClose}
         width={props.width || 'medium'}
         height={props.height || 'large'}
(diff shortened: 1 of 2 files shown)


src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -52,6 +52,10 @@ const SelectPanel = props => {
   // sync open state
   React.useEffect(() => setInternalOpen(props.open), [props.open])
 
+  const onInternalOpen = () => {
+    if (props.open === undefined) setInternalOpen(true)
```
gr2m, line 56: Undo change of current behavior

```suggestion
    setInternalOpen(true)
```

siddharthkp:
Hi!

> Dustin suggested to add an onOpen callback but I couldn't get it to work even with that, but I might have missed something?

I guess a callback could work for the minimal case by making a few tweaks, but it probably isn't the solution for use cases where the list is fetched async after the menu is opened. [This story for example](https://primer.style/react/storybook/?path=/story/drafts-components-selectpanel--b-with-suspended-list)

> Our understanding is that while items should not be re-sorted while the user interacts with the list, the selected items should be moved to the top when the panel is closed and opened again

I tried an implementation of this but wasn't happy with how the user land code worked, so I scraped it and added it to the list of things for this Q. 

I don't know the solution of the top of my head, I'd have to try again. Is it okay if I give you an answer once I get to [the issue for this](https://github.com/github/primer/issues/2644#issuecomment-1766413283)?

gr2m:
> I don't know the solution of the top of my head, I'd have to try again. Is it okay if I give you an answer once I get to [the issue for this](https://github.com/github/primer/issues/2644#issuecomment-1766413283)?

sure, thank you! 👍🏼

PR #3838, opened 2023-10-17
fix(Link): update underline={false} to disable underline

joshblack:
This is a follow-up to the latest release which included support for underline preference in the `Link` component: https://github.com/primer/react/pull/3720

I noticed when trying to update code upstream that `underline={false}` does not work with this preference enabled. This PR changes this behavior so that `underline={false}` is respected regardless of if the preference is enabled.

I'm curious what folks feel about this change 🤔 I couldn't think of an alternative if someone wanted to disable underlines in a specific area (like for cards in a project).

#### Changed

- Update `underline={false}` branch in Link to set `text-decoration` to `'none`' if the value is `false`
- Add `No Underline` Link story
- Add e2e tests for `No underline` Link story

### Rollout strategy

- [x] Patch release


mperrotti (approved):

mperrotti:
> I'm curious what folks feel about this change 🤔 I couldn't think of an alternative if someone wanted to disable underlines in a specific area (like for cards in a project).

I really like this approach. The only thing that feels a little strange is that we don't usually differentiate between an `undefined` (or falsey) prop value and an explicit `false` prop value.

mperrotti:
I pushed a commit to update `src/Link/Link.docs.json`. Feel free to drop it or push new changes.

src/Link/Link.tsx
```diff
@@ -22,11 +22,23 @@ const hoverColor = system({
 
 const StyledLink = styled.a<StyledLinkProps>`
   color: ${props => (props.muted ? get('colors.fg.muted')(props) : get('colors.accent.fg')(props))};
-  text-decoration: ${props => (props.underline ? 'underline' : 'var(--prefers-link-underlines, underline)')};
+  text-decoration: ${props => {
+    if (props.underline === undefined) {
+      return 'var(--prefers-link-underlines, underline)'
```
siddharthkp, line 27: Assuming `BaseStyles` is an optional wrapper... (correct me if it's not!)

Right now, in `main` we set the default to `none`.

```ts
text-decoration: props.underline ? 'underline' : 'none'
```

Instead of changing the default, should we keep that default and let folks change the default by _switching-on_ `--prefers-link-underlines`? 

```ts
text-decoration: props.underline ? 'underline' : 'var(--prefers-link-underlines, none)'
```

PR #3839, opened 2023-10-17
Revert "Support link underline preference"

joshblack:
Reverts primer/react#3720


siddharthkp (approved):
Trust this is a revert PR, don't need to review file changes.

PR #3842, opened 2023-10-19
chore(docs): update docs site to latest doctocat

joshblack:
This is a follow-up to: https://github.com/primer/react/pull/3696

Update our dependency on `@primer/gatsby-theme-doctocat` to `5.0.0`. This is the latest release for doctocat and upgrades Gatsby internally to v4 🥳 

#### Changed

- Update `styled-components` to `5.x` in `devDependencies`
  - This seemed mandatory for things to work as expected, with v4 I could not get the build to work
  - This also seemed okay in the interim as we move to v5 as the baseline in v36
- Update `jest-styled-components` to v7
- Update `NODE_OPTIONS` to include `--no-experimental-fetch` to help with WASM errors in develop and build

### Rollout strategy

None

### Testing & Reviewing

- [ ] Go through the docs site and make sure things work as expected 😅 
- [ ] Make sure `npm run develop` works locally in `docs`
- [ ] Make sure `npm run build` and `npm run build:preview` works locally in `docs`


pksjce (commented):
Looks like the `styled-components` upgrade affected snapshots. Internal class structure has changed. I think it is safe to update them though slightly annoying.

Do you think you need to update this in `/examples` and `/nextjs` too?

joshblack:
Thanks @pksjce! Updated the snapshots 👍 Thankfully it should be good still for examples but I'll keep an eye out for any issues 👀

pksjce:
I think I worded my previous comment wrong. I meant to say maybe you've to upgrade `styled-components` to latest in `/examples` folder too? Is that why the install is failing in ci?

pksjce (approved):

joshblack:
@pksjce makes sense! Definitely seems related, thankfully just running `npm i` seemed to do the trick and now the lockfile is up-to-date.

PR #3843, opened 2023-10-19
Vertically align cell contents in `DataTable`

thyeggman:
### Changelog

This PR vertically aligns cell content for the `DataTable` component.  When a row has actions, the typical pattern is to use an `IconButton`.  However, the button is taller than the row when the row contents are text.  Currently, by default, cell contents are top-aligned, so this creates a disparity between the top and bottom margins for cell content.  This is best demonstrated by [this storybook](https://primer.style/react/storybook/?path=/story/drafts-components-datatable-features--with-row-action-menu&globals=colorScheme:light):

DataTable row with action highlighting vertical alignment mismatch

#### New

Added styles to `TableCell` and `TableHeader`

#### Changed

NA

#### Removed

NA

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


thyeggman:
Here's a screenshot from the [storybook](https://primer-7dfacb37d6-13348165.drafts.github.io/storybook/?path=/story/drafts-components-datatable-features--with-row-action-menu) showing the new alignment:

data table with row contents vertically aligned to the middle

pksjce (approved):
Really like this change. The table cell items look a lot better. I ran the storybooks and compared them with the original and couldn't really find any regressions so far. I will approve but please don't merge till @joshblack or @mperrotti (original component authors) approve

src/DataTable/Table.tsx
```diff
@@ -66,7 +66,10 @@ const StyledTable = styled.table<React.ComponentPropsWithoutRef<'table'>>`
   .TableHeader,
   .TableCell {
     text-align: start;
+    display: flex;
+    align-items: center;
```
mperrotti, line 70: I'm a little worried about how this might affect the readability of rows where some cells have 1 line of text, and other cells wrap to 2+ lines.

Compare these two tables:
vertically centered rows

flush top rows

mperrotti (approved):
I'm open to merging this with the assumption that most table cells will not wrap to multiple lines.

src/DataTable/Table.tsx
thyeggman: Yeah that's a good point.  I'm not sure if the best solution is something that involves proper padding if there's a row action available, or increasing the base height of a row when there is no action.  I don't think anyone but us is using `DataTable` at the moment, and our current plan is to truncate row content when it overflows a cell.

dylanatsmith: Late on this but I think `align-items: baseline` would solve for that specific issue, though not sure if it would introduce problems with other elements like IconButtons or Labels.

Issue #3845, opened 2023-10-20
Tracking issue: implicit role cleanup


keithamus:
### Description

A lot of our React components apply `role=` for elements which have an implicit role. In a lot of these cases, the `role=` assigned is that of the implicit role, in other words the attribute is entirely redundant. 

We can safely remove these roles that are explicitly set, and rely on the implicit role as defined by the HTML spec.

### Steps to reproduce

This effects a variety of components and so there are many reproductions. Visit some components which assign a role, compare the tag/role to the implicit role as defined in the specs.

For example:

`Table.tsx` uses `StyledTable` with `role=table`:

https://github.com/primer/react/blob/216d2a9f57b8acb0701ab4e04a23e057fc325c90/src/DataTable/Table.tsx#L279

However `StyledTable` is itself a `<table>` and so the role is redundant:
 https://github.com/primer/react/blob/216d2a9f57b8acb0701ab4e04a23e057fc325c90/src/DataTable/Table.tsx#L24

A more direct example; `<TableHead>` is a `<thead>` with `role=rowgroup`, but that's also redundant: https://github.com/primer/react/blob/216d2a9f57b8acb0701ab4e04a23e057fc325c90/src/DataTable/Table.tsx#L298

### Version

v35.32.1

### Browser

Chrome, Safari, Firefox, Edge, iOS Safari

keithamus:
https://github.com/primer/react/pull/2996/commits/2c21c81cedf8a58fc7b409485db67fc1ebbdf502 added the roles to DataTable, the comment on some lines:

```js
    // We need to explicitly pass this role because some ATs and browsers drop table semantics
    // when we use `display: contents` or `display: grid` in the table
    // eslint-disable-next-line jsx-a11y/no-redundant-roles
```

lesliecdubs:
@keithamus do you think this issue remains valid and should land in the Primer teams backlog? Asking in part because [continued discussion about these attributes](https://github.slack.com/archives/C01L618AEP9/p1697809185967159?thread_ts=1697794152.996809&cid=C01L618AEP9) uncovered their usefulness in ensuring compatibility across various screen reader/OS/browser combinations.

keithamus:
Sure, let's close it.

PR #3846, opened 2023-10-20
remove role=table from table

keithamus:
Refs #3845 

This removes `role=table` from our `<Table>` component, as it is redundant. `<table>` elements already have an implicit role of `table`, so setting `role=table` does nothing.

#### Removed

Removed `role=table` for `<table>` elements.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/DataTable/Table.tsx
+++ b/src/DataTable/Table.tsx
@@ -276,7 +276,6 @@ const Table = React.forwardRef<HTMLTableElement, TableProps>(function Table(
         aria-labelledby={labelledby}
         data-cell-padding={cellPadding}
         className={clsx('Table', className)}
-        role="table"
         ref={ref}
         style={{'--grid-template-columns': gridTemplateColumns} as React.CSSProperties}
       />


joshblack:
@keithamus if I remember right, this was added due to usage of `display: contents` which does not preserve original element roles in certain screen readers (or at least at the time this was the behavior).

If the HTML structure needs to be updated, it would be super helpful if we opened a PR to update the spec: https://github.com/github/primer/blob/main/apis/data-table-api.md and include @ericwbailey to double-check that this is no longer the case 👍

ericwbailey:
I'd like to not do this. The role additions were specifically added to ensure desktop/mobile VoiceOver worked, and without them they will not or may regress due to Safari's/VoiceOver's instability. 

While we can file bugs against their issue tracker, we have no insight into if they'll be addressed or not, or when. Given that the roles are redundant at best, I would emphasize guaranteeing functionality for the user rather than semantic purity.

keithamus:
Yep I’ll close this

PR #3848, opened 2023-10-20
Support link underline preference (#3720)

joshblack:
This is a follow-up to https://github.com/primer/react/pull/3720 after it was reverted for the latest release. Original description included below.

---

This was initiated by our need for underlined links in blocks of text, and supports dotcom's new user setting to render links with underlines. The link underline setting is currently being tested on dotcom behind a feature flag.

A new setting has been added to the Storybook toolbar that allows us to preview what a component would look like when the user has specified that they prefer link underlines.

Closes https://github.com/github/primer/issues/2536

#### User who turned link underlines on in their GitHub settings
Screenshot 2023-09-14 at 5 45 07 PM

### Additional context about link underline settings

Issue: https://github.com/github/releases/issues/3340
Dotcom PR: https://github.com/github/github/pull/278701

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack:
Quick rollout plan notes from @mperrotti:

1. make sure BaseStyles is being imported or that --prefers-link-underlines is being set
2. merge https://github.com/primer/react/pull/3838 that updates how underline works and cut a PRC release with those changes
3. upgrade PRC in dotcom and request a review lab deployment to test things out

src/BaseStyles.tsx
```diff
@@ -26,6 +26,13 @@ const GlobalStyle = createGlobalStyle<{colorScheme?: 'light' | 'dark'}>`
   details-dialog:focus:not(:focus-visible):not(.focus-visible) {
     outline: none;
   }
+
+  /* Used to fake conditional styles using a technique by Lea Verou: https://lea.verou.me/blog/2020/10/the-var-space-hack-to-toggle-multiple-values-with-one-custom-property/ */
```
pksjce, line 30: omg so clever!

src/BaseStyles.tsx
```diff
@@ -26,6 +26,13 @@ const GlobalStyle = createGlobalStyle<{colorScheme?: 'light' | 'dark'}>`
...
+  /* We have to use a zero-width space character (\u200B) as the value instead of a regular whitespace character because styled-components strips out properties that just have a whitespace value. */
+  :root {--prefers-link-underlines: \u200B;}
+  [data-a11y-link-underlines='true'] {
+    --prefers-link-underlines: initial;
```
pksjce, line 34: what do you think of using the off/on variables like Lea suggested in her blogpost for a clearer understanding of this variable assignment?

src/Link/Link.tsx
```diff
@@ -22,9 +22,9 @@ const hoverColor = system({
 
 const StyledLink = styled.a<StyledLinkProps>`
   color: ${props => (props.muted ? get('colors.fg.muted')(props) : get('colors.accent.fg')(props))};
-  text-decoration: ${props => (props.underline ? 'underline' : 'none')};
+  text-decoration: ${props => (props.underline ? 'underline' : 'var(--prefers-link-underlines, underline)')};
```
siddharthkp, line 25: Assuming `BaseStyles` is an optional wrapper... (correct me if it's not!)

Instead of changing the default, should we keep that default and let folks change the default by _switching-on_ `--prefers-link-underlines`? 

```ts
text-decoration: props.underline ? 'underline' : 'var(--prefers-link-underlines, none)'
```

If `--prefers-link-underlines` is not defined, for example when BaseStyles is not imported, the fallback would be none

src/BaseStyles.tsx
```diff
@@ -26,6 +26,13 @@ const GlobalStyle = createGlobalStyle<{colorScheme?: 'light' | 'dark'}>`
   details-dialog:focus:not(:focus-visible):not(.focus-visible) {
     outline: none;
   }
+
+  /* Used to fake conditional styles using a technique by Lea Verou: https://lea.verou.me/blog/2020/10/the-var-space-hack-to-toggle-multiple-values-with-one-custom-property/ */
+  /* We have to use a zero-width space character (\u200B) as the value instead of a regular whitespace character because styled-components strips out properties that just have a whitespace value. */
+  :root {--prefers-link-underlines: \u200B;}
```
siddharthkp, line 32: Very cool hack! But we're only using this for one property: text-decoration so I'm not sure if we should prefer the explicit `none` here?

src/Link/Link.tsx
siddharthkp: @joshblack @pksjce Does this sound right or am I missing the goal here?

joshblack:
Superseded by: https://github.com/primer/react/pull/3946

PR #3853, opened 2023-10-23
Remove `jsx-a11y/click-events-have-key-events` and `jsx-a11y/no-static-element-interactions` disables

pksjce:
Closes https://github.com/github/primer/issues/2793

### Problem

In the `treeView` component, the structure looks like

```
li(treeItem) [keydown, focus, blur]
- div(container) [onClick, onAuxClick]
-- div(levelIndicator)
-- div(subtree chevron) [onClick]
-- content
```
Right now we have multiple`onClick` events on `div` tags which are non-interactive elements with no `treeitem` role on them as well. This needs to be managed using the eslint disables as mentioned in the issue title

### Solution

My solution is to convert above to below structure and move all the event handlers to the `treeitem` role `li` tag.
```
li(treeItem) [keydown, focus, blur, onClick, onAuxClick]
- div(container) 
-- div(levelIndicator)
-- div(subtree chevron)
-- content
```

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

I have tested this in storybook and it seems to unproblematic.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
+++ b/docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
@@ -7,9 +7,13 @@ import {Placeholder} from '@primer/react/Placeholder'
 import React from 'react'
 import State from '../../../components/State'
 
-const ReactRouterLink = ({to, ...props}) => {
+const ReactRouterLink = ({to, children, ...props}) => {
   // eslint-disable-next-line jsx-a11y/anchor-has-content
-  return <a href={to} {...props} />
+  return (
+    <a href={to} {...props}>
+      {children}
+    </a>
+  )
 }
 
 // Exclude octicons-react's default export because it's deprecated
--- a/docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
+++ b/docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
@@ -8,7 +8,6 @@ import React from 'react'
 import State from '../../../components/State'
 
 const ReactRouterLink = ({to, children, ...props}) => {
-  // eslint-disable-next-line jsx-a11y/anchor-has-content
   return (
     <a href={to} {...props}>
       {children}
--- a/src/TreeView/TreeView.tsx
+++ b/src/TreeView/TreeView.tsx
@@ -453,8 +453,19 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
             event.stopPropagation()
           }}
           onBlur={() => setIsFocused(false)}
+          onClick={event => {
+            if (onSelect) {
+              onSelect(event)
+            } else {
+              toggle(event)
+            }
+          }}
+          onAuxClick={event => {
+            if (onSelect && event.button === 1) {
+              onSelect(event)
+            }
+          }}
         >
-          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
           <div
             className="PRIVATE_TreeView-item-container"
             style={{
@@ -463,35 +474,17 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
               contentVisibility: containIntrinsicSize ? 'auto' : undefined,
               containIntrinsicSize,
             }}
-            onClick={event => {
-              if (onSelect) {
-                onSelect(event)
-              } else {
-                toggle(event)
-              }
-            }}
-            onAuxClick={event => {
-              if (onSelect && event.button === 1) {
-                onSelect(event)
-              }
-            }}
           >
             <div style={{gridArea: 'spacer', display: 'flex'}}>
               <LevelIndicatorLines level={level} />
             </div>
             {hasSubTree ? (
-              // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
               <div
                 className={clsx(
                   'PRIVATE_TreeView-item-toggle',
                   onSelect && 'PRIVATE_TreeView-item-toggle--hover',
                   level === 1 && 'PRIVATE_TreeView-item-toggle--end',
                 )}
-                onClick={event => {
-                  if (onSelect) {
-                    toggle(event)
-                  }
-                }}
               >
                 {isExpanded ? <ChevronDownIcon size={12} /> : <ChevronRightIcon size={12} />}
               </div>
--- a/src/TreeView/TreeView.tsx
+++ b/src/TreeView/TreeView.tsx
@@ -479,12 +479,22 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
               <LevelIndicatorLines level={level} />
             </div>
             {hasSubTree ? (
+              // This lint rule is disabled due to the guidelines in the `TreeView` api docs.
+              // https://github.com/github/primer/blob/main/apis/tree-view-api.md#the-expandcollapse-chevron-toggle
+              // This has specific advice that the chevron be available only to pointer event.
+              // If they take up a button role, they become unnecessary and numerous tab stops.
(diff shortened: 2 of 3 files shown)


.changeset/beige-schools-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Remove `jsx-a11y/click-events-have-key-events` and `jsx-a11y/no-static-element-interactions` disables
```
siddharthkp, line 5: If we don't expect any user changes, we can skip changeset :)

src/TreeView/TreeView.tsx
```diff
@@ -463,35 +474,17 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
...
             </div>
             {hasSubTree ? (
-              // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
               <div
                 className={clsx(
                   'PRIVATE_TreeView-item-toggle',
                   onSelect && 'PRIVATE_TreeView-item-toggle--hover',
                   level === 1 && 'PRIVATE_TreeView-item-toggle--end',
                 )}
-                onClick={event => {
```
siddharthkp, line 490: This might still be needed for the subtree handle to work in the controlled story: https://primer-109b1de022-13348165.drafts.github.io/storybook/?path=/story/components-treeview-features--controlled

https://github.com/primer/react/assets/1863771/3afe6a6c-25f2-47ed-8280-b3641a3cb622

But it is strange that it is a div element 🤔 

Maybe the fix is that this should not be a div or have an explicit role instead?

siddharthkp (changes requested):

src/TreeView/TreeView.tsx
pksjce: Thanks for looking into this. I was not sure what edge case this handled.
Ideally I'd say it needs to have a `role='button'`? But I'm not sure what do we prefer? Do we want `button` element styled to suit the tree or just add the role to the existing element?

siddharthkp: > Ideally I'd say it needs to have a role='button'?

That would have been my guess as well, but it looks like that's not the case.

See this [example of treeview from ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) and the [spec @ericwbailey wrote for our TreeView component](forhttps://github.com/github/primer/blob/main/apis/tree-view-api.md#the-expandcollapse-chevron-toggle)

ericwbailey: Thanks for the tag here! Tree nodes are a little weird in that they're interaction actions, which might suggest buttons (that's what I originally led with), but it's more a special kind of content unto itself. 

The tree pattern it's mapped to an operating system File Explorer tree, so that's a good mental model to use here. The entire tree itself is focusable, but then individual nodes are not buttons to allow for easier traversal, as well as a less polluted interactive element list if navigating via an element list of similar technique.

joshblack: This structure is definitely tough with `jsx-a11y/no-static-element-interactions` since I think we need a way to distinguish between clicking on the functional label for the treeitem versus both the label and subtree within a `role="treeitem"` 😬 In this case, the container for the label is expecting these events to propagate so that it can do specific behavior when interacting with the label.

In general, this rule feels like it can flag in unexpected ways and the way it suggests or indicates to fix the issue seem to be incorrect 😞 (e.g. suggesting to make the item an interactive role)

pksjce: @siddharthkp - Thanks so much for the link! Really enjoyed reading through such a thorough document.

The spec seems to indicate that the chevron icon exists only to be clicked upon (input pointer only). Keyboard and Screenreaders don't need to know about this interaction. If it was a button, it would be an unnecessary burden for Assistive Technology? This is my interpretation. @ericwbailey 

As for the `jsx-a11y/no-static-element-interactions` rule, I feel like I could build a case for it to be ignored or deprioritised on our accessibility scorecard. I would include `jsx-a11y/click-events-have-key-events` in the list too. WDYT @joshblack ?

ericwbailey: >If it was a button, it would be an unnecessary burden for Assistive Technology? This is my interpretation. @ericwbailey

Yup, pretty much. The reason behind the approach here is say there's a Treeview with 1,500 files, as well as nested nodes to contain them. Providing buttons here would create 1,500+ tab stops to get through the component.

By making only the Treeview itself focusable and then relying on ARIA announcements and non-<kbd>Tab</kbd> keyboard traversal, we reduce this to one tab stop. This helps out a ton with a lot of different motor control-related accessibility considerations.

pksjce: Ok, so for now I've removed this code change. Elaborated the reasons in the comments and recorded this info in the  issue https://github.com/github/primer/issues/2793

src/TreeView/TreeView.tsx
```diff
@@ -463,23 +474,15 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
...
-                onSelect(event)
-              } else {
-                toggle(event)
-              }
-            }}
-            onAuxClick={event => {
-              if (onSelect && event.button === 1) {
-                onSelect(event)
-              }
-            }}
```
pksjce, line 477: Hey @siddharthkp - How do you feel about this change? I saw that all storybooks worked when i moved these events to the parent.

siddharthkp:
@pksjce The controlled story is a bit broken, i'm not sure if that's an issue in the component or just the story 🤔 

https://github.com/primer/react/assets/1863771/f22603e3-7b3b-4926-892e-a1787c935515

Tested this link: https://primer-109b1de022-13348165.drafts.github.io/storybook/?path=/story/components-treeview-features--controlled

src/TreeView/TreeView.tsx
siddharthkp: Not confident 😓 https://github.com/primer/react/pull/3853#issuecomment-1794546357

I'm wondering if we're keeping once instance of eslint-ignore of these rules, should we keep the other one as well.

Bonus: This component is very detailed, should we write some tests instead of manual tests to bring confidence?

mperrotti:
> Completely optional/bonus: This component is very detailed, should we write some tests instead of manual tests to increase confidence in the refactor?

Agree with Sid on this one. TreeView is a complex beast, and I'd feel a lot less scared making updates if we had more thorough tests.

docs/src/@primer/gatsby-theme-doctocat/live-code-scope.js
```diff
@@ -7,9 +7,12 @@ import {Placeholder} from '@primer/react/Placeholder'
...
 
-const ReactRouterLink = ({to, ...props}) => {
-  // eslint-disable-next-line jsx-a11y/anchor-has-content
-  return <a href={to} {...props} />
+const ReactRouterLink = ({to, children, ...props}) => {
+  return (
+    <a href={to} {...props}>
+      {children}
+    </a>
+  )
```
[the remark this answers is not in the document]
pksjce: Yes, Lets do this. I'll create a new PR for this.

pksjce:
@siddharthkp - You were right about the controlled component subtree selection bug. Thats exactly what this code does 😓 
Upon revert it got fixed. 
This component could definitely use a lot more tests for these edge cases.

pksjce:
@siddharthkp - I made a commit fixing the issue along with the relevant test case. If you remove the `stopPropagation`, you can see that the new test fails. The component does have a ton of tests but it was lacking one for selection of child elements.

siddharthkp (approved):
Looking good now! 

Just need to remove the debug statements, approving in advance

PR #3386, opened 2023-06-07
Remove PageLayout.Pane position prop.

radglob:
One part of https://github.com/github/primer/issues/1199. This actually removes the prop from the codebase. This is intended to be a breaking change.

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated documentation
- [X] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3367, opened 2023-06-05
Use createRoot instead of ReactDOM.render for React 18 compatibility.

radglob:
Updated ConfirmationDialog to use React 18's createRoot API instead of ReactDOM.render.

Closes https://github.com/primer/react/issues/2990.

We may want to look into using the Portal API to avoid rendering a separate React app in the future.

### Screenshots

Please provide before/after screenshots for any visual changes

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated documentation
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [X] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #3858, opened 2023-10-24
next-major: Update docs and changeset for #3386 and #3367

siddharthkp:
- In https://github.com/primer/react/pull/3386, we deprecated `position` prop for `PageLayout.Pane`, 
  - Updated the docs to remove the deleted props
   - Updated the changeset to say what the new behavior is
   - Removed `positionWhenNarrow` from props because it's not actually used

- In https://github.com/primer/react/pull/3367,
  - Add component name
  - Update type from patch → major

### DIFF
--- a/src/PageLayout/PageLayout.tsx
+++ b/src/PageLayout/PageLayout.tsx
@@ -493,21 +493,6 @@ const isPaneWidth = (width: PaneWidth | CustomWidthOptions): width is PaneWidth 
 }
 
 export type PageLayoutPaneProps = {
-  /**
-   * @deprecated Use the `position` prop with a responsive value instead.
-   *
-   * Before:
-   * ```
-   * position="start"
-   * positionWhenNarrow="end"
-   * ```
-   *
-   * After:
-   * ```
-   * position={{regular: 'start', narrow: 'end'}}
-   * ```
-   */
-  positionWhenNarrow?: 'inherit' | keyof typeof panePositions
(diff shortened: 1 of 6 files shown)


src/SplitPageLayout/SplitPageLayout.docs.json
```diff
@@ -68,12 +68,6 @@
     {
       "name": "SplitPageLayout.Pane",
       "props": [
-        {
-          "name": "position",
-          "type": "| 'start' | 'end' | { narrow?: | 'start' | 'end' regular?: | 'start' | 'end' wide?: | 'start' | 'end' }",
-          "defaultValue": "'start'",
-          "description": ""
-        },
```
siddharthkp, line 76: This is being passed down to PageLayout but PageLayout rejects it 😓

src/PageLayout/PageLayout.tsx
```diff
@@ -555,8 +535,6 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
     {
       'aria-label': label,
       'aria-labelledby': labelledBy,
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      positionWhenNarrow = 'inherit',
```
siddharthkp, line 559: Note for reviewer: Only present in props, not used in the body of the component or passed to another component

joshblack (approved):

PR #3860, opened 2023-10-24
fix(PageLayout): use column-gap property for horizontal spacing in content area

### LINKED ISSUE #3859: Fix missing gap with PageLayout.Pane when coming before content
Context: https://github.com/primer/react/pull/3384#issuecomment-1776975023

There was a regression when a pane comes before content in `PageLayout` where a gap that originally was there between content is no longer present.

  | Before | After |
  |--------|--------|
  | space between pane and content | no space between pane and content |

joshblack:
Closes #3859 

When the `position` prop is removed, the way we configure horizontal spacing using margin or padding no longer works. This PR attempts to replicate this behavior with `column-gap` on the broader flex container which contains the content and pane areas.

#### Changed

- Update `PageLayout` to configure the column gap at the layout level
- Remove usage of `order` property from layouts

### Rollout strategy

- [x] Major release, included in v36


src/PageLayout/PageLayout.tsx
```diff
@@ -435,7 +445,6 @@ const Content: React.FC<React.PropsWithChildren<PageLayoutContentProps>> = ({
         {
           display: isHidden ? 'none' : 'flex',
           flexDirection: 'column',
-          order: REGION_ORDER.content,
```
pksjce, line 438: I think this is breaking storybooks. The "Custom Pane Widths" story seems to have the Pane position changed up.

joshblack:
Closing this as we decided not to move forward with trying to fix this in `next-major` and will restore the original behavior 👀

PR #3864, opened 2023-10-25
SelectPanel2: Implement aria-labelledby and aria-describedby

siddharthkp:
- Built on top of https://github.com/primer/react/pull/3863, merge after that!

| Before | After |
|--------|--------|
| chrome devtools, aria attributes are empty | chrome devtools, aria attributes are set and you can see name and description |


src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -190,10 +200,14 @@ const SelectPanelHeader: React.FC<React.PropsWithChildren> = ({children, ...prop
           {/* heading element is intentionally hardcoded to h1, it is not customisable 
             see https://github.com/github/primer/issues/2578 for context
           */}
-          <Heading as="h1" sx={{fontSize: 14, fontWeight: 600}} {...props}>
+          <Heading as="h1" id={`${panelId}--title`} sx={{fontSize: 14, fontWeight: 600}} {...props}>
```
broccolinisoup, line 203: Seems like we render `Heading` regardless if the title prop is being empty or not. This could lead to label the overlay with an empty string. Should we have a default title as we have have now in SelectPanel v1 or do we need a different enforcement here?

broccolinisoup (commented):
Left a question about the title prop, let me know what you think!

pksjce (approved):

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -190,10 +200,14 @@ const SelectPanelHeader: React.FC<React.PropsWithChildren> = ({children, ...prop
...
             {title}
           </Heading>
-          {description ? <Text sx={{fontSize: 0, color: 'fg.muted', display: 'block'}}>{description}</Text> : null}
+          {description ? (
```
pksjce, line 206: I think it would be nice to have a story with this property? I couldn't find it. Did I miss it?

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: > do we need a different enforcement here?

Oh yep! title will be a required prop for SelectPanel

siddharthkp:
Strange that the preview doesn't deploy after the base branch automatically changed to `main`. (It was pointing a different branch before which is now merged)
 
> Branch "drafts-selectpanel-aria-attributes" is not allowed to deploy to github-pages due to environment protection rules

Tested locally though!

Issue #3856, opened 2023-10-23
Data table pagination breaks when setting defaultPageIndex


apataky:
### Description

Created a data table with pagination, and when I set the `defaultPageIndex` the pagination completely bugs out. See screenshots below:

Settings the pagination:

![Image](https://github.com/primer/react/assets/36776987/115fce50-a5ae-4da3-8839-674fe1c9f19f)

Result:

![Image](https://github.com/primer/react/assets/36776987/6416d534-f47e-429c-b8ef-22215b93bcb3)

Expected to see pages with the correct page numbers (page 8 is the maximum so no 9, 10, 11, 12, 8 again), and page 7 selected and shown

### Steps to reproduce

Create a data table with a pagination component setting the `defaultPageIndex` as follows:

```tsx
import {DataTable, Table} from '@primer/react/drafts'

<Table.Pagination
  aria-label="Pagination for table"
  pageSize={10}
  totalCount={75}
  onChange={pageInfo => { }}
  defaultPageIndex={5}
/>
```

### Version

whatever is in gh/gh

### Browser

Chrome

apataky:
@strackoverflow something else I noticed today that might be related or might be another bug--the table pagination only updates the value the first time `defaultPageIndex` is set. For example if I set it using a hook and it rerenders with a different page index it wont update the UI to show the correct page. I have to force it by updating the key on the component.

This functionality is important to me because I'm trying to tie the page index to a query string, so when a user clicks the back button and the page changes I need the correct pagination to show

cs25-esc:
hi! is the issue resolved?

lesliecdubs:
Hi @cs25-esc 👋  This work is currently in progress and @strackoverflow has linked a PR (refer to the timeline item just before your comment) that may resolve it. Are you looking for an opportunity to contribute to Primer React?

cs25-esc:
Hi @lesliecdubs indeed !!
i have already forked this project and working on issue - TextInput action hover state is incorrect #3867 .
i looking for someone to assign me TextInput action hover state is incorrect #3867 .

apataky:
@strackoverflow do you know how long it takes after the PR is merged to see the fix on dotcom?

strackoverflow:
@apataky yes, we typically publish new releases weekly (though sometimes it ends up taking two weeks if there's an especially complex rollout). So you should hopefully see this fix within a week or two. 

[...]

PR #3866, opened 2023-10-25
fix(DataTable): Fix pagination step generation when `defaultPageIndex` is provided

strackoverflow:
Closes https://github.com/primer/react/issues/3856

This PR fixes a bug in DataTable's pagination where providing a `defaultPageIndex` would sometimes create unexpected steps. 

Here's the before/after when given the following props (attempting to pre-select the 5th page):

```jsx
<Table.Pagination
  pageSize={10}
  totalCount={100}
  defaultPageIndex={4}
/>
```

| Before | After |
|--------|--------|
| before |  after | 

There were two problems that needed to be fixed here:

1. When leading truncation is present, the generated steps would be off by one. To fix this, the leading and trailing truncation elements were moved out of the step creation loop
2. When given a starting index that is within the truncation limit of the last page (usually 7 steps), the loop generating steps would add more steps than there are pages (causing the appearance of "page 11" in the example above)

#### Changed

- Fix occasional incorrect pagination steps being shown when `defaultPageIndex` is provided

### Rollout strategy

- [x] Patch release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/DataTable/Pagination.tsx
+++ b/src/DataTable/Pagination.tsx
@@ -189,6 +189,13 @@ export function Pagination({
     pageSize,
     totalCount,
   })
+  const truncatedPageCount = pageCount > 2 ? Math.min(pageCount - 2, MAX_TRUNCATED_STEP_COUNT) : 0
+  const [offsetStartIndex, setOffsetStartIndex] = useState(() => {
+    return getDefaultOffsetStartIndex(pageIndex, pageCount, truncatedPageCount)
+  })
+  const offsetEndIndex = offsetStartIndex + truncatedPageCount - 1
+  const hasLeadingTruncation = offsetStartIndex >= 2
+  const hasTrailingTruncation = pageCount - 1 - offsetEndIndex > 1
   const getViewportRangesToHidePages = useCallback(() => {
     if (typeof showPages !== 'boolean') {
       return Object.keys(showPages).filter(key => !showPages[key as keyof typeof viewportRanges]) as Array<
@@ -203,70 +210,6 @@ export function Pagination({
     }
   }, [showPages])
 
-  // Build the pagination steps
-  const truncatedPageCount = pageCount > 2 ? Math.min(pageCount, MAX_TRUNCATED_STEP_COUNT) : 0
-
-  // The page index is 0-based, but we want to display 1-based page numbers
-  const page = pageIndex + 1
-
-  // To keep the array strictly numeric types, truncation steps will be indicated by -1
-  // example output given a current page of 6 (pageIndex of 5): [1, -1, 4, 5, 6, 7, 8, -1, 10]
-
-  // We want to start with the currently selected page in the array,
-  // and we're gonna build the array from the middle out, so the currently selected page
-  // will always in the middle when there's truncation on both sides.
-  const steps = [page]
-  const firstPage = 1
-  const lastPage = pageCount
-
-  // Build the rest of the page steps from the middle out,
-  // limited to the maximum number of steps we want to show (including truncation)
-  while (steps.length < truncatedPageCount) {
-    if (steps[0] > firstPage) {
-      // add a step before the first step
-      steps.unshift(steps[0] - 1)
-    }
-    if (steps[steps.length - 1] < lastPage) {
-      // add a step after the last step
-      steps.push(steps[steps.length - 1] + 1)
-    }
-  }
-
-  // Figure out if we will need to display truncation steps (...) on either side
-  const hasLeadingTruncation = steps[0] >= firstPage + 1
-  const hasTrailingTruncation = steps[steps.length - 1] <= lastPage - 1
-
-  if (hasLeadingTruncation) {
-    // replace the current first page with a truncation step (-1 indicates truncation in this array)
-    if (steps[0] !== firstPage + 1) {
-      if (hasTrailingTruncation) {
-        // if there's already trailing truncation, we will *replace* the first step
-        steps[0] = -1
-      } else {
-        // otherwise, insert the truncation step
-        steps.unshift(-1)
-      }
-    }
-
-    // insert the actual first page, which is always shown
-    steps.unshift(firstPage)
-  }
-
-  if (hasTrailingTruncation) {
-    if (steps[steps.length - 1] !== lastPage - 1) {
-      // replace the current last page with a truncation step (-1 indicates truncation in this array)
-      if (hasLeadingTruncation) {
-        // if there's already leading truncation, we will *replace* the last step to retain the correct amount of steps
-        steps[steps.length - 1] = -1
-      } else {
-        steps.push(-1)
-      }
-    }
-
-    // insert the actual last page, which is always shown
-    steps.push(lastPage)
-  }
-
   return (
     <LiveRegion>
       <LiveRegionOutlet />
@@ -278,12 +221,18 @@ export function Pagination({
               className="TablePaginationAction"
               type="button"
               data-has-page={hasPreviousPage ? true : undefined}
+              aria-disabled={!hasPreviousPage ? true : undefined}
               onClick={() => {
                 if (!hasPreviousPage) {
                   return
                 }
 
                 selectPreviousPage()
+                if (hasLeadingTruncation) {
+                  if (pageIndex - 1 < offsetStartIndex + 1) {
+                    setOffsetStartIndex(offsetStartIndex - 1)
+                  }
+                }
               }}
             >
               {hasPreviousPage ? <ChevronLeftIcon /> : null}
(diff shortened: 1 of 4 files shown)


src/DataTable/Pagination.tsx
```diff
@@ -288,6 +294,7 @@ export function Pagination({
                 )
               })
             : null}
+          {hasTrailingTruncation ? <TruncationStep key="truncation-1" /> : null}
```
strackoverflow, line 297: Previously these `TruncationSteps` had keys that aligned with the loop iteration, but I'm assuming the value here doesn't actually matter in this case.

pksjce (commented):
The workings of truncation for pagination seems a bit complex to me and I'm still trying to wrap my head around it.
Meanwhile I created a story in `DataTable.features.stories.tsx` in addition to the `WithPagination` story.
```
export const WithTruncatedPagination = () => {
  const pageSize = 30
  const [pageIndex, setPageIndex] = React.useState(0)
  const start = pageIndex * pageSize
  const end = start + pageSize
  const rows = repos.slice(start, end)
  return (
    <Table.Container>
      <Table.Title as="h2" id="repositories">
        Repositories
      </Table.Title>
      <Table.Subtitle as="p" id="repositories-subtitle">
        A subtitle could appear here to give extra context to the data.
      </Table.Subtitle>
      <DataTable
        aria-labelledby="repositories"
        aria-describedby="repositories-subtitle"
        data={rows}
        columns={[
          {
            header: 'Repository',
            field: 'name',
            rowHeader: true,
          },
          {
            header: 'Type',
            field: 'type',
            renderCell: row => {
              return <Label>{uppercase(row.type)}</Label>
            },
          },
          {
            header: 'Updated',
            field: 'updatedAt',
            renderCell: row => {
              return <RelativeTime date={new Date(row.updatedAt)} />
            },
          },
          {
            header: 'Dependabot',
            field: 'securityFeatures.dependabot',
            renderCell: row => {
              return row.securityFeatures.dependabot.length > 0 ? (
                <LabelGroup>
                  {row.securityFeatures.dependabot.map(feature => {
                    return <Label key={feature}>{uppercase(feature)}</Label>
                  })}
                </LabelGroup>
              ) : null
            },
          },
          {
            header: 'Code scanning',
            field: 'securityFeatures.codeScanning',
            renderCell: row => {
              return row.securityFeatures.codeScanning.length > 0 ? (
                <LabelGroup>
                  {row.securityFeatures.codeScanning.map(feature => {
                    return <Label key={feature}>{uppercase(feature)}</Label>
                  })}
                </LabelGroup>
              ) : null
            },
          },
        ]}
      />
      <Table.Pagination
        defaultPageIndex={12}
        aria-label="Pagination for Repositories"
        pageSize={pageSize}
        totalCount={repos.length}
        onChange={({pageIndex}) => {
          setPageIndex(pageIndex)
        }}
      />
    </Table.Container>
  )
}
```

I still see the two bugs
1. 1 off error in the selected page.
2. More pages than in the params.

I'm not completely sure if the story is passing the right props to the `Pagination` component or if I'm missing something else.

Also just wondering why we don't use the existing top level `Pagination` component here. From stories it does look similar.

strackoverflow:
@pksjce hmm, I'm not able to reproduce that. I added your story to my branch and got the expected truncation:

screenshot

I wouldn't be surprised if I missed something, though... 🤔

joshblack (commented):
Looking great! 🔥 Thanks for tackling the gnarly pagination logic 😅 

I think one requirement that is not clear with this component is that the visual spec specifies that the number of steps between the first page and last page matches the constant `MAX_TRUNCATED_STEP_COUNT` that we have. In other words, there should be max 7 items in between the first and last page steps.

I think when we move the first and last position out of that array we end up with some extra steps that show up, tried to make a quick video to show the difference (top is this PR, bottom is what's in `main`)

https://github.com/primer/react/assets/3901764/550fdc21-da10-413f-b11b-3f1f55600a03

strackoverflow:
@joshblack thanks for that clarification! 

I wanted to see if I could refactor this pagination component a bit and try to make it a little bit easier to follow and also adhere to the design requirements of displaying no more than 7 items (including truncation steps). 

During my monthly pairing session with former Hubber `@seejohnrun` today, we were able to [come up with something](https://github.com/primer/react/pull/3866/commits/87c44f182f9226330f225b6b8defd1c20c032fe4) that feels pretty solid!

The tldr of this refactor is that we now are just building one array that looks something like this (page 6 is the current page in this example, and the -1 values indicate where the truncation steps should be placed):

```
pages = [1, -1, 4, 5, 6, 7, 8, -1, 10]
```

The end result is a nice pagination stepper where the currently selected page is always in the middle when there's both leading and trailing truncation:

https://github.com/primer/react/assets/431533/8276a0ca-1b90-484b-8e48-d56bf1903a22

This refactor removes a lot of the "off by one" footguns that we had in this component and makes adhering to the 7 item limit a bit more straightforward. It also DRYs up the component a bit by removing the need for hardcoded first and last page elements. 

Curious to hear yours and anyone else's thoughts on this!

pksjce:
wow! I really like the new logic. It's easy to read and seems to work smoothly as well! Thanks for the refactor ✨ 
All original bugs seem to be solved.

I did have a couple of nits
1. The actual selected page is +1 of the pageIndex. If I provide a `defaultPageIndex` of say `8`, the actual selected page is `9`. I understand this logic but I felt it was not what I expected when I invoked the component.
2. I'd still insist on a storybook for this specific case ie with a `defaultIndex`. It really makes it easy to debug and test.

strackoverflow:
@pksjce Thank you! I added a new story demonstrating the usage of `defaultPageIndex`, thanks for the suggestion! 

As for the `defaultPageIndex` prop being 0-based and the page numbers being 1-based... I agree it's confusing... I'm wondering if we need to have a discussion about replacing the `defaultPageIndex` prop with something like `defaultPageNumber` if we want to switch to 1-based page indexes? Is that something we should address separately from this fix?

joshblack:
Thanks for taking the time to clean things up @strackoverflow! 🙏 

> The end result is a nice pagination stepper where the currently selected page is always in the middle when there's both leading and trailing truncation

Something that I think we found challenging with this pattern is when the visually active page is different than the page which has focus. This can cause issues when using a screen reader since it can be confusing that the number changes out from under you.

This seems to happens when you select a page, like page 52 in the example, so that it is focused but then the contents of the button updates to be page 54 (or some other page number). This change isn't announced across screen readers like VO or NVDA, in particular with the live region, and so you can get lost thinking you're on page 52 still (the page you selected) but the currently focused button is actually 54.

This is what lead to the current pattern where what you select is the active and focused page. The downside is definitely that it makes clicking on future pages impossible if there is no buffer, but without testing it's hard to tell if this is a problem or if people will just press next 🤷 (from what I remember)

Curious what you think 🤔 The pagination pattern with the active page centered seems to make a lot of sense when the pagination is server-side (since you never have to deal with this focus != active problem as the pages are links) but when the pagination is client-side the pattern seemed to run into that problem above. I wonder if there is something I'm missing here or a way to solve that which you've done in the past 👀

joshblack:
@strackoverflow good question, I think probably our easiest bet for the original issue is option 3. We can always revisit and do option 1 or 2 but I think that will end up being another kind of issue (changing the behavior) and would likely require some review from design and the a11y team. It could also mean that we need to re-request an a11y audit of the behavior (not sure about this, though!). Hope that's fair to say, let me know if I'm misunderstanding anything.

For the offset issue itself, would something like this work for the bounds check?

```tsx
function getDefaultOffsetStartIndex(pageIndex: number, pageCount: number, truncatedPageCount: number): number {
  // When the current page is closer to the end of the list than the beginning
  if (pageIndex > pageCount - 1 - pageIndex) {
    if (pageCount - 1 - pageIndex >= truncatedPageCount) {
      return pageIndex - 3
    }
    return pageCount - 1 - truncatedPageCount
  }

  // When the current page is closer to the beginning of the list than the end
  if (pageIndex < pageCount - 1 - pageIndex) {
    if (pageIndex >= truncatedPageCount) {
      return pageIndex - 3
    }
    return 1
  }

  // When the current page is the midpoint between the beginning and the end
  if (pageIndex < truncatedPageCount) {
    return pageIndex
  }
  return pageIndex - 3
}
```

With the following usage in the component:

```tsx
  const [offsetStartIndex, setOffsetStartIndex] = useState(() => {
    return getDefaultOffsetStartIndex(pageIndex, pageCount, truncatedPageCount)
  })
```

Totally get if this might end up being off 😅 Was feeling bad about the problems earlier so was hoping to help with finding a quick way to fix this lol.

strackoverflow:
@joshblack oh cool, that works. I undid the refactor that I committed and we can just move forward with this 🚢

pksjce (approved):
This looks great to me 👍

joshblack (approved):

PR #3869, opened 2023-10-26
Adapts Blankslate to render proportionally in narrow areas

mperrotti:
Relates to https://github.com/github/primer/issues/2678

**Before:**
Screenshot 2023-10-26 at 2 15 42 PM

**After:**
Screenshot 2023-10-26 at 2 16 03 PM

#### New

Nothing new

#### Changed

- Upgrades `styled-components` so we can use the `@container` rule block
- Adds container queries to adapt the sizing and spacing based on the available area
- Uses Primer Primitives instead of relying on browser-native styles for text elements

#### Removed

Nothing removed

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- Open any Blankslate story and resize the viewport so the Blankslate is < 545px
- Confirm that the sizing and spacing shrinks
- Confirm that no colors have changed

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Blankslate/Blankslate.tsx
```diff
@@ -54,18 +23,26 @@ export type BlankslateProps = React.PropsWithChildren<{
   spacious?: boolean
 }>
 
+const StyledBlankslate = styled.div`
+  --borderColor-default-local-fallback: ${get('colors.border.default')};
+  --fgColor-muted-local-fallback: ${get('colors.fg.muted')};
```
mperrotti, line 28: We don't have the new Primitives colors yet, so we need to provide fallbacks using JS.

.changeset/many-ants-pump.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/react': patch
```
siddharthkp, line 2: Good news: This is the first component in primer/react (outside of drafts) that uses css modules. ✨  

To get the css in dotcom, we need to import [a single css file for primer/react](https://unpkg.com/browse/@primer/react@0.0.0-20231027133432/css/) (similar to primer/css)

Bad news:

This is a breaking change because now we require you to include a css file globally for an existing component that we didn't earlier? Without that stylesheet, existing BlankSlate instances would become unstyled.

Probably fine to squeeze into dotcom, but not for open source?

mperrotti: Do you think there's any way we can use [BaseStyles](https://github.com/primer/react/blob/main/src/BaseStyles.tsx) to import CSS modules?

If this has to be a breaking change, should we hold off on this until the next major release?

mperrotti: > one that adds styles for narrow areas but still uses styled-components (not breaking change)

This isn't going to be possible since `styled-components` doesn't recognize `@container`, and it will get stripped out.

---

I think doing a quick v37 release right after v36 is the best path forward. I don't think there's any way to handle these updates without introducing a breaking change.

siddharthkp: > This isn't going to be possible since styled-components doesn't recognize @container, and it will get stripped out.

As far as I know, styled-components doesn't validate syntax though and strip it out. Here's an example of container queries with styled-components: https://codesandbox.io/s/zen-hill-v939m3?file=/src/App.js

mperrotti: Really?? I wasn't able to get it working.  I tried looking at the CodeSandbox, but I get an error: "Something went wrong Sandbox not found"

siddharthkp: My bad, changed privacy setting to public!

mperrotti: @siddharthkp - [your CodeSandbox](https://codesandbox.io/s/zen-hill-v939m3) is using `styled-components@6.1.0` but we're still on `styled-components@5.1.26`.  A `@container` appears to work in `6.1.0`,  but not `5.1.26`. Here's a CodeSandbox where I tested it: https://codesandbox.io/s/confident-visvesvaraya-9jcpkx?file=/src/App.js

I'm going to try to upgrade us to 6.1.0, but I'm scared about what that might break...

mperrotti:
@siddharthkp - I upgraded `styled-components` so we can use `@container` and removed the Blankslate CSS module.

.changeset/many-ants-pump.md
siddharthkp: Ooooh, good catch! That's my bad

> I'm going to try to upgrade us to 6.1.0, but I'm scared about what that might break...

Same! I wouldn't recommend it given that [doctom is locked to styled-components@5.3.6](https://github.com/github/github/blob/master/ui/packages/react-core/package.json#L57) as well. 😓 We probably should look into that as well, but shouldn't get blocked by it for this change.  

### The problem / Source diving

Trying the same sandbox with v5, it looks like container is applied, however there the 

container style on card container


v6: 

```css
@container (max-width: 600px) {  
  .fKDzaV .sc-bdfCDU { flex-direction:column } 
}
```

v5: 

```css
@container (max-width:600px) 
  .sc-aXZVg { flex-direction:column } 
}
```

which should have been

```diff
- @container (max-width:600px) 
+ @container (max-width:600px) {
/*                             ↑ add missing {      */
  .sc-aXZVg { flex-direction:column } 
}
```

This bug comes from the css parsing library that styled-components uses: [stylis](https://github.com/thysultan/stylis). styled-components@5.3.11 (last in v5) uses `@emotion/stylis` which is a built version of stylus@3.5.4 (from Nov, 2018!!). 

Both emotion and styled-components have dropped `@emotion/stylis` since. This specific bug was fixed in stylus@v4.1.3, which is only available in styled-components@v6 not v5 😢 

### Possible solutions

Because we only need that one opening brace `{` to unblock this work 😅, I have an idea that we can explore if it can work.

Instead of using styled-components to insert the css in head, we use `<style type="text/css">` to inline the 2 lines we need into the BlankSlate dom

working sandbox with styled-components@5: https://codesandbox.io/s/container-styled-v5-fix-sgpf2d?file=/src/App.js:1268-1328

mperrotti: @siddharthkp - so your official suggestion is to add ```<style type="text/css">{ContainerQuery}</style>``` inline?

siddharthkp: To unblock: Yes (hoping that doesn't cause problems in dotcom)

Follow up would be to think about how and when can we do another major release to make a css import required

joshblack: Sorry if I missed this in the thread earlier, but would another option here be to use `createGlobalStyle`? e.g.

```tsx
import { createGlobalStyle } from 'styled-components';

const ContainerQueryStyles = createGlobalStyle`
  .BlankspaceContainer {
    container: blankslate / inline-size;
  }

  @container blankslate (min-width: 34rem) {
    /* ... */
  }
`;

function Blankslate({border, children, narrow, spacious}: BlankslateProps) {
  return (
    <>
      <ContainerQueryStyles />
      <StyledBlankslate data-border={border} data-narrow={narrow} data-spacious={spacious}>
        {children}
      </StyledBlankslate>
    </>
  );
}
```

Was curious if this ended up working as expected. I think this has an advantage of having a shared stylesheet instead of having a `<style>` tag per instance of Blankslate (if it works, that is 😅)

src/Blankslate/Blankslate.tsx
```diff
@@ -1,71 +1,148 @@
...
+  }
+
+  ${StyledBlankslate} .Blankslate-Action:first-of-type {
+    margin-top: var(--stack-gap-normal);
+  }
+
+  ${StyledBlankslate} .Blankslate-Action:last-of-type {
+    margin-bottom: calc(var(--stack-gap-condensed) / 2);
+  }
+  }
```
siddharthkp, line 122: Tested this PR with dotcom and other than this invalid brace, everything else passes CI!

.changeset/many-ants-pump.md
mperrotti: @joshblack - I think using `createGlobalStyle` would still results in `@container (max-width:600px) ` getting written without an open curly brace. Check out @siddharthkp 's [comment above](https://github.com/primer/react/pull/3869#discussion_r1392523030) for more context.

joshblack: Ah, makes sense! Thanks @mperrotti for clarifying

mperrotti:
@siddharthkp - I used [this integration test PR](https://github.com/github/github/pull/301976) and confirmed everything works as expected.

I checked using the `Blankslate` that appears when your custom property filter doesn't return any matches. The path is `/organizations/actions/settings/custom-properties?tab=set-values`.

siddharthkp (approved):
- No breaking changes
- Mike has tested our hack works with dotcom
 
ship it!

PR #3885, opened 2023-10-31
Rename status "drafts" to "experimental"

siddharthkp:
One tiny part of implementing [ADR 010: Merge drafts status into experimental](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-010-drafts-are-experimental.md#adr-010-merge-drafts-status-into-experimental). 

This would update https://primer.style/guides/status

- Matching doctocat PR: https://github.com/primer/design/pull/652

### DIFF
--- a/src/UnderlineNav/UnderlineNav.docs.json
+++ b/src/UnderlineNav/UnderlineNav.docs.json
@@ -1,7 +1,7 @@
 {
-  "id": "drafts_underline_nav2",
+  "id": "experimental_underline_nav2",
   "name": "UnderlineNav",
-  "status": "draft",
+  "status": "experimental",
   "a11yReviewed": true,
   "stories": [],
   "props": [
--- a/script/components-json/component.schema.json
+++ b/script/components-json/component.schema.json
@@ -63,7 +63,7 @@
     "status": {
       "type": "string",
       "description": "The status of the component.",
-      "enum": ["draft", "alpha", "beta", "stable", "deprecated"]
+      "enum": ["draft", "experimental", "alpha", "beta", "stable", "deprecated"]
     },
     "a11yReviewed": {
       "type": "boolean",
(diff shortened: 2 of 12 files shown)


joshblack (commented):
Quick question, for the default story we've been adding a prefix over in: https://github.com/primer/react/blob/main/script/components-json/build.ts#L42-L49 for the component json stuff, with this change will we need to make sure the stories are also updated with this new status so the prefix ends up working out?

src/UnderlineNav/UnderlineNav.docs.json
```diff
@@ -1,7 +1,7 @@
 {
-  "id": "drafts_underline_nav2",
+  "id": "experimental_underline_nav2",
   "name": "UnderlineNav",
-  "status": "draft",
+  "status": "experimental",
```
broccolinisoup, line 4: This is not related to this PR but we should update the status of UnderlineNav to Beta (I'll push a PR)

broccolinisoup:
>Quick question, for the default story we've been adding a prefix over in: https://github.com/primer/react/blob/main/script/components-json/build.ts#L42-L49 for the component json stuff, with this change will we need to make sure the stories are also updated with this new status so the prefix ends up working out?
 
➕ 1  And I think we will also need to make sure the ids are updated in the Playwright tests i.e. from being `drafts-components-tooltip-default` to `experimental-components-tooltip-default`

siddharthkp:
> Quick question, for the default story we've been adding a prefix over in: https://github.com/primer/react/blob/main/script/components-json/build.ts#L42-L49 for the component json stuff, with this change will we need to make sure the stories are also updated with this new status so the prefix ends up working out?

Oooh good catch! I was hoping we can address the stories in a follow up PR, looks like we can't!

siddharthkp:
This is bigger than I wanted to tackle on that day. Closing PR as outdated

Issue #3599, opened 2023-08-07
Unable to style Select with sx prop


robaish:
### Description

Hi! I'm unable to style the `Select` [component](https://github.com/primer/react/blob/main/src/Select.tsx) with the `sx` prop as follows:

`<Select sx={{ borderRadius: '4px', border: '1px solid #eee' }}>`

Looking at the component, it's border-radius is set to `inherit`, but shouldn't the `sx` prop override styling?

### Steps to reproduce

1. Import `Select` from @primer/react
2. Style with `sx` prop
3. See no change

### Version

v35.26.1

### Browser

Chrome

lesliecdubs:
Thanks for reporting, @robaish!

@radglob would you be willing to take a quick look into this and see why `sx` isn't getting applied on `Select`? We think it might be a straightforward fix that the prop just isn't getting passed down.

robaish:
Thanks @lesliecdubs ! 🙏

Nagendra88688:
Hi @lesliecdubs ,
Is this issue solved or should I pick this task ?
Please let me know.

lesliecdubs:
@Nagendra88688 I believe this issue is still outstanding. I would invite you to review the [contribution docs](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md) if you'd like to try contributing a fix.

PR #3888, opened 2023-10-31
Apply `sx` prop to Select's parent for custom styles

strackoverflow:
Closes https://github.com/primer/react/issues/3599

This PR fixes a bug where the `sx` prop for the `Select` component didn't have any effect. 

Given the following custom styles, here is the before/after:

```jsx
<Select sx={{borderRadius: '12px', border: '1px dashed #000000'}}>
```

| Before | After |
|--------|--------|
| before | after | 

It seems as though the custom styles need to be applied to the `TextInputWrapper` component rather than the `StyledSelect` component, but I'm wondering if there are any drawbacks or complications from applying all of the custom styles to this element.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/Select.tsx
```diff
@@ -81,6 +84,7 @@ const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
             fill: disabled ? 'GrayText' : 'FieldText',
           },
         },
+        ...sx,
```
siddharthkp, line 87: Hi! This is 99% there.

`...sx` does a shallow merge with the existing styles so it might accidentally override nested style objects. We have a `merge` utility for sx that deep merges. [Here is a another component that uses merge for reference](https://github.com/primer/react/blob/main/src/ActionList/Divider.tsx#L17-L25).

---

For example, with shallow merge:

```tsx
<Select sx={{
 '@media screen and (forced-colors: active)': {
    color: 'fg.muted'
 }
}}>
```

`...props.sx` will completely overwrite the '@media screen and (forced-colors: active)' key and remove svg styles

siddharthkp (changes requested):
Almost there! Just a tiny change needed

siddharthkp:
> ⚠️ No Changeset found

Here's the format we follow for changesets:

```
---
"@primer/react": patch
---

Select: Add support for `sx` prop

<!-- Changed components: Select -->
```

src/Select.tsx
strackoverflow: Thanks! Makes sense!

siddharthkp (approved):
Looks great!

Feel free to merge it whenever you're ready!

PR #3891, opened 2023-11-01
ActionList: Update the generated HTML for group heading and group list to match with what is recoomended by accessibility

broccolinisoup:
- add ActionList.GroupHeading component
- update action list semantics depending on the given role

Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -158,7 +158,10 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         borderColor: 'var(--divider-color, transparent)',
       },
       // show between 2 items
-      ':not(:first-of-type)': {'--divider-color': theme?.colors.actionListItem.inlineDivider},
+      // all elements except the "first of type" amongst all the data-component="ActionList.Item"
+      ':not(:first-of-type) + &[data-component="ActionList.Item"]': {
+        '--divider-color': theme?.colors.actionListItem.inlineDivider,
+      },
       // hide divider after dividers & group header, with higher importance!
       '[data-component="ActionList.Divider"] + &': {'--divider-color': 'transparent !important'},
       // hide border on current and previous item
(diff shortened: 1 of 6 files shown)


src/ActionList/Item.tsx
```diff
@@ -212,13 +209,16 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
+    const containerProps = _PrivateItemWrapper ? {} : menuItemProps
     const wrapperProps = _PrivateItemWrapper ? menuItemProps : {}
 
     return (
       <ItemContext.Provider value={{variant, disabled, inlineDescriptionId, blockDescriptionId}}>
-        <LiBox
+        <Box
+          data-component="ActionList.Item--Container"
+          // If role is specified on the item, or gathered from the list, we render the wrapper as a div, otherwise as a li
+          as={role || itemRole ? undefined : 'li'}
```
siddharthkp, line 221: While it does a bit icky that we are overriding the ul semantics when we use `<ul role="listbox">`, we don't have to go out of our way to make it a div either 😅 

(if using a div or ul is all the same) I would recommend picking the element that leads to the simplest implementation. 

Related: Does the type for forwardedRef also change based on role? If yes, that feels like an implementation detail that is leaking out (especially when it's not explicitly set by the developer using the component, but internally inside an ActionMenu). Would it help to make that simpler by always rendering `ul`?

src/ActionList/Item.tsx
```diff
@@ -107,6 +104,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       '&[aria-disabled]': {
         cursor: 'not-allowed',
         '[data-component="ActionList.Checkbox"]': {
+          cursor: 'not-allowed',
```
broccolinisoup, line 107:
```suggestion
```

PR #3900, opened 2023-11-03
ActionList: Add ActionList.GroupHeading component and update the existing internal `Header` component for better semantics

broccolinisoup:
Closes #

#### New

As accepted in https://github.com/github/primer/pull/2484, added `ActionList.GroupHeading` API to be able to generate more accessible semantics for the action list.

##### Internal component

Header component in ActionList/Group.tsx renamed to be `GroupHeading` and exported to be used for the new `ActionList.GroupHeading` API but still supports the old API`<ActionList.Group title="group title" >...</ActionList.Group>`

###### for list role

```diff
<h2 id="list-id">Filter By</h2>
<ul role="list" aria-labelledby="list-id">
  <li>
-  <div role="presentation" aria-hidden="true"><span id="group-heading-path">Path</span></div>
+  <h3 id="group-heading-path">Path</h3>
    <ul aria-labelledby="group-heading-path">
      <li>app/assets/modules</li>
      <li>src/react/components</li>
      <li>memex/shared-ui/components</li>
    </ul>
  </li>
 <li>
    <h3 id="group-heading-advanced">Advanced</h3>
    <ul aria-labelledby="group-heading-advanced">
      <li>Owner</li>
      <li>Symbol</li>
      <li>Exclude archived</li>
    </ul>
  </li>
</ul>
```

##### For listbox and menu role

```diff
<h2 id="list-id">Reviewers</h2>
<ul role="listbox" aria-labelledby="list-id">
  <li role="none">
     <div role="presentation" aria-hidden="true"><span id="group-heading-everyone">Everyone</span></div>
-    <ul aria-labelledby="group-heading-everyone" role="group">
+    <ul aria-label="Everyone" role="group"> 
      <li role="option">broccolinisoup</li>
      <li role="option">siddharthkp</li>
      <li role="option">pksjce</li>
    </ul>
  </li>
</ul>
```

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [x] Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))


src/ActionList/ActionList.test.tsx
```diff
@@ -5,6 +5,7 @@ import theme from '../theme'
 import {ActionList} from '.'
 import {behavesAsComponent, checkExports} from '../utils/testing'
 import {BaseStyles, ThemeProvider, SSRProvider, ActionMenu} from '..'
+import data from '../drafts/SelectPanel2/mock-data'
```
broccolinisoup, line 8:
```suggestion
```

src/ActionList/Group.tsx
broccolinisoup, line 1: @joshblack as we chatted and agreed on the action list semantics in https://github.com/github/primer/pull/2787, I tried updating the implementation in https://github.com/primer/react/pull/3891 but it ended up [breaking dotcom tests](https://github.com/github/github/actions/runs/6740064747/job/18322633560?pr=296342)

After this exploration, I tried to splitting the heading and list semantics. In this PR, I am only addressing the heading semantic issues as you will see in the PR description that we still render HTML elements that lose their semantic meanings as their roles override (i.e. ul with role="listbox"), or still undoing semantics. I plan to address them all in a different PR (with a roll out strategy to sotcom). I think with fixing the heading issues, we will be able to close [the root issue](https://github.com/github/accessibility-audits/issues/2939) but I'll plan addressing the rest of the semantics too.  How does it all sound to you? Do you have any concerns? 

I'd love if you could 👀 the PR as well when you get a chance 🙏🏻

siddharthkp (approved):
GroupHeading looks great!

src/ActionList/Group.tsx
joshblack: Sounds good @broccolinisoup!

src/ActionList/ActionList.features.stories.tsx
```diff
@@ -96,6 +96,146 @@ export const WithVisualListHeading = () => (
   </ActionList>
 )
 
+export const ListWithNewGroupHeadingAPI = () => (
```
joshblack, line 99: From a documentation perspective, would it help to have the story name be more like `ListWithHeading` or something similar? Trying to think of when someone might be looking for how to do X (where X is adding a heading for a list) and having a story name that makes sense to them.

The terminology "NewGroupedHeadingAPI" feels very clear for maintainers but from a user perspective might not make as much sense.

src/ActionList/ActionList.features.stories.tsx
```diff
@@ -96,6 +96,146 @@ export const WithVisualListHeading = () => (
...
+        <ActionList.LeadingVisual>
+          <FileDirectoryIcon />
+        </ActionList.LeadingVisual>
+        views/assets/modules
+      </ActionList.Item>
+    </ActionList.Group>
+
+    <ActionList.Group>
+      <ActionList.GroupHeading as="h3">Group 2 Heading</ActionList.GroupHeading>
+      <ActionList.Item onClick={() => {}}>
```
joshblack, line 132: I'm not sure what our conventions are, but when would we use an empty handler for events versus using storybook actions?

src/ActionList/Group.tsx
```diff
@@ -58,32 +79,64 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
       }}
       {...props}
     >
-      {title && <Header title={title} variant={variant} auxiliaryText={auxiliaryText} labelId={labelId} />}
+      {(slots.groupHeading || title) && (
```
joshblack, line 82: Since we're dealing with types where the false-y value may be unexpected (like empty string), would it help to use a ternary here so the false-y branch reliably outputs `null`? Reference for the broader idea here: https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

It may not actually be that helpful here so totally understand if not, just noticed the values themselves aren't booleans and had this come up as a thought 👍

src/ActionList/Group.tsx
```diff
@@ -58,32 +79,64 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
...
+        <GroupHeading
+          title={title}
+          variant={variant}
+          auxiliaryText={auxiliaryText}
+          groupHeadingId={groupHeadingId}
+          as={slots.groupHeading?.props.as}
+        >
+          {slots.groupHeading ? slots.groupHeading.props.children : null}
+        </GroupHeading>
```
joshblack, line 91: In this case, are `GroupHeading` and `slots.groupHeading` the same component? If so, would `cloneElement` work in this scenario?

```suggestion
        {cloneElement(slots.groupHeading, {
          title,
          variant,
          auxiliaryText,
          groupHeadingId,
        })}
```

src/ActionList/Group.tsx
```diff
@@ -58,32 +79,64 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
...
 
-export type HeaderProps = Pick<ActionListGroupProps, 'variant' | 'title' | 'auxiliaryText'> & {
-  labelId: string
-}
+export type GroupHeadingProps = Pick<ActionListGroupProps, 'variant' | 'title' | 'auxiliaryText'> &
+  Omit<ActionListHeadingProps, 'as'> &
+  SxProp &
+  React.HTMLAttributes<HTMLElement> & {
+    as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
+    groupHeadingId?: string
```
joshblack, line 116: Is it possible for someone to pass this as a prop and have it update properly in the parent component? If not, or we don't want folks to be passing it, would this makes sense as something that is read from context?

joshblack (commented):
Looking great! Just left a couple comments/questions, can't wait to hear what you think 👀

src/ActionList/ActionList.features.stories.tsx
broccolinisoup: This is a really good question. I am not really sure either. I find storybook actions useful when the story is about the action like the [data table action stories](https://primer.style/react/storybook/?path=/story/drafts-components-datatable-features--with-row-action) you added are great examples for this use case in my opinion. The stories I added or the ones I mentioned [here](https://github.com/primer/react/pull/3900#discussion_r1385803143) to update are not focusing on actions, they are focusing on semantics, visuals, and having group titles. This is why, I didn't worry about the action itself but I am more than happy to update them to use storybook actions if it is going to be helpful. 
Re-thinking about it now, it is probably helpful to log something (storybook actions could be very beneficial in this case) when the click event is trigger at least so empty event handler isn't very helpful here. 

What are your thoughts??

src/ActionList/Group.tsx
broccolinisoup: I love that, thank you for sharing! I think I am so caught up in the convenience of using `&&` sometimes I forget about javascript quirks. Yes, empty string would mess this up. I'll update the code 💖

src/ActionList/Group.tsx
broccolinisoup: Ooo great point 🔥 Yes we don't want folks to be passing this id in, so totally makes sense to get it from context!

src/ActionList/Group.tsx
broccolinisoup: They are the same components yes.

When I tried cloning though it is complaining about `slots.groupHeading` might be undefined which is a valid case. If ActionList is using the "to-be-deprecated" API which is `<ActionList.Group title="Group heading">` , `slots.groupHeading` is going to be undefined because I use `GroupHeading` component to render both (`title="Group heading"` and `<ActionList.GoupHeading>Group heading</GroupHeading>`),

They are almost the same component except one is taking `title` prop and the other is taking `children` prop to render the heading. I can try splitting them up, or if you have any suggestion I would love to hear it!

src/ActionList/Group.tsx
broccolinisoup: committed [here](https://github.com/primer/react/pull/3900/commits/b929c0046bffc86cc94625993afc2c398cf49516) 🙂

src/ActionList/Group.tsx
broccolinisoup: committed [here](https://github.com/primer/react/pull/3900/commits/b929c0046bffc86cc94625993afc2c398cf49516) 🙂

src/ActionList/Group.tsx
```diff
@@ -58,32 +82,57 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
...
       {...props}
     >
-      {title && <Header title={title} variant={variant} auxiliaryText={auxiliaryText} labelId={labelId} />}
-      <GroupContext.Provider value={{selectionVariant}}>
+      <GroupContext.Provider value={{selectionVariant, groupHeadingId}}>
+        {slots.groupHeading || title ? (
+          <GroupHeading title={title} variant={variant} auxiliaryText={auxiliaryText} as={slots.groupHeading?.props.as}>
+            {slots.groupHeading ? slots.groupHeading.props.children : null}
+          </GroupHeading>
+        ) : null}
```
joshblack, line 90:
```suggestion
        {title && !slots.gropuHeading ? (
          <GroupHeading title={title} variant={variant} auxiliaryText={auxiliaryText} />
        ) : null}
        {!title && slots.groupHeading ? cloneElement(slots.groupHeading, {
          title,
          variant,
          auxiliaryText,
        }) : null}
```

joshblack (approved):
Looks great! Just left a comment with that idea from Slack, figured it'd be more helpful than the snippet I shared there 😂

src/ActionList/ActionList.features.stories.tsx
broccolinisoup: I'm going to take these stories back in this PR. I am thinking to do a follow up one that adds a `@deprecated` notation to the `title` prop and update all title usages with the new API in stories and docs. Let me know if there is any concern!

PR #3902, opened 2023-11-03
Address additional ToggleSwitch a11y feedback

camertron:
Closes https://github.com/github/primer/issues/2520

#### Changed

1. `aria-labelledby` is now a required prop.
2. Disabled switches now use `aria-disabled` instead of `disabled`.
3. A screen reader-only "Loading" label is now displayed after a configurable time period has elapsed to let screen reader users know the toggle is still in progress.
4. Fixed toggling via the On/Off label when the switch is disabled.

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan'

A [PR this summer](https://github.com/primer/react/pull/3520) removed the required `aria-describedby` attribute, and this PR adds it back in. It is possible new instances of `ToggleSwitch` in dotcom omitted this attribute in the interim, so care may be necessary when integrating.

EDIT: It appears there are no missing `aria-labelledby` attributes in dotcom code (see the [integration PR](https://github.com/github/github/pull/296457)). There are [two failing dotcom tests](https://github.com/github/github/actions/runs/6748626423/job/18347400466?pr=296457) that can be fixed by checking for `aria-disabled` instead of calling `toBeDisabled()` and `toBeEnabled()`.

### Testing & Reviewing

The loading-related changes are most easily testable using the "Loading with Delay" story under "Features." You can adjust the amount of toggle delay as well as the amount of delay before the "Loading" label appears and is announced by screen readers.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge
- [x] Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

Re: integration tests, see above.


joshblack:
@camertron would there be a way to have `aria-labelledby` be optional instead of required? I think having a prop become required would fall under our major change bucket 😞 If we could do optional (even have a warning be emitted) and then required that would be awesome

camertron:
@joshblack unfortunately no, that prop needs to be required. I was hoping we could sneak it in, as it _used_ to be required a few months ago and was mistakenly removed.

src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -187,7 +195,7 @@ const ToggleKnob = styled.div<{checked?: boolean; disabled?: boolean}>`
   }
 
   ${props => {
-    if (props.disabled) {
+    if (props['aria-disabled']) {
       return css`
         border-color: ${get('colors.switchTrack.disabledBg')};
       `
```
TylerJDev, line 201:
```suggestion
      return css`
        @media (forced-colors: active) {
          color: GrayText;
        }
        
        border-color: ${get('colors.switchTrack.disabledBg')};
      `
```

Adds the disabled state to be visually present in high contrast mode.

src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -127,7 +134,7 @@ const SwitchButton = styled.button<SwitchButtonProps>`
   }
 
   ${props => {
-    if (props.disabled) {
+    if (props['aria-disabled']) {
       return css`
         background-color: ${get('colors.switchTrack.disabledBg')};
         border-color: transparent;
```
TylerJDev, line 140:
```suggestion
      return css`
        @media (forced-colors: active) {
          border-color: GrayText;
        }
        
        background-color: ${get('colors.switchTrack.disabledBg')};
        border-color: transparent;
```

Adds the disabled state to be visually present in high contrast mode.

src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -249,7 +290,14 @@ const ToggleSwitch = React.forwardRef<HTMLButtonElement, React.PropsWithChildren
...
       >
-        {loading ? <Spinner size="small" /> : null}
+        {isLoadingLabelVisible ? (
+          <VisuallyHidden>
+            <span aria-live="polite" id={loadingLabelId}>
+              Loading
+            </span>
+          </VisuallyHidden>
+        ) : null}
+        {loading ? <Spinner size="small" aria-describedby={loadingLabelId} /> : null}
```
TylerJDev, line 300: Can we move the `aria-describedby={loadingLabelId}` onto the `SwitchButton` instead? Since we're removing the label off of the icon, let's add `aria-hidden` to `<Spinner>`.

src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -31,6 +35,9 @@ export interface ToggleSwitchProps extends Omit<React.HTMLAttributes<HTMLDivElem
    * **This should only be changed when the switch's alignment needs to be adjusted.** For example: It needs to be left-aligned because the label appears above it and the caption appears below it.
    */
   statusLabelPosition?: CellAlignment
+  /** If the switch is in the loading state, this value controls the amount of delay in milliseconds before
+   * the word "Loading" is announced to screen readers. Default: 2000. */
+  loadingLabelDelay?: number
```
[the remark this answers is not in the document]
camertron: I believe any prop added here will get automatically documented on primer.style 👍

src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -166,11 +177,12 @@ const SwitchButton = styled.button<SwitchButtonProps>`
   ${sx}
   ${sizeVariants}
 `
-const ToggleKnob = styled.div<{checked?: boolean; disabled?: boolean}>`
+const ToggleKnob = styled.div<{checked?: boolean}>`
```
broccolinisoup, line 180: We could add the `aria-disabled` type here as well

```
const ToggleKnob = styled.div<{checked?: boolean; 'aria-disabled': React.AriaAttributes['aria-disabled']}>`
```

src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -127,8 +134,12 @@ const SwitchButton = styled.button<SwitchButtonProps>`
   }
 
   ${props => {
-    if (props.disabled) {
+    if (props['aria-disabled']) {
```
broccolinisoup, line 137: we briefly chatted about ToggleSwitch with @siddharthkp today and one question we had was when the toggle is disabled, should we have still have visual cues? Also should the cursor be not allowed on the "On/off" label too?

Quick recording from Sid:

https://github.com/primer/react/assets/1446503/370289fc-4529-4ba9-8210-78af9a042c25

_Video description: Toggle switch disabled story is open and the mouse is moving on the toggle switch displaying not allowed cursor and subtle hover styles appear on the control as well. Then mouse moves over to the Off label and the cursor is pointer_

broccolinisoup (commented):
Left a couple of questions, looking forward to hearing your thoughts 🙌🏻

src/ToggleSwitch/ToggleSwitch.tsx
TylerJDev: I agree with @broccolinisoup, since this region is usually interactive, we'll want to add a visual cue (`cursor: not-allowed`) to indicate that it is currently unavailable when the component is disabled or loading.

TylerJDev:
> @broccolinisoup Also I am not really sure if I understand this prop. How would consumers know how long they should delay the label? 🤔 I might be missing some context though, I would appreciate any insights 🙏🏻

I think as long as there is a default, this is okay. A consumer probably shouldn't change this value unless they have a good reason to. This prop is to ensure that "loading" doesn't announce if the loading state is near instantaneous, so changing it should come sparingly.

TylerJDev (approved):
Outside of @broccolinisoup's comments, looks great! Once those are addressed, we should be good to go! 🎉

src/ToggleSwitch/ToggleSwitch.tsx
broccolinisoup: Thanks for adding the `cursor: not-allowed` to the label as well @camertron! I think the hover state change is not needed when the switch is disabled. See the [PVC example for reference](https://view-components-storybook.eastus.cloudapp.azure.com/view-components/lookbook/inspect/primer/alpha/toggle_switch/disabled).

```
 &:hover:not([aria-disabled]),
  &:focus:focus-visible:not([aria-disabled]) {
    background-color: ${get('colors.switchTrack.hoverBg')};
  }
```

I think that should remove the hover background change when it is disabled. Let me know if you have any concerns!

broccolinisoup (commented):
Thanks for following up with the comments! I left one more comment [here](https://github.com/primer/react/pull/3902#discussion_r1423317925). 

Also, regarding the fixes required at dotcom - this is still a new practise so please share any feedback you have - we are trying to include dotcom changes in the roll out plan section in the template. 

>EDIT: It appears there are no missing aria-labelledby attributes in dotcom code (see the https://github.com/github/github/pull/296457). There are [two failing dotcom tests](https://github.com/github/github/actions/runs/6748626423/job/18347400466?pr=296457) that can be fixed by checking for aria-disabled instead of calling toBeDisabled() and toBeEnabled().

This is great to know that there are changes required when the PR is merged into the release branch and it would be great to include a commit id to make the release coordinator's job easier when debugging the integration tests. Is that reasonable? What do you think?

camertron:
Thanks for taking another look @broccolinisoup!

> it would be great to include a commit id to make the release coordinator's job easier when debugging the integration tests. Is that reasonable? What do you think?

It sounds like a great idea to let the release coordinator know what to fix, but I'm not sure what you mean. Where should the commit ID go?

broccolinisoup:
>It sounds like a great idea to let the release coordinator know what to fix, but I'm not sure what you mean. Where should the commit ID go?

@camertron commit hash is probably a better term sorry 🙂 For example, I created [this integration test PR](https://github.com/github/github/pull/302812) for https://github.com/primer/react/pull/3816 and because there are some changes required at dotcom when the PRC pull request is merged, I fixed those issues in the integration PR and you can see the commits [here](https://github.com/github/github/pull/302812/commits). This is still WIP but once it is ready to ship, I plan to share this [commit hash](https://github.com/github/github/pull/302812/commits/15302c3c6414a8f18717e08caf7088efce2fd357) (probably fixing the linting and force pushing everything into once commit) with the release coordinator, so that they can cherry-pick it on the PR that is updating the primer react version at dotcom. Is that helpful? And this is just one way really, let us know if you prefer a different way to share the changes or if you have any feedback about this "process" 🙌🏻

camertron:
@broccolinisoup ahh ok I get it! Great idea, I can work on that 👍

TylerJDev:
Hey @camertron 👋 

I'm coming back to this to say I recently merged a PR that was in a similar state, and it initially had test failures in dotcom. I created the integration test PR, and [added all of the fixes needed in that PR](https://github.com/github/github/pull/304867#pullrequestreview-1778367991) to make sure the tests were all green. From there, the release conductor was able to implement those changes in the [release PR](https://github.com/github/github/pull/307671).

I think all we'd need to do is get the integration test PR to a point where all tests are passing, and provide that information to whomever is doing the release that week. 

@broccolinisoup, can you confirm the above? I think overall this PR should be ready to go once we know those test failures can be fixed easily without friction. Also, I'm wondering if we want to formalize this process somewhere, if we haven't already? 🤔

PR #3903, opened 2023-11-05
Make resize vertical splitter keyboard accessible

pksjce:
Closes https://github.com/github/primer/issues/2810

### Context

Earlier [guidance for SplitPageLayout accessibility](https://github.com/primer/react/pull/2593) was to create a visually hidden form with range input to determine the position of the splitter. However this hidden form swallows the focus when tabbing through the page layout. 

#### New

The proposed change here is to follow [Window Splitter a11y guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/) and support keyboard navigation on the vertical splitter.

https://github.com/primer/react/assets/417268/1100611d-ae34-4b79-a63f-15403a9e012f

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Tab through the `Resizable Pane` storybook under `PageLayout.features.stories` and navigate to focus on the vertical splitter. Use the left and right arrow keys to move it sideways. Use Voiceover to here the announcement on current pane width updated with every move.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/PageLayout/PageLayout.tsx
+++ b/src/PageLayout/PageLayout.tsx
@@ -30,7 +30,7 @@ const PageLayoutContext = React.createContext<{
   padding: keyof typeof SPACING_MAP
   rowGap: keyof typeof SPACING_MAP
   columnGap: keyof typeof SPACING_MAP
-  paneRef?: React.RefObject<HTMLDivElement>
+  paneRef: React.RefObject<HTMLDivElement>
   enableStickyPane?: (top: number | string) => void
   disableStickyPane?: () => void
   contentTopRef?: (node?: Element | null | undefined) => void
@@ -39,6 +39,7 @@ const PageLayoutContext = React.createContext<{
   padding: 'normal',
   rowGap: 'normal',
   columnGap: 'normal',
+  paneRef: {current: null},
 })
 
 // ----------------------------------------------------------------------------
@@ -240,11 +241,11 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
   const [currentWidth, setCurrentWidth] = React.useState(0)
 
   React.useEffect(() => {
-    if (paneRef?.current !== null) {
-      const paneStyles = getComputedStyle(paneRef?.current as Element)
+    if (paneRef.current !== null) {
+      const paneStyles = getComputedStyle(paneRef.current as Element)
       const maxPaneWidthDiffPixels = paneStyles.getPropertyValue('--pane-max-width-diff')
       const minWidthPixels = paneStyles.getPropertyValue('--pane-min-width')
-      const paneWidth = paneRef?.current.getBoundingClientRect().width
+      const paneWidth = paneRef.current.getBoundingClientRect().width
       const maxPaneWidthDiff = Number(maxPaneWidthDiffPixels.split('px')[0])
       const minPaneWidth = Number(minWidthPixels.split('px')[0])
       const viewportWidth = window.innerWidth
@@ -315,7 +316,7 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
       window.removeEventListener('mouseup', handleDragEnd)
       document.body.removeAttribute('data-page-layout-dragging')
     }
-  }, [isDragging, isKeyboardDrag])
+  }, [isDragging, isKeyboardDrag, currentWidth, minWidth, maxWidth])
 
   return (
     <Box
@@ -655,9 +656,7 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
 
     const isHidden = useResponsiveValue(responsiveHidden, false)
 
-    const {rowGap, columnGap, enableStickyPane, disableStickyPane} = React.useContext(PageLayoutContext)
-    let {paneRef} = React.useContext(PageLayoutContext)
-    paneRef = paneRef || useRef<HTMLDivElement>(null)
+    const {rowGap, columnGap, enableStickyPane, disableStickyPane, paneRef} = React.useContext(PageLayoutContext)
 
     React.useEffect(() => {
       if (sticky) {
@@ -787,7 +786,7 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
           }}
           // Ensure `paneWidth` state and actual pane width are in sync when the drag ends
           onDragEnd={() => {
-            const paneRect = paneRef?.current?.getBoundingClientRect()
+            const paneRect = paneRef.current?.getBoundingClientRect()
             if (!paneRect) return
             updatePaneWidth(paneRect.width)
           }}
--- a/src/PageLayout/PageLayout.tsx
+++ b/src/PageLayout/PageLayout.tsx
@@ -314,6 +314,8 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
     return () => {
       window.removeEventListener('mousemove', handleDrag)
       window.removeEventListener('mouseup', handleDragEnd)
+      window.removeEventListener('keydown', handleKeyDrag)
+      window.removeEventListener('keyup', handleKeyDragEnd)
       document.body.removeAttribute('data-page-layout-dragging')
     }
   }, [isDragging, isKeyboardDrag, currentWidth, minWidth, maxWidth])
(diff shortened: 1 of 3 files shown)


siddharthkp:
Hi! 

(Probably an old bug that's only surfacing now) Ran into a strange issue where the position would move more and more every keystroke and found that handleKeyDragEnd is running more and more times with every keystroke 🤔 

https://github.com/primer/react/assets/1863771/14bb4cb2-54a2-4c3d-90a2-459edbedb22d

siddharthkp (changes requested):
Implementation looks great!

Found one strange bug, other than that, looks good!

pksjce:
Hi @siddharthkp - Thanks for flagging that! I noticed this bug too. The fix was to remove the event listeners in the `useEffect`'s return function. Seems to be more stable now 👍

siddharthkp:
> Seems to be more stable now 👍

Yes, feels much better now!

siddharthkp:
Optional side note: We change the cursor when dragging with pointer, should we also change it for keyboard or not? 🤔 

Feels a tiny bit janky based on your keystrokes, but maybe it's fine. Leaving it up to you

https://github.com/primer/react/assets/1863771/d05f489b-1da4-473b-9247-3ff56fd3c6bd

siddharthkp (approved):
Approving from my side ✅ 

Should check with accessibility team as well

pksjce:
> Optional side note 1: We change the cursor when dragging with pointer, should we also change it for keyboard or not? 🤔

Feels a tiny bit janky based on your keystrokes, but maybe it's fine. Leaving it up to you

So I thought about this but imagined that if somebody is using only keyboard, they wouldn't use the mouse at all (which I believe gives that feeling.So they wouldn't see a mouse pointer? I'm not sure how usage for this will be like.

> Not for this PR, but a future optimisation that would be nice is to make the resizing incrementally bigger the longer you press an arrow key (and then reset it after cool off period)

Great idea! I was kinda bored while testing this and this addition would be perfect. It would be fun to implement. Do you think it might have any accessibility penalty?

Eric is on leave, whom would you recommend I tag for accessibility review?

siddharthkp:
> So I thought about this but imagined that if somebody is using only keyboard, they wouldn't use the mouse at all (which I believe gives that feeling.So they wouldn't see a mouse pointer? I'm not sure how usage for this will be like.

Ah yes, good point. I genuinely don't know if that's the case or not 😇

> Eric is on leave, whom would you recommend I tag for accessibility review?

@ichelsea or @TylerJDev maybe?

ichelsea (commented):
Thanks for looking for ways to continuously improve our components!

I'm not finding any issues with using the splitter with just my keyboard (and I'm also behind the idea to leave the cursor change alone when controlling with just the keyboard).

However, I'm unable to control this splitter component with VoiceOver on Safari or NVDA on Edge. It's read out to me, but the arrow keys just move me to the next element and there isn't an intuitive way to figure out how to use the control with a screen reader. I'm unsure of advice to give on this one as it dips into a more technical area that I'm unfamiliar with, so I'm going to tag in our accessibility engineering team to have a look while Eric is out 🙂.

kendallgassner:
@ichelsea What you are noticing is that the screen reader uses certain keys, in this case arrow keys, to interact with the virtual buffer. Essentially, the screen reader is using the arrow keys to move around the accessibility tree and not passing those key events back to the javascript.   

 

Link to reading: [Time to revisit accesskey?](https://tink.uk/time-to-revisit-accesskey/)

TylerJDev:
I really like all the work you've put into this fix! ✨

> However, I'm unable to control this splitter component with VoiceOver on Safari or NVDA on Edge.

I took a look at this, and I'm thinking that the behavior with NVDA (and potentially other screen readers) is due to the role of the resizable control. It seems that `role="separator"` does not toggle NVDA's "focus mode", which means that the arrow keys are not treated as you'd expect when you focus onto the widget.

This can be adjusted by entering focus mode when you're on the separator control (<kbd>Insert</kbd> (NVDA modifier) + <kbd>space</kbd>). Arrow keys should work once you've toggled focus mode on. While this does allow you to resize the control, I don't believe this is optimal as you'd have to know to switch modes in order to use the control beforehand. 

Using a different role _could_ be the option. Some roles will default to focus mode when focused, meaning that the user wouldn't have to manually switch. I see it was [suggested to utilize a `range` input](https://github.com/primer/react/pull/2593#issuecomment-1325528633) to operate the splitter. The closest ARIA role that maps to that seems to be the [`slider` role](https://www.w3.org/TR/wai-aria-1.2/#slider). This _might_ fix the issue with the arrow keys in NVDA, but would require some additional testing with other AT to make sure it's properly supported, which I'm not too confident in right now.

There are a few other solutions that we might be able to use, but I think @kendallgassner's suggestion of Office Hours is also a great way to get more opinions. I'll loop back to this once I can do some more testing.

> However this hidden form swallows the focus when tabbing through the page layout.

@pksjce, Does this mean that the focus outline wasn't shown, or is this a different issue? If it's the former then I'm wondering if `:focus-within` might've worked here?

siddharthkp:
> > However this hidden form swallows the focus when tabbing through the page layout.
> Does this mean that the focus outline wasn't shown, or is this a different issue? If it's the former then I'm wondering if :focus-within might've worked here?

Inside the pane is a visually hidden form which gets focus 😅 

https://github.com/primer/react/assets/1863771/dd63cc6d-f219-4ddf-9eab-28428b8a22d8

TylerJDev:
> Inside the pane is a visually hidden form which gets focus

Did we switch from the form due to there not being a focus indicator while using it?

pksjce:
@TylerJDev - The form itself is supposed to be used only by the screenreader. Sighted users are not supposed to know the form's existence. This raised an eslint error for the rule`github/a11y-no-visually-hidden-interactive-element`. This change in design is to 
1. Fix this eslint error
2. Move to a [more standard solution for such use cases](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/).

I have not tested this feature in NVDA as of yet. Would love to pair to do this if possible?
On voiceover, this was how it announced https://www.loom.com/share/04ef6b9e0d664ccaa17b166ed8b9f016

Would also love some help on process to bring this issue to the next office hours.

TylerJDev:
> This raised an eslint error for the rule `github/a11y-no-visually-hidden-interactive-element`.

That's a good point. I see in the current implementation, we use a generic `<input type="text">` with a `<button>` which is used to submit. I think the `<button>` is causing the lint violation to be thrown. I'm wondering if `<input type="slider">` would fix this, as we wouldn't need to use the button anymore.

> Would also love some help on process to bring this issue to the next office hours.

Definitely! I can help out with this :grin:

pksjce:
@TylerJDev - The `input` also contributes to the error.

I created a request for the next office hours. https://github.com/github/accessibility/issues/5101

pksjce:
Hi @ericwbailey , @ichelsea - 
This is in response to guidance on showing the resize mouse pointer on keyboard focus on the splitter component. 
Unfortunately its almost impossible to detect that the cause for focus is keyboard interaction. I'm unable to implement it as suggested. However, the mouse pointer does turn to the resize pointer when the drag is in progress as before.

src/PageLayout/PageLayout.tsx
```diff
@@ -286,17 +339,30 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
...
+              bg: isDragging || isKeyboardDrag ? 'accent.fg' : 'transparent',
               transitionDelay: '0.1s',
               '&:hover': {
-                bg: isDragging ? 'accent.fg' : 'neutral.muted',
+                bg: isDragging || isKeyboardDrag ? 'accent.fg' : 'neutral.muted',
               },
             }}
-            role="separator"
+            role="slider"
+            aria-label="Draggable splitter"
```
pksjce, line 349: @ericwbailey - Can you suggest the proper label to use here?

pksjce:
Taking this comment as confirmation to merge this.
https://github.com/github/accessibility/issues/5101#issuecomment-1824232088

> @pksjce perfect. confirming that the latest change (with the doubled-up arrow key behaviour and removal of aria-orientation) works great in all tested browser/SR combinations

PR #3910, opened 2023-11-06
Tooltip2: hide the tooltip from AT when it is not open

broccolinisoup:
Closes https://github.com/github/primer/issues/2822

#### Changed

Initially we were rendering the tooltip element visually hidden even when it is not opened but we received a feedback from the sign-off review that the tooltip should be hidden from screen readers when it is not visible for visual users too. 
In the meantime, the popover-polyfill released [a major change](https://github.com/oddbird/popover-polyfill/releases/tag/v0.3.1) that injects the styles with JS, therefore no need to tweak anything in the CSS itself. This PR;

1) Removes the visually hidden styles and leverages the default popover stylesheet styles that has `display: none` already. 
2) Removes the unnecessary CSS polyfill as they are injected via JS.

As always, shout out to @keithamus for his guidance here 🎉 
**Before**

| Before | After |
|--------|--------|
| tooltip text is in the accessibility tree when it is not opened | tooltip text is not in the accessibility tree when it is not opened |

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "@github/paste-markdown": "^1.4.0",
     "@github/relative-time-element": "^4.1.2",
     "@lit-labs/react": "^1.1.1",
-    "@oddbird/popover-polyfill": "^0.2.3",
+    "@oddbird/popover-polyfill": "^0.3.1",
     "@primer/behaviors": "^1.3.4",
     "@primer/octicons-react": "^19.8.0",
     "@primer/primitives": "^7.11.11",
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "@github/paste-markdown": "^1.4.0",
     "@github/relative-time-element": "^4.1.2",
     "@lit-labs/react": "^1.1.1",
-    "@oddbird/popover-polyfill": "^0.3.1",
+    "@oddbird/popover-polyfill": "^0.2.3",
     "@primer/behaviors": "^1.3.4",
     "@primer/octicons-react": "^19.8.0",
     "@primer/primitives": "^7.11.11",
--- a/src/drafts/Tooltip/Tooltip.tsx
+++ b/src/drafts/Tooltip/Tooltip.tsx
@@ -13,6 +13,7 @@ import {isSupported, apply} from '@oddbird/popover-polyfill/fn'
 
 // Reusable styles to use for :popover-open (Chrome, Edge) and \:popover-open (Safari, Firefox) classes
 const popoverStyles = `
+    display: block;
     padding: 0.5em 0.75em;
     width: max-content;
     height: fit-content;
@@ -32,22 +33,16 @@ const animationStyles = `
 `
 
 const StyledTooltip = styled.div`
-  /* tooltip element should be rendered visually hidden when it is not opened.  */
-  width: 1px;
-  height: 1px;
-  padding: 0;
-  margin: -1px;
-  overflow: hidden;
-  clip: rect(0, 0, 0, 0);
+  /* tooltip element shouldn't be in the DOM when it is not open  */
+  display: none;
   white-space: nowrap;
   position: fixed;
   font: normal normal 11px/1.5 ${get('fonts.normal')};
   -webkit-font-smoothing: subpixel-antialiased;
   color: ${get('colors.fg.onEmphasis')};
   text-align: center;
-
   word-wrap: break-word;
-  background: ${get('colors.neutral.emphasisPlus')}; //bg--emphasis-color
+  background: ${get('colors.neutral.emphasisPlus')};
   border-radius: ${get('radii.2')};
   border: 0;
   opacity: 0;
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -99,7 +99,7 @@
     "@github/paste-markdown": "^1.4.0",
     "@github/relative-time-element": "^4.1.2",
     "@lit-labs/react": "^1.1.1",
-    "@oddbird/popover-polyfill": "^0.3.1",
+    "@oddbird/popover-polyfill": "^0.2.3",
```
keithamus, line 102: This looks wrong.

src/drafts/Tooltip/Tooltip.tsx
```diff
@@ -13,6 +13,7 @@ import {isSupported, apply} from '@oddbird/popover-polyfill/fn'
 
 // Reusable styles to use for :popover-open (Chrome, Edge) and \:popover-open (Safari, Firefox) classes
 const popoverStyles = `
+    display: block;
```
keithamus, line 16: Perhaps removing this altogether will help

src/drafts/Tooltip/Tooltip.tsx
```diff
@@ -32,44 +20,31 @@ const animationStyles = `
...
-
-  word-wrap: break-word;
-  background: ${get('colors.neutral.emphasisPlus')}; //bg--emphasis-color
-  border-radius: ${get('radii.2')};
-  border: 0;
-  opacity: 0;
-  max-width: 250px;
-  inset: auto;
+  /* Overriding the default popover styles */
+  &[popover] {
```
broccolinisoup, line 24: This was needed for firefox. Otherwise the styles wasn't applied due to `[popover]` having more specificity than the styled component's class name.

broccolinisoup:
> I'm going to mark this as request changes because I want to keep a close eye on these changes before they ship.
> 
> I think now we're using `@oddbird/popover-pollyfill@0.3.x` - which injects stylesheets in JS - we can _remove_ a lot of the CSS and the polyfill will do the correct accessible thing.
> 
> I _think_ the correct diff for this PR would look like removing the `popoverStyles` const, and related polyfill code, and I don't think we'll need to add any additional CSS.

Thanks for the review @keithamus! I really appreciate it. I pushed another commit based on your suggestions, and everything worked ([with just a tiny adjustment](https://github.com/primer/react/pull/3910#discussion_r1384248267)) as you described 🎉

src/drafts/Tooltip/Tooltip.tsx
broccolinisoup: This trick caused a bit of a flashing though. 

https://github.com/primer/react/assets/1446503/cde9f8cf-6ded-45ce-b1a9-59dc8b77e57e
_Video description: There is a "Delete" button and the page is reloaded. The tooltip text without any styles appears right under the button very briefly like flashing then disappears_

And if I load the style component (the tooltip element itself) with `display: 'none'` and then add `display: 'block'` when it is open, the flash is gone ([commit](https://github.com/primer/react/pull/3910/commits/7552c3bc4bf80a92898ba1222f31336f7991f3e6) for changes) but I am not sure if this is a good way to solve this. Let me know if you have any a way of solving this already or any other thoughts!

keithamus (approved):
This now looks great!

:shipit:

src/drafts/Tooltip/Tooltip.tsx
keithamus: This looks reasonable for now. The polyfill injecting CSS means that we suffer from a FOUC, which is also true for dotcom today. I'm looking into ways to minimise or remove the FOUC but given it's only Firefox, and Firefox will soon be shipping native popovers, it's not a blocking concern.

siddharthkp: > And if I load the style component (the tooltip element itself) with display: 'none' and then add display: 'block' when it is open

I kinda like this as a temporary measure but also if the flash only happens in firefox, we can choose to ignore it. No strong feelings, up to you

broccolinisoup: Just to clarify, the fix that needed for firefox was adding the `[popover]` attribute selector to the class styling but this addition caused flashing in all browsers. So I guess it is better to keep it until Firefox catches up.

broccolinisoup:
@siddharthkp Could I get your approval please if it looks good? Need primer-react approval to merge 👀

siddharthkp (approved):
Ship it!

siddharthkp:
> @siddharthkp Could I get your approval please if it looks good? Need primer-react approval to merge 👀

Oh, my bad! Didn't realise that.

patrickhlauke:
:+1:

PR #3913, opened 2023-11-06
Adds inactive state to ActionList items

mperrotti:
Relates to https://github.com/github/primer/issues/2677

### Screenshots

https://github.com/primer/react/assets/2313998/80e94eeb-5b8b-42c0-aec0-cc35012bb663

Screenshot 2023-11-09 at 4 53 35 PM
Screenshot 2023-11-06 at 5 07 22 PM
Screenshot 2023-11-17 at 4 09 04 PM

### Changelog

Supports inactive ActionList items by letting users pass the required message to the `inactiveText` prop

#### New

`inactiveText` prop on ActionList.Item

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- Check out the new stories for `ActionList`, `ActionMenu`, and `NavList` that demonstrate inactive list items

### Merge checklist

- [x] Added/updated tests
- [?] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
  - Not sure. I thought none of our ActionList stuff was SSR compatible due to our use of slots
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


src/ActionList/Item.tsx
```diff
@@ -165,49 +169,55 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
         onSelect(event, afterSelect)
       },
-      [onSelect, disabled, afterSelect],
+      [onSelect, disabled, inactive, afterSelect],
     )
 
     const keyPressHandler = React.useCallback(
       (event: React.KeyboardEvent<HTMLLIElement>) => {
-        if (disabled) return
+        if (disabled || inactive) return
```
mperrotti, line 192: I'm not calling any handlers passed to `onSelect` for inactive items. This would also make it impossible to render a dialog when the user clicks the item. I think this is a reasonable limitation to begin with, and we can always update the implementation to support click/selection handlers so users can show a `Dialog` if we find we need to.

Do you agree that it's safe to prohibit `onSelect` for now?

src/ActionList/Item.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
 import styled from 'styled-components'
+import {AlertIcon} from '@primer/octicons-react'
 import Box, {BoxProps} from '../Box'
+import {Tooltip} from '../drafts'
```
mperrotti, line 5: I'm using the new, accessible Tooltip. Is this safe to use even though it's still in the experimental phase?

mperrotti, line 5: When I try and run tests locally and import the "new" Tooltip, I got this mysterious error:

```
 FAIL  src/ActionList/ActionList.test.tsx
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'LeadingVisual')

      226 | export type NavListLeadingVisualProps = ActionListLeadingVisualProps
      227 |
    > 228 | const LeadingVisual = ActionList.LeadingVisual
          |                                  ^
      229 |
      230 | LeadingVisual.displayName = 'NavList.LeadingVisual'
      231 |

      at Object.LeadingVisual (src/NavList/NavList.tsx:228:34)
      at Object.require (src/NavList/index.ts:1:1)
      at Object.require (src/drafts/index.ts:51:1)
      at Object.require (src/ActionList/Item.tsx:18:1)
      at Object.require (src/ActionList/index.ts:3:1)
      at Object.require (src/index.ts:49:1)
      at Object.require (src/utils/testing.tsx:7:1)
      at Object.require (src/utils/test-matchers.tsx:7:1)
```

Any idea what might be causing this? I couldn't figure it out.

src/ActionList/Item.tsx
mperrotti: Here's a draft PR to update the interface guidelines: https://github.com/primer/design/pull/654

I'll close it if we think we _do_ need to support click/selection behavior on inactive items.
I'll merge it if we don't think we need to support click/selection behavior on inactive items.

src/ActionList/Item.tsx
siddharthkp: cc @broccolinisoup

src/ActionList/Item.tsx
```diff
@@ -228,7 +240,37 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         >
           <ItemWrapper {...wrapperProps}>
             <Selection selected={selected} />
-            {slots.leadingVisual}
+            {
+              // If it's inactive and isn't rendered in a container like ActionMenu or SelectPanel, render a leading visual with a tooltip
+              inactive && container === undefined ? (
+                <Tooltip text={inactiveText}>
```
siddharthkp, line 246: As the cursor is on the entire item, should the tooltip also be on the whole item or just the icon?

https://github.com/primer/react/assets/1863771/0add3f8d-b5f1-418e-b1e1-d8ce6cb985b2

siddharthkp:
What is the overlap with `disabled` state?  

`<ActionList.Item disabled={true}>`: https://primer.style/react/storybook/?path=/story/components-actionlist-features--disabled-item

src/ActionList/Item.tsx
mperrotti: Just the icon

mperrotti:
> What is the overlap with disabled state?

Disabled items don't need to meet the minimum contrast and don't need a message explaining why it's disabled. cc @alexislucio

mperrotti:
> Just to confirm, I see in ActionList, we always add an alert icon. But in ActionMenu, we only add an alert icon if there is already a leading visual. Is that intentional?

ActionMenu actually doesn't need to show the alert icon at all since we show the message inline. I just kept it as another visual signal that the item doesn't work. I could easily change it so that ActionMenu items never render the alert icon. What do you think?

mperrotti:
> When the other items don't have icons, the alert icon moves the position of the label. Should we try to keep the labels aligned? For example, by using trailingVisual for alert instead of leading?

This is such a rare state that I didn't think we need to worry about the alignment as much as we do for things like [single select](https://primer.style/react/storybook/?path=/story/components-actionlist-features--single-select).

However, I like your idea about using the trailing visual slot. Let me play with it and see how it feels.

mperrotti:
> Is there a use case for disabled items with low contrast that's not covered by inactive? 🤔

If it's disabled and doesn't need to be conveyed to screen reader users. @alexislucio or somebody from a11y could probably give a more detailed answer. We have some info in the interface guidelines: https://primer.style/components/action-list#tooltips-and-dialogs-on-inactive-items

mperrotti:
I took the suggestion to render the inactive indicator in the trailing visual slot to preserve the alignment. Good call @siddharthkp !

src/ActionList/Item.tsx
broccolinisoup: In terms of using the new draft Tooltip, I think it is safe. It is unlikely to have API changes but still possible since it is in the experimental stage but it wouldn't be a big deal for us anyway because we could always update the internals if Tooltip API is changed. 

Tooltip is still going through the accessibility sign-off review but compared to the stable one it solves multiple problems and it is more accessible. I think trying out the new tooltip in Primer rather than dotcom first seems like a really good idea to me.

src/ActionList/Item.tsx
```diff
@@ -13,15 +15,39 @@ import {GroupContext} from './Group'
...
-import {LeadingVisual, TrailingVisual} from './Visuals'
+import {LeadingVisual, TrailingVisual, VisualProps} from './Visuals'
 
 const LiBox = styled.li<SxProp>(sx)
 
+const InactiveIndicator: React.FC<{
+  text: TooltipProps['text']
+  visualComponent: React.FC<React.PropsWithChildren<VisualProps>>
+}> = ({text, visualComponent: VisualComponent}) => (
+  <Tooltip text={text}>
```
broccolinisoup, line 26: Just reading through [the accessibility section on conveying inactive state to screen reader users](https://primer.style/components/action-list#tooltips-and-dialogs-on-inactive-items), and wanted to comment to make sure we are rending the right type of Tooltip here. In the new Tooltip, there are [two types](https://primer.style/components/tooltip#options): label type and description type. In general, icon buttons are label type. (The type matters because we do the internal connection between the trigger and the tooltip using `aria-labelledby` or `aria-describedby` depending on the type) 

In this usage, the button doesn't have an accessible name and it is described by the tooltip text. Is that expected? Or do we want to use the tooltip text to label the visual component?

src/ActionList/Item.tsx
broccolinisoup: Posted a comment/question about the usage as well https://github.com/primer/react/pull/3913#discussion_r1386025044 - very keen to hear what you think!

src/ActionList/Item.tsx
mperrotti: Great question. I don't think I got specific guidance from a11y about the "type" of Tooltip to use. @alexislucio - what do you think?

siddharthkp:
> If it's disabled and doesn't need to be conveyed to screen reader users. @alexislucio or somebody from a11y could probably give a more detailed answer. We have some info in the interface guidelines: https://primer.style/components/action-list#tooltips-and-dialogs-on-inactive-items

These guidelines sound like they are describing the new inactive state. 😓 I am very confused between the difference between disabled and inactive. Will wait for @alexislucio's reply.

ericwbailey:
>These guidelines sound like they are describing the new inactive state. 😓 I am very confused between the difference between disabled and inactive. Will wait for @alexislucio's reply.

Hiya! Filling in for @alexislucio while they're out. 

The thinking here is that inactive is a discrete, purposeful state separate from a more generic disabled state, and only to be used when there is a catastrophic failure mode present. By dividing these into two separate concerns, the hope is disabled state won't be accidentally misused (or vice-versa). 

The guidance around how that affects visuals and compliance speaks to a couple of specific concerns:

- Controls in an inactive state may still have functionality, just not the intended functionality (i.e. showing a tooltip or dialog). Because of that, they are not disabled per se, just repurposed to explain why they don't work.
  - This is a weird, niche situation, but in terms of WCAG disabled controls do not need to pass color contrast requirements. However, controls that do something do need to.
- Additionally, while we should provide a global site banner to indicate the outage state, we want to ensure the inactive controls are perceivable for low vision users. They may be viewing the screen through magnification, and may miss the global site banner area.
  - In this situation, there would be confusion about why expected controls have "disappeared" from view, adding stress to an already non-ideal situation. 

The specifics of the implementation (what is clicked/hovered/focused) are built off of this, but we're trying to thread a very specific and niche needle here of (hopefully) rare application state and WCAG nitty-gritty 😅

src/ActionList/Item.tsx
ericwbailey: Hey hi, filling in for @alexislucio while they're away. I think we'd want `aria-describedby`. My thinking is with the current code:

1. The `button` that wraps the `svg` icon uses an `aria-describedby` to point to the tooltip's `id`.
2. The `svg` does not have an accessibile name, as it has `aria-hidden="true"` applied to it.
3. The tooltip's content would read something like, "Not available due to an outage"

This leaves the `button` without an accessible name. What we can do about this is point to the ActionList item's title (ex: "Octicons") and use `aria-labelledby` to provide the accessible name for the tooltip button (ex: `<button aria-labelledby="actionlist-item-title-octoicons">`). 

We can then use `aria-describedby` to supplement the announcement, so it could read something along the lines of:

```text
Octicons. Button. Not available due to an outage.
```

mperrotti: Thanks @ericwbailey ! I updated this PR to follow that guidance.

siddharthkp:
@ericwbailey Couple of follow up questions

> The thinking here is that inactive is a discrete, purposeful state separate from a more generic disabled state, and only to be used when there is a catastrophic failure mode present. By dividing these into two separate concerns, the hope is disabled state won't be accidentally misused (or vice-versa).

I think I understand the intent, can you talk a little more about the accidental misuse we would want to avoid?


The second question is for the generic disabled state, right now, when an item is marked as disabled, there is no way to focus on it with a keyboard, which means it's not possible to communicate why an item is disabled.

Should we change how we render disabled items so that they are part of the accessible tree? and what that cause the behavior of inactive and disabled items to be similar?

An example from [slack](https://github.slack.com/archives/C01L618AEP9/p1699612423528549),

> Hey team :meow-wave: I'm looking to use disabled items in a SelectPanel, but noticed that the current implementation adds tabindex="-1", which makes the item inaccessible and makes the description not readable. Is this issue already in your sights?
For context, [the scenario is for our Project picker in Issues to show when a memex project has hit the item limit](https://github.com/github/issues/issues/7112#issuecomment-1801958989) and can have no more items linked to it.

SelectPanel with an item that is disabled, it has red text that explains the reason for disabling the item but is not accessible to screen readers

ericwbailey:
>Should we change how we render disabled items so that they are part of the accessible tree? 

I think so, yes. My mental model here would be applying `aria-disabled` to menu items, the way that the toolbar on macOS/Windows can get conditionally disabled. So, like, disabling the option to past in the Edit menu if there is nothing on the clipboard. 

The reasoning is `aria-disabled` still allows the menu item to be discoverable, and announced as disabled. This lets someone know it is possible to use, just not in the current application state.

>and what that cause the behavior of inactive and disabled items to be similar? (we can still choose to have different names to avoid misuse)

I think this is a letter-of-the-law WCAG compliance thing, but the reason we're not disabling the controls in this situation is that they're still doing something, just not what the user might normally expect. If there were no tooltips, dialogs, or other content that appeared as a result of user action it might be a candidate for disabling.

mperrotti:
@siddharthkp or @broccolinisoup - I could use some help figuring out why tests fail to run when I import the new `Tooltip` from `drafts`

broccolinisoup:
@mperrotti I had a look at it last week couldn't figure out but will look at it again today no worries! I'll make sure to unblock you

broccolinisoup:
@mperrotti If you could import the Tooltip explicitly from the source folder rather than through draft, the tests should pass.

```
import {Tooltip, TooltipProps} from '../drafts/Tooltip/Tooltip'
```

I am confused why just the draft import doesn't work though and I'll look into it. In the meantime, I hope this unblocks you!

src/drafts/Tooltip/Tooltip.tsx
```diff
@@ -273,9 +273,9 @@ export const Tooltip = React.forwardRef(
             React.cloneElement(child as React.ReactElement<TriggerPropsType>, {
               ref: triggerRef,
               // If it is a type description, we use tooltip to describe the trigger
-              'aria-describedby': type === 'description' ? `tooltip-${tooltipId}` : undefined,
+              'aria-describedby': type === 'label' ? `tooltip-${tooltipId}` : child.props['aria-describedby'],
```
broccolinisoup, line 276: I think this PR provided a great opportunity to validate some of the implementation details! Thanks again for using Tooltip here 🙌🏻

I am not sure if updating the string `description` to `label` was intentional here but it unfortunately introduces an issue to the Tooltip. We need to make sure we render 'aria-describedby' for the description type tooltips and this line does that. So I think if we could update this line as 

```suggestion
              'aria-describedby': type === 'description' ? `tooltip-${tooltipId}` : child.props['aria-describedby'],
```

it will add `aria-describedby` attribute to the child element for description type, or use the child's existing `aria-describedby` attribute if there is one and not description type tooltip

With this update, current inactive story suffice the criteria Eric described [here](https://github.com/primer/react/pull/3913#discussion_r1388600092). Let me know if I am missing anything

broccolinisoup (changes requested):
Left a comment about Tooltip! Looking forward to hearing your thoughts

src/drafts/Tooltip/Tooltip.tsx
mperrotti: Oops - not intentional, just a copy/paste mistake. Your suggestion is what I meant 😅

src/ActionList/Item.tsx
```diff
@@ -228,12 +265,29 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         >
           <ItemWrapper {...wrapperProps}>
             <Selection selected={selected} />
-            {slots.leadingVisual}
+            {
```
mperrotti, line 268: By the way, I cleaned up all this logic while I was working on the [loading state](https://github.com/github/primer/issues/2681). I put it in a component, made the logic a lot easier to follow, and was able to get rid of the non-null assertion.

If you'd like me to introduce that component in this PR instead of the loading state PR (not yet created), I'm ok with that.

src/ActionList/Item.tsx
```diff
@@ -251,8 +305,39 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
                   </Box>
                   {slots.inlineDescription}
                 </ConditionalBox>
-                {slots.trailingVisual}
+                {
```
mperrotti, line 308: I was able to clean up this logic. See https://github.com/primer/react/pull/3913/files#r1409960232

pksjce:
Hi @mperrotti - In the examples storybook (everything example), the ones with inactiveText don't seem to show any tooltip. Do you think this is a bug? Also the focus outline on the alert icon has an 8 px extra margin on one side. Is that something we want to fix?

mperrotti:
Thanks for flagging that @pksjce 

This only happens in that "All Combinations" story, so I did some investigating and found a bug with the new `Tooltip`. The positioning values are getting calculated incorrectly when the trigger is rendered outside the viewport boundaries.

https://github.com/primer/react/assets/2313998/de0cffeb-430a-420c-b626-9f483c556a04

Screenshot 2023-12-07 at 1 46 13 PM

cc @broccolinisoup

src/ActionList/shared.ts
```diff
@@ -52,18 +58,21 @@ type MenuItemProps = {
   'aria-labelledby'?: string
   'aria-describedby'?: string
   role?: string
+  'data-inactive'?: string
```
pksjce, line 61: I don't think this should be a prop. We set this based on `inactiveText` value right?

src/ActionMenu/ActionMenu.features.stories.tsx
```diff
@@ -122,3 +122,115 @@ export const MultiSelect = () => {
...
+              <BookIcon />
+            </ActionList.LeadingVisual>
+          </ActionList.LinkItem>
+        </ActionList.Group>
+      </ActionList>
+    </ActionMenu.Overlay>
+  </ActionMenu>
+)
+
+// TODO: Uncomment this story when we have inactive buttons
```
pksjce, line 181: Since inactive buttons have been merged this can be uncommented?

pksjce (approved):
Thanks for the prompt replies @mperrotti!
Left a couple of comments mainly
1. `data-inactive` does not need to be a prop

Would be great to see the following in a follow-up PR
- Uncomment the story with the inactiveButton
- Add more comprehensive documentation for `inactiveText` ie
   - Its shown as tooltip on icon in `ActionList` and as description in `ActionMenu`
   - It replaces LeadingVisual if one is already present and adds/replaces TrailingVisual if not. I feel this will be a bit of a surprise to consumers otherwise.
   - In case of a link item, it behaves in a completely different manner.
   
Also look forward to the `Tooltip` related future bugfix that is independent of this PR.

Other than that this looks good to me. I will also close out the associated discussion with a few comments for posterity. Thanks for all your efforts!

mperrotti:
> In case of a link item, it behaves in a completely different manner.

I'm not sure what you mean about the behavior being completely different on link items.

broccolinisoup (approved):
Tooltip usage is looking good! ✨

PR #3922, opened 2023-11-08
docs(contributors): add ADR 018

joshblack:
Closes https://github.com/github/primer/issues/2823

Add ADR 018 which supercedes ADR 009 around interaction testing. We no longer use Chromatic and instead prefer Playwright for authoring interaction tests.

Let me know if this is a good flow to use when updating an ADR! I wasn't sure what the exact flow was so I ended up just adding a new ADR with this decision.

Also feel free to provide any feedback if this approach is incorrect or merits a larger discussion!

### DIFF
--- a/contributor-docs/adrs/adr-009-interaction-tests.md
+++ b/contributor-docs/adrs/adr-009-interaction-tests.md
@@ -1,5 +1,8 @@
 # ADR 009: Use Interaction testing in storybook to maintain components
 
+> **Warning**
+> This ADR is superceded by [`ADR 018`](./adr-018-interaction-tests-revisited.md)
+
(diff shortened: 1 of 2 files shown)


broccolinisoup (approved):
This looks great! 

>Let me know if this is a good flow to use when updating an ADR! I wasn't sure what the exact flow was so I ended up just adding a new ADR with this decision.

To my understanding, ADRs are records of decisions that were made at a certain time so I imagine we want to keep all the records historically so I think creating new one is perfect.

siddharthkp (approved):
Good catch!

pksjce (approved):

Issue #3486, opened 2023-07-05
ActionMenu overflow when there are too many ActionList items


lowply:
## Summary

As shown in the following screen recording, ActionMenu overflows when there are too many ActionList items and blocks users from selecting items in the hidden part of the list.

https://github.com/primer/react/assets/18643/b9dc539d-5afe-43d5-bcb3-2ac334f43425

In the regular HTML `select` element this doesn't happen because the menu renders over the browser window. ActionMenu is, OTOH, a React component which cannot be rendered over the browser window. To mitigate this issue, there should be a scroll bar on the right side of the list so users can view and select all items regardless of the number of the items in the list.

This was originally reported in https://github.com/github/helphub/issues/4215.

### Steps to reproduce

See https://ghe.io/lowply/action-menu-example/blob/main/src/ActionMenuComponent.tsx

### Version

v35.26.0

### Browser

Safari

amareshvarma:
I would like to work upon this issue

amareshvarma:
@lowply The steps to reproduce is redirecting to github enterprise, can we have to available with out redirecting to github enterprise

siddharthkp:
Hi @lowply, How do you feel about using [SelectPanel](primer.style/react/SelectPanel) instead? It has better UX (like search) for long lists

lowply:
@amareshvarma Can you see this? https://lowply-org.github.io/action-menu/

@siddharthkp Thanks for the idea! TIL about SelectPanel and I like its scroll UI and filtering functionality. However in our use-case we can't have the selectable items - we need out users to jump to a certain URL when they click an item in the pull-down menu.

lesliecdubs:
@lowply can you share any more information about how you're using the `ActionMenu`, and what you would expect for this experience to be for a user utilizing assistive technology like a screenreader? 

If it's a long list of items, we were wondering how you were thinking about making sure the UX of selecting an item was not too much trouble.

lowply:
@lesliecdubs Hi 👋🏻 Please see https://github.com/github/helphub/issues/4215 for more context about this issue. Feel free to ping me on Slack if you have any other questions!

mperrotti:
I'm wondering if a [SelectPanel](https://primer.style/design/components/selectpanel) would make more sense for this use-case. The filtering behavior would make it easier to select from a long list of options.

[...]

PR #3926, opened 2023-11-09
Support overflow scrolling in ActionMenu

strackoverflow:
Closes https://github.com/primer/react/issues/3486 

When looking into https://github.com/primer/react/issues/3486, it occurred to me that while we support passing a `maxHeight` prop, we don't enable overflow scrolling when that prop is provided.

This PR adds an optional `overflow` prop to `Overlay` which allows developers to specify the overflow behavior of the menu. 

I think ideally we should automatically apply `overflow: auto` whenever a max height is set (or even for all ActionMenu instances) but at least for now this can unblock this use case without impacting existing usage of Overlay.

| Demo |
|--------|
| scrolling menu | 

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


strackoverflow:
Closing in favor of https://github.com/primer/view_components/pull/2383

strackoverflow:
Re-opening, I meant to close a different PR 😄

siddharthkp:
> I think ideally we should automatically apply overflow: auto whenever a max height is set (or even for all ActionMenu instances)

Same! Are there any risks in doing that? (for example: menus that suddenly have an unexpected scroll?)

siddharthkp (approved):
Looks good!

strackoverflow:
> Are there any risks in doing that? (for example: Overlays that suddenly have an unexpected scroll?)

@siddharthkp this is the exact thing I want to be cautious about. I agree, we can merge this basic step for now to unblock anyone who needs to enable scrolling but we can revisit to make the DX a little better. Thanks!

src/Overlay/Overlay.tsx
```diff
@@ -64,7 +65,7 @@ const StyledOverlay = styled.div<StyledOverlayProps>`
   max-height: ${props => props.maxHeight && heightMap[props.maxHeight]};
   width: ${props => widthMap[props.width || 'auto']};
   border-radius: 12px;
-  overflow: hidden;
+  ${props => (props.overflow ? `overflow: ${props.overflow};` : '')}
```
strackoverflow, line 68: I just realized we should probably retain the existing overflow behavior if the prop isn't provided so that we don't cause any unexpected side effects. I'll push up a fix for that.

strackoverflow:
@siddharthkp 👋 would you mind taking another look at this PR? I was about to merge it when [I realized](https://github.com/primer/react/pull/3926#discussion_r1424466530) that this changed the default overflow behavior of Overlay, which was not intentional.

Let me know if the current diff looks good to merge now!

broccolinisoup (approved):
Looks great! ✨

PR #3929, opened 2023-11-10
SelectPanel2: Types!

siddharthkp:
- Add types for experimental SelectPanel
- Removed a bunch of ts-ignores :)


src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -23,16 +24,16 @@ import {useFocusZone} from '../../hooks/useFocusZone'
...
 const SelectPanelContext = React.createContext<{
   title: string
-  description: string
+  description?: string
   panelId: string
   onCancel: () => void
   onClearSelection: undefined | (() => void)
   searchQuery: string
-  setSearchQuery: () => void
+  setSearchQuery: React.Dispatch<React.SetStateAction<string>>
```
broccolinisoup, line 32: Are we defining its type here? 🤔  As far as I understand, we set the default values for the context here but this seems like doing something else or do I miss something?

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -41,50 +42,87 @@ const SelectPanelContext = React.createContext<{
...
+  title: string
+  description?: string
+  selectionVariant?: ActionListProps['selectionVariant'] | 'instant'
+  id?: string
+
+  defaultOpen?: boolean
+  open?: boolean
+  anchorRef?: React.RefObject<HTMLButtonElement>
+
+  onCancel?: () => void
```
broccolinisoup, line 55: Do you think we need to type the event argument here? Something like this or not needed?

```
event: React.MouseEvent<HTMLButtonElement> | React.KeyboardEvent<HTMLButtonElement>
```

broccolinisoup (commented):
Just left some questions!

Feel free to get another review, I am still trying to get my head around all the type stuff 😄

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: There are multiple elements and events to cancel: click/press cancel if it's visible, click/press the cross icon (always present), click outside. So, I'm making `onCancel` a callback for all of these without passing event. 

I don't think there is a use case yet to access the event or prevent it, let's see if we get feedback based on usage around this!

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Yes, we can do both here.

`React.createContext<ContextTypes>(defaultValue)`

siddharthkp:
Draft → Ready for review!

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -331,11 +361,10 @@ const SelectPanelFooter = ({...props}) => {
     </Box>
   )
 }
-SelectPanel.Footer = SelectPanelFooter
 
-// @ts-ignore todo
-SelectPanel.SecondaryButton = props => {
-  return <Button {...props} size="small" type="button" block />
+// TODO: is this the right way to add button props?
+const SelectPanelSecondaryButton: React.FC<ButtonProps> = props => {
```
joshblack, line 366: Honestly this might be fine here (and is all React.FC is doing anyways) 🤷 

```suggestion
const SelectPanelSecondaryButton = (props: ButtonProps) => {
```

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -243,20 +277,16 @@ const SelectPanelHeader: React.FC<React.PropsWithChildren> = ({children, ...prop
     </Box>
   )
 }
-SelectPanel.Header = SelectPanelHeader
 
-// @ts-ignore todo
-const SelectPanelSearchInput = props => {
+const SelectPanelSearchInput: React.FC<TextInputProps> = ({onChange: propsOnChange, ...props}) => {
```
joshblack, line 281: For `onChange`, do they need access to the event itself or could we provide the value of the input instead? (Or a similar `onValueChange`). This might help to simplify the types if `onChange` (or `onValueChange`) is something like `onChange({ value }: { value: string }): void;`

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -142,7 +179,7 @@ const SelectPanel = props => {
             }}
           >
             {/* render default header as fallback */}
-            {slots.header || <SelectPanel.Header />}
+            {slots.header || <SelectPanelHeader />}
```
joshblack, line 182: I think of these as equivalent so really not meant as a suggestion to change, just is always fun to use new features when an opportunity comes up 😂 

```suggestion
            {slots.header ?? <SelectPanelHeader />}
```

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -41,50 +43,87 @@ const SelectPanelContext = React.createContext<{
...
       return null
     }
     return child
   })
 
-  const [internalOpen, setInternalOpen] = React.useState(props.defaultOpen)
+  const [internalOpen, setInternalOpen] = React.useState(defaultOpen)
   // sync open state
-  React.useEffect(() => setInternalOpen(props.open), [props.open])
+  React.useEffect(() => setInternalOpen(propsOpen || false), [propsOpen])
```
joshblack, line 101: It could be worth keeping track of the previous value in state instead of in an effect, I feel like this is similar to the `getDerivedStateFromProps` example from the hooks FAQ: https://legacy.reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops 

Curious if you agree or if there is a difference here 👀

src/drafts/SelectPanel2/SelectPanel.tsx
joshblack: Broadly speaking, how would you two feel about borrowing the `createContext` pattern from radix to help out with context usage like this across PRC? https://github.com/radix-ui/primitives/blob/main/packages/react/context/src/createContext.tsx#L3

I think the tl;dr is that this pattern has the context types itself be nullable (e.g. `ContextValueType | null`) but they leverage a hook to enforce that the value is always defined so they don't have to deal with default values.

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -41,50 +43,87 @@ const SelectPanelContext = React.createContext<{
...
+  onSubmit?: (event?: React.FormEvent<HTMLFormElement>) => void
+
+  // TODO: move these to SelectPanel.Overlay or overlayProps
+  width?: AnchoredOverlayProps['width']
+  height?: AnchoredOverlayProps['height']
+
+  children: React.ReactNode
+}
+
+const Panel: React.FC<SelectPanelProps> = ({
```
joshblack, line 67: Super random question, is there a convention that is used to decide between `React.FC<Props>` versus annotating the props directly e.g. `const Component = (props: Props) => ...`?

joshblack (commented):
Just left a couple comments/questions! Cool to see this all come together 🥳

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Not yet, I don't have strong opinions on this, happy to go either way.

Do we want to avoid `React.FC`?

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: >  the context types itself be nullable (e.g. ContextValueType | null) but they leverage a hook to enforce that the value is always defined so they don't have to deal with default values.

That's cool! Can I pass on this PR and try it in a follow up PR?

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Good call! Updated :)

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Will sync with what we decide for https://github.com/primer/react/pull/3929/files#r1407598117

joshblack (approved):

PR #3936, opened 2023-11-13
Wrap generating e2e script into async func

broccolinisoup:
#### Changed

Running the `script/generate-e2e-tests.js` throws an error due to prettier.format retuning a Promise object, not string. ([docs reference](https://prettier.io/docs/en/api.html#prettierformatsource-options))

```
node:internal/fs/utils:905
  throw new ERR_INVALID_ARG_TYPE(
  ^

TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of Promise
    at Object.writeFileSync (node:fs:2314:5)
    at Object.<anonymous> (/Users/broccolinisoup/GitHub/DesignSystems/react/script/generate-e2e-tests.js:1605:6)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12)
    at node:internal/main/run_main_module:23:47 {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

I wrapped the code into an async function and awaiting the prettier.format. Let me know if there is a better way to do it!

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Could be test adding a new item to the `components` array with an id and story name to check if the script generates a test file. 

### DIFF
--- a/script/generate-e2e-tests.js
+++ b/script/generate-e2e-tests.js
@@ -1521,15 +1521,17 @@ const components = new Map([
   ],
 ])
 
-for (const [component, info] of components) {
-  const filepath = path.join(E2E_DIR, path.format({name: `${component}.test`, ext: '.ts'}))
+async function generateE2ETests() {
+  try {
+    for (const [component, info] of components) {
+      const filepath = path.join(E2E_DIR, path.format({name: `${component}.test`, ext: '.ts'}))
 
-  if (fs.existsSync(filepath)) {
-    continue
-  }
+      if (fs.existsSync(filepath)) {
+        continue
+      }
 
-  const stories = info.stories.map(story => {
-    return `test.describe('${story.name}', () => {
+      const stories = info.stories.map(story => {
+        return `test.describe('${story.name}', () => {
   for (const theme of themes) {
     test.describe(theme, () => {
       test('default @vrt', async ({page}) => {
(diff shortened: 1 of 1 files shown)


joshblack:
@broccolinisoup just as another option, you can also replace the import to `@prettier/sync` and the API will be the same. We had to do this for the codemods so the dependency should already be installed too 👍

script/generate-e2e-tests.js
```diff
@@ -1556,20 +1558,28 @@ for (const [component, info] of components) {
...
+      })
+      fs.writeFileSync(filepath, formatted, 'utf8')
+    }
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.log(error)
+  }
 }
+
+generateE2ETests()
```
joshblack, line 1585: One suggestion, instead of the try-catch block we could instead add `.catch()` to the promise call and log/exit from there.

```suggestion
generateE2ETests()
  .catch((error) => {
    // eslint-disable-next-line no-console
    console.log(error)
    process.exit(1)
  });
```

PR #3945, opened 2023-11-15
ActionList.Item: Bug fix to support both inline and block description at the same time

siddharthkp:
Thanks to ton to @joshblack for his help with the types!! ✨

Found a regression that was introduced in [6 months ago](https://github.com/primer/react/pull/3173), but hasn't been reported yet :)

To fix that, added a backward compatible "test condition" to slots config, so that the component can be in two different based on props. Only ActionList.Item has this use case with inline and block description.

Example:
Reviewers can have both inline and block description

```diff
const [slots, childrenWithoutSlots] = useSlots(props.children, {
      leadingVisual: LeadingVisual,
      trailingVisual: TrailingVisual,
-     description: Description,
+     blockDescription: [Description, props => props.variant === 'block'],
+     inlineDescription: [Description, props => props.variant !== 'block'],
    })
```

<details>

<summary>Show code for preview</summary>

```jsx
<ActionList.Item>
  Only I
  <ActionList.Description>inline description</ActionList.Description>
</ActionList.Item>
<ActionList.Item>
  Only B
  <ActionList.Description variant="block">block description</ActionList.Description>
</ActionList.Item>
<ActionList.Item>
  I + B
  <ActionList.Description variant="inline">inline description</ActionList.Description>
  <ActionList.Description variant="block">block description</ActionList.Description>
</ActionList.Item>
<ActionList.Item>
  The everything bagel
  <ActionList.LeadingVisual><StarIcon/></ActionList.LeadingVisual>
  <ActionList.Description variant="inline">inline description</ActionList.Description>
  <ActionList.Description variant="block">block description</ActionList.Description>
  <ActionList.TrailingVisual><StarIcon/></ActionList.TrailingVisual>
</ActionList.Item>

```
</details>

| Before | After |
|--------|--------|
| before: only one of inline or block description is visible | after: both inline and block description are visible | 

See also conversation from slack: https://github.slack.com/archives/C01L618AEP9/p1700085056205969


langermank:
Hey @siddharthkp! Before I respond to the margin question, can you clarify how this is a regression? Its my understanding that its one or the other: inline or block. They shouldn't co-exist.. is there a use case I'm not aware of?

siddharthkp:
> Hey @siddharthkp! Before I respond to the margin question, can you clarify how this is a regression? Its my understanding that its one or the other: inline or block. They shouldn't co-exist.. is there a use case I'm not aware of?

When Vini did the initial inventory, we saw a few use cases for it, example in the reviewer selection:

Reviewers can have both inline and block description

.playwright/snapshots/components/ActionList.test.ts-snapshots/ActionList-Inline-Description-dark-colorblind-linux.png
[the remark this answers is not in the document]
langermank: Thanks for tagging me! I think the change you made is correct, mostly just by looking at the snapshots. 🚀

langermank (approved):

src/ActionList/Item.tsx
```diff
@@ -36,8 +36,10 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     const [slots, childrenWithoutSlots] = useSlots(props.children, {
       leadingVisual: LeadingVisual,
       trailingVisual: TrailingVisual,
-      description: Description,
+      blockDescription: [Description, props => props.variant === 'block'],
+      inlineDescription: [Description, props => props.variant !== 'block'],
```
pksjce, line 40: Something I don't understand is why not have two different slots for block and inline description? Is it to avoid breaking changes?

siddharthkp: Yep, that would be a breaking change. We used to support it before the regression.

But also, this feels like an implementation detail, I don't personally like the idea of it leaking out into the API with `ActionList.InlineDescription` and `ActionList.BlockDescription` 😅

pksjce (approved):

PR #3946, opened 2023-11-15
Link: Add inline prop to tag links inside a text block

siddharthkp:
From [A more readable GitHub](https://thehub.github.com/news/2023-10-18-a-more-readable-github/#what-you--need-to-know),

> Links embedded in text are most perceivable when they have distinct styling. Sufficient contrast with the surrounding text is good, but should not be used as the only method of visually distinguishing a link from other text content.

> To remediate the problem we have implemented a new [accessibility setting](https://github.com/settings/accessibility) with two options: show or hide link underlines in text blocks. Those who need or prefer stylistic differentiation can now choose to enable them. The default is to underline links.

This PR implements the short term strategy from https://github.com/github/primer/discussions/2836#discussioncomment-7533878

Note: inline=true overrides underline=false and adds an underline

Story with all combinations: https://primer-091e4ade04-13348165.drafts.github.io/storybook/?path=/story/components-link-features--inline


src/Link/Link.features.stories.tsx
```diff
@@ -19,3 +19,77 @@ export const Underline = () => (
     Link
   </Link>
 )
+
+export const Inline = () => (
```
joshblack, line 23: This seems like a good candidate for a dev story, similar to what we have for button (I think?). What would you think of having the example in the feature here be a link in a text block and this would be its own dev story?

joshblack (approved):
Looks great! Just had a comment for the story and also wanted to ask if this should be added to the `Link.docs.json` as well for the site?

src/Link/Link.features.stories.tsx
siddharthkp: Good idea, done!

Issue #3947, opened 2023-11-16
ActionMenu: Default behaviour wraps titles instead of using the available space


maximedegreve:
### Description

An issue has arisen with the ActionMenu, where the titles in the menu are wrapping even when there's sufficient room to display them without wrapping. You can see the details in this internal [issue #9265](https://github.com/github/repos/issues/9265).

![Menu in commits](https://github.com/primer/react/assets/980622/1dd0f16b-8fd0-4c00-a548-effcc8eb4e88)

Menus without a set fixed width should expand automatically, unless there's no space left on the screen. It's better to avoid setting a fixed width, as menu item labels can change based on their state, which might result in odd, unused space at the end of titles.

### Steps to reproduce

1. Go to [this feature internally behind a feature flag](https://github.com/github/github/commits/master/)
2. Click the 3 dot button in one of the items
3. The first item is wrapped

### Version

v36.2.0

### Browser

Safari, Chrome

siddharthkp:
Looking at the code, ActionMenu does not grow beyond `min-width: 192px` on it's own. 

You can make it grow either 
1. add width prop to `<ActionMenu.Overlay width="medium">`. [One of | 'small' | 'medium' | 'large' | 'xlarge' | 'xxlarge' | 'auto'](https://primer.style/react/Overlay#:~:text=Sets%20the%20width%20of%20the%20Overlay)
2. Add `white-space: nowrap` to the Item: `<ActionList.Item sx={{'white-space': 'nowrap'}}>` which will force it grow beyond the min-width. 

I'm wondering if we should always add `nowrap` on the label. 🤔 There are cases where you might want to truncate it though. @langermank run into that last week, i think!

maximedegreve:
@siddharthkp, I've explored various menu implementations, like the one found at [here](https://www.radix-ui.com/primitives/docs/components/dropdown-menu), and noticed they all use an automatic width setting. I'm not sure why there would be a need for manual width adjustment, but perhaps I'm missing something.

siddharthkp:
> I'm not sure why there would be a need for manual width adjustment, but perhaps I'm missing something

I don’t think its intentional 😅

lesliecdubs:
Assigning @siddharthkp and moving to in progress since there's a draft PR in progress. Sid, if the draft doesn't pan out please move this issue back to the teams backlog "Inbox" for further discussion.

PR #3959, opened 2023-11-20
[Do not merge yet] Portal: Make portals 100% width to support children with width:auto

siddharthkp:
- Fixes https://github.com/primer/react/issues/3947
- Includes https://github.com/primer/react/pull/3958

> [!WARNING]
> **Possible breaking changes?**
> 
> This PR affects Portals, so all overlays with width:auto would now be able to grow like they should have. It's hard to know which instances will this PR breaks.
>
> Components that use Portal: 
> - Overlay (+ AnchoredOverlay, ActionMenu, Autocomplete, InlineAutocomplete, ~[LabelGroup](https://github.com/primer/react/blob/main/src/LabelGroup/LabelGroup.tsx#L112) (not used in dotcom)~)
> - ~Dialog v2 (+ ConfirmationDialog). (dialog never uses width:auto)~
> 
> Side note: There are multiple instances where [width has been set to small or medium](https://musical-adventure-wlr3n3k.pages.github.io/?name=actionmenu.overlay&attribute=width) that can be cleaned up now in favor of auto

&nbsp;

The issue here was that the `Portal` we created by default did not have any width which meant floating components like Overlay don't grow even with `width:auto`

| Before | After |
|--------|--------|
| long text wraps in 2 rows | long text making the menu wider |


siddharthkp:
Now that we have feature flags, this can be staff shipped! 🎉

PR #3962, opened 2023-11-21
SelectPanel2: Clean up stories

siddharthkp:
- Built on top of https://github.com/primer/react/pull/3929
- Divided stories between features and examples
- Set up playground


src/drafts/SelectPanel2/stories/SelectPanel.default.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  const itemsToShow = query ? filteredLabels : data.labels.sort(sortingFn)
+
+  return (
+    <>
+      <SelectPanel
+        title="Select labels"
+        onSubmit={onSubmit}
+        onCancel={() => {
```
broccolinisoup, line 67: Is this a good place to utilize storybook actions or is this story generic enough that we don't need actions?

I don't think I'm utilizing them as much, Josh brought it to my attention in one of the code reviews and I am curious what you think about them as well?

broccolinisoup (approved):
The stories are all looking good to me overall. Just left a question. Let me know if there is any specific feedback you are after in the stories!

src/drafts/SelectPanel2/stories/SelectPanel.default.stories.tsx
siddharthkp: > I don't think I'm utilizing them as much

Same! I rarely think of them 😅 

I've added them to the playground now. Let's see if they are useful

PR #3967, opened 2023-11-23
Set initial sort order of DataTable based on initialSortColumn and initialSortDirection props

### LINKED ISSUE #3951: Alphanumeric sorting initial sort
### Description

We're using the DataTable component and have a column that is using the 'alphanumeric' sort. Clicking the column header sorts the table correctly. Howver, on load, we have set this column as the `initialSortColumn` and the table is not initially sorted using the 'alphanumeric' sort, but using the 'basic' algo. 

### Steps to reproduce

1. Implement a new DataTable component. Adding columns, with one column using the 'alphanumeric' sort
2. Have data that has integers in the values, like var1, var11, var103, var2, var22, var203 etc.
3. On the DataTable component, set the `initialSortColumn` to the column that has the 'alphanumeric' sort set
4. On load the table will be sorted using a 'basic' algo. Clicking on the table column header will make it sort using the alphanumeric algo.

### Version

latest

### Browser

_No response_

joshfarrant:
Closes #3951

### Note

I spotted the above issue and thought I'd take a look at the implementation. This is my first contribution to this repo and I hope I'm not treading on any toes by picking this up and raising a PR! I was keen to have a look at how the design system was structured, as well as how the tooling worked, and I thought picking up an issue would be a good way to dig into that. Again, hope you don't mind this PR out of the blue! 🙂

### Changelog

This PR modifies the behaviour of the `initialSortColumn` and `initialSortDirection` props of the DataTable component. Previously, I believe they were only used to _describe_ the sorting of the intial data provided by the implementor. Now, they can be used to _set_ how the initial data should be sorted.

Eg, if the DataTable is rendered as follows:

```tsx
<DataTable
  data={[
    { id: 1, fieldOne: 'a', fieldTwo: 'c' },
    { id: 2, fieldOne: 'b', fieldTwo: 'b' },
    { id: 3, fieldOne: 'c', fieldTwo: 'a' },
  ]}
  columns={[
    {
      header: 'Field One',
      field: 'fieldOne',
      sortBy: true
    },
    {
      header: 'Field Two',
      field: 'fieldTwo'
    },
  ]}
  initialSortColumn="fieldOne"
  initialSortDirection="DESC"
/>
```

Then the rendered table will now be sorted like this:

| Field One ↓ | Field Two |
| --- | --- |
| c | a |
| b | b |
| a | c |

Note that the sorting of the resulting table is different to the sorting of the provided data, as the sorting is now defined by the `initialSortColumn` and `initialSortDirection` props. Previously, the table would be sorted in the same way as the initial data regardless of what `initialSortColumn` and `initialSortDirection` were set to.

Since I've modified the implementation and removed the manual tracking of `rowOrder` and `prevData` in favour of recalculating the sorted data any time the `data`, `columns`, and `sortByColumns` props change, I'd recommend that the implementor memoizes `columns` and `data` before passing them into the DataTable component. This could be seen as best practise anyway, but it might be worth documenting it explicitly. How/where would you recommend documenting this?

This may not be the direction you want to go with these props (in which case let me know what the desired behaviour should be and I'll modify the PR appropriately) however this seemed to match the desired functionality described in #3951

#### Changed

- `initialSortColumn` and `initialSortDirection` props now set the initial sort state of the DataTable component.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Modify the `initialSortColumn` and `initialSortDirection` props in the DataTable component's _Playground_ story and hit the remount button in the top left. Note that when the table mounts, the sort order should match the specified column and direction.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

⬆️ This isn't something I have access to.

### DIFF
--- a/.changeset/eleven-cooks-wonder.md
+++ b/.changeset/eleven-cooks-wonder.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Set initial sort order of DataTable based on `initialSortColumn` and `initialSortDirection` props
--- a/src/DataTable/DataTable.stories.tsx
+++ b/src/DataTable/DataTable.stories.tsx
@@ -215,6 +215,7 @@ export const Playground = (args: DataTableProps<UniqueRow> & ColWidthArgTypes) =
             minWidth: args.minColWidth0,
             maxWidth: args.maxColWidth0,
             align,
+            sortBy: 'alphanumeric',
           },
           {
             header: 'Type',
@@ -237,6 +238,7 @@ export const Playground = (args: DataTableProps<UniqueRow> & ColWidthArgTypes) =
             minWidth: args.minColWidth2,
             maxWidth: args.maxColWidth2,
             align,
+            sortBy: 'datetime',
           },
           {
             header: 'Dependabot',
@@ -323,6 +325,20 @@ Playground.argTypes = {
     type: {
       name: 'enum',
       value: ['condensed', 'normal', 'spacious'],
+    },
+  },
+  initialSortColumn: {
+    control: {
+      type: 'text',
+    },
+  },
+  initialSortDirection: {
+    control: {
+      type: 'radio',
+    },
+    type: {
+      name: 'enum',
+      value: ['ASC', 'DESC'],
     },
   },
   ...getColumnWidthArgTypes(5),
--- a/src/DataTable/DataTable.tsx
+++ b/src/DataTable/DataTable.tsx
@@ -39,9 +39,7 @@ export type DataTableProps<Data extends UniqueRow> = {
   columns: Array<Column<Data>>
 
   /**
-   * Provide the id or field of the column by which the table is sorted. When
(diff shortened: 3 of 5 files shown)


joshblack:
Hey @joshfarrant! 👋 Just wanted to say thanks so much for taking the time to make this Pull Request 🙌 We appreciate it so much.

When looking at the proposed changes and the description you wrote out, one thing that stood out is that I do think we want to preserve the existing behavior where folks have the option to memoize, or not, when passing in props like `data` and `columns`. Additionally, if folks only provided a stable reference for one of these (like `data`) then a goal was to not cause a re-render if the other prop, in this case `columns`, did not have a stable reference.

This is where the existing setup came from to try and satisfy these scenarios and it'd be great if we could continue to support those, if possible 👀 Would love to hear what you think, and just wanted to say thanks again for taking the time to do this!

PR #3972, opened 2023-11-24
[Fix] Non-standard formatted tasklist items are not identified as tasklist items

iulia-b:
- Closes https://github.com/github/issues/issues/8246

The MD editor is not considering a valid tasklist item a line considering of an item that contains multiple spaces between the `-` delimiter and the `[ ]` textbox. This used to work as expected in the Rails experience - because of the differences in implementation, the `[ ]` there had a data attribute decorated with the position in the list and was not parsing again the text line, when the textbox was checked.

This PR is changing the regex that parses the line and takes into account 1 or 2 spaces that can be present there. i used `{1, 2}` instead of `*` for the repetitions so that we don't consider to be correct tasklist items that are too deformed. 

#### Changed

Regex for identifying a task in a markdown list in order to accept multiple spaces between the delimiter and the box

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/drafts/MarkdownEditor/_useListEditing.ts
```diff
@@ -45,7 +45,7 @@ const isNumericListItem = (item: ListItem | null): item is NumericListItem => ty
 export const parseListItem = (line: string): ListItem | null => {
   const result = listItemRegex.exec(line)
   if (!result) return null
-  const [, leadingWhitespace = '', fullDelimeter, itemNumberStr = '', taskBox = null, text] = result
+  const [, leadingWhitespace = '', fullDelimeter, itemNumberStr = '', , taskBox = null, text] = result
```
iansan5653, line 48: Here you are skipping the capture group for the one or two spaces, meaning that the updated Markdown will always have only one space. Do we maybe want to preserve the number of spaces like we do with `leadingWhitespace` here?

src/drafts/MarkdownEditor/_useListEditing.ts
```diff
@@ -29,7 +29,7 @@ const calculateNextListItemStarter = ({leadingWhitespace = '', delimeter, taskBo
  *  3. Task box (optional)
  *  4. Everything following
  */
-export const listItemRegex = /^(\s*)([*-]|(\d+)\.)\s(?:(\[[\sx]\])\s)?(.*)/i
+export const listItemRegex = /^(\s*)([*-]|(\d+)\.)(\s{1,2})(?:(\[[\sx]\])\s)?(.*)/i
```
iansan5653, line 32: Are tab characters valid here? Maybe more correct would be space only (` `). Not sure though. 

```suggestion
export const listItemRegex = /^(\s*)([*-]|(\d+)\.)( {1,2})(?:(\[[\sx]\])\s)?(.*)/i
```

iansan5653 (approved):
Looks good to me! Just some small questions that I don't think would be significant/noticeable issues.

src/drafts/MarkdownEditor/_useListEditing.ts
iulia-b: Addressed in [ad9c47f](https://github.com/primer/react/pull/3972/commits/ad9c47fc25f14fd1c5456de51560c7766b883512)

src/drafts/MarkdownEditor/_useListEditing.ts
iulia-b: `\s` is identifying tabs as well, i just checked

src/drafts/MarkdownEditor/_useListEditing.ts
```diff
@@ -45,22 +46,23 @@ const isNumericListItem = (item: ListItem | null): item is NumericListItem => ty
...
     taskBox: taskBox as '[ ]' | '[x]' | null,
   }
 }
 
 export const listItemToString = (item: ListItem) =>
   `${item.leadingWhitespace}${typeof item.delimeter === 'number' ? `${item.delimeter}.` : item.delimeter}${
-    item.taskBox ? ` ${item.taskBox}` : ''
-  } ${item.text}`
+    item.middleWhitespace
+  }${item.taskBox ? ` ${item.taskBox}` : ''} ${item.text}`
```
iansan5653, line 65: Since we are preserving the `middleWhitespace` now we don't need this additional space:

```suggestion
  }${item.taskBox || ''} ${item.text}`
```

iulia-b: ah good catch

joshblack (approved):

Issue #2457, opened 2022-10-19
leadingIcon and similar props should pass React.ReactNode instead of components


mattcosta7:
When passing props like the `leadingIcon` and `trailingIcon` props on a button, it's not uncommon to pass something that relates to props from the containing function. 

Currently, to do this one needs to wrap this in a function

```jsx
<Button leadingIcon={() => getAnIconBasedOnSomething(props)} />
```

This leads to `Button` needing to render more often than necessary

we could always memoize this, of course

```jsx
<Button leadingIcon={useCallback(() => getAnIconBasedOnSomething(prop1, prop2), [prop1, prop2])} />
```

But react already has a built in mechanism for dealing with this, components. 

```jsx
<Button leadingIcon={<IconBasedOnSomething prop={prop} />} />
```

This manages these icons, and similar (I assume in other components) in an internal way, similar to `children` by default, allows the parent context to control what gets rendered, rather than internalizing that to the button, and avoids the need to either bloat code by memoizing what are effectively inline components or handling these inefficiently by default

https://github.com/primer/react/blob/c3eedb23e632d73d77aa65e973c21382b8921ce3/src/Button/ButtonBase.tsx#L29-L42

could be minimally changed to support that

```tsx
const ButtonBase = forwardRef<HTMLElement, ButtonProps>(
  ({children, as: Component = 'button', sx: sxProp = defaultSxProp, ...props}, forwardedRef): JSX.Element => {
    const {leadingIcon, trailingIcon, variant = 'default', size = 'medium'} = props
    const {theme} = useTheme()
    const baseStyles = useMemo(() => {
      return merge.all([getButtonStyles(theme), getSizeStyles(size, variant, false), getVariantStyles(variant, theme)])
    }, [theme, size, variant])
    const sxStyles = useMemo(() => {
      return merge(baseStyles, sxProp as SxProp)
    }, [baseStyles, sxProp])

    return (
      <StyledButton as={Component} sx={sxStyles} {...props} ref={forwardedRef}>
        {leadingIcon && (
          <Box as="span" data-component="leadingIcon" sx={iconWrapStyles}>
            {leadingIcon}
          </Box>
        )}
        {children && <span data-component="text">{children}</span>}
        {trailingIcon && (
          <Box as="span" data-component="trailingIcon" sx={trailingIconStyles}>
            {trailingIcon}
          </Box>
        )}
      </StyledButton>
    )
  }
) as PolymorphicForwardRefComponent<'button' | 'a', ButtonProps>
[...]

PR #3979, opened 2023-11-27
refactor(Button): add support for React.ReactNode for slot props

joshblack:
Closes https://github.com/primer/react/issues/2457

Update `leadingVisual`, `trailingVisual`, and `trailingAction` to support `React.ReactNode` in addition to the existing `React.ElementType`. This change also updates the examples to use `React.ReactNode` as this will be the recommended practice (and potentially the only option) moving forward.

#### Changed

- Extend type for `leadingVisual`, `trailingVisual`, and `trailingAction` to include `React.ReactNode`
- Update docs for button components to use `React.ReactNode`

### Rollout strategy

- [x] Minor release

This is an additive change that teams can move to over time and is potentially an opportunity for codemods.

### Testing & Reviewing

- Verify that there is no loss of functionality with the changes to types and how these slots are rendered
- Would also love your feedback on how we're checking for element types with the `react-is` package and if there would be a better way to support this 👀

### DIFF
--- a/src/Button/ButtonBase.tsx
+++ b/src/Button/ButtonBase.tsx
@@ -1,4 +1,5 @@
 import React, {ComponentPropsWithRef, forwardRef, useMemo} from 'react'
+import {isValidElementType} from 'react-is'
 import {ForwardRefComponent as PolymorphicForwardRefComponent} from '../utils/polymorphic'
 import Box from '../Box'
 import {BetterSystemStyleObject, merge} from '../sx'
@@ -76,7 +77,7 @@ const ButtonBase = forwardRef(
             <Box as="span" data-component="buttonContent" sx={getAlignContentSize(alignContent)}>
               {LeadingVisual && (
                 <Box as="span" data-component="leadingVisual" sx={{...iconWrapStyles}}>
-                  <LeadingVisual />
+                  {isValidElementType(LeadingVisual) ? <LeadingVisual /> : LeadingVisual}
                 </Box>
               )}
               {children && <span data-component="text">{children}</span>}
@@ -86,13 +87,13 @@ const ButtonBase = forwardRef(
                 </Box>
               ) : TrailingVisual ? (
                 <Box as="span" data-component="trailingVisual" sx={{...iconWrapStyles}}>
-                  <TrailingVisual />
+                  {isValidElementType(TrailingVisual) ? <TrailingVisual /> : TrailingVisual}
                 </Box>
               ) : null}
             </Box>
             {TrailingAction && (
               <Box as="span" data-component="trailingAction" sx={{...iconWrapStyles}}>
-                <TrailingAction />
+                {isValidElementType(TrailingAction) ? <TrailingAction /> : TrailingAction}
               </Box>
             )}
           </>
--- a/src/Button/types.ts
+++ b/src/Button/types.ts
@@ -52,17 +52,17 @@ export type ButtonProps = {
   /**
    * The leading visual which comes before the button content
    */
-  leadingVisual?: React.ElementType | null
+  leadingVisual?: React.ReactNode | React.ElementType | null
 
   /**
    * The trailing visual which comes after the button content
    */
-  trailingVisual?: React.ElementType | null
+  trailingVisual?: React.ReactNode | React.ElementType | null
 
   /**
    * Trailing action appears to the right of the trailing visual and is always locked to the end
    */
-  trailingAction?: React.ElementType | null
+  trailingAction?: React.ReactNode | React.ElementType | null
 
   children?: React.ReactNode
 
--- a/.changeset/odd-goats-compare.md
+++ b/.changeset/odd-goats-compare.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Add support for React.ReactNode to leadingVisual, trailingVisual, and trailingAction for Button components
--- a/src/Button/Button.features.stories.tsx
+++ b/src/Button/Button.features.stories.tsx
@@ -12,9 +12,9 @@ export const Danger = () => <Button variant="danger">Danger</Button>
 
 export const Invisible = () => <Button variant="invisible">Invisible</Button>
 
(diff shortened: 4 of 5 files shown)


joshblack:
cc @mattcosta7 and @iansan5653, would love any thoughts/opinions you all have on this approach and if you'd recommend anything different 👀

src/Button/types.ts
```diff
@@ -52,17 +52,17 @@ export type ButtonProps = {
   /**
    * The leading visual which comes before the button content
    */
-  leadingVisual?: React.ElementType | null
+  leadingVisual?: React.ReactNode | React.ElementType | null
```
mattcosta7, line 55: Quickly skimming this:

`ReactNode` = ` string | number | boolean | react.ReactElement<any, string | react.JSXElementConstructor<any>> | Iterable<react.ReactNode> | react.ReactPortal | react.PromiseLikeOfReactNode | null | undefined`

Do we sufficiently handle things like string, number, true, etc here? Do we need additional tests/stories for that, or would we want to limit this a bit more in types?

joshblack: @mattcosta7 my assumption from the parent issues was to support the generic `React.ReactNode` but am happy to constrain it if that would be helpful 👍 I think `React.ReactElement` in particular would satisfy the cases that have been brought up. In particular:

```tsx
import * as React from 'react';

function Component() {
  return 'example';
}

const Memo = React.memo(Component)

const ForwardRef = React.forwardRef(Component)

const Lazy = React.lazy(async () => {
  return {
    default: Component
  };
})

const T1: React.ReactElement = <p>Hello</p>;
const T2: React.ReactElement = <>hi</>;
const T3: React.ReactElement = <Lazy />;
const T4: React.ReactElement = <Memo />;
const T5: React.ReactElement = <ForwardRef />;
```

src/Button/ButtonBase.tsx
```diff
@@ -76,7 +77,7 @@ const ButtonBase = forwardRef(
             <Box as="span" data-component="buttonContent" sx={getAlignContentSize(alignContent)}>
               {LeadingVisual && (
                 <Box as="span" data-component="leadingVisual" sx={{...iconWrapStyles}}>
-                  <LeadingVisual />
+                  {isValidElementType(LeadingVisual) ? <LeadingVisual /> : LeadingVisual}
```
iansan5653, line 80: TIL of `isValidElementType`. One odd example I see in [the docs](https://github.com/facebook/react/tree/main/packages/react-is) is that it looks like an element tag name will be `true`:

```ts
ReactIs.isValidElementType("div"); // true
```

If the function actually works like this, it creates some potential weirdness where if the consumer passes a string that happens to match a tag name, we will try to render that tag name instead. But if they pass any other string we will render the string. 

This could be considered expected behavior - `<div />` is valid JSX after all - but it's a pretty strange gotcha. And to be fair I don't expect consumers to pass strings here much since that would make for an odd-looking button, but I think if we support it in the types we should support rendering it.

I wonder if maybe we should check `isValidElementType(LeadingVisual) && typeof LeadingVisual !== "string"`. In which case, maybe we should abstract this logic out to a shared helper util / type.

src/Button/types.ts
iansan5653: FWIW I think it's a good idea to support all of `ReactNode`, even if it takes a bit more logic, because having to reason whether the JSX you're working with is `ReactNode` vs `ReactElement` can be frustrating. For example, you might want to do something like `leadingVisual={someBoolean && <Icon />}` or `leadingVisual={children}` and it's annoying / less readable to have to wrap those in fragments.

iansan5653 (commented):
:rocket: awesome, I'm happy to see Primer moving towards more consistency here! I noticed one small potential bug but other than that it looks great.

src/Button/types.ts
mattcosta7: Agreed re supporting all of react node, but just making sure we test for all of the edges there

the one you mentioned around specific strings behaving properly, for instance being a good case to handle. 

what do the booelans true/false do, etc.  With tests for all of the edges here I think we should be solid

src/Button/ButtonBase.tsx
mattcosta7: I think that's a good call to force a string to be a string, instead of a maybe element type.   Good call out!

src/Button/types.ts
joshblack: When thinking about the edge cases coming up, I definitely feel more confident for things like `React.ReactElement` than things like strings, numbers, or booleans. I'm not sure what kind of scenarios would exist for these kinds of primitives values so I would be unsure what to test for or document in storybook. If there are examples to use then I could definitely pursue that but without those I'm not sure what to add in order to cover those cases 👀 

From the `boolean` perspective, it would be great to use a ternary over the logical operators. From the `string` or `number` perspective, I assume that `number` would be analogous to the `count` prop and `string` would likely need an element anyways to be styled as a visual.

joshblack: That makes total sense, @mattcosta7. Definitely agreed that usage-wise folks can kind of get away with anything here (like in your example, if a number is not supported well they could return a function that renders a number 😅). This is definitely where I got tripped up on looking for edge cases as, at the end of the day, these aren't supported paths necessarily and we don't really have a way to restrict them.

I could do something like a helper similar to:

```tsx
// This would be the `leadingVisual`, `trailingVisual`, etc type
type Slot = React.ReactNode | React.ElementType;

function renderSlot(slot: Slot): React.ReactNode {
  if (typeof slot === 'string') {
    return slot;
  }

  if (typeof slot === 'function') {
    return createElement(slot);
  }

  if (typeof slot === 'boolean') {
    return null;
  }

  return slot;
}
```

That would make it more obvious what these edge cases may be and test around those? But from the context of a button I'm not sure and would love any advice/suggestions if this is going down a route that isn't really what you were thinking of.

mattcosta7: Yeah, in that case, just limiting to elementtype might at least be clear enough - we don't necessarily need to handle every possible case internally.  at some point, users gonna do what users gonna do

PR #3982, opened 2023-11-28
refactor(useId): update useId to use React.useId, update ssr utils

joshblack:
Background for these changes: https://github.com/github/primer/issues/1715#issuecomment-1830411602

Update SSRProvider, useSSRSafeId to use the native React 18 useId() instead of @react-aria/ssr

#### Changed

- Update `SSRProvider` to noop
- Update `useSSRSafeId` to use our internal `useId`
- Update `useId` to use the native `React.useId()` hook
- Update snapshots to use new useId generation

#### Removed

- Remove `@react-aria/ssr` from the project

### Rollout strategy

- [x] Minor release

Hopefully this can be contained in a minor release as the changes should all be compatible 🤞 This PR does end up removing support for `@react-aria/ssr` and replaces it but we could also keep the dependency in the project and remove it later if it ends up being problematic. This work should definitely be tested out in dotcom to see what the impact is.

### Testing & Reviewing

- Would appreciate a look at the type signatures for the changed exports to make sure they are equivalent
- Verify that id generation still works as expected in components which use this hook
- Double-check the rollout plan to see if I missed anything 👀


broccolinisoup (approved):
TL;DR

Looks beautiful! ✨ 

> Would appreciate a look at the type signatures for the changed exports to make sure they are equivalent

All makes sense to me ✅ 

>Verify that id generation still works as expected in components which use this hook

Checked the Tooltip and UnderlineNav, both are golden ⭐ 
> Double-check the rollout plan to see if I missed anything 👀

Based on https://github.com/github/primer/issues/1715#issuecomment-1830411602, I just wanted to say that I LOVE the migration plan. Great example of how to gradually move from A to B and from B to C 😊 

One question about the deprecating the `SSRProvider ` and the `useSSRSafeId `, should we do it in this PR or a follow up PR is better? Curious to hear your thoughts.

siddharthkp (approved):
Very nice!

PR #3989, opened 2023-11-30
Add aria-disabled to the storybooks to highlight issues

pksjce:
Hi, so during my investigation I added `aria-disabled` to the `Button` component and found a couple of styling issues
- `aria-disabled='false'` and `aria-disabled='true'` are treated the same. aria-disabled='false' should not add disabled styles to the button

Image

- Invisible button does not respect `aria-disabled` values at all

Image


pksjce:
True. Closing this now

PR #3990, opened 2023-11-30
Add stories to show Tooltip and Dialog usage in Inactive Buttons

pksjce:


pksjce:
@mperrotti - I don't think there is any harm showcasing specific expectations for the inactive button. 
Also as @langermank pointed out, I can't get the dialog and tooltip to work with this implementation. Since this is on top of your PR, I was hoping you'd have a fix off the top of your mind? Else I'm planning on debugging this and making sure it works. 
I also wanted to highlight that adding a tooltip optionally is a little awkward and if devs who were working on gracefully degrading using the inactive button had questions, we could point them to this little implementation.

pksjce:
Closing this as we will update stories for these use cases if questions etc arise on usage of inactive component states.

PR #4000, opened 2023-12-01
Allow keyboard navigation to dialog's scroll region

strackoverflow:
Closes https://github.com/github/accessibility-audits/issues/5565 (GitHub staff only)

This pull request fixes an accessibility issue in Dialog where the dialog's scrollable body could not be focused with the keyboard, preventing keyboard users from navigating to the dialog content.

#### Changed

Fix an issue where the scrollable Dialog body could not be focused with the keyboard.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/internal/components/ScrollableRegion.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import Box from '../../Box'
+import {useOverflow} from '../hooks/useOverflow'
+
+type ScrollableRegionProps = React.PropsWithChildren<{
+  'aria-labelledby'?: string
```
joshblack, line 6: One opportunity that I realized when going over this is that we could make `aria-labelledby` required here so that we can make sure usage follows [this guidance](https://w3c.github.io/aria/#region:~:text=Authors%20MUST%20give%20each%20element%20with%20role%20region%20a%20brief%20label%20that%20describes%20the%20purpose%20of%20the%20content%20in%20the%20region.)

> Authors MUST give each element with role region a brief label that describes the purpose of the content in the region.

Curious what you think 👀 Would this mess anything up if it was required?

```suggestion
  'aria-labelledby': string
```

joshblack (approved):
Looks great! Just had one comment 👀

src/internal/components/ScrollableRegion.tsx
strackoverflow: I think that sounds great! The table and the dialog are the only consumers of this internal component so we should be good changing this requirement.

strackoverflow: @joshblack actually, I am now realizing that we would probably also have to make `aria-labelledby` a required prop on the `Table` component as well (since it gets passed through to the scrollable region), and that feels like it could be a breaking change?

joshblack: @strackoverflow ah, great catch, you're totally right. I'll add it to our list to consider for v37 then and totally ignore me for this 😂

PR #4009, opened 2023-12-04
Update release-schedule.yml

joshblack:
Update our release schedule to run at the end of a release schedule (16:30 GMT+0000). Currently it's running too soon which is causing the timing problems

#### Changed

- Change CRON trigger of `release-schedule.yml`

### Rollout strategy

None

### Testing & Reviewing

- Verify that the CRON will run at 16:30 GMT+0000 every Monday


broccolinisoup (approved):

joshblack:
@broccolinisoup exactly (or at least that's the intent lol)

PR #4020, opened 2023-12-05
SelectPanel2: Use html dialog

siddharthkp:
- Drop AnchoredOverlay in favor of html dialog
- using `useAnchoredPosition` for positioning now
- Implemented click outside behavior suggested by @maximedegreve, [see code comment for full context](https://github.com/primer/react/pull/4020/files#diff-be5ba0fdc0d9e5e7cdbfcdf60786509127cdb59f9668e019e96c992088413fd6R177-R183)

https://github.com/github/primer/assets/1863771/f2a202cb-e2fc-4b9a-ac87-65d1d0815210

Alt text: When a SelectPanel is opened, it renders a dialog with a focus trap, you see that the accessibility panel only shows a dialog. Pressing the `Esc` key closes the panel. Looking at the elements panel, you can see that the dialog is rendered in a `top layer`

---

Question for @maximedegreve related to html dialog:

[Answered] We decided clicking outside the panel should not close the panel, we still hold the same opinion? Is there something we can do to make this not feel like a bug? [See demo in storybook](https://primer-70364bb629-13348165.drafts.github.io/storybook/?path=/story/drafts-components-selectpanel--default). 

_Update: If the user clicks outside, we nudge the user to use the actions in the footer. [See demo](https://primer-70364bb629-13348165.drafts.github.io/storybook/iframe.html?args=&id=drafts-components-selectpanel--default&viewMode=story)_


maximedegreve:
Hey @siddharthkp 

I've previously raised this concern [within Slack](https://github.slack.com/archives/CSGAVNZ19/p1700146835337559?thread_ts=1698788240.443079&cid=CSGAVNZ19) as well. 

The cancel and save buttons are visually clear to indicate confirming is required in the dialog. However, this pattern isn't obvious to everyone, which can be frustrating. This is particularly true for long-time users, who have been on our platform for over 10 years and may find these changes annoying. Additionally, clicking on the background without any response can make it seem as though we've deliberately worsened the user experience, especially for those who are not familiar with those accessible patterns.

The best thing I think we can do now is introducing a gelatine or bounce effect on the buttons when this annoyance occurs? This could be an interesting and engaging way to guide users towards understanding and adopting the new behavior, without giving the impression that we've overlooked something that is an accessibility requirement.

https://github.com/primer/react/assets/980622/13f50498-66d3-43a3-b16c-53eaafc5b25d

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -135,26 +148,77 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+        sx={{
+          ...position,
+          // reset dialog default styles
+          border: 'none',
+          padding: 0,
+          margin: 0,
+          '::backdrop': {background: 'transparent'},
+
+          '& [data-selectpanel-primary-actions]': {
+            animation: footerAnimationEnabled ? 'selectpanel-gelatine 0.5s linear' : 'none',
```
mperrotti, line 210: I think this feels a little nicer when we speed it up to `350ms`

mperrotti (approved):

mperrotti:
I really like @maximedegreve's suggestion to "highlight" the buttons when somebody tries to close by clicking the background scrim.

I think we should also add this functionality to `Dialog`. @maximedegreve and other reviewers - how do you feel about this? This doesn't have to be part of this PR - we could create a follow-up issue or PR to do this.

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -303,6 +366,7 @@ const SelectPanelSearchInput: React.FC<TextInputProps> = ({onChange: propsOnChan
         <TextInput.Action
           icon={XCircleFillIcon}
           aria-label="Clear"
+          tooltipDirection="w"
```
broccolinisoup, line 369: Totally nitpick and out of curiosity: Is this to hack the broken tooltip in TextInput.Action? (I created an issue for [this](https://github.com/primer/react/issues/4091))

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -135,26 +148,77 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+    },
+    [anchorRef.current, dialogRef.current],
+  )
+
+  /* 
+    We don't close the panel when clicking outside.
+    For many years, we used to save changes and closed the dialog (for label picker)
+    which isn't accessible, clicking outside should discard changes and close the dialog
+    Fixing this a11y bug would confuse users, so as a middle ground,
+    we don't close the menu and nudge the user towards the footer actions
```
broccolinisoup, line 182: This is cool! I think this behaviour is one if the first ones in Primer, right?

broccolinisoup (approved):
Looks great ✨ - just left a few notes

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Done!

Here's a video to compare durations:

https://github.com/primer/react/assets/1863771/44c8ba47-c05a-4d4c-9747-4a0a64abb48c

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: > Is this to hack the broken tooltip in TextInput.Action?

It is! Thank you.

Going to use Tooltip v2 in a new PR, so we might not need the issue long term

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: It is!  We should consider this experimental for now!

src/drafts/SelectPanel2/SelectPanel.tsx
maximedegreve: Love that you actually did this in the end! ❤️

maximedegreve: I couldn't see it in the code but is it disabled when reduced motion is on?

Issue #3992, opened 2023-11-30
The resizable pane for PageLayout is triggered by right click


joshblack:
### Description

When going through the updates to the Page layout pane for resizable panes, I noticed that both the existing implementation and the updated one responded to a right click as a click event which caused the resize handle to unintentionally stick to the cursor.

### Steps to reproduce

1. Visit the storybook: https://primer.style/react/storybook/?path=/story/components-pagelayout-features--resizable-pane
2. Right click on the resizable pane
3. Dismiss the context menu and move the mouse around to observe the issue

https://github.com/primer/react/assets/3901764/1e4c4e21-fe38-4b83-b9ae-f4dc4e3023f9

### Version

v36.2.0

PR #4022, opened 2023-12-05
fix(PageLayout): update onMouseDown to check which button is pressed

joshblack:
Closes https://github.com/primer/react/issues/3992

#### Changed

- Update `onMouseDown` in the vertical resize handler to only start dragging if the "main" mouse button is active during the event

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- Verify that the resizable pane no longer fires when right click is used before dragging


pksjce (approved):
Nice fix 👍

PR #4037, opened 2023-12-08
avoid creating a div on every render in portal

mattcosta7:
### Changelog

Portal avoids `createElement` call on every render

#### Changed

Portal

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Portal/Portal.tsx
+++ b/src/Portal/Portal.tsx
@@ -65,14 +65,18 @@ export const Portal: React.FC<React.PropsWithChildren<PortalProps>> = ({
   onMount,
   containerName: _containerName,
 }) => {
-  const hostElement = document.createElement('div')
+  const elementRef = React.useRef<HTMLDivElement | null>(null)
+  if (!elementRef.current) {
+    const div = document.createElement('div')
+    // Portaled content should get their own stacking context so they don't interfere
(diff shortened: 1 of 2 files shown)


src/Portal/Portal.tsx
```diff
@@ -65,14 +65,18 @@ export const Portal: React.FC<React.PropsWithChildren<PortalProps>> = ({
...
+    elementRef.current = div
+  }
 
-  // Portaled content should get their own stacking context so they don't interfere
-  // with each other in unexpected ways. One should never find themselves tempted
-  // to change the zIndex to a value other than "1".
-  hostElement.style.position = 'relative'
-  hostElement.style.zIndex = '1'
-  const elementRef = React.useRef(hostElement)
+  const element = elementRef.current
```
mattcosta7, line 79: since we only used the ref, which is created on the first render, we can just create it on the first render. 

this is still not SSR safe, however, and we should probably change this to account for that, but at least we're not wasting memory for no reason?

mperrotti (approved):

PR #4039, opened 2023-12-09
Sync color mode prior to subscribing in ThemeProvider

mattcosta7:
related to https://github.com/github/issues/issues/8432

### Changelog

Ensure theme syncs between initial render and setting up listeners in useEffect. This should handle the unlikely but possible scenario that in between initial render and the effect, the user's preference for theme color has changed. This team is not large, but potentially happens for users who use automatic theme switching based on time/schedule

#### Changed

ThemeProvider setColorMode gets called when setting up subscription

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

It's not super straightforward to validate, but generally, when setting up subscriptions in an efffect, it's important to ensure that the time between render and the effect was accounted for.

In this case, I believe that occasionally when this effect occurs far enough after that initial render, the match state can change prior to listening to the change event, leading to some ui tearing. 

see: `useSyncExternalStore`'s shim implementation (which is basically a subscription to an event) - https://github.com/facebook/react/blob/40f653d13c363c6f81b13de67ce391991fb1f870/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js#L103-L122

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/fresh-hotels-begin.md
+++ b/.changeset/fresh-hotels-begin.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Sync theme in effect
--- a/src/ThemeProvider.tsx
+++ b/src/ThemeProvider.tsx
@@ -160,15 +160,20 @@ function useSystemColorMode() {
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     const media = window?.matchMedia?.('(prefers-color-scheme: dark)')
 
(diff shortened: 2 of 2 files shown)


joshblack (approved):

mattcosta7:
gh/gh integration tests are 🟢

PR #4048, opened 2023-12-11
Dialog: Full screen & Bottom sheet support

maximedegreve:
### Closes
- https://github.com/github/primer/issues/2414
- https://github.com/github/primer/issues/2413

Note: I've chosen not to tackle the[ issues mentioned here](https://github.com/github/primer/issues/2372) to avoid causing any disruptive changes.

#### New

1️⃣ `action-sheet` dialog

We've implemented a range slider to easily adjust the size of the dialog using only a keyboard, enhancing accessibility. We've used a similar approach for our [SplitPageLayout resizer](https://github.com/primer/react/pull/3903). I've also implemented reduced-motion support.

https://github.com/primer/react/assets/980622/c55fc0cb-1cea-4dab-94cb-e9239a870431

https://github.com/primer/react/assets/980622/babadea0-8794-40b1-a812-cf9b6f7617f8

2️⃣  `full-screen` dialog

This is quite simple, as it closely resembles the existing dialog behavior, with the only difference being that it's full-screen.

https://github.com/primer/react/assets/980622/eb043543-8a06-4854-bf0b-fc6c446213d2

#### Changed

- Storybook features always display dialogs in an open state so that snapshots can effectively capture any changes.
- Removed the ability to change props in features for simplicity and consistency with other components
- In the example called 'WithCustomRenderers', I've placed the header, content, and footer components inside the 'WithCustomRenderers' component. This change improves readability, especially when highlighting the code.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


joshblack:
Hey @maximedegreve! Sorry for the delay on this, was spending some time reading the related issues to get the context to understand the changes 👀 

Just wanted to check-in with you to make sure I understood things right, the corresponding issues are talking about two things: 

1. The ability for a dialog to go full screen (the `full-screen` type in this PR)
2. A new presentation for a dialog called an action sheet (the `action-sheet` type in this PR)

And it looks we're supporting using these types responsively through the `type` prop on `Dialog` to get something like:

```tsx
<Dialog type={['full-screen', 'default']} />
```

Hope that makes sense 🤔 Let me know if I'm missing anything.

With this in mind, I had a couple of questions come up:

- Are these new types restricted to any specific viewport sizes or should they be visible across viewports?
- Does action sheet in this PR correspond to bottom sheets in the design guidance or are they a separate type?

joshblack:
@maximedegreve nice, makes sense. Would it be helpful to limit usage of these variants to only specific breakpoints (if feasible) so that these are only used in appropriate contexts? Or even just roll these into the components themselves as default behavior? (This would like require a specific component for BottomSheet but could be worth it)

maximedegreve:
@joshblack 💯 feasible. At first, I thought the same. But I have concerns about limiting it like this. Some dialogs already are playing with a full screen mode right now in desktop and bottom sheets are also very similar to side sheets. I would assume side sheets would be an addition to the future. Will our API become inconsistent and harder to read if we allow side sheets on desktop but not bottom sheets? In this case we would need `type: sidesheet bottomsheetOnMobile: true` or `type:default bottomSheetOnMobile:true`.  In all our other components we use the {narrow: value, default: value...} pattern.

Red arrow pointing to a full screen icon a 'New conversation' header.

I do believe that components using the dialog, like SelectPanel, might prefer these built-in defaults though. 

It could be a good idea to check with the rest of team if we want to maintain flexibility at the component level here or not.

src/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -17,8 +17,7 @@ export interface ConfirmationDialogProps {
    * Required. This callback is invoked when a gesture to close the dialog
    * is performed. The first argument indicates the gesture.
    */
-  onClose: (gesture: 'confirm' | 'close-button' | 'cancel' | 'escape') => void
-
+  onClose: (gesture: 'confirm' | 'cancel' | 'close-button' | 'escape' | 'drag' | 'overlay') => void
```
mperrotti, line 20: Would we ever even use an action sheet or full screen confirmation dialog?

src/Dialog/Dialog.docs.json
```diff
@@ -37,14 +37,19 @@
...
+      "description": "This method is invoked when a gesture to close the dialog is used (either an Escape key press, clicking/tapping on the backdrop, clicking/tapping the 'X' in the top-right corner or dragging away a bottom sheet). The gesture argument indicates the gesture that was used to close the dialog."
     },
     {
       "name": "role",
       "type": "'dialog' | 'alertdialog'",
       "description": "The ARIA role to assign to this dialog."
     },
+    {
+      "name": "type",
+      "type": "'default' | 'full-screen' | 'bottom-sheet'",
```
mperrotti, line 50: I'm wondering if we should just make a `'bottom-sheet'` dialog it's own component instead of rendering it using a `Dialog`. Then, we could set `'full-screen'` in the `width` and `height` props. If `'full-screen'` is passed to _either_ `width` or `height`, the dialog would be full screen.

Curious to hear what other reviewers think.

src/Dialog/Dialog.tsx
```diff
@@ -96,12 +101,14 @@ export interface DialogProps extends SxProp {
...
-   * (either 'close-button' or 'escape').
+   * This method is invoked when a gesture to close the dialog is used
+   * (either an Escape key press, clicking/tapping on the backdrop,
+   * clicking/tapping the 'X' in the top-right corner or dragging away a
+   * bottom sheet). The gesture argument indicates the gesture that was
+   * used to close the dialog.
    */
-  onClose: (gesture: 'close-button' | 'escape') => void
+
+  onClose: (gesture: 'close-button' | 'escape' | 'drag' | 'overlay') => void
```
mperrotti, line 111: Some concern as I posted in `ConfirmationDialog`:

> It could be confusing if somebody tries to pass 'drag' or 'overlay' on something that doesn't use those gestures. I'm not sure of a better way to handle those, but maybe we can explicitly mention which variants accept which gestures in the JSDoc comment and prop docs.

src/Dialog/Dialog.tsx
```diff
@@ -421,9 +504,11 @@ const DialogCloseButton = styled(Button)`
   color: ${get('colors.fg.muted')};
   padding: ${get('space.2')};
   align-self: flex-start;
+  pointer-events: auto; // Allows clicking when combined with a draggable bottom-sheet header
```
mperrotti, line 507: What happens if we don't set a `pointer-events` style at all? I'd assume it would be the same as `pointer-events: auto;` unless we style `pointer-events` somewhere else that I missed.

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,327 @@
...
+ * Props to customize the rendering of the Dialog.
+ */
+export interface DialogBottomSheetProps extends SxProp {
+  /**
+   * This method is invoked when a gesture to close the dialog is used
+   */
+  onClose: (gesture: 'close-button' | 'escape' | 'drag' | 'overlay') => void
+
+  /**
+   * Passed through to use as the draggable area of the dialog.
```
mperrotti, line 22: Non-blocking nitpick: I'd reword this to "Content that appears in the draggable header area." The current description kind of sounds like it lets you change what we render instead of the the blue drag "handle".

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+  /**
+   * Identifies the element (or elements) that describes the element on which the attribute is set.
+   */
+  ariaDescribedby: string
+
+  /**
+   * Indicates whether an element is modal when displayed.
+   */
+  ariaModal: boolean
```
mperrotti, line 46: This seems right to me, but I'd double-check with a11y folks. Some of the `aria-` attributes have inconsistent behavior on different screen readers.

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+  const dragStop = (e: MouseEvent | TouchEvent) => {
+    if (!dialogRef.current) return
+
+    e.preventDefault()
+
+    isDragging.current = false
+    const sheetHeight = parseInt(dialogRef.current.style.height)
+
+    const isReduced = prefersReducedMotion()
```
mperrotti, line 105: Non-blocking nitpick: `isReduced` should already be in scope and shouldn't need to be redefined. I don't think it would change between render and `onDragStop`.

src/internal/components/ScrollableRegion.tsx
```diff
@@ -1,13 +1,15 @@
...
 import Box from '../../Box'
 import {useOverflow} from '../hooks/useOverflow'
+import {SxProp} from '../../sx'
 
 type ScrollableRegionProps = React.PropsWithChildren<{
   'aria-labelledby'?: string
   className?: string
-}>
+}> &
+  SxProp
```
mperrotti, line 10: I think I'm missing something. Why do we need to start supporting `sx` on `ScrollableRegion`?

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,327 @@
...
+        aria-describedby={ariaDescribedby}
+        aria-modal={ariaModal}
+        sx={sx}
+      >
+        <Box sx={{position: 'relative'}}>
+          <DraggableRegion
+            onMouseDown={dragStart}
+            onTouchStart={dragStart}
+            onChange={onSliderChange}
+            type="range"
```
mperrotti, line 201: Neat technique!

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,327 @@
...
+        ref={dialogRef}
+        role={role}
+        open={open}
+        aria-labelledby={ariaLabelledby}
+        aria-describedby={ariaDescribedby}
+        aria-modal={ariaModal}
+        sx={sx}
+      >
+        <Box sx={{position: 'relative'}}>
+          <DraggableRegion
```
mperrotti, line 197: I think it might be a little nicer if the close button was the first focusable element in the DOM. This way, arrow buttons don't move the sheet around until I explicitly focus the drag handle.

mperrotti, line 197: The drag target is layered above the close button, so it isn't clickable.

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,327 @@
...
+      onClose(gesture)
+    } else {
+      setFireDelayedOnClose(gesture)
+    }
+  }
+
+  const updateSheetHeight = (height: number) => {
+    if (!dialogRef.current) return
+    dialogRef.current.style.height = `${height}dvh`
+    setSnappedHeight(height)
```
mperrotti, line 92: We might get better performance if we always render it at full height and use `translate: transformY();` to position it in and out of the viewport.

Changing `height` triggers reflow (and maybe repaint?).

mperrotti:
Also, we need to add tests: unit tests (if applicable) and visual regression tests.

src/Dialog/Dialog.docs.json
maximedegreve: My concern is that you'll need to wrap different components in a `<Box/>` for responsive behaviour. For instance, we're planning to integrate side sheets shortly too. In this scenario, I'd prefer using a side sheet on desktop and a bottom sheet on mobile. This means I'll be required to proceed in this manner:

```javascript
<Box sx={{display: ['block', 'block', 'none']}}><BottomSheet /></Box>
<Box sx={{display: ['none', 'none', 'block']}}><SideSheet /></Box>
``` 

Another situation where I need to follow this approach is when I want to display a bottom sheet on mobile, but a standard dialog on desktop.

```javascript
<Box sx={{display: ['block', 'block', 'none']}}><BottomSheet /></Box>
<Box sx={{display: ['none', 'none', 'block']}}><Dialog /></Box>
``` 

This adds extra complexity because an open `<Dialog/>` that turns into `<BottomSheet/>` on mobile would need to stay open. If the user applies separate states for each component, it might lead to unintended effects.

The proposed API really simplifies above and ensures consumers still understand the underlying semantics is a dialog.

```javascript
<Dialog title="Your title" type={{
    narrow: 'full-screen',
    regular: 'default',
    wide: 'default'
}} >
```

or

```javascript
<Dialog title="Your title" type={{
    narrow: 'bottom-sheet',
    regular: 'default',
    wide: 'default'
}} >
```

src/Dialog/Dialog.tsx
maximedegreve: Sadly if I remove it doesn't work anymore.

src/internal/components/ScrollableRegion.tsx
maximedegreve: Yes, because if we don't, the footers in the bottom sheets and full screens won't stay at the bottom if the content isn't long enough. That's why I'm using this method:

`<ScrollableRegion aria-labelledby={dialogLabelId} className="DialogOverflowWrapper" sx={{flex: 1}}>`

src/Dialog/Dialog.tsx
maximedegreve: I believe it's clearer in this context because the dialogue supports it with the bottom sheet variant. But I'm curious if those gestures are currently in used. Why would you need to know how the user closed the dialogue? What do you think...

src/Dialog/DialogBottomSheet.tsx
maximedegreve: Fixed

src/ConfirmationDialog/ConfirmationDialog.tsx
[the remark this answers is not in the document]
maximedegreve: Fixed

src/Dialog/DialogBottomSheet.tsx
maximedegreve: Makes sense. Fixed.

src/Dialog/DialogBottomSheet.tsx
maximedegreve: @ichelsea @ericwbailey is that something you could double check for me?

src/Dialog/DialogBottomSheet.tsx
maximedegreve: Whoops yes this is overkill. Fixed.

src/Dialog/DialogBottomSheet.tsx
maximedegreve: Are you sure it will reflow/repaint because the content is scrollable therefore the width and height of the scrollable content won't change. I haven't observed any performance issues. Not animating the height is kinda complex because the scrollable area needs to resize either way. Using `transform` I would have to set a height before animating to something beyond the height of the viewport and at the end of the animation then set the final height to ensure the scrollable area is correct. Things get even more tricky while you're dragging.

src/Dialog/DialogBottomSheet.tsx
maximedegreve: Fixed.

src/Dialog/Dialog.tsx
keithamus: We should avoid extending these. The gestures aren't checked in consuming code, and in the future the `onClose()` call will be called whenever the dialog is closed and will not pass a gesture back. See https://github.com/primer/react/pull/3840/files#diff-7c9e4f2540e063c93f543e6eaa44cac7dc3b0e55a8fe81030c9e571f72950a3eR97-R102

src/Dialog/DialogBottomSheet.tsx
mperrotti: Also tagging @alexislucio since she's this week's a11y FR.

src/Dialog/DialogBottomSheet.tsx
mperrotti: I checked it out in Chrome's performance dev tools, and it seems to behave just fine on my work computer. Let's leave this as-is unless we notice jank on mobile or other less powerful devices. 🙂 

---

> Are you sure it will reflow/repaint because the content is scrollable therefore the width and height of the scrollable content won't change.

I'm pretty sure the browser does a reflow whenever you change a height or width, even if no content is actually affected. Browsers even trigger a reflow when you check the dimensions of an element using `.getBoundingClientRect()` 😢

mperrotti (commented):
Looking good to me! There's one other thing I noticed in [this story](https://primer-bfeba1f965-13348165.drafts.github.io/storybook/?path=/story/components-dialog-features--bottom-sheet): the dialog doesn't dismiss when I click the close button, but it _does_ dismiss when I click the backdrop.

src/Dialog/DialogBottomSheet.tsx
maximedegreve: Oh right. I've learned something today. 🧠

src/Dialog/Dialog.tsx
maximedegreve: @keithamus Makes a lot of sense. I've removed it and always return `close-button` for now.

src/Dialog/Dialog.tsx
```diff
@@ -310,61 +298,211 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
   const body = (renderBody ?? DefaultBody)(defaultedProps)
   const footer = (renderFooter ?? DefaultFooter)(defaultedProps)
 
-  return (
-    <>
+  if (responsiveType === 'full-screen') {
```
keithamus, line 301: There is a lot of duplication going on here. It might benefit from hoisting the element into a variable. Something like:


```js
let TheDialog = NormalDialog;
if (responsiveType === 'full-screen') TheDialog = FullScreenDialog
else if (responsiveType === 'bottom-sheet') TheDialog = BottomSheetDialog
```

I haven't looked at the duplication within each of these components but it might be worth looking into that too.

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,318 @@
...
+  }
+  ::-moz-range-thumb {
+    appearance: none;
+    visibility: hidden;
+  }
+  ::-webkit-slider-thumb {
+    appearance: none;
+  }
+`
+const Content = styled.div<
```
keithamus, line 297:
```suggestion
const Content = styled.dialog<
```

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,318 @@
...
+  // Animates the bottom sheet in on mount
+  useEffect(() => {
+    showBottomSheet()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  const currentSliderValue = snappedHeight === HALF_HEIGHT ? 1 : 2
+
+  return (
+    <FullScreenContainer
```
keithamus, line 177: Why does this container exist?

src/__tests__/Dialogv2.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    Component: Dialog,
+    options: {skipAs: true, skipSx: true},
+    toRender: () => (
+      <SSRProvider>
+        <Dialog onClose={() => {}} />
+      </SSRProvider>
+    ),
+  })
+
+  it('should match snapshot', () => {
```
mperrotti, line 58: We don't do snapshot tests anymore because the errors potentially caught in snapshots are caught by VRTs.

It looks like we never had any unit tests for Dialog v2, so I don't think you should have to add them in this PR.

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,318 @@
+import React, {useEffect, useRef, useState, PropsWithChildren, MouseEvent, TouchEvent, ChangeEvent} from 'react'
```
mperrotti, line 1: We should add unit tests for this component. We would test things like whether the `onClose` prop is called when passed. You can use `npm run test:coverage -- src/Dialog/DialogBottomSheet.tsx` to get ideas for what we might want to test for.

src/Dialog/DialogBottomSheet.tsx
maximedegreve: It doesn't have to, but there are subtle differences that cause them to behave differently. When animation is used in the code, it can be easier to understand if you organize the different states by their names. Of course, this is a matter of personal preference, and you could also consider adding an 'isFullScreen' boolean property here.

``` JavaScript
const NormalDialog = styled.div<StyledDialogProps>`
  ...
  box-shadow: ${get('shadows.overlay.shadow')};
  min-width: 296px;
  max-width: calc(100vw - 64px);
  max-height: calc(100vh - 64px);
  ...
  width: ${props => widthMap[props.width ?? ('xlarge' as const)]};
  height: ${props => heightMap[props.height ?? ('auto' as const)]};
  border-radius: 12px;
  ...
  @keyframes overlay--dialog-appear {
    0% {
      opacity: 0;
      transform: scale(0.5);
    }
    100% {
      opacity: 1;
      transform: scale(1);
    }
  }

  ${sx};
`

const FullScreenDialog = styled.div<StyledDialogProps>`
  ...
  width: 100vw;
  height: 100vh;
  ...
  animation: overlay--dialog-appear ${ANIMATION_DURATION} ${get('animation.easeOutCubic')};
  @keyframes overlay--dialog-appear {
    0% {
      opacity: 0;
    }
    100% {
      opacity: 1;
    }
  }
  ${sx};
`
```

src/Dialog/Dialog.tsx
maximedegreve: It doesn't have to, but there are subtle differences that cause them to behave differently. When animation is used in the code, it can be easier to understand if you organize the different states by their names. Of course, this is a matter of personal preference, and you could also consider adding an 'isFullScreen' boolean property here.

``` JavaScript
const NormalDialog = styled.div<StyledDialogProps>`
  ...
  box-shadow: ${get('shadows.overlay.shadow')};
  min-width: 296px;
  max-width: calc(100vw - 64px);
  max-height: calc(100vh - 64px);
  ...
  width: ${props => widthMap[props.width ?? ('xlarge' as const)]};
  height: ${props => heightMap[props.height ?? ('auto' as const)]};
  border-radius: 12px;
  ...
  @keyframes overlay--dialog-appear {
    0% {
      opacity: 0;
      transform: scale(0.5);
    }
    100% {
      opacity: 1;
      transform: scale(1);
    }
  }

  ${sx};
`

const FullScreenDialog = styled.div<StyledDialogProps>`
  ...
  width: 100vw;
  height: 100vh;
  ...
  animation: overlay--dialog-appear ${ANIMATION_DURATION} ${get('animation.easeOutCubic')};
  @keyframes overlay--dialog-appear {
    0% {
      opacity: 0;
    }
    100% {
      opacity: 1;
    }
  }
  ${sx};
`
```

src/Dialog/DialogBottomSheet.tsx
maximedegreve: Same answer as: https://github.com/primer/react/pull/4048#discussion_r1455511524

src/__tests__/Dialogv2.test.tsx
maximedegreve: That makes life much easier!

src/Dialog/DialogBottomSheet.tsx
mperrotti: I can try and help too. Just push up your changes.

maximedegreve: Done

mperrotti:
I'm working on fixing your Jest issues, but it also looks like some styles broke.

Screenshot 2024-01-17 at 3 18 06 PM

src/Dialog/Dialog.tsx
```diff
@@ -310,61 +298,211 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
...
+    100% {
+      opacity: 1;
+      transform: scale(1);
+    }
+  }
+
+  ${sx};
+`
+
+const FullScreenDialog = styled.div<StyledDialogProps>`
```
mperrotti, line 456: Is this meant to be a `dialog` instead of a `div`?

src/Dialog/DialogBottomSheet.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {mediaQueries} from '../utils/layout'
+
+let matchMedia: MatchMediaMock
+
+describe('Dialog', () => {
+  beforeAll(() => {
+    matchMedia = new MatchMediaMock()
+  })
+
+  afterAll(() => {
```
mperrotti, line 15: If you plan to use `matchMedia` to test things at different media queries, you probably want to change this to `afterEach`

src/Dialog/DialogBottomSheet.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    options: {skipAs: true, skipSx: true},
+    toRender: () => (
+      <Dialog onClose={() => {}}>
+        <div>Hidden when narrow</div>
+      </Dialog>
+    ),
+  })
+
+  checkExports('Dialog', {
+    default: Dialog,
```
mperrotti, line 30: I don't think this has a default export.

```suggestion
    default: undefined,
```

src/Dialog/Dialog.tsx
mperrotti: If these _are_ meant to be `dialog`s, we'll need to update the ref types to `HTMLDialogComponent` like I suggested in `src/Dialog/DialogBottomSheet.tsx`

mperrotti (commented):
I played around with this and I think I have a way forward (or at least a workaround) to resolve your Jest issues. Maybe we can pair? Ideally with another PRC engineer present 😅

src/Dialog/Dialog.tsx
```diff
@@ -310,61 +298,211 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
...
+    0% {
+      opacity: 0;
+    }
+    100% {
+      opacity: 1;
+    }
+  }
+`
+
+const NormalDialog = styled.div<StyledDialogProps>`
```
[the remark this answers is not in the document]
maximedegreve: Yes whoops!

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,318 @@
...
+   */
+  ariaDescribedby: string
+
+  /**
+   * Indicates whether an element is modal when displayed.
+   */
+  ariaModal: boolean
+}
+
+export default React.forwardRef<HTMLDivElement, PropsWithChildren<DialogBottomSheetProps>>((props, forwardedRef) => {
```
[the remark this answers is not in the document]
maximedegreve: Actually I had to revert this again because the role is defined through the `role` prop.

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,318 @@
...
+
+  // 🔄 STATES
+
+  const [open, setIsOpen] = useState<boolean>(false)
+  const [snappedHeight, setSnappedHeight] = useState<number>(HALF_HEIGHT)
+  const [fireDelayedOnClose, setFireDelayedOnClose] = useState<boolean>(false)
+
+  // 📎 REFERENCES
+
+  const dialogRef = useRef<HTMLDivElement>(null)
```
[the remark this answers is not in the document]
maximedegreve: Actually I had to revert this again because the role is defined through the `role` prop.

src/Dialog/DialogBottomSheet.tsx
```diff
@@ -0,0 +1,318 @@
...
+  const startY = useRef(0)
+  const startHeight = useRef(0)
+  const isDragging = useRef(false)
+
+  useRefObjectAsForwardedRef(forwardedRef, dialogRef)
```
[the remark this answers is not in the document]
maximedegreve: Actually I had to revert this again because the role is defined through the `role` prop.

src/Dialog/Dialog.tsx
maximedegreve: Actually I had to revert this again because the role is defined through the `role` prop.

src/Dialog/DialogBottomSheet.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <Dialog onClose={() => {}} type="bottom-sheet">
+        <div>Hello World</div>
+      </Dialog>
+    ),
+  })
+
+  // Arrow keys support in the testing library is currently not implemented and therefore can't be tested
+  // https://github.com/testing-library/user-event/issues/871
+
+  it('expands/contracts when using the range slider', async () => {
```
mperrotti, line 33: If this merges before https://github.com/primer/react/pull/4048, I mighjt update my PR to move these tests into the same test file.

mperrotti (commented):
It's looking good! Just one little thing we might to address: we don't delay calling `onClose` when clicking the "x" or keying <kbd>Esc</kbd>

https://github.com/primer/react/assets/2313998/cf97acba-38ca-4298-a222-01a69f23a813

If this is intentional, don't worry about it.

maximedegreve:
Yeah that was intentionally because I thought you aren't swiping it away.

mperrotti (approved):
Looking good. Since this is such a big PR, it might be nice to get a review from somebody who's "officially" and engineer before merging. Not required though 🙂

Issue #3062, opened 2023-03-22
Focus not appearing on non-standard dialog button (re-opened after revert)


ctmayn:
### Description

Focus does not appear at all on Primary and Danger buttons in a Dialog Footer

### Steps to reproduce

1. Go to https://primer.style/react/storybook/?path=/story/components-dialog--stress-test
2. Open the Dialog
3. Use tab to move through the footer buttons
4. Notice that once you get to the danger and primary buttons, focus is no longer visible

### Version

v35.22

### Browser

Chrome

langermank:
This should be a quick fix. The Dialog component is using the deprecated Button component, which doesn't have the new focus styles. Here's the [source code](https://github.com/primer/react/blob/51cdbb0c3a30953ba98ff26b937ee1d13c6543aa/src/Dialog/Dialog.tsx) for Dialog.

tallys:
@langermank assigning your way, let me know if you can take a look

siddharthkp:
Thank you @mperrotti! 

This was reopened because of https://github.com/primer/react/pull/3527

siddharthkp:
Copying reasons for reverting from https://github.com/primer/react/pull/3527, 

Issues found: 

1. (link is github staff only) [ui/packages/memex/src/client/pages/insights/components/insights-upsell-dialog.tsx:66](https://github.com/github/github/blob/master/ui/packages/memex/src/client/pages/insights/components/insights-upsell-dialog.tsx#L66) uses Button as="a", the types for which are not supported

   Reported in https://github.com/github/github/actions/runs/5573804589/jobs/10181608881:

   ```
   ui/packages/memex/src/client/pages/insights/components/insights-upsell-dialog.tsx(66,11): 
   error TS2322: Type '{ as: string; buttonType: "primary"; content: JSX.Element; href: string; sx: { ml: number; }; target: string; rel: string; onClick: () => void; }' is not assignable to type 'DialogButtonProps'.
   ```

2. Checking the [storybook for the pull request](https://primer-fbbec81156-13348165.drafts.github.io/storybook/?path=/story/components-dialog--basic-dialog), it looks like the focus bug was not fixed

lesliecdubs:
@broccolinisoup during planning, we assumed this would be part of the handover of this week's release. Does that sound right to you? Can you help handle this as @joshblack takes over Release Conductor duties?

broccolinisoup:
[...]

PR #4049, opened 2023-12-11
Replace deprecated `Button` in `Dialog`

TylerJDev:
Closes https://github.com/github/primer/issues/2470, https://github.com/primer/react/issues/3062

#### Changed

Replaces deprecated `Button` component in `Dialog.tsx` with newest version.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

For a successful rollout in `github`, we'll need to adjust some items that rely on the deprecated version of `Button`. The changes needed can be viewed in the [integration test PR](https://github.com/github/github/pull/304867#pullrequestreview-1778367991).

### Testing & Reviewing

Ensure that a focus indicator is present when focused on any of the dialog's footer buttons.

[Dialog default Storybook example](https://primer-fd43f55c13-13348165.drafts.github.io/storybook/?path=/story/components-dialog--default).

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


.changeset/eight-pillows-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
```
joshblack, line 2: Since the public API of this component is changing (specifically `buttonType` is being broadened to include a new union member) could we mark this as `minor`?

TylerJDev: Definitely! Switched it to `minor`

joshblack (approved):
These changes look great to me and the upstream PR makes it seem like we only need to update tests 💯 I'll also request @broccolinisoup who I think may also have some helpful context for this work if I remember right

broccolinisoup:
> These changes look great to me and the upstream PR makes it seem like we only need to update tests 💯 I'll also request @broccolinisoup who I think may also have some helpful context for migrating this work if I remember right (I think she had helped an outside contributor trying to do this in the past and I wasn't sure what the hiccups ended up being there)

Thanks @joshblack for the ping, I appreciate it! We chatted about it in our call with Tyler and the reason that we reverted this PR is addressed in the integration PR 🤗 And the solution is approved by the code owners. Seems like we are good to go!

broccolinisoup (approved):
Looks great! Thank you so much for fixing that and providing the integration fixes as well ✨

Issue #3942, opened 2023-11-14
`Spinner` is inaccessible by default


iansan5653:
The [`Spinner`](https://primer.style/react/Spinner) component is, at first glance, a convenient way to render a loading state:

```tsx
<Box>
  {isLoading ? <Spinner /> : content}
</Box>
```

This appears straightforward, but in reality this is a bad idea because it's completely inaccessible. `Spinner` has no label by default, so the box will simply appear empty to screen readers.

The US Government CMS design system renders [`Spinner`](https://design.cms.gov/components/spinner/?theme=core) as a `status` with hidden "Loading" text. I think we probably should do the same by default, rendering it like so (hiding the SVG to avoid duplication):

```tsx
<Box role="status">
  <VisuallyHidden>Loading</VisuallyHidden>
  <SpinnerSvg aria-hidden />
</Box>
```

This would make the accessible path much easier to follow.

lesliecdubs:
Thanks for the suggestion, this does seem like a nice improvement to me as the current `Spinner` doesn't appear to have any label at all!

@TylerJDev do you concur?

iansan5653:
One alternative might be to [label the SVG properly](https://github.com/github/accessibility/blob/189ac82bda7cc9cccae0f1d22ca875f720566875/docs/wiki/svgs.md?plain=1#L4), however I'm not sure that's the best approach in this case because the spinner being an image is not useful context. `image labelled "Loading"` makes less sense than just `"Loading"`.

One concern I do have with this approach is that while it announces a loading state, it doesn't announce loading completion. Unmounting the `Spinner` removes the `status` from the UI, which results in no announcement. I'm not sure if this is a big deal -- do we need to announce loading completion? One approach could be to do `useEffect(() => () => announce("Loading completed"), [])` where `announce` causes an `aria-live` announcement.

TylerJDev:
Definitely agree. The `<Spinner>` component as it is now could be improved for accessibility.

> The US Government CMS design system renders [Spinner](https://design.cms.gov/components/spinner/?theme=core) as a status with hidden "Loading" text.

[...]

PR #4051, opened 2023-12-11
Adds loading state to ActionList items

mperrotti:
Relates to https://github.com/github/primer/issues/2681

https://github.com/primer/react/assets/2313998/2d15f422-1e1f-428e-b6ac-99e30ef4328b

#### New

- `ActionList.item` accepts `loading` prop

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


mperrotti:
This will be a draft until https://github.com/primer/react/pull/3913 is merged

broccolinisoup (commented):
Hi @mperrotti 👋🏻 I have a question before reviewing the code but let me know if I am missing any context on this or just behind of catching up with things 🙌🏻 

I am curious to know if we have explored possible API design for loading states? Just looking at [Treeview](https://primer.style/components/tree-view/react#treeviewsubtree) for example, the states are handled differently with a `state` prop rather than individual props for each state. I am curious to know if we have an holistic approach in terms of API design for loading states.

mperrotti:
@broccolinisoup - I think TreeView is actually the outlier. Here are the other components where we use a `loading` prop:
- Autocomplete.Menu
- Button/IconButton ([PR](https://github.com/primer/react/pull/3582))
- FilteredActionList
- InlineAutocomplete
- MarkdownViewer
- TextInput
- ToggleSwitch

src/ActionList/Item.tsx
```diff
@@ -266,27 +243,25 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
-              showInactiveIndicator && slots.leadingVisual ? (
-                // using a non-null assertion for `inactiveText` since we check for it in `showInactiveIndicator`
-                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-                <InactiveIndicator labelId={labelId} text={inactiveText!} visualComponent={LeadingVisual} />
-              ) : (
-                // If it's not inactive, just render the leading visual slot
-                slots.leadingVisual
-              )
-            }
+            <VisualOrIndicator
```
broccolinisoup, line 246: When I test this with VoiceOver, action list item and the loading indicator are announce as in totally separate things. I am not sure what would be the solution here, sorry I can't provide more in that regard however I am struggling to understand the connection between the item and its state. 

In [ToggleSwitch](https://github.com/primer/react/pull/3902), the toggle button is described by the loading text when it is in the loading state. I wonder if we need to make such connection here? 🤔

broccolinisoup (commented):
I left some questions mainly around accessibility. Looking forward to hearing your thoughts 🙌🏻 

I also think it would be great to have @TylerJDev review the PR when gets a chance 🤗

src/Spinner/Spinner.tsx
```diff
@@ -12,30 +13,34 @@ const sizeMap = {
 export interface SpinnerInternalProps {
   /** Sets the width and height of the spinner. */
   size?: keyof typeof sizeMap
+  srText?: string
 }
 
-function Spinner({size: sizeKey = 'medium', ...props}: SpinnerInternalProps) {
+function Spinner({size: sizeKey = 'medium', srText = 'Loading', ...props}: SpinnerInternalProps) {
```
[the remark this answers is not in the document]
mperrotti: > First and foremost, I am not 100% sure if the implementation is clear based on https://github.com/primer/react/issues/3942 and https://github.com/github/primer/issues/2203 - There are still https://github.com/primer/react/issues/3942#issuecomment-1815205526 and I am not sure if there is an alignment in terms of the recommended implementation from both issues. Is there something I am missing here?

Great point. I'll bring this to a11y office hours.

---

> There are Spinner [usages (GitHub staff only link)](https://musical-adventure-wlr3n3k.pages.github.io/?=spinner%22&name=%22Spinner%22&attribute=aria-label) that utilizes aria-label and the way we spread the props here ignores the aria-label for visually hidden text. One thing we could do, we can maybe keep using the aria-label for visually hidden text. Seems like people already adopted this and introducing a new prop that serves a similar purpose might lead confusion.

Maybe we could check if `aria-label` has been passed. If it has, we could prevent `srText` from being rendered and log a warning.

src/ActionList/Item.tsx
mperrotti: > When I tested this with VoiceOver, action list item and the loading indicator are announce as totally separate things.

I think this is ok. Let's see what a11y has to say. @alexislucio is this week's a11y FR.

TylerJDev: I think this example is different due to the `Spinner` changes being moved, but I was wondering if there'd be any point in time where there could be more than one item in a loading state? I think if multiple items could be loading at once, we'd want to make the text unique as to convey what exactly is loading.

If only one item can be loading at a given time, then I think there should be enough context on what is loading due to the user triggering the thing that puts it in a loading state.

src/ActionList/Visuals.tsx
```diff
@@ -75,3 +77,56 @@ export const TrailingVisual: React.FC<React.PropsWithChildren<VisualProps>> = ({
...
+    (!itemHasLeadingVisual && position === 'leading') // it doesn't have a leading visual, and it's in the leading position
+  ) {
+    // => so we don't render the indicator here
+    return children
+  }
+
+  return inactiveText ? (
+    <Tooltip text={inactiveText}>
+      <Box
+        as="button"
```
TylerJDev, line 111: Just curious, are we using the `Tooltip` + `Button` combo here as a last option? I know this is an anti-pattern, so I just wanted to confirm!

Also, I see we're not using this pattern for when the `Spinner` is active. Is the difference here being that the `Spinner` state is temporary rather than permanent?

src/ActionList/Visuals.tsx
```diff
@@ -75,3 +77,56 @@ export const TrailingVisual: React.FC<React.PropsWithChildren<VisualProps>> = ({
...
+        sx={{
+          background: 'none',
+          color: 'inherit',
+          border: 'none',
+          padding: 0,
+          font: 'inherit',
+          cursor: 'pointer',
+        }}
+        aria-labelledby={labelId}
```
TylerJDev, line 120: I see we're using the `text` prop without defining the type, so it's defaulting to `description`. I'm thinking it might be better to switch this for the `label` type and having the `aria-labelledby` on the button here be `aria-describedby` instead. This would make the tooltip the accessible name, with the ActionList.Item inner text content being the description (e.g. "{inactiveText}, {ActionList.Item Text}").

TylerJDev (commented):
Awesome work! I just had a few questions, none are blocking!

src/ActionList/Visuals.tsx
mperrotti: > Just curious, are we using the Tooltip + Button combo here as a last option? I know this is an anti-pattern, so I just wanted to confirm!

Yes, this is the direction I got from a11y office hours: https://github.com/github/accessibility/issues/4779

---

> Also, I see we're not using this pattern for when the Spinner is active. Is the difference here being that the Spinner state is temporary rather than permanent?

Correct.

src/ActionList/Visuals.tsx
mperrotti: This sounds counter to the feedback I got during the same a11y office hours I mentioned in your previous comment.

TylerJDev: Thanks for the link :grin: I took a peek at the feedback, and I see that it's mentioned to use the `aria-describedby` version of the `Tooltip`, which I believe would still apply even if we did swap the values of `aria-describedby` and `labelledby`. Right now, the current usage of `aria-labelledby` means that we're relying on the text directly within `ActionList.Item` `<span>` as the accessible name, this causes redundancy since the previous element in the DOM carries the same accessible name.  

(e.g. current example
```html
<div>
  <span id="item-1--label">Action list item text contents</span>
  <button aria-labelledby="item-1--label" aria-describedby="tooltip-:r23:">
    <span><Icon /></span>
  </button>
  <div role="tooltip" id="tooltip-:r23:">Alert contents</div>
</div>
```
)

If a user is navigating through the DOM, "Action list item text contents" will be announced twice, once while on the `<span>` and second when on the `<button>`, which could be redundant. The second potential issue is that is that the purpose of the "alert" button is now secondary to the text within `ActionList.Item#item-(n)--label`, meaning that the text that describes the alert could be missed since it's only referred to with `aria-describedby`. 

I think ideally, we'd want the above example to announce something like this when a user is on the `<button>`: "`Accessible name`: Alert contents, `Accessible description`: Action list item text contents". 

(e.g. suggested example
```html
<div>
  <span id="item-1--label">Action list item text contents</span>
  <button aria-labelledby="tooltip-:r23:" aria-describedby="item-1--label">
    <span><Icon /></span>
  </button>
  <div role="tooltip" id="tooltip-:r23:">Alert contents</div>
</div>
```
)
The above example would remove the redundancy of "Action list item text contents", and ensures that the accessible name of the `<button>` is the actual purpose of the control. Screen reader users will be read the intent of the error message and then will be read what the error message refers to via the accessible description.

Tagging @primer/accessibility-design FR to make sure that this suggestion is aligned with the feedback given during office hours!

src/ActionList/Item.tsx
mperrotti: We discourage showing too many loading indicators on the page at once, but it's certainly possible. I think the most common example (an ActionList with multiple items in a loading state) will be covered by the loading ActionList PR.

If you think we need another example of multiple Spinner loading indicators on the page at once, I'd love some ideas for use cases we can show in Storybook.

mperrotti: Yes, it's possible that multiple items could be loading at once. If we can't even know the title of the list item, it will be difficult to come up with unique text for each loading item. Any ideas how we can convey that? My first thought is "Loading item {n}" or "Loading item {n} under the {section title} section".

src/ActionList/Visuals.tsx
mperrotti: I implemented this pattern and pushed a commit. I can easily revert if needed.

emilybrick (approved):
visual regression test shots lgtm 👍

src/ActionList/Visuals.tsx
TylerJDev: Is the `Tooltip` still present? I'm trying to gain back context from my previous comments, but I was unable to find a story that had active tooltips in `ActionList` 🤔

TylerJDev (commented):
Nothing really blocking, just a few comments!

* I noticed that the "loading" state isn't communicated to AT, I'm assuming the updates to `spinner` will fix this, but just wanted to make note of it.
* We might need to disallow the new `ActionList.TrailingAction` when an item is loading

packages/react/src/ActionList/Visuals.tsx
```diff
@@ -77,3 +80,56 @@ export const TrailingVisual: React.FC<React.PropsWithChildren<VisualProps>> = ({
...
+        aria-describedby={labelId}
+      >
+        <VisualComponent>
+          <AlertIcon />
+        </VisualComponent>
+      </Box>
+    </Tooltip>
+  ) : (
+    <VisualComponent>
+      <Spinner size="small" />
```
TylerJDev, line 132: I see that `<Spinner>` now provides the "loading" text, but `ActionList.Item` isn't referencing it 😭 This is an issue with `ActionList.Item` utilizing `aria-labelledby` and referencing only the container for the visual text.

I think we have a few options we could take:
* Add `<VisuallyHidden>Loading</VisuallyHidden>` in `Item.tsx` that is only present when in loading state
* Add generated ID to `VisualOrIndicator` that is only used when in loading state that [we add to the `aria-labelledby`](https://github.com/primer/react/blob/main/packages/react/src/ActionList/Item.tsx#L335)

packages/react/src/ActionList/LinkItem.tsx
```diff
@@ -21,53 +21,60 @@ type LinkProps = {
...
-      {props.children}
-    </Item>
-  )
-}) as PolymorphicForwardRefComponent<'a', ActionListLinkItemProps>
+    return (
+      <Item
+        active={active}
+        sx={{paddingY: 0, paddingX: 0}}
+        inactiveText={inactiveText}
+        loading={loading}
```
TylerJDev, line 50: Should links be allowed to have a loading state as they'd technically always be active? Are we expecting the link `href` to change during or after loading has completed?

packages/react/src/ActionList/ActionList.features.stories.tsx
```diff
@@ -442,6 +442,20 @@ export const InactiveItem = () => {
   )
 }
 
+// TODO: test inactive _and_ loading - inactive should take precedence
```
TylerJDev, line 445: Do we still need this? 🤔

TylerJDev (commented):
Just a few comments! The only thing I could add is an integration test in Dotcom to see if anything is breaking!

packages/react/src/ActionList/Item.tsx
```diff
@@ -421,19 +396,15 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
-                  // This preserves the left alignment of item text.
-                  showInactiveIndicator && !slots.leadingVisual ? (
-                    <InactiveIndicator labelId={labelId} text={inactiveText!} visualComponent={TrailingVisual} />
-                  ) : (
-                    // If it's not inactive, or it has a leading visual that can be replaced,
-                    // just render the trailing visual slot.
-                    trailingVisual
-                  )
-                }
+                <VisualOrIndicator
```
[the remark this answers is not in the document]
mperrotti: @TylerJDev - good catch! I made an update to hide the trailing action when loading:
```
{!inactive && !loading && !menuContext && Boolean(slots.trailingAction) && slots.trailingAction}
```

packages/react/src/ActionList/Visuals.tsx
mperrotti: I went with the `<VisuallyHidden>` direction

packages/react/src/ActionList/ActionList.features.stories.tsx
mperrotti: No, I just forgot to remove it.

packages/react/src/ActionList/LinkItem.tsx
mperrotti: Good point! We've moved away from loading states for links, and instead we've been advocating for consumers to just hide loading links. I'm going to remove it.

mperrotti:
Updating snapshots since we updated the stories.

packages/react/src/NavList/NavList.stories.tsx
```diff
@@ -225,7 +225,32 @@ export const WithInactiveItems: StoryFn = () => (
   </PageLayout>
 )
 
-export const WithGroup = () => (
+export const WithLoadingItems: StoryFn = () => (
+  <PageLayout>
+    <PageLayout.Pane position="start">
+      <NavList>
+        <NavList.Item href="#" loading>
```
TylerJDev, line 232: Now that we don't have `loading` for links, should this prop & story be removed from `NavList`? I noticed that this story wasn't working, and I think it's because all `NavList.Item`(s) utilize `ActionList.LinkItem` by default.

packages/react/src/ActionList/Item.tsx
```diff
@@ -456,7 +428,8 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
               {slots.blockDescription}
             </Box>
           </ItemWrapper>
-          {!inactive && !menuContext && Boolean(slots.trailingAction) && slots.trailingAction}
+          {!inactive && !loading && !menuContext && Boolean(slots.trailingAction) && slots.trailingAction}
+          {loading === true && <VisuallyHidden>Loading</VisuallyHidden>}
```
TylerJDev, line 432: I think we might need to put this [in the `Box` above](https://github.com/primer/react/blob/main/packages/react/src/ActionList/Item.tsx#L411) so that it is referenced to via `aria-labelledby`, or we could include it in [the `aria-labelledby` list](https://github.com/primer/react/blob/main/packages/react/src/ActionList/Item.tsx#L336) and generate a unique ID for the `VisuallyHidden` component that we could add to the list.

TylerJDev (approved):
I have two suggestions, but otherwise looks good!

packages/react/src/NavList/NavList.stories.tsx
mperrotti: Thanks - I missed this one.

packages/react/src/ActionList/Item.tsx
mperrotti: Ahh ok I understand.

PR #4053, opened 2023-12-12
Use custom aria label for the MD editor

iulia-b:
- Closes https://github.com/github/issues/issues/8856
- Connected to https://github.com/github/github/pull/304978/files#diff-b90fd791c914cc3cdac126aeefd5839a7cc6bfa509a053a686dcb6228aa32303R285
- Reported in https://github.com/github/accessibility-audits/issues/6850

#### Changed

Updated the Markdown editor to use the 'describedBy' property, if specified, for setting aria-label to the `Textarea` component inside markdownEditor. This is the component that is focused by screenreaders and announced

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Testing & Reviewing

Replacing "Markdown value" with a custom value, provided by the consumer 

image

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/drafts/MarkdownEditor/_MarkdownInput.tsx
```diff
@@ -149,7 +151,8 @@ export const MarkdownInput = forwardRef<HTMLTextAreaElement, MarkdownInputProps>
           value={value}
           onKeyDown={onKeyDown}
           disabled={disabled}
-          aria-label="Markdown value"
+          aria-label={describedBy ? undefined : 'Markdown value'}
+          aria-describedby={describedBy}
```
siddharthkp, line 155: If there is a description, it feels odd that we remove the label. What's the thought process behind that?

siddharthkp (commented):
Left a comment for clarification, also added @iansan5653 for a review

src/drafts/MarkdownEditor/_MarkdownInput.tsx
iulia-b: The prop, which is coming from `MarkdownEditor`, is being used for the `fieldset`'s `aria-desbribedBy`. But this is not read by screenreader when tabbing on the page. Issue's comment box, for ex, is using this field to add an accessible name  to the Textarea, which is missing atm. There is a report after the a11y  audit mentioning that. 'MD value' is not a good accessible name. I don't want to change too much in the API of the MDEditor  so i am using the `describedBy` arg, if existing, to describe the text area. Should i use **labelledBy** instead?

I should have added this to the PR description, will do so in a moment.

siddharthkp: Based on the updated PR description, 

> Updated the Markdown editor to use the 'describedBy' property, if specified, for setting aria-label to the Textarea component inside markdownEditor. This is the component that is focused by screenreaders and announced

If your goal is to set a label, my recommendation would be to `aria-label` because `aria-describedby` wouldn't do the job. Voiceover on Mac will use the placeholder as label (I'm not sure if this behavior is consistent across screen readers) and only announce the description based on the user's verbosity settings. (it does not for me)

> I don't want to change too much in the API of the MDEditor so i am using the describedBy arg, if existing, to describe the text area.
> There is a report after the a11y audit mentioning that. 'MD value' is not a good accessible name. 

That is fair. However, looking at the current default of "markdown value", it seems better to add an `aria-label` prop (can make it optional to avoid breaking changes) and then fallback to the default browser behavior of using placeholder as label instead of "markdown value". 

@iansan5653 what do you think about this solution? ^

--- 

> But this (fieldset description) is not read by screenreader when tabbing on the page.

Yes, because it is not an interactive element, but you can still use a screen reader to reach the fieldset. However, doing some digging, description is not consistently read for static elements [across screen readers](https://w3c.github.io/using-aria/#p10) :( 

As an alternate to adding description to textarea, we have the option of changing the fieldset description to a label. I'm not sure which one of these is preferred, getting some guidance from the accessibility team would be useful here!

iulia-b:
Closing this as discussed in https://github.slack.com/archives/C01L618AEP9/p1702477860598739?thread_ts=1702462838.612959&cid=C01L618AEP9 . Wainting for the MDEditor to be migrated to gh/gh and will contribute there

PR #4046, opened 2023-12-11
IconButton: Add tooltip to icon buttons by default

broccolinisoup:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #4056, opened 2023-12-13
Tooltip: Do not wrap the tooltip and the trigger in a div

broccolinisoup:
#### Changed

Removed the div wrapper because it was causing styling issues on the icon buttons ([WIP PR](https://github.com/primer/react/pull/4046)) and added mouse enter and leave events to the tooltip to make sure it stays open when it is hovered. I don't think we need it but let me know if I am missing anything.

```diff
-<div> 👈 wrapper 
  <button>tooltip trigger</button>
  <div>tooltip itself</div>
-</div>
```

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/ActionList/Item.tsx
+++ b/src/ActionList/Item.tsx
@@ -34,6 +34,7 @@ const InactiveIndicator: React.FC<{
         padding: 0,
         font: 'inherit',
         cursor: 'pointer',
+        display: 'flex',
       }}
       aria-labelledby={labelId}
     >
--- a/.changeset/empty-starfishes-relate.md
+++ b/.changeset/empty-starfishes-relate.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Tooltip2: Do not wrap the tooltip span and its trigger in a div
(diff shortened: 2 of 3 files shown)


src/ActionList/Item.tsx
```diff
@@ -34,6 +34,7 @@ const InactiveIndicator: React.FC<{
         padding: 0,
         font: 'inherit',
         cursor: 'pointer',
+        display: 'flex',
```
broccolinisoup, line 37: Due to removing the wrapper in the tooltip, I needed to add this to align the inactive visual in center. Let me know if you prefer a different way to do it @mperrotti

siddharthkp (approved):
Definitely makes sense to avoid an extra wrapper!

Am I right in assuming that because tooltip is now in top layer, removing this wrapper makes no difference to positioning?

broccolinisoup:
>Am I right in assuming that because tooltip is now in top layer, removing this wrapper makes no difference to positioning?

Yes! Because we use popover, tooltip is always on the top layer. Also because we use [anchor positioning](https://github.com/primer/behaviors/blob/92e06b516eb3751bcf864c3678fd2722df885f4f/src/anchored-position.ts#L1) to set top and left for the tooltip div, it makes no difference how it is placed in the dom. The tests and stories are looking too ✨

PR #4023, opened 2023-12-05
Use `aria-required` instead of `required` on required form elements

mperrotti:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #4058, opened 2023-12-13
Let aria-required be undefined if not passed

mperrotti:
Follow up to https://github.com/primer/react/pull/4023

Does not pass `aria-required` attribute to input or textarea if it is undefined. This fixes some tests that were breaking in dotcom.

#### Changed

Before:
`aria-required` is set to `"false"` if the `required` prop is not passed to `TextInput` or `Textarea` .

After:
`aria-required` is not set at all if the `required` prop is not passed to `TextInput` or `Textarea` .

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/TextInput/TextInput.tsx
```diff
@@ -144,7 +144,7 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(
           onFocus={handleInputFocus}
           onBlur={handleInputBlur}
           type={type}
-          aria-required={required ? 'true' : 'false'}
+          aria-required={required}
```
mperrotti, line 147: We still use `aria-required={required ? 'true' : 'false'}` on Radio and Checkbox. Should we update those as well? Or are we worried this could also break tests in dotcom?

broccolinisoup, line 147: Just a quick question do we want to have `aria-required="false"` if required is passed but as false? (Example [usage](https://github.com/github/internal-statuspage/blob/30b1e7bc7a53daeec20e4c802578007e3654f9ae/frontend/src/incidents/DeclareIncidentDialog/DeclareIncidentForm.tsx#L82C9-L82C25))

Or do we only want to have `aria-required` when it is only true?

broccolinisoup: What are the broken tests?? 👀

siddharthkp: > Or do we only want to have aria-required when it is only true?

yes :)

This is the failing test: https://github.com/github/github/actions/runs/7202649701/job/19621310998?pr=304951

siddharthkp: There are some [tests in dotcom that compare the dom](https://github.com/github/github/actions/runs/7176291085/job/19540971389), so any additional props even when false fail CI 🤦

Ideally they should not depend on attributes from within primer components, but it's common :(

siddharthkp: > Or are we worried this could also break tests in dotcom?

Only one way to find out 😓 Going to run the integration tests for this PR

siddharthkp (approved):
Looks good, thanks!

PR #4062, opened 2023-12-14
Allow `Button` to break-word for extra long labels (edge cases)

langermank:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
I have a path forward for preserving the top and bottom padding for buttons with wrapping lines of text.

We can set the following styles on the "buttonContent" container element:
```
align-self: stretch;
padding-block: calc(var(--control-{size}-paddingBlock, {fallback}) - 2px)
```

These styles force `.buttonContent` to fill the height of the button, then adds the correct amount of padding minus the top and bottom borders.

mperrotti:
Talked to Katie about my idea on Slack and got the go-ahead to push up changes.

mperrotti (approved):
Idk if I can approve these changes since I pushed a commit, but I'm trying it anyway.

@langermank - we just need a changeset.

pksjce (commented):
Hi 👋  Thanks for the changes @langermank!
As far as I understand this change is a potential fix for https://github.com/github/accessibility-audits/issues/4690

On deeper investigation, I found that the particular bug is probably a rails issue because it looks like [this is the source code](https://github.com/github/github/blob/b53d706fce358cd68e48ca56b6cad0845125476d/app/views/settings/account_preferences/show.html.erb#L39) for the bug mentioned in the issue. So I don't think this particular PR solves any reported issues.

Other things to consider are
1. This could potentially be an issue in PRC too. Esp with a `Button` in a container like a `Dialog`. We can see it is broken in the `Dialog` stories which currently uses the deprecated `Button` component. We'd need to refactor it to the latest `Button` to confirm this. 
2. If a `Button` is present directly on a page, it's not an issue because even on the highest zoom, the page will comfortably overflow and the user will be able to scroll across to view all of the button text (unlike in a dialog where it simply gets cut off).
3. A cursory investigation in dotcom shows that buttons can have long and dynamic texts frequently. I wonder how this PR affects them. Do you think we can get some input from product devs to get an insight into this? Do you know whom we can add to this review?

src/Button/styles.ts
```diff
@@ -247,14 +247,19 @@ export const getBaseStyles = (theme?: Theme) => ({
...
+    width: 'var(--control-medium-size, 32px)',
     minWidth: 'unset',
   },
   '&[data-size="small"]': {
-    padding: '0 8px',
-    height: '28px',
+    minHeight: 'var(--control-small-size, 28px)',
     gap: '4px',
     fontSize: '0',
+    paddingInline: 'var(----control-small-paddingInline-condensed, 0.5rem)',
```
langermank, line 257:
```suggestion
    paddingInline: 'var(--control-small-paddingInline-condensed, 0.5rem)',
```

PR #4083, opened 2023-12-18
Flash child SVG fill color fix

mperrotti:
Closes https://github.com/github/primer/issues/2799

⚠️ This is a breaking change because it will remove the fill colors for SVGs that are passed as children of Flash without specifying the correct `fill` color(s) on the icon component. ⚠️

Before:
Screenshot 2023-12-18 at 3 13 36 PM

After:
Screenshot 2023-12-18 at 3 12 42 PM

#### New

Adds a `renderDefaultVisual` prop to `Flash` that will render the correct icon with the correct fill color depending on what is passed to the `variant` prop.

#### Changed

Prevents all SVG children of Flash from getting an automatic fill color

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

#### Roll-out plan:
This is a low-risk change as long as we update existing uses of `Flash` that manually pass one of the variant-specific leading visuals.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mperrotti:
> ⚠️ This is a breaking change because it will remove the fill colors for SVGs that are passed as children of Flash without specifying the correct fill color(s) on the icon component. ⚠️

Does anybody have any ideas for resolving this bug _without_ introducing breaking changes?

---

If we have no way around introducing a breaking change, what should I set the parent branch to?

joshblack:
@mperrotti just wanted to clarify, are there scenarios where folks would want all SVGs in the banner to be styled in this way or do you think it's always a bug?

I think if it's always considered a bug then we could safely ship this, potentially, since there wouldn't be a case that we would need to cover. However, if there are cases where folks would want that then I think we just need to do a migration plan and go from there.

src/Flash/Flash.tsx
```diff
@@ -46,6 +47,7 @@ const variants = variant({
 type StyledFlashProps = {
   variant?: 'default' | 'warning' | 'success' | 'danger'
   full?: boolean
+  renderDefaultVisual?: boolean
```
pablonete, line 50: This name seems hard to discover and understand. Maybe `renderLeadingIcon`?

src/Flash/Flash.tsx
```diff
@@ -61,7 +63,7 @@ const StyledFlash = styled.div<StyledFlashProps>`
     margin-bottom: 0;
   }
 
-  svg {
+  [data-component='leadingVisual'] {
```
pablonete, line 66: This will break the layout of existing usages of Flash where icons relied on this style.
What's the plan to smoothly roll this out? Are we expected to fixing all Flash usages across the consumer codebase at the same time that we upgrade to this version? If so, how do we track this?

src/Flash/Flash.tsx
```diff
@@ -71,8 +73,26 @@ const StyledFlash = styled.div<StyledFlashProps>`
 
 export type FlashProps = ComponentProps<typeof StyledFlash>
 
-const Flash = React.forwardRef(function Flash({as, variant = 'default', ...rest}, ref) {
-  return <StyledFlash ref={ref} as={as} variant={variant} {...rest} />
+const Flash = React.forwardRef(function Flash({as, children, variant = 'default', renderDefaultVisual, ...rest}, ref) {
+  const defaultLeadingVisuals = {
```
pablonete, line 77: Nit: should we move this const object out to the module level?

pablonete (commented):
The fix makes sense to me, I have questions about its rollout process.

mperrotti:
> just wanted to clarify, are there scenarios where folks would want all SVGs in the banner to be styled in this way or do you think it's always a bug?

@joshblack - if somebody is manually passing the correct icon as a leading visual but doesn't pass a `fill` prop, I think they'd want it to be the color set by the component styles.

It's not a big bug, but it's unexpected.

src/Flash/Flash.tsx
mperrotti: I've marked this as a breaking change, so it wouldn't go out until the next major release. We'd either handle the migration in dotcom ourselves or instruct folks on how to handle it themselves.

src/Flash/Flash.tsx
mperrotti: Yea, I like that

mperrotti:
Closing this PR until we get closer to v37

PR #4089, opened 2023-12-19
test(e2e): use @prettier/sync for e2e tests

joshblack:
Uses `@prettier/sync` instead of `prettier` for formatting as the `format` method is now async in later versions of `prettier`.

#### Changed

- Update `script/generate-e2e-tests.js` to use `@prettier/sync`

### Rollout strategy

### Testing & Reviewing

- Verify you can run this script locally 😅

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@primer/react",
-  "version": "36.4.0",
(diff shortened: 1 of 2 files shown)


langermank (approved):
Nice!

PR #4092, opened 2023-12-20
Only show tooltip on :focus-visible, not :focus

siddharthkp:
- Closes https://github.com/github/primer/issues/2926


siddharthkp:
- Part of https://github.com/primer/react/pull/4224 now

PR #4095, opened 2023-12-20
Add internal `ConditionalWrapper` component

langermank:
I have a use case for this in `Button`, and we have this [same component](https://github.com/primer/view_components/blob/main/app/components/primer/conditional_wrapper.rb) in PVC.

I renamed it to match up with PVC, but no changes beyond that. I've also tested it already in #3582.


mperrotti (approved):

joshblack:
I think one downside of this technique that wouldn't exist if using ternaries or refactoring into smaller components is that things are evaluated even if the condition is false-y.

So when we have code like:

```tsx
<ConditionalWrapper if={conditional} sx={{ /* ... */ }}>
  <ChildComponent sx={{ /* ... */ }}>
    <SomeOtherChild />
  </ChildComponent>
</ConditionalWrapper>
```

Things like the `sx` props are being generated along with anything that are `children` within the component. This can make some types of code awkward if you're relying on something from the `if` block to be defined for a child component and TS won't properly narrow like it would when using ternaries or refactoring into sub-components.

```tsx
<ConditionalWrapper if={hasTheThing}>
  {/* this will be evaluated even if `hasTheThing` is false-y */}
  {/* it also won't be narrowed so it will require a guard */}
  <SomeComponent>{hasTheThing.title}</Subcomponent/>
</ConditionalWrapper>
```

I think with this in mind I'd love to encourage alternatives that we could use instead that wouldn't have these limitations. Ternaries are definitely something that come to mind but they are so inconvenient/hard to understand with a lot of nesting. In those cases, it may be worth us trying to refactor things into sub-components that are local to the file to help avoid the challenges above while still being able to reason about why things are rendering.

langermank:
Putting this on pause as we look into alternative solutions ✨

mperrotti:
Dismissing my review until we agree on a path forward from [this discussion](https://github.com/github/primer/discussions/2942)

mperrotti:
It's worth mentioning that we use a similar component in `ActionList.Item`: https://github.com/primer/react/blob/main/src/ActionList/Item.tsx#L353

mperrotti (approved):

PR #4096, opened 2023-12-20
Fix missing selection attributes in ActionList items

strackoverflow:
Addresses the following a11y audit issue: https://github.com/github/accessibility-audits/issues/4484 (GitHub staff only)

This PR fixes an accessibility issue with `ActionList` items not reporting their selection state to screen readers. 

The `aria-selected` (or `aria-checked`) attributes were not being applied to `ActionList.Item` because they were only being read from `ActionListContainerContext`, but `ActionList` instances don't provide that context.

The solution I chose here was to automatically infer the selection attribute type based on the item's role: `menuitemcheckbox` and `menuitemradio` roles get an inferred selection attribute of `aria-checked`, while `option` roles will get `aria-selected`. 

Of course, if a selection attribute is specified through `ActionListContainerContext`, that will be preferred.

Interestingly, selection state was [asserted](https://github.com/primer/react/blob/c2a53a003fb392b97cd33aa5eea0329e4f726874/src/ActionList/ActionList.test.tsx#L95-L96) in one of the component tests, but the `aria-selected` attribute was [manually provided](https://github.com/primer/react/blob/c2a53a003fb392b97cd33aa5eea0329e4f726874/src/ActionList/ActionList.test.tsx#L47) in the test stub. Removing that allows the test to assert the actual selection behavior.

## Demo

This screenshot demonstrates the NVDA screenreader properly announcing the selection state with this change:

> screenshot of NVDA properly announcing selection state

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/src/__tests__/__snapshots__/Autocomplete.test.tsx.snap
+++ b/src/__tests__/__snapshots__/Autocomplete.test.tsx.snap
@@ -789,6 +789,7 @@ exports[`snapshots renders a menu that contains an item to add to the menu 1`] =
       >
         <li
           aria-labelledby="0--label "
+          aria-selected={false}
           className="c2 c3"
           data-id="0"
           id="0"
@@ -819,6 +820,7 @@ exports[`snapshots renders a menu that contains an item to add to the menu 1`] =
         </li>
         <li
           aria-labelledby="1--label "
+          aria-selected={false}
           className="c2 c3"
           data-id="1"
           id="1"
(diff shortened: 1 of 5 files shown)


src/__tests__/__snapshots__/Autocomplete.test.tsx.snap
```diff
@@ -789,6 +789,7 @@ exports[`snapshots renders a menu that contains an item to add to the menu 1`] =
       >
         <li
           aria-labelledby="0--label "
+          aria-selected={false}
```
strackoverflow, line 792: It looks like Autocomplete might have had a similar issue with list items not having `aria-selected` attributes, and this fix applies here as well.

joshblack (approved):

Issue #4091, opened 2023-12-20
TextInput.Action tooltip is clipped


broccolinisoup:
### Description

Not sure what update cause this but currently with the default direction of the tooltip, the text is not visible because it is clipped. We should update it to use Tooltip v2 and also add a VRT test for this story.

![Text input screenshot with the trailing action is in focus and only the tooltip's caret is visible and the text of the tooltip is clipped](https://github.com/primer/react/assets/1446503/9edf4900-89ac-4b47-925a-4bfeb106b7c8)

### Steps to reproduce

1. Go to https://primer.style/react/storybook/?path=/story/components-textinput-features--with-trailing-action
2. Focus on the `X` (trailing action) or hover over it
3. Observe that the tooltip appears but it is clipped

### Version

v36.4.0

lesliecdubs:
Assigning to @broccolinisoup and moving to "In Progress" since I saw an open PR: https://github.com/primer/react/pull/4097

PR #4097, opened 2023-12-21
TextInput: Update TextInput.Action internal component to use Tooltip v2

broccolinisoup:
Closes https://github.com/primer/react/issues/4091

#### Changed

- Update to use the new tooltip for TextInput's trailing action
- Update the default direction of `tooltipDirection` prop to `s` because it is what is defaulted in Tooltip

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Changes can be tested in the storybook

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/internal/components/TextInputInnerAction.tsx
```diff
@@ -93,7 +93,7 @@ const TextInputAction = forwardRef<HTMLButtonElement, TextInputActionProps>(
     return (
       <Box as="span" className="TextInput-action" marginLeft={1} marginRight={1} lineHeight="0">
         {icon && !children ? (
-          <Tooltip direction={tooltipDirection} aria-label={ariaLabel}>
+          <Tooltip direction={tooltipDirection} text={ariaLabel as unknown as string} type="label">
```
broccolinisoup, line 96: `text={ariaLabel as unknown as string}` this is not great but some sort of type conversion or "hack" is needed here because `text` is a required prop for Tooltip but I see we can't make `ariaLabel` required for TextInput's trailing action because it could be an icon button or a text button. I am open to feedback if we want to handle this in another way

mperrotti: Would this be a little easier to read/understand?

```jsx
text={ariaLabel || ''}
```

src/TextInput/TextInput.docs.json
```diff
@@ -128,7 +128,7 @@
         {
           "name": "tooltipDirection",
           "type": "'n' | 'ne' | 'e' | 'se' | 's' | 'sw' | 'w' | 'nw'",
-          "defaultValue": "n",
+          "defaultValue": "s",
```
mperrotti, line 131: This might be the default on `Tooltip2`, but I think we should explicitly set `'s'` as the default value of `tooltipDirection` in the `TextInput.Action` component.

This way the docs will always be accurate, even if we decide to change the default direction in `Tooltip2`.

broccolinisoup: >This might be the default on Tooltip2, but I think we should explicitly set 's' as the default value of tooltipDirection in the TextInput.Action component.

No worries, that sounds good. I'll update it now.

src/internal/components/TextInputInnerAction.tsx
broccolinisoup: Yeah that reads better. Thank you. Although both options will end up rendering an empty tooltip. I wonder if we should conditionally render the tooltip for icon button too? Though this solution will only work until we bring the tooltip to icon buttons by default https://github.com/primer/react/issues/2008. What are your thoughts on that?

src/internal/components/TextInputInnerAction.tsx
```diff
@@ -93,7 +93,7 @@ const TextInputAction = forwardRef<HTMLButtonElement, TextInputActionProps>(
     return (
       <Box as="span" className="TextInput-action" marginLeft={1} marginRight={1} lineHeight="0">
         {icon && !children ? (
-          <Tooltip direction={tooltipDirection} aria-label={ariaLabel}>
+          <Tooltip direction={tooltipDirection ?? 's'} text={ariaLabel ?? ''} type="label">
```
joshblack, line 96: Would this be something that we'd want to restrict in v37 so that `aria-label` is a required prop? I think this is fine since it mirrors the current behavior but in a future major release it'd be great to have this required so we don't have a non-descriptive tooltip

broccolinisoup: >Would this be something that we'd want to restrict in v37 so that aria-label is a required prop?

I think we could. The only thing I can think of a bit of a blocker is that the `TextInput.Action` also accepts children and if children is passed, the tooltip becomes optional. (Based on just reading the code). I guess the case would be a text button as trailing action that doesn't need a tooltip. ([related additional comment](https://github.com/primer/react/pull/4097#discussion_r1433327122))

Just looking at the [current use cases at dotcom](https://musical-adventure-wlr3n3k.pages.github.io/?name=textinput.action), all seem to have icons. I wonder if it would make sense to re-visit the text button use case? 🤔 Or do we have clear uses cases for this? @mperrotti

src/internal/components/TextInputInnerAction.tsx
mperrotti: Ah ok. Yea, I think we can keep what you have then.

src/internal/components/TextInputInnerAction.tsx
mperrotti: I don't think trailing _actions_ need to accept text, just the trailing visuals. For example: we may use trailing visual text to convey units like "minutes". See the timeout settings in https://github.com/settings/codespaces for an example.

mperrotti (approved):

src/internal/components/TextInputInnerAction.tsx
broccolinisoup: This is great. We can then refactor this conditional tooltip and make `aria-label` mandatory in the next major. I'll add this to the v37 issue.

PR #4125, opened 2024-01-08
ci(deploy): update deploy production workflow to use latest hash from shared workflow

joshblack:
Our production deployment workflow is broken as a result of the changes to actions/deploy-pages before break. This PR updates the hash used of our shared workflow to pickup the latest sets of changes that should address this.

#### Changed

- Update hash to shared deploy workflow in `primer/.github`

### Rollout strategy

### Testing & Reviewing

This one will likely require us to manually call the workflow_dispatch event after merge in order to test this out and see if it fails again.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -51,7 +51,7 @@ jobs:
     name: Production
     needs: [guard]
     if: ${{ needs.guard.outputs.should_deploy == 'true' }}
-    uses: primer/.github/.github/workflows/deploy.yml@0cec9b9914f358846163f2428663b58da41028c9
+    uses: primer/.github/.github/workflows/deploy.yml@bb0b18fb589038a93a49a24624be0559c01ed0b9
     secrets:
       gh_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


camertron (approved):
Thanks!

pksjce (approved):

.github/workflows/deploy_production.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     name: Production
     needs: [guard]
     if: ${{ needs.guard.outputs.should_deploy == 'true' }}
-    uses: primer/.github/.github/workflows/deploy.yml@0cec9b9914f358846163f2428663b58da41028c9
+    uses: primer/.github/.github/workflows/deploy.yml@bb0b18fb589038a93a49a24624be0559c01ed0b9
```
pksjce, line 54: How did you find this hash?

joshblack: @pksjce honestly just looking at the latest commit in the repo 😂 

It'd be great if we could tag our stuff in `primer/.github` so that we don't have to use hashes and so that dependabot can pick up on updates

PR #4127, opened 2024-01-08
Update Link `inline` prop description

dipree:
https://github.com/github/primer/issues/2099

Updates the Link `inline` property description and set the `underline` property to "deprecated".

### DIFF
--- a/.changeset/nice-seals-decide.md
+++ b/.changeset/nice-seals-decide.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Deprecate the `underline` property of the Link component in favor of the new `inline` property to better handle link visibility and accessibility when adjacent to text.
--- a/src/Link/Link.docs.json
+++ b/src/Link/Link.docs.json
@@ -15,25 +15,26 @@
       "name": "muted",
       "type": "boolean",
       "defaultValue": "false",
-      "description": "Uses a less prominent shade for Link color, and the default link shade on hover"
+      "description": "Uses a less prominent shade for Link color, and the default link shade on hover."
     },
     {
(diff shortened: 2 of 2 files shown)


mperrotti (approved):

joshblack:
Just wanted to leave a couple notes for the deprecation process that I believe we're hoping to follow:

- If possible, the deprecated item should have `/** @deprecated */` as an annotation with a corresponding notice for what to use instead
- The deprecated item should be marked as deprecated in the corresponding docs page
- When a prop is deprecated, it would fall under a `minor` semver doc per the [versioning guide](https://github.com/primer/react/blob/main/contributor-docs/versioning.md#:~:text=A%20prop%20is%20deprecated)
- If possible, we should have a lint rule to manage deprecations (and offer fixes) over in: https://github.com/primer/eslint-plugin-primer-react

What would be a good place for us to include/document this info so that it's more apparent when making deprecation changes? Would that versioning guide be a good place? Just want to make sure we've captured all of these things so that things don't slip through the cracks 👀

dipree:
@joshblack what's the best to do now? Revert and merge again as `minor` with and additional note for "what to use instead" and adding a `/** @deprecated */` [here](https://github.com/primer/react/blob/c21b939701240ccc93225dd53122c616c24dc3bd/src/Link/Link.tsx#L32)?

I had a quick look at [this](https://github.com/primer/react/blob/main/contributor-docs/deprecating-components.md) but then just looked at the options available and used the `deprecated` option.

dipree:
Revert: https://github.com/primer/react/pull/4134
Follow-up: https://github.com/primer/react/pull/4135

PR #4131, opened 2024-01-09
SelectPanel2: Fix bug calling onCancel

siddharthkp:
Bug: Calling `onSubmit` calls `internalClose` which calls `onCancel` 🐞! 

This PR fixes that by separating `internalClose` and `internalCancel`


src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -114,13 +114,18 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+  }
+
+  const onInternalCancel = () => {
+    onInternalClose()
     if (typeof propsOnCancel === 'function') propsOnCancel()
   }
 
   const onInternalSubmit = (event?: React.FormEvent<HTMLFormElement>) => {
     event?.preventDefault() // there is no event with selectionVariant=instant
-    if (propsOpen === undefined) setInternalOpen(false)
```
siddharthkp, line 123: don't worry, this line is part of `onInternalClose`

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -150,13 +155,32 @@ const Panel: React.FC<SelectPanelProps> = ({
 
   /* Dialog */
   const dialogRef = React.useRef<HTMLDialogElement>(null)
-  if (internalOpen) dialogRef.current?.showModal()
-  else dialogRef.current?.close()
+
+  // sync dialog open state (imperative) with internal component state
+  React.useEffect(() => {
```
siddharthkp, line 160: Moved these statements into an effect to avoid conflict with the subsequent lines in the component in the same render

dialog was closing and then the other lines in the component would try to access dialogRef.current which throw an error on the console :)

mperrotti (approved):

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -150,13 +155,32 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+  }, [internalOpen])
 
   // dialog handles Esc automatically, so we have to sync internal state
-  React.useEffect(() => dialogRef.current?.addEventListener('close', onInternalClose))
+  // but it doesn't call onCancel, so have another effect for that!
+  React.useEffect(() => {
+    const dialogEl = dialogRef.current
+    dialogEl?.addEventListener('close', onInternalClose)
+    return () => dialogEl?.removeEventListener('close', onInternalClose)
+  })
```
broccolinisoup, line 171: Do we need these effects to run every render? 🤔

src/drafts/SelectPanel2/SelectPanel.test.tsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react'
+import {ThemeProvider, ActionList} from '../../'
+import {render, RenderResult} from '@testing-library/react'
+import userEvent, {UserEvent} from '@testing-library/user-event'
```
broccolinisoup, line 4: Oo UserEvent type, I've never used it before 🔥

siddharthkp: It's really nice!

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Good question! I'll double check

broccolinisoup (approved):

PR #4135, opened 2024-01-09
Update Link `inline` prop description and deprecate `underline`

dipree:
https://github.com/github/primer/issues/2099

Updates the Link `inline` property description and set the `underline` property to "deprecated".


src/Link/Link.tsx
```diff
@@ -29,6 +29,7 @@ const StyledLink = styled.a<StyledLinkProps>`
   text-decoration: none;
 
   /* You can add one by setting underline={true} */
+  /** @deprecated */
```
joshblack, line 32:
```suggestion
```

I think the notice could live on `StyledLinkProps` above like:

```tsx
type StyledLinkProps = {
  hoverColor?: string
  muted?: boolean
  /** @deprecated use `inline` to specify the type of link instead */
  underline?: boolean
  // Link inside a text block
  inline?: boolean
} & SxProp
```

Hope that makes sense! This should mean that folks get this kind of error in their IDE of choice:

An editor showing the deprecation message for the underline prop in a code snippet

joshblack (changes requested):
Thanks for doing this! Just left a comment on where `@deprecated` could live, hope it makes sense and let me know if you have any questions!

dipree:
@joshblack I see how that works now, thanks and updated.

joshblack (approved):

Issue #4137, opened 2024-01-10
Remove `StyledOcticon` reference in `Tooltip.mdx`


joshblack:
> [!NOTE]
> This issue was first reported by @DavidMeu [over in Slack ](https://github.slack.com/archives/CSGAVNZ19/p1704899157705119)

This component was deprecated in v35 and removed in v36 but is still being referenced in [this example](https://primer.style/react/drafts/Tooltip#as-a-label). We should remove this from this page in our docs along with updating the side nav to be in alphabetical order (Octicon is currently in the "S" grouping in the nav)

### Checklist

- [ ] Remove `StyledOcticon` from `Tooltip.mdx`
- [ ] Update side nav to have alphabetical order with `Octicon`

broccolinisoup:
Hi @joshblack 👋 Since you have an open PR, I assigned this issue to you but let me know if you have any concern!

PR #4138, opened 2024-01-10
docs: update StyledOcticon references in docs

joshblack:
Closes https://github.com/primer/react/issues/4137

#### Changed

- Change `StyledOcticon` to `Octicon` in `Tooltip.mdx`
- Update side nav in docs to be alphabetical

### Rollout strategy

### Testing & Reviewing

- Verify in the deploy preview that the side nav has been updated
- Verify in the deploy preview that the Tooltip example renders as expected


DavidMeu (approved):
Thank you for pushing this.

Small concerns:
1. Maybe it is good to also rename `StyledOcticon` --> `Octicon` [in the component](https://github.com/primer/react/blob/main/src/Octicon/Octicon.tsx).

2. Also I'm not sure if we need to duplicate this component with the `StyledOcticon` name and put it together with [all the other depreacted components](https://github.com/primer/react/tree/main/src/deprecated).
(Maybe because it's only renaming it is ok not to have it there 🤔)

broccolinisoup (approved):

PR #4140, opened 2024-01-10
Convey Spinner to assistive technologies

mperrotti:
Closes https://github.com/primer/react/issues/3942

#### New

- Spinner accepts `srText` prop (default value is "Loading"

#### Changed

- Deprecated `aria-label` on Spinner
- Restricts Spinner to only accept specified props, not just any HTML attribute

### Rollout strategy

- [ ] Patch release
- [x] Minor release

When we ship the next major release, we can remove `aria-label` from Spinner and provide a codemod to change it to `srText`.

### Merge checklist

- [x] Added/updated tests
- [unsure] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Spinner/Spinner.tsx
```diff
@@ -1,45 +1,57 @@
...
-      <path
-        d="M15 8a7.002 7.002 0 00-7-7"
-        stroke="currentColor"
-        strokeWidth="2"
-        strokeLinecap="round"
-        vectorEffect="non-scaling-stroke"
-      />
-    </svg>
+    /* inline-flex removes the extra line height */
+    <Box sx={{display: 'inline-flex'}} role={hasSrAnnouncement ? 'status' : undefined}>
```
TylerJDev, line 30: Would we want to add the `role` value as a prop instead of a static value? If so, we would only have two values to pick from, `status` and `alert`, with `status` being the default value.

mperrotti:
I'm converting this back to a draft until we decide on a direction based on feedback in this discussion: https://github.com/github/primer/discussions/2976

mperrotti:
The aria-live issues should be resolved by https://github.com/primer/react/pull/4174

TylerJDev:
Hey @mperrotti 👋!  With [`live-region-element`](https://github.com/primer/react/blob/main/packages/react/src/internal/components/Announce.tsx) being available, do you think this PR is good to go?

lindseywild:
@mperrotti Hi! Just wanted to check in on the status of this PR 🙏🏻

mperrotti:
Sorry @lindseywild - this PR got lost and stale. Going to update today.

Related PVC PR: https://github.com/primer/view_components/pull/2884

mperrotti:
@lindseywild @TylerJDev @joshblack - should I remove the announcements and follow @owenniblock's PVC implementation until we've figured out how to handle multiple announcements?

[Comment from the PVC PR](https://github.com/primer/view_components/pull/2884#discussion_r1630926078):

> I would not feel comfortable if the Spinner announced by default, I think that's too risky given the number of Spinners we have in the wild already


[Thread in the ADR PR](https://github.com/primer/react/pull/4611#discussion_r1607307132)

TylerJDev:
> should I remove the announcements and follow @owenniblock's PVC implementation until we've figured out how to handle multiple announcements?

I think that's fair. Unless we can make it opt-in, where the live region only exists if the consumer turns it on themselves via a prop. Even without the live region, this PR will fix the accessibility issue on there being no context by default on the `<Spinner>` component.

packages/react/src/Spinner/Spinner.tsx
```diff
@@ -1,46 +1,62 @@
...
+          d="M15 8a7.002 7.002 0 00-7-7"
+          stroke="currentColor"
+          strokeWidth="2"
+          strokeLinecap="round"
+          vectorEffect="non-scaling-stroke"
+        />
+      </svg>
+      {hasSrAnnouncement ? (
+        <VisuallyHidden>
+          <Status>{srText || ariaLabel}</Status>
```
TylerJDev, line 52: Per https://github.com/primer/react/pull/4140#issuecomment-2154532654, we could make this opt-in and keep the live region, but I'll leave this up to you! It would probably involve an additional prop to toggle the live region being on/off with the default being off. We could also remove it and come back to it in the future, but there might be value in having it in now as opt-in.

TylerJDev (approved):
Left some comments, but this looks good even if we remove the live region!

mperrotti:
@TylerJDev - after talking to @joshblack, I decided to remove the announcement until we land on a solution for using `Status` in components without accidentally triggering too many announcements.

PR #4143, opened 2024-01-12
docs: update versioning guide with prop spread

joshblack:
Add a section to our versioning docs to handle a scenario for changing the element onto which props are spread.

#### Changed

- Add a section to our versioning docs to handle a scenario for changing the element onto which props are spread.

### Rollout strategy

### Testing & Reviewing

Would love feedback around having this case in our versioning guide 👀 

- Is this a situation worth documenting in our versioning guide?
- Does this feel like an appropriate version bump for this situation?
- Do the examples clearly convey why the version bump is what it is?
- Are there any scenarios that are missing that should be added?

### DIFF
--- a/contributor-docs/versioning.md
+++ b/contributor-docs/versioning.md
@@ -5,14 +5,15 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 ## Table of Contents
 
-* [Overview](#overview)
-* [Changes](#changes)
-* [Reference](#reference)
-  * [The type of a prop is broadened](#the-type-of-a-prop-is-broadened)
-  * [The type of a prop is narrowed](#the-type-of-a-prop-is-narrowed)
-  * [The `display` property used for the container of `children` is changed](#the-display-property-used-for-the-container-of-children-is-changed)
-  * [A component includes a landmark role](#a-component-includes-a-landmark-role)
-  * [A component no longer includes a landmark role](#a-component-no-longer-includes-a-landmark-role)
+- [Overview](#overview)
+- [Changes](#changes)
+- [Reference](#reference)
+  - [The type of a prop is broadened](#the-type-of-a-prop-is-broadened)
+  - [The type of a prop is narrowed](#the-type-of-a-prop-is-narrowed)
+  - [The `display` property used for the container of `children` is changed](#the-display-property-used-for-the-container-of-children-is-changed)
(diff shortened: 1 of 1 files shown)


broccolinisoup (approved):
>Is this a situation worth documenting in our versioning guide?

Definitely - I think this is something might look very trivial but can potentially break things

>Does this feel like an appropriate version bump for this situation?

Yes
>Do the examples clearly convey why the version bump is what it is?

Absolutely 🔥 
>Are there any scenarios that are missing that should be added?

I can't think of another example from the top of my head but we can always add as we see.

Overall looks great! Thanks for adding this. We should definitely link this guide to the pull request template

joshblack:
Thanks so much for taking a look @broccolinisoup! Really appreciate it.

>  We should definitely link this guide to the pull request template

It should be available in this line under the rollout section: https://github.com/primer/react/blob/main/.github/pull_request_template.md?plain=1#L25 but it may be more hidden than intended 😅 Do you have any recommendations for where it might be better placed or how to make it more obvious?

broccolinisoup:
>It should be available in this line under the rollout section: https://github.com/primer/react/blob/main/.github/pull_request_template.md?plain=1#L25 but it may be more hidden than intended 😅 Do you have any recommendations for where it might be better placed or how to make it more obvious?

@joshblack Oh, it is commented out! I missed that 🙈 I don't know if it would be helpful but I _think_ just taking the last sentence out of the comment block would make it more visible. People can see in the preview as well. What do you think?

```diff
-<!-- How do you recommend this change to be rolled out? Refer to [contributor docs on Versioning](https://github.com/primer/react/blob/main/contributor-docs/versioning.md) for details. -->
+<!-- How do you recommend this change to be rolled out? --> Please refer to [contributor docs on Versioning](https://github.com/primer/react/blob/main/contributor-docs/versioning.md) for details.
```

Screenshot of the Rollout strategy section in the pull request template and it has the versioning documentation link is above the versioning options.

siddharthkp:
> Is this a situation worth documenting in our versioning guide?

Yes! 

> Does this feel like an appropriate version bump for this situation?

I think so. We might try to find workarounds to release things sooner. But, as a rule or starting point, this is good!

> Do the examples clearly convey why the version bump is what it is?

Yes

> Are there any scenarios that are missing that should be added?

This may be covered implicitly, but changing the container type (li → div, or div → li) can lead to similar type errors as the second example

siddharthkp (approved):

PR #4146, opened 2024-01-13
SelectPanel2: Only show clear action in search with text

siddharthkp:
Clear action is only visible when there is something to clear

https://github.com/primer/react/assets/1863771/a777043d-e2a7-4851-a24e-9c4ed2792b93


src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -377,12 +377,7 @@ const SelectPanelSearchInput: React.FC<TextInputProps> = ({onChange: propsOnChan
           }}
         />
       }
-      sx={
-        {
-          /* TODO: uncommenting this breaks keyboard navigation, that's odd */
-          // '& input:empty + .TextInput-action': {display: 'none'},
-        }
-      }
+      sx={{'&:has(input:placeholder-shown) .TextInput-action': {display: 'none'}}}
```
siddharthkp, line 380: Note: this was fun! was able to use a [new css feature](https://caniuse.com/css-has) and a [really old css feature](https://caniuse.com/css-placeholder-shown)

broccolinisoup (approved):
Looks great!

One tiny comment: I know this is nitpicking but I just want to mention. I don't think `placeholder` is an exposed prop in the API but funny thing is that when the `placeholder=''` instead of `placeholder='Search'`, the css logic works as expected in Chrome but not in Safari. In Safari, when the placeholder is an empty string, the clear action is visible. I found out while I was like 'Ohh `placeholder-shown`, let me play with this' 😄

siddharthkp:
> the css logic works as expected in Chrome but not in Safari. 

Interesting, this is what I see in Arc. I imagine it's because there is no placeholder to show, so it's now shown 😅

no placeholder, action is shown

PR #4158, opened 2024-01-18
Adds tests for Dialog v2

mperrotti:
I'm adding these missing tests because regressions to Dialog were accidentally introduced in https://github.com/primer/react/pull/4048 and we almost didn't catch them.

### Changelog

No component changes are being shipped

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
  - These are just changes to tests

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react'
+import {fireEvent, render, waitFor} from '@testing-library/react'
```
broccolinisoup, line 2: Curious to hear if there is a reason to use fireEvent here instead of userEvent 👀 I have been using userEvent mostly based on what I learnt from the differences from fireEvent ([docs reference](https://testing-library.com/docs/user-event/intro/#differences-from-fireevent)) but curious to hear what you think 🙌🏻

mperrotti: I just copied some stuff over from the old Dialog tests. I'll update to use `userEvent` instead 🙂

broccolinisoup (commented):
I think we have a lint and snapshot failing error but other than that it looks good!

mperrotti:
Thanks @broccolinisoup - everything should be correct now 🤞

src/Dialog/Dialog.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+    matchMedia = new MatchMediaMock()
+  })
+
+  afterEach(() => {
+    matchMedia.clear()
+  })
+
+  behavesAsComponent({
+    Component: Dialog,
+    options: {skipAs: true, skipSx: true, htmlRenderer: true},
```
broccolinisoup, line 22:
```suggestion
    options: {skipAs: true, skipSx: true},
```

broccolinisoup (approved):
There was a [type error](https://github.com/primer/react/actions/runs/7807556039/job/21296179178) but pushed a commit to fix it, I hope it is okay! It looks great now 🎉 Thanks!

PR #4160, opened 2024-01-19
Remove `MarkdownEditor` and related code

### LINKED ISSUE #3604: Remove `MarkdownEditor` from `@primer/react`

iansan5653:
Closes #3604

Removes the `MarkdownEditor` and related components and hooks. I also reviewed our dependencies and removed any that were only used by this code. Not much to say here - it was mostly just deleting everything and running `npm i`.

#### Removed

Deprecated experimental components and hooks:

- `MarkdownEditor`
- `MarkdownViewer`
- `InlineAutocomplete`
- `useCombobox`
- `useDynamicTextareaHeight`
- `useIgnoreKeyboardActionsWhileComposing`
- `useSafeAsyncCallback`
- `useSyntheticChange`
- `useUnifiedFileSelect`

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

Rollout for GitHub will be straightforward for this change as all internal code has already migrated to the internal versions of these components and hooks.

> [!NOTE]
> Although this is all experimental code and we can delete it at any time since experimental code doesn't follow semver, I think it's more polite to any downstream consumers to delete this in a major version, giving this change more visibility. There's no rush to make this change as the code is already deprecated, so we can hold off merging this until we're ready to make another major bump (or we can merge to a release branch if we have one).

### Merge checklist

- [x] Added/updated tests
- NA: ~~[ ] Tested in Chrome~~
- NA: ~~[ ] Tested in Firefox~~
- NA: ~~[ ] Tested in Safari~~
- NA: ~~[ ] Tested in Edge~~


broccolinisoup (commented):
Thanks for pushing this PR @iansan5653 - I'll refrain from approving to make sure it doesn't get merged accidentally. Let me know if you have any concern.

PR #4166, opened 2024-01-22
Tentative focus fix in action list item

pksjce:
Closes https://github.com/github/accessibility-audits/issues/4641

**Bug description**
The bug specifically occurs in one case where the action menu does not have a visible anchor and is programmatically opened through a keyboard shortcut combination. Specifically in the code viewer, on hitting the SHIFT + CTRL + C combination, you can see the actionmenu emerge at the cursor position. This action menu can also be accessed at the line level through an anchor thats only visible on focus. 
The actual bug is that the focus does not shift from the cursor position into the action menu's first item as expected. 
This bug has been especially challenging because it is hard to reproduce in it's entireity. One can fiddle around with tab button and fix the focus params such that the bug becomes nonexistent. Also I was not successful in creating a minimum reproduction of the issue in a storybook. https://github.com/primer/react/pull/4025 . Any advise on fixing this is welcome.

**Root cause analysis**
The current theory is that as per the spec on `focus-visible` https://drafts.csswg.org/selectors/#indicate-focus , these lines tell us what could the problem be.

> "If the previously-focused element [indicated focus](https://drafts.csswg.org/selectors/#indicate-focus), and a script causes focus to move elsewhere, indicate focus on the newly focused element.
> Conversely, if the previously-focused element did not [indicate focus](https://drafts.csswg.org/selectors/#indicate-focus), and a script causes focus to move elsewhere, don’t indicate focus on the newly focused element."

So if the focus shifts programmatically then the spec does not require to apply the focus-visible pseudo class.

**Proposed fix**
After much deliberation, I think our best chance is to fix the issue to our best of knowledge and see if it fixes the actual issue.
We will make sure that the first item in action menu always receives focus when it opens by more robust css. This should be the right solution.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


joshblack:
@pksjce was trying to re-create this issue again on dotcom and am having trouble getting the shortcut to trigger the menu 🤔 I just get the screen jumping all the way to the bottom. Was curious if others are observing the same or if it was just me.

TylerJDev:
@joshblack, yup I'm getting the same behavior. [I've asked this in one of the slack channels](https://github.slack.com/archives/C0357UQ03KN/p1706302846106509) to see if this functionality was removed or not.

pksjce:
Hi @TylerJDev @joshblack , I see that they seemed to have fixed this. Though I'm still having some challenges getting it working. I'm trying to get a review lab out so we can test the fix

TylerJDev:
@pksjce, @joshblack, I'm now able to open the menu in code view. Let me know if you need an additional look within the review lab too!

joshblack:
@TylerJDev is this an implementation behavior of Safari where it does not show focus if the menu is opened with a click and then keyboard is used? The list items clearly have focus, and `:focus-visible` styling works in Safari, but it does not seem to be triggered in this case 🤔 

The fix above totally works because the `focus-visible` polyfill adds the appropriate classes/attributes so the styling kicks in but was wondering if this was a behavior that Safari has implemented or is something else at play.

joshblack (approved):
LGTM to fix this in the interim while we figure out what on earth is going on 😅

TylerJDev (approved):

PR #4168, opened 2024-01-22
Refactor/remove focus visible polyfill

joshblack:
In our maintainer sync today, it came up that we no longer need to polyfill `focus-visible` based on GitHub's browser support range. This PR removes this polyfill and any references to the explicit `.focus-visible` class

#### Changed

- Instances of `.focus-visible` to just `:focus-visible`

#### Removed

- The `focus-visible` polyfill

### Rollout strategy

- [x] Minor release

I selected a minor release here but could also see this considered a breaking change since we don't explicitly call out our browser compatability (unless we indicate that it should match GitHub's?)

### Testing & Reviewing

- This PR changes some BaseStyles so we should confirm that none of these impact major functionality
- In most cases, this PR removes CSS selectors which had a fallback already with `:focus-visible`. We should make sure no remaining `.focus-visible` selectors remain or are in use

### DIFF
--- a/.changeset/tricky-planets-drum.md
+++ b/.changeset/tricky-planets-drum.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Remove support for focus-visible polyfill
--- a/@types/focus-visible/index.d.ts
+++ b/@types/focus-visible/index.d.ts
@@ -1,1 +1,0 @@
-declare module 'focus-visible'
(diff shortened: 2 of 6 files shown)


keithamus (approved):
:fire: :fire:

TylerJDev:
I think this might affect the changes that could happen in https://github.com/primer/react/pull/4166. Does this mean we'd need to find a new solution to fix the referenced bug? 🤔

siddharthkp (approved):
🎉 

Let's test this with dotcom to be extra extra sure :)

> I selected a minor release here but could also see this considered a breaking change since we don't explicitly call out our browser compatability (unless we indicate that it should match GitHub's?)

Happy with minor

joshblack:
Closing as we currently need this polyfill for some odd Safari behavior 🤔

PR #4170, opened 2024-01-22
SelectPanel2: Fix height for fit-content in Safari

siddharthkp:
✅ Tested storybook and dotcom in Arc/Chrome, Safari and Firefox

By default, SelectPanel has "large" height (432px) and maintains that when search results need less space:

| Header | Header |
|--------|--------|
| SelectPanel showing all items in a scrollable list | SelectPanel showing only 2 items, still has the same height |

However, for really short panels, you would want it to adjust to the contents, `<SelectPanel height="fit-content">` does the job for that:

https://github.com/primer/react/assets/1863771/dbc59014-d9e4-49ed-9a63-47cde89acc6c

(height transition is part of story, not select panel)

### DIFF
--- a/.changeset/selectpanel-fix-height.md
+++ b/.changeset/selectpanel-fix-height.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+experimental/SelectPanel: Fix height for `fit-content`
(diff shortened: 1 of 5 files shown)


broccolinisoup (approved):
I think we should make sure to add the new `height` prop value `fit-content` to the docs.json other than that looks great ✨ 

One more comment though, I know this is not about the component itself but just to have the most accurate story examples and have the best accessible practises, we might want to add a todo or note about the dynamic content addition (when a checkbox is checked, another content becomes visible) in this new story and see how we can make sure screen readers announce this as well.

PR #4173, opened 2024-01-23
SelectPanel2: Update docs

siddharthkp:
- [x] Update props
- [x] Add stories to docs, restructure folder to match structure
- [x] Update components script to allow adding "examples" stories in docs


script/components-json/build.ts
```diff
@@ -56,17 +56,19 @@ const components = docsFiles.map(docsFilepath => {
   // Get path to feature story file
   // Example: src/components/Box/Box.docs.json -> src/components/Box/Box.features.stories.tsx
   const featureStoryFilepath = docsFilepath.replace(/\.docs\.json$/, '.features.stories.tsx')
+  const exampleStoryFilepath = docsFilepath.replace(/\.docs\.json$/, '.examples.stories.tsx')
 
   // Get source code for each feature story
   const featureStorySourceCode = getStorySourceCode(featureStoryFilepath)
+  const exampleStorySourceCode = getStorySourceCode(exampleStoryFilepath)
```
broccolinisoup, line 63: I am wondering if we should check this with the rest of the Primer folks?

Because looking at the rails examples on the primer.style, they seems to have only features of the component albeit the title of the section is "Examples" - My only question/concern is that would it be too many if we include features and examples and potentially not help folks to find the "feature" that they are looking for? 🤔

broccolinisoup (commented):
Left a comment about the build script- curious to hear what your thoughts are 🙌🏻

script/components-json/build.ts
siddharthkp: > I am wondering if we should check this with the rest of the Primer folks?

Sure!

> they seems to have only features of the component 

Hmm, I personally would want to include examples on common usage as well. I think it makes sense to show examples in the docs

>  My only question/concern is that would it be too many if we include features and examples and potentially not help folks to find the "feature"

Good callout! That's valid, I don't think we should add all the examples for sure.

I've sorted the stories in the order I think they should be shown. I think I can move some of the examples into features though, so that we show features first and examples after that 🤔

siddharthkp:
@broccolinisoup Didn't get any concerns on slack, going forward with this :)

broccolinisoup (approved):
Thanks for following up! Looks great 🎉

siddharthkp:
The vrt job seems to be failing because of a different PR 🤔 (https://github.com/primer/react/actions/runs/7824236577)

PR #4174, opened 2024-01-23
feat(live-region): add support for <live-region> element

joshblack:
## Overview

This is a proof of concept for a `live-region` custom element that is used in coordination with our components to provide accessible announcements in live regions.

## Problem

When making announcements through live regions, the container for a live region must exist in the document **before** the announcement in made. When this is not the case, the announcement will be dropped.

The core challenge for making announcements is when a component is rendered conditionally. In these scenarios, we cannot rely on a stable live region _within_ the component. Instead, we would require one of the following approaches:

- Require a live region to be provided via context at the root level
- Dynamically find, or create, a live region and buffer messages sent to that live region

This PR introduces a direction for both approaches through a combination of components and hooks.

## Approach

This PR introduces components, hooks, and a `<live-region>` custom element in order to solve the problems mentioned above.

### Components

`LiveRegion` and `LiveRegionOutlet` are two internal components that are used to render and provide a `<live-region>` for a tree of React components. These components can be used at the top-level of an application or can be used within a component.

```tsx
function MyApp({ children }) {
  return (
    <LiveRegion>
      {children}
      <LiveRegionOutlet />
    </LiveRegion>
  );
}
```

In the example above, `LiveRegion` acts as a context provider that provides `children` with access to a `live-region` element rendered by `LiveRegionOutlet`.

### Hooks

`useLiveRegion` and `useAnnounce` are two internal hooks used to either get or interact with the `live-region` element in context.

- `useLiveRegion(): LiveRegionElement` provides access to the `live-region` in context
  - Note: this hook will dynamically find, or create, a `live-region` element if one does not exist in context
- `useAnnounce()` provides an `announce()` method to directly announce the message
  - Note: the value returned by this hook is a constant reference which makes it easier to use within `useEffect()` hooks

```tsx
function MyComponent() {
  const announce = useAnnounce();
  return (
    <button onClick={() => { 
      announce('This is an example')
    }}>
      Test announcement
    </button>
  );
}
```

### `<live-region>` custom element

The `live-region` custom element is an attempt at using a Custom Element to provide a common interop point between Primer and github/github. This custom element mirror methods used upstream in github/github, specifically `announce()` and `announceFromElement()`, and provides them from the element class directly.

```tsx
const liveRegion = document.createElement('live-region');
document.documentElement.appendChild(liveRegion);

liveRegion.announce('Test announcement');

liveRegion.announce('Test announcement with delay', {
  delayMs: 500,
});

liveRegion.announce('Test assertive announcement', {
  politeness: 'assertive',
});
```

### `<live-region>` custom element design

Under the hood, this live region uses the following structure:

```html
<style>
:host {
  clip-path: inset(50%);
  height: 1px;
  overflow: hidden;
  position: absolute;
  white-space: nowrap;
  width: 1px;
}
</style>
<div id="polite" aria-live="polite" aria-atomic="true"></div>
<div id="assertive" aria-live="assertive" aria-atomic="true"></div>
```

## Feedback


joshblack:
Note: looking into the deploy preview problem now to see what we can do. Unfortunately we directly wire things up to `@primer/react` from docs so the node export conditions aren't being picked up 😅

src/internal/components/LiveRegion/live-region-element/element.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (!container) {
+      throw new Error('Unable to find container for politeness message')
+    }
+    return container.textContent
+  }
+
+  setMessage(message: string, politeness: Politeness = 'polite') {
+    const container = this.shadowRoot?.getElementById(politeness)
+    if (!container) {
+      throw new Error('Unable to find container for message')
```
mperrotti, line 153: How would I even hit this error?

src/internal/components/LiveRegion/live-region-element/element.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    if (container.textContent === message) {
+      container.textContent = `${message}\u00A0`
```
mperrotti, line 157: Why do we need a non-breaking space here? Not bothered by it, just want to learn.

src/internal/components/LiveRegion/live-region-element/live-region-element.d.ts
```diff
@@ -0,0 +1,10 @@
+import {LiveRegionElement} from './element'
+import * as React from 'react'
+
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'live-region': React.DetailedHTMLProps<React.HTMLAttributes<LiveRegionElement>, LiveRegionElement>
```
mperrotti, line 7: Is this adding `<live-region>` to be identified as a JSX element? Or is it doing something else? I don't have a problem with this, I just want to educate myself.

mperrotti (commented):
Wow, excellent work! I want to "approve", but I (unfortunately) don't have much experience with custom elements yet, and would love to get feedback from folks who do.

mperrotti:
> What did you think of the problem statement?

It is clear and I've run into the challenges identified multiple times.

---

> What did you think of the approach? Do you feel like there are other approaches worth exploring in this space?

I love this approach. We could also explore React portals, but I think this is better.

---

> What did you think of the design? Did it make sense? Anything you would change?

Before looking at the source, my impression was that the hook's API design is very straightforward and intuitive. However, I found the following things to be a little confusing about the components:

- I was confused about what `<LiveRegion>` does and why a `<LiveRegionOutlet>` is also needed. Which one creates a live region?
- When would I want to use `<Status>` instead of `<LiveRegion>`?
- Why would I want to use a `<LiveRegion>` instead of just using the `useAnnounce()` hook?
- Do I need to render a `<LiveRegion>` in order to use the `useAnnounce()` hook?

After reviewing the PR, I think I can answer all of those questions. However, I still have these questions:

- When would I use `<Message>`?
- Why don't we use dot notation for `<LiveRegionOutlet>` like we do for other components with parent-child relationships?
- Should we consider exporting some of these hooks and components for Primer consumers to use in their applications?

---

> How do you feel about moving forward with this approach? Anything you're worried about or anything you'd want addressed beforehand?

I feel good about it assuming that we're going to do some screen reader testing. I really think we should add some usage examples to Storybook.

src/Spinner/Spinner.tsx
```diff
@@ -1,41 +1,50 @@
...
-      <path
-        d="M15 8a7.002 7.002 0 00-7-7"
-        stroke="currentColor"
-        strokeWidth="2"
-        strokeLinecap="round"
-        vectorEffect="non-scaling-stroke"
-      />
-    </svg>
+    <>
+      {loadingMessage !== undefined ? <Status message={loadingMessage} /> : null}
```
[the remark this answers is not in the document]
joshblack: > So Status will buffer the announcement so that it doesn't get missed

Exactly 💯 when we run into a case where a live region is not on the page yet, we have make a `live-region` custom element that buffers announcements until it is in the DOM.

Thankfully this should help to address issues with both creating the live region and having a message at the same time by separating them 🤞

src/internal/components/LiveRegion/LiveRegion.tsx
```diff
@@ -0,0 +1,56 @@
...
+// `LiveRegion` in order for this to occur.
+//
+// If possible, we use a declarative shadow dom for `live-region` during
+// server-side rendering
+function LiveRegionOutlet() {
+  const setLiveRegion = useOutlet()
+  return (
+    <live-region ref={setLiveRegion} suppressHydrationWarning>
+      {canUseDOM ? null : (
+        <template
```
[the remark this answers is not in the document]
joshblack: Should be fixed in the latest commit!

src/internal/components/LiveRegion/Message.tsx
```diff
@@ -0,0 +1,35 @@
...
+    savedLiveRegion.current = liveRegion
+  }, [liveRegion])
+
+  useEffect(() => {
+    if (committedRef.current !== true) {
+      return
+    }
+
+    const cancel = savedLiveRegion.current.announce(value, {
+      delayMs: 750,
```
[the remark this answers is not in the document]
joshblack: I think that value for `Message` come from data table work specifically but I don't know if it's broadly applicable. `Message` here really is for DataTable to not break and hopefully we can just update that component to use `Status` or `useAnnounce` 🤞

src/internal/components/LiveRegion/live-region-element/define.ts
```diff
@@ -0,0 +1,3 @@
+import {LiveRegionElement} from './element'
+
+customElements.define('live-region', LiveRegionElement)
```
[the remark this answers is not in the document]
joshblack: Yup! At least that's the hope 😅 They should be replaced with the shim in Node.js environments

src/internal/components/LiveRegion/live-region-element/element.ts
joshblack: God, good question. I think if the shadow dom was changed in some way is really the only case (or the shadow root wasn't initialized properly) but hopefully this never happens 😅

src/internal/components/LiveRegion/live-region-element/element.ts
joshblack: @mperrotti from what I understand, this case comes up when the current message and the last message are the same. The message needs to be different (in this case the non breaking space totally counts) in order to be re-announced.

src/internal/components/LiveRegion/live-region-element/live-region-element.d.ts
joshblack: Yup, exactly 👍

joshblack:
@mperrotti thanks so much for the review! Appreciate it a ton. Will respond to some of the questions below 👇 

> When would I use <Message>?

Hopefully never 😅 This is in place just so DataTable keeps working.

> Why don't we use dot notation for <LiveRegionOutlet> like we do for other components with parent-child relationships?

We could definitely use this style if we expose it in our public API. Right now it's an internal component which doesn't typically use that convention.

> Should we consider exporting some of these hooks and components for Primer consumers to use in their applications?

Definitely! I think it'd be incredibly helpful. Depending on how the feedback goes for this I would love to bring it to accessibility and see if this would be a good interop point for announcements at GitHub 🤞 

Some of the other ones that I thought were great questions and will add them to the questions section in the PR 🔥 

> I was confused about what <LiveRegion> does and why a <LiveRegionOutlet> is also needed. Which one creates a live region?

It can be helpful to think of `LiveRegion` as the context provider (or even `LiveRegionProvider`) and `LiveRegionOutlet` as what is actually rendering the `live-region` element.

Totally get that this is confusing naming-wise and we can definitely change that, maybe to `LiveRegionProvider` and `LiveRegion` to make it more obvious?

mperrotti:
> Totally get that this is confusing naming-wise and we can definitely change that, maybe to `LiveRegionProvider` and `LiveRegion` to make it more obvious?

Yea, that name change makes things a lot clearer 🙂

joshblack:
@mperrotti awesome, done! ✨

joshblack:
One quick update, the `Status` component has been updated after some feedback. This component, along with a new `Alert` component, now announce their contents instead of a dedicated `message` from.

For more info, checkout the [Status, Alert](https://github.com/primer/react/pull/4174#:~:text=before%20JS%20hydrates.-,Status%2C%20Alert,-The%20Status%20and) section above.

owenniblock:
@joshblack I like this, feels more "React-y" that what we're using already. We should discuss this with the accessibility team if we haven't already since it'll presumably replace the existing `aria-live.ts` stuff that we use in dotcom at the moment. I wonder if there's any way to bridge the gap between the two to allow Rails developers to use the same `<aria-live>` element 🤔

PR #4178, opened 2024-01-24
SelectPanel2: Add story for "create new item"

siddharthkp:
- WIP: Still have doubts about the focus management
- Mockup: [Figma](https://www.figma.com/file/S3xgRPvRuFVCDGxPlIGxaP/SelectPanel-%26-Filtering?type=design&node-id=1406-130691&mode=design)
- With upcoming dialog refactor: https://github.com/primer/react/pull/4179

https://github.com/primer/react/assets/1863771/240f2ad7-0ba5-46f1-8c01-fd1c40c6ce23

PR #4179, opened 2024-01-24
[Do not merge] SelectPanel2: Add story for "create new item" + <dialog> refactor

siddharthkp:
- Merged https://github.com/primer/react/pull/4178 and https://github.com/primer/react/pull/3840
- Mockup: [Figma](https://www.figma.com/file/S3xgRPvRuFVCDGxPlIGxaP/SelectPanel-%26-Filtering?type=design&node-id=1406-130691&mode=design)

Open question on keyboard navigation: Should the "Create new label" button be reachable by arrow keys (like it's part of list) or with Tab (like it's outside of list). [See storybook](https://primer-d27cbe6e58-13348165.drafts.github.io/storybook/iframe.html?id=drafts-components-selectpanel-examples--create-new-row&viewMode=story)

https://github.com/primer/react/assets/1863771/240f2ad7-0ba5-46f1-8c01-fd1c40c6ce23


siddharthkp:
@maximedegreve @ichelsea Question about focus management for you: Should the "Create new label" button be reachable by arrow keys (like it's part of list) or with Tab (like it's outside of list). [See demo on storybook ↗](https://primer-d27cbe6e58-13348165.drafts.github.io/storybook/iframe.html?id=drafts-components-selectpanel-examples--create-new-row&viewMode=story)

maximedegreve:
@siddharthkp [This was answered here.](https://www.figma.com/file/S3xgRPvRuFVCDGxPlIGxaP?node-id=2571:90595&mode=design#679843165)

ichelsea:
@maximedegreve I left a comment in the Figma but I thought we settled on since the button lives outside of the `listbox`, we shouldn't allow arrowing down to it or back up to the input (if there are no results in the `listbox`). The arrow controls should only work inside of the `listbox` and a concession to allow arrowing down to the `listbox` from the input (but not back up) as a carry-over from previously expected user interaction.

PR #4181, opened 2024-01-25
test: remove expect.extend calls to toHaveNoViolations

joshblack:
This is a small refactor to remove calls to `expect.extend(toHaveNoViolations)` in individual test suites since we have a setup call that includes this by default.

#### Changed

- Remove calls to `expect.extend(toHaveNoViolations)` in individual test suites

### Rollout strategy

This is a change to tests


siddharthkp (approved):
Sweet!

PR #4182, opened 2024-01-25
theme.js -> theme.ts

jaked:
Fixed broken link: the `theme.js` file was renamed to `theme.ts` in `primer/react`


joshblack (approved):
Thanks! Appreciate it 🙏

PR #4195, opened 2024-01-30
Scroll AnchoredOverlay when it's inside a scrolling child of the body

owenniblock:
Closes https://github.com/github/primer/issues/1421

Adds the code @siddharthkp created for [his PR to solve this problem](https://github.com/primer/react/pull/2200). When you scroll within a parent element that isn't the main body element, the Anchor now updates.

### Original functionality

A video showing a panel of buttons, the user opens an overlay showing a cat by pressing one of the buttons. When the user scrolls the panel, the overlay cat remains fixed in place.

https://github.com/primer/react/assets/4696224/81404a1d-3fbf-432b-ae9c-a2cf7618fc62

### New functionality

A video showing a panel of buttons, the user opens an overlay showing a cat by pressing one of the buttons. When the user scrolls the panel, the overlay follows the button that was pressed. There's an issue shows where the overlay moves position based on the clipping of the parent rect.

https://github.com/primer/react/assets/4696224/0a162092-d1e1-4bca-bd24-eb7de5c78993

#### New

This adds a new `useElementObserver` to observe when the rect for an element changes.

#### Changed

Changes `useAnchoredPosition` and `AnchoredOverlay` to use the new hook.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

I'm not totally happy with this but couldn't see a better way to do it. I removed the `instanceof Element` because if broke the gatsby SSR build (Element isn't available server side). I've tried to use the [solutions suggested by StackOverflow](https://stackoverflow.com/questions/55971033/es6-import-breaks-gatsby-build-webpackerror-referenceerror-element-is-not-de) however excluding the library from the build isn't possible (we have to exclude the whole component, but then the Javascript still gets build for every other component so it _still_ fails - I think some work on src/hooks/index.ts might resolve this but that was a rabbit hole I didn't have time to explore) and trying to only load the code client-side was proving to be fiddly and hacky given the structure of the code. In the end, I opted for the simplest solution 🤷 

Do we need to worry about the increase to the size of `browser.*.js`?

I had a look at the weird issue whereby the overlay jumps position when the scroll takes the anchor button into a position where `updatePosition` returns a different orientation. We can fix this, but it's a much larger lift (and I believe this issue also exists for other overlays which don't suffer from this problem). To fix it, we'll probably need to rework `useAnchoredPosition` so that `updatePosition` is only called when the overlay is first opened, subsequent changes should move the overlay based on the original orientation of the overlay (resetting when closed).

Finally: do we want to add this functionality to `AutocompleteOverlay` and/or `SelectPanel`?

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


owenniblock:
The build error here might be due to [the same issue as described on this StackOverflow post](https://stackoverflow.com/questions/55971033/es6-import-breaks-gatsby-build-webpackerror-referenceerror-element-is-not-de).

owenniblock:
🤔 It looks like the change has surfaces an axe error where there wasn't one previously. It _looks_ like the only axe error on the AnchoredOverlay is due to the sentinel elements being focuable ([you can find this error with a test of this page with the overlay open](https://primer.style/react/storybook/iframe.html?globals=colorScheme:light&viewMode=story&id=components-anchoredoverlay--default)).

Maybe this fix has made the AnchoredOverlay open correctly when `initiallyOpen={true}`

owenniblock:
Converting this back to a draft. Might put it down for now and take a look in spare moments next week.

keithamus (commented):
I don't think observing changes is the right way to handle this. It causes some quite poor performance issues by causing layout thrashing and bridging JS binding - hence you adding a debounce, which then causes jankyness. I'm fairly sure it can be fixed in CSS.

siddharthkp:
> I'm fairly sure it can be fixed in CSS.

@keithamus That would be really cool! My initial exploration was a quite a while ago, so I don't remember if I explored css solution or I was only trying to extend getAnchoredPosition 🤔

owenniblock:
Closing, we can always reopen if the CSS investigation doesn't bear fruit.

siddharthkp:
Hi again!

Does anyone have this on their list? Don't want to completely drop this either.

owenniblock:
Not on my list @siddharthkp

PR #4198, opened 2024-01-30
SelectPanel2: Export props + pass props

siddharthkp:
- Closes https://github.com/github/primer/issues/3024

### Rollout strategy

- [x] Patch release

### Merge checklist

- [dont have tests yet!] Added/updated tests
- [x] Tested in Chrome
- NA] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))


broccolinisoup (approved):

joshblack (approved):

ayy-bc:
@siddharthkp we would also need the `SelectPanelSecondaryActionProps` to be exported to use it with the item picker for cases where we need secondary actions (related: https://github.com/github/collaboration-workflows-flex/issues/423, https://github.com/github/github/pull/310856/files). 

Do you want me to create another issue to track that or can that also go as a part of this PR?

src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -250,6 +250,7 @@ const Panel: React.FC<SelectPanelProps> = ({
             '100%': {transform: 'scale(1, 1)'},
           },
         }}
+        {...props}
```
JelloBagel, line 253: @siddharthkp Does the typescript on `props` also allow `StyledOverlay` props to be passed through? 

For context, I'm asking case where the prop `maxHeight` is needed to fix a styling issue with the overlay height: https://github.com/github/collaboration-workflows-flex/issues/449
```
<SelectPanel
      height="fit-content"
      maxHeight="large"
      ...
/>
```

siddharthkp:
@ayy-bc Sure thing, I can add them to this PR

src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: >  Does the typescript on props also allow StyledOverlay props to be passed through?

Not yet!

> I'm asking case where the prop maxHeight is needed to fix a styling issue with the overlay height: https://github.com/github/collaboration-workflows-flex/issues/449

That's an interesting issue, not sure if maxHeight is the solution. Going to reply on the issue first itself

PR #4199, opened 2024-01-30
SelectPanel2: Improve keyboard navigation from search input

siddharthkp:
When the focus is on search input,
1. Pressing ArrowDown should focus the first element in the list
2. Pressing Tab should should focus the last focused element in the list

When the focus is in the list,
1. Pressing Up should not come out of the list
2. Pressing Shift + Tab should move focus to the list

https://github.com/primer/react/assets/1863771/9abc6626-2eca-4e21-be70-6faf90736368


joshblack (commented):
Just wanted to ask a broader clarification question before reviewing, what kind of role/semantics are we looking for between the input and the listbox? At first glance, this seems like a case where `role="combobox"` may be appropriate but wanted to check-in to see since the intent here is to shift the focus to the list box.

siddharthkp:
> At first glance, this seems like a case where role="combobox" may be appropriate but wanted to check-in to see since the intent here is to shift the focus to the list box.

Interesting! I can see how that might work. Until now, we've looked at it as a listbox with dynamic content 🤔 

cc @ichelsea for clarification though!

selectpanel parts labelled, the list part is labelled as listbox

ichelsea:
I'm like 90% sure `role=combobox` can't be used here back when I was working with James, that definitely came up. I can't recall the reason off of the top of my head but I can deep dive into past issues if we need one! (If I recall it partially has to do with multi-select support?)

siddharthkp:
@ichelsea Thank you! That's very helpful

@joshblack Back for your review!

broccolinisoup (commented):
I was testing it on different stories and the focus doesn't move to the list on arrow down when there are sections in the list. Is that intentional or do we want the same focus model when there are sections as well? 


https://github.com/primer/react/assets/1446503/48820e9d-3608-4c1e-9381-86ac2ac28c8c

_A selectpanel pop up is open and displays two sections with headings and options in each. Pressing the arrow down when the focus is on the search input doesn't do anything._

siddharthkp:
> I was testing it on different stories and the focus doesn't move to the list on arrow down when there are sections in the list. Is that intentional or do we want the same focus model when there are sections as well?

That's not intentional at all! You've found a bug with groups :D

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -161,6 +163,14 @@ const Panel: React.FC<SelectPanelProps> = ({
     [internalOpen],
   )
 
+  // used in SelectPanel.SearchInput
+  const moveFocusToList = () => {
+    const selector = 'ul[role=listbox] li:not([role=none])'
```
broccolinisoup, line 168: The selector looks a bit too fragile, makes me a bit nervous 😬 but the query makes a lot of sense. Can we add tests/stories to make sure we always get the intended first element? (It is not a blocker, feel free to merge the PR)

broccolinisoup (approved):
It does work now with different variants of the selectbox - thanks for fixing!

Left one comment about adding tests but looks good!

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Will take it up in a new PR!

PR #4200, opened 2024-01-31
Tooltip v2: Allow external id to be passed down in tooltip so that the trigger can be labelled/described by the that id

broccolinisoup:
#### New

Allow Tooltip to use the custom id to label/describe the trigger element

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

In the [story](https://primer-24c81a19c4-13348165.drafts.github.io/storybook/?path=/story/drafts-components-tooltip-examples--custom-id), you can confirm that the custom id that is passed in the code is present in the dom.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


src/drafts/Tooltip/Tooltip.tsx
```diff
@@ -273,9 +273,9 @@ export const Tooltip = React.forwardRef(
             React.cloneElement(child as React.ReactElement<TriggerPropsType>, {
               ref: triggerRef,
               // If it is a type description, we use tooltip to describe the trigger
-              'aria-describedby': type === 'description' ? `tooltip-${tooltipId}` : child.props['aria-describedby'],
```
broccolinisoup, line 276: I don't think we need this prefix.

src/drafts/Tooltip/Tooltip.tsx
```diff
@@ -186,7 +186,7 @@ export const TooltipContext = React.createContext<{tooltipId?: string}>({})
 
 export const Tooltip = React.forwardRef(
   ({direction = 's', text, type = 'description', children, ...rest}: TooltipProps, forwardedRef) => {
-    const tooltipId = useId()
+    const tooltipId = useId(rest.id)
```
joshblack, line 189: Wanted to ask, is there a specific reason to not destructure this off of props? It doesn't seem like `id` as an identifier is used and could prevent potential bugs in the future if the position of `{...rest}` changes (like if someone accidentally moves it after where `id={tooltipId}` is set.

broccolinisoup: Great point! I used `rest.id` because id isn't a defined prop like text or children but I guess it doesn't matter though and it is not very safe. Thanks for the comment! I updated it and added tests ✨  https://github.com/primer/react/pull/4200/commits/c93cffe0a6b6f5a58c5429696f336eb927612e70

PR #4201, opened 2024-01-31
refactor(npm): use mjs as extension for import condition

joshblack:
This PR makes some of the changes suggested by [`publint`](https://publint.dev/@primer/react@36.7.0) for the package. The results can be viewed in the URL before or run locally by running `npx publint`.

The important context behind the `.mjs` extension change for ESM is that `.js` files are, by default, evaluated to be the same semantics as their package context. In our case, the root `package.json` has an implicit `"type": "commonjs"` and therefore these files would be interpreted as CommonJS when loaded by the native Node.js module resolution flow (even if they are imported using `import`)

Changing the extension to `.mjs` forces the ESM semantics that seem implied by the `import` export condition present

#### Changed

- Change extensions of ESM entrypoints to `.mjs`

### Rollout strategy

- [x] Minor release

This PR definitely needs an integration check upstream to make sure it doesn't break anything 😅 

### Testing & Reviewing

- Feel free to run `npx publint` locally to view the before and after of these changes
- Make sure that rollup is correctly generated `.mjs` files for ESM
- Question: should we include `npx publint` under our "lint" step in `ci.yml`?
- Do you have any questions or feedback around why this change is needed? 👀

### DIFF
--- a/.changeset/real-hairs-sniff.md
+++ b/.changeset/real-hairs-sniff.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Update package.json exports to use the .mjs extension for "import" export conditions
--- a/package.json
+++ b/package.json
@@ -2,29 +2,30 @@
   "name": "@primer/react",
   "version": "36.7.0",
   "description": "An implementation of GitHub's Primer Design System using React",
+  "type": "commonjs",
   "main": "lib/index.js",
   "module": "lib-esm/index.js",
   "exports": {
     ".": {
-      "import": "./lib-esm/index.js",
+      "import": "./lib-esm/index.mjs",
       "require": "./lib/index.js"
     },
     "./experimental": {
-      "import": "./lib-esm/experimental/index.js",
+      "import": "./lib-esm/experimental/index.mjs",
       "require": "./lib/experimental/index.js"
     },
     "./drafts": {
-      "import": "./lib-esm/drafts/index.js",
+      "import": "./lib-esm/drafts/index.mjs",
       "require": "./lib/drafts/index.js"
     },
     "./deprecated": {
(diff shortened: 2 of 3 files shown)


joshblack:
Note for reviewers, I'm pretty sure the failing check is due to styled-components still being CommonJS 😅 will check to confirm and see if it's a worthwhile fix or if this whole jump should happen in a major release instead of piecemeal

broccolinisoup (approved):
Thanks for raising this pr and adding a great description to it @joshblack! 🙌🏻  
The changes looks good to me!

>Question: should we include npx publint under our "lint" step in ci.yml?

Just curious to understand what kind of issues would violate https://publint.dev/ linting rules 🤔 Like the changes on tooling or would it be more than that? I hope my question makes sense 😄

joshblack:
> Just curious to understand what kind of issues would violate https://publint.dev/ linting rules. Like the changes on tooling or would it be more than that?

I definitely think this would catch changes related to a package's `package.json` along with published package files (very much so related to rollup and TS output). This is what it catches today, for reference: https://publint.dev/@primer/react@36.7.1

Hope I understood what you meant, let me know if not!

broccolinisoup:
> I definitely think this would catch changes related to a package's `package.json` along with published package files (very much so related to rollup and TS output). This is what it catches today, for reference: https://publint.dev/@primer/react@36.7.1

Yeah this is great! I had no idea such linting existed!

> Hope I understood what you meant, let me know if not!

That is exactly what I asked, thanks! I think it is a good idea to add the linting into the CI suit, otherwise it is hard to remember it to do it manually once in a while.

joshblack:
Sadly going to close this one out since I can't really figure out the dual type situation mentioned above 😞

PR #4205, opened 2024-02-01
Bugfix: Honor `sx` on draft `Dialog.Header`

pablonete:
`sx` was supported but it wasn't applied on top of standard `Header` styles.
```tsx
<Dialog.Header sx={{borderBottomWidth: 0}}>
```
This does nothing because `sx` is listed on the props but it's ignored on the component. It's the only component missing it (see `Title`, `Subtitle`, `Body`, and `Footer`).

### Changelog

No API changes.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

When using `renderHeader`, we couldn't use `Dialog.Header` to change the style, so we had to use a `Box` and copy & change the styles from the `Dialog.Header` definition.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/src/Dialog/Dialog.tsx
+++ b/src/Dialog/Dialog.tsx
@@ -342,6 +342,7 @@ const Header = styled.div<SxProp>`
   padding: ${get('space.2')};
   z-index: 1;
   flex-shrink: 0;
+  ${sx};
 `
 
 const Title = styled.h1<SxProp>`


broccolinisoup (approved):
Hello  @pablonete  👋🏻 Thanks so much for raising this PR. It looks good to me! 👍🏻 

We appreciate the feedback as well. I'll share it with the team, it is a super reasonable point 🙏🏻

siddharthkp:
@pablonete That's very good feedback, thanks! `renderHeader` is not a great API because it recommends ejecting out of future updates :(

re: `headerSx`, there is another way to do that already which might work for you. You can target the header from an `sx` on the Dialog like this:

```
<Dialog 
  sx={{
    [Dialog.Header]: {boxShadow: 'none'},
}}/>
```

PR #4212, opened 2024-02-02
ActionListItem: Better handle modifier keys

### LINKED ISSUE #4194: `<ActionList.Item>` doesn't correctly handle keyboard events
### Description

[<ActionList.Item>](https://github.com/primer/react/blob/main/src/ActionList/Item.tsx) does not properly handle keyboard selection events when a modifier key is used.

Currently `<ActionList.Item>` registers a handler for the [`onKeyPress` event](https://github.com/primer/react/blob/691adc6ff5593f3e7e847c42bffc8bdb29d7bfab/src/ActionList/Item.tsx#L252). `keypress` events are fired when a key -or- key combination produces a character value. This works fine in most cases, but on Windows for example, <kbd>Alt+Enter</kbd> does not produce a character value and an event is not emitted.

Additionally, `keypress` is now considered an [obsolete](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#obsolete_events) event and is [no longer recommended](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) for use. 

Separately, on Mac, the key combination <kbd>Option+Space</kbd> results in a [non-break space](https://en.wikipedia.org/wiki/Non-breaking_space) character value, which is different than typical <kbd>Space</kbd> resulting in the conditional check for `' '` on [line 231](https://github.com/primer/react/blob/691adc6ff5593f3e7e847c42bffc8bdb29d7bfab/src/ActionList/Item.tsx#L231) to fail.

### Steps to reproduce

1. Create a new [GitHub Project](https://docs.github.com/en/issues/planning-and-tracking-with-projects/creating-projects/creating-a-project)
2. Ensure there is a [table view](https://docs.github.com/en/issues/planning-and-tracking-with-projects/customizing-views-in-your-project/managing-your-views) on the project
3. Click the view's option button and then choose sort by

#### Mac:
Use the keyboard to select a secondary sort by using <kbd>Option+Space</kbd> and notice selection does not get made.

#### Windows
Use the keyboard to select a secondary sort by using <kbd>Alt+Enter</kbd> and notice selection does not get made.

### Version

v36.6.0

### Browser

Chrome

tmelliottjr:
Closes #4194

### Changelog

- Updated the handling of keyboard events within the ActionListItem component. Previously, `onKeyPress` was used. This has been replaced with `onKeyDown` to move away from now deprecated APIs.
- Additionally, instead of checking the produced character value in `event.key`, these changes now use `event.code` to check the actual key pressed.

#### New

N/A

#### Changed

N/A

#### Removed

N/A

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Using macOS, using <kbd>Option + Space</kbd> should now results in correct functionality. 

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


broccolinisoup (commented):
Hello @tmelliottjr 👋🏻 Thanks for raising this PR! From the first look, it looks good to me. I just wanted to check if it is ready for a review or still WIP since it is a draft PR. Let me know!

tmelliottjr:
@broccolinisoup I haven't had a chance to come back to this, but the last I saw this had CI errors. Which it doesn't now... 😕 

I looked at it briefly when it _was_ failing and it seemed that my changes adding the `keydown` handler had inadvertently affected the AutoComplete component.

Issue #3331, opened 2023-05-26
VerticalDivider for PageLayout.Pane always initially renders as style 'none'


Bestra:
### Description

I'm using React Router to render several different pages that are laid out almost identically. They generally look like this:
```jsx
<PageLayout>
  <PageLayout.Pane divider="line">
    {someContent}
  </PageLayout.Pane>
</PageLayout>
```
Each route renders a component with the same `PageLayout.pane` with a 'line' divider. When I switch routes, most of the DOM appears stable to my eyes, but there's a brief period where the vertical divider seems to disappear. 

I investigated and found that this happens on initial render, and that I could recreate it via the storybook locally. In the video below I've set a breakpoint when the `VerticalDivider` renders. You'll see that in the first few renders, the divider doesn't show at all. On the third render it pops into view as a line the way I'd expect.

https://github.com/primer/react/assets/2043348/eaffd9e6-7027-4fc0-a4b6-6f5b38af4800

From the React devtools profiler , the `VerticalDivider` renders a total of 4 times. On the third render it finally becomes visible.

<img width="484" alt="The react devtools profiler. The `VerticalDivider` component's third render of 4 is selected. 'Why did this render?' says 'Hooks 5 and 8 changed'" src="https://github.com/primer/react/assets/2043348/cd5d57b1-7fd0-41db-822a-9d3ebb5f1b8a">

I set a logpoint in the `PageLayout.Pane` component to record its props over its renders.  You'll see the `responsiveVariant` value changes from `none` to `line`. Each logpoint shows up twice since we're in strict mode.

![`console.log` results. There are 6 lines. `variant: {narrow: 'none', narrow: 'line'}` is present on every line. `responsiveValue` is `none` on the first 4 lines, but `line` on the last 2. `sx: {marginLeft: 0}` is present on every line](https://github.com/primer/react/assets/2043348/af5bfa55-217e-429c-81ba-8cd5c97fdf0f)

This corresponds to the `useMedia()` hooks used by `useResponsiveValue()` going from `false` to `true`.  

![The hooks tab from the react devtools component window. Hooks 5 and 8 are highlighted as `true`. The functions for these hooks are the state of the`isRegularViewport` and `isWideViewport` from `useResponsiveValue`](https://github.com/primer/react/assets/2043348/aad4023d-8d4b-49bf-afee-2ac7afe6b705)

Under the hood `useResponsiveValue()` calls `useMedia()` with a defaultState value of `false` for all three viewport widths
[...]

PR #4215, opened 2024-02-02
PageLayout: Fix layout shift for divider=line

siddharthkp:
- Closes https://github.com/primer/react/issues/3331

#### Problem

When using `<PageLayout.Pane divider={{regular: 'line', narrow: 'none'}}>`, we need to figure out which responsive value to use (regular, narrow or wide)

The hook `useResponsiveValue` we use renders the default (no divider) before the passed variant (like `line`) is rendered. This causes some layout shift (see video below)

https://github.com/primer/react/assets/1863771/6ba438ad-3050-4b00-9705-693c4c9a6e19

This shift is slightly worse when the page is rendered on the server because this hook is not SSR compatible:

https://github.com/primer/react/assets/1863771/4e450df2-d6f1-4d1b-8892-f80d77841173

Note: This shift is also visible even when we have the same variant across screen sizes (`<PageLayout.Pane divider="line">`)

&nbsp;

#### Changed

The ideal solution here would be to make this change css only, and not need any client side javascript. This however isn't very trivial and requires some time.

This PR has a pragmatic solution just for PageLayout to eliminate layout shift for the most common use case of `<PageLayout.Pane divider="line">` by rendering a transparent divider for `divider=none`. This will hopefully buy us more time to fix the actual cause.

&nbsp;

Before: There is no divider rendered at first, and then it renders and pushes the content away causing a layout shift

https://github.com/primer/react/assets/1863771/25dc12a9-c25f-47ab-a85f-598e64a5ff97

After: There is a transparent divider rendered at first, which does not cause any layout shift

https://github.com/primer/react/assets/1863771/506f3dd4-1cbd-4ee8-b113-38bc8e2ab78f

&nbsp;

### Rollout strategy

- [x] Patch release

### Merge checklist

- [x] Added/updated tests
~- [ ] Added/updated previews (Storybook)~
- [x] Tested in Chrome
~- [ ] Tested in Firefox~
~- [ ] Tested in Safari~
~- [ ] Tested in Edge~


siddharthkp:
@joshblack Nah, those are all wrong 😢 Converted back to drafts

siddharthkp:
This was bigger than I initially anticipated, abandoning this

PR #4221, opened 2024-02-05
`<UnderlineNav>` menu now uses the Anchored Position on smaller screen sizes to not clip.

### LINKED ISSUE #4059: The popover for `UnderlineNav` is clipped in mobile viewports
### Description

The popover for `UnderlineNav` is clipped in mobile viewports

### Steps to reproduce

I noticed this when using the latest repo beta opt-in 👀 

1. Visit a repo: https://github.com/primer/react
2. Scroll down to the UnderlineNav usage for the README, Code of Conduct, etc
3. Narrow the viewport` to a mobile viewport size (so that the More button is visible)
4. Activate the button to trigger the popover
5. Observe that the popover is clipped by the viewport

### Version

v36.x

### Browser

Chrome

Rebstorm:
Closes #4059

#### New

This PR fixes the issue with the `<UnderlineNavMenu>` on smaller screens using the menu. Where it would normally overflow to the left, this makes sure we align the absolute item using the [AnchoredPosition API](https://github.com/primer/behaviors/blob/main/docs/anchored-position.md#L1)  that you have provided. 

Old Behavior:
bad_behavior

New Behavior (321px >):
happy_321-360

On very small screens (not supported, but hey - why not):
happy_path_very_small

#### Changed

- Updates to `menuStyles` in the `<UnderlineNav>`
It has become more complex, as the style has become complex as well. I am forced to check now for how many pixels we have available in the `listRef`. I am also using the `getAnchoredPosition` API now to calculate the position of the element, but only for smaller screens, as to not mess with the `min-width` that was previously set.

#### Removed

- Nothing has been removed. 

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


src/UnderlineNav/styles.ts
```diff
@@ -135,17 +137,47 @@ export const menuItemStyles = {
...
+    maxWidth: '640px',
+  }
+
+  /** Is the width above 320px? If so, disregard the calculation below, as we need the minWidth to make the menu look pretty. **/
+  if (listRef && listRef.clientWidth > 320) {
+    return baseRules
+  }
+
+  /**
+   * We're dealing with very small screens, below 320px width. Here, we need to make sure that the menu is centered, or at least in view.
```
broccolinisoup, line 174: >We're dealing with very small screens, below 320px width.

Sorry if the bug wasn't very clear or if we provided misleading information there. Our [responsive screen sizes guideline](https://primer.style/foundations/responsive#viewport-size) mentions the starting point for width is 320px. Less than 320px is not considered practical to adopt. 

With that in mind, I don't think we need a dynamic width. We are still happy to have the 192 px min width for the menu. 

Another thing that I noticed on the storybook is that the menu is cut on the right side this time, probably because not positioning the anchor properly. (This is 360px width)

Overflow menu is open and the counters are barely visible due to not positioning the menu to be in the viewport

Sharing the same case on the production at 360px width for comparison.

Overflow menu is visible without an issue

broccolinisoup (changes requested):
Hi @Rebstorm 👋🏻 Thank you so much for raising this PR! We really appreciate it 🙏🏻 I left a comment, let me know if there is anything I can elaborate or if you have any questions/concerns!

Rebstorm:
**Woop woop!** 🎉

Thank you a billion for the feedback, @broccolinisoup. It really helps with the implementation and understanding what the aim is 🙏.

Here's how I suggest we fix this:

I kind of like the `margin: 0` solution which forces the menu to the right. It is simple and works. However, when the container does not offer enough room (as the more button is left-aligned), I suggest we use `getAnchoredPosition` like you suggested.

I've updated the code to do the following:

- **If we do not have enough space:**

  We use `getAnchoredPosition` to position our elements. The only real combination I found that would work would be to combine `{align: 'center', side: 'outside-left'}`. But if I understand the [documentation correctly](https://github.com/primer/behaviors/blob/main/docs/anchored-position.md#positionsettings-interface), that should be the right setting for popovers/menus.

- **If we have enough space:**

  Keep using `right: 0` to make it right-aligned.
  
  

Again, thanks for the feedback! If I missed something, or, misunderstood something - please let me know and ill get to it :pray:

PR #4234, opened 2024-02-08
`<UnderlineNav>` menu now uses the Anchored Position on smaller screen sizes to not clip, or go out of bounds. 

### LINKED ISSUE #4059: The popover for `UnderlineNav` is clipped in mobile viewports

Rebstorm:
Closes #4059

#### New

This PR fixes the issue with the `<UnderlineNavMenu>` on smaller screens using the menu. Where it would normally overflow to the left, this makes sure we align the absolute item using the [AnchoredPosition API](https://github.com/primer/behaviors/blob/main/docs/anchored-position.md#L1)  that you have provided. 

Old Behavior:
bad_behavior

New Behavior (321px >):
happy_321-360

On very small screens (not supported, but hey - why not):
happy_path_very_small

#### Changed

- Updates to `menuStyles` in the `<UnderlineNav>`
It has become more complex, as the style has become complex as well. I am forced to check now for how many pixels we have available in the `listRef`. I am also using the `getAnchoredPosition` API now to calculate the position of the element, but only for smaller screens, as to not mess with the `min-width` that was previously set.

#### Removed

- Nothing has been removed. 

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


Rebstorm:
Sorry :pray: . I seem to have done a wonky-merge. Recreated the PR, feedback and conversation can be found [here.](https://github.com/primer/react/pull/4221)

Rebstorm:
Copy+Pasting my last message:


Woop woop! 🎉

Thank you a billion for the feedback, @broccolinisoup. It really helps with the implementation and understanding what the aim is 🙏.

Here's how I suggest we fix this:

I kind of like the margin: 0 solution which forces the menu to the right. It is simple and works. However, when the container does not offer enough room (as the more button is left-aligned), I suggest we use getAnchoredPosition like you suggested.

I've updated the code to do the following:

    If we do not have enough space:

    We use getAnchoredPosition to position our elements. The only real combination I found that would work would be to combine {align: 'center', side: 'outside-left'}. But if I understand the [documentation correctly](https://github.com/primer/behaviors/blob/main/docs/anchored-position.md#positionsettings-interface), that should be the right setting for popovers/menus.

    If we have enough space:

    Keep using right: 0 to make it right-aligned.

Again, thanks for the feedback! If I missed something, or, misunderstood something - please let me know and ill get to it 🙏

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -338,7 +338,7 @@ export const UnderlineNav = forwardRef(
                   selectionVariant="single"
                   ref={containerRef}
                   id={disclosureWidgetId}
-                  sx={menuStyles}
+                  sx={menuStyles(containerRef.current, listRef.current)}
```
broccolinisoup, line 341: I would suggest not to calculate the menu styles dynamically when it is not needed (i.e. when listRef's width is higher than the min width (192px))

We can maybe refactor this to be like below?

```
sx={listRef.current?.clientWidth && listRef.current.clientWidth >= 192 
? baseRules 
: menuStyles(containerRef.current, listRef.current)}
```

It would be good to take the 192 into const and reuse it (my example is bit of a pseudo code sorry 😄 )

broccolinisoup: We will need to export the `baseRules` as well from `styles.ts`.

packages/react/src/UnderlineNav/styles.ts
```diff
@@ -135,17 +137,34 @@ export const menuItemStyles = {
...
+    zIndex: 1,
+    top: '90%',
+    boxShadow: '0 1px 3px rgba(0, 0, 0, 0.12), 0 1px 2px rgba(0, 0, 0, 0.24)',
+    borderRadius: '12px',
+    backgroundColor: 'canvas.overlay',
+    listStyle: 'none',
+    // Values are from ActionMenu
+    minWidth: `${minStyleWidth}px`,
+    maxWidth: '640px',
+  }
```
broccolinisoup, line 160:
```suggestion
  export const baseRules = {
    position: 'absolute',
    zIndex: 1,
    top: '90%',
    boxShadow: '0 1px 3px rgba(0, 0, 0, 0.12), 0 1px 2px rgba(0, 0, 0, 0.24)',
    borderRadius: '12px',
    backgroundColor: 'canvas.overlay',
    listStyle: 'none',
    // Values are from ActionMenu
    minWidth: `${minStyleWidth}px`,
    maxWidth: '640px',
    right: '0'
  }
```

broccolinisoup: exporting the baseRules with the `right: '0'` for the default styles.

packages/react/src/UnderlineNav/styles.ts
```diff
@@ -135,17 +137,34 @@ export const menuItemStyles = {
...
+
+  // Do we have less room in the UnderlineNav container then what we have reserved with our minWidth?
+  // If so, do not position the element using right, as it will overflow. Let's use `getAnchoredPosition` instead.
+  if (containerRef && listRef && listRef.clientWidth < minStyleWidth) {
```
broccolinisoup, line 164: Because we are doing the calculation earlier, we won't need this here anymore.

I think the code should like this

```jsx
const {left} = getAnchoredPosition(containerRef, listRef, {align: 'start', side: 'outside-bottom'})
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  const {right, ...rest} = baseRules
  return {...rest, left}
```

packages/react/src/UnderlineNav/styles.ts
```diff
@@ -135,17 +137,34 @@ export const menuItemStyles = {
...
+    const {left} = getAnchoredPosition(containerRef, listRef, {align: 'center', side: 'outside-left'})
```
broccolinisoup, line 165: I also played with these align and side values and updated the side to be `outside-bottom` because I think `outside-left` was causing an unexpected shift when the menu is open. (I am not 100% sure so let me know if I am missing anything here.) Here is a screen recording to explain the behaviour.

https://github.com/primer/react/assets/1446503/4097b7a7-ab5f-452c-b5b6-ab0dc9fd5a7d

_Video preview: example underlinenav menu has 'README', 'Code of Conduct', 'MIT Licence', 'Security' options with the two icon buttons aligned on the right side. The overflow menu appears when there is an overflow as the screen shrinks towards the left side. Overflow menu pops up towards the left side with its right edge aligning with the more button's right edge. When the screen gets small enough, the menu shifts towards the right side and its left edge now is aligning with the more button's right edge. As screen gets smaller, the menu repositions itself to bottom center._

```jsx
const {left} = getAnchoredPosition(containerRef, listRef, {align: 'start', side: 'outside-bottom'})
```

Also updating `align`  to `'start'` helps us make sure the menu is always aligning with the beginning of the list when there is not enough padding/margin around the navigation.

broccolinisoup (commented):
Hello @Rebstorm 👋🏻 Thanks so much for following up with the review comments 🙏🏻 

I love your thinking on the solution! This is great. I have two main comments, one is about the position of the menu and the other is the styling performance. Please let me know if there is anything I can elaborate or if I misunderstand anything from your suggested solution. Thanks so much for your time and your contribution 🙏🏻

packages/react/src/UnderlineNav/UnderlineNav.tsx
Rebstorm: Alright. Sounds good to me. I've moved the calculation to a ternary (like you suggested) outside of the function, as to not force the calculation for each render. 

Ive set the 192 to a const that I called `baseMenuMinWidth`. I hope that explains the variable well enough.

packages/react/src/UnderlineNav/styles.ts
Rebstorm: I added `right: '0'` back to the `baseMenuStyles`. :100:

packages/react/src/UnderlineNav/styles.ts
Rebstorm: Oh thank you thank you thank you! I was struggling with this, I couldnt find a combination that would work better then aligning it `center` and using `outside-bottom`. Your solution seem to have fixed what I was struggling with however, appreciate it :pray:

Rebstorm:
To summarize the current fix, based on @broccolinisoup's feedback:

1. I moved the clientWidth calculation outside of the menuStyles to not force the getAnchoredPosition to run on each render.
2. Created a constant for the min-width called baseMenuMinWidth. The baseMenuMinWidth is now referenced as the value we check, use, or test the 192px min-width. That way, we only use a single value.
3. Created a simple constant for the base menuStyles called baseMenuStyles, which contains right: '0'.
4. Simplified the test case as the menuStyles is now simpler.
5. The getAnchoredPosition now uses {align: 'start', side: 'outside-bottom'} for placement.

Again, thank you for a clear response and great feedback. This has been fun working on (and I hope it did not cause too much overhead for you), and if there's something missing - let me know!

broccolinisoup (commented):
Thanks so much for following up with the comments and pushing this PR to the end @Rebstorm! It looks great 🎉 

Just a note: I'll hold off merging this (this is why I am commenting on the PR instead of approving) for a few days to test some action issues on the fork PRs. 

Thanks again for raising this pull request and contributing to Primer 🥳 🎈

broccolinisoup:
@joshblack I was testing https://github.com/primer/react/pull/4235, it cancels the deploy preview (fork) action in this PR. Do you happen to know why that is failing to run? Also what is the best way you recommend to debug cancelled actions? I usually have a hard time debugging them, well, because there is not much proper logs 🫠

broccolinisoup (approved):
Hi @Rebstorm 👋🏻 Apologise for the delay here. I am going ahead merging the PR 🎉 

Thanks so much again for your contribution 🙏🏻

PR #4235, opened 2024-02-08
ci: update deploy_preview_forks workflows

joshblack:
Noticed this after the merge recently, this PR updates our workflow for deploying a preview from forks to only run `npm ci` at the top level and no longer install from the docs workspace specifically

#### Changed

- Update `deploy_preview_forks.yml` workflow to only install dependencies once

### Rollout strategy

This is a change to our CI processes

### Testing & Reviewing

- Not sure how to specifically test this without using a PR from a fork. Thankfully we have some PRs opened now from forks that we can use to test this once merged

### DIFF
--- a/.github/workflows/deploy_preview_forks.yml
+++ b/.github/workflows/deploy_preview_forks.yml
@@ -21,9 +21,6 @@ jobs:
           cache: 'npm'
       - name: Install dependencies
         run: npm ci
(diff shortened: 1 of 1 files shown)


broccolinisoup (approved):
Looks good! 🚢  I'll test this on https://github.com/primer/react/pull/4234 and report back if there is any issues.

PR #4236, opened 2024-02-08
refactor(react): remove circular dependencies

joshblack:
This PR removes circular dependencies that we have in `@primer/react`. It also updates our rollup config to throw an error when we encounter one when bundling. Circular dependencies in an ESM context create issues where you are unable to use an imported identifier because it has not been defined in the module context yet.

#### Changed

- Update imports that point directly to `src/index.ts` and replaced them with imports directly to the module
- Update cycles by moving shared code into a shared file, either an existing one or by creating a new one 

### Rollout strategy

- [x] Minor release

Technically this should be a refactor and no behavior should change but, just to be safe, I think this should go out under a minor release to make sure

### Testing & Reviewing

- Verify that the refactor doesn't break any references or functionality 😅


broccolinisoup (approved):
This is great! Thanks so much for taking care of this ✨

PR #4237, opened 2024-02-08
refactor(tsconfig): enable verbatimModuleSyntax

joshblack:
Based on changes over in: https://github.com/github/github/pull/293430

This PR enables the [`verbatimModuleSyntax` option](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) in our base `tsconfig`. This change prevents some of the challenges we were running into where Gatsby or Storybook were reporting issues importing/finding certain imports that were using type elision (where TS detects that you're importing a type but bundlers may not be aware that it is a type).

Note: this still allows mixed imports as long as `type` is included before the named import. For example:

```tsx
import { ExampleExport, type ExampleExportProps } from './example';
```

#### Changed

- Enable `verbatimModuleSyntax` option in `tsconfig.base.json`
- Enable eslint rule for this syntax
- Run autofix for eslint rule to refactor imports 

### Rollout strategy

I believe no public change should be perceivable through this refactor

### Testing & Reviewing

- Would love to hear what folks think of enabling this rule and its impact on PRC
- Are there any downsides to enabling this rule?


siddharthkp (approved):
TIL! 

I don't know the tradeoffs of this, but trust you with it!

Just confirming, I understand this does not change how types can imported outside the package, right?

joshblack:
@siddharthkp correct! Should be no change at all downstream. I think that, for us, the only drawback is that we have to explicitly annotate type imports.

PR #4238, opened 2024-02-08
docs(examples): add vite template for codesandbox

joshblack:
Add an example that we can use in Codesandbox to automatically create an environment for using `@primer/react`. This could be for issues, playing around with ideas, or anything else and is automatically linked the source of this package in our project.

Any example can be tied to a Codesandbox URL, for this PR the URL is: https://codesandbox.io/p/devbox/github/primer/react/tree/docs/add-codesandbox-example/examples/codesandbox

In the future, we would use:

```
https://codesandbox.io/p/devbox/github/primer/react/tree/main/examples/<example-folder>
```

#### New

- Add codesandbox to examples using vite as a template

### Rollout strategy

This is a change to our internal examples

### Testing & Reviewing

- [ ] What do you think of this strategy for working with `@primer/react`?
- [ ] Do you like using `vite` for the example? Should we use something else? 
- [ ] Do you want anything included in the example to make it easier to re-create issues or work with?


joshblack:
Note, this PR most likely needs: https://github.com/primer/react/pull/4236 to work locally since there are some circular dependencies that it cannot seem to resolve 😞

joshblack:
Depends on: https://github.com/primer/react/pull/4261

siddharthkp (approved):
Love that!

> What do you think of this strategy for working with @primer/react?

Yes please!

> Do you like using vite for the example? Should we use something else?

Super idealistic, we should use similar setup to gh/gh. But as a playground, I'm happy to use whatever is fast!

> Do you want anything included in the example to make it easier to re-create issues or work with?

Minimal is good! (non-blocking) Do you think we should make it SSR by default?

joshblack:
@siddharthkp great point with the SSR, we do have an App Router example from Next.js as an alternative that could make that process easier. Happy to switch the codesandbox example over to that. When looking at [Vite's example for SSR](https://github.com/bluwy/create-vite-extra/tree/master/template-ssr-react) it seemed to involve some extra steps that doesn't make it as worth it, at least to me. Curious what your thoughts are

PR #4241, opened 2024-02-09
Adds new component TabPanels

owenniblock:
### This Work

Closes https://github.com/github/primer/issues/3043

Adds a new TabPanels component to mirror the Primer View Component TabPanels. We will need to retrospectively go back and update PVC to use the new `<tab-container>` which supports shadow DOM.

#### Screenshots

Three tabs, the first is selected with the first Panel visible
Three tabs, the first is selected, the first panel is visible. Additional content is a button after the tab controls and some text after the tab page
Three tabs, the second is selected with the second Panel visible

### Changelog

* Adds new TabPanels alpha component to provide non-navigational tabs with pages.

#### New

* TabPanels component

#### Changed

* Seems like we were missing some snapshots for Dialog-v1

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

I'd particularly like guidance on the following:

* If any of my CSS could be improved
* If there are any additional features of `tab-container-element` which we should support
* If I've missed any tests

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


owenniblock:
[dotcom - primer integration workflow](https://github.com/github/github/actions/runs/7990186657)

owenniblock:
We'll need to retrospectively update the PVC components to use the new shadow DOM version of `@github/tab-container-element`. @keithamus do we want to track that as part of the Epic? Or separately?

joshblack (commented):
Super exciting to see! 🥳 Congrats on the contribution @owenniblock Had ~~some~~ way too many notes as I was going through this, would love to hear what you think 👀 Also let me know if there is a better way to talk through these 😅 

## Package

From a package perspective, it'd be great if this lived in `src/drafts/index.ts` or `src/experimental/index.ts` since this is an alpha component. Components in the `src/index.ts` entrypoint are considered stable and would need to follow semver

## Accessibility

Had a couple of observations that I was curious about from an a11y lenses:

- Would it be possible to make the `aria-label` or `aria-labelledby` required so that the `role="tablist"` element always has a label?
- Should the element with `role="tabpanel"` have `aria-labelledby` set to the corresponding tab?
- Could we make the tabindex on the `role="tabpanel"` dynamic with respect to if it has interactive content? So if there is no interactive content in the panel, it has a tabindex. If it does have interactive content, then it does not
- While it's only supported in JAWS, would it be worth adding `aria-controls` to the `role="tab"` elements?
- I had some issues in JAWS getting the tab panel to announce 🤔 Curious what the expected flow would be here (or if my test procedure was wrong!)
- Would it be possible for us to address the axe issue with `role="tab"` not being in `role="tablist"` so that downstream users won't also have to disable this rule?

## Server-side rendering, initial render

- There seems to be a flash of unstyled content on slower network speeds or in browsers like Firefox

https://github.com/primer/react/assets/3901764/f0adb260-b9fc-434c-b083-ebae553a3aca

- There seem to be hydration errors when SSR'd

image

- I also ran into issues when using this with Next.js' App Router but couldn't really determine why 🤔 It may be unrelated but here was the message

```
Error: Could not find the module "[...]/primer/react/packages/react/lib-esm/index.js#TabPanels#Panel" in the React Client Manifest
````

## Customization

Would it be possible to configure the tablist itself? For example, if we wanted to wrap it in a `nav` tab for a navigation tabs setup (or if this not preferred?)

## Interop with React patterns

- It'd be great if we could make `onTabContainerChange` to simply `onChange`
- Do we want to support a version of controlled state instead of uncontrolled?

**Comment by owenniblock** [comment]
@joshblack thanks for those extensive notes, that's super helpful! I'll start going through them over the next few days. Will let you know pairing / a chat would help.

**Comment by owenniblock** [comment]
> it'd be great if this lived in src/drafts/index.ts

Done!

**Comment by owenniblock** [comment]
Re Accessibility stuff:

I've been thinking that it would be good to bring this to an Accessibility Office Hours (maybe next week). Here are my notes on the things you raised 🙂 

> make the aria-label or aria-labelledby required

I like this idea! We're aiming to make this component as close to the Primer View Component version as possible (which also has an optional aria-label I believe). We could back-port the required feature which would be a breaking change in PVC so might need to tread carefully there (I'd appreciate @keithamus' thoughts here also).

> Should the element with role="tabpanel" have aria-labelledby set to the corresponding tab?

Yes! It used to. Maybe we lost that functionality during the conversion to shadow DOM magic. I'll check!

> Could we make the tabindex on the role="tabpanel" dynamic with respect to if it has interactive content? So if there is no interactive content in the panel, it has a tabindex. If it does have interactive content, then it does not

I like this idea, we'd need to make a change to `@github/tab-container-element` to support this (which consequently would mean that the same change gets reflected in PVC).

> While it's only supported in JAWS, would it be worth adding aria-controls to the role="tab" elements?

Not sure, I'll investigate! 👍 

> I had some issues in JAWS getting the tab panel to announce 🤔 Curious what the expected flow would be here (or if my test procedure was wrong!)

Could be related to the missing labelledby 🤔 I'll do some digging. I don't have JAWS on my Windows machine (but I guess I should!)

> Would it be possible for us to address the axe issue with role="tab" not being in role="tablist" so that downstream users won't also have to disable this rule?

This is tricky because it's not _actually_ an issue. The Accessibility Tree is correct but because it uses shadow dom and slots etc, axe doesn't recognise it correctly. It _might_ be possible to contribute a fix to axe so that it works with components leveraging shadow dom but I'm not certain.

**Comment by keithamus** [comment]
> > There seems to be a flash of unstyled content on slower network speeds or in browsers like Firefox
> 
> Ah, @keithamus maybe we need to have users set `hidden` to the correct tab panels initially 🤔 I was hoping we wouldn't have to but I'm pretty that that's what's causing this.

We could explore CSS options here. E.g. `tab-panel:not(:defined) [role=panel]:not(:first-child) { display: none }` would probably get us 99% of the way there. (There'd be a potential flash if e.g. the second tab was selected, perhaps some more CSS could fix that).

> I _think_ we might need to make a small change to `@github/tab-container-element` so that it removes `hidden` from the panel in the DOM when adding it to the shadow slot (unless it already does this, I will check!)

It does this already.

**Comment by keithamus** [comment]
>     * Could we make the tabindex on the `role="tabpanel"` dynamic with respect to if it has interactive content? So if there is no interactive content in the panel, it has a tabindex. If it does have interactive content, then it does not

This seems very tricky to properly determine, and seems like it would create an undesirable experience? For example in use on the CommentEditor the `Preview` pane _sometimes_ has interactive content if there is a link or code block in the message, while other times it doesn't. I'm not certain we want this but I'd love to hear more.

>     * There seem to be hydration errors when SSR'd

We might have to add `suppressHydrationWarning` for this component as the custom element will set things like tabindex, which is not possible to do without mutating the DOM (unless we also reapply the same logic to the React component but that's doing work twice). We can selectively add it to the tabs and panels or just on the whole component.

> Would it be possible to configure the tablist itself? For example, if we wanted to wrap it in a `nav` tab for a navigation tabs setup (or if this not preferred?)

This is not preferred. For those uses people should reach for `TabNav`. We should document accordingly. Having said that this is entirely possible in the custom element, you can supply your own `role=tablist` and it'll respect that.

>     * It'd be great if we could make `onTabContainerChange` to simply `onChange`

Arguably we should do this in the custom element too.

**Comment by owenniblock** [comment]
@joshblack I spent some time with @keithamus this morning experimenting with CSS options for the flash on content issue (thanks Keith!)

Would love you to give it a test, we've had to compromise in a couple of places because the structure of the tablist & additional content elements doesn't exist yet. So there are the following limitations:

* The underline of the tabs only covers the tabs themselves and not the full width of the component (also noticed that there's a very small gap in the line between each tab, might look at that 🤔)
* If you have 11 or more tabs and load the component with the 11th (or higher) tab selected, the tab panel doesn't show until the component loads.

I'll need to retest in various browsers, so let me know if this approach seems viable to you and then I'll do that 🙂

File: docs/content/TabPanels.mdx [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+
+```jsx live
+<TabPanels aria-label="Select a tab">
+  <TabPanels.Tab>Tab 1</TabPanels.Tab>
+  <TabPanels.Tab>Tab 2</TabPanels.Tab>
+  <TabPanels.Tab>Tab 3</TabPanels.Tab>
+  <TabPanels.Panel>Panel 1</TabPanels.Panel>
+  <TabPanels.Panel>Panel 2</TabPanels.Panel>
+  <TabPanels.Panel>Panel 3</TabPanels.Panel>
+</TabPanels>
```
iansan5653, line 25: Suggestion: if every `Panel` should always have a `Tab` and vice versa (a 1:1 relationship) could we combine the two into a single component? This would enforce safe usage ensuring the 1:1 relationship is always maintained and things are properly linked.

With this API it's pretty easy to accidentally break the 1:1 relationship & linking (as it is here actually - the tabs are not linked to the panels by `aria-labelledby` in this example).

In other words, the API would look something like this:

```suggestion
<TabPanels aria-label="Select a tab">
  <TabPanels.TabPanel name="Tab 1">
    Panel 1
  </TablPanels.TabPanel>
  <TabPanels.TabPanel name="Tab 2">
    Panel 2
  </TablPanels.TabPanel>
  <TabPanels.TabPanel name="Tab 3">
    Panel 3
  </TablPanels.TabPanel>
</TabPanels>
```

Under the hood it should be pretty straightforward to have `TabPanel` render `Tab` and `Panel`:

```tsx
const TabPanel = ({name, children}) => {
  const id = useId()
  
  return <>
    <TabPanel.Tab id={id}>{name}</TabPanel.Tab>
    <TabPanel.Panel aria-labelledby={id}>{children}</TabPanel.Tab>
  </>
}
```

Although if the order of the elements is important (panels must follow tabs) it would get a bit more complicated. But even then I think it would still be worth it to have a cleaner API.

keithamus: We could offer this API just within React quite easily but as @owenniblock points out we lose out on the ability to easily add content before or after the tabs, without doing a bunch of juggling of props and I don't know how prevalent it is to stick JSX in props but it doesn't make for a great API (IMO stuffing JSX in a prop would be an anti-pattern). In other words for very simple cases (as described above) it looks like a more desirable API but as we start to do the things we need the JSX devolves into a bit of a mess:

```jsx
<TabPanels aria-label="Select a tab">
  {/* each one of these could have icons or other markup that's kind of janky as a prop */}
  <TabPanels.TabPanel name={<><OctocatIcon /> Tab 1</>}>
    Panel 1
  </TablPanels.TabPanel>
  <TabPanels.TabPanel name={<><OctocatIcon /> Tab 2</>}>
    Panel 2
  </TablPanels.TabPanel>
  {/* I guess the best way to handle the Before/After tabs stuff would be another sub-component */}
  <TabPanels.AfterTabs>
    <MarkdownToolbar/>
  </TabPanels.AfterTabs>
  {/* Then because _eveything_ is out of tree order, anything after tab panels also needs a sub-component */}
  <TabPanel.AfterPanels>
    <ExtraButtons/>
  </TabPanel.AfterPanels>
</TabPanels>
```

vs 

```jsx
<TabPanels aria-label="Select a  tab">
  {/* Tabs don't need JSX props, can just render children */}
  <TabPanels.Tab><OctocatIcon /> Tab 1</TabPanels.Tab>
  <TabPanels.Tab><OctocatIcon /> Tab 2</TabPanels.Tab>
  {/* Stuff after tabs goes after the tabs in the tree */}
  <MarkdownToolbar/>
  <TablPanels.Panel>
    Panel 1
  </TablPanels.Panel>
  <TabPanels.Panel>
    Panel 2
  </TablPanels.Panel>
  {/* Stuff after panels goes after the panels in the tree */}
  <ExtraButtons/>
</TabPanels>
```

owenniblock:
Signed up for [Wednesday's Accessibility Office Hours](https://github.com/github/accessibility/issues/5586) to check it all looks OK 🤞

docs/content/TabPanels.mdx
iansan5653: > we're aiming for feature parity with Primer View Components here, so any API change we add here will need to be ported over to PVC in future

I think this is an unfortunate approach because ViewComponents and React are different paradigms. ViewComponents are consumed in HTML, which is not typechecked and doesn't have embedded JS. React components are consumed in typechecked JavaScript. This naturally means the APIs should be different - writing a React API to be the same a ViewComponent API is going to result a worse developer experience.

> IMO stuffing JSX in a prop would be an anti-pattern

FWIW this is a common practice and I personally don't think it's problematic at all, though admittedly it's not beautiful. This was actually recommended by the official React docs until very recently (it got refactorted out as they built the new site). It's flexible and allows for accepting strings for the basic case while still accepting full JSX for advanced cases.

In practice, the example with the current API really looks like this:

```tsx
const tab1Id = useId()
const tab2Id = useId()

<TabPanels aria-label="Select a  tab">
  <TabPanels.Tab id={tab1Id}><OctocatIcon /> Tab 1</TabPanels.Tab>
  <TabPanels.Tab id={tab2Id}><OctocatIcon /> Tab 2</TabPanels.Tab>
  <MarkdownToolbar/>
  <TablPanels.Panel aria-labelledby={tab1Id}>
    Panel 1
  </TabPanels.Panel>
  <TabPanels.Panel aria-labelledby={tab2Id}>
    Panel 2
  </TabPanels.Panel>
  <ExtraButtons/>
</TabPanels>
```

While it is flexible, it's problematic:

- It's difficult to remember where the IDs go (tabs or panels?)
- It's easy to mix up the IDs and link the wrong tabs with panels
- There's no clear link between the tabs & panels - as a reader who hasn't seen this before, would you intuitively understand that the tabs control the panels?

If the current level of flexibility is a requirement, I have a few suggestions that I think will improve the current API:

1. Make the `id` and `aria-labelledby` props required in the TypeScript types so they cannot be left out accidentally
2. Rename the `aria-labelledby` prop to `tabId`
3. Throw a runtime error if there is not a valid 1:1 link between tabs and panels, so we can catch these issues early
4. (This one is a little more complex) Create a new hook that returns tab and panel components already linked together, generating a unique ID under the hood

In practice (3) might look like this:

```tsx
const [Tab1, TabPanel1] = useTabPanel()
const [Tab2, TabPanel2] = useTabPanel()

<TabPanels aria-label="Select a  tab">
  <Tab1><OctocatIcon /> Tab 1</Tab1>
  <Tab2><OctocatIcon /> Tab 2</Tab2>
  <MarkdownToolbar/>
  <TabPanel1>
    Panel 1
  </TabPanel1>
  <TabPanel2>
    Panel 2
  </TabPanel2>
  <ExtraButtons/>
</TabPanels>
```

Alternatively if we prefer namespacing it could look like this:

```tsx
const Tab1 = TabPanels.useTab()
const Tab2 = TabPanels.useTab()

<TabPanels aria-label="Select a  tab">
  <Tab1.Tab><OctocatIcon /> Tab 1</Tab1.Tab>
  <Tab2.Tab><OctocatIcon /> Tab 2</Tab2.Tab>
  <MarkdownToolbar/>
  <Tab1.Panel>
    Panel 1
  </Tab1.Panel>
  <Tab2.Panel>
    Panel 2
  </Tab2.Panel>
  <ExtraButtons/>
</TabPanels>
```


I think this is more readable, clearly shows the link between the tabs and panels, and has fewer pitfalls. It means the consumer doesn't have to think about creating an ID, ensuring that it's unique, and knowing where to put it.

iansan5653:
A small accessibility concern I noticed in the demo - it doesn't look like we are wrapping the tabs in a [`tablist`](https://w3c.github.io/aria/#tablist) container. This is a required parent for the [`tab`](https://w3c.github.io/aria/#tab) role.

mperrotti:
Just a heads-up, we just started a project to wrangle all of our different kinds of navigation. The current direction I'm exploring removes `TabNav` and `TabPanels` and just uses `UnderlineNav` and `UnderlinePanels`.

[Google doc for context](https://docs.google.com/document/d/1SzoM3SyWoPK3NRMJ6KKfnGdBLwlSCHkF-GJTi9yq_YI/edit#heading=h.my5g4lufu5ka)
[WIP Explorations in Figma](https://www.figma.com/file/VwrXIELc8jtXXhY2AiHsfQ/Navipocalypse-explorations?type=design&node-id=89-48639&mode=design&t=Cl8yEhHhAdirlxkL-4)

docs/content/TabPanels.mdx
owenniblock: @iansan5653 thanks for the valuable thoughts, I'm working on a change so that the `TabPanels` element requires an `id` and then that `id` is used to auto generate ids for the tabs, and wire up the `aria-labelledby` as well.

I'll let you know when that's pushed up so you can check it out and let me know if you think that addresses some of this feedback.

owenniblock:
> A small accessibility concern I noticed in the demo - it doesn't look like we are wrapping the tabs in a [tablist](https://w3c.github.io/aria/#tablist) container. This is a required parent for the [tab](https://w3c.github.io/aria/#tab) role. This might need to be something we fix in the custom element?

@iansan5653 this is actually handled in the custom element, the `tablist` role is added in the Shadow DOM and isn't required in the Light DOM. I recently fixed an issue where I was setting the `tablist` role on a `slot` element (which is invalid) and now this passes accessibility checks. We should get additional feedback from screen reader users soon as well.

owenniblock:
@mperrotti I'll look through those items. Is there a good place for feedback and/or discussion? I have some concerns that we'll lose important nuances around the underlying semantics of the controls we already have. For example I know we added a separate component in Rails (TabPanels and TabNav) because folks didn't know to only use TabNav for navigation elements. I'll dig into this in more detail wherever is the best place for the discussion 🙂

mperrotti:
@owenniblock - you can add feedback in the Figma or send me feedback on Slack. I also brought this to a11y office hours earlier this week https://github.com/github/accessibility/issues/5623#issuecomment-1967930432

We could still have "nav" vs "panels" distinction for the nuances around the underlying semantics.

iansan5653:
> > A small accessibility concern I noticed in the demo - it doesn't look like we are wrapping the tabs in a [tablist](https://w3c.github.io/aria/#tablist) container. This is a required parent for the [tab](https://w3c.github.io/aria/#tab) role. This might need to be something we fix in the custom element?
> 
> @iansan5653 this is actually handled in the custom element, the `tablist` role is added in the Shadow DOM and isn't required in the Light DOM. I recently fixed an issue where I was setting the `tablist` role on a `slot` element (which is invalid) and now this passes accessibility checks. We should get additional feedback from screen reader users soon as well.

Got it, thanks! I was confused by the DOM in the browser devtools, but looking at the accessibility tree confirms that there is in fact a `tablist`:

Chrome accessibility tree inspector showing a 'tablist' labelled 'Select a tab' with three labeled tabs. 'Tab 1' is selected. Outside the tablist is a 'tabpanel' labelled 'Tab 1'.

packages/react/src/drafts/TabPanels/TabPanels.tsx
```diff
@@ -0,0 +1,161 @@
...
+import {TabContainerElement} from '@github/tab-container-element'
+import {createComponent} from '../../utils/custom-element'
+import sx, {type SxProp} from '../../sx'
+import type {ComponentProps} from '../../utils/types'
+import getGlobalFocusStyles from '../../internal/utils/getGlobalFocusStyles'
+
+const TAB_CLASS = 'TabPanel-tab'
+
+const TabContainer = createComponent(TabContainerElement, 'tab-container')
+const StyledTabContainer = styled(TabContainer)`
```
owenniblock, line 15: @keithamus let me know if there's a better way to do this. Initially I tried wrapping the previous line but it gave me an error:

> Type instantiation is excessively deep and possibly infinite

Moving the call to a separate line seems to confuse the TypeScript typing stuff less 🤔

owenniblock:
We were going to change the events to `onChange` and `onChanged` - do we want to do that as part of this PR?

owenniblock:
[Integration fails](https://github.com/github/github/actions/runs/8189497014/job/22394464686?pr=316000)

```   routes/HistoryComparisonPage.tsx(95,15): error TS2339: Property 'relative-time' does not exist on type 'JSX.IntrinsicElements'.
  routes/HistorySummaryPage.tsx(86,21): error TS2339: Property 'relative-time' does not exist on type 'JSX.IntrinsicElements'.
  npm ERR! Lifecycle script `tsc` failed with error: 
  npm ERR! Error: command failed 
  npm ERR!   in workspace: @github-ui/repos-rules 
  npm ERR!   at location: /workspace/github/ui/packages/repos-rules ```

File: packages/react/src/drafts/TabPanels/TabPanels.tsx [hunk 3]
```diff
@@ -0,0 +1,160 @@
...
+function TabPanels({children, id, ...props}: TabPanelsProps) {
+  // Loop through the chidren, if it's a tab, then add id="{id}-tab-{index}"
+  // If it's a panel, then add aria-labelledby="{id}-tab-{index}"
+  let tabIndex = 0
+  let panelIndex = 0
+
+  const childrenWithProps = React.Children.map(children, child => {
+    if (React.isValidElement<TabPanelsTabProps>(child) && child.type === Tab && !child.props.id) {
+      if (props.selectedTabIndex === tabIndex) {
+        return React.cloneElement(child, {id: `${id}-tab-${tabIndex++}`, selected: true})
```
iansan5653, line 86: It looks like if `id` is not set, this will be `undefined-tab-1` since the value will be `undefined`. That could be problematic if it happens twice on a page. 

Maybe we can use `useId` to generate a guaranteed unique fallback? 

```tsx
const id = props.id ?? React.useId()
```

packages/react/src/drafts/TabPanels/TabPanels.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+function TabPanels({children, id, ...props}: TabPanelsProps) {
+  // Loop through the chidren, if it's a tab, then add id="{id}-tab-{index}"
+  // If it's a panel, then add aria-labelledby="{id}-tab-{index}"
+  let tabIndex = 0
+  let panelIndex = 0
+
+  const childrenWithProps = React.Children.map(children, child => {
+    if (React.isValidElement<TabPanelsTabProps>(child) && child.type === Tab && !child.props.id) {
+      if (props.selectedTabIndex === tabIndex) {
```
iansan5653, line 85: Because the selection logic is inside the conditional with `!child.props.id` it appears we won't assign `selected: true` if the `id` prop is set on the `Tab` (ie, if the consumer is using a custom ID here for some reason). It looks like this would break the tab activation behavior?

packages/react/src/drafts/TabPanels/TabPanels.docs.json
iansan5653, line 1: I'm not sure where the best place to call this out is, but I think it's really important that we clearly document how the tabs and panels are linked - it's uncommon for the orders to be dependent on each other, so we should be clear that the tabs need to be in the same order as the panels. I think it's also worth calling out any other restrictions. Specifically: Do the panels need to follow the tabs, or could you put the tabs right next to their corresponding panels? And what happens to content you put before/after/in between that isn't a tab or panel?

packages/react/src/utils/custom-element.ts
```diff
@@ -26,4 +26,7 @@ export const createComponent = <I extends HTMLElement, E extends EventNames = {}
     },
   )
 
+  return output as React.ComponentType<React.ComponentProps<typeof output> & {sx?: unknown}>
```
iansan5653, line 29: should `sx` be the standard `BetterSystemStyleObject` type here? From `SxProp`?:

```suggestion
  return output as React.ComponentType<React.ComponentProps<typeof output> & SxProp>
```

packages/react/src/drafts/TabPanels/TabPanels.docs.json
```diff
@@ -0,0 +1,72 @@
...
+      "type": "string",
+      "description": "The id of the tab container, used to generate child ids."
+    },
+    {
+      "name": "selectedTabIndex",
+      "type": "number",
+      "description": "The 0-based index of the tab that is selected."
+    },
+    {
+      "name": "onTabContainerChange",
```
keithamus, line 29: This has an alias to `onChange` right?

packages/react/src/drafts/TabPanels/TabPanels.docs.json
```diff
@@ -0,0 +1,72 @@
...
+      "type": "function",
+      "description": "Callback fired when the tab container changes (bubbles, cancelable): fired on `<tab-container>` before a new tab is selected and visibility is updated. `event.tab` is the tab that will be focused and `tab.panel` is the panel that will be shown if the event isn't cancelled."
+    },
+    {
+      "name": "onTabContainerChanged",
```
keithamus, line 34: Likewise here, aliased to `onChanged`, no?

packages/react/src/drafts/TabPanels/TabPanels.docs.json
```diff
@@ -0,0 +1,72 @@
...
+      "type": "function",
+      "description": "Callback fired when the tab container changes (bubbles): fired on `<tab-container>` after a new tab is selected and visibility is updated. `event.tab` is the tab that is now active (and will be focused right after this event) and `event.panel` is the newly visible tab panel."
+    },
```
iansan5653, line 37: I don't see where these handlers get bound - does that happen automatically with `createComponent`?

iansan5653 (commented):
I like this API direction much better! Just a few notes on some potential bugs + documentation.

packages/react/src/drafts/TabPanels/TabPanels.docs.json
keithamus: yes. Specifically [we made them props on the component](https://github.com/github/tab-container-element/blob/984ef06309b29797ed52965f08135fcb94b52a8e/src/tab-container-element.ts#L52-L68).

packages/react/src/drafts/TabPanels/TabPanels.docs.json
```diff
@@ -0,0 +1,72 @@
...
+      "name": "sx",
+      "type": "SystemStyleObject"
+    }
+  ],
+  "subcomponents": [
+    {
+      "name": "TabPanels.Tab",
+      "props": [
+        {
+          "name": "selected",
```
joshblack, line 48: Is this prop still available? I couldn't seem to get it to work and wanted to check 👀

packages/react/src/drafts/TabPanels/TabPanels.docs.json
```diff
@@ -0,0 +1,72 @@
...
+      "description": "Callback fired when the tab container changes (bubbles): fired on `<tab-container>` after a new tab is selected and visibility is updated. `event.tab` is the tab that is now active (and will be focused right after this event) and `event.panel` is the newly visible tab panel."
+    },
+    {
+      "name": "sx",
+      "type": "SystemStyleObject"
+    }
+  ],
+  "subcomponents": [
+    {
+      "name": "TabPanels.Tab",
```
joshblack, line 45: Should `Tab` have support for the `id` prop?

packages/react/src/drafts/TabPanels/TabPanels.docs.json
```diff
@@ -0,0 +1,72 @@
+{
+  "id": "tab_panels",
+  "name": "TabPanels",
+  "status": "draft",
+  "a11yReviewed": false,
+  "stories": [],
+  "props": [
+    {
+      "name": "aria-label",
```
joshblack, line 9: Could we could make `aria-label` (and I guess `aria-labelledby`) required in the types for `TabPanels` so that one of these must be set?

packages/react/src/drafts/TabPanels/TabPanels.docs.json
```diff
@@ -0,0 +1,72 @@
...
+      "type": "string",
+      "description": "Used to set the `aria-label` on the `role=\"tablist\"` element."
+    },
+    {
+      "name": "children",
+      "type": "React.ReactNode",
+      "description": "The content of the component, can contain Tabs, Panels but also content before and after Tabs and after the Panels."
+    },
+    {
+      "name": "id",
```
joshblack, line 19: Could we make the `id` prop optional here and use the `useId` hook to optionally generate one / use an `id` if it's provided?

packages/react/src/drafts/TabPanels/TabPanels.features.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+import type {ComponentProps} from '../../utils/types'
+import {Button} from '../../Button'
+
+export default {
+  title: 'Drafts/Components/TabPanels/Features',
+  component: TabPanels,
+} as Meta<ComponentProps<typeof TabPanels>>
+
+export const Selected = () => (
+  <TabPanels aria-label="TabPanels example" id="tab-panels" selectedTabIndex={1}>
```
joshblack, line 13: For the controlled case, what's the intended way to synchronize the current selected tab with the index being provided? It seems like `onTabContainerChange` provides the tab but not the index so would that mean we would need to look it up in some way?

packages/react/src/drafts/TabPanels/TabPanels.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+export type TabPanelsProps = ComponentProps<typeof TabContainer>
+
+function TabPanels({children, id, ...props}: TabPanelsProps) {
+  // Loop through the chidren, if it's a tab, then add id="{id}-tab-{index}"
+  // If it's a panel, then add aria-labelledby="{id}-tab-{index}"
+  let tabIndex = 0
+  let panelIndex = 0
+
+  const childrenWithProps = React.Children.map(children, child => {
```
joshblack, line 83: I think this technique would make it so that people couldn't wrap `Tab` in a custom component as then it wouldn't be able to infer the props from the children needed to setup `id` (or `aria-labelledby` in the panel case). For example:

```tsx
function MyCustomTab({ children }) {
  return <TabPanels.Tab onClick={/* ... */}>{children}</TabPanels.Tab>
}

<TabPanels>
  <MyCustomTab>Tab one</MyCustomTab>
  <MyCustomTab>Tab two</MyCustomTab>
  <MyCustomTab>Tab three</MyCustomTab>
  <TabPanels.Panel>Panel one</TabPanels>
  <TabPanels.Panel>Panel two</TabPanels>
  <TabPanels.Panel>Panel three</TabPanels>
</TabPanels>
```

packages/react/src/drafts/TabPanels/TabPanels.tsx
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  &:not(:defined) :not([role='tabpanel']) {
+    vertical-align: top;
+  }
+
+  ${sx};
+`
+
+export type TabPanelsProps = ComponentProps<typeof TabContainer>
```
joshblack, line 75: I'm having a hard time getting these types to suggest anything, wanted to ask is this specific to my setup or are others seeing this as well? I think a specific case is that `onTabContainerChange` is not showing up and, if defined, the `event` type is `any`.

packages/react/src/drafts/TabPanels/TabPanels.tsx
```diff
@@ -0,0 +1,160 @@
...
+import {TabContainerElement} from '@github/tab-container-element'
+import {createComponent} from '../../utils/custom-element'
+import sx, {type SxProp} from '../../sx'
+import type {ComponentProps} from '../../utils/types'
+import getGlobalFocusStyles from '../../internal/utils/getGlobalFocusStyles'
+
+const TAB_CLASS = 'TabPanel-tab'
+
+const TabContainer = styled(createComponent(TabContainerElement, 'tab-container'))`
+  & > :not([role='tabpanel']) {
```
joshblack, line 15: I'm noticing that the border will flash with our styles since we don't have access to the tablist yet and so it can only span our buttons that are inline-block. Is there any way we could have that node defined?

joshblack (commented):
Thanks for taking in all the feedback from the session this week! Left some notes on the PR, also wanted to capture some markup/behavior stuff below too.

1. There seems to be a focus stealing bug on page load so that the selected tab becomes focused on page load. Let me know if this is environment specific for me or if you're seeing it too 🤔 

    https://github.com/primer/react/assets/3901764/7c3c7cee-4232-4fee-abfc-5c94a81f34d3

2. I wanted to ask what the takeaway was from office hours around the tabindex for tabpanel behavior as I couldn't remember. APG recommends [this](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/#accessibilityfeatures:~:text=It%20is%20recommended%20that%20all%20tabpanel%20elements%20in%20a%20tab%20set%20are%20focusable%20if%20there%20are%20any%20panels%20in%20the%20set%20that%20contain%20content%20where%20the%20first%20element%20in%20the%20panel%20is%20not%20focusable.) but I know these aren't always the best recommendation.

3. TypeScript types seem to be missing for common props that might be passed into the elements, e.g. things like `onClick`. Let me know if this is more my setup versus something else

4. For uncontrolled/controlled, one pattern we have is that `default*` props are used to initialize state. So something like `defaultIndex` would initialize the state value for the first render but nothing after. For the controlled case, `selectedIndex` would be used and would change the state

5. When using it controlled with `selectedTabIndex`, it seems like changing this value will cause focus to shift to the newly selected tab. For controlled, I don't think we would want this behavior as it's not really intended to shift folks over but to change the state value in some way for when the user interacts with the component

Finally, just wanted to chime in on the ordering stuff that Ian commented on above. I feel like there are a couple of groups to tab components implementations out there (at least in React) and I was curious which technique resonated with everyone the most. Some of them:

- Used `index` and position to link tabs and panels
  - https://reach.tech/tabs/
  - https://polaris.shopify.com/components/navigation/tabs
- Used a `value` prop to keep tab and panel in sync
  - https://mui.com/material-ui/react-tabs/
  - https://www.radix-ui.com/primitives/docs/components/tabs
  - https://ui.shadcn.com/docs/components/tabs
  - https://mantine.dev/core/tabs/
- Used `id` (or equivalent) to keep in sync
  - https://ariakit.org/reference/tab-provider
  - https://workday.github.io/canvas-kit/?path=/docs/components-containers-tabs--basic

Also let me know I missed any!

---

Thanks again for doing this @owenniblock! Happy to chat about any of this on Slack or through Zoom, whatever is most helpful.

joshblack:
One separate question I wanted to ask was how should a selected tab present in WHCM? I think the border radius can tell you what tab is selected but it's a little subtle so wanted to double-check.

Screenshot 2024-03-07 at 4 21 50 PM

This 100% feels like a separate issue but wanted to bring it up to see what you all thought since I was going through it during testing.

packages/react/src/drafts/TabPanels/TabPanels.tsx
owenniblock: Yeah, I've been wondering about this. Really we don't *want* consumers to set the child ids 🤔 How would you feel about just overriding the id whether the user sets it or not?

I'd have to document it and add how the ids are generated so users could reason about what the ids would be.

packages/react/src/drafts/TabPanels/TabPanels.docs.json
owenniblock: Good call on the documentation.

You can use either pattern:

```
<TabPanels>
  <Tab />
  <Tab />
  <Panel />
  <Panel />
 ```
 
 Or:
 
 ```
 <TabPanels>
  <Tab />
  <Panel />
  <Tab />
  <Panel />
  ```
  
I'd like to encourage people to use the first pattern (because it makes adding the additional content simpler, hence all the examples and docs using that pattern). If you think it's important, we could document the second pattern and note that the first is recommended when adding additional content around the tabs etc?

File: packages/react/src/drafts/TabPanels/TabPanels.docs.json (hunk 8 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.docs.json]: [I've added an alias called `onChange` and `onChanged`](https://github.com/github/tab-container-element/blob/main/src/tab-container-element.ts#L34) so we're documenting that now instead (same binding, it's a prop on the underlying web component).

File: packages/react/src/drafts/TabPanels/TabPanels.docs.json (hunk 10 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.docs.json:45]: It's available, but really I don't think we _want_ the user to set it. [See this comment for context](https://github.com/primer/react/pull/4241#discussion_r1517591820).

File: packages/react/src/drafts/TabPanels/TabPanels.tsx (hunk 15 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.tsx:75]: I think maybe I need to redefine the events using the createContainer `Events` stuff 🤔

File: packages/react/src/drafts/TabPanels/TabPanels.docs.json (hunk 9 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.docs.json:48]: Ah, it's overwritten by the logic in the parent. I'll remove it since we only ever want to use the `selectedTabIndex`.

File: packages/react/src/drafts/TabPanels/TabPanels.docs.json (hunk 11 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.docs.json:9]: I added `aria-labelledby` and throw an error if one of them isn't set. Let me know if there's a better way to do this!

File: packages/react/src/drafts/TabPanels/TabPanels.docs.json (hunk 12 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.docs.json:24]: _done_

File: packages/react/src/drafts/TabPanels/TabPanels.features.stories.tsx (hunk 13 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.features.stories.tsx:13]: @keithamus how would you feel about adding the tabIndex to the event object that gets raised?

File: packages/react/src/drafts/TabPanels/TabPanels.tsx (hunk 14 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.tsx:92]: @joshblack can you think of a good way to do this? We could probably do something with `prototype` and `Object.getPrototypeOf` or something like that but I think that's only going to support the case where there's a single parent level. So something like this won't be possible:

```
function MyCustomParentTab({ children }) {
  return <TabPanels.Tab onClick={/* ... */}>{children}</TabPanels.Tab>
}

function MyCustomTab({ children }) {
  return <MyCustomParentTab onChange={/* ... */}>{children}</MyCustomParentTab>
}

<TabPanels>
  <MyCustomTab>Tab one</MyCustomTab>
  <MyCustomTab>Tab two</MyCustomTab>
  <MyCustomTab>Tab three</MyCustomTab>
  <TabPanels.Panel>Panel one</TabPanels>
  <TabPanels.Panel>Panel two</TabPanels>
  <TabPanels.Panel>Panel three</TabPanels>
</TabPanels>
```

File: packages/react/src/drafts/TabPanels/TabPanels.tsx (hunk 16 above)
Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.tsx:15]: I'd welcome your suggestions on this. I've got as close as I can without adding HTML additional DOM elements. We could add a `<div role="tablist">` pattern but then we'd lose the simplicity of the structure (and the ability to define `<tab><panel><tab><panel>`).

I don't feel super strongly about either of those options but if you can think of any way of doing this with what we've got right now, let me know!

Comment (owenniblock) [reply packages/react/src/drafts/TabPanels/TabPanels.tsx:15]: Having said that, when you use that alternative structure - the content shift is very noticeable... Perhaps it's best if we _don't_ allow that pattern anyway 🤔 Will discuss with @keithamus

File: packages/react/src/drafts/TabPanels/TabPanels.tsx (hunk 4 above)
Comment (iansan5653) [reply packages/react/src/drafts/TabPanels/TabPanels.tsx:98]: Hmmmm, I feel like if we have an `id` prop in the type definition, we have to respect it. However, we could 'lie' about / hide the props by excluding `id` from the prop types for the `Tab` and `Panel` components. It would still get assigned when cloned because the underlying styled component would still handle it.

File: packages/react/src/drafts/TabPanels/TabPanels.docs.json
Comment (iansan5653) [reply packages/react/src/drafts/TabPanels/TabPanels.docs.json:1]: I think the second pattern is likely to be more popular in practice because it keeps the related content close together; particularly for use-cases where there's no additional content. I definitely think it's worth noting both and the advantages/disadvantages of each.

File: packages/react/src/drafts/TabPanels/TabPanels.tsx (hunk 14 above)
Comment (iansan5653) [reply packages/react/src/drafts/TabPanels/TabPanels.tsx:96]: To be fair, we do already have this limitation with other PRC components that use the `slots` pattern, so I guess it's acceptable. But I agree I've never loved this restriction - it's frustrating because it prevents abstraction.

I'm not sure what a better solution might be in this case though - if you are manipulating `children`, you naturally run into these problems unfortunately. The better way would be to find a solution that avoids manipulating the children [as noted in the API docs](https://react.dev/reference/react/Children) but that's a tricky problem in this case 🤔

File: packages/react/src/drafts/TabPanels/TabPanels.features.stories.tsx (hunk 13 above)
Comment (keithamus) [reply packages/react/src/drafts/TabPanels/TabPanels.features.stories.tsx:13]: The "look it up in some way" is `event.target.selectedTabIndex`. But I'm fine to expose it as `event.selectedTabIndex`.

File: packages/react/src/drafts/TabPanels/TabPanels.tsx (hunk 14 above)
Comment (joshblack) [reply packages/react/src/drafts/TabPanels/TabPanels.tsx:96]: Totally agreed @iansan5653, it's something that it'd be great to avoid in new components as it makes composition and abstraction limited 😞 

@owenniblock I think one approach could be to use context for this. In situations where people use indexes, the context value may be the index itself. For example:

```tsx
const ItemContext = createContext();

function List({ children }) {
  return (
    <div>
      {Children.map(children, (child, index) => {
        return <ItemContext.Provider value={index}>{child}</ItemContext.Provider>
      })}
    </div>
  );
}
```

This allows for any number of custom components around the item but that item will still know what position it is in. The downside of this, unfortunately, is that things like fragments won't work.

Another strategy that seems common is that items have a unique identifier prop, such as `value` or `id`, that is then used to communicate with a root level context when something changes.

```tsx
const ListContext = createContext();

function List({ children }) {
  return <ListContext.Provider value={/* ... */}>{children}</ListContext.Provider>
}

function Item({ children, value }) {
  const context = useContext(ListContext);

  return <button onClick={() => context.someAction(value)}>{children}</button>
}
```

Hope those examples make sense!

packages/react/src/drafts/TabPanels/TabPanels.tsx
joshblack: Sadly nothing comes to mind on my end @owenniblock 😞 That block level element is definitely what would give that full border effect. From an API perspective, I definitely don't mind having a tab list component and it seems common across other implementations.

packages/react/src/drafts/TabPanels/TabPanels.docs.json
joshblack: I think `id` is a tricky one since generally we allow for folks to provide it and have a fallback if they don't. This pattern makes use of the `useId()` hook in the codebase (the one in `internal/hooks` versus the one directly from React) so when authoring it roughly looks like:

```tsx
function SomeComponent({ id }) {
  const someId = useId(id);
}
```

I think if it's a part of the public API then the component should follow this behavior but if it's not part of the public API then we wouldn't need to worry about it until someone requests it 👍

packages/react/src/drafts/TabPanels/TabPanels.docs.json
joshblack: @owenniblock what would you think of something like this at the type level that could be a part of the component props? Was wondering if this would work or if there was a case that this wouldn't cover 👀 

```ts
type Label = {
  'aria-label': string;
  'aria-labelledby'?: never;
}

type Labelledby = {
  'aria-label'?: never;
  'aria-labelledby': string;
}

type Labelled = Label | Labelledby;
```

packages/react/src/drafts/TabPanels/TabPanels.tsx
iansan5653: IMO, I still think the most straightforward API with the least drawbacks is using a single component for the tabs and panels, like:

```tsx
<TabPanel tab={<><Icon /> Tab name</>}>
  Tab contents
</TabPanel>
```

I [proposed](https://github.com/primer/react/pull/4241#discussion_r1501104692) this earlier and the suggestion was rejected because:

1. Putting JSX in a prop isn't considered a great API
2. It doesn't provide as much flexibility around inserting custom content in to the tab bar

For (1) we can consider the React `Suspense` API which does exactly the same thing with its `fallback` prop. Perhaps it's not beautiful but it's standard and idiomatic. 

(2) is interesting but I think it's worth reconsidering how much customization we really need. Can we just say that any non-tab content is rendered in the tab bar where it goes? 

```tsx
Content before panel
<TabPanels>
  Content before tabs in tab bar
  <TabPanel tab={<><Icon /> Tab name</>}>
    Tab contents
  </TabPanel>
  Content between tabs in tab bar
  <TabPanel tab={<><Icon /> Tab name</>}>
    Tab contents
  </TabPanel>
  Content after tabs in tab bar
</TabPanels>
Content after panel
```

Is there anything that this approach really doesn't support?

owenniblock: I wonder if these discussions should happen after we release this as a draft and before we move to alpha? 🤔 I'm aware that it's taken a long time to get us this far and our priority should be getting the `CommentEditor` experience out and in line with the Rails experience.

If we feel like we need to get the API hammered out before we release this as a PRC component, perhaps an alternative approach could be to move this work into dotcom as an experimental component and then iterate there before moving it back to Primer later?

Happy with either approach, just don't want to spend too much longer this bit of work and lose sight of our goals.

If neither of those approaches is acceptable, I guess I just need a clear idea of what our definition of done is here so I know what I need to finish off to get this merged.

owenniblock: In the meantime @joshblack I've changed this logic to branch on `child.props.role` instead of the type of the child, hopefully this will enable consumers to create wrapper elements for the tabs and tabpanels.

iansan5653: > I guess I just need a clear idea of what our definition of done is here so I know what I need to finish off to get this merged.

Yeah I should clarify that for me personally, while I have strong opinions about React APIs, my opinions are definitely not blocking. Feel free to disregard my comments if you have a need to get this shipped 🙂 



> In the meantime @joshblack I've changed this logic to branch on `child.props.role` instead of the type of the child, hopefully this will enable consumers to create wrapper elements for the tabs and tabpanels.

I don't think this will actually solve the problem unfortunately. As a basic example:

```tsx
const PreviewTab = ({markdown}) => {
  const content = useRenderMarkdown(markdown)

  return <TabPanel>{content}</TabPanel>
}


const CommentEditor = ({value}) => (
  <TabPanels>
    <TabPanels.Tab>Preview</TabPanels.Tab>
    <PreviewTab markdown={value} />
  </TabPanels>
)
```

In this example, when iterating over the children you'll see a `PreviewTab` instance. `child.props` will be an object in the shape `{markdown: string}` - there's no `role` here. 

If we have to do this manipulation, a better approach might be to directly manipulate the DOM instead of working with the `Children` API. For example, you could, in a `useEffect` after rendering, use `querySelectorAll("[role=tab]")` to get all the tab elements, then update them. This allows any components to be used since it only depends on the DOM structure.

However it's worth noting that this is just as much of a hack as using the `Children` API and you have to be very careful to keep reapplying the changes as React updates the DOM. It's also not SSR-compatible; the manipulation has to be done after client-side render. It's very much not ideal but it is a lot more flexible than manipulating `children`.

The 'idiomatic' way to do it would still be the prop slot API I proposed in my previous comments, since that requires no manipulation outside the standard React render cycle.

owenniblock:
[Integration test for dotcom passes](https://github.com/github/github/pull/317019) ✅

owenniblock:
Not sure why design reviewers is on this PR, I've added engineer reviewers 🤷

joshblack (approved):

owenniblock:
CI was green when I reran the failing stuff so going to try one last time to merge...

Issue #4240, opened 2024-02-09
Source links of the components are broken on the primer.style/react


broccolinisoup:
### Description

We need to update our components' source links to reflect the new file paths after the monorepo changes.

As an example https://github.com/primer/react/blob/main/src/Tooltip/Tooltip.tsx should be https://github.com/primer/react/blob/main/packages/react/src/Tooltip/Tooltip.tsx

### Steps to reproduce

1. Go to https://primer.style/react
2. Go to any component page (Directly linking the component page i.e. https://primer.style/react/Tooltip redirects to the new website and view source link is not present on the new docs side)
3. Click the source link to access the source code of the component
4. Observe that the link is broken.
### Version

v36.7.2

joshblack:
Thanks for catching this! Opened a PR over at: https://github.com/primer/react/pull/4243 to address this

PR #4243, opened 2024-02-09
docs: update source urls to point to new location in monorepo

joshblack:
Closes https://github.com/primer/react/issues/4240

Updates the source url for files to the new monorepo structure

#### Changed

- Source url for mdx component files have been changed

### Rollout strategy

This is a docs related change

### Testing & Reviewing

- [ ] Verify I didn't miss any links 😓
- [ ] Verify that the new locations are correct

### DIFF
--- a/docs/content/ActionList.mdx
+++ b/docs/content/ActionList.mdx
@@ -2,7 +2,7 @@
 componentId: action_list
 title: ActionList
 status: Beta
-source: https://github.com/primer/react/tree/main/src/ActionList
+source: https://github.com/primer/react/tree/main/packages/react/src/ActionList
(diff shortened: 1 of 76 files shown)


broccolinisoup (changes requested):
Ah I just realised, there are some links that includes `blob` instead of `tree` in the url and they still point to the old one. 

i.e. [Truncate](https://primer-cb7bf1b4f6-13348165.drafts.github.io/Truncate) 

Sorry this is a bit tedious 😞

joshblack:
Thanks for catching that @broccolinisoup! So helpful 🙌

broccolinisoup (approved):
Looks great! Thanks for updating them 🙌🏻

PR #4245, opened 2024-02-09
SelectPanel2: Automatically adjust height based on contents

siddharthkp:
- Fixes https://github.com/github/collaboration-workflows-flex/issues/449

I was under the false impression that SelectPanel should retain it's height, but we do want it to resize with contents!

Before: 

https://github.com/primer/react/assets/1863771/ac1f08a2-bac9-44c5-8035-e7ea94f54bab

After: 

https://github.com/primer/react/assets/1863771/139e1cf6-cfe5-41a7-a1ad-9cb7281f2508


joshblack (approved):

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -65,7 +65,8 @@ export type SelectPanelProps = {
 
   // TODO: move these to SelectPanel.Overlay or overlayProps
   width?: OverlayProps['width']
-  height?: OverlayProps['height'] | 'fit-content'
+  height?: 'fit-content' // not used, keeping it around temporary for backward compatibility
```
broccolinisoup, line 68: Just out of curiosity, what was the reason of wanting to keep this value? 
Since SelectPanel is an experimental component, I would think we can remove this value to make the code simpler and update the existing use case if there is any.

broccolinisoup (approved):

Issue #4262, opened 2024-02-12
`CheckboxGroup` has `aria-conditional-attr` violation


joshblack:
### Description

The `CheckboxGroup` and `Checkbox` components have an [`aria-conditional-attr` violation](https://dequeuniversity.com/rules/axe/4.8/aria-conditional-attr?application=AxeChrome) from axe due to usage of `aria-checked`

The value of `aria-checked` should mirror the value of the input's `checked` property

### Steps to reproduce

1. Visit the [component storybook](https://primer.style/react/storybook/iframe.html?args=&id=components-checkboxgroup--default&viewMode=story)
2. Run axe
3. View issue titled: "ARIA attributes must be used as specified for the element's role"

### Version

v36.x

PR #4263, opened 2024-02-12
fix(Checkbox): mirror `aria-checked` to checked property of checkbox input

joshblack:
Closes https://github.com/primer/react/issues/4262

#### Changed

- Mirror the `aria-checked` attribute to the `checked` property on the checkbox input

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- [ ] Verify that the axe issue from the original issue is no longer present
- [ ] Do you have any ideas for better handling this change? 👀 Unfortunately, `defaultChecked` was the root of the issue here and I couldn't think of a better way to keep things in sync since this component can be both controlled and uncontrolled


joshblack:
bump @pksjce if you have a sec to review!

broccolinisoup (approved):
>Verify that the axe issue from the original issue is no longer present

Yeah, all gone!
>Do you have any ideas for better handling this change? 👀 Unfortunately, defaultChecked was the root of the issue here and I couldn't think of a better way to keep things in sync since this component can be both controlled and uncontrolled

I think this is great. The way I thought is very similar to yours, I think yours is much cleaner and certainly avoids nested ternaries 😄  

```jsx
checked={indeterminate ? false : defaultChecked || checked}
aria-checked={indeterminate ? 'mixed' : defaultChecked ? defaultChecked : checked ? 'true' : 'false'}
```

Thanks for fixing this 🚀

PR #4264, opened 2024-02-12
feat: remove dedicated type entrypoints

joshblack:
Remembered this when reviewing #4225 that we have dedicated subfolders for TypeScript types to resolve correctly and that the `package.json` files in these folders have a link to an issue.

It appears that this issue is closed and so I wanted to test it out to see if it worked as intended. If so, we can remove these folders in preference of the types in `exports`

#### Changed

- Add `types` to conditional exports in `package.json`

#### Removed

- Remove `drafts`, `experimental`, and `deprecated` stub folders

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

- [ ] Verify that these folders are no longer needed in case I missed anything 😅 
- [ ] Double check that CI passes
- [ ] For testing upstream, I opened up a PR and they seem to be okay and are failing due to unrelated changes

### DIFF
--- a/.changeset/lovely-tools-share.md
+++ b/.changeset/lovely-tools-share.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Replace subfolders used for TypeScript types with "types" field in "exports"
--- a/packages/react/deprecated/package.json
+++ b/packages/react/deprecated/package.json
@@ -1,9 +1,0 @@
-{
-    "_comment1": "this is required only for typescript. once this is fixed https://github.com/microsoft/TypeScript/issues/33079 we can remove this hack",
-    "name": "@primer/react/deprecated",
-    "types": "../lib-esm/deprecated/index.d.ts",
-    "main": "../lib-esm/deprecated/index.js",
-    "type": "module",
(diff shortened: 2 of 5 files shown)


siddharthkp (approved):
Hi!

I don't fully understand the change 😅  but trust that you have tested this with dotcom!

joshblack:
@siddharthkp I think the notable thing is that we had these folders in place for when TypeScript didn't fully support the `"exports"` config in `package.json`. The linked issue in the `package.json` files that we had has since been closed and the capabilities for this are now behind module resolution options like node16/bundler but may not be present if someone is using just `node` as the resolution option.

The big change in this PR is shifting types over to the export conditions using the same syntax as other entrypoints. The drawback of this is that tools like publint will warn that the types linked to in the `"import"` block will still be interpreted as CommonJS and not ESM since the parent `package.json` file has an implicit `"type": "commonjs"` 😬 This seems to be the main downside of dual bundling and would require us having to emit these files as `.d.mts` files.

Anyways, just wanted to add some more context here lol. There's a lot of moving pieces so just wanted to share before merging to get your thoughts.

PR #4266, opened 2024-02-13
ci: update release, release_candidate workflows

joshblack:
Update our `release` and `release_candidate` workflows to work with the new monorepo structure. Unfortunately, this means that we inline the shared workflows in order to unblock the current release. In the future, we can move to a shared workflow that works with multiple packages with the built-in `changesets` functionality

#### Changed

- Update `release.yml` with logic from shared workflow
- Update `release_canary.yml` with logic from shared workflow

### Rollout strategy

### Testing & Reviewing

Sadly I think we can only test this out during the release 😞 In the interim, we should scan through the code in this PR to confirm that:

- [ ] the version for the pre-release will no longer be `null`
- [ ] there are no more reference to workflow inputs
- [ ] all secrets are accounted for and included

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,14 +10,37 @@ jobs:
   release-main:
     if: github.ref_name == 'main'
     name: Main
-    uses: primer/.github/.github/workflows/release_with_app.yml@main
-    secrets:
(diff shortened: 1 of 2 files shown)


broccolinisoup (approved):
Thanks so much! Looks great. I'll test this today.

Issue #4228, opened 2024-02-06
`PointerBox` example is not working in primer.style


joshblack:
The [`PointerBox`](https://primer.style/components/pointer-box/react/alpha) example is currently broken on https://primer.style due to an error in the story itself. When [visiting the story](https://primer.style/react/storybook/?path=/story/components-pointerbox--default&globals=colorScheme:light), we see the following error:

```
TypeError: Cannot read properties of undefined (reading 'every')
    at U (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:454:20021)
    at b (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:454:22257)
    at cr (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:454:27010)
    at Dt (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:454:27848)
    at https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:454:27644
    at Array.map (<anonymous>)
    at Te (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:454:27606)
    at et (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:454:27742)
    at Ht (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:455:9884)
    at Yr (https://primer.style/react/storybook/4516.3cb9eab0.iframe.bundle.js:455:10065)
```

This appears to be a case with internal storybook logic and may be related to the `args` or `argTypes` on the story.

PR #4268, opened 2024-02-14
Fix PointerBox storybook

pksjce:
Closes #https://github.com/primer/react/issues/4228

This PR gets rid of storybook error that prevented this particular story from rendering in both docs and storybook


broccolinisoup (approved):
Oh wow very interesting that removing the component attribute fixes this!

pksjce:
Yes, i think its no longer required in storybooks

PR #4272, opened 2024-02-14
Fixes for `ActionList` semantics

TylerJDev:
Fixes issue in: https://github.com/github/accessibility-audits/issues/2939

Based off of @broccolinisoup's work in https://github.com/primer/react/pull/3971 

Ensures that the default semantics of `ActionList.Item` rely on `<button>` instead of `<li tabindex="0">`. If the `ActionList` itself should act as a list or menu, `<li>` will be used with the appropriate roles.

#### Changed

* Default element type of `ActionList.Item` is now `button`
* Minor type changes (i.e. `HTMLLIElement` to `HTMLButtonElement`)
* Small ARIA addition to `NavList`, `AutocompleteSuggestions`.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Integration test PR: https://github.com/github/github/pull/317139

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/Item.tsx
```diff
@@ -250,8 +256,22 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     const inlineDescriptionId = `${itemId}--inline-description`
     const blockDescriptionId = `${itemId}--block-description`
     const inactiveWarningId = inactive && !showInactiveIndicator ? `${itemId}--warning-message` : undefined
+    const validRole = listRole === 'listbox' || listRole === 'menu' || listRole === 'list' || inactive
```
TylerJDev, line 259: I'm using this conditional as a backwards compatible escape hatch. If an `ActionList` is using one of these roles, treat it as a list, or something that carries similar semantics. This includes `Menu`, where can retain the `<li>` semantics. If developers need to utilize this component as a _list_ they can do so by providing the proper ARIA role.

packages/react/src/ActionList/Item.tsx
```diff
@@ -272,20 +292,31 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       id: itemId,
     }
 
-    const containerProps = _PrivateItemWrapper ? {role: itemRole ? 'none' : undefined} : menuItemProps
+    const containerProps = _PrivateItemWrapper
+      ? {role: itemRole ? 'none' : undefined, ...props}
+      : // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        (validRole && {...menuItemProps, ...props, ref: forwardedRef}) || {}
```
TylerJDev, line 298: Curious if there's a better way to do this. ESLint was throwing an odd error that didn't seem to be valid, but I'm not too sure what an alternative would look like. Essentially, if `_PrivateItemWrapper` is true, default to no role, and pass the props onto the wrapper. Whereas if `validRole` is true, pass both props, and ref to the list item (`<li>`). If `validRole` is false, don't provide anything, as this will be handled in `ButtonItemWrapper`.

packages/react/src/ActionList/Item.tsx
```diff
@@ -272,20 +292,31 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
       <ItemContext.Provider
         value={{variant, disabled, inactive: Boolean(inactiveText), inlineDescriptionId, blockDescriptionId}}
       >
         <LiBox
-          ref={forwardedRef}
-          sx={merge<BetterSystemStyleObject>(styles, sxProp)}
+          sx={merge<BetterSystemStyleObject>(
+            validRole || _PrivateItemWrapper ? styles : listItemStyles,
+            validRole || _PrivateItemWrapper ? sxProp : {},
+          )}
```
TylerJDev, line 317: I was having issues with the styles in other components that use `ActionList` (e.g. `ActionMenu`, `NavList`), which is why we have this conditional. Ensures that styles are properly passed to `<LiBox>` depending on if it's a `validRole`, or `_PrivateItemWrapper`.

TylerJDev:
@broccolinisoup, I would love your review on this PR while it's still in draft state. I started from your PR and tried to shape it in a way that would be (mostly) backwards compatible with existing versions, so I'd love any additional input you might have.

packages/react/src/ActionList/Item.tsx
broccolinisoup: I'd like to understand the reason of having `list` role in this condition 🤔 

When I think about the semantics 

```jsx
<ul role='menu'>
  <li role='menuitem'></li>
  <li role='menuitem'></li>
</ul>
```
or 
```jsx
<ul role='listbox'>
  <li role='option'></li> 
  <li role='option'></li>
</ul>
```

The element that user interacts is the `li` element. So keeping the current semantics and keeping the onSelect item on the li element make sense to me. 

However, when I think about the `list`, I imagine a list of action items and the semantics in my head is looking like this;

```jsx
<ul role='list'>
  <li role='listitem'>
    <button>action 1</button>
  </li>
  <li role='listitem'>
      <button>action 2</button>
  </li>
</ul>
```
In this case, user will interact with the button so we will need to make sure the proper ARIA attributes and onSelect element is triggered on the button element not the list. 

So I'd like to understand why we would like to keep the semantics same for the `list` role as well. Also as far as I know, when there is no role is specified on the `<ul>` element, it renders as `list`. So I am a bit confused and would like to understand why we are treating the `list` role as same as `listbox` and `menu`

Let me know if I misunderstood anything here or anything I can clarify on my questions. Thank you 💖

broccolinisoup (commented):
Thanks so much for this PR @TylerJDev, really appreciate it 🙌🏻 I left some questions to start with. Happy to discuss them here async or in our call tomorrow. Looking forward to hearing your thoughts.

packages/react/src/ActionList/Item.tsx
TylerJDev: Thank you for the write-up! I agree with your comment on `role="list"`. My idea for including `role="list"` here was as an escape hatch. Using `role="list"` means that the markup will purely be a list, e.g. 

```html
<ul role="list"> <!-- ActionList rendered markup -->
  <li>Item 1</li> 
  <li>Item 2</li>
</ul>
```

In addition, I added `role="list"` here to improve our backwards compatibility, as it would allow developers to retain the same semantics as before, _if_ their use case was valid. A good example of this is with `NavList`. In this component we utilize `ActionList.Item` as a button (`as="button"`), and then we add our own custom `<li>`. 

```html
<Box as="li" sx={{ ... }}>
  <ActionList.Item as="button">{ .... }</ActionList.Item>
  <div>{subNav}</div>
</Box>
```

This will render the following markup:

```html
<li> <!-- Box as="li" -->
  <button>Item 1</button> <!-- ActionList.Item -->
  <div> <!--- SubNav --->
    <ul>{...}</ul> <!-- SubNav `ActionList` -->
  </div>
</li>
```

`NavList` uses `<Box as="li">` as a container for both `ActionList.Item` and `SubNav` items. This is so that the `SubNav` is a sibling to the `ActionList.item as="button"`. Without `role="list"` here, it will default to the new semantics, which is okay, but comes with a few issues: 

* Duplicate list items (`<li>`) due to both `<Box as="li">` and `<LiBox>` being used.
* Removing `<Box as="li">` will mean that the `SubNav` content is no longer a sibling to `ActionList.Item` `<button>`.

I wasn't entirely sure how to fix the above without making more changes to the API of either `NavList` or `ActionList.item`, so I chose the path of least resistance 😅. I'm open to any additional feedback, or changes we could make here.

packages/react/src/ActionMenu/ActionMenu.examples.stories.tsx
```diff
@@ -346,7 +346,7 @@ export const MultipleSections = () => {
 export const DelayedMenuClose = () => {
   const [open, setOpen] = React.useState(false)
   const [copyLinkSuccess, setCopyLinkSuccess] = React.useState(false)
-  const onSelect = (event: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>) => {
+  const onSelect = (event: React.MouseEvent<HTMLButtonElement> | React.KeyboardEvent<HTMLButtonElement>) => {
```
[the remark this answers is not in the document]
TylerJDev: Definitely! Great point, I'll change this!

packages/react/src/ActionList/Item.tsx
broccolinisoup: Thanks so much for the great explanation!

It does makes sense to tweak (include `list` into the valid roles) the implementation to make sure `Navlist` works as expected because otherwise we would need some additional changes in the implementation. 

>Using role="list" means that the markup will purely be a list

This is where I am a little unsure how to communicate between an explicit `role="list"` vs implicit `role="list"`. Because when there is no `role` prop is specified on `ActionList`, it also render the list as `list`. Based on this,  we end up having two "list" lists:

1. interactive list items - Explicit (role="list" is passed as a prop)

```jsx
<ul> 
  <li>Item 1</li>
  <li>Item 2</li>
</ul>
```

2. List items that have buttons - Implicit (no role prop is passed)

```jsx
<ul> 
  <li>
    <button>Item 1</button>
  </li>
  <li>
    <button>Item 2</button>
  </li>
</ul>
```

I am not sure why/where we would need the 1. option except supporting existing use cases which is `NavList` and [CustomImagesList](https://github.com/github/github/blob/master/ui/packages/custom-images/routes/CustomImagesList.tsx#L47) plus open source use cases if any.

[CustomImagesList](https://github.com/github/github/blob/master/ui/packages/custom-images/routes/CustomImagesList.tsx#L47) looks like a static list, they might be using `ActionList` just for styling maybe? 🤷🏻‍♀️ 

All considered, I am thinking would it make sense to exclude `NavList` use case so we don't have to distinguish explicit vs implicit role? Exclude I mean, it could be reading the context of ActionList and if it is `NavList` we can keep the semantics same? It is very similar to what your original thinking is, it is just a different condition I am suggesting or thinking loudly to be honest 😄 

I'll also tag @siddharthkp to hear his suggestion since he has great context on ActionList 🙌🏻

TylerJDev: > it could be reading the context of ActionList and if it is NavList we can keep the semantics same

We could definitely try this route! Even if we wanted to have this condition based through a prop, I think it would work the same more or less. I'll dig on this one, but if we can all agree on it then I'm fine with losing `role="list"`.

siddharthkp: > This is where I am a little unsure how to communicate between an explicit role="list" vs implicit role="list".

This is where I got confused as well! 😅 

I need to poke around this a bit more to understand it better. I'll be back 😇

packages/react/src/SegmentedControl/SegmentedControl.tsx
```diff
@@ -129,10 +129,10 @@ const Root: React.FC<React.PropsWithChildren<SegmentedControlProps>> = ({
                 <ActionList.Item
                   key={`segmented-control-action-btn-${index}`}
                   selected={index === selectedIndex}
-                  onSelect={(event: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>) => {
+                  onSelect={(event: React.MouseEvent | React.KeyboardEvent) => {
                     isUncontrolled && setSelectedIndexInternalState(index)
                     onChange && onChange(index)
-                    child.props.onClick && child.props.onClick(event as React.MouseEvent<HTMLLIElement>)
+                    child.props.onClick && child.props.onClick(event as React.MouseEvent<HTMLButtonElement>)
```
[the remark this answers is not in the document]
TylerJDev: Yup, good callout! This should be `li` instead of `button`. Will update!

packages/react/src/NavList/NavList.tsx
```diff
@@ -33,7 +33,7 @@ const NavBox = styled.nav<SxProp>(sx)
 const Root = React.forwardRef<HTMLElement, NavListProps>(({children, ...props}, ref) => {
   return (
     <NavBox {...props} ref={ref}>
-      <ActionList>{children}</ActionList>
+      <ActionList role="list">{children}</ActionList>
```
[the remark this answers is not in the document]
TylerJDev: Yup, exactly! I used it here as an "escape-hatch" so that `NavList`, (and potentially others) could keep the default `ActionList` semantics.

packages/react/src/ActionList/shared.ts
```diff
@@ -11,7 +11,7 @@ export type ActionListItemProps = {
    * Callback that will trigger both on click selection and keyboard selection.
    * This is not called for disabled or inactive items.
    */
-  onSelect?: (event: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>) => void
+  onSelect?: (event: React.MouseEvent | React.KeyboardEvent) => void
```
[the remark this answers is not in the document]
TylerJDev: That is very true 😭, It's been something I've been dealing with in the [integration test PR](https://github.com/github/github/pull/315374). The main issue I had was, I couldn't find a decent way to express what element type `ActionList.Item` was based on the different conditions we have (i.e. passing a specific role to `ActionList` may cause `ActionList.Item` to be `HTMLLiElement` instead of `HTMLButtonElement`).

packages/react/src/ActionList/Item.tsx
```diff
@@ -137,6 +137,12 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       },
     }
 
+    const listItemStyles = {
+      display: 'flex',
+      // show between 2 items
+      ':not(:first-of-type)': {'--divider-color': theme?.colors.actionListItem.inlineDivider},
```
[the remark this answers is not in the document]
TylerJDev: I believe this was added to fix some styles when `<button>` was added. @broccolinisoup might have more context here!

packages/react/src/NavList/NavList.tsx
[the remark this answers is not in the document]
siddharthkp: Can you say more, what are the default ActionList semantics?

packages/react/src/ActionList/ActionList.examples.stories.tsx
```diff
@@ -327,7 +327,7 @@ const SortableItem: React.FC<React.PropsWithChildren<SortableItemProps>> = ({opt
   return (
     <ActionList.Item
       role={role}
-      ref={element => dragRef(element) && dropRef(element)} // merge refs
+      ref={(element: HTMLLIElement) => dragRef(element) && dropRef(element)} // merge refs
```
siddharthkp, line 330: Is this required? Does it throw a TS error without this? (asking to check if we would need to do this in dotcom as well?)

packages/react/src/ActionList/shared.ts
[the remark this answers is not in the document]
siddharthkp: I'm guessing there are TS errors with something like as well?

```
onSelect?: (event: React.MouseEvent<HTMLLIElement | HTMLButtonElement> | React.KeyboardEvent<HTMLLIElement | HTMLButtonElement>) => void
```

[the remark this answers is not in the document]
TylerJDev: Nope, I don't get any TS errors in PRC! This could be worth a try in the integration PR :eyes:. Would prefer this, especially if it doesn't require any additions/removals to existing types in dotcom.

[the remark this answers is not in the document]
broccolinisoup: @siddharthkp sorry for ghosting you on the ts question!

>@broccolinisoup is there typescript magic that we can do to infer between button and list so that we can make the types: React.MouseEvent<HTMLLIElement | HTMLButtonElement> instead?
I actually remember we talked about maybe updating dotcom to use `ActionListItemProps['onSelect']`and then it infer the types. @TylerJDev should we give this a go?

Also could you share the integration PR with us and I can look at the logs and run it on the code space as well to have some more thoughts?

packages/react/src/ActionList/Item.tsx
broccolinisoup: Do we have any updates on this escape hatch solution? 👀

packages/react/src/ActionList/ActionList.examples.stories.tsx
TylerJDev: Just removed it, as it wasn't required for TS!

packages/react/src/ActionList/shared.ts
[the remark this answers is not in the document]
TylerJDev: Definitely, here's the integration PR: https://github.com/github/github/pull/317139. I'll add this in the description too.

packages/react/src/NavList/NavList.tsx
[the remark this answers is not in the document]
TylerJDev: The default ActionList semantics would be `<li><button>Item</button></li>` with any props being placed on the button rather than the list item. The criteria to maintain the same semantics (i.e. only `<li>`, no `<button>`) are:

* If either ARIA role is provided (`role="list"`, `role="listbox"`, `role="menu"`)
* If `inactive` is passed to `ActionList.Item`
* If a valid value of `_PrivateItemWrapper` is passed
* If `ActionList` is used within an `ActionMenu`

So we'd have this (default in this PR):

```
<ul>
  <li>
    <button {...props}>Item 1</button>
  </li>
</ul>
```

and the semantics if any of the criteria is met (default currently outside of this PR):
```
<ul>
  <li tabindex="0" {...props}>
    item 1
  </li>
</ul>
```

TylerJDev:
Wanted to loop back here to plan a potential rollout. Since this PR consists of a change that would revamp the semantics of `ActionList` by default, it could be potentially breaking. Here are some ideas I had:

### Make new semantics opt-in via a prop

Essentially, we could remove the logic that checks to see if the `ActionList` is a certain ARIA role (e.g. `role="list"`, `role="menu"`), and instead give access to the new semantics via a prop that would then render the `ActionList` with the buttons rather than just list items (e.g. `renderButtonItems`). This allows existing implementations that are inaccessible due to the current semantics to change to the more accessible markup via the prop. 

We could then aim to make the semantics available by default without the prop, potentially in our next major release.

### Add prop to all existing `ActionList` components that disables the new semantics

This would be similar to the rollout of Tooltip v2. The aim would be to slowly reduce the amount of usage of the old semantics, while any new implementations of `ActionList` would use the new semantics by default. We could slowly enforce usage of the new semantics overtime, by encouraging consumers of `ActionList` to remove the prop and rely on the new semantics.

### Release new semantics as-is

The goal for this PR is to be backwards compatible with existing implementations, meaning that any current `ActionList` now, should perform the same with the new semantics. It further reduces the amount of risk by limiting what _type_ of `ActionList` receives the new semantics (i.e. some ARIA roles won't receive the new semantics). We could further gain confidence by testing existing `ActionList` implementations across dotcom, (there are numerous occurrences, so we wouldn't' be able to test all of them).

I'm open to any additional ideas, or thoughts :grin:

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -382,7 +382,8 @@ describe('ActionList', () => {
 
   it('should render ActionList.Item as button when feature flag is enabled', async () => {
     const {container} = HTMLRender(
-      <FeatureFlags flags={{'action-list-item-as-button': true}}>
+      // eslint-disable-next-line camelcase
+      <FeatureFlags flags={{action_list_item_as_button: true}}>
```
siddharthkp, line 386: Heads up! Feature flag names should start with `primer_react_`

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -378,4 +379,47 @@ describe('ActionList', () => {
     const heading = getByText('Group Heading')
     expect(list).toHaveAttribute('aria-label', heading.textContent)
   })
+
+  it('should render ActionList.Item as button when feature flag is enabled', async () => {
+    const {container} = HTMLRender(
+      // eslint-disable-next-line camelcase
+      <FeatureFlags flags={{action_list_item_as_button: true}}>
```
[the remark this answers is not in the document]
siddharthkp: cc @joshblack Interesting thing to note here, we would need to eslint disable camelcase for our feature flags 😅 

Not a big deal, tiny papercut though

TylerJDev:
Switching this PR to "ready for review". Now that we've added feature flags, I believe we can smoothly release this change and monitor it 🥳 

I've tested briefly in Dotcom and it appears that some of the implementations that this PR would touch work with no issues!

.changeset/hot-mayflies-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Utilizes `<button>` inside of `ActionList.Item` instead of relying on `<li>` as an interactive item.
```
TylerJDev, line 5: What would be the best way to convey that this feature is feature flagged and not available outside of the flag? 🤔

packages/react/src/ActionList/Item.tsx
```diff
@@ -259,8 +268,24 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     const inlineDescriptionId = `${itemId}--inline-description`
     const blockDescriptionId = `${itemId}--block-description`
     const inactiveWarningId = inactive && !showInactiveIndicator ? `${itemId}--warning-message` : undefined
+    const validRole = listRole === 'listbox' || listRole === 'menu' || listRole === 'list' || inactive
```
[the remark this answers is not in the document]
broccolinisoup: It is interesting that I can see you replaced in the [commit](https://github.com/primer/react/pull/4272/commits/6774f937c5b531f75eb35f755aafee8c2f361fe0) but it is not reflected here. I am confused 😅

.changeset/hot-mayflies-pump.md
broccolinisoup: Curious to learn that as well!

broccolinisoup (commented):
I haven't been able to fully reviewed it but will follow up with the rest tomorrow 🙌🏻

packages/react/src/ActionList/Item.tsx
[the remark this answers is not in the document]
TylerJDev: That's a good catch. I must've accidentally added it back somehow? Should be removed for good now!

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -378,4 +379,45 @@ describe('ActionList', () => {
     const heading = getByText('Group Heading')
     expect(list).toHaveAttribute('aria-label', heading.textContent)
   })
+
+  it('should render ActionList.Item as button when feature flag is enabled', async () => {
+    const {container} = HTMLRender(
+      <FeatureFlags flags={{primer_action_list_item_as_button: true}}>
```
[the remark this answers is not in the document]
TylerJDev: There was [a recent PR](https://github.com/primer/react/pull/4615) by @joshblack to address this! I just needed to add `primer_react` instead of just `primer`. Should be good now!

packages/react/src/ActionList/Item.tsx
```diff
@@ -259,8 +268,24 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
+              {...props}
+            >
+              {children}
+            </Box>
+          )
+        }) as PolymorphicForwardRefComponent<React.ElementType, ActionListItemProps>)
+      : React.Fragment
 
-    const ItemWrapper = _PrivateItemWrapper || React.Fragment
+    const ItemWrapper = _PrivateItemWrapper || (validRole || !buttonSemantics ? React.Fragment : ButtonItemWrapper)
```
broccolinisoup, line 288: I think we also need to add NavList context check into this condition, don't we?


```jsx
let DefaultItemWrapper = React.Fragment 
if(buttonSemantics) {
  DefaultItemWrapper = validRole || navListContext ? React.Fragment : ButtonItemWrapper
} 

const ItemWrapper = _PrivateItemWrapper || DefautlItemWrapper
```

I don't mean to re-write the condition, it wasn't straightforward to read when I added the NavList (at least for me 😅) Do it however you feel comfortable 💖

packages/react/src/ActionList/Item.tsx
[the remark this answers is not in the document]
broccolinisoup: I don't remember sorry 😢 It has been a while since I pushed that draft PR. We can comment it out and see how the stlyes are different to figure out if it still works fine in this PR.

broccolinisoup (commented):
I left some comments and questions! Looking forward to hearing what you think 🙌🏻 

Also, should we enable the rule in stories and see how the changes are working?

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -378,4 +379,45 @@ describe('ActionList', () => {
     const heading = getByText('Group Heading')
     expect(list).toHaveAttribute('aria-label', heading.textContent)
   })
+
+  it('should render ActionList.Item as button when feature flag is enabled', async () => {
+    const {container} = HTMLRender(
+      <FeatureFlags flags={{primer_react_action_list_item_as_button: true}}>
```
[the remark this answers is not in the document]
TylerJDev: Good idea! I've added your suggestion!

packages/react/src/ActionList/Item.tsx
```diff
@@ -259,8 +268,24 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     const inlineDescriptionId = `${itemId}--inline-description`
     const blockDescriptionId = `${itemId}--block-description`
     const inactiveWarningId = inactive && !showInactiveIndicator ? `${itemId}--warning-message` : undefined
+    const validRole = listRole === 'listbox' || listRole === 'menu' || inactive
```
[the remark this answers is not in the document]
TylerJDev: That is fair, I can add both, but let me know if we'd rather go with one or the other!

packages/react/src/ActionList/Item.tsx
```diff
@@ -281,20 +306,44 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
+    }
 
     return (
       <ItemContext.Provider
         value={{variant, disabled, inactive: Boolean(inactiveText), inlineDescriptionId, blockDescriptionId}}
       >
         <LiBox
-          ref={forwardedRef}
-          sx={merge<BetterSystemStyleObject>(styles, sxProp)}
+          ref={buttonSemantics || validRole ? forwardedRef : null}
```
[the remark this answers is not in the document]
TylerJDev: It switches between the inner `<button>` and the list item depending on the context. When the `<button>` is applied, we pass the props and `forwardedRef` to it, whereas if it's a list item only (i.e. `validRole` is true), then we provide the `forwardedRef` to that list item. I did it this way so that the `forwardedRef` will always be on the thing that you can pass props to.

packages/react/src/ActionList/Item.tsx
```diff
@@ -259,8 +268,24 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     const inlineDescriptionId = `${itemId}--inline-description`
     const blockDescriptionId = `${itemId}--block-description`
     const inactiveWarningId = inactive && !showInactiveIndicator ? `${itemId}--warning-message` : undefined
+    const validRole = listRole === 'listbox' || listRole === 'menu' || inactive
+
+    const ButtonItemWrapper = buttonSemantics
+      ? (React.forwardRef(({as: Component = 'button', children, ...props}, forwardedRef) => {
+          return (
+            <Box
```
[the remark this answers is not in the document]
TylerJDev: `<Box>` is used here to provide a non-styled `<button>`, this also goes for anything that's passed to it via `as`. We could use a `<Button>` here, but I'm not sure if it'll provide more than the current implementation, but I may be wrong here!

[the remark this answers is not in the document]
broccolinisoup: I think you are right, because both render as buttons semantically at the end of the day. Unless we need to utilize Button props, I think we are good with the Box!

packages/react/src/ActionList/Item.tsx
```diff
@@ -259,8 +268,24 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     const inlineDescriptionId = `${itemId}--inline-description`
     const blockDescriptionId = `${itemId}--block-description`
     const inactiveWarningId = inactive && !showInactiveIndicator ? `${itemId}--warning-message` : undefined
+    const validRole = listRole === 'listbox' || listRole === 'menu' || inactive
+
+    const ButtonItemWrapper = buttonSemantics
```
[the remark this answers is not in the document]
TylerJDev: Yup, that is fair! Thank you for the suggestion! 😁

.changeset/hot-mayflies-pump.md
siddharthkp, line 5: How about:

```suggestion
(Behind feature flag) ActionList: Utilizes `<button>` inside of `<li>` for interactive items.
```

packages/react/src/ActionList/Item.tsx
TylerJDev: You're right! This is easier to read too. Thank you! I adjusted `validRole` to handle what `navListContext` would be here.

.changeset/hot-mayflies-pump.md
TylerJDev: This is great, thank you! :grin:

packages/react/src/ActionList/Item.tsx
```diff
@@ -268,24 +268,28 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     const inlineDescriptionId = `${itemId}--inline-description`
     const blockDescriptionId = `${itemId}--block-description`
     const inactiveWarningId = inactive && !showInactiveIndicator ? `${itemId}--warning-message` : undefined
-    const validRole = listRole === 'listbox' || listRole === 'menu' || inactive
+    // Ensures ActionList.Item retains list item semantics if a valid ARIA role is applied, or if item is inactive
```
broccolinisoup, line 271: ❤️

packages/react/src/ActionList/ActionList.features.stories.tsx
```diff
@@ -710,3 +712,25 @@ export const GroupWithFilledTitle = () => {
     </ActionList>
   )
 }
+
+export const ActionListWithButtonSemantics = () => {
```
broccolinisoup, line 716: I wonder if we can enable the flag in all our stories and see how the changes are working?

I haven't tried it but maybe wrapping the story [here](https://github.com/primer/react/blob/main/packages/react/.storybook/preview.js#L215) with the feature flag component? 🤔 

```js
<BaseStyles>
  <FeatureFlags flags={{primer_react_action_list_item_as_button: true}}>
    <Story {...context} />
    {context.globals.colorScheme === 'all' && <p className="theme-name">{theme}</p>}
  </FeatureFlags>
</BaseStyles>
```

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -378,4 +379,49 @@ describe('ActionList', () => {
...
+    expect(button).toHaveTextContent('Item 1')
+
+    // Ensure passed prop "disabled" is applied to the button
+    expect(button).toHaveAttribute('aria-disabled', 'true')
+
+    const listItems = container.querySelectorAll('li')
+    expect(listItems.length).toBe(2)
+  })
+
+  it('should render ActionList.Item as li when feature flag is disabled', async () => {
```
broccolinisoup, line 407: Would we want to test some of the listSemantics conditions as well or is it not necessary? For example;

- Flag is enabled and the action list is a SelectBox -> expected to see the list semantics

Just a suggestion!

packages/react/src/ActionList/Item.tsx
```diff
@@ -95,7 +98,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
 
     const onSelect = React.useCallback(
       (
-        event: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>,
+        event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>,
```
broccolinisoup, line 101: I wonder if loosening the type here (which I think is a good way to eliminate the integration issue with dotcom) might have any "tangible" disadvantages? 🤔

packages/react/src/ActionList/Item.tsx
```diff
@@ -281,20 +310,44 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
       >
         <LiBox
-          ref={forwardedRef}
-          sx={merge<BetterSystemStyleObject>(styles, sxProp)}
+          ref={buttonSemantics || listSemantics ? forwardedRef : null}
+          sx={
+            buttonSemantics
+              ? merge<BetterSystemStyleObject>(
+                  listSemantics || _PrivateItemWrapper ? styles : listItemStyles,
+                  listSemantics || _PrivateItemWrapper ? sxProp : {},
```
broccolinisoup, line 345: >`listSemantics || _PrivateItemWrapper ? sxProp : {},`

I am not sure if I understand this line, sorry. Does that mean we don't allow custom overrides for `listItemStyles` ? In other words, we don't want to merge `listItemStyles` with `sxProp` when `listSemantics || _PrivateItemWrapper` is false? I hope I am making sense 😅

packages/react/src/ActionList/ActionList.features.stories.tsx
TylerJDev: We definitely could! But would we want to keep this on even after we ship this PR, or would it just during this PR?

packages/react/src/ActionList/Item.tsx
TylerJDev: It's definitely not as strict, but it does ease the integration in dotcom. I think with the feature flag it makes it a bit trickier as the type is dependent on the FF being enabled. The issue I had when I was dealing with TS early on was that there are two potential things it could be; a list item or a button. In addition to that, the `<li>` will always be the parent to anything else within the `item`, so technically it is an `HTMLLIElement`, but also a `HTMLButtonElement` since everything is being applied to the button, not the `<li>`.

I did try adding both types `HTMLLIElement | HTMLButtonElement` [and I got odd issues in dotcom](https://github.com/primer/react/pull/4272#discussion_r1529075714) 🤔

packages/react/src/ActionList/Item.tsx
TylerJDev: No worries! This is basically saying that, if the `ActionList.Item` is going to be rendered with list semantics or if it's a private item wrapper, apply the `styles` and `sxProp`, as these are the styles that you'd expect when the items are being rendered as list items only. This retains what we have right now in [`ActionList/Item.tsx`](https://github.com/primer/react/blob/main/packages/react/src/ActionList/Item.tsx#L294). 

When it's not `listSemantics` or `_PrivateItemWrapper` we need to adjust the styles slightly, [as the styles are placed on the `<button>` instead](https://github.com/primer/react/pull/4272/files#diff-21bd620cf266fd1b3eb8d73399ee032c406b9bf49eac6f598ad65c9389ddb515R278), meaning that we don't want to have duplicate styles on the `<li>` and `<button>`.

Let me know if this still makes sense, or if you have suggestions on how we can make this more readable 🤔

packages/react/src/ActionList/ActionList.test.tsx
TylerJDev: That is a great suggestion! Just added a new test!

packages/react/src/ActionList/Item.tsx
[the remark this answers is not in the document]
TylerJDev: I went ahead and removed it, and it seems like it was needed for cases where [a divider is used](https://primer-657d0c9a88-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--item-dividers) and the "button" semantics are applied. I think it's worth keeping for this 🤔

TylerJDev:
@siddharthkp, @broccolinisoup - Hey y'all! I think this is in a good state for a review! I addressed all of the feedback and we should only need [the changes listed in this comment](https://github.com/github/github/pull/317139#issuecomment-2004140416) to roll this out in Dotcom (mainly typescript adjustments). Would love one last review from y'all ❤️

packages/react/src/ActionList/ActionList.features.stories.tsx
broccolinisoup: Sorry for the late response! I was thinking enabling the ff in this PR to make sure the changes are working as expected before releasing it to dotcom. It is like testing it in the small surface area (primer/react) and once we confirm that the changes are good, we can release. 

This is only a suggestion though. If you think we can directly release it that is okay too!

packages/react/src/ActionList/Item.tsx
broccolinisoup: Yeah I remember that! Let's see how this type works then and since it is under ff, we can always work on it!

broccolinisoup (approved):
Sorry for stalling this! I think this PR overall is in a good shape and because it is under ff and only semantic changes, I think we can start testing with confidence. 🙌🏻 

This was a huge work, thanks so much for working on this and your patience with my  nit-pick comments and delayed responses 💖

packages/react/src/ActionList/ActionList.features.stories.tsx
TylerJDev: Added! Should now be enabled for all stories!

packages/react/src/ActionList/Item.tsx
joshblack: Hello from the future! This came up during release conductor stuff this week in dotcom, it seems like this change will cause typed handlers to fail since they are using the more specific type.

Since we've switched to a more generic type, these more specific ones will fail 😭

TylerJDev: Hey Josh! This was a tricky situation since the current type `HTMLLIElement` is technically correct, but also isn't depending on the context. Adding both the LI type and Button type caused a different issue ☹️. 

I made the types more generic in dotcom in my integration test PR that should clean up the failures you're seeing. It does involve making the type a bit more generic, but we don't seem to lose out on too much. Let me know what you think though! 

* [Original integration test PR with more context](https://github.com/github/github/pull/317139#issuecomment-2004140416)
* [Most recent integration test PR with changes](https://github.com/github/github/pull/329335#issuecomment-2181252241)

PR #4276, opened 2024-02-15
feat: add support for feature flags

joshblack:
Add an exploration for introducing feature flags to `@primer/react`. This PR introduces a new component, `FeatureFlags`, that is responsible for enabling or disabling flags in a specific tree of components. It also introduces hooks, `useFeatureFlag` and `useFeatureFlags`, as ways for maintainers to determine if a feature is flagged in or not.

```tsx
<FeatureFlags flags={flags}>
  <ExampleReactApp />
</FeatureFlags>
```

```tsx
function ExampleComponent(props) {
  const enabled = useFeatureFlag('enable_new_feature');

  return <button onClick={() => { enabled ? newBehavior() : oldBehavior() }}>...</button>;
}
```

This PR also includes some docs for how to approach this for things like TypeScript, testing, and (in the future) hopefully I way to interact with flags in Storybook.

#### New

- Add `FeatureFlags` component
- Add `useFeatureFlag` hook
- Add `useFeatureFlags` hook

### Rollout strategy

- [x] Minor release

Introduce in `experimental` entrypoint for testing


joshblack:
Note from UI mob pairing: add support for providing a (sync) function to check if something is enabled

siddharthkp:
Great idea! Love to see it.

Short feedback: I think we make the first iteration a little simpler for ourselves. I'm a little hesitant to support nested scopes right off the bat because usage in gh/gh can quickly get out of our hands. (I'm thinking inconsistencies between pages, untested behavior with combination of feature flags, single instances of a component usage with feature flag)

Ideally, I'd like to start with a broad boundary (adjacent to ThemeProvider) and once we have some experience with it (and related tooling), make it more advanced.

Note for future: It would be nice if we had a common ThemeProvider across gh/gh and it was called something like PrimerProvider / PrimerContainer / PrimerBoundary (with exceptions for charting etc.)

joshblack:
@siddharthkp would love to do something like `PrimerApp` or some kind of "App-level" component like what you described 👍 it would make it simple to add in things as "defaults" to projects that use Primer as a result.

And happy to defer on nesting if that would be helpful 👍 I think the only thing that felt critical is that we want a set of "default" feature flags (in this case the global ones) and that this should be merged in with what is given at the `FeatureFlags` level. This seemed important because otherwise call sites to `FeatureFlags` need to manage all flags and their default value, even if it's unrelated to what is being authored.

siddharthkp (approved):
Looking good!

I know you're going to write tests, approving in advance

PR #4279, opened 2024-02-16
test(e2e): add e2e test for SelectPanel2 default story

joshblack:
Context https://github.com/github/primer/issues/3025

This PR adds the first e2e test for SelectPanel2 for its default story. It also addresses the axe violation that shows up for the `role="listbox"` by labeling it by the title id for the dialog

#### New

- Add e2e test for SelectPanel2

#### Changed

- Add `listLabelledBy` as context value for `ActionList`

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- [ ] Does the change for the `aria-labelledby` make sense? will this always be available or is this something teams should provide?


packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -299,6 +299,7 @@ const Panel: React.FC<SelectPanelProps> = ({
                   selectionAttribute: 'aria-selected',
                   selectionVariant: selectionVariant === 'instant' ? 'single' : selectionVariant,
                   afterSelect: internalAfterSelect,
+                  listLabelledBy: `${panelId}--title`,
```
siddharthkp, line 302: very nice!

siddharthkp (approved):
These look great!

PR #4287, opened 2024-02-19
Feature/3606 enhance avatar image src

### LINKED ISSUE #3606: Add support for StaticImageData with Avatar Components (NextJS)
In NextJS, the most common way to import and use an image, at least for local images, is to do as the documentation suggests ([the documentation](https://nextjs.org/docs/pages/building-your-application/optimizing/images#local-images)):

```ts
import MyImage from "../public/me.png"
```

When we import a local image this way, the `MyImage` variable is an object from the [`StaticImageData`](https://github.com/vercel/next.js/blob/195d1f1a2b6dc794ca9b84be318c90b47f44a85a/packages/next/src/shared/lib/get-img-props.ts#L10-L17) interface.  

For now, a way I found to use the Avatar component in NextJs with the approach I cited above is to pass the `src` attribute of StaticImageData instead of the whole object. Would it be possible to add support for it?

timges:
Closes #3606 

#### Changed

 - ***Avatar***: Expanded the `src` prop to allow not only strings but also image data objects as a source for images. These objects must contain a src key.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

<strong>QUESTION</strong>: The created `ImageData` type is basicaly just a proxy for the StaticImageData type which next.js provides. I did this because I was not sure if installing the actual types of next js is a valid option, or not. What do you think?

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


broccolinisoup (commented):
Hello @timges 👋🏻 Thank you so much for raising this PR! We really appreciate it. 

When I look at the API in the example that is shared in the reported issue, I can see that the `src` prop accepts StaticImageData type and [automatically determine the width and height of the image based on the imported file.](https://nextjs.org/docs/pages/building-your-application/optimizing/images#local-images:~:text=Next.js%20will,the%20imported%20file. )

```
import Image from 'next/image'
import profilePic from '../public/me.png'
 
export default function Page() {
  return (
    <Image
      src={profilePic}
      alt="Picture of the author"
      // width={500} automatically provided
      // height={500} automatically provided
      // blurDataURL="data:..." automatically provided
      // placeholder="blur" // Optional blur-up while loading
    />
  )
}
```

As far as I understand, in this PR, the `src` attribute is extracted from the given obj and the rest is ignored. When nextjs extracts the data from the given image, and the data is provided to the Avatar component, consumers would still need to specify the `size` prop explicitly in the Avatar component

```
import Avatar from '@primer/react'
import profilePic from '../public/me.png'
 
export default function Page() {
  return (
    <Avatar
      src={profilePic}
      alt="Picture of the author"
      size={50} // still need to be explicitly applied
    />
  )
}
```

I am thinking maybe we can use `width` and `height` attribute in the obj to form the `size` prop and automatically apply by default. If it is not the desired size, they can explicitly pass another value. At least this is what I understand from the NextJS API and it would be nice to have that addition for more intuitiveness. 

I'll tag @mperrotti to get his opinion on forming a default size from the extracted data before we go any further 🙌🏻 

Thanks so much again for your time 🙏🏻  Let me know if you have any concern/questions or anything I can clarify further.

timges:
Hey @broccolinisoup ✌🏼. Thank you so much for the review! 😇

Your thoughts on integrating the image data's `width` and `height` make a lot of sense and feel quite natural. If this would be **plain Image component** i would 100% agree with you!

However, as I thought about it some more, I figured that it might mess with the **Avatar's** consistency. 

Right now, the Avatar's default `size` is set at 20x20px, probably chosen to suit most common uses (just guessing here). It also ensures that the avatar image has equal `width` and `height` values, so that the user will always see a nice circular or squared shape rather than an elliptical or rectangular one, when the values differ. 
By exclusively applying the default size when the `src` is a string, users might see different results depending on the type of `src` they choose. It's not a huge deal since users can define the `size` anyway, but it just doesn't sit quite right with me. I would love to get your opinion on this! 🙂

I see two options, which make sense here:
1. Remove the `DEFAULT_AVATAR_SIZE` completely and **calculate** a `size` value based on the `width` and `height` to maintain the squared/circular shape
2. Keep it as is and apply the `DEFAULT_AVATAR_SIZE`, not matter what the type of the `src` is.

Currently i feel like option 2 would be the best fit, as it is non breaking an keeps the component consistent in its default configuration. I might of course be missing better solutions!

Thanks you for being so invested. Excited about your thoughts. 😊

broccolinisoup:
Hi @timges 👋🏻 Sorry for the late response. 

I see your point about not automatically passing the `width` and `height` to the avatar component and it makes a lot of sense! I now have concerns about the need for passing an object though if we are not leveraging passing the width and height and all these additional props 🤔 

I'd like to take this back to chat about the API with my team and get back to you with a little clearer feedback. Thanks for your understanding 🙏🏻

timges:
Hi @broccolinisoup ✌🏼,

Don't worry; the PR isn't going to escape us 😬

I see where you're coming from and agree that more discussion is necessary. Since this feature is mainly focused on improving convenience for a specific set of users, it doesn't seem pressing anyways. Should you require more input, perspectives, or assistance, feel free to reach out, and I'll be happy to chip in 😌👍

lukasoppermann (approved):
LGTM

PR #4290, opened 2024-02-20
Bug/auto complete change event not emitted

### LINKED ISSUE #4289: (Autocomplete) - Last typed character is lost on blur, when the value matches a suggestion
### Description

https://github.com/primer/react/assets/13063135/a908fe24-f3e3-4729-ad1b-93932a03c052

When the string you type into the autocomplete input field exactly matches a suggestion, the last character is lost on blur

### Steps to reproduce

1. Go here https://primer.style/components/autocomplete/react/alpha#examples
2. focus the autocomplete & clear the input
3. type 'css'
4. hit tab

-> last character is lost

### Version

v36.9.0

### Browser

Chrome, Safari, Firefox, Edge

timges:
Closes #4289

#### Changed

Updated the autocomplete-input component to now emit a change event, once a user input makes the input value matches an autocomplete suggestion. This was not the case before, because of the way the autocomplete functionality is implemented: 

-> To render the suggestion into the input field, the value of the underlying input-ref is set to match the next suggestion (if available) after every user input. Thus when the actual input of the user matches the value which is already stored in the input-ref, no change is emitted.

I fixed that by intercepting the exact keystroke where this issue would occur and implemented some custom logic instead of the default behaviour. React brings a private property `_valueTracker` which is used to track changes and act as a trigger for reacts `onChange`. I utilized this to manually trigger a change event.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

The implementation utilizes a private property which is supposd to be some react internal object. Using it might not be the most elegant solution.
Another approcah i thought of was to just intercept the keystroke as is but call the `setInputValue` directly, instead of using the `_valueTracker`

```typescript
[...]
if(inputValue + event.key === autocompleteSuggestion (...)) {
  event.prevenDefault()
  setIputValue(inputRef.current.value)
}
[...]
```

User outcome would basically be the same, but there would be no `changeEvent` being emitted. This lead to my decision of going for the `_valueTracker` approach. Would love to hear youre thoughts on this 😊

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


joshblack (commented):
Hi there! 👋 I'm so sorry about the delay on my end for this, thank you for being so patient 🙏 

I think my only question was around the usage of the private property and if there was any way to get around that 😅 I think that's the only thing that stood out when taking a look.

I'll also include someone else to help me review the PR to help out. Just wanted to say thanks again for opening this PR!

joshblack:
cc @siddharthkp just wanted to include you for another set of 👀

timges:
Hi @joshblack,

No worries—the issue won’t run away 😄 And if it does, maybe that’s for the best.

It's been a few weeks, so I would need to refresh my memory on the context. From what I recall, there's no way to resolve the issue without modifying the current autocompletion logic.

This stems from the input field's value being set to the autocomplete suggestion after each keystroke. Consequently, when the final character is typed, no change is detected because the value has already been updated. Using a private property is a straightforward method to force change detection, although it's clearly not the most elegant solution.

Calling setInputValue(**) might fix the bug visually, but it would still prevent the dispatch of a change event to the user, which isn't ideal either.

I’m not sure I’ll have the time to delve deeper into this at the moment, so for now, we might have to either stick with the provided solution or you could investigate further on your own. Sorry about that, but I’m currently quite busy with my main job 😅

Anyways, thank you so much for taking the time to review this PR! 🫶🏼 I’m excited to see this product evolve—even though I’m not even using it, lol. As someone working in the development of a corporate design system myself, it's inspiring to see what you created. Both design- and development wise. Kudos to you and your team!

PR #4293, opened 2024-02-21
Add 36.9.0 to changelog.md

siddharthkp:
Sync https://github.com/primer/react/releases/tag/%40primer%2Freact%4036.9.0 with changelog.md

We don't usually do this manually, only this one time when the tooling got confused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,36 @@
 # @primer/components
 
+## 36.9.0
+
+### Minor Changes
+
+-   [#4250](https://github.com/primer/react/pull/4250) [`f064534bf6f15261a3b381ce1f7b3828daf98c84`](https://github.com/primer/react/commit/f064534bf6f15261a3b381ce1f7b3828daf98c84) Thanks [@broccolinisoup](https://github.com/broccolinisoup)! - Add a new entry point 'next' and export the Tooltip v2 from it with an updated documentation
+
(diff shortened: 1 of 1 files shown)


joshblack (approved):

joshblack:
Oh wait, is this file here now: https://github.com/primer/react/blob/main/packages/react/CHANGELOG.md ? Next to the package

broccolinisoup (commented):
We have two Changelog files now one is under the root, the one it is edited in this PR and also under `packages/react` which is sync with the v36.9.0. Which one does the changeset refer to? 🤔

siddharthkp:
Updated! 

Added a symlink, hope this is the way to go: https://github.com/primer/react/tree/update-changelog-md-manually

broccolinisoup:
> Updated!
> 
> Added a symlink, hope this is the way to go: https://github.com/primer/react/tree/update-changelog-md-manually

Hmm, not sure if I am missing anything but I see the `packages/react/CHANGELOG.md` as the content of the root changelog instead of the actual changelog 🤔 (Below recording describes this behaviour)


https://github.com/primer/react/assets/1446503/613e0f6b-8469-4f55-8ea0-9beb1f282244

siddharthkp:
Yeah, that's just how GitHub shows symlinks, not sure if that defeats the purpose 🤔 

File that is a symbolic link

joshblack (approved):

broccolinisoup:
> Yeah, that's just how GitHub shows symlinks, not sure if that defeats the purpose and we should just put an anchor tag there 🤔

Right, okay. I didn't know that GH shows the symlinks like that. If it is a generally known thing we can leave it like this I guess, but I think anchor tag is a safer choice. Up to you!

broccolinisoup (approved):

PR #4307, opened 2024-02-26
docs(examples): update App Router example to avoid hydration mismatch

joshblack:
Update the App Router example to avoid the initial hydration mismatch error.

#### Changed

- Add `build` and `start` scripts to `app-router` example
- Add `suppressHydrationWarning` to `html` due to focus-visible changes

### Rollout strategy

### Testing & Reviewing

- Run `npm install`
- Run `npm run develop` in `examples/app-router`
- Verify that no warning or error appears in the console

### DIFF
--- a/examples/app-router/package.json
+++ b/examples/app-router/package.json
@@ -3,7 +3,9 @@
   "private": true,
   "scripts": {
     "clean": "rimraf .next",
+    "build": "next build",
     "develop": "next",
+    "start": "next start",
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


peterbe (commented):
Does `suppressHydrationWarning` only set it for the `<html>` tag?
What if there are other juicy warnings?

peterbe:
By the way, I confirmed that this solves the immediate problem.

joshblack:
@peterbe great question, I believe it will only suppress mismatches for the `html` tag (and maybe one more level deep) and so any other mismatches down the tree will still be reported.

Let me know if that's not the case though!

broccolinisoup (approved):

PR #4313, opened 2024-02-27
feat: add live region support with Alert, Status, and Announce

joshblack:
Add support for using the `@primer/live-region-element` package through three components:

- `Alert`, analogous to `role="alert"`, used for assertive announcements
- `Status`, analogous to `role="status"`, used for polite announcements
- `Announce`, a wrapper component that makes announcements when its text contents have changed using the `<live-region>` element

These components use `Box` under the hood and can be styled and customized as-needed. They can be used as the visual parts of a component. They can also be hidden when wrapped with `VisuallyHidden`.

> [!NOTE]
> Wanted to add a section in case you're wondering why we aren't using `role="alert"` and `role="status"` directly 👀 

The core challenge we're trying to work around is adding the corresponding live region and message _at the same time_.

Put another way, support for rendering and announcing the following snippet is not great:

```html
<div role="status">Test message</div>
```

Broadly speaking, making sure that the element exists first before adding the message has better support than doing both at the same time. This is what the components in this PR, and corresponding `@primer/live-region-element`, help to accomplish.

#### New

- Add `Alert`, `Status`, and `Announce` as internal components

#### Changed

- Add the file path to `Announce` to the list of side effects to the package due to its usage of `@primer/live-region-element/define`

### Rollout strategy

This is adding components to our internal set of components and does not impact the public API of `@primer/react`

### Testing & Reviewing

- [ ] Checkout the storybook stories for the `Status` component and verify that the live region works as expected (let me know if I can help out with instructions for this!)
- [ ] Go through the tests to make sure they have accurate coverage for the component
- [ ] Try to use this component in the `app-router` example and verify that it can be called in a server component, has no SSR errors, and appropriately announces on render
- [ ] Let me know if you have any feedback on the approach here! Very open to feedback, suggestions, and ideas 👀


packages/react/src/internal/components/Alert.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import {Announce, type AnnounceProps} from './Announce'
+
+export type AlertProps = AnnounceProps
+
+export function Alert({children, ...rest}: AlertProps) {
+  return (
+    <Announce {...rest} politeness="assertive">
```
siddharthkp, line 8: (not blocking) This component feels pretty minimal, should we reduce just 1 Announce component where politeness can be a required prop?

packages/react/src/internal/components/Announce.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * `children` passed in to screen readers using the given politeness level. It
+ * will also announce any changes to the text content of `children`
+ */
+export function Announce({children, politeness = 'polite', ...rest}: AnnounceProps) {
+  const ref = useRef<ElementRef<'div'>>(null)
+  const savedPoliteness = useRef(politeness)
+
+  useEffect(() => {
+    savedPoliteness.current = politeness
+  }, [politeness])
```
siddharthkp, line 26: (non blocking) Why do we save this in a ref, but also update it whenever it changes?

Is this the equivalent of using it from props or do we intentionally only want to use it after an effect lifecycle?

packages/react/src/internal/components/Announce.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  // Announce the initial message, this is wrapped in `useEffectOnce` so that it
+  // does not announce twice in StrictMode
```
siddharthkp, line 29: (not blocking) haters gonna say this is an anti pattern. ship it!

packages/react/src/internal/components/Announce.tsx
```diff
@@ -0,0 +1,72 @@
...
+      characterData: true,
+    })
+
+    return () => {
+      observer.disconnect()
+    }
+  }, [])
+
+  return (
+    <Box {...rest} ref={ref}>
```
siddharthkp, line 68: (non blocking) Ooh, interesting! Is this a box because we want folks to be able to build a sub-component using this?

packages/react/src/internal/components/Status.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return <Status>{message}</Status>
+}
+
+export const VisuallyHiddenStory: StoryObj = {
+  name: 'VisuallyHidden',
+  render: () => {
+    return (
+      <>
+        <p>This is an example</p>
+        <VisuallyHidden>
```
siddharthkp, line 32: (non blocking) Ah, I was wondering if this should be a prop on the component or the default on the component. I guess we can find out once we start using it!

siddharthkp (approved):
Left a few comments, none of them are blocking. Approving in advance

packages/react/src/internal/components/Alert.tsx
joshblack: The big motivator for `Alert` and `Status` was that they correspond to the equivalent roles that people tend to look for. I almost would want `Announce` as an implementation detail in case we discover that we can implement `Alert` and `Status` differently in the future lol.

I'd love to keep them as a result, and encourage their use over `Announce`, since it'd be more clear what `Alert` or `Status` would be internally than `Announce`, especially to new contributors. It'd also let us swap out semantics in case we want to switch them up in the future

packages/react/src/internal/components/Announce.tsx
joshblack: @siddharthkp I don't know if it necessarily has to be a `Box`, just wanted a quick way to use the `sx` prop 😂 

Basically just want `Status` and `Alert` to be nodes that people can style as-needed since they can be used for visual content.

packages/react/src/internal/components/Status.stories.tsx
joshblack: Ideally, I want people to think of `Status` as `<div role="status">`. It won't do anything more or less than that role, but under-the-hood we make it work. So you can style it if you're using it visually, or if you want to hide then you use `VisuallyHidden`.

packages/react/src/internal/components/Alert.tsx
siddharthkp: Interesting, makes sense!

packages/react/src/internal/components/Announce.tsx
siddharthkp: Ah I see!

packages/react/src/internal/components/Status.stories.tsx
siddharthkp: > Ideally, I want people to think of Status as <div role="status">

Fair... but also do you see folks using it with visible elements? Maybe in Banner?

joshblack: @siddharthkp definitely, I think so.

PR #4250, opened 2024-02-10
Add a new entry point as `next` and update draft Tooltip to export it from next

broccolinisoup:
Based on the [Tooltip v2 adoption discussion](https://github.com/github/primer/discussions/2754), this PR adds a new entry point as `next` and export Tooltip from it with an updated documentation. The discussion is still open for feedback for docs consolidation, so please feel free to share any ideas/thoughts/concerns 🙌🏻  

#### New

- `next` entry point

#### Changed

- Renamed the draft/experimental Tooltip in docs and storybook to become Tooltip v2

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Testing & Reviewing

- Is it clear why there are two version of Tooltips under the components?
- Is it visible enough that the import path of Tooltip v2 is different than the usual import path?
- Checking out the Tooltip and Tooltip v2 documentation and storybook and make sure they reflect the right version of the component
- If possible testing the export would be great. i.e `import {Tooltip} from '@primer/react/next'` and make sure it imports the component properly.
- [Integration PR (Github stuff only)](https://github.com/github/github/pull/309975)

### Merge checklist

- [ ] Added/updated tests
- [x] Added/updated documentation
- [x] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #4317, opened 2024-02-28
Tooltip: Take the `position:absolute` back and make sure the v1 and v2 ids are same for new docs

broccolinisoup:
`position:absolute` on the tooltip element was introduced to in https://github.com/primer/react/pull/4250 to fix an issue on the [Tooltip V2 react docs](https://primer-c47bc7bb55-13348165.drafts.github.io/TooltipV2) (tooltip is not visible) but it ended up causing a problem at dotcom (specifically memex), flickering and shifting when the tooltip appears. Without this change, everything except [the docs](https://primer-c47bc7bb55-13348165.drafts.github.io/TooltipV2) is working fine. I think the issue is not on the tooltip but on docs, maybe? I'd like to take this change back to unblock the version upgrade and can look into further later. Though, all react docs is going to be migrated over to the new docs and the way that examples are rendered differently (from stories) on the new site so I'll see if the same issue persistent on the new docs first.

#### Changed

- Updated the tooltip v2's id to be the same as v1 for consolidation on the new docs site

#### Removed

- Removed `position: absolute` from the tooltip element.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- Can view the stories and make sure the tooltip appears correctly
- Unfortunately, (as far as I know) there is no way of testing docs changes in the new site until the release is out

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/docs/content/TooltipV2.mdx
+++ b/docs/content/TooltipV2.mdx
@@ -1,5 +1,5 @@
 ---
-componentId: tooltip_v2
+componentId: tooltip
 title: Tooltip v2
 status: Beta
 a11yReviewed: true
--- a/packages/react/src/TooltipV2/Tooltip.tsx
+++ b/packages/react/src/TooltipV2/Tooltip.tsx
@@ -44,12 +44,10 @@ const StyledTooltip = styled.div`
   }
   /* class name in chrome is :popover-open */
   &[popover]:popover-open {
-    position: absolute;
     display: block;
   }
   /* class name in firefox and safari is \:popover-open */
   &[popover].\\:popover-open {
-    position: absolute;
     display: block;
   }
 
--- a/packages/react/src/TooltipV2/Tooltip.tsx
+++ b/packages/react/src/TooltipV2/Tooltip.tsx
@@ -23,7 +23,6 @@ const StyledTooltip = styled.div`
   /* Overriding the default popover styles */
   display: none;
   &[popover] {
-    position: absolute;
     padding: 0.5em 0.75em;
     width: max-content;
     margin: auto;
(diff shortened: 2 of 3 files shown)


packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -23,7 +23,6 @@ const StyledTooltip = styled.div`
   /* Overriding the default popover styles */
   display: none;
   &[popover] {
-    position: absolute;
```
broccolinisoup, line 26: This change was introduced to in https://github.com/primer/react/pull/4250 to fix an issue on the Tooltip V2 react docs but it ended up causing a problem at dotcom (specifically memex), flickering and shifting when the tooltip appears. Without this line, everything except the docs is working fine. I think the issue is not on the tooltip but on docs, maybe? I'd like to take this change back to unblock the version upgrade and can look into further later. Though, all react docs is going to be migrated over to the new docs and the way that examples are rendered differently (from stories) on the new site so I'll see if the same issue persistent on the new docs first.

docs/content/TooltipV2.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-componentId: tooltip_v2
+componentId: tooltip
```
broccolinisoup, line 2: Found out from Cameron that to be able to consolidate the versions of the same component on the new docs site, the ids need to be the same.

packages/react/src/TooltipV2/Tooltip.tsx
broccolinisoup: The docs problem happens because the examples come after the first scroll, so this is why we added `position:absolute`. However, when I removed the `position:absolute` the instances at dotcom that come after the the first scroll are still the same. Seems like taking it back will fix things and not introduce any other issues as well.

joshblack (approved):

pksjce (approved):

Issue #4315, opened 2024-02-27
`SelectPanel` should compare Objects based on `key` or `id` instead of the Objects themselves


ajhenry:
### Description

When passing in objects to the `SelectPanel` component, the selected items may not render correctly since they are doing direct object comparison:

https://github.com/primer/react/blob/ea44386e4d7afae7bfd07a679604baaecca55965/packages/react/src/SelectPanel/SelectPanel.tsx#L135-L155

If the items are defined in the component, they could rerender and create new object references. Then this comparison check will always return false as the original objects are held in `useState`.

Instead, we should rely on comparing the `key` or `id` props as these are guaranteed* to be unique per object.

So our comparison function would instead look something like the following:

```typescript
onAction: (itemFromAction, event) => {
          item.onAction?.(itemFromAction, event)

          if (event.defaultPrevented) {
            return
          }

          if (isMultiSelectVariant(selected)) {
            const otherSelectedItems = selected.filter(selectedItem => selectedItem.key !== item.key)
            const newSelectedItems = selected.find(selectedItem => selectedItem.key === item.key) ? otherSelectedItems : [...otherSelectedItems, item]

            const multiSelectOnChange = onSelectedChange as SelectPanelMultiSelection['onSelectedChange']
            multiSelectOnChange(newSelectedItems)
            return
          }

          // single select
          const singleSelectOnChange = onSelectedChange as SelectPanelSingleSelection['onSelectedChange']
          singleSelectOnChange(item.key === selected.key ? undefined : item)
          onClose('selection')
        },
```

cc: @ipc103 

### Steps to reproduce

Minimal Repo
https://codesandbox.io/p/sandbox/dreamy-cache-msfqyq

### Version

v35.26.0

### Browser

Chrome

ajhenry:
@ipc103 and I are happy to take this on 😄

PR #4324, opened 2024-02-28
Allow non-memoized item lists

ipc103:
Closes https://github.com/primer/react/issues/4315

CC @ajhenry 

### Changelog
Fixes a bug in the `SelectPanel` component which required the same object to be stored in the `items` and `selectedItems` lists.

#### New
N/A

#### Changed

Fixes a bug where `SelectPanel` wouldn't update state correctly when the `items` list contained different objects on re-render. 

#### Removed

N/A

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing
See the updated test cases for a minimal repro example. 

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -124,9 +124,24 @@ export function SelectPanel({
     }
   }, [placeholder, renderAnchor, selected])
 
+  const testIsItemSelected = React.useCallback((selectedItem: ItemInput, item: ItemInput) => {
+    const itemType = typeof selectedItem
+    if (itemType === 'object') {
+      if (selectedItem.hasOwnProperty('id')) {
+        return selectedItem.id === item.id
+      }
```
ajhenry, line 132: I think we should also try testing for `key` since that's React's way of differentiating component changes

ipc103: I thought about that, but I actually think it's better to add the `key` when rendering over the list instead of trying to add it preemptively. Not super opinionated though if we want to use the key here!

siddharthkp:
Heads up: We can test if these changes are compatible with instances in gh/gh by creating an integration PR with this action: https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md

siddharthkp:
Followup: Here's the integration PR https://github.com/github/github/pull/314966

but I'm expected it to fail today because there are a couple unmerged fixes with gh/gh 😓

ipc103:
Thanks @siddharthkp! Let me know if/how I can help.

siddharthkp:
Update: Here's the new integration PR: https://github.com/github/github/pull/315122

The CI is failing here. Is it because of the changes in this PR or unrelated? 🤔

ipc103:
It does seem like some of those are related to this PR. I wonder if there are spots where we're working around the current behavior, or otherwise depending on it. I'll take a look and see if I can make this a non-breaking change!

ipc103:
Looking more closely I'm not convinced those failures are related. They seem to be about unrelated components. I'm going to re-run just to make sure they aren't flakes or something.

ajhenry:
Hey @siddharthkp anything we can help out with here? Looking through the test failures, it doesn't _seem_ like this change should be involved but happy to debug where we can!

siddharthkp:
> Hey @siddharthkp anything we can help out with here? Looking through the test failures, it doesn't seem like this change should be involved but happy to debug where we can!

Hey! sorry for the silence! Yeah, it's always tricky to tell when the integration tests have other errors as well 😢 

If this is not urgent, I'm going to wait for the next release to go out so that I can have a clean slate to test this! Hope that's okay!

ipc103:
Thanks @siddharthkp - that should work for us. The workaround is to just make sure that the list is memoized, so we can keep that up in the meantime.

ipc103:
@siddharthkp 👋 Hi! What do you think about this one - any chance we could cut another integration test branch and see if this change plays more nicely with the newer release?

ipc103:
@siddharthkp I went ahead and created a new integration PR: https://github.com/github/github/pull/325842 I used the latest release candidate since it seems like canary builds are on pause at the moment. Did I do it right? If so, I think everything is passing 🤞

siddharthkp:
- Implemented in https://github.com/primer/react/pull/5073

PR #4325, opened 2024-02-28
chore(project): update to use Node.js v21

joshblack:
This is a follow up to #4316. This updates our local Node.js version to be v21 which does not seem to show the same problems as v20.

#### Changed

- Update GitHub Actions to use Node.js v21
- Update `.nvmrc` to use Node.js v21
- Update project root `engines` field specify v21

### Rollout strategy

This is a change to the package manager for the project

### Testing & Reviewing

- Verify that CI installs aren't flaky 😅


broccolinisoup (approved):
🎉

joshblack:
Ugh, nevermind. Seeing the same `ECONNRESET` errors more reliably now.

PR #4327, opened 2024-02-29
TooltipV2: Set the tooltip position when the popover open not when the toggle event is triggered

broccolinisoup:
### Changelog

Tooltip element by default has `position: fixed` but this doesn't work cases where the tooltip element is after the first scroll. So the position needs to be `position: absolute`. However with this change, there is another bug occurs where the tooltip element first appears in a wrong `left` and `right` position therefore causes screen to wiggle. The recording below shows this behaviour. 

I think this occurs because we set the position when the `toggle` event is triggered and the `toggle` event is triggered _after_ we set the tooltip element to be visible. I updated the position logic to be called in the same function we set the visibility (tooltip.showPopover()) rather than being called as a callback in the `toggle` event. I ran the tests and seems like the bug is gone. 

https://github.com/primer/react/assets/1446503/df569781-a55d-4c6d-9bc9-7321ed4886ad

#### Changed

- Position setting was used to be called as a callback function in the the `toggle` event listener but now it is called when the `:popover-open` pseudo class is added to the tooltip element.

#### Removed

- Removed `toggle` event listening. 

I don’t know why we would need to listen to the toggle event, since we know exactly when the tooltip is shown. 🤷🏻‍♀️  It seems like one is enough [if I read the docs correctly](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using#showing_popovers_via_javascript)

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -196,7 +197,23 @@ export const Tooltip = React.forwardRef(
 
     const openTooltip = () => {
       if (tooltipElRef.current && triggerRef.current && !tooltipElRef.current.matches(':popover-open')) {
-        tooltipElRef.current.showPopover()
+        const tooltip = tooltipElRef.current
+        const trigger = triggerRef.current
+        tooltip.setAttribute('popover', 'auto')
```
keithamus, line 202: This should always be set so that the browser can correctly expose the right AT nodes. We should not set it as the tooltip is being opened.

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -23,6 +23,7 @@ const StyledTooltip = styled.div`
   /* Overriding the default popover styles */
   display: none;
   &[popover] {
+    position: absolute;
```
keithamus, line 26: AIUI this is the main thing that is doing all the heavy lifting. I think you could just apply this line and ignore the rest and it would still work?

packages/react/src/TooltipV2/Tooltip.tsx
broccolinisoup: That makes sense. Although [the tooltip is not exposed to AT](https://github.com/primer/react/pull/3910) when it is not opened but I think it still makes sense to set the initial attributes in the render rather than when it is opened.

packages/react/src/TooltipV2/Tooltip.tsx
broccolinisoup: Yeah definitely this line does all the heavy lifting and actually the rest is not _directly_ related to this change but the existing issue* just became more visible when we added this line (no idea why 🤷🏻‍♀️ ). 

*The existing issue (at least it is an issue in my opinion) is where we set the position. 

Previously the flow was:
1) hover/focus event is triggered
2) `tooltip.showPopover()` function is called
3) `:popover-open` attribute is added to the tooltip div
4) Adding `:popover-open` triggers the `toggle` event as well as triggers the animation that changes the tooltip's opacity from 0 to 1
5) Position is set in the `toggle` event's callback function.

I believe the wiggling issue (video in the PR description) happens because we set the position a little too late (step 5) but we make the tooltip visible in the step 3. So in this PR I am setting the position in Step 3 and that makes the toggle event redundant. 

Let me know if you have any concern or suggestions.

siddharthkp (approved):
Change in order of functions makes sense to me!

PR #4331, opened 2024-02-29
Add `word-break` to ActionList

langermank:
Allow long strings without natural breaks to `break-word` in ActionList item.

Before:
dont break word

After:
break word


siddharthkp:
Surprising that this doesn't update any visual tests! Do you mind adding one just so that we have it for the future!

langermank:
Thanks @siddharthkp, good call. I added a story 

break word

mperrotti (approved):

Issue #3020, opened 2023-03-13
Flash has auto styling that interfere with SVG elements


liranelisha:
### Description

I am writing to report an issue with a Flash component that I have been using on my website. The component appears to have an auto-styling feature that is interfering with the SVG elements on my web pages.

Specifically, the auto-styling feature is causing the SVG elements to be displayed incorrectly with fixed marginRight of 2 (8px). This is an issue, as the SVG elements are a part of the design and functionality of my pages.

I have tried adjusting the CSS styles for the SVG elements to no avail, as the auto-styling seems to get a priority. The only way to resolve the issue seems to be to add the `!important`.

I would greatly appreciate it if you could look into this issue and provide a solution as soon as possible. If necessary, I can provide you with access to my website and the code for the Flash component and the SVG elements.

Thank you for your prompt attention to this matter.

![image](https://user-images.githubusercontent.com/76681072/224741082-c1b93cf7-ad0e-4c2a-bfab-66884253ef09.png)

### Steps to reproduce

create a `StyledOcticon` inside a `Flash` component

### Version

35.21.0

dejnon:
Here is another reproduction:
<img width="993" alt="image" src="https://user-images.githubusercontent.com/873549/224742754-adaf5853-807f-4064-bc34-d7956f8f4ae5.png">

liranelisha:
this issue was seen in the devportal
example: https://devportal.githubapp.com/devportal/apps/github?tab=deployments
when we have the "external link" icon when the text include a link to external source

lesliecdubs:
Thanks for the report, this does seem to be a bug. 

Internal note: I am removing the `react` label from this issue for now so that Primer Design can hopefully prioritize this. @tallys let us know if that's not doable.

tallys:
@lukasoppermann let's take a look at this as part of our work on flash/banner

lukasoppermann:
Yes. 👍

liranelisha:
sill happening...

lukasoppermann:
@liranelisha we will keep this in mind for the new component.

lukasoppermann:
@liranelisha could you verify that the issue is fixed in the new component?

https://primer.style/components/banner/react/draft

liranelisha:
i think we are currently unable to upgrade to version 36.27.0 as it cause some unit testing fails on our side. but i'll keep on my list to test it and reopen if needed once upgraded

TylerJDev:
Hey @liranelisha! 👋 

[...]

PR #4335, opened 2024-03-01
feat(Banner): add banner component

joshblack:
Closes https://github.com/github/primer/issues/3200

Add support for the Banner component to Primer React.

#### New

- A new Banner component is available under experimental/drafts
- Supporting stories, tests, and docs for Banner

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

- [ ] Verify implementation matches API design in [API Document](https://github.com/github/primer/issues/3117)
- [ ] What do you think of the API design? Is it flexible enough for us to iterate on? Does it have any gaps or are there opportunities we should consider?
- [ ] Verify responsive behavior for buttons
- [ ] Verify that all the limitations for banner are incorporated here and tested, for example when custom icons are allowed or when banners are allowed to be dismissible


joshblack:
Note for reviewers, would love feedback on visual details since it's been a while since I touched this based on the Figma spec 👀 I also feel like there are some things that I might have mistranslated so would appreciate an extra eye on.

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,323 @@
...
+  description?: React.ReactNode
+
+  /**
+   * Provide an icon for the banner.
+   * Note: Only `variant="info"` banners should use custom icons
+   */
+  icon?: React.ReactNode
+
+  /**
+   * Optionally provide a handler to be called when the banner is dismissed.
```
khiga8, line 24: It looks like we're expecting the consumer to manage focus upon dismiss and that worries me a little bit given it seems very easy to forget.

I'm curious to hear your thoughts on:

* why we would or wouldn't want to build focus on dismiss into the component.
* how we might encourage people to implement focus management correctly.

I'd love if we can leverage some of the programmatic focus management logic that Lindsey introduced to the global flash. 🤔

packages/react/src/Banner/Banner.examples.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const bannerRef = React.useRef<React.ElementRef<typeof Banner>>(null)
+  const focus = useFocus()
+
+  return (
+    <>
+      {hasError ? (
+        <Banner
+          ref={bannerRef}
+          title="Error"
+          description={<Alert>Something went wrong. Please try again later.</Alert>}
```
khiga8, line 32: Though I understand that heading semantics won't be conveyed through a live region, I'm wondering whether it would make sense to encourage including it as part of the live region announcement since it may contain information that isn't part of the description. 🤔 

What would that look like?

packages/react/src/Banner/Banner.features.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+  component: Banner,
+} satisfies Meta<typeof Banner>
+
+export default meta
+
+export const Critical = () => {
+  return (
+    <Banner
+      title="Critical"
+      description={
```
khiga8, line 19: For my own React learning, I'm curious why you went with introducing `description` as a prop, versus a slot!

packages/react/src/Banner/Banner.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+    render(<Banner title="test" />)
+    expect(screen.getByRole('region', {name: 'test'})).toBeInTheDocument()
+  })
+
+  it('should label the landmark element with the title prop', () => {
+    render(<Banner title="test" />)
+    expect(screen.getByRole('region')).toEqual(screen.getByLabelText('test'))
+  })
+
+  it('should default the title to a h2', () => {
```
khiga8, line 17: I am a bit worried that having a default heading level will result in banners being added with incorrect heading hierarchy. I worry most people will set `title` without making the connection that it render a heading with a level that should be adjusted for their usecase.

What do you think?

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,323 @@
...
+  }, [titleId])
+  const dismissible = variant !== 'critical' && onDismiss
+  const hasActions = primaryAction || secondaryAction
+
+  if (__DEV__) {
+    // Note: __DEV__ will make it so that this hook is consistently called, or
+    // not called, depending on environment
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    useEffect(() => {
+      const title = document.getElementById(titleId)
```
khiga8, line 79: To circle back on - https://github.com/github/primer/pull/3090#discussion_r1523866786!

khiga8 (commented):
WOO! Excited to see this. Kicking off with thoughts/questions!

e2e/components/Banner.test.ts
```diff
@@ -0,0 +1,93 @@
+import {test, expect} from '@playwright/test'
```
khiga8, line 1: For my learning... is it fair to say that the majority of things (e.g. keyboard interactions, markup behavior) can be well-tested with jest unit tests? 

I'm curious if there are any plans or any needs to expand e2e testing coverage for Primer react component beyond axe and snapshot tests.

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,323 @@
...
+    --banner-bgColor: ${get('colors.success.subtle')};
+    --banner-borderColor: ${get('colors.success.muted')};
+    --banner-icon-bgColor: ${get('colors.success.fg')};
+  }
+
+  &[data-variant='upsell'] {
+    --banner-bgColor: ${get('colors.done.subtle')};
+    --banner-borderColor: ${get('colors.done.muted')};
+    --banner-icon-bgColor: ${get('colors.done.fg')};
+  }
```
langermank, line 151: Once https://github.com/primer/react/pull/4491 merges, you can access `upsell` colors. However, these are just CSS variables and won't really have a fallback. If you want to keep using the `get` configuration, you can just use the `done` colors as the fallback. Otherwise maybe we just move to pure CSS variables? 

```
--fgColor-upsell
--bgColor-upsell-muted
--borderColor-upsell-muted
```

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,323 @@
...
+    --banner-borderColor: ${get('colors.attention.muted')};
+    --banner-icon-bgColor: ${get('colors.attention.fg')};
+  }
+
+  /* BannerIcon ------------------------------------------------------------- */
+
+  .BannerIcon {
+    display: grid;
+    place-items: center;
+    padding: 0.5rem;
```
langermank:
```suggestion
    padding: 0.375rem 0.5rem;
```
I'm pulling this from PVC, where we went pretty specific here to align the title with the trailing actions labels. Do you hate it? Is it okay? I think it's acceptable 😆

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,323 @@
...
+    column-gap: 0.25rem;
+  }
+
+  /* BannerContent ---------------------------------------------------------- */
+
+  .BannerContent {
+    display: grid;
+    row-gap: 0.25rem;
+    grid-column-start: 1;
+    margin-block: 0.5rem;
```
langermank:
```suggestion
    margin-block: 0.375rem;
```
Matched the BannerIcon alignment

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  /* BannerDismiss ---------------------------------------------------------- */
+
+  .BannerDismiss {
+    display: grid;
+    place-items: center;
+    padding: 0.5rem;
+    width: 2rem;
+    height: 2rem;
```
langermank, line 238:
```suggestion

```
I don't think we need this since IconButton already has explicit h/w

packages/react/src/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+          to enable two-factor authentication as an additional security measure.
+        </>
+      }
+      primaryAction={<Banner.PrimaryAction>Button</Banner.PrimaryAction>}
+      secondaryAction={<Banner.SecondaryAction>Button</Banner.SecondaryAction>}
+    />
+  )
+}
+
+export const Playground: StoryObj<typeof Banner> = {
```
langermank, line 34: What do you think about adding some more control options to the Playground? I think it would be nice to have...

- description (text)
- title (text)
- maaaybe do a custom icon slot, but its annoying and hard in Storybook.. 
- maybe toggle the secondary actions to show/hide?

langermank (commented):
Just a few little nitpicks, otherwise this looks SO FREAKING GREAT

packages/react/src/Banner/Banner.examples.stories.tsx
joshblack: @khiga8 would something like this maybe work?

```tsx
description={
  <Alert>
    <VisuallyHidden>Error:</VisuallyHidden> Something went wrong. Please try again later.
  </Alert>
}
```

e2e/components/Banner.test.ts
joshblack: @khiga8 great questions 🔥 Thankfully I think with: https://www.npmjs.com/package/@testing-library/user-event we can get pretty robust keyboard interaction testing in Jest.

No plans so far to extend what is in the e2e tests, I think the only other case that has come up is when we need to test behavior when the browser is resizing (like for underline nav).

packages/react/src/Banner/Banner.test.tsx
joshblack: Makes total sense to me! Would the preference then be for this to always be an explicit heading level that the user of the component provides?

packages/react/src/Banner/Banner.features.stories.tsx
joshblack: @khiga8 love this question. Just want to double-check, by slot did you mean the html `<slot>` or more so the PVC slots idea?

I think the tl;dr for this is that this setup is (at least as I understand it) how slots are typically done for React components. I'm definitely thinking of [this article](https://engineering.atspotify.com/2023/05/multiple-layers-of-abstraction-in-design-systems/) when I'm saying that though so I'm not sure how ubiquitous that is. Specifically, they have a good breakdown that I love to keep in mind in the React world:

A row breakdown of the different types of Component API decisions that you can make. On the left side, there is a config pattern where the component takes in dedicated props and strings that are their values for a Card component. In the middle is the slots pattern that expands the values of props to include other react components. On the far right side, they use a custom pattern where each of the props are now children of the card component

Hope that makes sense, would love to know what you think! 👀

packages/react/src/Banner/Banner.tsx
joshblack: @khiga8 great points, thanks for bringing them up. Maybe a small change we could make here to accommodate what your suggesting is to make is so that anytime `onDismiss` is provided that we have an explicit `returnFocusRef` (or something similar) prop so that focus is always explicitly managed? We could also make it so that `onDismiss` expects a return value which is the element to return focus to.

Happy to explore either option, with potentially the latter feeling more similar to what I'm assuming people will write (similar to the storybook example).

Let me know what you think! Happy to bake in this, or other approaches, into the component 👍

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,323 @@
...
+    display: flex;
+    column-gap: 0.5rem;
+    align-items: center;
+  }
+
+  .BannerActions [data-hide-on-sm] {
+    display: none;
+  }
+
+  @media screen and (min-width: 544px) {
```
lukasoppermann, line 211: Hey @joshblack I see that you went with a media query instead of a container query. Is this due to baseline support?

We need to support something like this though, how do we do it?

CleanShot 2024-04-12 at 08 52 45@2x

joshblack: Hey @lukasoppermann! 👋 Great question, my plan is to follow up with container query support since we need to make sure media query support works as a fallback because container queries aren't available in all the browsers we target. I'll make an issue for it in the parent epic if that works for you 👍 

Sadly, like you pointed out, in the fallback case it won't correctly support those scenarios 😞 Is there anything you would like me to try out or any ideas to make sure we degrade gracefully here?

joshblack: @lukasoppermann my plan was definitely as a first pass to make sure things worked with `@media` and then add in container support as a progressive enhancement because of that baseline support 😞 

Great question for these scenarios, they seem tricky to do without any changes on the user end without `@container`. Maybe for these scenarios what we could do is add an explicit prop to override any implicit layout? Could be something as simple as "direction="column" or we could do something like a `type` or `variant` prop while we wait for our baseline to improve.

Let me know what you think!

packages/react/src/Banner/Banner.stories.tsx
joshblack: Great callout,  added the following:

- [x] title
- [x] description
- [x] primaryAction
- [x] secondaryAction

No icon support for now 😓 hope that's okay lol

packages/react/src/Banner/Banner.examples.stories.tsx
khiga8: To reduce redundancy around the heading text appearing in the banner twice for screen reader users, would something like this be possible?

```.tsx
<Banner>
  <Status>
    <Banner.Title>Subscription renewed</Banner.Title>
    <Banner.Description>Your subscription has been successfully renewed until May 5, 2024. </Banner.description>
  </Status>
</Banner>
```

cc: @ichelsea if you have other thoughts!

packages/react/src/Banner/Banner.test.tsx
khiga8: Yes, I think that would be the most sure-way to encourage consumers to be mindful about the heading level and have proper heading hiearchy. I can't think of why we wouldn't want to require it, but let me know if there are concerns I'm not thinking of.

Another way to approach this would be to extend this lint rule - [a11y-explicit-heading](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/rules/a11y-explicit-heading.js).

packages/react/src/Banner/Banner.tsx
khiga8: > We could also make it so that onDismiss expects a return value which is the element to return focus to.

Are you thinking that the component would then handle placing focus on whatever the return value is?

joshblack: > Are you thinking that the component would then handle placing focus on whatever the return value is?

Exactly 👍 

We could also totally follow this behavior when `onDismiss` is called (specifically the `getElementToFocus` helper) and restore it that way, as well. Would it be worth making an issue for this for us to explore/try out different implementations or is this something that should be incorporated in this PR?

packages/react/src/Banner/Banner.examples.stories.tsx
joshblack: @khiga8 I bet we could make it work, we could also announce the live region itself if that would be desirable (e.g. something like `Banner as={Status}`

packages/react/src/Banner/Banner.test.tsx
joshblack: Sounds good to me 👍 I think the only thing that comes up for this is that naming the prop on `Banner` is awkward (compared to on `Banner.Title`). 

Here are some options that came to mind, curious if any of them resonate or if there is a pattern that you would recommend following for this scenario:

- `titleHeadingLevel`
- `titleHeading`
- `heading`
- `headingLevel`

These could be props that accept a number (2-6) or the explicit tag (h2-h6).

packages/react/src/Banner/Banner.tsx
joshblack: @lukasoppermann just wanted to provide an update, added support for a container query along with a flexbox fallback!

packages/react/src/Banner/Banner.tsx
khiga8: > Would it be worth making an issue for this for us to explore/try out different implementations

This sounds good to me! :)

joshblack: Added over at: https://github.com/github/primer/issues/3222

Would love to pair on this if you're interested! Also wanted to check to see if it would be okay if this work was in a separate PR?

packages/react/src/Banner/Banner.test.tsx
joshblack: `titleAs` could definitely work! I also don't mind restricting these since `as` is technically for _any_ element and we really do want these specifically to only be headings in a specific range. Super torn on this, but if I had to pick I guess I would like:

```tsx
<Banner headingLevel={2}></Banner>

<Banner>
  <Banner.Title headingLevel={2}
</Banner>
```

Curious what you think

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,360 @@
...
+ * that value.
+ */
+const StyledBanner = styled.div`
+  display: grid;
+  grid-template-columns: auto minmax(0, 1fr) auto;
+  align-items: start;
+  background-color: var(--banner-bgColor);
+  border: var(--borderWidth-thin, 1px) solid var(--banner-borderColor);
+  padding: var(--base-size-8, 0.5rem);
+  border-radius: var(--borderRadius-medium, ${get('radii.2')});
```
lukasoppermann, line 128: @joshblack do we need the fallback here?

packages/react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+  @supports (container-type: inline-size) {
+    container: banner / inline-size;
+  }
+
+  &[data-variant='critical'] {
+    --banner-bgColor: ${get('colors.danger.subtle')};
```
lukasoppermann, line 135: @joshblack Can we use v8 tokens instead and avoid `get` all together?

```suggestion
    --banner-bgColor: var(--bgColor-danger-muted);
```

lukasoppermann:
We need to increase the spacing below the buttons.

It is 6px to short (in figma there is an inner padding in the action container when the buttons are below the content).

CleanShot 2024-04-19 at 13 38 22@2x

lukasoppermann:
Hey @joshblack,

this is really great already. Do you think we could add the _in sidebar_ example to the playground as well? 
https://primer-65defb5e3c-13348165.drafts.github.io/storybook/?path=/story/drafts-components-banner--playground

E.g. like this:

Mockup for both banner sizes next to each other

lukasoppermann:
When the title is disabled, there is a larger space below the actions in the banner than above.

space below buttons

packages/react/src/Banner/Banner.tsx
joshblack: @lukasoppermann sadly not yet 😞 this will likely be a part of v37 when we officially shift over.

packages/react/src/Banner/Banner.tsx
joshblack: I believe so, I think any time we refer to a custom property there is a chance that it is not defined because in PRC we don't require a team to set them (yet) like in github/github

packages/react/src/Banner/Banner.tsx
khiga8: Definitely down to pair! 

My understanding is that this is a draft/first implementation and something that will be iterated on based on learnings :)

packages/react/src/Banner/Banner.test.tsx
khiga8: I like that! That's a great point that `as` is technically for all any elements, and I do like the idea of moving away from it.

joshblack:
Thanks @lukasoppermann! I believe I addressed the following:

- [x] When the title is disabled, there is a larger space below the actions in the banner than above.
- [x] Do you think we could add the in sidebar example to the playground as well?
- [x] Another thing, whenever a title is present, the actions (not close button) should be moved below the text.
- [x] It is 6px to short (in figma there is an inner padding in the action container when the buttons are below the content).

langermank (approved):

lukasoppermann:
@joshblack we should add a test to make sure svgs work to avoid this issue: https://github.com/primer/react/issues/3020

joshblack:
@lukasoppermann we should be good to go thankfully! What we have right now will only be behind the `.BannerIcon` class name so should completely avoid the problem 🤞

Issue #4336, opened 2024-03-01
`colorSchemes` calculation slows down initial load


manuelpuyol:
Primer React parses color schemes on runtime using a reduce on a pretty big object:

https://github.com/primer/react/blob/32b0cf77fcb4112f00da6ffb93a52875d23b9370/packages/react/src/theme.ts#L64-L76

This is run whenever Primer is loaded in a JS bundle, which means that clients will have to run this expensive calculation before even starting to render anything.

Primer could pre-compile `colorSchemes` and include it in the bundle instead of having clients calculating it :)

I wrote two scripts comparing load times between the two methods here https://gist.github.com/manuelpuyol/f0b3c140c579e195374b30f25d5da833

In my tests, the load times went from ~15ms to <1ms!

PR #4337, opened 2024-03-01
Add precompile step for colorSchemes

manuelpuyol:
Closes #

https://github.com/primer/react/issues/4336

### Changelog

Precompile colorSchemes to make load times faster.

#### New

- Added a `precompile-color-schemes` build step that will output a `.ts` file with the precompiled object.

#### Removed

- Removed colorSchemes compilation step from runtime

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp:
This looks promising! 

Let's test it with gh/gh for any accidental errors

joshblack (commented):
Thanks for taking the time to do this! Really appreciate you exploring this to improve performance in this area.

One question I had was if it was possible to statically update `src/legacy-theme/ts/colors/index.ts` into the intended format that the precompile step is taking or if this is something that will always need to execute, whether that's on load or at a precompile step.

manuelpuyol:
### Bundle diff

before:

CleanShot 2024-03-06 at 10 18 33

after:

CleanShot 2024-03-06 at 10 18 43

The raw bundle is slightly bigger, but smaller when Parsed or Gzipped

> src/legacy-theme/ts/colors/index.t

hey @joshblack technically we don't need the compile step, we could just copy the output and put it in a static file. If those colors rarely change I think that could be a good idea.

I made it a compile step simply because I don't know how often they change and I thought it could be easier for a developer to edit the base files and then the compile step taking care of converting everything into a big object

Issue #3710, opened 2023-09-05
Update `@lit-labs/react` from 1.1.1 to 2.0.1


joshblack:
This is a follow-up to our revert here: https://github.com/primer/react/pull/3708

Unfortunately, we had to rollback updating `@lit-labs/react` from v2 to v1 due to how we bundle in ESM-only modules currently in `@primer/react`.

For background, `@primer/react` ships with support for both ESM and CommonJS. However, not all the dependencies we consume provide a CommonJS compatible output. Since we depend on these dependencies and need to continue supporting ESM and CommonJS, an interim solution was created that inlined a CommonJS-compatabile version of these dependencies only for the CommonJS entrypoint.

However, packages like `@lit-labs/react` make use of the `exports` field in their `package.json` file in order to properly coordinate entrypoints depending on the environment. Specifically, v2 of this package will use an entrypoint only for Node.js if that environment is detected and one for the client in order to avoid usage of `useLayoutEffect`.

This information is lost in the current transformation and we would need to find a way to restore it in order to support `@lit-labs/react` v2 while continuing to support CommonJS.

lesliecdubs:
Discussed at the PRC planning meeting and while we don't want this dep to get too out-of-date, it seems like there isn't an immediate need to take this on. Moving to unprioritized backlog for now, but please feel free to counter or add more context if/when we should reconsider.

PR #4346, opened 2024-03-04
simplify custom-element react bindings with util

### LINKED ISSUE #4347: Simplify custom-element react bindings with util
Between #4345, #4241 and #2484 it might be worthwhile simplifying the adoption of Custom Elements by creating a util which combines the various primitives: `styled`, `sx`, `createComponent`.

keithamus:
Closes #4347

### Changelog

This is an internal change to add a private helper.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

This is an internal change to add a private helper.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -87,7 +87,7 @@
     "@github/markdown-toolbar-element": "^2.1.0",
     "@github/paste-markdown": "^1.4.0",
     "@github/relative-time-element": "^4.1.2",
-    "@lit-labs/react": "^1.1.1",
+    "@lit-labs/react": "1.2.1",
     "@oddbird/popover-polyfill": "^0.3.1",
     "@primer/behaviors": "^1.5.1",
     "@primer/octicons-react": "^19.8.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -448,7 +448,7 @@
     "examples/app-router": {
       "name": "example-app-router",
       "dependencies": {
-        "@primer/react": "36.9.0",
+        "@primer/react": "36.10.0",
         "next": "^14.1.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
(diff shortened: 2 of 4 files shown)


packages/react/package.json
```diff
@@ -87,7 +87,7 @@
     "@github/markdown-toolbar-element": "^2.1.0",
     "@github/paste-markdown": "^1.4.0",
     "@github/relative-time-element": "^4.1.2",
-    "@lit-labs/react": "^1.1.1",
```
keithamus, line 90: Lit's react package has graduated from labs. We should use the latest version. From their [npm package readme](https://www.npmjs.com/package/@lit-labs/react):

> This package has graduated from labs! It is now available as @lit/react. This package is just a proxy that re-exports @lit/react. As a result, while it will no longer be updated, it will continue to work and get updates through the ^1.0.0 version range of @lit/react. This should reduce duplication of code while the ecosystem migrates their imports away from @lit-labs/react.

joshblack: @keithamus just a heads up, we ran into an issue when moving to newer versions of `@lit-labs/react` (I think v2+?) which is what graduated into `@lit/react` from what I can tell. Specifically, it was when the packages moved to ESM only.

I tried to capture the issues over in: https://github.com/primer/react/issues/3710. Mostly they're due to us having to bundle ESM-only deps as CommonJS since we still have to support that in the current major version. When we do that, the way they do conditional exports unfortunately no longer works 😞 

We have plans to move to ESM only in the next major AFAIK but couldn't think of something in the current major that would support this. Would love to know what you think

keithamus: I'll just bump it on the minor then.

joshblack (approved):

PR #4350, opened 2024-03-04
chore(deps): remove @react-aria/ssr package

joshblack:
Remove the `@react-aria/ssr` package from `@primer/react` as it is no longer used as of v36. This also removes the SSR docs as they are no longer relevant as of v36.

#### Removed

- Remove `@react-aria/ssr` from `package.json

### Rollout strategy

This change removes a dependency from the project that is not imported or used and should not impact the public API


joshblack:
bump @siddharthkp when you get a sec 👀

siddharthkp (approved):
Oh right! That's good news

joshblack:
Ooo great point @siddharthkp, PR opened: https://github.com/primer/design/pull/743

PR #4357, opened 2024-03-05
 chore(project): update lockfile for install times

joshblack:
Regenerate the `package-lock.json` file to try and help out with the increased timeouts related to `npm i` when using npm v10. This change appears to be resolving more consistently than previous attempts and hopefully restores our install times back to what they used to be 🤞 

#### Changed

- Regenerate `package-lock.json`
- Update `@size-limit/preset-big-lib` to match current version of `size-limit`
- Remove all instances of installing npm v9

### Rollout strategy

### Testing & Reviewing

- Pull down the PR and try it out locally by running:
  - `npm run clean:all`
  - `npm i`
- Let me know if you get a timeout error!
- Verify that tests and deploys still work as expected due to lockfile changes
- Very that there are no steps that install npm v9 in CI


broccolinisoup (approved):
No timeout 🎉

PR #4377, opened 2024-03-11
SelectPanel2: Add story for "create new item"

siddharthkp:
- Mockup: [Figma](https://www.figma.com/file/S3xgRPvRuFVCDGxPlIGxaP/SelectPanel-%26-Filtering?type=design&node-id=1406-130691&mode=design)

https://github.com/primer/react/assets/1863771/240f2ad7-0ba5-46f1-8c01-fd1c40c6ce23

### DIFF
--- a/packages/react/src/drafts/SelectPanel2/SelectPanel.examples.stories.tsx
+++ b/packages/react/src/drafts/SelectPanel2/SelectPanel.examples.stories.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import {SelectPanel} from './SelectPanel'
-import {ActionList, ActionMenu, Avatar, Box, Button, Text, Octicon, Flash} from '../../index'
+import {ActionList, ActionMenu, Avatar, Box, Button, Text, Octicon, Flash, FormControl, TextInput} from '../../index'
 import {Dialog} from '../../drafts'
 import {
   ArrowRightIcon,
@@ -12,6 +12,7 @@ import {
   GitPullRequestIcon,
   GitMergeIcon,
   GitPullRequestDraftIcon,
+  PlusCircleIcon,
 } from '@primer/octicons-react'
 import data from './mock-story-data'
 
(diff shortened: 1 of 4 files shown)


packages/react/src/drafts/SelectPanel2/SelectPanel.examples.stories.tsx
```diff
@@ -846,6 +847,219 @@ export const NestedSelection = () => {
...
+
+  /* 
+    Controlled state + Create new label Dialog 
+    We only have to do this until https://github.com/primer/react/pull/3840 is merged
+  */
+  const [panelOpen, setPanelOpen] = React.useState(false)
+  const [newLabelDialogOpen, setNewLabelDialogOpen] = React.useState(false)
+
+  const openCreateLabelDialog = () => {
+    setPanelOpen(false)
```
siddharthkp, line 909: Note: 

This is a bit unfortunate, you have to control the open state of SelectPanel because `SelectPanel` uses `<dialog>` (top layer) but `Dialog` uses portal not `<dialog>` so Dialog would render under the panel 🤦 

This would change once we pick up https://github.com/primer/react/pull/3840 again

packages/react/src/drafts/SelectPanel2/SelectPanel.examples.stories.tsx
```diff
@@ -846,6 +847,219 @@ export const NestedSelection = () => {
...
+            setNewLabelDialogOpen(false)
+            setPanelOpen(true)
+          }}
+        />
+      )}
+    </>
+  )
+}
+
+const CreateNewLabelDialog = ({
```
siddharthkp, line 1007: Note: The code here isn't ideal, I tried to make it a little more accessible in the story but the underlying component needs work 😢 

This isn't the focus of the PR/epic though, so I just added a `Flash` banner telling people not to copy this part of code 😅

joshblack (approved):

PR #4379, opened 2024-03-12
Bug fix: default Button disabled bg-color

langermank:
- Adds `background-color` for the default Button disabled state
- Extends disabled Storybook stories for better snapshot tests (will be more useful when [this PR](https://github.com/primer/react/pull/4352) merges)

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Button/__tests__/__snapshots__/Button.test.tsx.snap
```diff
@@ -871,6 +874,7 @@ exports[`Button respects the small size prop 1`] = `
 
 .c0:disabled [data-component=ButtonCounter] {
   color: inherit;
+  background-color: var(--button-default-bgColor-disabled,undefined);
```
jonrohan, line 877: This seems weird, but noticing the rest of the snapshot has similar `undefined` 🤷🏻

jonrohan (approved):

packages/react/src/Button/__tests__/__snapshots__/Button.test.tsx.snap
langermank: Oh.. that's not good. Investigating

langermank: Actually, this looks like a normal thing that happens. I tested it in Storybook and the value is there. Its not great but we're getting rid of snaps soon anyways. Thanks for flagging!

packages/react/src/Button/styles.ts
```diff
@@ -18,6 +18,7 @@ export const getVariantStyles = (variant: VariantType = 'default', theme?: Theme
       '&:disabled': {
         color: 'primer.fg.disabled',
         borderColor: `var(--button-default-borderColor-disabled, ${theme?.colors.btn.border})`,
+        backgroundColor: `var(--button-default-bgColor-disabled, ${theme?.colors.input.disabledBg})`,
```
mperrotti, line 21: Are we sure `theme.colors.input.disabledBg` is the right path for the fallback? It's rendering as `undefined` in the Jest snapshot tests.

mperrotti (approved):

packages/react/src/Button/__tests__/__snapshots__/Button.test.tsx.snap
mperrotti: I made a very similar comment. Ignore me if this is expected behavior.

packages/react/src/Button/styles.ts
langermank: It looks like the border color above is also undefined, so I assumed there's something up with the test

PR #4386, opened 2024-03-13
Add support for nested submenus to `ActionMenu`

iansan5653:
`ActionMenu` has [Primer designs and docs](https://primer.style/components/action-menu) for nesting, but has never actually supported it. Nested submenus should meet the following spec:

1. The anchor should match surrounding menu items but render a visual indicator of a submenu
2. When activated, the anchor menu item should open a submenu and keep the root menu open
3. Menus should be navigable by left/right arrow keys, opening submenus with right and closing them with left
4. The entire stack of menus should close when the user presses <kbd>Tab</kbd> or activates a menu item
5. Only the top menu should close when the user presses <kbd>Escape</kbd>
6. The anchor menu item should have `aria-haspopup` and `aria-expanded` (when expanded)

Fortunately meeting these requirements with the existing `ActionMenu` architecture is actually pretty straightforward, so I went ahead and just did it:

- A new component `ActionMenu.MenuItemAnchor` renders an `ActionList.Item` with a trailing visual as the menu's anchor
- Using context, `ActionMenu` internally determines if it is a submenu in order to close parent menus when necessary and to bind left/right navigation
- Everything else is pretty much taken care of by existing logic - ie, `AnchoredOverlay` already assigns the right ARIA attributes to the menu item anchor

The result is an intuitive submenu API that 'just works':

```jsx
<ActionMenu>
  <ActionMenu.Button>Edit</ActionMenu.Button>
  <ActionMenu.Overlay>
    <ActionList>
      <ActionList.Item>Copy</ActionList.Item>
      <ActionList.Item>Cut</ActionList.Item>
      <ActionList.Item>Paste</ActionList.Item>
      <ActionMenu>
        <ActionMenu.MenuItemAnchor>Paste special</ActionMenu.MenuItemAnchor>
        <ActionMenu.Overlay>
          <ActionList>
            <ActionList.Item>Paste plain text</ActionList.Item>
            <ActionList.Item>Paste formulas</ActionList.Item>
            <ActionList.Item>Paste with formatting</ActionList.Item>
            <ActionMenu>
              <ActionMenu.MenuItemAnchor>Paste from</ActionMenu.MenuItemAnchor>
              <ActionMenu.Overlay>
                <ActionList>
                  <ActionList.Item>Current clipboard</ActionList.Item>
                  <ActionList.Item>History</ActionList.Item>
                  <ActionList.Item>Another device</ActionList.Item>
                </ActionList>
              </ActionMenu.Overlay>
            </ActionMenu>
          </ActionList>
        </ActionMenu.Overlay>
      </ActionMenu>
    </ActionList>
  </ActionMenu.Overlay>
</ActionMenu>
```

The only thing I'm not so sure of is the naming of `ActionMenu.MenuItemAnchor`. It's not consistent with `ActionMenu.Button`, but I figured just calling it `ActionMenu.MenuItem` would be confusing. Open to suggestions here.

- Related: https://github.com/github/primer/issues/3120

#### New

- Adds support for nested submenus in `ActionMenu`

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/ActionMenu/ActionMenu.docs.json
+++ b/packages/react/src/ActionMenu/ActionMenu.docs.json
@@ -10,7 +10,7 @@
       "type": "React.ReactElement[]",
       "defaultValue": "",
       "required": true,
-      "description": "Recommended: `ActionMenu.Button` or `ActionMenu.Anchor` with `ActionMenu.Overlay`"
+      "description": "Recommended: `ActionMenu.Button`, `ActionMenu.MenuItemAnchor`, or `ActionMenu.Anchor`, with `ActionMenu.Overlay`"
     },
     {
       "name": "open",
@@ -46,6 +46,22 @@
       "passthrough": {
         "element": "Button",
         "url": "/react/Button"
+      }
+    },
+    {
+      "name": "ActionMenu.MenuItemAnchor",
+      "props": [
+        {
+          "name": "children",
+          "type": "React.ReactElement",
+          "defaultValue": "",
+          "required": true,
+          "description": ""
+        }
+      ],
+      "passthrough": {
+        "element": "ActionList.Item",
+        "url": "/react/ActionList"
       }
     },
     {
--- a/packages/react/src/ActionMenu/ActionMenu.tsx
+++ b/packages/react/src/ActionMenu/ActionMenu.tsx
@@ -194,7 +194,7 @@ type MenuOverlayProps = Partial<OverlayProps> &
 const Overlay: React.FC<React.PropsWithChildren<MenuOverlayProps>> = ({
   children,
   align = 'start',
-  side = 'outside-bottom',
+  side,
   'aria-labelledby': ariaLabelledby,
   ...overlayProps
 }) => {
@@ -222,7 +222,7 @@ const Overlay: React.FC<React.PropsWithChildren<MenuOverlayProps>> = ({
       onOpen={onOpen}
       onClose={onClose}
       align={align}
-      side={side}
+      side={side ?? (isSubmenu ? 'outside-right' : 'outside-bottom')}
       overlayProps={overlayProps}
       focusZoneSettings={{focusOutBehavior: 'wrap'}}
     >
--- a/packages/react/src/ActionMenu/ActionMenu.tsx
+++ b/packages/react/src/ActionMenu/ActionMenu.tsx
@@ -19,13 +19,16 @@ import type {ForwardRefComponent as PolymorphicForwardRefComponent} from '../uti
 import {Tooltip} from '../TooltipV2/Tooltip'
 import {ActionList, type ActionListItemProps} from '../ActionList'
 
-type MenuCloseHandler = (gesture: 'anchor-click' | 'click-outside' | 'escape' | 'tab' | 'item-select') => void
+export type MenuCloseHandler = (
+  gesture: 'anchor-click' | 'click-outside' | 'escape' | 'tab' | 'item-select' | 'arrow-left',
+) => void
 
 export type MenuContextProps = Pick<
   AnchoredOverlayProps,
   'anchorRef' | 'renderAnchor' | 'open' | 'onOpen' | 'anchorId'
 > & {
   onClose?: MenuCloseHandler
+  isSubmenu?: boolean
 }
 const MenuContext = React.createContext<MenuContextProps>({renderAnchor: null, open: false})
 
@@ -60,7 +63,7 @@ const Menu: React.FC<React.PropsWithChildren<ActionMenuProps>> = ({
     gesture => {
       setCombinedOpenState(false)
 
-      // Only close the parent stack when an item is selected or the user tabs out of the menu entirely
+      // Close the parent stack when an item is selected or the user tabs out of the menu entirely
       switch (gesture) {
         case 'tab':
         case 'item-select':
@@ -122,7 +125,18 @@ const Menu: React.FC<React.PropsWithChildren<ActionMenuProps>> = ({
   })
 
   return (
-    <MenuContext.Provider value={{anchorRef, renderAnchor, anchorId, open: combinedOpenState, onOpen, onClose}}>
+    <MenuContext.Provider
+      value={{
+        anchorRef,
+        renderAnchor,
+        anchorId,
+        open: combinedOpenState,
+        onOpen,
+        onClose,
+        // will be undefined for the outermost level, then false for the top menu, then true inside that
+        isSubmenu: parentMenuContext.isSubmenu !== undefined,
+      }}
+    >
       {contents}
     </MenuContext.Provider>
   )
@@ -186,13 +200,18 @@ const Overlay: React.FC<React.PropsWithChildren<MenuOverlayProps>> = ({
 }) => {
   // we typecast anchorRef as required instead of optional
   // because we know that we're setting it in context in Menu
-  const {anchorRef, renderAnchor, anchorId, open, onOpen, onClose} = React.useContext(MenuContext) as MandateProps<
-    MenuContextProps,
-    'anchorRef'
-  >
+  const {
+    anchorRef,
+    renderAnchor,
+    anchorId,
+    open,
+    onOpen,
+    onClose,
+    isSubmenu = false,
+  } = React.useContext(MenuContext) as MandateProps<MenuContextProps, 'anchorRef'>
 
   const containerRef = React.useRef<HTMLDivElement>(null)
-  useMenuKeyboardNavigation(open, onClose, containerRef, anchorRef)
+  useMenuKeyboardNavigation(open, onClose, containerRef, anchorRef, isSubmenu)
 
   return (
     <AnchoredOverlay
--- a/packages/react/src/ActionMenu/ActionMenu.tsx
+++ b/packages/react/src/ActionMenu/ActionMenu.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useContext} from 'react'
 import {TriangleDownIcon} from '@primer/octicons-react'
 import type {AnchoredOverlayProps} from '../AnchoredOverlay'
 import {AnchoredOverlay} from '../AnchoredOverlay'
@@ -13,11 +13,13 @@ import type {MandateProps} from '../utils/types'
 import type {ForwardRefComponent as PolymorphicForwardRefComponent} from '../utils/polymorphic'
 import {Tooltip} from '../TooltipV2/Tooltip'
 
+type MenuCloseHandler = (gesture: 'anchor-click' | 'click-outside' | 'escape' | 'tab' | 'item-select') => void
+
 export type MenuContextProps = Pick<
   AnchoredOverlayProps,
   'anchorRef' | 'renderAnchor' | 'open' | 'onOpen' | 'anchorId'
 > & {
-  onClose?: (gesture: 'anchor-click' | 'click-outside' | 'escape' | 'tab') => void
+  onClose?: MenuCloseHandler
 }
 const MenuContext = React.createContext<MenuContextProps>({renderAnchor: null, open: false})
 
@@ -44,9 +46,23 @@ const Menu: React.FC<React.PropsWithChildren<ActionMenuProps>> = ({
   onOpenChange,
   children,
 }: ActionMenuProps) => {
+  const parentMenuContext = useContext(MenuContext)
+
   const [combinedOpenState, setCombinedOpenState] = useProvidedStateOrCreate(open, onOpenChange, false)
   const onOpen = React.useCallback(() => setCombinedOpenState(true), [setCombinedOpenState])
-  const onClose = React.useCallback(() => setCombinedOpenState(false), [setCombinedOpenState])
+  const onClose: MenuCloseHandler = React.useCallback(
+    gesture => {
+      setCombinedOpenState(false)
+
+      // Only close the parent stack when an item is selected or the user tabs out of the menu entirely
+      switch (gesture) {
+        case 'tab':
+        case 'item-select':
+          parentMenuContext.onClose?.(gesture)
+      }
+    },
+    [setCombinedOpenState, parentMenuContext],
+  )
 
   const menuButtonChild = React.Children.toArray(children).find(
     child => React.isValidElement<ActionMenuButtonProps>(child) && (child.type === MenuButton || child.type === Anchor),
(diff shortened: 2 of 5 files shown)


broccolinisoup (commented):
Hey @iansan5653  👋🏻  Thanks so much for creating this PR 🙏🏻 

I am not sure if we have had any discussions around the API, please let me know if I miss any context on that but I wonder if we have explored other options before settling on creating a new sub components. I think the functionality works great, thanks so much again, I just wanted to make sure we are not overloading the component with sub components if we have other feasible alternatives.  Like using an existing anchor component maybe 🤷🏻‍♀️ 

```jsx
<ActionMenu.Anchor>
     <ActionList.Item>
         Paste special
         <ActionList.TrailingVisual>
              <ChevronRightIcon />
         </ActionList.TrailingVisual>
     </ActionList.Item>
  </ActionMenu.Anchor>
```

vs

```
<ActionMenu.MenuItemAnchor>Paste special</ActionMenu.MenuItemAnchor>
```

Especially since we are relying on context to determine if the menu is a submenu. This option is more manually constructing it but I wonder if it feels intuitive enough to folks without introducing more sub components to  ActionMenu. Or maybe introducing a new sub component reduces complexity under the hood, it is just food for thoughts to talk about the api options. 

Tagging @siddharthkp to get his thoughts around the API

Thanks again for pushing this PR, super super helpful! 🙌🏻

.changeset/wild-students-bow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@primer/react": feature
+"docs": patch
```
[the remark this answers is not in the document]
joshblack: Yeah! Thankfully we can typically ignore it since it's not a published package 👍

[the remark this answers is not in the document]
iansan5653:
```suggestion
```

Might as well delete it

broccolinisoup:
Hey @iansan5653 thanks for the reply!

> Although I do prefer to provide the trailing visual by default because I'm concerned we might end up with inconsistent iconography otherwise. It's important to have a consistent visual here so we can intuitively indicate (to sighted users) that a submenu is available. If some apps use KebabHorizontalIcon and some use ArrowRightIcon and some use TriangleRightIcon and finally others use ChevronRightIcon, it's going to be a poor experience.

That is true, I underestimated the trailing visual issue. 

>One option might be for Anchor to detect if its child is an ActionList.Item and clone it to add the icon. I really want to avoid this because it's fragile and generally not a good idea to manipulate children (in React or IRL 😄).

😄 Yeah that is not a good enough reason to manipulate the children, I agree.

>This could work - we could easily change ActionMenu.Button to return an ActionList.Item instead of a Button when it detects that it is a submenu anchor. 

This could definitely work and make it intuitive enough when we update the name to make it more generic as you mentioned like `ActionMenu.DefaultAnchor`. Although, since this is going to be a major change, it will add another item to the major upgrade list and I am now thinking its value. Would it worth the effort? I guess my question would be do we want to add another sub component and increase the API surface area or do we want to keep the API lean and increase the code complexity under the hood? Especially with the major release headache, I am leaning more towards what you initially built `ActionMenu.MenuItemAnchor` and I am glad we had this conversation. Thanks for following up with the feedback. Obviously, this is totally my personal opinion and I'd love to make sure at least one of my other peers seen the API change and provide if they have any feedback 🙌🏻  (I nudged Sid on the DM 🫠)

pksjce:
Hi @iansan5653 ! 
Thanks so much for working on this feature for us.
This helps me a lot with `ActionBar` where we have this requirement. 
I have created a PR branching off of this branch here. https://github.com/primer/react/pull/4421 . I would love if you had any feedback on this.
It seems to work pretty well except for two critiques
1. There is no way to provide `MenuItemAnchor` with a `LeadingVisual`. Using `ActionList.LeadingVisual` is a warning since its not under the right parent.
2. I agree that the naming of `MenuItemAnchor` could be better. Maybe something that includes `SubMenu` in it?

iansan5653:
> Using `ActionList.LeadingVisual` is a warning since its not under the right parent.

This will work in practice even if it's a lint error, because `ActionMenu.MenuItemAnchor` is rendering `children` directly into `ActionList.Item` without wrapping them with anything. I think it's just a matter of updating our linting configuration to indicate that it will work.

iansan5653:
I went ahead and pushed fb103fb21f7cd06b75e6cb547e617b4a34fb9571 which deletes `MenuItemAnchor` in favor of instead automatically assigning the keyboard shortcut and trailing visual. The logic works like this:

- If this is a submenu:
  - Open the menu when <kbd>→</kbd> is pressed (regardless of anchor type)
  - If the component inside `Anchor` is `ActionList.Item`:
    - Add the trailing icon to that component's `children`

This has some great benefits:

- It doesn't require any special consideration to wire up the keyboard shortcut properly
- The trailing icon will _usually_ work great
- The API surface is kept the same as before
- The API is intuitive: `<ActionMenu.Anchor><ActionList.Item>...</ActionList.Item></ActionMenu.Anchor>`
- There's no need to update linting logic to [allow for leading visuals](https://github.com/primer/react/pull/4386#issuecomment-2012407386)

The resulting consumer code looks like this:

<details><summary>Expand code block</summary>

```jsx
  <ActionMenu>
    <ActionMenu.Button>Edit</ActionMenu.Button>
    <ActionMenu.Overlay>
      <ActionList>
        <ActionList.Item>Cut</ActionList.Item>
        <ActionList.Item>Copy</ActionList.Item>
        <ActionList.Item>Paste</ActionList.Item>
        <ActionMenu>
          <ActionMenu.Anchor>
            <ActionList.Item>
              <ActionList.LeadingVisual>
                <SparkleFillIcon />
              </ActionList.LeadingVisual>
              Paste special
            </ActionList.Item>
          </ActionMenu.Anchor>
          <ActionMenu.Overlay>
            <ActionList>
              <ActionList.Item>Paste plain text</ActionList.Item>
              <ActionList.Item>Paste formulas</ActionList.Item>
              <ActionList.Item>Paste with formatting</ActionList.Item>
              <ActionMenu>
                <ActionMenu.Anchor>
                  <ActionList.Item>Paste from</ActionList.Item>
                </ActionMenu.Anchor>
                <ActionMenu.Overlay>
                  <ActionList>
                    <ActionList.Item>Current clipboard</ActionList.Item>
                    <ActionList.Item>History</ActionList.Item>
                    <ActionList.Item>Another device</ActionList.Item>
                  </ActionList>
                </ActionMenu.Overlay>
              </ActionMenu>
            </ActionList>
          </ActionMenu.Overlay>
        </ActionMenu>
      </ActionList>
    </ActionMenu.Overlay>
  </ActionMenu>
```

</details>

There are some downsides though: this does require cloning the anchor and manipulating its `children`, something we already [called out](https://github.com/primer/react/pull/4386#issuecomment-2011151396) as not great. However, it's worth noting that we are already cloning the anchor element anyway to assign the right ref and wire up the click handlers, so this isn't as big of a change as you might think.

The trailing icon logic is somewhat fragile: it will not work unless the `ActionList.Item` is a _direct_ descendant of `ActionMenu.Anchor`. Ie, this will not work: `<ActionMenu.Anchor><SomeWrappedMenuItem /></ActionMenu.Anchor>`. I think this is OK since the trailing icon is, while desirable, not _critical_ to the functionality of the component. The consumer can always add the icon themselves in these edge cases.

---

I think this path is pretty acceptable and leads to a nice API with a good balance of pros and cons, but I also understand the drawbacks may make us want to go a different route. I went ahead and pushed the commit as a proof of concept but I'm happy to remove it if we don't want to take this path.

iansan5653:
Update: found a way to keep the improved API without manipulating the `ActionList.Item`'s `children`, by using context to apply a default trailing visual to any wrapped `ActionList.Item` component. I think this is a great solution with few significant compromises.

.changeset/wild-students-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": feature
```
broccolinisoup, line 2: I don't know if changeset action understands `feature`s ([reference](https://github.com/changesets/action/blob/main/src/utils.ts#L8-L12)) - we might need to go with `minor`

packages/react/src/ActionMenu/ActionMenu.docs.json
```diff
@@ -10,7 +10,7 @@
       "type": "React.ReactElement[]",
       "defaultValue": "",
       "required": true,
-      "description": "Recommended: `ActionMenu.Button` or `ActionMenu.Anchor` with `ActionMenu.Overlay`"
+      "description": "Recommended: `ActionMenu.Button`, or `ActionMenu.Anchor`, with `ActionMenu.Overlay`"
```
broccolinisoup, line 13:
```suggestion
      "description": "Recommended: `ActionMenu.Button` or `ActionMenu.Anchor` with `ActionMenu.Overlay`"
```

Should we revert it to its original state? Do we need a comma before or? 🤔

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -100,15 +119,52 @@ const Menu: React.FC<React.PropsWithChildren<ActionMenuProps>> = ({
...
+    [children, isSubmenu, onOpen],
+  )
+
+  // Add right chevron icon to submenu anchors rendered using `ActionList.Item`
+  const parentActionListContext = useContext(ActionListContainerContext)
+  const thisActionListContext = useMemo(
+    () =>
+      isSubmenu ? {...parentActionListContext, defaultTrailingVisual: <ChevronRightIcon />} : parentActionListContext,
+    [isSubmenu, parentActionListContext],
+  )
```
broccolinisoup, line 157: I wonder if we need the full context here? 

Could we do something like?

```
<ActionListContainerContext.Provider value={isSubmenu ? {defaultTrailingVisual: <ChevronRightIcon />} : {}}>
      {React.cloneElement(children, {
        ...anchorProps,
        ref: anchorRef,
        onKeyDown: openSubmenuOnRightArrow,
      })}
    </ActionListContainerContext.Provider>
```

We should still memoise it and not inline but just as a pseudo code. Let me know if I underestimate any of the existing context though or missunderstand anything!

broccolinisoup (commented):
Hey @iansan5653 👋🏻 

I love the API 🤤 And I love that the default trailing visual comes out of the box 🎉 

+1 to using context so that it also works with custom components. I don't have any feedback around the API, it is great! I left some comments about the implementation. Looking forward to hearing what you think of them 🙌🏻 

Could we please also add a story/test case for using a custom element that wraps `ActionList.Item` as an anchor? 

Also, on narrow screen sizes, due to lack of real estate, the sub sub menu opens in the left side and I intuitively used right arrow key to close it but it didn't work because we only wired the left arrow key for closing if I am not mistaken. I attached a video. I don't know if we should handle that case as well or was that just me expected to have the right arrow key close the menu if it is opened on the left side. 

https://github.com/primer/react/assets/1446503/a7fc0f15-c9df-4a72-9208-b0f5133e82d8

.changeset/wild-students-bow.md
iansan5653: huh, this was generated by the changesets CLI. I can change it to minor

packages/react/src/ActionMenu/ActionMenu.docs.json
iansan5653: good catch :+1:

packages/react/src/ActionMenu/ActionMenu.tsx
iansan5653: Yeah, I think we do need the full context because this anchor is still a list item inside a parent list, and so it still needs the info from that list. We just want to override the trailing visual, not the entire context, and React won't merge them by default

iansan5653:
> Could we please also add a story/test case for using a custom element that wraps `ActionList.Item` as an anchor?

Will do 👍 

> Also, on narrow screen sizes, due to lack of real estate, the sub sub menu opens in the left side and I intuitively used right arrow key to close it but it didn't work because we only wired the left arrow key for closing if I am not mistaken. I attached a video. I don't know if we should handle that case as well or was that just me expected to have the right arrow key close the menu if it is opened on the left side.

I don't think there's a good solution for this unfortunately. The arrow key navigation pattern is prescribed by web accessibility standards and we have to stick with that because screen readers cannot 'see' which direction the submenu opens. At the same time, we can't force the menu to always open to the right because then it will overflow off the screen. So I think we have to accept that it will feel a little backwards in this case. 

This approach is consistent with operating system menus:


https://github.com/primer/react/assets/2294248/ce95f507-fced-433a-9142-4be9d8b38cb4

.changeset/wild-students-bow.md
iansan5653: Fixed in [6a00e67](https://github.com/primer/react/pull/4386/commits/6a00e67e8cdd3b072d88653a1307582f3896c924)

packages/react/src/ActionList/Item.tsx
```diff
@@ -74,14 +74,17 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       inlineDescription: [Description, props => props.variant !== 'block'],
     })
 
+    const {container, afterSelect, selectionAttribute, defaultTrailingVisual} =
+      React.useContext(ActionListContainerContext)
+    const trailingVisual = slots.trailingVisual ?? defaultTrailingVisual
```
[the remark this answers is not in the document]
iansan5653: Ooh, nice catch! I needed to wrap the default one in `TrailingVisual` for styling

screenshot showing leading and trailing visuals rendering with matching alignment and color

packages/react/src/ActionMenu/ActionMenu.features.stories.tsx
```diff
@@ -181,3 +190,45 @@ export const InactiveItems = () => (
...
+      <ActionList>
+        <ActionList.Item>Cut</ActionList.Item>
+        <ActionList.Item>Copy</ActionList.Item>
+        <ActionList.Item>Paste</ActionList.Item>
+        <ActionMenu>
+          <ActionMenu.Anchor>
+            <ActionList.Item>
+              <ActionList.LeadingVisual>
+                <SparkleFillIcon />
+              </ActionList.LeadingVisual>
```
pksjce, line 207: 🥳

pksjce (approved):

broccolinisoup:
>This approach is consistent with operating system menus:

Right!! That is great to know. Yeah I thought that it is not very straightforward to solve this but wanted to mention anyway. I was mainly worry about the sighted keyboard users but seems like this is a common learnt pattern, I guess we are good here as well 👍🏻 

> turns out this doesn't actually work with wrapped components unless they forward all props to the underlying component

Yeah this is what I was thinking as a test case. Like https://github.com/github/github/blob/master/ui/packages/copilot-chat/components/ChatReference.tsx#L90. I tried with a pseudo story and it works 👍🏻

broccolinisoup:
@iansan5653 I just realised that the tests are failing. Could you please make sure they are all green as well before going ahead with merge?

PR #4372, opened 2024-03-08
Adds FormControl's "Auto-Wiring" to ActionMenu Component

JeffreyKozik:
Followup to previous discussions: 
- https://github.com/github/github/pull/315892/files#r1515116984
- https://github.com/github/accessibility-audits/issues/6390#issuecomment-1984664269
- https://github.com/github/accessibility/issues/5801#issuecomment-1985966997
- https://github.slack.com/archives/CSGAVNZ19/p1709836721563139

#### Changed

Made use of the the new `useFormControlForwardedProps` hook (introduced by https://github.com/primer/react/pull/3632) to automatically wire the `ActionMenu` component to the accessibility and validation provided by the `FormControl` component it's nested within. If `ActionMenu` isn't nested within `FormControl` the hook has no effect.

https://github.com/primer/react/blob/f7dcae96362c038a0ce2d3b44a6db41bf1d480c1/packages/react/src/FormControl/_FormControlContext.tsx#L24-L39

Within a form the accessible name is now the text within the corresponding `FormControl.Label` tag
Screenshot as evidence for the statement above
Outside of a form the accessible name is still the text within the `ActionMenu.Button` itself
Screenshot as evidence for the statement above

I also updated Jest snapshots for the `Token` component (all it did was rearrange class names). I'm unsure why it did this since it seems unrelated to the changes I made, but I'm not too concerned about it since we seem to be moving away from snapshot testing: https://github.com/primer/react/blob/cb54f4298700d369e616aaefb2e7df4830231736/contributor-docs/testing.md#snapshots.

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #4389, opened 2024-03-13
Adds FormControl's "Auto-Wiring" to SelectPanel v2 Component

JeffreyKozik:
Followup to [this conversation in Slack](https://github.slack.com/archives/C01L618AEP9/p1710344884849579) and https://github.com/primer/react/pull/4372. Related to https://github.com/github/primer/issues/2396.

#### Changed

Made use of the the new useFormControlForwardedProps hook (introduced by https://github.com/primer/react/pull/3632) to automatically wire the SelectPanel v2 component to the accessibility and validation provided by the FormControl component it's nested within. If SelectPanel v2 isn't nested within FormControl the hook has no effect.

Within a form the accessible name is now the text within the corresponding FormControl.Label tag and the button itself
Screenshot of computed properties
Screenshot as storybook UI

Outside of a form the accessible name is still the text within the SelectPanel.Button itself
Screenshot of computed properties
Screenshot of storybook UI

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -315,7 +328,8 @@ const Panel: React.FC<SelectPanelProps> = ({
 }
 
 const SelectPanelButton = React.forwardRef<HTMLButtonElement, ButtonProps>((props, anchorRef) => {
-  return <Button ref={anchorRef} {...props} />
+  const inputProps = useFormControlForwardedProps(props)
```
JeffreyKozik, line 331: Followed the pattern suggested in this slack thread: https://github.slack.com/archives/CSGAVNZ19/p1709843761424779?thread_ts=1709836721.563139&cid=CSGAVNZ19

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -3,7 +3,20 @@ import {SearchIcon, XCircleFillIcon, XIcon, FilterRemoveIcon, AlertIcon, ArrowLe
...
+  Heading,
+  Box,
+  Tooltip,
+  TextInput,
+  Spinner,
+  Text,
+  Octicon,
+  Link,
+  Checkbox,
+  useFormControlForwardedProps,
```
JeffreyKozik, line 18: Using the hook recently created by this PR: https://github.com/primer/react/pull/3632. This hook is also being used in the `InlineAutocomplete` draft component: https://github.com/primer/react/blob/cb54f4298700d369e616aaefb2e7df4830231736/packages/react/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx#L118

siddharthkp (approved):
Looks good!

Merging to main is blocked right now (temporarily) for another release. I can merge this once it opens

JeffreyKozik:
> Looks good!
> 
> Merging to main is [blocked today for another release](https://github.slack.com/archives/C02NUUQ9C30/p1710343612526019). I can merge this tomorrow.

Sounds great, thank you!

adrianababakanian (approved):
🎉

iansan5653:
🤔 Overall the approach looks great, but I have concerns about the accessibility implications. If a button's label is set using `aria-label` or `aria-labelledby`, the text inside the button is totally overridden and becomes invisible to screen readers. 

This is problematic for several reasons:

- There may be additional context / instructions in the panel's placeholder value that's shown when no selection is active, and designers may rely on users being able to access this content 
- Once a selection is made, screen reader users will be unable to determine what it is (or even that anything is selected)
- Voice control users will see a button and think that its content is its label, but actually the label is something different (this may be more of an inherent problem with the way we style `SelectPanel` anchors as buttons instead of inputs)

---

A potential alternative: What if instead of setting `aria-labelledby`, we set the button's `aria-label` to something like `Label Text: Button Text`? This would be a bit of a hack; we'd have to do something roughly like `` `${document.getElementById(labelledBy).textContent}: ${buttonText}` ``.

However there are risks with this approach too: will users be confused by the accessible label of the element changing when the selected value changes? I don't know how this would affect VoiceOver users.

---

A third option might be to explore a different `role` altogether for the anchor of a `SelectPanel`. I'm not totally sure that `button` is actually ideal here. On one hand yes, it's an activatable button that opens a panel; but on the other hand it's also an input with a value. Native `<select>` elements get the `combobox` role; maybe we should explore that approach? Maybe that approach makes sense only in forms, like in this case?

I think it's worth digging in to this a bit further - maybe in accessibility office hours?

JeffreyKozik:
@iansan5653 thank you for the detailed and thoughtful review.

As we use the `FormControl.Label` to label other components within forms such as [radio groups](https://primer.style/components/radio-group), [text inputs](https://primer.style/components/text-input), and [select menus](https://primer.style/components/select), why is the labelling able to work in an accessible way for those components, but not for this one? Does it ultimately come down to those being `role=input` and this being `role=button`?

If that's the case would it either a) be appropriate to change to `role=input` for the entire SelectMenu or b) be appropriate to have a prop that allows developers to set to `role=input` for situations such as this one?

Alternatively, the 2nd option you propose seems acceptable as well.

I'm happy to go to a11y office hours to discuss this further if needed as well. Let me know the next one you're available for.

iansan5653:
> Does it ultimately come down to those being `role=input` and this being `role=button`?

Yep, pretty much. While both roles are widgets with accessible names and text content, screen readers intentionally ignore the text content of a button in favor of its accessible name. 99% of the time these two things are identical (because it's not really standard to have a button with an external label), so if a screen reader were to read out both it would be annoying and unnecessary the vast majority of the time .

> If that's the case would it either a) be appropriate to change to `role=input` for the entire SelectMenu or b) be appropriate to have a prop that allows developers to set to `role=input` for situations such as this one?

Unfortunately not. `role=input` implies much more than just 'this button acts similarly to an input'. Screen readers will expect an `input` role to act _exactly_ like a text input - that is, they will expect it to accept text content and not provide a list of options.

Native browser `<select>` elements get the role [`combobox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role), which is much a closer candidate. I _think_ the only thing `SelectPanel` is missing from the `combobox` pattern is handling alphanumeric keys when focus is on the anchor (this should probably open the panel and put those characters in the filter). Here's an example of an [accessible select-like custom `combobox`](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/).

It's honestly possible that we should consider always giving the `SelectPanel` anchor the role of `combobox`, or maybe automatically in certain situations. I really don't know for sure about that and I'm by no means an expert, so I'd definitely recommend talking to the accessibility team about that. I'd be happy to show up to any of their office hours - just let me know when you'll be there!

JeffreyKozik:
@iansan5653 that's interesting context. It seems like giving SelectPanel the `combobox` role makes a lot of sense here, although I'm not sure the implications of that for the other use cases of SelectPanel such as when it's outside of a form. @siddharthkp what do you think about giving the `SelectPanel` anchor the role of `combobox`?

TylerJDev:
I believe it was mentioned that `SelectPanel` couldn't be a combobox, even though it's a similar pattern, as it handles certain aspects differently than a true combobox would. 

[We do have something else in the works](https://github.com/github/primer/issues/3066) that could act as the middle ground between `SelectPanel` and `ActionMenu`, that's specifically meant to be a true "combobox". Since this is still in a proof of concept state, I think it'd be best to try a different route for now, and iterate on it later.

I'm wondering if we could keep the usage of `aria-labelledby`, but also reference the button itself. For example:

```html
<label id="label1">SelectPanel within FormControl</label>
<button id="label2" aria-labelledby="label1 label2">Choose a tag</button>
```

This means that the button references both the label and the contents within itself (i.e. "Choose a tag"), making the accessible name: "**SelectPanel within FormControl**, **Choose a tag**". This is using [multiple `aria-labelledby` id(s) to concatenate a label](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA16#example-3-a-label-from-multiple-sources).

We could also change the order so that the current value is announced first (e.g. `aria-labelledby="label2 label1"` = "**Choose a tag**, **SelectPanel within FormControl**"). If we went this route, this would have to be unique to `SelectPanel`, as other the other components that support `FormControl` linking now do not require additional ARIA.

I'd still recommend running this by the direct Accessibility team, as I may be unaware of any other implications, such as how someone using voice might interact with this control, and Ian's comments around potential confusion with the label. I don't mind following up with them too!

siddharthkp:
This is great @iansan5653, thanks for bringing this up! I totally missed it.

> Without selection: "SelectPanel within FormControl: Choose a tag"
> With selection: "SelectPanel within FormControl: v35.28.0"

This looks perfect! There is a really old discussion that might be helpful to read through: https://github.com/github/primer/discussions/696

> I'm wondering if we could keep the usage of aria-labelledby, but also reference the button itself. For example:
> 
> <label id="label1">SelectPanel within FormControl</label>
> <button id="label2" aria-labelledby="label1 label2">Choose a tag</button>

This would work nicely! 

@TylerJDev, Just checking, both of these information should be considered as labels right (not description)?

JeffreyKozik:
@TylerJDev @siddharthkp thanks for providing more details on this issue! I'll take a swing at this change later today, but if I get pulled into other issues and need you to help out here on Monday I'll let you know, thanks @siddharthkp!

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -330,18 +329,33 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+      }
+    }
+  }, [inputProps.id])
+
+  const selectPanelButtonId = `${inputProps.id}--select-panel-button`
+  if (inputProps.id && labelId) {
+    return (
       <Button
         ref={anchorRef}
+        aria-labelledby={`${labelId} ${selectPanelButtonId}`}
```
siddharthkp, line 373: @TylerJDev, this looks correct. However, it this is read as one long strong: "SelectPanel within FormControl Choose a tag".

screen reader preview with the text: SelectPanel within FormControl Choose a tag

Should we artificially try to add a comma in there? "SelectPanel within FormControl, Choose a tag"

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -337,7 +350,34 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+      <Button
+        ref={anchorRef}
+        aria-labelledby={`${labelId} ${selectPanelButtonId}`}
+        {...inputProps}
+        id={selectPanelButtonId}
+      />
+    )
+  } else {
+    return <Button ref={anchorRef} {...props} />
+  }
```
JeffreyKozik, line 380: The other implementation I tried was
```tsx
return (
    <>
      {inputProps.id && (
        <VisuallyHidden id={`${inputProps.id}--select-panel-button-children`}>{inputProps.children}</VisuallyHidden>
      )}
      <Button
        ref={anchorRef}
        {...inputProps}
        aria-labelledby={`${inputProps.id} ${inputProps.id}--select-panel-button-children`}
      />
    </>
  )
```
However I wasn't thrilled about using `VisuallyHidden` because it can have [unintended consequences](https://github.com/github/primer/issues/2865#issuecomment-1999910555), although it is used in the [discussion](https://github.com/primer/react/pull/4389) @siddharthkp referenced so it might not be problematic. 

The solution I ended up implementing is similar to what @TylerJDev suggested:
```html
<label id="label1">SelectPanel within FormControl</label>
<button id="label2" aria-labelledby="label1 label2">Choose a tag</button>
```

The only wrinkle is that I had to give the `Button` a new `id` (so it's no longer linked to the `for` tag in the `label`).

Is there a more elegant solution here, perhaps involving forwarding React refs?

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
TylerJDev: Great point! I do think we should add something here, I feel that the comma would be a good option. It should provide a slight pause for most screen readers, such as with other punctuation like (`:`, `;`, etc).

JeffreyKozik: Is there a good way to do this? If I do
```tsx
aria-labelledby={`${labelId}, ${selectPanelButtonId}`}
```
It tries to find an element with the `id` `${labelId},` and can't so then it only ends up using the `${selectPanelButtonId}` label. The only way I can do it is by adding `VisuallyHidden` text with a comma and adding that as a third label in between the other two like:
```tsx
aria-labelledby={`${labelId}, ${selectPanelButtonId}`}
```
const selectPanelButtonId = `${inputProps.id}--select-panel-button`
  const punctuationId = `${inputProps.id}--select-panel-button-punctuation`
  const punctuation = ','
  if (inputProps.id && labelId) {
    return (
      <>
        <VisuallyHidden id={punctuationId}>{punctuation}</VisuallyHidden>
        <Button
          ref={anchorRef}
          aria-labelledby={`${labelId} ${punctuationId} ${selectPanelButtonId}`}
          {...inputProps}
          id={selectPanelButtonId}
        />
      </>
    )
```

Comment (siddharthkp) [reply packages/react/src/drafts/SelectPanel2/SelectPanel.tsx:373]: I think that's unfortunately the only way to do it 😢

File: packages/react/src/drafts/SelectPanel2/SelectPanel.tsx (hunk 4 above)
Comment (JeffreyKozik) [reply packages/react/src/drafts/SelectPanel2/SelectPanel.tsx:380]: When I tried reusing the id that's generated by the FormControl for this purpose it ended up giving it the name: "SelectPanel within FormControl SelectPanel within FormControl" since the label overrides the inner text of the button when linking with the `for` attribute

File: packages/react/src/drafts/SelectPanel2/SelectPanel.tsx (hunk 3 above)
Comment (JeffreyKozik) [reply packages/react/src/drafts/SelectPanel2/SelectPanel.tsx:373]: Alright I pushed that change. It also doesn't do "a, b" it does "a , b" so the spacing of the comma is a little bit off.

Comment (TylerJDev) [reply packages/react/src/drafts/SelectPanel2/SelectPanel.tsx]: Is it possible to add the comma to the end of the first label? So if `A` is "SelectPanel within FormControl", we'd append the comma to the end of the label, rather than having 3 separate references for `aria-labelledby`? That'd way we'd have only `A` and `B` being referenced.

Comment (JeffreyKozik) [reply packages/react/src/drafts/SelectPanel2/SelectPanel.tsx]: That's a good idea, but then wouldn't show a comma in the UI for the label as well?

Comment (TylerJDev) [reply packages/react/src/drafts/SelectPanel2/SelectPanel.tsx]: We could have it visually hidden as well, for example: 

```
<Label>
  SelectPanel within FormControl<VisuallyHidden>,</VisuallyHidden>
</Label>
```

This should still announce for screen readers, and will not be visually present within the label. In addition, if we could only add the comma conditionally when there's more than one value in the `aria-labelledby`, that would be best, as there's no need for punctuation if there's only one label.

Comment (JeffreyKozik) [reply packages/react/src/drafts/SelectPanel2/SelectPanel.tsx]: I'm able to do this
```tsx
      label.innerHTML = label.innerHTML.trim()
      label.innerHTML += `
        <span style="
          position: absolute;
          width: 1px;
          height: 1px;
          padding: 0;
          margin: -1px;
          overflow: hidden;
          clip: rect(0, 0, 0, 0);
          white-space: nowrap;
          border-width: 0;"
        >,</span>
      `
```
Although for some reason it still puts a space in between the label text and the ","
image
image

There may be some method of removing this space, but either way it seems at the worst dangerous and the best very non-React to force set the innerHTML like that 🤔

TylerJDev: I don't think we need to remove the space - as long as the comma is at the end, and it's not affecting the visual look of the label, then the announcement should be fine with or without the space.

JeffreyKozik: Which approach do you think is better? 1) Manually setting the `innerHTML` or 2) having three values in `aria-labelledby` and one of them is `VisuallyHidden`? Personally I prefer the 2nd approach because the first approach seems like it could cause problems in edge cases when manually setting `innerHTML`.

JeffreyKozik:
@siddharthkp @TylerJDev what are the next steps here to get this change ready to merge?

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: Sorry, I've been busy with first responder duties. I need to check if this is safe/fine.

@TylerJDev can you validate this for me, please?

siddharthkp:
@JeffreyKozik Sorry, just need to verify the final detail here: https://github.com/primer/react/pull/4389#discussion_r1535723774

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
TylerJDev: Let me run this by the Accessibility team, I'm not 100% sure if there's any implications we'd have from referencing three values, but if not, it'd be preferable to use this rather than setting `innerHTML` directly.

JeffreyKozik: Thanks! If you could cc me when you ask them that would be great, I'm interested in what are a11y practices are in this scenario 😄

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
TylerJDev: > When I tried reusing the id that's generated by the FormControl for this purpose it ended up giving it the name: "SelectPanel within FormControl SelectPanel within FormControl" 

I'm curious about where it gave it the name twice? Could we not utilize the ID `FormControl` provides, instead of using `selectPanelButtonId`? I see the potential issue where the `for` _could_ duplicate the label text, but when I tested, it stated both properly 🤔

JeffreyKozik: I'll take a video and sent it over. When I was doing it it was duplicating the text

JeffreyKozik: Oh interesting. 

When it's this order it works
```tsx
aria-labelledby={`${labelId} ${inputProps.id}`}
```

When it's this order it duplicates "SelectPanel within FormControl`
```tsx
aria-labelledby={`${inputProps.id} ${labelId}`}
```

https://github.com/primer/react/assets/29404461/3a61cd2c-b8a4-465e-9bf7-149af9741db2

JeffreyKozik: Alright I went ahead and removed `selectPanelButtonId` in favor of `inputProps.id` to avoid and a11y issues.

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
JeffreyKozik: Are there any a11y issues with visually hiding punctuation like this and only using it as part of the aria-labelledby attribute?

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -337,7 +351,35 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+      }
+    }
+  }, [inputProps.id])
+
+  const punctuationId = `${inputProps.id}--select-panel-button-punctuation`
+  const punctuation = ','
+  if (inputProps.id && labelId) {
+    return (
+      <>
+        <VisuallyHidden id={punctuationId} aria-hidden="true">
```
JeffreyKozik, line 374: Made this `aria-hidden="true'` so screen readers aren't distracted by this punctuation. Instead, it's only used in the aria-labelledby attribute

JeffreyKozik:
I actually think there's a cleaner way to do this with aria-label instead of aria-labelledby I'll push the change on Monday

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
iansan5653: 🤔 I wonder if it would be more straightforward to just build the label ourselves:

```jsx
const labelText = document.getElementById(labelId)?.textContent ?? ""
const buttonText = document.getElementById(inputProps.id)?.textContent ?? ""
const label = `${labelText}${punctuation} ${buttonText}`

return <Button ref={anchorRef} aria-label={label} {...inputProps} />
```

iansan5653: Haha I see this is exactly what you did 😄

JeffreyKozik: Yeah ended up going with your suggestion from this comment: https://github.com/primer/react/pull/4389#issuecomment-1998145910 😄

JeffreyKozik:
@siddharthkp @TylerJDev if this approach looks acceptable to you I can merge this today

TylerJDev:
I see on the trigger button, there's an `aria-labelledby` attached that links directly to the `<span>` within. Since we're relying on `aria-label` instead of `aria-labelledby`, we'd need to remove `aria-labelledby` usage altogether here.

Example: 
```HTML
<button aria-labelledby=":r19:-label" aria-label="SelectPanel within FormControl, Choose a tag" id=":r19:">
 <span data-component="text" id=":r19:-label">Choose a tag</span>
</button>
```

This `aria-labelledby` takes precedence over `aria-label`, and makes the accessible name of the control the text contents of the span (i.e. "Choose a tag"). I don't believe this was an addition of this PR, but for this case in particular (forms context), we'd would need to remove the `aria-labelledby` in favor of the `aria-label`.

Looking at it closely, this doesn't seem related to `SelectPanel`, instead it appears this was a recent add to our [`Button` component](https://github.com/primer/react/blob/main/packages/react/src/Button/ButtonBase.tsx#L65). I'll follow up on this!

JeffreyKozik:
@TylerJDev thanks for taking a look! It looks like `aria-labelledby` was added to the `Button` component just last week
Image
I just pulled the most recent changes from `main` and now I'm able to reproduce what you're describing. I'll go ahead and remove `aria-labelledby` from the `Button` in this situation.

JeffreyKozik:
@TylerJDev I was able to fix this issue by leaving the `aria-label` and adding an `aria-labelledby` attribute that references the `Button` itself. Let me know if this looks appropriate 👍

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -337,7 +350,27 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+    if (label?.textContent) {
+      setLabelText(label.textContent)
+    }
+  }, [inputProps.id])
+
+  if (labelText && inputProps.children) {
+    return (
+      <Button
+        ref={anchorRef}
+        aria-label={`${labelText}, ${(anchorRef as MutableRefObject<HTMLButtonElement>).current.textContent}`}
```
TylerJDev, line 366: Could we flip this, so that the button text content is first?

```suggestion
        aria-label={`${(anchorRef as MutableRefObject<HTMLButtonElement>).current.textContent}, ${labelText}`}
```

This way, the contents of the button is announced first, and the label second. We'd want this mainly because the most important content is the button's text content, whereas the label itself would be secondary to that, after it has been read.

JeffreyKozik: Sure I can go ahead and flip that 👍

TylerJDev:
> I was able to fix this issue by leaving the aria-label and adding an aria-labelledby attribute that references the Button itself. Let me know if this looks appropriate 👍

Thanks for all the work on this 🙏. I'm getting some inconsistent behavior with one of the screen readers I tested this implementation with. Using NVDA + Chrome/FF, the label announces as expected, with both label and button text contents being included. With VoiceOver and Chrome specifically, only the button's text contents is announced. This seems like an issue with Chrome + VO specifically, and not an issue with this PR.

I'd say short-term, we could probably keep this attribute, as it provides a better experience now compared to if the `aria-labelledby` targets the `<span>` within. Long-term we'll need to figure out a way to handle `aria-labelledby` within the `<button>` component in cases where we utilize `aria-label`. Once that happens, we can remove the `aria-labelledby` here.

TylerJDev (approved):
Other than my suggestion, I think this looks good from an accessibility perspective. Let's get @siddharthkp to review this once more before we merge, but I don't have any additional concerns. Thanks a ton for all work here :grin:

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
JeffreyKozik: Changes are pushed

JeffreyKozik:
> With VoiceOver and Chrome specifically, only the button's text contents is announced. This seems like an issue with Chrome + VO specifically, and not an issue with this PR.

Interesting, I guess different screen reader + browser combinations handle an aria-labelledby referencing itself differently. It's good that it works in most scenarios though 👍 

> Long-term we'll need to figure out a way to handle `aria-labelledby` within the `<button>` component in cases where we utilize `aria-label`.

I agree. That definitely makes sense to handle at the `Button` level because it affects all `Button` usage not just in `SelectPanel`. Right now it basically sounds like you can't use an `aria-label` at all with the `Button` component since it'll just get overridden by the `aria-labelledby`.

JeffreyKozik:
@siddharthkp if this approach looks suitable to you, I think it's ready to merge 👍

JeffreyKozik:
Trying to create the integration tests and run them again with my updated code, however the workflow keeps failing: https://github.com/github/github/actions/workflows/primer-react-pr-test.yml so I'm not even able to run the integration tests. Not sure if this is related to what's ongoing with npm, or a separate issue

JeffreyKozik:
Integration tests passed 🎉 https://github.com/github/github/pull/322332

Only tests failing seem related to `github-licenses` which I believe is expected because it's an integration test PR

PR #4393, opened 2024-03-14
Update README.md

maximedegreve:
We wanted to make our template more discoverable to avoid that internal employees have to set up their own project and run into unexpected issues.

- [💬 Internal Slack thread](https://github.slack.com/archives/C02NUUQ9C30/p1710415630143979)

### DIFF
--- a/.changeset/ten-candles-melt.md
+++ b/.changeset/ten-candles-melt.md
@@ -1,0 +1,5 @@
+---
+ "@primer/react": patch
+ ---
+
+ Add our react template to the README.md
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ npm install @primer/react
 yarn add @primer/react
 ```
 
+## Template
+
+The fastest way to try or make a prototype with Primer React without setting up a new project is by using our [react template](https://github.com/primer/react-template).
+
 ## Roadmap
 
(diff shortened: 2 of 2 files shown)


.changeset/ten-candles-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+ "@primer/react": patch
```
broccolinisoup, line 2: We usually release the docs changes without a changeset but did you want to announce this in the release notes?

maximedegreve: That was my plan. I was getting a warning though because there wasn't a changeset?

README.md
```diff
@@ -38,6 +38,10 @@ npm install @primer/react
 yarn add @primer/react
 ```
 
+## Template
+
+The fastest way to try or make a prototype with Primer React without setting up a new project is by using our [react template](https://github.com/primer/react-template).
```
owenniblock, line 43: Minor wording suggestion, non-blocker and probably personal preference :D

```suggestion
The fastest way make a prototype or try Primer React without setting up a new project is by using our [react template](https://github.com/primer/react-template).
```

owenniblock (approved):

.changeset/ten-candles-melt.md
broccolinisoup: Yeah the CI job will fail if there is no changeset. You could either add a changeset like you did here or you can add `skip changeset` label.

PR #4395, opened 2024-03-15
ActionList.GroupHeading roll out improvements

broccolinisoup:
#### New

1. Adds a `@deprecated` TS notice to the `title` prop on `ActionList.Group` sub component.
2. Adds an invariant check for cases where the action list is rendered as menu or listbox but `as` prop is provided on the `ActionList.GroupHeading` component. This is not needed and in fact it could be misleading. Because action lists whose roles are menu or listbox, will render their group headings are divs, not actual headings. (Reefer to the api docs for this accessibility info for now.)

```jsx
<ActionList role= "menu">
  <ActionList.Group>
    <ActionList.GroupHeading as="h2"><ActionList.GroupHeading> //👈 `as` is not needed here and it is misleading. 
  </ActionList.Group>
</ActionList>
```

#### Changed

1. Update the `warning` to become error (invariant) for missing `as` prop on `<ActionList.GroupHeading>` for `list` role action lists. 

```jsx
<ActionList>
  <ActionList.Group>
    <ActionList.GroupHeading as="h2"><ActionList.GroupHeading> //  👈 `as` is necessary here to explicitly set the heading level
  </ActionList.Group>
</ActionList>
```

2. Update the `title` prop on `ActionList.GroupHeading` to `_internalBackwardCompatibleTitle,` because it is used for to distinguish the old api vs new api for backwards compatibility and we don't want consumers to mistake and use that instead of children. https://github.com/primer/react/pull/4395/files#r1533123778

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/Group.tsx
```diff
@@ -85,8 +85,10 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
     >
       <GroupContext.Provider value={{selectionVariant, groupHeadingId}}>
         {title && !slots.groupHeading ? (
-          <GroupHeading title={title} variant={variant} auxiliaryText={auxiliaryText} />
+          // Escape hatch: supports old API <ActionList.Group title="group title"> in a non breaking way
+          <GroupHeading variant={variant} auxiliaryText={auxiliaryText} unsafeTitle={title} />
```
broccolinisoup, line 89: To make sure that the invariant check doesn't go through on the old API so that it doesn't break any usecases in downstream repos, I kept the `ActionList.GroupHeading` sub component's API flexible.  Especially, to distinguish where the group heading is coming from i.e. old api vs new api. 

If it comes from old api (i.e. if it has title prop), I render the heading like

```
<GroupHeading variant={variant} auxiliaryText={auxiliaryText} title={title} />
```

if it uses the new api (if children exists), 

```
<GroupHeading variant={variant} auxiliaryText={auxiliaryText} >{title}</GroupHeading>
```

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -262,8 +262,8 @@ describe('ActionList', () => {
     expect(heading).toBeInTheDocument()
     expect(heading).toHaveTextContent('Group Heading')
   })
-  it('should throw a warning if ActionList.Group is used without as prop when no role is specified (for list role)', async () => {
-    const spy = jest.spyOn(console, 'warn').mockImplementationOnce(() => {})
+  it('should throw an error if ActionList.GroupHeading is used without an `as` prop when no role is specified (for list role)', async () => {
```
broccolinisoup, line 265: This test fails https://github.com/primer/react/actions/runs/8368326991/job/22912262706?pr=4395 and I am so confused why - it is throwing and error and I am expecting it in the test 🤷🏻‍♀️ Am I missing something obvious?

packages/react/src/ActionList/Group.tsx
siddharthkp: Optional name idea, instead of `unsafeTitle`: `_internalBackwardCompatibleTitle`

packages/react/src/ActionList/ActionList.test.tsx
siddharthkp: I think you have to explicitly expect it to throw instead of spy

example in primer: https://github.com/primer/react/blob/main/packages/react/src/__tests__/Autocomplete.test.tsx#L444
stack overflow: https://stackoverflow.com/a/46155381

packages/react/src/ActionList/ActionList.features.stories.tsx
```diff
@@ -73,7 +74,8 @@ export const WithVisualListHeading = () => (
       </ActionList.Item>
     </ActionList.Group>
 
-    <ActionList.Group title="Advanced">
+    <ActionList.Group>
+      <ActionList.GroupHeading as="h4">Advanced</ActionList.GroupHeading>
```
siddharthkp, line 78: Just checking, do we have a story that still uses the old API? Just to keep it around for regressions

siddharthkp (commented):
Looking good! 🎉 

Will approve once the tests are passing!

packages/react/src/ActionList/ActionList.features.stories.tsx
broccolinisoup: Yes! I added them as dev stories https://github.com/primer/react/blob/main/packages/react/src/ActionList/ActionList.dev.stories.tsx and playwright snapshots them too to catch any regression 😈

packages/react/src/ActionList/ActionList.test.tsx
broccolinisoup: Ahh, right! it did the trick, thanks so much 🙌🏻

siddharthkp (approved):
Looking great! Ship it!

PR #4396, opened 2024-03-15
experimental/SelectPanel: Update anchored position after showModal

siddharthkp:
### The problem: 

When an html `<dialog>` is not "open", it has the dimensions of 0px by 0px, which means the anchored position is not able to calculate if the dialog is going beyond the window boundary.

https://github.com/primer/react/assets/1863771/df8c4137-b88d-41ad-a2da-a481b22f340f

### Solution

We need to update position after `dialog.showModal()` has "opened" the dialog and given it dimensions. This, however, creates a new problem of visible layout shift which can look janky.

In Safari, you see the dialog move but it's done before you can tell what happened:

https://github.com/primer/react/assets/1863771/a3e10383-b3a4-4582-8fb4-475a160260b3

In Chrome, you can really see the page glitching because for a tiny moment, Chrome tries to horizontally scroll the page to make sure the dialog is visible:

https://github.com/primer/react/assets/1863771/735fa254-45fe-4cc3-81fc-4428dee9f56c

###  Polish

We can solve this by adding a minuscule delay to the animating-in, this gives us just enough time to set the correct position (one tick)

https://github.com/primer/react/assets/1863771/03fd228a-7054-4201-8369-7074b04f7b32


packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -232,6 +241,10 @@ const Panel: React.FC<SelectPanelProps> = ({
         height="fit-content"
         maxHeight={maxHeight}
         sx={{
+          // to avoid a visible position shift, we delay the overlay animating-in
+          // to wait until the correct position is set (see useAnchoredPosition above for more)
+          animationDelay: '16ms',
```
siddharthkp, line 246: Note for reviewer: This isn't scientific at all, but we need to pick a number, so based it off a 60Hz refresh rate 😅 

From https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame, 

> The frequency of calls to the callback function will generally match the display refresh rate. The most common refresh rate is 60hz, (60 cycles/frames per second), though 75hz, 120hz, and 144hz are also widely used. requestAnimationFrame() calls are paused in most browsers when running in background tabs or hidden iframes, in order to improve performance and battery life.

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -164,11 +164,31 @@ const Panel: React.FC<SelectPanelProps> = ({
   /* Dialog */
   const dialogRef = React.useRef<HTMLDialogElement>(null)
 
+  /* Anchored position */
```
siddharthkp, line 167: Note for reviewer: This function already existed, just moved it above

packages/react/src/Overlay/Overlay.tsx
```diff
@@ -71,7 +71,9 @@ export const StyledOverlay = styled.div<StyledOverlayProps>`
   border-radius: 12px;
   overflow: ${props => (props.overflow ? props.overflow : 'hidden')};
   animation: overlay-appear ${animationDuration}ms ${get('animation.easeOutCubic')};
+  animation-fill-mode: forwards;
 
+  opacity: 0;
```
siddharthkp, line 76: Note for reviewer: Need to add this if we want to delay animation

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
siddharthkp: An alternative approach would be to "watch" for dialog open. Either in a recursive timeout or intersection observer, but that felt overkill!

broccolinisoup (approved):
Same problem as we had on [tooltip](https://github.com/primer/react/pull/4327)!  I think delaying it with an animation makes sense to me.

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -164,11 +164,31 @@ const Panel: React.FC<SelectPanelProps> = ({
...
+
   // sync dialog open state (imperative) with internal component state
   React.useEffect(() => {
-    if (internalOpen) dialogRef.current?.showModal()
-    else if (dialogRef.current?.open) dialogRef.current.close()
-  }, [internalOpen])
+    if (internalOpen) {
+      dialogRef.current?.showModal()
+      updatePosition() // update the position once the dialog is open
+    } else if (dialogRef.current?.open) dialogRef.current.close()
```
keithamus, line 190: This could skip the check

```suggestion
    } else dialogRef.current?.close()
```

siddharthkp:
Note for self: [This change](https://github.com/primer/react/pull/4396/files#r1526539341) seems to break tests (both in this repo and dotcom) where we instantly check if Overlay contents are visible (while they are animating)

Need to reduce the surface area of this change just to SelectPanel

PR #4415, opened 2024-03-19
ci(release): update concurrency group to no longer cancel wip

joshblack:
This is a slight change to our canary release workflow to no longer cancel in-progress workflows in the same concurrency group (in this case `npm-canary`). The workflow will cancel changes to itself if the PR is updated.

This will help when we have multiple PRs (or merge groups) attempting to publish a canary version at the same time.

#### Changed

- Add concurrency group for release canary workflow to cancel updates to the same PR
- Update concurrency options for publishing to no longer cancel in flight publish steps

### Rollout strategy

### DIFF
--- a/.github/workflows/release_canary.yml
+++ b/.github/workflows/release_canary.yml
@@ -6,6 +6,10 @@ on:
       - 'next-major'
       - 'changeset-release/**'
 
+concurrency:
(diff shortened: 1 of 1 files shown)


camertron (approved):
Wow, TIL about the `concurrency: group:` option.

PR #4418, opened 2024-03-20
Docs: Add stories from features as fallback

siddharthkp:
The new docs have a dropdown for examples: 

[See live ↗](https://primer.style/components/tree-view/react/beta)

docs showing a dropdown for examples underneath examples

#### Problem 

These are generated from the [`stories` array in docs.json](https://github.com/primer/react/blob/main/packages/react/src/TreeView/TreeView.docs.json#L6)

Unfortunately, only 1 component has this field populated. The rest of the components only show the default example and no dropdown:

docs with the examples dropdown missing

#### Problem 

Instead of filling this field for all the components manually, I added a fallback to use "feature" stories when stories are not filled in.

Pros:
- It's better than showing only the default stories
- Don't have to manually update this field when a story is added, removed or renamed

Cons:
- Adds all the features stories, no matter how repetitive
- Does not automatically add example stories

If you want, you can still manually update the `stories` array in docs.json and use that instead of the default! I think this is a good solution for now instead of showing only the default story.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [x] None: Docs only release

### Merge checklist


joshblack (approved):
Love it! Makes sense to me 👍

PR #4423, opened 2024-03-21
[Draft] Stack component

langermank:
Add support for the experimental `Stack` component to Primer React.

#### New

- Add support for the experimental `Stack` component to the drafts and experimental entrypoint
- Add a new test helper, `matrix`, to help with generating combinations of controls for e2e tests
- Add a new utility, `getResponsiveAttributes`, that helps with generating the correct data attributes for props that support values and responsive values

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

- [ ] Go through the storybook stories, verify that
  - [ ] A default story exists
  - [ ] A playground story exists
  - [ ] Relevant feature stories exist
  - [ ] Controls correctly update the playgrounds tory
- [ ] Look at the API of `Stack`, does it make sense to you? Anything that you would edit or change before we release it?
- [ ] What do you think of our CSS strategy here? Specifically around our use of data attributes and our responsive styling strategy
- [ ] What do you think of our matrix testing style in Playwright? Is it over the top? Are there better ways to do VRT for these scenarios?
- [ ] Verify the docs.json file for this component is valid


packages/react/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,544 @@
...
+  /**
+   * Specify the gap between children elements in the stack
+   */
+  gap?: Gap
+
+  /**
+   * Specify the orientation for the stack container
+   * @default vertical
+   */
+  orientation?: Orientation
```
siddharthkp, line 388: should we call this `direction` to match what you'd expect with in css?

packages/react/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,544 @@
...
+   * Sets whether items are forced onto one line or can wrap onto multiple lines
+   * @default nowrap
+   */
+  wrap?: Wrap
+
+  /**
+   * Sets how items will be distributed in the stacking direction
+   * @default start
+   */
+  spread?: Spread
```
siddharthkp, line 406: Should we call this `justify` instead of spread to match what you'd expect it to be with css?

packages/react/src/Stack/Stack.tsx
```diff
@@ -0,0 +1,544 @@
...
+
+  /**
+   * Sets the padding of the stack container
+   * @default none
+   */
+  padding?: Padding
```
siddharthkp, line 412: Reminder to update the eslint rule as well: https://github.com/primer/eslint-plugin-primer-react/blob/main/src/rules/no-system-props.js#L16

packages/react/src/experimental/index.ts
```diff
@@ -1,3 +1,6 @@
 'use client'
 
 export * from '../drafts'
+
+export {Stack, StackItem} from '../Stack'
+export type {StackProps, StackItemProps} from '../Stack'
```
siddharthkp, line 6: These should probably be in drafts because we export everything from drafts to experimental (but not the other way round)

Just until next major (when we can get rid of "drafts")

siddharthkp (commented):
I'm excited about this!

langermank:
The PR was updated with Copilot Workspace (v0.17). For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/primer/react/pull/4423?shareId=81334511-055c-4d8d-8ab3-5f711a7ea355).

langermank:
The PR was updated with Copilot Workspace (v0.17). For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/primer/react/pull/4423?shareId=fbfe9bf4-fcf3-406c-be5f-98c7c15719a9).

joshblack:
One thing I wanted to confirm, if we switch over to the CSS name for things do we also want the types to be updated? For example:

- For `orientation` to `direction`, would we also want `'horizontal' | 'vertical'` to go to `'column' | 'row'`
- For `spread` to `justify` (or maybe `justifyContent`?), would we also want to update these values to match the values for the corresponding CSS property?

joshblack (approved):
🚀

ericwbailey (approved):
🚢

PR #4424, opened 2024-03-22
ActionBar: Add stories for couple of edge cases

pksjce:
Closes #https://github.com/github/primer/issues/3133 and #https://github.com/github/primer/issues/3134

### Changelog

Adding two stories for issues that were a problem with the webcomponent in the rails version of this component.

### DIFF
--- a/.changeset/sweet-icons-stare.md
+++ b/.changeset/sweet-icons-stare.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+ActionBar: The overflow menu was earlier bootlegged with heavily customised ActionList. This is being replaced with ActionMenu which is cleaner and more robust.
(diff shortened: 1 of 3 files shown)


pksjce:
Thanks for the review! I should not have made the assumption that draft components don't need changesets. 
But we can make larger refactors and API change in this component without needing a `major` right? Just a `patch` changeset to document the change to the component should be enough?

siddharthkp:
> I should not have made the assumption that draft components don't need changesets.

I see! If it's not used anywhere yet, then I'd say it's nice to have a changeset but optional for now.

> But we can make larger refactors and API change in this component without needing a major right? Just a patch changeset to document the change to the component should be enough?

patch/minor sounds good in this case :)

siddharthkp (approved):
Ship it!

PR #4425, opened 2024-03-22
Dialog v2: Add support for returnFocusRef

siddharthkp:
In some rare cases, you need to return focus to a different instance of a button element than the one that triggered the Dialog (Dialog v1 supports this) 

One example of this is when you want to hide the trigger element temporarily to do some asynchronous work. Added an example in story. 

[See conversation on slack](https://github.slack.com/archives/C01L618AEP9/p1710947439592729) | [See affected code in gh/gh](https://github.com/github/github/blob/9b2f41812e3096df777bdd14105d09eb8117a7ef/ui/packages/issue-create/CreateIssueButton.tsx#L34-L51)

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


keithamus (commented):
We might want to consider what this looks like when we migrate to native dialog, and delete the focustrap code. At that point the effective way to alter focus would be to listen for the close/cancel events and call focus on a new element. Given that's the case, and given this seems to be an edge case, why don't we document that as a pattern, rather than implementing this feature?

siddharthkp:
> We might want to consider what this looks like when we migrate to native dialog, and delete the focustrap code. At that point the effective way to alter focus would be to listen for the close/cancel events and call focus on a new element. Given that's the case, and given this seems to be an edge case, why don't we document that as a pattern, rather than implementing this feature?

@keithamus I'd like to treat native vs synthetic dialog as an implementation. The feature here is to return focus to an element different than the one that opened the dialog, which would still be relevant (and in my opinion) a feature we can support with our Dialog wrapper over the native dialog element

keithamus:
I guess I'm just curious about why we're enabling explicit support for rare cases? Does having a ref vs using a callback sufficiently reduce complexity for upstream users?

siddharthkp:
> Does having a ref vs using a callback sufficiently reduce complexity for upstream users?

I think so.

If this wasn't already a feature in Dialog v1, I would have resisted it. But because it helps in transitioning from Dialog v1 to Dialog v2 seamlessly, I'm more inclined to accept this feature request. (Dialog v2 is the one we are making accessible)

.changeset/young-countries-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+"@primer/react": patch
```
pksjce, line 2: Just noting that its marked as `minor` on the description

siddharthkp: Ah, yes you're right! Updated

pksjce (approved):

Issue #3991, opened 2023-11-30
Apply 'disabled' styling when `aria-disabled` is set on a `Button` or `IconButton`


iansan5653:
In many cases it's not desirable to fully disable a button control because it can make that button difficult for some users to locate. A disabled button is not focusable and doesn't respond to interaction at all, and it doesn't have any indication of why it's disabled. 

While it's generally preferable to just avoid disabling buttons and rely on alternative methods instead, in _some_ cases it may be preferable to use `aria-disabled` along with CSS styling to disable a button. This is particularly useful in tandem with a tooltip that appears on focus (this tooltip would be inaccessible on a `disabled` button).

As a component library we should make accessible practices as easy as possible. So I think we shouldn't force consumers to build custom CSS styles in order to use this approach - we should make using `aria-disabled` as easy as `disabled`. With that in mind, I propose that `aria-disabled` should, by default, have the same styling as `disabled` buttons. 

I've focused on buttons here but this actually probably applies to all interactive components that can be disabled.

mperrotti:
@lesliecdubs - if the group agrees with this change, it would fit great in this PR: https://github.com/primer/react/pull/3582

If you prefer we do it in a separate PR, I'm cool with that too.

broccolinisoup:
I think we should re-open this because the button loading state PR was reverted and it is being re-introduced here https://github.com/primer/react/pull/4485 (It is referenced in https://github.com/primer/react/pull/4426 )

broccolinisoup:
@mperrotti I'm assigning this issue to you since you have done the work already in https://github.com/primer/react/pull/4485 but feel free to un-assign yourself if there is anything I misunderstand here.

lesliecdubs:
Should this issue be closed now? I see #4485 was merged.

langermank:
This has been fixed with CSS Modules work on Button 👍

PR #4426, opened 2024-03-22
Allow user to override `Button` and `IconButton` `aria-disabled`

iansan5653:
Closes https://github.com/primer/react/issues/3991

It looks like in https://github.com/primer/react/pull/3582 we added support for `aria-disabled` styling on button components, which is awesome! Except that we accidentally introduced a regression where consumers cannot set the `aria-disabled` attribute on these components. This is because we are setting `aria-disabled` in `ButtonBase` _after_ we spread the consumer's props.

The fix here is simply to move this line earlier, so the consumer can override the attribute.

Alternatively, we could instead force `aria-disabled` to be `true` when `loading`, regardless of what the consumer sets, but then respect the consumer's prop all other times. That's not a hard change to make if we prefer to go that route.

#### Changed

- Fixes a bug where consumers cannot override buttons' `aria-disabled` attribute

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/smart-cameras-sip.md
+++ b/.changeset/smart-cameras-sip.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Fixes a bug where consumers cannot override buttons' aria-disabled attribute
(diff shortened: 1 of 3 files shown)


langermank (approved):
Awesome, thank you!

PR #4446, opened 2024-03-28
Tooltip V2: Update the threshold value to match with the offset

broccolinisoup:
Reported in the communities (https://github.com/orgs/community/discussions/113546) and also internally that the elements that have tooltips on are not fully clickable. The reported issue is on PVC tooltip but React doesn't have the right value either. We think this is happening because the threshold (12px) is higher than the offset value (4px) and tooltip ends up blocking out the bottom part of its trigger. 

Corresponding PVC PR: https://github.com/primer/view_components/pull/2733

#### Changed

Updated the height of the after pseudo element (Between tooltip and its trigger) to match with the offset value ([default](https://github.com/primer/behaviors/blob/main/src/anchored-position.ts#L250)) so that tooltip don't overlap with its trigger.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/TooltipV2/Tooltip.tsx
+++ b/packages/react/src/TooltipV2/Tooltip.tsx
@@ -62,7 +62,7 @@ const StyledTooltip = styled.span`
     display: block;
     right: 0;
     left: 0;
-    height: 8px;
+    height: var(--overlay-offset, 0.25rem);
     content: '';
   }
 
--- a/.changeset/old-shirts-train.md
+++ b/.changeset/old-shirts-train.md
@@ -1,0 +1,6 @@
+---
+"@primer/react": patch
+---
+
+Tooltipv2: Update the threshold value to match with the offset
+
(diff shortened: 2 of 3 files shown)


packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -62,7 +62,7 @@ const StyledTooltip = styled.span`
     display: block;
     right: 0;
     left: 0;
-    height: 8px;
+    height: var(--overlay-offset, 0.25rem);
```
broccolinisoup, line 65: @langermank should this be 4px or 0.25rem?

keithamus (approved):

siddharthkp (approved):
Very nice!

packages/react/src/TooltipV2/Tooltip.tsx
langermank: What you have is perfect!

PR #4448, opened 2024-03-28
Fix types of useAnchoredPosition to avoid type assertions.

### LINKED ISSUE #4447: The `useAnchoredPosition` hook returns generic refs for Element, which always requires a type assertion by the user
### Description

The useAnchoredPosition hook will create refs if they are not passed in. However, when it does so, it creates a generic `React.RefObject<Element>` ref and returns that. Therefore, when it is used on any JSX node, it must be accompanied by a type assertion, e.g. `<div ref={floatingElementRef as React.RefObject<HTMLDivElement>`.

This can be fixed by doing 2 things:

1. Update the `useAnchoredPosition` hook to take optional type parameters, `TAnchorElement` and `TFloatingElement`. These type parameters get passed to `useProvidedRefOrCreate` in order to get back the correctly-typed ref object, and in turn, returned by the `useAnchoredPosition` hook.
2. Default the new type parameters to HTMLDivElement. `div` is likely the most common element type for the floating element, and possibly for anchor as well (the other possibility for anchor would be `button`).

This will allow consumers to use the refs returned by the hook without adding on a type assertion. See other codebase changes as an example.

It's possible this is a very minor breaking change: If a consumer is using a type assertion to fit a ref to an element other than a div, this change may break them. The downstream fix is to 1) remove the type assertion, and 2) pass the correct type param into `useAnchoredPosition`.

### Steps to reproduce

```tsx
function MyComponent() {
  const {anchorElementRef, floatingElementRef, position} = useAnchoredPosition();
  return (
    <div>
        { /* the type assertions below should not be necessary */ }
        <div ref={anchorElementRef as React.RefObject<HTMLDivElement>}>The anchor</div>
        <div ref={floatingElementRef as React.RefObject<HTMLDivElement>}>The floating element</div>
    </div>
  );
}
```

### Version

latest

### Browser

_No response_

T-Hugs:
Closes #4447 

#### Changed

`useAnchoredPosition` types updated to remove user type assertions

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Autocomplete/AutocompleteOverlay.tsx
+++ b/packages/react/src/Autocomplete/AutocompleteOverlay.tsx
@@ -57,7 +57,7 @@ function AutocompleteOverlay({
       preventFocusOnOpen={true}
       onClickOutside={closeOptionList}
       onEscape={closeOptionList}
-      ref={floatingElementRef as React.RefObject<HTMLDivElement>}
+      ref={floatingElementRef}
       top={position?.top}
       left={position?.left}
       sx={{
--- a/packages/react/src/hooks/useAnchoredPosition.ts
+++ b/packages/react/src/hooks/useAnchoredPosition.ts
@@ -1,8 +1,9 @@
-import React, {useLayoutEffect} from 'react'
+import React from 'react'
 import {getAnchoredPosition} from '@primer/behaviors'
 import type {AnchorPosition, PositionSettings} from '@primer/behaviors'
 import {useProvidedRefOrCreate} from './useProvidedRefOrCreate'
 import {useResizeObserver} from './useResizeObserver'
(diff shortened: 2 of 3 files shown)


broccolinisoup (commented):
Hello @T-Hugs 👋🏻 Sorry for late review! Thanks so much for pushing this PR 🙌🏻 Left one question - and would recommend [running integration tests at dotcom](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md) to make sure we are not introducing anything breaking. Let me know if you have any questions or concerns 🙌🏻

PR #4469, opened 2024-04-04
fix(Portal): blur activeElement children when unmounting

joshblack:
Context https://github.com/github/issues/issues/10142

Update our `Portal` behavior to check for an activeElement and, if it has one, blur the element. This seems to help address a timing issue where we are removing a portal and focusing an element like in the linked issue.

Specifically, in Safari it seems that even though we restore focus that it doesn't restore the page position without going through this.

#### Changed

- Update `Portal` to blur activeElement children when removing an element

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- [ ] Visit the default SelectPanel storybook
- [ ] Add enough content on the page to cause it to scroll
- [ ] Observe that dismissing the SelectPanel with escape does not cause the page to scroll


siddharthkp (approved):
Good digging! 

This _feels_ like a hack but a very nicely self-contained hack, so happy to ship it if it fixes a bug for us

Let's test this with dotcom to make sure there are no regressions (maybe a good candidate to discuss staff shipping)

joshblack:
Following up, this fails downstream tests suite typically with an `act()` violation error 🤔 So it may be that this change is causing effects to happen where they do not typically in these downstream tests and they need `act()` to remedy it.

PR #4471, opened 2024-04-05
Tooltip2: Prevent closing other overlays when tooltip has the focus and ESC is hit

broccolinisoup:
Due to handling escape events in tooltip via browser (popover api), this causes an issue where tooltip is used on an element on an overlay, hitting esc key closes all overlays. Expected behaviour is that, when tooltip has the focus, hitting escape only close the tooltip.

Video previews describe the current behaviours as before and behaviour after the fix as after. 

## Before

https://github.com/primer/react/assets/1446503/3a5b4e25-953f-44ef-a78c-f880781ab283

## After

https://github.com/primer/react/assets/1446503/9d882828-1e97-4daf-94dd-b0b37d422195

#### New

- Added the escape hook to take the responsibility of handling the escape events in the component rather than leaving it to the browser so that we can intercept with it and prevent escape event bubbling up.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp (approved):
Makes sense to me!

It's a little unfortunate but only because we have both portals and popovers together. I think we'll be able to remove this code once all Overlays are also dialog or popover 👍

PR #4476, opened 2024-04-08
Add fallback from `dist` folder

lukasoppermann:
This is needed so we can remove the remove the `tokens-next-private` folder.
It does not have any effect on the output.

We need to cut this release https://github.com/primer/primitives/pull/855 before we can merge.


langermank:
Not needed anymore since this code was removed 👍

PR #4459, opened 2024-04-03
Button: only add fallback aria-labelledby if aria-label isn't present

siddharthkp:
- Found this bug in https://github.com/primer/react/pull/3582 in integration tests for release candidate: https://github.com/github/github/pulls?q=is:pr+is:open+head:primer/react-integration

### Changelog

In https://github.com/primer/react/pull/3582, we introduced an `aria-labelledby` for all instances of Button. This however isn't a good idea when `aria-label` is already present on the instance.

**aria-labelledby takes precedence over aria-label, so we only want to add a fallback aria-labelledby when aria-label isn't present**

Skipping changeset because this is a bug fix for an unreleased PR

### Rollout strategy

- [x] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Testing & Reviewing

Tested with gh/gh by making the changes locally

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #4485, opened 2024-04-10
Un-revert "Add `loading` prop for `Button` and `IconButton` (#3582)"

mperrotti:
Brings back https://github.com/primer/react/pull/3582

----

> Copied from #3582:

Adds a `loading` state to both `Button` and `IconButton`. The markup follows the recommended path forward from accessibility design, and mirrors the implementation from the `Export CSV` button as part of GH org security coverage pages.

**Implementation details**

- We use a visually hidden label instead of changing the text of the button to handle text announcements better, and to not change the size of the button.
- `aria-disabled` instead of `disabled` to maintain focus management

**New props:**
`loading`: boolean
`loadingAnnouncement`: default to `Loading` or provide a custom message to be announced on ATs

Closes https://github.com/github/primer/issues/2680

### Screenshots

https://github.com/primer/react/assets/18661030/426caee6-4cb6-478f-b5f8-a84ba26e356e

https://github.com/primer/react/assets/18661030/67d49170-974b-4ed3-8e1b-022a6feabfca

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp:
Hi! There is another bug when aria-label is present, tried to address that in this PR (not merged): https://github.com/primer/react/pull/4459

Should I copy those changes over here?

mperrotti:
@siddharthkp - yes, please copy those changes over as well

siddharthkp:
Sorry, while copying over my changes, I ran into a couple questions

On this line: https://github.com/primer/react/pull/4485/files#diff-515525e4a59d9a55e13c6c60155ec6782c78a026aae4c540c324cd8e11c83a48R102-R105,

> aria-labelledby is needed because the accessible name becomes unset when the button is in a loading state.

Looking at this comment, is this still true? It looks like we set `aria-describedby` not label aria-label or aria-labelledby 🤔

mperrotti:
@siddharthkp I'm not sure I understand your question because we set all three depending on the situation.

We set `aria-label` for `IconButton` (see [Storybook example](https://primer.style/react/storybook/?path=/story/components-iconbutton--default))

We set `aria-labelledby` for either of these cases:
- `loading` prop is `true`
- `aria-labelledby` prop has been passed

We set `aria-describedby` for either of these cases:
- `loading` prop is `true`
- `aria-describedby` prop has been passed

siddharthkp:
Sorry, I wasn't clear!

Looking at this block of code: https://github.com/primer/react/pull/4485/files#diff-515525e4a59d9a55e13c6c60155ec6782c78a026aae4c540c324cd8e11c83a48R102-R105: 

```ts
// aria-labelledby is needed because the accessible name becomes unset when the button is in a loading state.
// We only set it when the button is in a loading state because it will supercede the aria-label when the screen
// reader announces the button name.
aria-labelledby={loading ? `${uuid}-label` : ariaLabelledBy}
```

and wondering why/when _accessible name becomes unset when the button is in a loading state_?

I removed it that line and saw no changes which really confused me. I'm probably missing something very obvious here, thanks for your patience with me!

siddharthkp:
> I had the same experience in the browser, but it caused the following test to fail:

Oh that's rough! Ideally, I'd like to optimise it for the browser and fix/workaround the jest error. UNLESS, we also see similar jest errors in dotcom integration tests?

cc @joshblack @broccolinisoup @pksjce (nerd sniping 🤭) have you seen something like this before?

packages/react/src/Button/ButtonBase.tsx
```diff
@@ -64,45 +84,65 @@ const ButtonBase = forwardRef(
...
+          data-loading={Boolean(loading)}
+          data-no-visuals={!LeadingVisual && !TrailingVisual && !TrailingAction ? true : undefined}
+          data-size={size === 'small' || size === 'large' ? size : undefined}
+          aria-describedby={[loadingAnnouncementID, ariaDescribedBy]
+            .filter(descriptionID => Boolean(descriptionID))
+            .join(' ')}
+          // aria-labelledby is needed because the accessible name becomes unset when the button is in a loading state.
+          // We only set it when the button is in a loading state because it will supercede the aria-label when the screen
+          // reader announces the button name.
+          aria-labelledby={loading ? `${uuid}-label` : ariaLabelledBy}
```
TylerJDev, line 105: I'm wondering if we could add `ariaLabelledBy` to the first condition too? 

(i.e. `${uuid}-label ${ariaLabelledBy}` : `ariaLabelledBy`)

This allows instances that rely on `aria-labelledby` to remain covered, even if the button is in a loading state.

mperrotti: Oh I didn't realize we could have 2 values in aria-labelledby. That could work! I'll give it a try...

mperrotti:
@siddharthkp - do you think @TylerJDev 's suggestion will fix our problem?

TylerJDev:
> @siddharthkp - do you think @TylerJDev 's suggestion will fix our problem?

I believe the problem would still persist here if there's no `ariaLabelledBy` being passed. I'd say utilizing `aria-labelledby` as you're doing now is definitely a way to go about it, while also adding the `ariaLabelledBy` prop to the first condition. 

> We have a component for that: https://github.com/primer/react/blob/main/packages/react/src/internal/components/VisuallyHidden.tsx

This would actually be a good alternative! We wouldn't need to reference the inner span anymore if we used this, and it _should_ be visually equivalent to `visibility: hidden`. I might prefer this route, as we wouldn't need to handle `aria-labelledby` directly anymore.

packages/react/src/Button/Button.examples.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+      return
+    }
+
+    setSuccess(actionResult)
+  }
+
+  return (
+    <>
+      <VisuallyHidden aria-live="polite">{!loading && success ? 'Export completed' : null}</VisuallyHidden>
+      <Button loading={loading} leadingVisual={DownloadIcon} onClick={onClick('error')}>
```
khiga8, line 39:
```suggestion
      <Button loading={loading} leadingVisual={DownloadIcon} onClick={onClick('success')}>
```
Should this be `success`? I noticed that the live region doesn't get updated with the `Export completed` announcement in [the storybook success example](https://primer-ef8e619bcb-13348165.drafts.github.io/storybook/iframe.html?args=&id=components-button-examples--loading-status-announcement-successful&viewMode=story). 

It seems to be getting updated for the error example though!

packages/react/src/Button/Button.examples.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+        Export (success)
+      </Button>
+    </>
+  )
+}
+
+export const LoadingStatusAnnouncementError = () => {
```
khiga8, line 46: In this example, the failure state is communicated to assistive technologies via announcement, but there's no visual indication that anything failed. 

Do we have guidance around communicating errors visually? If so, should this example include that?

packages/react/src/Button/Button.examples.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+      setError(actionResult)
+      return
+    }
+
+    setError(!actionResult)
+  }
+
+  return (
+    <>
+      <VisuallyHidden aria-live="polite">{!loading && error ? 'Export failed' : null}</VisuallyHidden>
```
khiga8, line 71: Should the announcement utility components be used in these examples instead, or are they only recommended for use internally? (cc: @joshblack)

```suggestion
      <VisuallyHidden><Status>{!loading && error ? 'Export failed' : null}</Status></VisuallyHidden>
```

packages/react/src/Button/Button.examples.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+import {DownloadIcon} from '@primer/octicons-react'
+import {VisuallyHidden} from '../internal/components/VisuallyHidden'
+
+const meta: Meta<typeof Button> = {
+  title: 'Components/Button/Examples',
+} as Meta<typeof Button>
+
+export default meta
+
+export const LoadingStatusAnnouncementSuccessful = () => {
```
khiga8, line 13: In these examples, when the loading state is activated, the focus gets lost and removed from the button. I believe this is a [2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html) issue, and may result in some screen reader users ending up at the top of the page. Could we avoid focus loss?

khiga8 (commented):
Took a look because I am looking at an adjacent topic of [buttons with dynamically updated labels](https://github.com/github/accessibility-audits/issues/5629)! :)

lindseywild:
@mperrotti is there anything the a11y v-team can help with to get this PR along the finish line?

packages/react/src/Button/Button.examples.stories.tsx
mperrotti: Great point! I'll take a look through our docs and update this.

packages/react/src/Button/Button.examples.stories.tsx
mperrotti: I don't think we had these when I first opened this PR. I can update and use them.

packages/react/src/Button/Button.examples.stories.tsx
mperrotti: Good catch. I'll take a look.

lindseywild:
@mperrotti Thanks! I see that Kate and Tyler both weighed in already (they're both on the v-team 😄 ) so let me know if you need anything else once their feedback is addressed!

packages/react/src/Button/ButtonBase.tsx
```diff
@@ -65,46 +85,76 @@ const ButtonBase = forwardRef(
...
-                <TrailingAction />
-              </Box>
-            )}
-          </>
+            </>
+          )}
+        </StyledButton>
+        {loading && (
+          <VisuallyHidden>
+            <Status id={loadingAnnouncementID}>{loadingAnnouncement}</Status>
```
TylerJDev, line 154: Not a blocker, I think we can add this functionality after this PR has been merged!

We may want to add functionality that will only announce `loadingAnnouncement` once (n) seconds have passed. For example, if the loading state is toggled for more than 2 seconds, announce `loadingAnnouncement` through the live region by conditionally rendering the region after that amount of time has been passed.

TylerJDev (approved):
LGTM with 1 non-blocking comment!

packages/react/src/Button/ButtonBase.tsx
mperrotti: @joshblack is working on something to support that in https://github.com/primer/react/pull/4673

Going to wait until that merges, then update here. Thanks!

siddharthkp:
👋 Wanted to test with dotcom but would have to wait until integration tests are clean: https://github.com/github/github/pull/329278

siddharthkp:
@mperrotti New integration PR: https://github.com/github/github/pull/331055

I see a few failing tests. Can you please confirm if these are related to the PR or not?

mperrotti:
I see some color contrast tests failing, but we didn't change any colors in this PR: https://github.com/github/github/actions/runs/9713401583/job/26810192180?pr=331055

Also, I tested the prod-deployed Storybook against this preview and didn't see color differences.

mperrotti:
These failures don't seem related to buttons at all:
- https://github.com/github/github/actions/runs/9713401583/job/26810188969?pr=331055#step:15:10043
- https://github.com/github/github/actions/runs/9713401583/job/26810188969?pr=331055#step:15:10056

mperrotti:
@siddharthkp - I took a look at those failures, and they don't _seem_ to be related. Should we do a more in-depth test like a Review Lab deployment? Or does this seem safe to you?

mperrotti:
I think the failed VRT was a fluke because I tested it in prod, and it matches what we get in the "Actual" screenshot. Going to update snapshots.

Screenshot 2024-07-03 at 4 09 14 PM

mperrotti:
@siddharthkp @TylerJDev - I tried opening a new integration test PR with these changes, but I keep getting a mysterious failure in the `commit-and-push` job. Latest attempt: https://github.com/github/github/actions/runs/9783972307

siddharthkp:
> @siddharthkp @TylerJDev - I tried opening a new integration test PR with these changes, but I keep getting a mysterious failure in the commit-and-push job. Latest attempt: https://github.com/github/github/actions/runs/9783972307

Looking at the logs, I think It failed only on the last step to ping you on slack, it should have created the PR before that though

mperrotti:
@siddharthkp - that's what I thought too, but I don't see any PR created https://github.com/github/github/pulls?q=is%3Apr+is%3Aopen+Integration+tests+for+PRC

siddharthkp:
ui/packages with [failing tests](https://github.com/github/github/pull/333742):

```
- @github-ui/filter
- @github-ui/pull-request-viewer
```

Triaging them now

siddharthkp:
Found the bug!


Code: 

```tsx
<Button leadingVisual={CommentIcon} count={3} />
```

| on main | on branch |
|--------|--------|
| button with comment icon and a count | button with comment icon but no count | 


I have pushed a story to this branch with the failing case

mperrotti:
@siddharthkp - everything should be good to go! I'll kick off another integration test PR as soon as we have a Canary release  available

mperrotti:
VRTs need to be updated now that we fixed the bug where whitespace characters were not being rendered between two elements. For example: the single-select ActionMenu button label: https://primer.style/react/storybook/?path=/story/components-actionmenu-features--single-select

mperrotti:
I kicked off an update to the integration test PR: https://github.com/github/github/actions/runs/10064634051

emilybrick (approved):
Looks good! ✨

PR #4496, opened 2024-04-12
Docs: publish docs for SelectPanel v2 and Dialog v2

siddharthkp:
Based on Tooltip v2, added `docsId` for Dialog v2 and SelectPanel v2

### DIFF
--- a/packages/react/src/Dialog/Dialog.docs.json
+++ b/packages/react/src/Dialog/Dialog.docs.json
@@ -1,6 +1,7 @@
 {
-  "id": "drafts_dialog",
-  "name": "Dialog",
+  "id": "dialog_v2",
+  "docsId": "dialog",
+  "name": "Dialog v2",
   "status": "draft",
   "a11yReviewed": false,
   "stories": [],
--- a/packages/react/src/Dialog/Dialog.docs.json
+++ b/packages/react/src/Dialog/Dialog.docs.json
@@ -1,7 +1,7 @@
 {
   "id": "dialog_v2",
   "docsId": "dialog",
-  "name": "Dialog v2",
+  "name": "Dialog",
   "status": "draft",
   "a11yReviewed": false,
   "stories": [],
--- a/packages/react/src/SelectPanel/SelectPanel.docs.json
+++ b/packages/react/src/SelectPanel/SelectPanel.docs.json
@@ -1,5 +1,6 @@
 {
   "id": "select_panel",
+  "docsId": "select_panel",
   "name": "SelectPanel",
   "status": "alpha",
   "a11yReviewed": false,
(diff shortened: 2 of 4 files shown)


packages/react/src/drafts/SelectPanel2/SelectPanel.docs.json
```diff
@@ -1,6 +1,7 @@
 {
-  "id": "selectpanel_v2",
-  "name": "SelectPanel",
+  "id": "select_panel_v2",
+  "docsId": "select_panel",
+  "name": "SelectPanel v2",
```
siddharthkp, line 4: @broccolini Should we call this `SelectPanel v2` or just `SelectPanel`?

Seems like [Tooltip v2 is titled just Tooltip in the docs](https://primer.style/components/tooltip/react/beta), even though [docs.json says "Tooltipv2"](https://github.com/primer/react/blob/main/packages/react/src/TooltipV2/Tooltip.docs.json#L3)

packages/react/src/SelectPanel/SelectPanel.docs.json
```diff
@@ -1,5 +1,6 @@
 {
   "id": "select_panel",
+  "docsId": "select_panel",
```
siddharthkp, line 3: @broccolinisoup Do we need docsId on both v1 and v2 docs.json?

packages/react/src/drafts/SelectPanel2/SelectPanel.docs.json
broccolinisoup: Great catch! - And this is a [temporary hack](https://github.com/primer/design/pull/752/commits/116966476a1ff08de10aee82d758ca95014fad96) until we release https://github.com/primer/react/pull/4483 

So, we should keep the name as `SelectPanel` 😊

packages/react/src/SelectPanel/SelectPanel.docs.json
broccolinisoup: Yes please. We need the unique docsId in all components that we want to group together.

packages/react/src/Dialog/Dialog.docs.json
```diff
@@ -1,6 +1,7 @@
 {
-  "id": "drafts_dialog",
-  "name": "Dialog",
+  "id": "dialog_v2",
+  "docsId": "dialog",
+  "name": "Dialog v2",
```
broccolinisoup, line 4: We can keep the name as `Dialog` here as well. 

The challenge with Dialog is that the story ids. Currently https://primer.style/components/dialog/react/alpha show the alpha dialog but the examples are belong to the draft dialog 😵  (If I am not mistaken) 

Can we take the dialog v2 stories under draft? and remove that `v1` from the alpha dialog to correct the story ids? I also remember there was a confusion around this and this is why we made this change, right? So we might maybe need to come up with custom solution for dialog if we don't want to revert the stories back.

broccolinisoup (commented):
Looks good! Left some comments 🙌🏻

siddharthkp:
@broccolinisoup Updated the names!

broccolinisoup (approved):

PR #4501, opened 2024-04-15
ActionBar: Make aria-label mandatory

pksjce:
Closes # (Didn't create an issue)

Voiceover announcement was wonky without a mandatory `aria-label` on the `ActionBar` component.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/lucky-zebras-stare.md
+++ b/.changeset/lucky-zebras-stare.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+ActionBar: Make it such that either aria-label or aria-labelledby is present
--- a/packages/react/src/drafts/ActionBar/ActionBar.stories.tsx
+++ b/packages/react/src/drafts/ActionBar/ActionBar.stories.tsx
@@ -30,7 +30,7 @@ export default {
 } as Meta<typeof ActionBar>
 
 export const Default = () => (
-  <ActionBar>
+  <ActionBar aria-label="Toolbar">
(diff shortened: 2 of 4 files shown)


broccolinisoup (approved):
Looks good!

One thing we did for UnderlineNav is to add a [runtime check](https://github.com/primer/react/blob/main/packages/react/src/UnderlineNav/UnderlineNav.tsx#L200C12-L200C12) for aria-label as well along with TS required notation. We might consider a similar practise here.

TylerJDev:
I'm wondering if we might want to also add `aria-labelledby` as an option, being that you'd be required to use one of the two, but not both, nor none.

pksjce:
@TylerJDev - Like this? https://github.com/primer/react/blob/0a8aff8fd55dfc0a7090711d4b5cb9f6ba7bfb6f/packages/react/src/Button/types.ts#L19 ?

I like it! Is it common that `aria-labelledby` will be required for a component like this?

TylerJDev:
> @TylerJDev - Like this?

Yup, exactly!

> Is it common that aria-labelledby will be required for a component like this?

It may not be as common, but could be the better option if the context for the `ActionBar` already exists on the page. I think the most common use case of this will be when there's a heading above the `ActionBar` that references the content underneath (similarly to how the CommentBox has the "Add a comment" heading above its toolbar).

TylerJDev (approved):

PR #4509, opened 2024-04-17
docs: Update React version referenced in contributor docs

smockle:
Node.js 16 is unsupported, and the current [`.nvmrc`](https://github.com/primer/react/blob/main/.nvmrc) references Node.js 20.


smockle:
> If these changes should not result in a new version, you're good to go.

Works for me; I don’t think we need to cut a new version for a contributor docs update.

joshblack (approved):
Thanks for this!

PR #4513, opened 2024-04-19
Export things we need for CommentEditor

owenniblock:
Because we're housing the MarkdownToolbar in dotcom, we need to export a couple of things that we use there.

#### New

Exports `createComponent`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/index.ts
```diff
@@ -45,6 +45,10 @@ export {useRefObjectAsForwardedRef} from './hooks/useRefObjectAsForwardedRef'
 export {useResizeObserver} from './hooks/useResizeObserver'
 export {useResponsiveValue} from './hooks/useResponsiveValue'
 
+// Utils
+export {createComponent} from './utils/create-component'
+export type {ComponentProps} from './utils/types/ComponentProps'
```
joshblack, line 50: Just wanted to ask, is there any reason to use this over `React.ComponentPropsWithoutRef`? (Or `React.ComponentProps`)

packages/react/src/index.ts
```diff
@@ -45,6 +45,10 @@ export {useRefObjectAsForwardedRef} from './hooks/useRefObjectAsForwardedRef'
 export {useResizeObserver} from './hooks/useResizeObserver'
 export {useResponsiveValue} from './hooks/useResponsiveValue'
 
+// Utils
+export {createComponent} from './utils/create-component'
```
joshblack, line 49: For stability, is this something that we feel like will be stable/unlikely to change without a breaking change or would it be helpful to have a "bake period" in drafts/experimental before promoting to stable?

Wasn't sure about the status of the helper and wanted to check-in 👍 Totally understand either way.

keithamus: It'll likely be removed when we upgrade to React 19. In lieu of that I think any changes can happen alongside major releases. I don't anticipate huge changes.

packages/react/src/index.ts
owenniblock: Good idea! I've done this, thanks @joshblack

joshblack (approved):

PR #4517, opened 2024-04-19
Fix `Stack` direction prop

langermank:
We moved from `orientation` to `direction` but missed this classname change 😄

### DIFF
--- a/packages/react/src/Stack/Stack.stories.tsx
+++ b/packages/react/src/Stack/Stack.stories.tsx
@@ -15,9 +15,33 @@ export default meta
 export const Default: Story = {
   render: () => (
     <Stack>
-      <div>First</div>
-      <div>Second</div>
-      <div>Third</div>
+      <div
+        style={{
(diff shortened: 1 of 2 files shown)


broccolinisoup (approved):
Looks good! 🚀 Is there a way to add a test for these type of cases? 👀 Just to make sure the prop name and the data attribute name aligns since this is [our strategy](https://github.com/primer/react/pull/4449) now.

joshblack (approved):

joshblack:
@broccolinisoup great question, might be worth a test helper to make these types of things easier to test. Something that is surprising about this to me is that we had tests over at: https://github.com/primer/react/blob/main/packages/react/src/Stack/__tests__/Stack.test.tsx#L74-L96 and that VRT is not failing so I wonder what's going on 🤔 

Side-note, @langermank do we want the values of `direction` to align with `flex-direction` values or is `horiziontal` and `vertical` okay?

siddharthkp (approved):

PR #4527, opened 2024-04-22
Add `labelWrap` prop to `Button`

langermank:
This has been an ongoing issue with the Button component across all frameworks. [Previously](https://github.com/primer/react/pull/4062), we had hoped to introduce this change that allows button labels to wrap by default if the label string is too long. However, after testing this behind a [feature flag](https://github.com/github/primer/issues/3123) for a bit we continued to find random issues across dotcom as the change can lead to unpredictable layout regressions.

In Rails, to unblock a few open accessibility issues I decided to introduce the change with a [param](https://github.com/primer/view_components/pull/2751).

The only way to make this change the default is with a major release, and even then we would need to babysit it and individually fix each area where layout breaks. I don't think this is realistic, and feel like introducing it as a prop should be sufficient for unblocking teams and mitigating accessibility issues.

Closes https://github.com/github/primer/issues/3123

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Button/Button.features.stories.tsx
```diff
@@ -135,3 +136,23 @@ export const Small = () => <Button size="small">Default</Button>
...
+      </Button>
+      <Button size="large" labelWrap>
+        This large button label will wrap if the label is too long
+      </Button>
+      <Button labelWrap leadingVisual={HeartIcon} trailingVisual={EyeIcon}>
+        This button label will wrap if the label is too long
+      </Button>
+      <Button labelWrap trailingAction={TriangleDownIcon}>
+        This button label will wrap if the label is too long
+      </Button>
```
jonrohan, line 155: Maybe include one that isn't wrapping to show along with the wrapping ones?

jonrohan (approved):
👍🏻 Pretty straight forward

PR #4531, opened 2024-04-23
Make sure Link is underlined when `inline={true}`

dipree:
Related to https://github.com/github/primer/issues/2099

Currently, the `inline` property only works in combination with the `data-a11y-link-underlines` attribute. However, the link should always be underlined if `inline` is set to true and no attribute is provided.

#### Changed

If `inline` is set to `true`, the Link will be underlined even if not using the `data-a11y-link-underlines` attribute.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Link/Link.tsx
+++ b/packages/react/src/Link/Link.tsx
@@ -27,16 +27,21 @@ const hoverColor = system({
 const StyledLink = styled.a<StyledLinkProps>`
   color: ${props => (props.muted ? get('colors.fg.muted')(props) : get('colors.accent.fg')(props))};
 
-  /* By default, Link does not have underline */
+  /* By default, Link is not underlined. */
   text-decoration: none;
 
-  /* You can add one by setting underline={true} */
+  /* If inline or underline are set to true Link is underlined. */
   text-decoration: ${props => (props.underline ? 'underline' : undefined)};
+  text-decoration: ${props => (props.inline ? 'underline' : undefined)};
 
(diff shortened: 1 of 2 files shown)


packages/react/src/Link/Link.tsx
```diff
@@ -27,17 +27,22 @@ const hoverColor = system({
...
   color: ${props => (props.muted ? get('colors.fg.muted')(props) : get('colors.accent.fg')(props))};
 
-  /* By default, Link does not have underline */
+  /* By default, Link is not underlined. */
   text-decoration: none;
 
-  /* You can add one by setting underline={true} */
+  /* If inline or underline are set to true Link is underlined. */
   text-decoration: ${props => (props.underline ? 'underline' : undefined)};
+  text-decoration: ${props => (props.inline ? 'underline' : undefined)};
```
joshblack, line 35: I think we'll need one line otherwise the `inline` value will always override, maybe something like:

```suggestion
  text-decoration: ${props => (props.underline || props.inline ? 'underline' : undefined)};
```

joshblack (changes requested):

packages/react/src/Link/Link.tsx
dipree: Thank you, it's not a problem but definitely better style. Remaining problem is the data attribute if set to `false` affecting the `underline: true`. We might want to deprecate `underline` first 🤔

PR #4535, opened 2024-04-25
feat: add experimental IssueLabel component

joshblack:
> [!NOTE]
> This is a part of our weekly mob programming session 👀 

Add experimental `IssueLabel` component


lukasoppermann:
Hey, I am a bit concern with the name `Label`. A label is different from what we want. The correct name would be `IssueLabel`, can we rename it?

langermank:
Notes for next pair session:

- Using this as a reference https://github.com/github/github/blob/1f7d3bba98e3d62a150a57ce4f72f4ec8631c018/ui/packages/label-token/LabelToken.tsx#L35
- We agreed to keep color2k

joshblack:
Cherry picked over into: https://github.com/primer/react/pull/4960 to make it easier to merge 👍

PR #4536, opened 2024-04-26
ActionBar: Add a few fixes and tests

pksjce:
Closes # No particular issue

### Changelog

- Add the `unsafeDisableTooltip` prop for IconButton
- Add proper aria-label in story
- Add visual tests


e2e/components/drafts/ActionBar.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        test('default @vrt', async ({page}) => {
+          await visit(page, {
+            id: 'drafts-components-actionbar--comment-box',
+            globals: {
+              colorScheme: theme,
+            },
+          })
+
+          // Default state
+          await page.getByText('Write').click()
```
broccolinisoup, line 18: Do we need to press the "Write" tab for the default state? We might want to take the screenshot of the default state first without performing any action.

packages/react/src/drafts/ActionBar/ActionBar.test.tsx
```diff
@@ -33,3 +33,11 @@ describe('ActionBar', () => {
     expect(results).toHaveNoViolations()
   })
 })
+
+/* Test suite
+1. Did it render all the iconbuttons?
```
broccolinisoup, line 38: I remember from UnderlineNav and it was easier to write those visual tests in Playwright than Jest but it is up to you 😊

packages/react/src/drafts/ActionBar/ActionBar.tsx
```diff
@@ -299,7 +299,7 @@ export const ActionBarIconButton = forwardRef((props: ActionBarIconButtonProps,
     const domRect = (ref as MutableRefObject<HTMLElement>).current.getBoundingClientRect()
     setChildrenWidth({text, width: domRect.width})
   }, [ref, setChildrenWidth])
-  return <IconButton ref={ref} size={size} {...props} variant="invisible" />
+  return <IconButton ref={ref} size={size} {...props} variant="invisible" unsafeDisableTooltip={false} />
```
broccolinisoup, line 302: 💖

e2e/components/drafts/ActionBar.test.ts
pksjce: Oops my bad! we don't need to click it!

packages/react/src/drafts/ActionBar/ActionBar.test.tsx
pksjce: Yes, I think all the responsive stuff is better off tested in playwright. It's definitely harder(is it even possible?) to wrangle it in jest.
Will make a better division of test cases!

e2e/components/drafts/ActionBar.test.ts
broccolinisoup: No worries, let me know when the snapshots are generated, I'll give another review 😊

broccolinisoup (approved):

TylerJDev (approved):

mperrotti (approved):

PR #4537, opened 2024-04-26
Fix: Ensure dynamically updated label is announced

khiga8:
Closes https://github.com/github/accessibility-audits/issues/5629

### Changelog

I'm addressing an audit ticket that points out how the [dynamically updated counter button in Storybook](https://primer.style/react/storybook/iframe.html?globals=&args=&id=components-button-features--trailing-counter&viewMode=story) label is not announced when incremented. It turns out there isn't really a standardized approach to ensure announcements happen when a button is dynamically updated. 

I ended up testing various approaches across different screen reader/browser combinations which you can check out in https://github.com/github/accessibility/pull/6222. (I welcome your feedback on that PR).

Ultimately through my testing, I found that dynamically updating the `aria-label` performs the best in terms of ensuring the button state is communicated. However, this is a brittle approach so I recommend using a live region announcement (unless an `aria-label` is already being set on the button).

This PR updates the storybook stories accordingly with a note.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

[See my changes 🔍 ](https://primer-9f1d0c5714-13348165.drafts.github.io/storybook/?path=/story/components-button-features--trailing-counter)
[Videos from testing across multiple screen reader/browsers](https://github.com/github/accessibility/blob/9cac264a5332c90fed00ee943e34f9d2186fd602/docs/wiki/screen-reader-testing/dynamically-updated-buttons-support-april-2024.md#2-updating-the-aria-label-in-addition-to-the-button-content)

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack:
@khiga8 just wanted to share [example usage of `count`](https://musical-adventure-wlr3n3k.pages.github.io/?name=button&attribute=count) if it was helpful! With the link you shared, it seemed like it was looking in `GitHub` which would be for GitHub UI packages versus Primer usage (sorry if that was confusing or if I'm misunderstanding something here 👀)

Something interesting is that there are two instances of this where both `aria-label` and `count` is used 🤔

packages/react/src/Button/Button.features.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import {EyeIcon, TriangleDownIcon, HeartIcon} from '@primer/octicons-react'
 import React, {useState} from 'react'
-import {Button} from '.'
+import {Button, Link} from '.'
```
joshblack, line 3: Would it be possible to import `Link` from `'../Link'`? This could help to prevent the circular reference in the `Button/index.js` entrypoint that's happening 👀

khiga8:
EEEP @joshblack thank you for the primer query link!!! I'll have to dive into it next week!

joshblack (approved):
Looks great, thanks for doing this! 👍 Just need snapshots updated and should be good to go

packages/react/src/Box/Box.features.stories.tsx
```diff
@@ -11,6 +11,12 @@ export const Border = () => (
   <Box sx={{borderWidth: 1, borderStyle: 'solid', borderColor: 'border.default', p: 3}}>Box with border</Box>
 )
 
+export const LiveRegionTest = () => (
+  <Box aria-live="polite" aria-atomic="true">
+    Test
+  </Box>
+)
+
```
khiga8, line 19:
```suggestion
```

packages/react/src/Button/Button.features.stories.tsx
```diff
@@ -16,44 +18,81 @@ export const LeadingVisual = () => <Button leadingVisual={HeartIcon}>Leading vis
...
+        </p>
+      </>
+    )
+  }
+}
 export const TrailingCounter = () => {
   const [count, setCount] = useState(0)
+  const onClick = () => {
+    setCount(count + 1)
+    announce(`Watch ${count + 1}`)
```
khiga8, line 46: Hey @joshblack! I decided to update the approaches in these examples to use a live region, based on the [conclusion](https://github.com/github/accessibility/blob/main/docs/wiki/screen-reader-testing/dynamically-updated-buttons-support-april-2024.md#takeaways) I came to that the `aria-live` hack shouldn't be used, unless it's already being considered for other reasons.

I am thinking the next steps for me after this PR are:
* a dedicated storybook page that dives into different approaches for communicating button state.
* adding a section to the interface guidelines.

lindseywild (approved):
Looks good from an accessibility perspective!

langermank (approved):

PR #4543, opened 2024-04-29
Allows asterisk in `FormControl.Label` to be adjustable

TylerJDev:
Closes: https://github.com/github/primer/issues/3263
https://github.com/github/accessibility-audits/issues/5724

Provides more flexibility to our required indicator (currently an asterisk). This PR adds the ability to change the text of the indicator, as well as hide it from AT via the new `requiredIndicator` prop.

#### New

Adds two new props in `FormControl.Label`:

* `requiredText` - Allows you to change the required indicator from '_*_', to the value provided. Defaults to '_*_'.
* `requiredIndicator` - Allows you to toggle between the indicator being present or not in the accessibility tree. Defaults to 'true'.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/FormControl/_FormControlLabel.tsx
```diff
@@ -8,12 +8,14 @@ export type Props = {
    * Whether the label should be visually hidden
    */
   visuallyHidden?: boolean
+  requiredText?: string
+  requiredIndicator?: boolean
```
TylerJDev, line 12: I'm wondering if this might be better handled on `<FormControl>` itself, as to apply the required indicator settings to all labels that exist within. The main con here is that there could be cases where you'd only want to adjust one of the indicators, not all.

packages/react/src/internal/components/InputLabel.tsx
```diff
@@ -55,7 +59,7 @@ const InputLabel: React.FC<React.PropsWithChildren<Props>> = ({
       {required ? (
         <Box display="flex" as="span">
           <Box mr={1}>{children}</Box>
-          <span>*</span>
+          {requiredText && <span aria-hidden={!requiredIndicator ? true : undefined}>{requiredText}</span>}
```
TylerJDev, line 62: Right now `requiredIndicator` only hides the indicator to the accessibility tree (i.e. `aria-hidden="true"`). This is because the indicator is usually important in discerning which form fields are required visually, whereas it being present in the accessibility tree is not.

You can still technically hide the indicator by using an empty string in `requiredText` (e.g. `requiredText=""`).

TylerJDev:
This is one of many ways to handle how we do required indicators in `FormControl.Label`. Would love any other opinions on how to make this easier for consumers :eyes:

More context: https://github.com/github/primer/issues/3263

packages/react/src/FormControl/FormControl.docs.json
```diff
@@ -56,6 +56,18 @@
...
+          "name": "requiredText",
+          "type": "string",
+          "defaultValue": "'*'",
+          "description": "The text to display when the field is required"
+        },
+        {
+          "name": "requiredIndicator",
+          "type": "boolean",
+          "defaultValue": "true",
+          "description": "Whether to show or hide the required text in the accessibility tree, but still visually show the required text."
```
owenniblock, line 69:
```suggestion
          "description": "Whether to show or hide the required text in the accessibility tree, the required text is still shown visually."
```

packages/react/src/FormControl/_FormControlLabel.tsx
owenniblock: It makes sense to me to add this to the `FormControl.Label` 🤔 Because that's where the indicator is. I could see users wanting to be able to set a default `requiredText` for an entire form but it looks like there's no `Form` element so that's not really possible. I guess in that instance an engineer would probably create a custom component which inherits from the `FormControl.Label`.

packages/react/src/internal/components/InputLabel.tsx
owenniblock: Do we want to allow an empty requiredText when the element is required 🤔 I _think_ I would be in favour here of that not being possible. I don't feel _super_ strongly about this but it could lead to audit issues in future if someone hides the requiredText but sets a field to be programatically required.

owenniblock (commented):
Looks good to me, one comment on the edge case of setting requiredText to be empty. Also a nitpick on the docs wording 🙂

packages/react/src/FormControl/_FormControlLabel.tsx
TylerJDev: Yeah I agree! I think if there's truly value here we could come back to it, but for now I think handling it on `FormControl.Label` is good enough.

packages/react/src/internal/components/InputLabel.tsx
TylerJDev: That's a good point, I'm not entirely sure if there are instances today where we'd want to opt out of a visual indicator for required fields. The only instance that I could think of, is if someone were to opt for marking optional fields visually rather than required fields (e.g. `(optional)`). I don't think we have a case like that in GH today, so I'm okay with defaulting to '*' if `requiredText` is an empty string.

packages/react/src/internal/components/InputLabel.tsx
```diff
@@ -52,10 +56,10 @@ const InputLabel: React.FC<React.PropsWithChildren<Props>> = ({
       }}
       {...props}
     >
-      {required ? (
+      {required || requiredText ? (
         <Box display="flex" as="span">
           <Box mr={1}>{children}</Box>
-          <span>*</span>
+          <span aria-hidden={!requiredIndicator ? true : undefined}>{!requiredText ? '*' : requiredText}</span>
```
owenniblock, line 62: Personal preference for inverting the input variables here to remove the `!` (and maybe using the Nullish coalescing operator 🤔)

```suggestion
          <span aria-hidden={requiredIndicator ? undefined : true}>{requiredText ?? '*'}</span>
```

Bit of a nitpick so will approve anyway ✅

owenniblock (approved):

PR #4545, opened 2024-04-30
Work on implementing code connect to allow for realistic code preview in figma

lukasoppermann:
### Changelog
- adds figma code connect files for some componenrts
- adds `figma.config.json` form code connect config
- adds deps and script to `packages/react/package.json`
- adds action to run code connect whenever a `.figma.tsx` file is changed
- adds script to generate code connect for icons

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

No effect on package output, just for figma, but it can't be version within figma.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/tsconfig.build.json
```diff
@@ -15,6 +15,7 @@
     "src/utils/story-helpers.tsx",
     "**/*.stories.tsx",
     "**/*.test.tsx",
+    "src/**/*.figma.tsx",
```
joshblack, line 18: Quick question, do the files for figma need to be included in the package or can they be used on push in the actions workflow and then discarded for the package build? 👀

packages/react/package.json
```diff
@@ -214,6 +216,7 @@
     "storybook-addon-turbo-build": "2.0.1",
     "styled-components": "5.3.11",
     "terser": "5.27.0",
+    "ts-node": "10.9.2",
```
joshblack, line 219: We have the `tsx` utility for running ts files if that helps! Sadly we had to move off of `ts-node` for some other scripts, just wanted to share that utility in case it wasn't obvious.

packages/react/tsconfig.build.json
lukasoppermann: They do NOT need to be included in the package.

They are also not really interpreted but only "read" by the figma tool, which is why it may be hard to get them 100% TS compatible.

packages/react/tsconfig.build.json
```diff
@@ -15,6 +15,7 @@
     "src/utils/story-helpers.tsx",
     "**/*.stories.tsx",
     "**/*.test.tsx",
+    "**/*.figma.tsx",
```
lukasoppermann, line 18: @joshblack neither this now the change in `packages/react/tsconfig.json` seems to make typescript ignore my files. Any idea why?

.github/workflows/figma_connect_publish.yml
```diff
@@ -0,0 +1,17 @@
+name: Figma Connect Publish on Change
+
+on:
+  push:
+    paths:
+      - packages/react/src/components/**/*.figma.tsx
+    branches:
+      - main
```
joshblack, line 8: Do we need to tie this to releases for the package or is it okay to do it on push to `main`? (Or do we need both?)

.github/workflows/figma_connect_publish.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  code-connect:
+    name: Code Connect
+    runs-on: ubuntu-latest
+    steps:
+      - run: npx figma connect publish
+        env:
+          FIGMA_ACCESS_TOKEN: ${{ secrets.FIGMA_ACCESS_TOKEN }}
```
joshblack, line 17: Just wanted to ask, how does this access token get generated?

packages/react/package.json
```diff
@@ -50,6 +50,7 @@
     "build:storybook:visual-testing": "storybook build",
     "build:components.json": "tsx script/components-json/build.ts",
     "build:precompile-color-schemes": "tsx script/precompile-color-schemes.ts",
+    "figma:sync": "export $(cat .env | xargs) && npx figma connect publish",
```
joshblack, line 53: Is it possible to do this without the local `.env` file and have folks testing it pass the ENV variable explicit?

```suggestion
    "figma:sync": "figma connect publish",
```

packages/react/src/Avatar/Avatar.figma.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * -- This file was auto-generated by `figma connect create` --
+ * `props` includes a mapping from Figma properties and variants to
+ * suggested values. You should update this to match the props of your
+ * code component, and update the `example` function to return the
+ * code example you'd like to see in Figma
+ */
+
+figma.connect(
+  Avatar,
+  'https://www.figma.com/file/GCvY3Qv8czRgZgvl1dG6lp/Primer-Web?type=design&node-id=3805-11&mode=design&t=n5UtEmPJTCzDEJKK-4',
```
joshblack, line 15: Just curious, are these URLs found by going to the Primer Web figma file and finding the specific component or is there something else to it? 👀

packages/react/tsconfig.build.json
joshblack: @lukasoppermann what are you looking for it to be ignored from? Pulled it down real quick and it seemed to be ignored from the build but just wanted to check-in to make sure I knew what you were looking to do 👀

packages/react/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "src/**/*.js",
     "src/**/*.ts",
     "src/**/*.tsx",
+    "!src/**/*.figma.tsx",
```
joshblack, line 13: Let me know if I'm misunderstanding, it seems like we would want this in our `type-check` task for CI to make sure that figma connect files are generated correctly but we would not want it in our package files, is that correct?

packages/react/figma.config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "codeConnect": {
+    "include": ["src/**/*.figma.tsx", "src/**/*.tsx"],
+    "exclude": ["test/**", "docs/**", "build/**"],
+    "react": {
+      "importPaths": {
+        "./src/*": "@primer/react"
```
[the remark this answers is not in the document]
lukasoppermann: Good question, I'd need to test it.

.github/workflows/figma_connect_publish.yml
lukasoppermann: I have to manually create it in Figma and than it has to be added to the repo. Sadly there is no org token available in Figma yet. The Figma folks are aware of this issue though.

.github/workflows/figma_connect_publish.yml
lukasoppermann: Hmm, releases is problematic if you want to update/fix stuff as we would not do a new release for a figma update. This is why I figured main is best.

packages/react/tsconfig.build.json
lukasoppermann: I want it to be ignored from the type checking as well. This would allow me to remove this line: https://github.com/primer/react/pull/4545/files#diff-5fe13549f218c483abf30bee0a4a3001cd58d382586bfedf07299cdbe7721961R2

packages/react/src/Avatar/Avatar.figma.tsx
lukasoppermann: Yes, this is exactly how to do it.

packages/react/tsconfig.json
lukasoppermann: I think neither.

- def. not in package file
- there are some issues with making it typescript compliant as it is "read as text", this is why I should probably also not be checked.

Happy to jump on a pairing session to see if there is some way around ignoring it.

joshblack: @lukasoppermann sounds great, would love to pair up on it if you have the time!

joshblack (approved):

PR #4546, opened 2024-04-30
Add `TreeView.LeadingAction` sub-component 

### LINKED ISSUE #4542: [TreeView] Add area for `dragHandle` to be shown before the expand/collapse toggle
- add an area in the grid to add the drag handle (before any other icon/button/component)
- drag handle should be visible when hovering over the `TreeView.Item` 
- only take up the space if drag handle is passed from the consumer to tree view
- when drag handle is passed from consumer, shift child level items to match alignment

https://github.com/github/github/pull/321236/files#diff-785247148fc29c46a71b4de392214511ae6709b4516cc4e921d5e9b37b5a23d3

showing a drag handle in front of Item 1

ayy-bc:
Closes #4542 

### Changelog

This PR adds `TreeView.LeadingAction` sub component (like the leading visual, trailing visual) to create a slot for elements that go before the toggle and can be used for various use cases such as a drag handle.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TreeView/TreeView.tsx
```diff
@@ -120,6 +121,12 @@ const UlBox = styled.ul<SxProp>`
         outline: 2px solid transparent;
         outline-offset: -2px;
       }
+
+      .PRIVATE_TreeView-item-drag-handle {
```
dusave, line 125: Is this a defined pattern anywhere else?

ayy-bc: Not that I'm aware of (regarding this being a defined pattern). 

Since we need to show the drag handle when the user hovers over the `TreeItem`, I decided to toggle the visibility of that class (that contains the drag handle) from the `hover` property of the entire container 🤔 

@dusave

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -488,6 +511,12 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
             <div style={{gridArea: 'spacer', display: 'flex'}}>
               <LevelIndicatorLines level={level} />
             </div>
+            {dragHandle ? (
+              // eslint-disable-next-line jsx-a11y/no-static-element-interactions
+              <div className="PRIVATE_TreeView-item-drag-handle" onMouseDown={() => setIsExpanded(false)}>
+                {dragHandle}
```
mperrotti, line 517: I don't think we should require folks to style their own drag handles. We should ship with a default recommended style.

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -337,6 +358,7 @@ export type TreeViewItemProps = {
   onExpandedChange?: (expanded: boolean) => void
   onSelect?: (event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>) => void
   className?: string
+  dragHandle?: ReactElement
```
mperrotti, line 361: I'm not sure we need to require folks to pass `dragHandle` if they're passing `dragAndDrop` on the parent.

If we have a tree view where some nodes are draggable and others are not, we could assume all nodes are draggable by default. Then, we could add a prop to `TreeView.Item` that allows people to disable drag-and-drop functionality.

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -488,6 +511,12 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
             <div style={{gridArea: 'spacer', display: 'flex'}}>
               <LevelIndicatorLines level={level} />
             </div>
+            {dragHandle ? (
+              // eslint-disable-next-line jsx-a11y/no-static-element-interactions
+              <div className="PRIVATE_TreeView-item-drag-handle" onMouseDown={() => setIsExpanded(false)}>
```
mperrotti, line 516: How would a keyboard user drag and drop?

mperrotti, line 516: After I interact with the drag handle, I can't get focus to go back into the tree view. Maybe I'm just doing something wrong - so you and other reviewers should check to make sure you can reproduce this bug.

mperrotti (commented):
I have some initial feedback, but I really think we need somebody from a11y to take a look. For example, I'm not sure the "right"/accessible way to add interactive controls inside of a `role="treeitem"`.

It also looks like consumers will have to write their own drag-and-drop functionality since there are no props to handle `onDragStart`/`onDragEnd` callbacks. Also, we should agree on the right designs/interactions for a tree view when it's actively being dragged. For example:
- What does the node look like when the user is dragging it? Probably just knock the opacity down to make it look like a "ghost" of the original node.
- How do we indicate the prospective new position of the node while it's being dragged? Probably a blue line that acts like a cursor to show what position the node will be in after it's dropped.

ericwbailey (approved):
>only show dragHandle when user hovers over TreeView.Item

One small thing I'd love to sneak in: can we also show the DragHandle via `:focus-visible`? That way it will also be shown to keyboard-only user.

>I have some initial feedback, but I really think we need somebody from a11y to take a look. For example, I'm not sure the "right"/accessible way to add interactive controls inside of a role="treeitem".

This interaction pattern is really unique to the web, so it's going to take some time to research this. The other thought here is that my understanding of the shape of this work is getting initial functionality in for demo purposes, and then follow up with accessibility tweaks so it does not block the work.

With that context, I think we can consider this merge-able once other things folks reviewing have pointed out have been addressed. I simply don't know what will be needed yet. We have [a plan in place for discovering that](https://github.com/github/accessibility/issues/6055), but it will take time.

So, I'd consider this an approval from accessibility at a high level with the idea that there may need to be larger structural updates to follow as we learn what they'll be.

packages/react/src/TreeView/TreeView.tsx
ericwbailey: +1 to this one as well. I don't think we'll have a scenario where individual tree nodes would be prevented from being dragged and dropped?

ayy-bc:
@mperrotti @ericwbailey the initial idea was to just have some space before the expand/collapse toggle where users can place any element. I used `dragHandle` as the example as that is what we were interested in. 

If we were to ship with a default recommended style to always have the `dragHandle` show, I believe the drag and drop logic should also reside within TreeView (at least for onDragStart/onDragEnd). This would complicate things quite a bit for our use-case.

I think the better way to put this is: do you have recommendations on what we could do if we just wanted to place an element before the TreeView.Item toggle? The consumer would handle what that element does (i.e clicking on it, dragging it, etc).

siddharthkp:
> The other thought here is that my understanding of the shape of this work is getting initial functionality in for demo purposes, and then follow up with accessibility tweaks so it does not block the work.

@ayy-bc @ericwbailey Can you say a little more about the demo purposes?

(If it's something urgent with user demos, we can ship this quickly as is and keep iterating on it. If it's internal demos not on production, when we can use a different method to unblock the team)

siddharthkp:
Curious to hear your thoughts here @mperrotti,

This PR is very specific to drag and drop behavior, which is fine, as long as we also bake-in user feedback for dragging state.

> the initial idea was to just have some space before the expand/collapse toggle where users can place any element. I used dragHandle as the example as that is what we were interested in.

On a pattern level, however, the requirement/request from https://github.com/primer/react/issues/4542 and @ayy-bc's comment seems to be 2 things:

1. an icon button "slot" that is placed before both the expand/collapse button and the leading visual

   This feels similar to how we support both `trailingVisual` and `trailingAction` at the same time on components like Button, TextInput and PageHeader. 

   Is this a good case for add a `leadingAction` slot along with the existing `leadingVisual` slot on `TreeView.Item`?

2. The ability to hide the button with css
    
    This part feels use case specific, so it would be better to put this at the usage site instead of baking it in the component. 
    
    To be able to target the leadingAction slot, we would need a reliable selector like `[data-component=TreeView.leadingAction]`. (We do this in [other components like Button, ActionList, UnderlineNav, etc. already](https://github.com/search?q=repo%3Aprimer%2Freact%20data-component&type=code))

@ayy-bc Does that seem correct or am I missing a crucial part here?

ayy-bc:
@siddharthkp thanks for looking into this. We really appreciate it.

> Can you say a little more about the demo purposes?

We are planning to ship the first iteration of drag and drop with sub-issues with our currently ongoing epic. The changes will be out in production for everyone flagged into sub-issues to try. Regarding the urgency, it would help us a ton to be able to ship this out soon as we're currently using a local copy of TreeView inside of nested list view to test out our changes. 

Regarding your [second comment](https://github.com/primer/react/pull/4546#issuecomment-2095602584): 
1. Yes, that's spot on. We need an icon button slot before the expand/collapse toggle so we can place the drag handle there.
2. A little confused about this one. Ideally, the updated grid (with an area for leadingAction ~dragHandle) would apply to all the `TreeView.Item`'s based on some prop/attribute (maybe at the parent `TreeView` level, and the updated grid with space for leading action is only used when the consumer passes some attribute/prop so as to not break other users of TreeView). Then, the user can pass the icon button (as leading action) to any `TreeView.Item` and for those that don't have the leading action passed, they would just show some empty space there so as to not break the tree layout (we'll only be rendering drag handles for first level items to begin with). 
I'm not the best at styling but if the end user/usage site can pass the leading action (~dragHandle) as a child to the `TreeView.Item` with the `[data-component=TreeView.leadingAction]` and if that places the handle in the right place, that should work with us

ericwbailey:
>@ayy-bc @ericwbailey Can you say a little more about the demo purposes?
>
>(If it's something urgent with user demos, we can ship this quickly as is and keep iterating on it. If it's internal demos not on production, when we can use a different method to unblock the team)

Ah, sorry about that! I was referring to internal demos and the eventual adding of more Hubbers to the feature flag. So, no scary public-facing thing we need to rush to address.

---

For what it's worth, I really like the idea behind `leadingAction`. I also wonder if we'll run into a future state where the design calls for both drag and drop and a leading action?

siddharthkp:
> For what it's worth, I really like the idea behind leadingAction. I also wonder if we'll run into a future state where the design calls for both drag and drop and a leading action?

Definitely possible! Till now, the pattern has been to put action (or multiple actions) in trailing slots instead of leading

> Regarding the urgency, it would help us a ton to be able to ship this out soon as we're currently using a local copy of TreeView inside of nested list view to test out our changes.

Understood. Also, that's a good workaround in the time being!

I'll wait for @mperrotti's thoughts on `leadingAction` before we conclude on the direction

mperrotti:
I like the flexibility of a `leadingAction`, but I can't think of other possible leading actions since we typically put actions after the item label (in a "trailing" slot).

I feel comfortable moving forward with `leadingAction` so we can quickly unblock and iterate, but I don't think we need to have both `leadingAction` and `leadingVisual` yet.

packages/react/src/TreeView/TreeView.features.stories.tsx
```diff
@@ -695,6 +696,89 @@ export const NestedTrees: Story = () => {
...
+  const [isLoading, setIsLoading] = React.useState(false)
+  const [asyncItems, setAsyncItems] = React.useState<string[]>([])
+
+  let state: SubTreeState = 'initial'
+
+  if (isLoading) {
+    state = 'loading'
+  } else if (asyncItems.length > 0) {
+    state = 'done'
+  }
```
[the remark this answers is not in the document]
siddharthkp: +1

siddharthkp:
Great, we have a direction then!

@ayy-bc, I have an implementation in mind, I'll try to finish it tomorrow.

Do you mind if I update your PR itself with the changes? (because we have reviewers here already)

packages/react/src/TreeView/TreeView.tsx
dusave: Sorry, my comment wasn't clear. I meant was the _naming convention_ a defined pattern anywhere else? I'm thinking we probably don't want to draw attention to something marked as `PRIVATE`

packages/react/src/TreeView/TreeView.tsx
dusave: > I don't think we'll have a scenario where individual tree nodes would be prevented from being dragged and dropped?

@ericwbailey we _could_. If the user didn't have access to see an item, they wouldn't be able to drag it.

ayy-bc:
@siddharthkp yeah please feel free to update this PR with the changes you have in mind. Thanks a ton for getting the ball rolling in this one :) I'll take a look at the PR whenever you have your changes in

siddharthkp:
Leaving notes for myself 

TODOs In the story:
- [ ] implement fold on click
- [ ] implement hide until hovered

TODOs in the component:
- [ ] check if leading action should be aria-hidden
- [ ] update snapshots
- [ ] make the API always render an icon button

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -829,6 +847,19 @@ const TrailingVisual: React.FC<TreeViewVisualProps> = props => {
 
 TrailingVisual.displayName = 'TreeView.TrailingVisual'
 
+// ----------------------------------------------------------------------------
+// TreeView.LeadingAction
+
+const LeadingAction: React.FC<IconButtonProps> = props => {
```
ayy-bc, line 853: Can we make the `LeadingAction` accept children similar to `leadingVisual` and `trailingVisual`? This lets us (consumer) pass in any component (and associated props to that component) to the LeadingAction

ayy-bc: Since we already have a drag trigger component that is accessible and works around some quirks of NVDA, we'd love to use that (instead of rewriting the logic again within `LeadingAction` button) if possible.

ayy-bc: > We need an icon button slot before the expand/collapse toggle so we can place the drag handle there.

Think I made a small mistake here when asking for an icon button slot. I got it a little mixed up but ideally, we'd want similar functionality to `LeadingVisual` i.e a slot to pass child components

ayy-bc: https://github.com/primer/react/compare/ayy-bc/addTreeViewDragHandle...ayy-bc/suggestions

ayy-bc: Merged those changes to this PR

siddharthkp:
@ericwbailey Do you know if should add an `aria-hidden` to the leading action until we know it's accessibility story? (There's no way to get to it with keyboard navigation either)

ericwbailey:
@siddharthkp Yup, that'd be great. Ultimately, we'd need it and also `tabindex="-1"` applied to [any interactive element](https://ui.githubapp.com/storybook/?path=%2Fdocs%2Frecipes-nestedlistview-documentation-accessibility-listitem--docs#suppression-of-secondary-info-and-actions) on the leading action to fully suppress it. If you have the capacity to sneak that in now with this PR that'd be great!

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -829,6 +850,25 @@ const TrailingVisual: React.FC<TreeViewVisualProps> = props => {
...
+
+const LeadingAction: React.FC<TreeViewVisualProps> = props => {
+  const {isExpanded, leadingActionId} = React.useContext(ItemContext)
+  const children = typeof props.children === 'function' ? props.children({isExpanded}) : props.children
+  return (
+    <>
+      <div className="PRIVATE_VisuallyHidden" aria-hidden={true} id={leadingActionId}>
+        {props.label}
+      </div>
+      <div className="PRIVATE_TreeView-item-leading-action" aria-hidden={true}>
```
siddharthkp, line 864:
```suggestion
      <div className="PRIVATE_TreeView-item-leading-action" aria-hidden={true} tabIndex="-1">
```

packages/react/src/TreeView/TreeView.tsx
siddharthkp: See also Eric's comment about tabIndex=-1: https://github.com/primer/react/pull/4546#issuecomment-2107683449

siddharthkp:
> Ultimately, we'd need it and also tabindex="-1" applied to [any interactive element](https://ui.githubapp.com/storybook/?path=%2Fdocs%2Frecipes-nestedlistview-documentation-accessibility-listitem--docs#suppression-of-secondary-info-and-actions) on the leading action to fully suppress it. If you have the capacity to sneak that in now with this PR that'd be amazing!

Just noticed the changes made since my last review, looks like that would have to live in the application instead of primer/react because primer does not control the contents of the action :(

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -459,7 +463,7 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
           id={itemId}
           role="treeitem"
           aria-labelledby={labelId}
-          aria-describedby={`${leadingVisualId} ${trailingVisualId}`}
+          aria-describedby={`${leadingActionId} ${leadingVisualId} ${trailingVisualId}`}
```
siddharthkp, line 466: Don't think the action should be used to describe the item, especially when the action itself is not accessible (aria-hidden + tab-index=-1) 🤔 

cc @ericwbailey for confirmation

ericwbailey: `aria-hidden` should hypothetically override this, but yes, I think we should remove it. 

We'll likely want to include a hint down the line in the `aria-label` about state, such as "draggable". That's something we'll be researching with TetraLogical.

packages/react/src/TreeView/TreeView.tsx
ayy-bc: Yeah, the drag trigger component that we use has a wrapper around the `IconButton` component

https://github.com/github/github/blob/master/ui/packages/drag-and-drop/components/SortableListTrigger.tsx

ayy-bc: And I assume its the same as using the leading/trailing visual so the consumer (us) can pass the `tabindex=-1` like we do when consuming TreeView in NestedListView.

packages/react/src/TreeView/TreeView.tsx
siddharthkp: @ayy-bc Final thing for you to look at :)

siddharthkp: > @siddharthkp I was just following what LeadingVisual and TrailingVisual in this case. Looking at it again, I think removing this makes sense and we can add the required labels to the components itself when we pass it.
> I can remove it or you can do so, whatever you prefer (lmk).

Yes please, I think actions are different

ayy-bc: @siddharthkp pushed the changes

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -45,6 +45,7 @@ const ItemContext = React.createContext<{
   setIsExpanded: (isExpanded: boolean) => void
   leadingVisualId: string
   trailingVisualId: string
+  leadingActionId: string
```
siddharthkp, line 48: @ayy-bc Don't think we need this anymore either, right?

siddharthkp:
@langermank Can I get a review for the snapshots

mperrotti (approved):

PR #4547, opened 2024-05-01
VRT: Mock avatars to return stable images

siddharthkp:
#### Context:

To make our stories realistic and useful, we use examples of real collaborations in our stories (instead of fake or generated data). This is very nice!

This also means when one of the collaborators updates their avatar, some of stories also update. This is also very nice!

But, this also means that our visual tests can fail because they have the old avatar in snapshots to compare. This is not nice at all! 😓  Most recently reported in https://github.com/primer/react/pull/4532#issuecomment-2087903001

Screenshot of before and after of ActionList when someone changes their avatar

#### Solution:

Added a route mock for our e2e tests (visual tests) that always returns the same image that is safe for visual comparisons. Chose the primer avatar for this. (could have also used a static image, but this feels as good)

Alternatives considered:
- Use static images in storybook instead of real avatars. Did not chose this option because it's easy to forget in future stories without linting/checks. Also, I personally like that our stories have real avatars :) 

### Rollout strategy

- [x] None; affects only tests

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack (approved):

langermank (commented):
😂

langermank (approved):

PR #4552, opened 2024-05-02
experimental/SelectPanel: Don't render form if nested inside form

siddharthkp:
> [!NOTE]  
> If we want, we can wait until we figure out the semantics for https://github.com/github/primer/issues/3183 to find out if we want to render a `<form>` element at all

#### Bug:

SelectPanel renders a `<form>` element inside a `<dialog>`. We also do allow the use of SelectPanel inside with FormControl (which is expected to be inside a `<form>` already)

This would create invalid DOM nesting of `<form>` nested inside `<form>`. react-dom throws a validation warning on the console

#### Fix:

If SelectPanel is nested inside FormControl, it will now render a `<div>` instead of `<form>`.

Note: If someone is using SelectPanel in a custom form without FormControl, this PR would not be able to catch that scenario.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


packages/react/src/drafts/SelectPanel2/SelectPanel.examples.stories.tsx
```diff
@@ -861,20 +859,22 @@ export const WithinForm = () => {
     <>
       <h1>Within Form</h1>
 
-      <FormControl>
-        <FormControl.Label>SelectPanel within FormControl</FormControl.Label>
-        <SelectPanel title="Choose a tag" selectionVariant="instant" onSubmit={onSubmit}>
-          <SelectPanel.Button leadingVisual={TagIcon}>{selectedTag || 'Choose a tag'}</SelectPanel.Button>
+      <form>
```
JeffreyKozik, line 862: It doesn't seem like this is triggering the error

> `<form>` cannot appear as a descendant of `<form>`.

Do we know why this is only occurring in Dotcom and not Storybook and if there's a way to achieve parity between the two to avoid future problems?

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -306,8 +312,8 @@ const Panel: React.FC<SelectPanelProps> = ({
               }}
             >
               <Box
-                as="form"
-                method="dialog"
+                as={isInsideForm ? 'div' : 'form'}
```
JeffreyKozik, line 315: Is it possible to write a test for this to ensure we're not getting the error again? Or is it not detectable through testing?

JeffreyKozik (approved):
Thanks for fixing this. As you mentioned, it might be better to not render a `<form>` element for SelectPanel at all, and that would fix this issue if someone had SelectPanel nested within a `<form>` other than FormControl. However, I'm not sure what the other implications of doing so are.

packages/react/src/drafts/SelectPanel2/SelectPanel.examples.stories.tsx
JeffreyKozik: Shouldn't this error have been showing without needing to wrap `<form>` around `<FormControl>`? Do we know why it was occurring in Dotcom and not Storybook and if there's a way to achieve parity between the two to avoid future problems?

joshblack (approved):

JeffreyKozik:
What's the timeline on when this will get deployed? I'm currently trying to resolve this accessibility audit issue: https://github.com/github/accessibility-audits/issues/6410 by converting our custom select component to the SelectPanelv2, but since it's a form within a form it's not working for me. If this gets merged to Primer that would unblock me and allow me to convert our custom component to SelectPanelv2.

siddharthkp:
Hi @JeffreyKozik!

Oops, the same conversation happened in 2 places, copying my message from slack :) 

> (I can merge the PR if you want to be unblocked)
>
> But, I wouldn't recommend migrating to SelectPanel v2 right now. There is a slight change in plans with SelectPanel. Instead of promoting SelectPanel v2 from experimental to stable, we are planning to make accessibility remediations to SelectPanel v1 for better adoption in the quarter starting next week. (see [ADR](https://github.com/github/primer/blob/main/adrs/2024-05-22-select-panel-component.md) for more info)
>
> If that dialog does not use either SelectPanel, I would recommend using SelectPanel v1

JeffreyKozik:
Okay thanks for the clarification! I didn't realize SelectPanel v1 was a recommended option at this time. I'll go ahead and use that then, thanks!

JeffreyKozik:
@siddharthkp are there plans to add the "dialog" version of the selectpanel to v1? For my use case, v1 works fine, but v2 as a dialog is much closer to our existing custom component.

siddharthkp:
Closing this PR because it's stale and we are not working on SelectPanel2 anymore.

PR #4554, opened 2024-05-02
feat(FeatureFlags): loosen feature flag type to accept undefined

joshblack:
This change makes it easier to integrate our `FeatureFlags` component in scenarios where a feature flag value may be `boolean | undefined`.

#### Changed

- The type of a flag has been loosened to `boolean | undefined`

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

- Verify that `FeatureFlags` now accepts a `flags` prop where the values can be `undefined` and they are correctly initialized in `FeatureFlagScope`

### DIFF
--- a/.changeset/grumpy-coats-worry.md
+++ b/.changeset/grumpy-coats-worry.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Loosen feature flag type for experimental FeatureFlags to accept undefined
--- a/packages/react/src/FeatureFlags/FeatureFlagScope.ts
+++ b/packages/react/src/FeatureFlags/FeatureFlagScope.ts
@@ -1,5 +1,5 @@
 export type FeatureFlags = {
-  [key: string]: boolean
+  [key: string]: boolean | undefined
 }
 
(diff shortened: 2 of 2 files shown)


.changeset/grumpy-coats-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
```
siddharthkp, line 2: patch?

```suggestion
'@primer/react': patch
```

siddharthkp (approved):
LGTM!

.changeset/grumpy-coats-worry.md
joshblack: @siddharthkp my thinking for picking `minor` was that this change seemed like it was adding functionality where a type was broadended to accept more values. Right now in our versioning doc we have that as a `minor` over at: https://github.com/primer/react/blob/main/contributor-docs/versioning.md#the-type-of-a-prop-is-broadened 

Let me know if this seems more like a bug fix though or if it'd be worth talking about updating the version doc 👍

PR #4555, opened 2024-05-02
Add `openOnFocus` prop to `AutocompleteInput`

JoyceZhu:
Unblocks github/accessibility-audits#7437

The audit issue notes that when an `AutocompleteInput` is the first field in a form with errors (and thus when it needs to be auto-focused after submission to help the user discover and address errors), the autocomplete menu popping up automatically on focus 1) causes NVDA+Firefox to miss reading out an associated error message/`FormControl.Validation` even when linked via `aria-describedby` and 2) visually obscures the error message, which makes discovering what went wrong harder.

Offering a prop to optionally disable the menu appearing on focus improves these two users experiences; the menu will still render when the user types in the autocomplete's input field.

#### New

- Adds an `openOnFocus` prop which defaults to `true` (current behavior) to `AutocompleteInput`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
I wonder if we'd want up/down arrow keys to open the menu if this prop is used. For example, if you set [`openOnFocus` to `false`](https://primer-3b8d52ba54-13348165.drafts.github.io/storybook/?path=/story/components-autocomplete--default&args=openOnFocus:!false) and clear the input field, pressing up/down arrow keys will select items from the menu but doesn't show the menu itself. I'm wondering if we'd want it to show the menu by default depending on when either up/down key is used?

JoyceZhu:
> I wonder if we'd want up/down arrow keys to open the menu if this prop is used

I think this makes sense as behavior! Let me implement that tomorrow!

packages/react/src/Autocomplete/AutocompleteInput.tsx
```diff
@@ -38,10 +54,12 @@ const AutocompleteInput = React.forwardRef(
...
       event => {
-        onFocus && onFocus(event)
-        setShowMenu(true)
+        if (openOnFocus) {
+          onFocus?.(event)
+          setShowMenu(true)
+        }
       },
-      [onFocus, setShowMenu],
+      [onFocus, setShowMenu, openOnFocus],
```
kendallgassner, line 62:
```suggestion
      [onFocus, setShowMenu, openOnFocus, setShowMenu],
```

packages/react/src/Autocomplete/AutocompleteInput.tsx
```diff
@@ -83,8 +101,11 @@ const AutocompleteInput = React.forwardRef(
           setInputValue('')
           inputRef.current.value = ''
         }
+        if (!showMenu && ARROW_KEYS_NAV.has(event.key) && !event.altKey) {
+          setShowMenu(true)
+        }
       },
-      [inputRef, setInputValue, setHighlightRemainingText, onKeyDown],
+      [inputRef, setInputValue, setHighlightRemainingText, onKeyDown, showMenu],
```
kendallgassner, line 108:
```suggestion
      [inputRef, setInputValue, setHighlightRemainingText, onKeyDown, showMenu, setShowMenu],
```

TylerJDev (approved):
LGTM!

kendallgassner (approved):

PR #4558, opened 2024-05-03
chore: add hydro analytics to storybook

joshblack:
Closes https://github.com/github/primer/issues/3269

Add support for hydro analytics to storybook in production deployments. This is only used to get page views of different URLs in storybook.

#### Changed

- Update storybook config to include hydro tags when deployed to production

### Rollout strategy

This is a change to our storybook

### Testing & Reviewing

- Verify that hydro events are being emitted in preview (temporarily enabled for this PR)

I'll make sure to remove support for page views in previews after we verify that it is wired up correctly 👍

### DIFF
--- a/packages/react/.storybook/main.js
+++ b/packages/react/.storybook/main.js
@@ -1,3 +1,7 @@
+'use strict'
+
+const {DEPLOY_ENV = 'development'} = process.env
+
 /**
  * @type {import('@storybook/core-common').StorybookConfig}
  */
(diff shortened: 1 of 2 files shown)


joshblack:
Verified analytics are being sent for the preview over in: https://github.slack.com/archives/C02NUUQ9C30/p1714755458846759?thread_ts=1714728330.477579&cid=C02NUUQ9C30

Updated it so that we now will only report in production and **not** previews 👍

PR #4567, opened 2024-05-07
Use dynamic height and width for dialogs

dgreif:
Dialog v2 in right/left position currently flows off the top of the screen on iOS devices. This is because we use `100vh`, which can be _larger_ than the visible view port. I'd like to fix this so dialogs never get cut off on iOS.

I updated both Dialogs to use `100dvh` and `100dvw` instead of `100vh` and `100vw`. These are supported on Safari 15+, which is all that we officially support on dotcom today.

| Before | After |
|--------|--------|
| A right positioned dialog with the header and footer cut off on iOS | A right positioned dialog with header and footer fully in view | 

#### Changed

Use dynamic view height/width for dialogs. This allows available visible space to be properly computed on iOS devices.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge

### DIFF
--- a/.changeset/tall-forks-bathe.md
+++ b/.changeset/tall-forks-bathe.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Use dynamic view height/width for dialogs. This allows available visible space to be properly computed on iOS devices.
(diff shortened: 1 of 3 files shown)


siddharthkp (approved):
Looks good to me!

PR #4571, opened 2024-05-08
Fix replacement URL for React getting started

MrChrisW:
#### Changed

The existing link shown with the deprecation banner on https://primer.style/react/getting-started points to https://primer.style/design/guides/development/react which is incorrect and results in a 404 not found.

This updates the replacement URL to the correct value of https://primer.style/guides/react ✨ 

### Rollout strategy

### Merge checklist


joshblack (approved):
Thanks for this!

PR #4572, opened 2024-05-08
`TreeView`: Align tree item toggle and visual icons to top of item

iansan5653:
The vast majority of the time, `TreeView` items are only one line and the content is truncated to fit. However, they can potentially wrap, as in https://github.com/github/collaboration-workflows-flex/issues/902 (internal issue). In this case, we want the chevron to still remain locked to the top of the item, rather than vertically centered.

This is slightly challenging because the height of items changes depending on the input mode, expanding automatically for touch controls. It wasn't complicated to move the `min-height` to a CSS variable and calculate a correct padding based on that: `min-height / 2 - icon-height`.

There is no visual change here for the non-wrapped item.

#### Changed

- Always align `TreeView` chevron icon to top of item

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TreeView/TreeView.tsx
```diff
@@ -163,9 +167,9 @@ const UlBox = styled.ul<SxProp>`
 
   .PRIVATE_TreeView-item-toggle {
     grid-area: toggle;
-    display: flex;
```
broccolinisoup, line 166: I think we should still keep the `display: flex` and `justify-content: center;` to keep the icon centred horizontally. The VRT are failing due to this.

broccolinisoup (commented):
Hey @iansan5653 👋🏻 Thanks for the PR! Left a comment about horizontal centring. The rest of it looks good to me. We should also get a designer review on this though -  tagging @mperrotti 🙌🏻

packages/react/src/TreeView/TreeView.tsx
iansan5653: Yep, working on that now :+1:

mperrotti:
@iansan5653 - can we add a story that demonstrates this change? Kind of like this...
Screenshot 2024-05-09 at 1 27 01 PM


Alternatively, we could update all existing stories to have an expandable node that has 2 lines of content. That seems unnecessary though...

broccolinisoup:
> Added a story. Now that I'm looking at it I'm thinking maybe we should top-align the leading and trailing visuals as well? I'm not totally sure how I feel about this approach given the leading/trailing visual appearance... maybe overriding the styling for the specific use case of sub-issues is actually better 🤔 But we don't have any official style overriding support, requiring us to reach in and select `PRIVATE` classes, which is hacky.

It makes sense o me to align all to the top - not the same case but we took a [similar approach for pageheader](https://primer-95439128bf-13348165.drafts.github.io/storybook/?path=/story/drafts-components-pageheader-devonly--large-variant-with-multiline-title) as well that we align all visuals to the top when the title is multiline

@mperrotti thoughts?

JelloBagel:
I updated the leading/trailing visual to be top aligned as well: [185b613](https://github.com/primer/react/pull/4572/commits/185b6131b92f6f3730a71ce9d15c6159502408db). I removed the `min-height` to properly align the visuals with the text but kept the same visual height by adding padding to the top and bottom of the text.

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -200,11 +210,16 @@ const UlBox = styled.ul<SxProp>`
     overflow: hidden;
     white-space: nowrap;
     text-overflow: ellipsis;
+    align-self: center;
```
broccolinisoup, line 213: If we remove the flex start and line height above, I don't think we would need this. Let me know if I am missing any context here though!

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -187,10 +193,14 @@ const UlBox = styled.ul<SxProp>`
   .PRIVATE_TreeView-item-content {
     grid-area: content;
     display: flex;
-    align-items: center;
+    align-items: flex-start;
```
broccolinisoup, line 196: I am not really sure if we need the flex-start here in the content 🤔 Same as line-height just above.

broccolinisoup (commented):
Hello! I left a few comments/questions. Let me know if there is anything I misunderstood or if there is anything I can clarify about my questions/comments 🙌🏻

mperrotti (approved):
These changes make sense and seem safe to me.

Issue #4587, opened 2024-05-13
`Stack` `gap` is overridden on wider layouts


iansan5653:
### Description

When `Stack` is rendered on a screen greater than `87.5rem` wide, the gap between items is always overridden to `1rem` - even if the `gap` is set to `none` or `condensed`.

### Steps to reproduce

```jsx
<Stack gap="none">
  <span>A</span>
  <span>B</span>
</Stack>
```

### Version

36.17.0

PR #4589, opened 2024-05-13
refactor(Stack): write to single custom property for gap changes

joshblack:
Closes https://github.com/primer/react/issues/4587

Update the strategy for setting gap in `Stack` to use a single custom property, `--stack-gap`, that is written to a stack gap value (e.g. `--stack-gap-normal`). This is an attempt at refactoring this code so that we are writing to this value across breakpoints so that tracing down why something is happening at a breakpoint is only one jump (finding where the property is set) instead of two jumps (finding the variable for the breakpoint, finding the value that is being set for it). Curious to hear if this approach seems more clear if they are the same 👀 

#### Changed

- Update how gap is set on `Stack`
- Address gap issues on larger breakpoints

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- Verify that gap is being set correct across breakpoints, for example if `none` is set then it should be `none` across breakpoints


packages/react/src/Stack/Stack.tsx
```diff
@@ -4,15 +4,16 @@ import type {ResponsiveValue} from '../hooks/useResponsiveValue'
 import {getResponsiveAttributes} from '../internal/utils/getResponsiveAttributes'
 
 const StyledStack = styled.div`
-  --Stack-gap-whenRegular: var(--stack-gap-normal, 16px);
-  --Stack-gap-whenNarrow: var(--stack-gap-normal, 16px);
-  --Stack-gap-whenWide: var(--Stack-gap-whenRegular);
+  --stack-gap-none: 0;
+  --stack-gap-condensed: 0.5rem;
+  --stack-gap-normal: 1rem;
+  --stack-gap-spacious: 2rem;
```
langermank, line 10: The only thing that feels confusing to me is that we're resetting these values that already exist globally 🤔 what if we update those values in primer/primitives at some point?

joshblack: Whoops 🤦‍♂️ I completely forgot about that, let me just use these directly and I'll add in the fallback each time I'm using it 👍

joshblack:
bump @langermank when you get a sec 👀

langermank (approved):
🚀

PR #4599, opened 2024-05-16
Check integration test results

siddharthkp:
### Context

I'd like to fetch and report the result of `golden-jobs-results` on the PR as a CI check and comment.

Strategy 1: Pull (implemented here)
1. Check if there is a PR related to current branch 
2. Check if the CI workflows have run
3. Check if `golden-jobs-results` has completed yet
4. Check results of `golden-jobs-results`

Strategy 2: Push (probably better!)
1. Inform PR has been created (based on label)
2. (maybe) inform `golden-jobs-results` has started/expected
3. Inform `golden-jobs-results` has completed with results (based on label)

Questions I need to think more about:
1. What triggers this workflow?
2. Can this workflow be re-triggered manually?
3. Is it workflow required or just recommended?
4. What happens when this workflow fails for reasons not related to your PR/change
5. Pushing another commit should invalidate the results, right?

### Rollout strategy

- [x] None; workflow only change.

### DIFF
--- a/.github/workflows/report-integration-results.yml
+++ b/.github/workflows/report-integration-results.yml
@@ -19,14 +19,14 @@ jobs:
           client-id: ${{ vars.PRIMER_ISSUE_TRIAGE_APP_CLIENT_ID_FOR_GITHUB }}
           client-secret: ${{ secrets.PRIMER_ISSUE_TRIAGE_APP_CLIENT_SECRET_FOR_GITHUB }}
           installation-id: ${{ vars.PRIMER_ISSUE_TRIAGE_APP_INSTALLATION_ID_FOR_GITHUB }}
-
       - name: Get workflow run
         uses: actions/github-script@v7
         env:
           # BRANCH: "prc-integration-test-for-${{github.event.pull_request.number}}"
           BRANCH: "prc-integration-test-for-4484"
           WORKFLOW_ID: "primer-react-pr-test.yml"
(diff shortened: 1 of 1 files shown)


.github/workflows/report-integration-results.yml
```diff
@@ -0,0 +1,101 @@
...
+          github-token: ${{ secrets.TEMP_SID_PAT }}
+          script: |
+            const { BRANCH } = process.env;
+            try {
+              const { data } = await github.rest.repos.getBranch({
+                owner: 'github',
+                repo: 'github',
+                branch: BRANCH
+              });
+              core.setOutput('branch', BRANCH);
```
joshblack, line 41: Instead of us calling out to the repo directly, do you think it would be possible for us to use repository dispatch to coordinate between github/github and primer/react? The advantage of this would be that we wouldn't need to have a token/app that has permissions to github/github in CI.

siddharthkp: Definitely possible! 👀 Hopefully we can make them as granular. 

> The advantage of this would be that we wouldn't need to have a token/app that has permissions to github/github in CI.

We would still need a token that can do the first repository dispatch to trigger `primer-react-pr-test` workflow (not included in this PR). But, that would need less permissions which is definitely better

siddharthkp: > The advantage of this would be that we wouldn't need to have a token/app that has permissions to github/github in CI.

How worried should we be about this?

PR #4611, opened 2024-05-20
[ADR] docs: add live-regions adr

joshblack:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/contributor-docs/adrs/adr-020-live-regions.md
+++ b/contributor-docs/adrs/adr-020-live-regions.md
@@ -1,0 +1,162 @@
+# ADR 020: Live Regions
+
+## Status
+
+| Stage    | Status      |
+| -------- | ----------- |
+| Approved | <!-- ✅ --> |
+| Adopted  | <!-- 🚧 --> |
+
+## Context
+
+There are several components in Primer React that make use of live regions.
+However, the library does not have a central pattern for making accessible live
+region announcements. This ADR documents the decision to use a central pattern
+for live region announcements that can be used across Primer and GitHub.
+
+ARIA Live regions fall into several buckets that we may want to use as component
+authors:
+
+- [ARIA live region roles](https://www.w3.org/TR/wai-aria-1.2/#live_region_roles), such as `alert`, `log`, or `status`
+- [The `aria-live` attribute](https://www.w3.org/TR/wai-aria-1.2/#aria-live) to turn an element into a live region
+
+In components, we see the following scenarios in Primer React:
+
+- Announce the contents of an element when it is rendered or on page load, like when a spinner is displayed or a form is submitted
+- Announce the changes to the content of an element, like when the count inside of a
+  button is incremented
+- Announce a message programmatically, such as the number of results for a query
+
+Currently, contributors may reach for roles such as `alert` or `status` to
+achieve these scenarios. They may also add `aria-live` to an element explicitly.
+However, both of these approaches do not announce consistently across screen
+readers.
+
+### Links & Resources for ARIA Live regions
+
+- https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-1/
+- https://www.scottohara.me/blog/2022/02/05/are-we-live.html
+
+## Decision
+
+In order to have a common interop point for live region announcents, Primer React will
+make use of a `live-region` custom element from `@primer/live-region-element`.
+This package will be included and published from the `primer/react` repo.
+
+The custom element exposes a way to make announcements that can be used
+across frameworks. This makes it useful not only for Primer but GitHub as a
+whole.
+
+In addition, `@primer/react` will leverage and export the following helpers for
+use within Primer React and GitHub:
+
+- `announce()` and `announceFromElement()`, programmatic helpers for making
+  announcements
+- The `Status` component to correspond with `role="status"`
+- The `Alert` component to correspond with `role="alert"`
+
+Within `@primer/react`, we should lint against usage of `aria-live` and the
+corresponding roles (if possible) and suggest using these alternatives instead.
+
+### Impact
+
+This decision will impact existing usage of `aria-live`, `role="status"`, and
+`role="alert"` for the components listed below. These components will need to be
+updated to use the new approach, using one of the following approaches:
+
+- Use `announce()` or `announceFromElement()`
+- Use `Status` or `Alert`
+
+#### Instances of `aria-live`
+
+- InlineAutocomplete
+- InputValidation
+- SelectPanel
+- TreeView
+
+### Instances of `role="status"`
+
+- LiveRegion
+- TreeView
+- ActionList examples
+
+### Instances of `role="alert"`
+
+None
+
+### Instances of `LiveRegion`
+
+- DataTable
+
+## Examples
+
+### Announce on page load
+
+```tsx
+import {Status} from '@primer/react'
+
+function ExampleComponent() {
+  return <Status>Example page load message</Status>
+}
+```
+
+### Announce on content change
+
+```tsx
+import {Status} from '@primer/react'
+import {useState} from 'react'
+
+function ExampleComponent() {
+  const [count, setCount] = useState(0)
+  return (
+    <button
+      type="button"
+      onClick={() => {
+        setCount(count + 1)
+      }}
+    >
+      <Status>Count {count}</Status>
+    </button>
+  )
+}
+```
+
+### Announce programmatically
+
+```tsx
+import {announce} from '@primer/react'
+import {useState} from 'react'
+
+function ExampleComponent() {
+  const [results, setResults] = useState(data)
+
+  return (
+    <>
+      <input
+        type="text"
+        onChange={event => {
+          const filteredResults = data.filter(item => {
+            /* ... */
+          })
+          setResults(filteredResults)
+          announce(`${filteredResults.length} results available`)
+        }}
+      />
+      {/* ... */}
+    </>
+  )
+}
+```
+
+## Alternatives
+
+## Unresolved questions
+
+- What should happen if a component makes excessive announcements due to a
+  component that it renders? Is it possible to "group" announcements?
+- For programmatic announcements, how do we interop with existing proposals in
+  this space for `ariaNotify`?
+- When should something use an ARIA Live region role component versus just using
+  the helper functions?
+
+## FAQ


khiga8 (commented):
Thank you for writing this!

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,165 @@
...
+This package will be included and published from the `primer/react` repo.
+
+The custom element exposes a way to make announcements that can be used
+across frameworks. This makes it useful not only for Primer but GitHub as a
+whole.
+
+In addition, `@primer/react` will leverage and export the following helpers for
+use within Primer React and GitHub:
+
+- `announce()` and `announceFromElement()`, programmatic helpers for making
```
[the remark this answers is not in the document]
joshblack: @khiga8 I think groups outside of Primer React could use `@primer/live-region-element` directly! This would just be as a convenience, but happy to make it so that folks only import/use `@primer/live-region-element` if that would be clearer.

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,165 @@
...
+`role="alert"` for the components listed below. These components will need to be
+updated to use the new approach, using one of the following approaches:
+
+- Use `announce()` or `announceFromElement()`
+- Use `Status` or `Alert`
+
+In addition, we should make sure that `<live-region>` is successfully included
+in GitHub.
+
+#### Instances of `aria-live`
```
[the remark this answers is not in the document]
joshblack: Sure! I can do that 👍

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,165 @@
...
+
+None
+
+### Instances of `LiveRegion`
+
+- DataTable
+
+## Examples
+
+### Announce on page load
```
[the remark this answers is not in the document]
joshblack: It definitely could and I think would be a good reason for it to _not_ announce by default since we don't want it to announce on load except in rare circumstances, right?

And happy to update to "Announce when content is shown" 👍 There is this special case of page load that can be hard to distinguish between which I think is the only thing to consider. In other words, if a component is rendered it's hard to tell if it's the first render of the page or just the component itself being rendered for the first time.

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Alternatives
+
+## Unresolved questions
+
+- What should happen if a component makes excessive announcements due to a
+  component that it renders? Is it possible to "group" announcements?
+- For programmatic announcements, how do we interop with existing proposals in
+  this space for `ariaNotify`?
+- When should something use an ARIA Live region role component versus just using
```
TylerJDev, line 162: Maybe when the region exists in a `Dialog`. This is because, in most screen readers / browsers, the live region wont announce unless it's directly inside of the dialog. Since our current utility doesn't handle this, I'd probably recommend using a live region directly. The only tricky thing about this is, if the live region doesn't exist at page load, there's the chance it won't announce.

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,165 @@
...
+  )
+}
+```
+
+## Alternatives
+
+## Unresolved questions
+
+- What should happen if a component makes excessive announcements due to a
+  component that it renders? Is it possible to "group" announcements?
```
TylerJDev, line 159: I guess this one depends. Would we weigh this based on frequency of announcements, or repetitiveness? (e.g. announcing "loading" multiple times due to multiple spinners being rendered on the page at once).

contributor-docs/adrs/adr-020-live-regions.md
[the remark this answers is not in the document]
khiga8: Yes I think that would make it more clear :)

contributor-docs/adrs/adr-020-live-regions.md
joshblack: Great points, maybe both? Would be great to find a reference page that has this problem and see what we could do to make it obvious when something is firing a lot 😅 

Thankfully some of these announcements will just be dropped if they are announced quickly but would be great to highlight the fundamental problem that is going on here.

One option down the road could be to create "grouping" live regions that could batch up messages. Basically something like:

```tsx
<StatusGroup>
  <SomeComponent />
  <SomeComponent />
  <SomeComponent />
</StatusGroup>
```

Where `SomeComponent` renders `Status`. This grouping component could then collect messages and announce them together (or could expose an api to "summarize") but this is way down the line and may not even be useful 😂

contributor-docs/adrs/adr-020-live-regions.md
joshblack: @TylerJDev one nice thing about `announce()` or `announceFromElement()` is that they accept a `from` argument that will try to find a live region and create one if it doesn't exist. The logic for this is over at: https://github.com/primer/live-region-element/blob/main/packages/live-region-element/src/query.ts#L3 so that if the `from` element is within a  `dialog` the live region will attach itself to that dialog. `Status` and `Alert` will also do this automatically if one doesn't exist.

Either way, the way these utilities work is that they will make sure a live region exists _before_ making the announcement and hopefully that helps improve the robustness here 🤞

joshblack:
cc @primer/engineer-reviewers if you have a sec to check this out! Curious to hear what you all think about this approach 👀

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,233 @@
...
+The custom element exposes a way to make announcements that can be used
+across frameworks. This makes it useful not only for Primer but GitHub as a
+whole. The `@primer/live-region-element` exports two helpers to use for making
+announcements: `announce()` and `announceFromElement`. Both helpers can be used
+when working in Primer and by teams at GitHub.
+
+In addition, `@primer/react` will leverage and export the following helpers for
+use within Primer React and GitHub:
+
+- The `Status` component to correspond with `role="status"`
```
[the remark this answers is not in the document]
joshblack: Updated! Thanks for the suggestion

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,233 @@
...
+### Links & Resources for ARIA Live regions
+
+- https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-1/
+- https://www.scottohara.me/blog/2022/02/05/are-we-live.html
+
+## Decision
+
+In order to have a common interop point for live region announcements, Primer React will
+make use of a `live-region` custom element from `@primer/live-region-element`.
+This package will be included and published from the `primer/react` repo.
```
mperrotti, line 47: So we're getting rid of `@primer/live-region-element` and moving it to this repo?

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,233 @@
...
+announcements: `announce()` and `announceFromElement`. Both helpers can be used
+when working in Primer and by teams at GitHub.
+
+In addition, `@primer/react` will leverage and export the following helpers for
+use within Primer React and GitHub:
+
+- The `AriaStatus` component to correspond with `role="status"`
+- The `AriaAlert` component to correspond with `role="alert"`
+
+Within `@primer/react`, we should lint against usage of `aria-live` and the
```
mperrotti, line 61: I'm down for this linting rule, but let's make sure we create an issue so we don't forget to do it.

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,233 @@
...
+
+#### Instances of `aria-live="assertive"`
+
+- InlineAutocomplete
+
+### Instances of `role="status"`
+
+- LiveRegion
+- TreeView
+- ActionList examples
```
mperrotti, line 98: I think Spinner would use `role="status"` too once [this PR](https://github.com/primer/react/pull/4140) is merged

```suggestion
- ActionList examples
- Spinner
```

mperrotti (commented):
Really like the direction this is headed. I'm a little stumped on the best way to handle announcement grouping, but I'd be happy to participate in a pairing session to help us figure it out.

contributor-docs/adrs/adr-020-live-regions.md
mperrotti: @owenniblock's API suggestion is interesting. Would the props be the same for when we don't expect a bunch of announcements to be rendered? Or is this a special API for when consumers think we might build up a bunch of announcements?

contributor-docs/adrs/adr-020-live-regions.md
joshblack: Basically this would mean that it wouldn't live in it's own repo and instead would be under `packages/live-region-element` as another package in the monorepo 👀 It would still be published as the `@primer/live-region-element` package.

contributor-docs/adrs/adr-020-live-regions.md
joshblack: Agreed!

contributor-docs/adrs/adr-020-live-regions.md
```diff
@@ -0,0 +1,234 @@
...
+function ExampleComponent() {
+  const [count, setCount] = useState(0)
+  return (
+    <button
+      type="button"
+      onClick={() => {
+        setCount(count + 1)
+      }}
+    >
+      <AriaStatus>Count {count}</AriaStatus>
```
khiga8, line 141: If a page renders with `<ExampleComponent />`, would `Count 0` be announced, or would we only hear announcements starting from the content change?

khiga8: Based on your note above, it sounds like `Count 0` would be announced.

I'm curious what markup you would recommend if we want to visually show `Count 0` on render and don't want it announced, but still want to announce/visually convey subsequent changes 🤔 

Following the guidance in:
> In cases where they should always be present, then the first message passed to the
> component should be an empty string. Changes to the content of the component
> will trigger subsequent announcements.

it sounds like we'd need to do something like:

```.tsx
  Count {count}
  <VisuallyHidden><AriaStatus></AriaStatus></VisuallyHidden>
```
and manage both separately. This feels kind of messy to me, as opposed to just using `role="status"` which would work for this particular scenario 🤔

contributor-docs/adrs/adr-020-live-regions.md
khiga8: Does this mean that PVC consumers could also pull this element into their projects?

contributor-docs/adrs/adr-020-live-regions.md
joshblack: > Based on your note above, it sounds like Count 0 would be announced.

Yes, you're spot on 👍 

The core challenge to me is that we can't distinguish between page load and a component rendering when deciding to make an announcement. Sometimes we want this component to announce when it's rendered and sometimes not (in particular during page load).

I think there are two ways to approach this:

1. Opt-in to announce on first render
2. Opt-out of announcing first render

So the `AriaStatus` would pick one as the default behavior and provide a prop to opt-out of that default. Based on what you shared, it seems like it'd be worth treating Option 1 as the default but I'm curious what you would say 👀

contributor-docs/adrs/adr-020-live-regions.md
owenniblock: @mperrotti I hadn't really thought that far forward, just had an idea and threw it out there 😅 

I guess those props would be optional and if there's no `announcement-id` then it assumes each item should be announced individually.

We could also split out the announcement into its own tag (something like):

```
<Announce  announcement-id="my-id" announcement-template="{n} workflows loading" announcement-frequency="2000">
   <Spinner />
</Announce>
```

Or maybe this API could form part of `<AriaStatus>` if that gains tractions.

contributor-docs/adrs/adr-020-live-regions.md
khiga8: Oh this is so tricky 😅. It could be unexpected for these React components to be called `AriaStatus` and `AriaAlert` but behave slightly different from how these native roles behave, but then again.. most people are not familiar with the nuances around how the native role behaves anyways.. I feel like either is fine as long as we make the behavior crystal clear to the consumer. Would it be overkill to make it a required prop? 🤔  

I would also be interested in @patrickhlauke thoughts, especially as he's also reviewing Primer components/docs and might have insights around what we designate as default behavior, and how it gets communicated to Primer consumers.

Another question - when a component is rendered with a `hidden` attribute, and then shown by removing the `hidden` attribute, is that considered "first render"? Or does this pattern usually not occur in React since things are typically conditionally rendered?

joshblack: > It could be unexpected for these React components to be called AriaStatus and AriaAlert

Totally agreed, thanks for bringing that up. Also wanted to make sure it's clear that if these abstractions aren't particularly helpful (especially compared to directly using `role="status"`) always happy to pursue a different route 👍 

Was revisiting what @patrickhlauke posted over at: https://github.com/w3c/aria/issues/2154 for this since it looks like we're running into this behavior where announcements are made only on _update_ and we are running into situations where we want to announce on insertion (such as a spinner).

With this in mind, we could have the default be what the corresponding role has:

- `status` will announce on update, by default
- `alert` will announce on insert, by default

And for `status` specifically we could have a prop like `announceOnShow` or similar that would provide that first announcement. If we prefer to just keep the roles mapped to `role="alert"` and `role="status"` we could also consider having a dedicated component just for this case, such as `Announce`, that will just announce its contents and that way we avoid having to make these kinds of considerations.

---

For the `hidden` question, great point. None of the components proposed are taking that into account and we could 100% add that in based on what you'd recommend. For example, each component can accept `hidden` and we only announce when `hidden={false}`.

contributor-docs/adrs/adr-020-live-regions.md
joshblack: Yeah! Definitely @khiga8

contributor-docs/adrs/adr-020-live-regions.md
joshblack: Made these updates over at: https://github.com/primer/react/pull/4673 FYI @khiga8

mperrotti (approved):

siddharthkp (approved):

khiga8 (approved):
Looking good! We can revisit the "Unresolved questions" as we learn along the way 🚀

TylerJDev (approved):

PR #4612, opened 2024-05-20
docs(versioning): update versioning guide with css custom property changes

joshblack:
Saw this scenario come up recently and wanted to make sure it was documented with respect to corresponding semver change 👀 

#### Changed

- Add section to versioning guide around changing CSS Custom Properties for a component

### Rollout strategy

This is a change to project documentation

### Testing & Reviewing

I would love to get your feedback on this entry in the versioning doc. Here are some questions if none come to mind to get things going 👀 

- What do you think of this decision? Is it too broad? Too specific?
- Are there any cases to changing a CSS Custom Property that this entry misses?
- Do you agree or disagree with the decision for the semver bump annotation for this type of change?

### DIFF
--- a/contributor-docs/versioning.md
+++ b/contributor-docs/versioning.md
@@ -13,6 +13,7 @@
   - [The type of a prop is broadened](#the-type-of-a-prop-is-broadened)
   - [The type of a prop is narrowed](#the-type-of-a-prop-is-narrowed)
   - [The `display` property used for the container of `children` is changed](#the-display-property-used-for-the-container-of-children-is-changed)
+  - [A component changes its usage of a CSS Custom Property](#a-component-changes-its-usage-of-a-css-custom-property)
   - [A component includes a landmark role](#a-component-includes-a-landmark-role)
   - [A component no longer includes a landmark role](#a-component-no-longer-includes-a-landmark-role)
   - [The element onto which props are spread is changed](#the-element-onto-which-props-are-spread-is-changed)
@@ -62,6 +63,7 @@ For a full list of releases, visit our [releases](https://github.com/primer/reac
 |               | The version of a dependency is increased to a newer minor or patch version                                                                    | `minor`             |
 |               | The version of a dependency is increased to a newer major version                                                                             | `major`             |
 | CSS           | [The `display` property used for the container of `children` is changed](#the-display-property-used-for-the-container-of-children-is-changed) | potentially `major` |
+|               | [A component changes its usage of a CSS Custom Property](#a-component-changes-its-usage-of-a-css-custom-property)                             | potentially `major` |
 | Accessibility | [A component includes a landmark role](#a-component-includes-a-landmark-role)                                                                 | potentially `major` |
 |               | [A component no longer includes a landmark role](#a-component-no-longer-includes-a-landmark-role)                                             | potentially `major` |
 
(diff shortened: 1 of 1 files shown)


joshblack:
Wanted to get your thoughts on this @langermank @mperrotti if you have the time! Let me know if I can provide any additional context for this.

Also wanted to include you @lukasoppermann as an FYI in case this was of interest due to overlap with primitives 👀 The work you did with Banner prompted this and I wanted to make sure this was captured in our versioning guide 👍

contributor-docs/versioning.md
```diff
@@ -132,6 +134,67 @@ In this situation, changing the layout of the `button` to `flex` would collapse
...
+
+semver bump: potentially **major**
+
+When a component changes how it uses a CSS Custom Property, the change could be
+considered breaking if the CSS Custom Property is a part of the component's
+public API or participates in a broader system for design tokens. The breakage
+occurs when a consumer of the component is setting the value of this CSS Custom
+Property in some way expecting a change in the component. For example:
+
+```diff
```
mperrotti, line 147: Woah, I didn't know there was a `diff` flag for code blocks. Nice!

contributor-docs/versioning.md
```diff
@@ -132,6 +134,67 @@ In this situation, changing the layout of the `button` to `flex` would collapse
 the whitespace present between the text "Save" and "changes" and would be
 considered a breaking change.
 
+### A component changes its usage of a CSS Custom Property
+
+semver bump: potentially **major**
+
+When a component changes how it uses a CSS Custom Property, the change could be
+considered breaking if the CSS Custom Property is a part of the component's
+public API or participates in a broader system for design tokens. The breakage
```
mperrotti, line 143: > if the CSS Custom Property is a part of the component's public API

Do we have examples of components where we intentionally provide CSS Custom Properties as part of the component's public API? I can't think of a time where I've used a CSS Custom Property as part of the component API.

contributor-docs/versioning.md
```diff
@@ -132,6 +134,67 @@ In this situation, changing the layout of the `button` to `flex` would collapse
...
+// This is not a breaking change if base-size-6 is defined in the current major
+// version of Primer Primitives
+.ExampleComponent {
+-  margin-block: var(--space-small, 0.375rem);
++  margin-block: var(--base-size-6, 0.375rem);
+}
+```
+
+```diff
+// This is a breaking change if base-size-6 is defined in the next major
```
mperrotti, line 185: So this is a breaking change even though `--space-small` and `--base-size-6` come from Primer Primtives _and_ we provided a fallback?

mperrotti (commented):
> What do you think of this decision? Is it too broad? Too specific?

I'm ok with it, but you could also use this logic to make the argument that changing most other CSS properties could also be considered a breaking change. For example: if we add or remove a `position: relative` from  a component's styles, we change the layout logic and stacking context that consumers might have been relying on in their external styles.

---

> Are there any cases to changing a CSS Custom Property that this entry misses?

Not that I can think of.

---

> Do you agree or disagree with the decision for the semver bump annotation for this type of change?

Agree, but I can easily see people forgetting to do the bump.

contributor-docs/versioning.md
joshblack: @mperrotti same, I can't think of a time either.

contributor-docs/versioning.md
joshblack: That's a great point about the fallback. You're totally right that the fallback would still apply so there would not be any breakage in the UI, in this case what I'm trying to think through is if this would be a breaking change when someone anticipated `--space-small` to be the token applied to a component and then it changes to a different token.

Put another way:

- `ExampleComponent` depends on `space-small`
- Consumer uses/leverage/extends `space-small`
- `ExampleComponent` no longer depends on `space-small`, uses `base-size-6`

I'm down either way on this, totally happy to say that the relationship between Component -> token can change and should not be a part of the contract here.

mperrotti: I think that the relationship between Component and token can chance and should not be part of the contract here.

PR #4613, opened 2024-05-20
Trigger onClose when Dialog backdrop is clicked

joshblack:
> [!NOTE]
> This is a re-open of https://github.com/primer/react/pull/4565 which was reverted as part of the last release

Relates to https://github.com/github/primer/issues/2531
Relates to https://github.com/github/repos/issues/9248
Alternative to https://github.com/github/primer/issues/1838

I'd like to allow the v2 Dialog to be closed by clicking on the backdrop overlay. This is currently not supported, though the v1 Dialog and PVC dialog support this behavior. 

Based on [the usage guidelines](https://primer.style/components/dialog#:~:text=To%20escape%20the%20focus%20trap%2C%20hitting%20the%20Esc%20key%20or%20clicking%20on%20the%20backdrop%20will%20close%20the%20dialog), there are some unique cases where clicking the backdrop should _not_ close the dialog. By passing in a new `backdrop` gesture, there is an escape hatch for these unique cases to _ignore_ `backdrop` events on a case-by-case basis.

#### New

Dialog and ConfirmationDialog can now be closed by clicking on the backdrop round the dialog. This will cause `onClose` to be called with a new `'backdrop'` gesture.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack:
wanted to bump you @keithamus when you have a sec 👀 For context, this was the PR for closing when the backdrop is clicked. You had proposed an alternate way for this but I'm not sure what that would look like. If you have a sec, or want to pair, would appreciate it a ton

keithamus:
The web platform will provide `closedby=any` (https://github.com/whatwg/html/pull/10157) as the value for closing via a click outside the dialog, and the `onclose` handler isn't told how the element is closed as there's not a strong use case for being able to differentiate. AIUI the monolith doesn't use the onclose hint that Primer provides and I'm hesitant to extend it especially if it means we're unable to move to the web platform native features which often are more accessible than we could possibly provide (for example allowing for operating system gestures to close the dialog too).

theinterned (approved):
This looks great to me!

I can certainly see @keithamus argument for staying within the API of the web platform, and honestly I can't think of a reason I'd distinguish between a close button, esc key, and a background click. I could potentially see the need for cancel vs confirm type actions. I suspect in web platform terms the event target could be used to make these distinctions? The same could be done in the react world right?

Still it seems like the removal of "gestures" would be a breaking API change and need not hold up this PR?

keithamus (approved):
This LGTM

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -98,9 +98,9 @@ export interface DialogProps extends SxProp {
 
   /**
    * This method is invoked when a gesture to close the dialog is used (either
-   * an Escape key press or clicking the "X" in the top-right corner). The
+   * an Escape key press, clicking the backdrop, or clicking the "X" in the top-right corner). The
```
dwilsonactual, line 101: should this reference to clicking the backdrop be removed since that information isn't actually communicated?

PR #4629, opened 2024-05-28
docs: add example with NavList.Group

khiga8:
Relates to: https://github.com/github/accessibility/issues/6109

#### New

I am starting to familiarize myself with the NavList components. 

I noticed that `NavList.Group` is documented, but there are no examples demonstrating its usage. 

This PR adds a storybook example using `NavList.group`

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

#### 🔈 QUESTION TO REVIEWER!!!!!

Here is the [WithGroup example](https://primer-ee7614f40c-13348165.drafts.github.io/storybook/?path=/story/components-navlist--with-group) that I've newly added.

What feels odd to me is that rather than having two separate lists, we have a single list which renders nested lists.
This strays from the [Primer Rails NavList implementation](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/beta/nav_list/show_more_item), where two separate lists are rendered for two groups... does anyone have insights on whether this markup is expected/accessible? @broccolinisoup I was told you might have insights here :) 

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


broccolinisoup (approved):
>What feels odd to me is that rather than having two separate lists, we have a single list which renders nested lists.
This strays from the [Primer Rails NavList implementation](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/beta/nav_list/show_more_item), where two separate lists are rendered for two groups... does anyone have insights on whether this markup is expected/accessible?

Yes PVC renders groups as 2 different lists whereas PRC has them as list items. The difference isn't intentional like one is better than the other; however both have accessible semantics as far as I understand.

When we revisited ActionList semantics, the original idea was to align with PVC; however the way we render the action lists in PRC is different and trying to match with PVC was going to require drastic changes in the component. Because of this, we explored alternative ways to make the semantics accessible without worrying too much about aligning it with PVC. [This was the PR](https://github.com/github/primer/pull/2787) that we documented the expected group semantics and [this is the render version](https://github.com/github/primer/blob/e71cd4834d127e4fb1dd55a3256e7760b57d9c7e/apis/action-list-api.md#actionlist-with-groups) of the ActionList API both in PRC and PVC.

Let me know if you have any questions/concerns 😊 

Regarding the PR itself, we usually document features under NavList.features.stories but I see we don't have this separation in NavList (Almost all of these components have that structure though). It is fine to have the story in one single story file for now and we can quickly refactor it later. Thanks for pushing this PR!! 🙌🏻

PR #4634, opened 2024-05-30
Introduce TrailingAction to ActionList

khiga8:
Depends on: https://github.com/primer/react/pull/4272

Fixes: https://github.com/github/primer/issues/1966 
Towards: https://github.com/github/accessibility-audits/issues/2942

#### New

This PR introduces a `TrailingAction` feature to the `ActionList` component. This allows for a secondary action to be displayed on the right side of an `ActionList.Item`.

The `ActionList.Item` can be rendered as either an `IconButton` or a `Button` depending on the presence of an `icon` prop. Semantically, the `ActionList.Item` can be either a `button` or an `a` tag.

We implemented this feature according to the agreed upon [proposed API](https://github.com/github/primer/pull/3314/). Kudos to @broccolinisoup for invaluable feedback!

⚠️ While working on this feature, we discovered an _existing_ issue of horizontal overflow occurring when a description is long. Given this is an existing problem, we've filed an issue for that to be investigated/addressed separately - https://github.com/primer/react/issues/4664.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/Item.tsx
```diff
@@ -390,19 +396,18 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
                   </Box>
                   {slots.inlineDescription}
                 </ConditionalWrapper>
-                {
+                {slots.trailingVisual &&
```
khiga8, line 399: @tylerjdev it looks like this change is resulting in the inactive button missing in the snapshots - [Prod version](https://primer.style/react/storybook/?path=/story/components-actionlist-features--inactive-multiselect). 

(Side note: I am confused on if this button is here, just for the purpose of showing this info tooltip.. but that seems out-of-scope for now)

TylerJDev: I don't believe that this was an intentional change. We should be able to revert back to the prod state! I don't think this is needed for `TrailingAction`.

packages/react/src/ActionList/ActionList.features.stories.tsx
```diff
@@ -734,3 +734,45 @@ export const ActionListWithButtonSemantics = () => {
...
+        <ActionList.Item>
+          Item 2 (with link TrailingAction)
+          <ActionList.TrailingAction as="a" href="#" aria-label="Some action 1" icon={ArrowRightIcon} />
+        </ActionList.Item>
+        <ActionList.Item>
+          Item 3<ActionList.Description>This is an inline description.</ActionList.Description>
+          <ActionList.TrailingAction aria-label="Some action 2" icon={BookIcon} />
+        </ActionList.Item>
+        <ActionList.Item>
+          Item 4<ActionList.Description variant="block">This is a block description.</ActionList.Description>
```
khiga8, line 758: Just realized block variant is supported for Description so added this `ActionList.Description variant="block"` example!

@langermank Is the Trailing Action alignment okay as is, or should it be vertically centered?

Screenshot of an ActionItem with a block description, and a TrailingAction. The TrailingAction has a default, start alignment..

TylerJDev: For reference, this is with a "Trailing Visual":

Focused Action List item with a trailing visual to the right of the text, the trailing visual is positioned up vertically

It's similar with the "inactive" button. If we should modify this instance, do we need to adjust the others?

packages/react/src/ActionList/Item.tsx
```diff
@@ -149,10 +159,38 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
+          outline: 'none',
+          border: `2 solid`,
+          boxShadow: `0 0 0 2px ${theme?.colors.accent.emphasis}`,
+        },
+        ':active:not([aria-disabled]):not([data-inactive])': {
+          backgroundColor: `actionListItem.${variant}.activeBg`,
+          color: getVariantStyles(variant, disabled, inactive).hoverColor,
+        },
+        ':hover, :focus, :focus-within': {
+          // Used when `showOnHover` is passed to `TrailingAction`
```
broccolinisoup, line 179: Do we still need this? Ref: https://github.com/github/primer/pull/3314/#discussion_r1635155551

packages/react/src/ActionList/TrailingAction.tsx
```diff
@@ -0,0 +1,59 @@
+import React, {forwardRef} from 'react'
+import Box from '../Box'
+import {Button, IconButton} from '../Button'
+import type {ForwardRefComponent as PolymorphicForwardRefComponent} from '../utils/polymorphic'
+
+export type ActionListTrailingActionProps = {
+  as?: 'button' | 'a'
+  href?: string
```
broccolinisoup, line 8: What would you think about defining types like this?

```tsx
{
  as?: 'button'
  href: never // href is invalid for button type
} | 
{
  as?: 'a'
  href: 'string' // href is required for a type
}
```

I have seen we did this before and I thought it is cool!

broccolinisoup (commented):
Thanks for this PR! 

I wonder if we can ship this separately from the ActionList semantics? Or what is the reason we want to ship with ActionList semantics, in other words? 😊 

I left some comments.

Could we also add the new sub component to the `ActionList/docs.json` ?

TylerJDev:
> I wonder if we can ship this separately from the ActionList semantics? Or what is the reason we want to ship with ActionList semantics, in other words? 😊

Technically we could, but the markup would be inaccessible as the trailing action would be within an interactive list item, e.g. 

```HTML
<li tabindex="0" ...> <!-- ActionList.Item -->
  <span data-component="ActionList.TrailingAction">
    <button>Action</button>> <!-- ActionList.TrailingAction -->
  </span>
</li>
```

This leads to an interactive element being within another. With the new semantics, the `TrailingAction` is adjacent to the button within the list item.

```HTML
<li> <!-- ActionList.Item -->
  <button>Item</button>
  <span data-component="ActionList.TrailingAction">
    <button>Action</button>> <!-- ActionList.TrailingAction -->
  </span>
</li>
```

packages/react/src/ActionList/Item.tsx
TylerJDev: We shouldn't need this anymore. Just removed it!

packages/react/src/ActionList/TrailingAction.tsx
khiga8: OOH TIL 😁

khiga8:
> I wonder if we can ship this separately from the ActionList semantics? Or what is the reason we want to ship with ActionList semantics, in other words? 😊

+1 to @tylerjdev! We branched this off of Tyler's semantics PR so we're building this feature on top of accessible semantics. 

I think this does mean that this TrailingAction won't be available until Tyler's semantics PR is merged, stability is confirmed behind the feature flag, and the feature flag is removed. Is there anything else needed to get that PR over the line?

packages/react/src/ActionList/TrailingAction.tsx
khiga8: ok this didn't quite work for me 😅

packages/react/src/ActionList/Item.tsx
```diff
@@ -313,6 +349,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
     let containerProps
     let wrapperProps
 
+
```
khiga8, line 352:
```suggestion
```

packages/react/src/ActionList/Item.tsx
```diff
@@ -340,6 +377,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
           ref={buttonSemantics || listSemantics ? forwardedRef : null}
           sx={
             buttonSemantics
+
```
khiga8, line 380:
```suggestion
```

packages/react/src/ActionList/Item.tsx
```diff
@@ -149,10 +159,32 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
+      },
+    }
+
     const listItemStyles = {
       display: 'flex',
       // show between 2 items
       ':not(:first-of-type)': {'--divider-color': theme?.colors.actionListItem.inlineDivider},
+      width: 'calc(100% - 16px)',
+      marginX: buttonSemantics ? '2' : '0',
+      ...(buttonSemantics ? hoverStyles : {}),
```
khiga8, line 187: @tylerjdev, do we still need this?

packages/react/src/ActionList/Item.tsx
```diff
@@ -237,6 +253,7 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
         '--divider-color': 'transparent',
       },
       ...(active ? activeStyles : {}),
+      ...(listSemantics || _PrivateItemWrapper ? hoverStyles : {}),
```
khiga8, line 256: @tylerjdev do we still need this?

packages/react/src/ActionList/TrailingAction.tsx
khiga8: @broccolinisoup, I continued to poke around with the syntax. See https://github.com/primer/react/commit/194f847351b216644a52d757e3a98a8c49c02b5d. 
I thought I was close, but typescript gets mad at the `href = null` thats passed into line 13 because `button` is assumed to be the default. 

I looked at other examples in the code and did not see any examples that achieve what we want. 🤔 

I would be open any ideas you have to get the types to work, but I don't think this should block us from moving forward!

packages/react/src/ActionList/Item.tsx
TylerJDev: Kinda, this gives consistency between the link items and the regular items when you hover over the `<li>`.

packages/react/src/ActionList/Item.tsx
TylerJDev: We don't! Just removed it.

joshblack (approved):

khiga8:
Thank you for the types suggestion @joshblack! It worked - [update types](https://github.com/primer/react/pull/4634/commits/b1f65d6663bfa6578a9c93dc6228422e731b3fea)! 😁

langermank (approved):

siddharthkp:
Hi! 👋 Please add the new prop to [ActionList.docs.json](https://github.com/primer/react/blob/b1f65d6663bfa6578a9c93dc6228422e731b3fea/packages/react/src/ActionList/ActionList.docs.json) as well

PR #4635, opened 2024-05-30
WIP Adds noTitle option to RelativeTime component

lindseywild:
Blocked until a new release of https://github.com/github/relative-time-element ships and is updated in primer/react.

Closes https://github.com/github/accessibility-audits/issues/5633

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/RelativeTime/RelativeTime.docs.json
+++ b/packages/react/src/RelativeTime/RelativeTime.docs.json
@@ -102,6 +102,12 @@
       "type": "React.RefObject<RelativeTimeElement>"
     },
     {
+      "name": "noTitle",
+      "type": "boolean",
+      "defaultValue": "",
+      "description": "Removes the `title` attribute provided on the element by default"
+    },
+    {
       "name": "as",
       "type": "React.ElementType",
       "defaultValue": "\"relative-time\""
--- a/packages/react/src/RelativeTime/RelativeTime.tsx
+++ b/packages/react/src/RelativeTime/RelativeTime.tsx
@@ -6,14 +6,14 @@ import {createComponent} from '../utils/create-component'
 const RelativeTimeComponent = createComponent(RelativeTimeElement, 'relative-time')
 
 const localeOptions: Intl.DateTimeFormatOptions = {month: 'short', day: 'numeric', year: 'numeric'}
-function RelativeTime({date, datetime, children, ...props}: RelativeTimeProps) {
+function RelativeTime({date, datetime, children, noTitle, ...props}: RelativeTimeProps) {
   if (datetime) date = new Date(datetime)
   return (
-    <RelativeTimeComponent {...props} date={date}>
+    <RelativeTimeComponent {...props} date={date} no-title={noTitle ? '' : undefined}>
       {children || date?.toLocaleDateString('en', localeOptions) || ''}
     </RelativeTimeComponent>
   )
 }
 
-export type RelativeTimeProps = ComponentProps<typeof RelativeTimeComponent>
+export type RelativeTimeProps = ComponentProps<typeof RelativeTimeComponent> & {noTitle?: boolean}
 export default RelativeTime
--- a/package-lock.json
+++ b/package-lock.json
@@ -317,7 +317,7 @@
     "examples/app-router": {
       "name": "example-app-router",
       "dependencies": {
-        "@primer/react": "36.19.0",
+        "@primer/react": "36.19.1",
         "next": "^14.1.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
@@ -336,7 +336,7 @@
         "react-dom": "^18.2.0"
       },
       "devDependencies": {
-        "@primer/react": "36.19.0",
+        "@primer/react": "36.19.1",
         "@types/react": "^18.2.55",
         "@types/react-dom": "^18.2.19",
         "@typescript-eslint/eslint-plugin": "^7.3.1",
@@ -391,7 +391,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@primer/octicons-react": "19.x",
-        "@primer/react": "36.19.0",
+        "@primer/react": "36.19.1",
         "next": "^14.1.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
@@ -61912,7 +61912,7 @@
     },
     "packages/react": {
       "name": "@primer/react",
-      "version": "36.19.0",
+      "version": "36.19.1",
       "license": "MIT",
       "dependencies": {
         "@github/combobox-nav": "^2.1.5",
--- a/packages/react/src/RelativeTime/RelativeTime.examples.stories.tsx
+++ b/packages/react/src/RelativeTime/RelativeTime.examples.stories.tsx
@@ -1,0 +1,31 @@
+import React from 'react'
+import RelativeTime from './RelativeTime'
+import {ClockIcon} from '@primer/octicons-react'
+import Link from '../Link'
+import {Tooltip} from '../TooltipV2'
+import {IconButton} from '../Button'
+
+export default {
+  title: 'Components/RelativeTime/Examples',
+  component: RelativeTime,
+}
+
+export const NoTitleAttribute = () => <RelativeTime date={new Date('2020-01-01T00:00:00Z')} noTitle={true} />
+
+export const LinkWithTooltip = () => (
+  <Tooltip text={new Date('2020-01-01T00:00:00Z').toString()}>
+    <Link href="https://www.github.com">
+      <RelativeTime date={new Date('2020-01-01T00:00:00Z')} noTitle={true} />
+    </Link>
+  </Tooltip>
+)
+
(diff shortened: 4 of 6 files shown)


lindseywild:
[Integration failures](https://github.com/github/github/actions/runs/9368314897/job/25790115581?pr=327453) are unrelated to this PR!

packages/react/src/RelativeTime/RelativeTime.tsx
```diff
@@ -6,14 +6,14 @@ import {createComponent} from '../utils/create-component'
...
   return (
-    <RelativeTimeComponent {...props} date={date}>
+    <RelativeTimeComponent {...props} date={date} no-title={noTitle ? '' : undefined}>
       {children || date?.toLocaleDateString('en', localeOptions) || ''}
     </RelativeTimeComponent>
   )
 }
 
-export type RelativeTimeProps = ComponentProps<typeof RelativeTimeComponent>
+export type RelativeTimeProps = ComponentProps<typeof RelativeTimeComponent> & {noTitle?: boolean}
```
smockle, line 18: I think this is no longer necessary, now that `@github/relative-time-element` has been updated. At least, when I remove it in my codespace, I don’t get type errors.

```suggestion
export type RelativeTimeProps = ComponentProps<typeof RelativeTimeComponent>
```

lindseywild: Ah, that makes sense! Thanks!

smockle (approved):

packages/react/src/RelativeTime/RelativeTime.docs.json
```diff
@@ -101,6 +101,12 @@
       "name": "ref",
       "type": "React.RefObject<RelativeTimeElement>"
     },
+    {
+      "name": "noTitle",
+      "type": "boolean",
+      "defaultValue": "",
+      "description": "Removes the `title` attribute provided on the element by default"
```
siddharthkp, line 108: Hi! 👋 

Can you also add when is it recommended to skip title?

siddharthkp:
Hi! 

Can i request additional context, what does passing `no-title` down do? When is it recommended to use + when is it not recommended to use?

joshblack:
Would it be preferable to always have the `title` not be present, at least from the Primer React side? Internally we could set `noTitle` on the relative time element but we wouldn't expose it as a prop to consumers to toggle.

Just wanted to ask to see if this would be helpful or if not because of what you shared about usage at GitHub and how there isn't a good alternative yet when the title is not displayed 👀

lindseywild:
> Would it be preferable to always have the title not be present, at least from the Primer React side? Internally we could set noTitle on the relative time element but we wouldn't expose it as a prop to consumers to toggle.

Yes, it would be preferable to never have a title attribute! I added a story that shows how we can instead use a Tooltip on interactive elements, but that doesn't solve the case of non-interactive elements :/ If you think it's ok I am totally fine removing the `title` attribute altogether, but I think that may ruffle some feathers and we'd would probably require user testing first 😅

packages/react/src/RelativeTime/RelativeTime.docs.json
lindseywild: In this description? Or are you talking about updating the [in-progress docs PR](https://github.com/primer/design/pull/795/files)?

siddharthkp:
> If it helps, I have a [PR for updated guidance](https://github.com/primer/design/pull/795/files) in the works that I will merge after this one is merged :)

Yes, this was very helpful! 

Please correct me if I understood this incorrectly:
- For visual users, the date is presented as "3 days ago" and then they can hover for exact date + time
- For screen reader users, the date is also presented as "3 days ago", but there is no way to know exact date + time because title is ignored.

(this is where I'm probably understanding incorrectly) 

To make both experiences similar, we need to replace title with an accessible element, like a tooltip. By removing `title` without replacing it with tooltip, we do make the experience the same, but don't we lose the detailed information about data and time?

---

> Ideally it'd be default behavior, but we'd need a larger conversation with stakeholders and design to come up with an accessible solution (we've attempted this in Office Hours, but there is nothing straightforward that we can do other than this without extensive testing).

That's fair! I feel a little worried about `noTitle` because I don't know how it or where will be used in the product. Can you share a little more on that.

How do you feel about primer components for RelativeTime (both PRC and PVC) using a tooltip instead of a title? It would lead to some inconsistency in the short term (between primer vs non-primer relative time), but hopefully will set the direction towards accessible + consistent

lindseywild:
You are mostly correct with how you're understanding it! The other user we're excluding is keyboard users -- even if a user can navigate to the element because it's focusable, like a link, the title attribute does not display for them. 

> but don't we lose the detailed information about data and time?

No, if we pass in the full date and time as the tooltip (check out the story [Link with Tooltip](https://primer-f33b80816f-13348165.drafts.github.io/storybook/?path=/story/components-relativetime-examples--link-with-tooltip)) that shows the full date in the tooltip, but the shortened version as the main text.

> I feel a little worried about noTitle because I don't know how it or where will be used in the product. Can you share a little more on that.

Sure! We can think of this as a "baby step" to a fully accessible solution. Since we don't have one right now, this is _better_ than what currently exists, but still not fully _ideal_. One example of where this can be used in the product is on PRs, where the timestamp acts as a link to that comment. We can use `noTitle` to not render a `title` attribute, and instead replace it with a Tooltip with the full time as it is in the example story.

I am also thinking that if audit issues start coming in, similar to the one linked in the PR, it gives teams a way to fix the audit issue with minimal impact or needing to reassign the issue to Primer -- they have a few options of how to address it. They can do one of the following:

* Pass in `noTitle` so they don't render a title (and we can let the teams decide if they want to replace it or go without it)
* For interactive elements, pass in `noTitle` and pass in the full date/time in a Tooltip
* For non-interactive elements, pass in `noTitle` and write as much of the full date/time out as possible

> How do you feel about primer components for RelativeTime (both PRC and PVC) using a tooltip internally instead of expecting it to be added in the product?

We considered this! The tricky part is that not all RelativeTime elements are interactive. Some are static text, which we can't render a Tooltip on in an accessible way. We considered having two components, RelativeTime (as-is) and RelativeTimeWithTooltip, but right now consumers wrap the RelativeTime component in a `<Link>` tag, so it'd be hard to determine in the component what it's parent is and whether or not it should render a Tooltip. Also, consumers may not know the difference between the two. But if you think otherwise, we can explore it.

siddharthkp:
Ah, I seeee! Thanks you going into detail, now it's very clear to me.

> The tricky part is that not all RelativeTime elements are interactive.

I completely missed this detail

---

Summarising for myself first:

Current scenario:
- RelativeTime renders an inline text element (usually short text like "3 days ago")
- We add a title attribute to the element which has the full datetime, but is only available to visual users, not to screen reader users

Ideal scenario:
- Case 1: If RelativeTime (inline text element) is wrapped in an interactive element like Link, the parent element can receive focus and we can add a Tooltip on the parent element with full datetime. **To make sure there isn't also a redundant browser tooltip, we need to remove title attribute**
- Case 2: If RelativeTime is standing on it's own, there is no way to show full datetime consistently between visual users and screen reader users. We need to modify the text to show more details by default. **We could keep the title attribute, but it's better to nudge developers to keep the experience consistent**

packages/react/src/RelativeTime/RelativeTime.docs.json
siddharthkp: Ideally both places, but either is fine for now :)

lindseywild:
@siddharthkp Everything you said looks good, one slight modification:

"We add a title attribute to the element which has the full datetime, but is only available to visual users, not to screen reader users **or keyboard users**"

And regarding:
> We could keep the title attribute, but it's better to nudge developers to keep the experience consistent

Correct! Eventually we hope to have a better solution, but this is just one small step to getting there :)

siddharthkp:
---

> We can think of this as a "baby step" to a fully accessible solution. Since we don't have one right now, this is better than what currently exists, but still not fully ideal. 
> I am also thinking that if audit issues start coming in, similar to the one linked in the PR, it gives teams a way to fix the audit issue with minimal impact

Very fair, happy to merge this as a _temporary_ "unblocker" for teams. I'd nudge you to follow up with a discussion post to find a long term fix? (I call it _temporary_, but once it's in dotcom, nothing is _temporary_ as we would need to keep it until it's replaced)

siddharthkp (approved):
Approved. See also: https://github.com/primer/react/pull/4635#issuecomment-2150350150

Ready to merge whenever you'd like

lindseywild:
@siddharthkp Thank you! That totally makes sense. That plan would also need to include a deprecation plan for `relative-time-element` since it's reliant on that... I can follow up with a Discussion post with the current findings and research we've done :)

lindseywild:
@siddharthkp posted! https://github.com/primer/react/discussions/4645

PR #4640, opened 2024-06-03
fix(SegmentedControl): explicitly set the `type` attribute for `SegmentedControl.Button`

joshblack:
Closes https://github.com/github/primer/issues/3308

#### Changed

- Explicitly set the `type` attribute of the `SegmentedControl.Button` component to "button" to prevent automatic submission in `form` elements

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

- [ ] Verify that the `SegmentedControl.Button` component renders a `<button>` element with `type="button"`

### DIFF
--- a/.changeset/few-paws-confess.md
+++ b/.changeset/few-paws-confess.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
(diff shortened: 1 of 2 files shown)


broccolinisoup (approved):
We just need a changeset but the PR is looking great!

PR #4646, opened 2024-06-05
chore(deps): update react, react-dom, testing-library

joshblack:
Update `react`, `react-dom`, `@testing-library` and corresponding `@types/*` dependencies to the latest version. This will only update our internal dependencies, not the dependencies that we expect from downstream users.

#### Changed

- Move `act` from `react-dom/test-utils` to `react`
- Update react, react-dom
- Update testing-library
- Update types

### Rollout strategy


packages/react/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
```diff
@@ -118,7 +118,7 @@ const InlineAutocomplete = ({
   const inputProps = useFormControlForwardedProps(externalInputProps)
 
   const inputRef = useRef<HTMLInputElement & HTMLTextAreaElement>(null)
-  useRefObjectAsForwardedRef(children.ref ?? noop, inputRef)
+  useRefObjectAsForwardedRef(children.ref && typeof children.ref !== 'string' ? children.ref : noop, inputRef)
```
siddharthkp, line 121: The string comparison seems like a bonus here, did this show up as validation error with the new version?

siddharthkp (approved):
Left 1 clarifying comment, everything else looks good!

Approving in advance

packages/react/src/drafts/InlineAutocomplete/InlineAutocomplete.tsx
joshblack: @siddharthkp yeah! Just showed up as a TS error

PR #4649, opened 2024-06-06
Bumps eslint-plugin-github and eliminates false positive a11y rule flag

khiga8:
Relates to: https://github.com/github/accessibility/issues/5304

### Changelog

This PR bumps the `eslint-plugin-github` to the [latest release](https://github.com/github/eslint-plugin-github/releases/tag/v5.0.1). With this release, the `no-redundant-roles` rule will stop flagging two false positive reports, because of an update we made in: https://github.com/github/eslint-plugin-github/pull/531.

These two lint disables have been showing up as needing resolution in Primer's Accessibility Automation scorecard. @pksjce and @TylerJDev provided very helpful feedback in https://github.com/github/accessibility/discussions/4921 that these are false positives and how setting `role="rowgroup"` explicitly is necessary in order to improve screen reader semantics. 

Based on this feedback, we excluded redundant use of `role="rowgroup"` from being flagged in the linting config since:
1. All reported instances were necessary to improve screen reader semantics to get around an AT/browser bug.
2. Redundant use of role isn't actually harmful for accessibility, and is more of a readability/best practice. It is worth continuing to enforce, but its not the end of the world if we don't flag it for `role="rowgroup"`.

Thank you very much for reporting this and we apologize for the delay in following-up. We hope to follow-up with the other instances as well.

#### Removed

* Two lint disables that are no longer relevant so it will not be counted in the accessibility scorecard.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "@typescript-eslint/parser": "^7.3.1",
         "eslint": "8.56.0",
         "eslint-import-resolver-typescript": "3.6.1",
-        "eslint-plugin-github": "4.10.1",
+        "eslint-plugin-github": "5.0.1",
         "eslint-plugin-jest": "27.2.1",
         "eslint-plugin-jsx-a11y": "6.7.1",
         "eslint-plugin-mdx": "3.0.0",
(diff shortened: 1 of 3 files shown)


TylerJDev (approved):
Thanks for updating this!

PR #4650, opened 2024-06-06
feat: add rollup-plugin-import-css

joshblack:
This is a follow up to the RFC in: https://github.com/primer/react/pull/4609

This PR adds the package proposed by the RFC with a slight name change: `rollup-plugin-import-css`. This plugin for Rollup allows us to author CSS modules in Primer React and generate CSS alongside the component in JavaScript. Each CSS module is compiled to a JavaScript file that includes the classes for the CSS file. This generated file also imports the generated CSS.

In the RFC, we used `Blankslate` as an example for this. Using this plugin would generate the following structure:

```
dist/Blankslate
├── Blankslate-c90a59c9.css
├── Blankslate.d.ts
├── Blankslate.js
├── Blankslate.module.css.js
└── index.d.ts
```

Here, `Blankslate.js` would import classes from `Blankslate.module.css.js` This module looks like:

```js
import './Blankslate-c90a59c9.css'
export default {
  exampleClassName: 'generated_example_class_name',
};
```

The main advantage of this approach is that we generate a CSS file per-component. This allows consumers to use a component without importing additional CSS that is not used by their project.

> [!NOTE]
> This PR **does not** change the Rollup config for `@primer/react`. This will be tackled separately when we begin work towards migrating CSS modules in that package.

#### New

- Add `rollup-plugin-import-css` to the project

### Rollout strategy


joshblack:
Wanted to include you @jonrohan as a reviewer with your work on CSS Modules stuff this quarter 👀

jonrohan (approved):

PR #4657, opened 2024-06-10
PageHeader: Move `variant` prop of `PageHeader.Title` to the parent `PageHeader` component to prevent layout shift

broccolinisoup:
Reported internally [on Slack (Hubbers link only)](https://github.slack.com/archives/C01L618AEP9/p1717754823228669) - Please note this is the fix of the`PageHeader.Actions` layout shift issue in the PageHeader component. Not the font size which I am still working on it. 

This is technically a breaking API change but PageHeader is still an experimental component and we got feedback about the layout shift after shipping https://github.com/primer/react/pull/4358. 

It is caused by setting the `data-size-variant={titleVariant}` on PageHeader component  (to set height for other sub elements such as leading and trailing visual and actions) in the useeffect. We had to do it this way in the useeffect because we only get the value of the `variant` prop _after_ rendering the `PageHeader.TitleArea` component. This PR moves the `variant` prop to the parent `PageHeader` component to prevent layout shift.

Let me know if you have a different idea to handle this.

I noticed this layout shift issue but thought it is due to the slowness of the storybook but unfortunately it is quite visible in the real life too :/ 

#### Changed

- Move the `variant` prop from `PageHeader.TitleArea` to `PageHeader` and rename it to `titleVariant`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/rude-wombats-buy.md
+++ b/.changeset/rude-wombats-buy.md
@@ -1,6 +1,5 @@
 ---
 "@primer/react": patch
-"docs": patch
 ---
 
 This is technically a breaking API change.
--- a/.changeset/rude-wombats-buy.md
+++ b/.changeset/rude-wombats-buy.md
@@ -1,0 +1,7 @@
+---
+"@primer/react": patch
+"docs": patch
+---
+
+This is technically a breaking API change.
+Move `variant` prop of `PageHeader.TitleArea` to the parent `PageHeader` component as `titleVariant` to prevent layout shifts
--- a/packages/react/src/PageHeader/PageHeader.dev.stories.tsx
+++ b/packages/react/src/PageHeader/PageHeader.dev.stories.tsx
@@ -18,19 +18,27 @@ export default meta
 
 export const LargeVariantWithMultilineTitle = () => (
   <Box sx={{padding: 3}}>
-    <PageHeader>
+    <PageHeader titleVariant="large">
       <PageHeader.LeadingAction>
         <IconButton aria-label="Edit" icon={PencilIcon} variant="invisible" />
       </PageHeader.LeadingAction>
-      <PageHeader.TitleArea variant="large">
+      <PageHeader.TitleArea>
         <PageHeader.LeadingVisual>
           <GitBranchIcon />
         </PageHeader.LeadingVisual>
-        <PageHeader.Title>
+        <PageHeader.Title
+          sx={{
+            // lineHeight: '1.25',
+            fontWeight: 'normal',
+            // fontSize: '32px',
+            fontSize: ['26px', '26px', '32px', '32px'],
+            // fontSize: ['26px', '26px', 'var(--text-title-size-large, 32px)', 'var(--text-title-size-large, 32px)'], // it doesn't support this format right now.
+          }}
+        >
           Title long title some extra loooong looong words here some extra loooong looong words here some extra loooong
           looong words here some extra loooong looong words here some extra loooong looong words here
         </PageHeader.Title>
-        <PageHeader.TrailingVisual>
+        <PageHeader.TrailingVisual sx={{height: '120px'}}>
           <Label>Beta</Label>
         </PageHeader.TrailingVisual>
       </PageHeader.TitleArea>
--- a/packages/react/src/PageHeader/PageHeader.dev.stories.tsx
+++ b/packages/react/src/PageHeader/PageHeader.dev.stories.tsx
@@ -26,19 +26,11 @@ export const LargeVariantWithMultilineTitle = () => (
         <PageHeader.LeadingVisual>
           <GitBranchIcon />
         </PageHeader.LeadingVisual>
-        <PageHeader.Title
-          sx={{
-            // lineHeight: '1.25',
-            fontWeight: 'normal',
-            // fontSize: '32px',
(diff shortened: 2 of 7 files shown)


.changeset/rude-wombats-buy.md
```diff
@@ -0,0 +1,7 @@
+---
+"@primer/react": patch
+"docs": patch
```
broccolinisoup, line 3:
```suggestion
```

packages/react/src/PageHeader/PageHeader.dev.stories.tsx
```diff
@@ -18,11 +18,11 @@ export default meta
 
 export const LargeVariantWithMultilineTitle = () => (
   <Box sx={{padding: 3}}>
-    <PageHeader>
+    <PageHeader titleVariant="large">
```
broccolinisoup, line 21: I feel good about this API change. Even though, it seems more title related, we also manage the `height` of the all these leading and trailing visuals and actions by this prop. Because when title font size, line height etc is changing depending on the size, the height of the other elements change too. So I think it makes better sense to me to have this prop higher in the hierarchy. 

Also, open to naming suggestions 🙌🏻

packages/react/src/PageHeader/PageHeader.tsx
```diff
@@ -61,13 +61,18 @@ const hiddenOnNarrow = {
...
+  {
+    'aria-label'?: React.AriaAttributes['aria-label']
+    as?: React.ElementType | 'header' | 'div'
+    titleVariant?: 'subtitle' | 'medium' | 'large' | ResponsiveValue<'subtitle' | 'medium' | 'large'>
+  } & SxProp
+>
+
+const Root = React.forwardRef(
+  ({children, sx = {}, as = 'div', titleVariant = 'medium', ...rest}: PageHeaderProps, forwardedRef) => {
+    const [fontSize, setFontSize] = useState<string | null | number | string[]>(null)
```
broccolinisoup, line 74: This part still causes a layout shift and it is the only way I found to handle [css-in-js style overrides](https://github.com/github/github/blob/master/ui/packages/issue-viewer/components/header/HeaderViewer.tsx#L71)


I wrote my thoughts on the [Slack thread (internal only)](https://github.slack.com/archives/C01L618AEP9/p1717996593042509?thread_ts=1717754823.228669&cid=C01L618AEP9) - I think we should bring some sort of limitation to what we allow as an override especially for PageHeader since the goal of this component is to bring consistency on GH headings.

packages/react/src/PageHeader/PageHeader.tsx
```diff
@@ -124,9 +143,21 @@ const Root = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageHeader
...
 
-      // // grab the data-size-variant attribute from the titleArea
-      const sizeVariant = titleArea.getAttribute('data-size-variant')
-      setTitleVariant(sizeVariant as string)
+      // find the title element
+      const title = Array.from(titleArea.children as HTMLCollection).find(child => {
+        return child instanceof HTMLElement && child.getAttribute('data-component') === 'PH_Title'
+      })
+
+      const styles = getComputedStyle(title as HTMLHeadingElement)
```
mperrotti, line 151: I'm confused about why we need to `getComputedStyle` on the title

broccolinisoup: I mentioned [here](https://github.com/primer/react/pull/4657/files#r1633020618) why I have this, but we should find another way so that we can remove this from the useeffect because it causes layout shift. 


@mperrotti The reason I have this solution is because to support array type overrides. 👇

```jsx
<PageHeader.Title
  sx={{
    fontSize: ['26px', '26px', 'var(--text-title-size-large, 32px)', 'var(--text-title-size-large, 32px)'],
  }}
>
  Issue Title
</PageHeader.Title>
```

When [we look for, for example, `lineHeight` in sx prop](https://github.com/primer/react/blob/main/packages/react/src/PageHeader/PageHeader.tsx#L456), we assign that value to a css variable `--custom-line-height` so that the default styles can be [overwritten](https://github.com/primer/react/blob/main/packages/react/src/PageHeader/PageHeader.tsx#L339) however when we have an array type like above, css var can't resolve that array. So we need some other way to resolve it and this is not the best way to resolve it, I know 😅 I just don't know how else we can solve this.  

I pushed another [PR](https://github.com/primer/react/pull/4658) that doesn't involve javascript but it is not great either.

I also posted a discussion to talk about overrides https://github.com/github/primer/discussions/3329 - Let me know what you think! We should maybe pair on it and chat further?

broccolinisoup:
Closing this because we found a better way without API change. I'll push the PR very soon!

packages/react/src/PageHeader/PageHeader.tsx
broccolinisoup: No worries about this problem anymore, [I am closing this PR](https://github.com/primer/react/pull/4657#issuecomment-2162694294) and I'll chat you tomorrow! @mperrotti

PR #4666, opened 2024-06-11
Utilize `aria-describedby` on all `ActionList` descriptions

TylerJDev:
Closes https://github.com/github/primer/issues/2501, https://github.com/github/primer/issues/3471

#### Changed

* Added `aria-labelledby` to `ActionList.TrailingVisual`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
@primer/engineer-reviewers - Could I get a review on this PR? :eyes:

packages/react/src/ActionList/Item.tsx
```diff
@@ -294,10 +295,11 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       'data-inactive': inactive ? true : undefined,
       'data-loading': loading && !inactive ? true : undefined,
       tabIndex: disabled || showInactiveIndicator ? undefined : 0,
-      'aria-labelledby': `${labelId} ${slots.inlineDescription ? inlineDescriptionId : ''}`,
-      'aria-describedby': slots.blockDescription
-        ? [blockDescriptionId, inactiveWarningId].join(' ')
-        : inactiveWarningId,
+      'aria-labelledby': `${labelId} ${slots.trailingVisual ? trailingVisualId : ''}`,
+      'aria-describedby':
```
siddharthkp, line 299: Untested code, that reads slightly better:

```suggestion
      'aria-describedby': [
        slots.blockDescription ? blockDescriptionId : undefined,
        inactiveWarningId ?? undefined,
        slots.inlineDescription ? inlineDescriptionId : undefined,
      ]
        .filter(String)
        .join(' '),
```

siddharthkp (changes requested):
Left some clarifying questions

packages/react/src/ActionList/Item.tsx
```diff
@@ -294,10 +295,11 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
       'data-inactive': inactive ? true : undefined,
       'data-loading': loading && !inactive ? true : undefined,
       tabIndex: disabled || showInactiveIndicator ? undefined : 0,
-      'aria-labelledby': `${labelId} ${slots.inlineDescription ? inlineDescriptionId : ''}`,
-      'aria-describedby': slots.blockDescription
-        ? [blockDescriptionId, inactiveWarningId].join(' ')
-        : inactiveWarningId,
+      'aria-labelledby': `${labelId} ${slots.trailingVisual ? trailingVisualId : ''}`,
```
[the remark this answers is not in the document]
TylerJDev: > How do we decide if trailing visual should be in label or description?

I've added it as a label via `aria-labelledby` instead of a description because if there's text content within trailing visual, it's usually short and connects directly to the main item's text (accessibl, e.g. 

```tsx
<ActionList>
  <ActionList.Item>Tokens Available 
    <ActionList.TrailingVisual>8</ActionList.TrailingVisual>
  </ActionList.Item>
</ActionList>
```
Accessible name would be: "Tokens Available 8". 

I mainly based this on examples I've seen in the wild. If we added it as a description, there's a chance it might be ignored, so I think `aria-labelledby` is a fair bet.

> Are there places in dotcom which would end up with the wrong label / description from this?

I'd say it's very unlikely, since any text within the trailing visual isn't attached to `ActionList.Item`'s accessible name/description currently.

> Re: moving inline description from label to description, can we create a different PR for that because I'm less confident about that change.

Yup that's fair! I think it's worth its own discussion to see if we'd want to switch it or not, as it definitely works both ways, as part of the accessible name, or as part of the description. I'll take that change out of this PR and create another so that we can discuss/see if we want to include it or not!

packages/react/src/ActionList/Item.tsx
TylerJDev: Looks like it worked! Thanks for the suggestion!

.changeset/lovely-days-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Adds `aria-describedby` to `ActionList` inline descriptions, adds `aria-labelledby` to `ActionList.TrailingVisual`
```
TylerJDev, line 5:
```suggestion
Adds `aria-labelledby` to `ActionList.TrailingVisual`, making it part of the accessible name of `ActionList.Item`
```

packages/react/src/ActionList/Item.tsx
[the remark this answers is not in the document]
TylerJDev: Should now only include the proposed changed for `TrailingVisual`!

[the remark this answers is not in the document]
siddharthkp: > I mainly based this on examples I've seen in the wild.

Does this also include within menus/selectpanels?

packages/react/src/ActionList/Item.tsx
```diff
@@ -294,10 +295,14 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
...
-        ? [blockDescriptionId, inactiveWarningId].join(' ')
-        : inactiveWarningId,
+      'aria-labelledby': `${labelId} ${slots.trailingVisual ? trailingVisualId : ''} ${
+        slots.inlineDescription ? inlineDescriptionId : ''
+      }`,
+      'aria-describedby':
+        [slots.blockDescription ? blockDescriptionId : undefined, inactiveWarningId ?? undefined]
+          .filter(String)
+          .join(' ')
+          .trim() || undefined,
```
siddharthkp, line 305: Optional: Don't think we need trim now that we are filtering strings

packages/react/src/NavList/__snapshots__/NavList.test.tsx.snap
```diff
@@ -339,7 +339,7 @@ exports[`NavList renders a simple list 1`] = `
       >
         <a
           aria-current="page"
-          aria-labelledby=":r2:--label "
+          aria-labelledby=":r2:--label  "
```
siddharthkp, line 342: Could me make the filtering string on aria-labelledby as well so that we don't get these extra spaces 😅

siddharthkp (approved):
Left a few tidying tasks, otherwise looks ready!

Approving in advance :)

packages/react/src/ActionList/Item.tsx
TylerJDev: I added this because we'd get `aria-describedby=" "` if none of the values were true. With the `trim() || undefined`, it will remove the attribute altogether since it's just an empty string at that point 🤔

packages/react/src/NavList/__snapshots__/NavList.test.tsx.snap
TylerJDev: Yes, definitely! I forgot I turned on auto merge 🤣, I can follow up in one of the PRs that I'll make for one of [the sev-3 issues](https://github.com/github/primer/issues/3558)!

PR #4668, opened 2024-06-12
use has prop to select the titlearea and different variants

broccolinisoup:
Reported internally [on Slack (Hubbers link only)](https://github.slack.com/archives/C01L618AEP9/p1717754823228669) - Please note this is the fix of the`PageHeader.Actions` layout shift issue in the PageHeader component. Not the font size which I am still working on it. 

The layout shift was caused by setting the `data-size-variant={titleVariant}` on PageHeader component  (to set height for other sub elements such as leading and trailing visual and actions) in the useeffect. We had to do it this way in the useeffect because we only get the value of the `variant` prop _after_ rendering the `PageHeader.TitleArea` component. However, pairing with @siddharthkp yesterday, he suggested to explore using `has` prop since we only change styles based on variants and nothing else, it simply worked! ✨ 

This PR removes reading the `variant` value from `PageHeader.TitleArea` after the render. Instead, it simply looks for elements that has `data-component="TitleArea"` and each related variant data attributes and set the styling for sub components under. Since we don't need to wait for any render, it simply prevents layout shift.

I noticed this layout shift issue before but thought it is due to the slowness of the storybook but unfortunately it is quite visible in the real life too :/ 

#### Changed

- Move the `variant` prop from `PageHeader.TitleArea` to `PageHeader` and rename it to `titleVariant`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/PageHeader/PageHeader.tsx
+++ b/packages/react/src/PageHeader/PageHeader.tsx
@@ -80,19 +80,19 @@ const Root = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageHeader
     `,
       //  --custom-height is a custom property (passed by sx) that can be used to override the set height.
       // We don't want these values to be overriden but still want to allow consumers to override them if needed.
-      '&[data-size-variant="large"]': {
+      '&:has([data-component="TitleArea"][data-size-variant="large"])': {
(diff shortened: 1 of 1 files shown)


broccolinisoup:
closing in favor of https://github.com/primer/react/pull/4669

PR #4669, opened 2024-06-12
PageHeader: Fixes layout issues on title and sub components

broccolinisoup:
Reported internally [on Slack (Hubbers link only)](https://github.slack.com/archives/C01L618AEP9/p1717754823228669) that there is a layout shift happening on titles and sub components. (font-size (for titles) and the height of the actions change after the initial render)

#### Changed

- Updates root styles to select elements that _has_ `data-component="TitleArea"` and each related variant data attribute, and sets the font styling and height for sub components (leading actions, actions etc) for each variant. Since we don't need to wait for any render as we did before, we don't see any layout shift. 
- Since we set the font styles on `TitleArea` (as described just above), we don't hit specificity issues when there is an override on the Title (like the reported issue), therefore no layout shift on the titles either.

#### Removed

- Removes retrieving the `variant` value from `PageHeader.TitleArea` in the useeffect.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [x] (GitHub staff only) Integration tests pass at github/github - https://github.com/github/github/pull/328861 Performed manual testing at dotcom and can confirm that all is working fine 👍🏻


packages/react/src/PageHeader/PageHeader.tsx
```diff
@@ -435,6 +395,9 @@ const Breadcrumbs: React.FC<React.PropsWithChildren<ChildrenPropTypes>> = ({
             return value ? 'none' : 'flex'
           }),
           alignItems: 'center',
+          fontWeight: 'initial',
+          lineHeight: 'var(--text-body-lineHeight-large, 1.5)',
```
mperrotti, line 399: Shouldn't we be using `var(--text-body-lineHeight-medium, 1.4285)`?

broccolinisoup: Oh sorry I missed to update that! Just pushed it.

mperrotti (approved):
In the VRT snapshots, the text next to the "Open" badge looks vertically off-center, but I'm not actually seeing this when I look at the preview deployment. I checked this in Chrome, Safari, and Firefox.

Screenshot 2024-06-25 at 4 32 11 PM

broccolinisoup:
@mperrotti yeah it is interesting! I can confirm that I see the same thing as you shared. It might be because of the font settings etc are different where these snapshots are generated.

PR #4673, opened 2024-06-14
refactor(live-region): update live region helpers to match ADR

joshblack:
Update our live region helpers to match the feedback and decisions from the ADR: https://github.com/primer/react/pull/4611

#### New

- Add support for `hidden` to `Announce`, `AriaStatus`, and `AriaAlert` so that messages are not announced when the container is hidden
- Add support for `delayMs` to `Announce`, `AriaStatus`, and `AriaAlert` to specify the delay before making an announcement
- Add `Announce`, `AriaStatus`, and `AriaAlert` as experimental components

### Rollout strategy

- [x] Minor release


khiga8 (approved):
This looks great! ❤️ Thank you for all your work on this! I think aligning the defaults with default behavior of ARIA is safe to start with, and we could revisit defaults if we need to.

Do you ideas around how we might validate that people are using the API as intended?

joshblack:
> Do you ideas around how we might validate that people are using the API as intended?

@khiga8 love this question, thanks for asking. I definitely want to have a chart over in the Primer Query dashboard not only for adoption but to see what props are being used. My hope is that based on this (and usage within PRC) we get a better idea of what is needed, whether that's lint rules, getting rid of props, or something to help with common gotchas.

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
```diff
@@ -604,7 +604,8 @@ const SelectPanelSecondaryAction: React.FC<SelectPanelSecondaryActionProps> = ({
 
 const SelectPanelLoading = ({children = 'Fetching items...'}: React.PropsWithChildren) => {
   return (
-    <Status
+    <AriaStatus
+      announceOnShow
```
jonrohan, line 608: Stupid question, are there styles attached to `<Status>` and is that different than `<AriaStatus>`

jonrohan (approved):
Scanned the code and checked out the story changes. Looks pretty good 👍🏻

packages/react/src/drafts/SelectPanel2/SelectPanel.tsx
joshblack: @jonrohan no styles, just behavior for now 👀 They're both setup as a `Box` though internally 😅

PR #4681, opened 2024-06-18
Figma connect files

lukasoppermann:
This PR adds some more code connect files to connect react components to figma implementations.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/FormControl/FormControl.figma.tsx
```diff
@@ -0,0 +1,40 @@
...
+import FormControl from '../FormControl'
+
+figma.connect(
+  FormControl.Label,
+  'https://www.figma.com/design/GCvY3Qv8czRgZgvl1dG6lp/Primer-Web?node-id=34462-69137&t=WXa19NDaeGGhlbLq-4',
+  {
+    props: {
+      label: figma.textContent('Label'),
+    },
+    example: ({label}) => <FormControl.Label>{label}</FormControl.Label>,
```
joshblack, line 12: Do you have a preference between disabling the lint rule for these examples specifically or wrapping them in a FormControl? 👀

joshblack (approved):
Looks great 👍 Just left a question about the ESLint errors popping up and also saw that some of the files have TS errors in CI. Let me know if I can help out at all with those!

packages/react/src/FormControl/FormControl.figma.tsx
lukasoppermann: Yes, we need to disable them, as they are nested in other components, that wrap everything in a Form.Control already.

lukasoppermann:
Hey @joshblack I think we need to disable the type check for those files, or do you know how we can fix it? We can't add an `as TYPE` statement, as this will me transferred to figma.

packages/react/src/UnderlineNav/UnderlineNav.figma.tsx
```diff
@@ -28,22 +28,26 @@ figma.connect(
...
         default: undefined,
-        hover: '"false"',
-        focus: '"false"',
+        hover: 'false',
+        focus: 'false',
       }),
       counter: figma.boolean('counter?', {
         false: undefined,
-        true: figma.children('CounterLabel'),
+        true: figma.string('CounterLabel'),
```
lukasoppermann, line 38: This has to be reverted, we need children here.

packages/react/src/UnderlineNav/UnderlineNav.figma.tsx
```diff
@@ -28,22 +28,26 @@ figma.connect(
     props: {
       label: figma.textContent('Label'),
       current: figma.enum('state', {
-        selected: '"page"',
```
lukasoppermann: We need double quotes, otherwise we end up with no quotes in figma.

packages/react/src/UnderlineNav/UnderlineNav.figma.tsx
```diff
@@ -28,22 +28,26 @@ figma.connect(
...
         false: undefined,
         true: figma.children('Icon'),
       }),
     },
     example: ({label, current, counter, leadingVisual}) => (
-      <UnderlineNav.Item aria-current={current} counter={counter} icon={leadingVisual}>
+      <UnderlineNav.Item
+        aria-current={current}
+        counter={counter}
+        icon={leadingVisual ? () => leadingVisual : undefined}
```
lukasoppermann, line 49: this does not work I think as the code is not executed. But I'll check.

joshblack:
@lukasoppermann I'll revert the commit based on your comments 👍 I'll add more info here for where the changes came from.

> This has to be reverted, we need children here.

The type error is because `counter` only accepts `string | number` in the component but we're giving it a `JSX.Element` from `figma.children`. 

> We need double quotes, otherwise we end up with no quotes in figma.

The type error here is that the value being passed on as `current` is not `'page'` but instead is `'"page"'` with the nested double quotes.

> this does not work I think as the code is not executed. But I'll check.

The type error here is because `leadingVisual` does not accept `JSX.Element`, it expects a `React.FunctionComponent` which is where `() => leadingVisual` comes from.

---

My assumption is that we're running into an issue between how we want to represent it in Figma and how it's structured over in TS world 👀 Let me know how you would like to resolve these. My fear for disabling TypeScript on these examples is that TS seems to be catching snippets that would be invalid if someone tried to copy and use them in code due to the type mismatches.

packages/react/src/UnderlineNav/UnderlineNav.figma.tsx
lukasoppermann: Nope, it does not work and moving it to the figma.boolean section also does not.
I reached out to Figma for this.

packages/react/src/UnderlineNav/UnderlineNav.figma.tsx
lukasoppermann: I think we can ignore the error here, as the output is fine, right?

CleanShot 2024-07-03 at 09 22 45@2x

lukasoppermann: It seems this was fixed on figmas side

PR #4685, opened 2024-06-19
fix(Spinner): use aria-label when defined, falling back to srText

joshblack:
Resolve an issue that is showing up in: https://github.com/github/github/pull/329788 where `Spinner` will always use `srText` instead of the provided `aria-label`.

#### Changed

- Update Spinner to render an sr-only div only if `aria-label` is not defined. In this case, use `srText` as the value for the sr-only div.

### Rollout strategy

- [x] Patch release

### DIFF
--- a/examples/app-router/tsconfig.json
+++ b/examples/app-router/tsconfig.json
@@ -12,7 +12,7 @@
     "incremental": true,
     "esModuleInterop": true,
     "module": "esnext",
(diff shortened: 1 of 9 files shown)


jonrohan (approved):
Any tests?

joshblack:
@jonrohan good call, added!

PR #4686, opened 2024-06-19
Add expand to `NavList`

TylerJDev:
Adds new component `NavList.ShowMoreItem`, allows native support for "expanding" content within a `NavList`.

Closes https://github.com/github/primer/issues/2637

#### Proposed API

[Basic example](https://primer-b750586425-13348165.drafts.github.io/storybook/?path=/story/components-navlist--with-expand):
```tsx
export const WithExpand = () => {
  const items = [
    {href: '#', text: 'Item 4'},
    {href: '#', text: 'Item 5'},
    {href: '#', text: 'Item 6'},
    {href: '#', text: 'Item 7'},
    {href: '#', text: 'Item 8'},
    {href: '#', text: 'Item 9'},
  ]

  return (
    <NavList>
      <NavList.Item href="#" aria-current="page">
        Item 1
      </NavList.Item>
      <NavList.Item href="#">Item 2</NavList.Item>
      <NavList.Item href="#">Item 3</NavList.Item>
      <NavList.GroupExpand label="Show more" items={items} />
    </NavList>
  )
}
```

[Multiple expands](https://primer-b750586425-13348165.drafts.github.io/storybook/?path=/story/components-navlist--with-multiple-expands):
```tsx
export const ExpandWithPages = () => {
  const items = [
    {href: '#', text: 'Item 4'},
    {href: '#', text: 'Item 5'},
    {href: '#', text: 'Item 6'},
    {href: '#', text: 'Item 7'},
    {href: '#', text: 'Item 8'},
    {href: '#', text: 'Item 9'},
  ]

  return (
    <NavList>
      <NavList.Item href="#" aria-current="page">
        Item 1
      </NavList.Item>
      <NavList.Item href="#">Item 2</NavList.Item>
      <NavList.Item href="#">Item 3</NavList.Item>
      <NavList.GroupExpand pages={2} label="Show more" items={items} />
    </NavList>
  )
}

```

[Group example (storybook)](https://primer-b750586425-13348165.drafts.github.io/storybook/?path=/story/components-navlist--with-group-expand)

#### New

* Adds new component `NavList.ShowMoreItem`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
Questions for reviewers!

* Should we rename `expand` with something else? I am indifferent to the name, but I do think that there are other possible names that could work.
  * collapsible
  * disclosure
  * expandable
  * ...etc, (note: some of the names might imply that the control can be "collapsed" when it is expanded, which isn't the case in the **current** proposed API.
* Would we ever want to support the control to be collapsible? Currently in PVC, [the control does not collapse](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/beta/nav_list/show_more_item) when expanded and [custom implementations of `NavList` with a "show more"](https://github.com/github/accessibility-audits/issues/4880#issuecomment-1706758444) do not collapse.
  * For now, the goal of this PR is parity with PVC and to unblock teams, of which only the "expand" functionality is required, but would love to hear any thoughts on this! 
* I added the ability to put expands inside of expands, which allows you to "nest" expands. This was added so that we'd have [parity with PVC and how it handles "pages"](https://primer-lookbook.github.com/view-components/lookbook/inspect/primer/beta/nav_list/show_more_item) (second group example).

packages/react/src/NavList/NavList.tsx
```diff
@@ -275,6 +275,52 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+      <ActionList.Item as="button" aria-expanded="false" ref={forwardedRef} onClick={expandList} {...props}>
+        {label}
+        <ActionList.TrailingVisual>
+          <PlusIcon />
+        </ActionList.TrailingVisual>
+      </ActionList.Item>
+    </Box>
+  ) : (
+    React.Children.map(children, (child, index) => {
+      const isFirstChild = index === 0 ? {'data-target--expand-focus': true} : null
```
TylerJDev, line 317: I'm curious if there's a better way to go about this? The alternative is to utilize a `ref` and pass that to the cloned child - but we'd need to ensure any `ref` passed by the consumer takes priority, while storing that reference somewhere 🤔 

The alternative is the current solution, which is utilizing the `ref` on the expand button to target the container, as `children` cannot have a wrapper (e.g. `<Box ref={ref}>{children}</Box>`). We're essentially adding a `data-*` attribute to the first child cloned, and focusing that when `expanded === true`. 

Would love to hear other ideas, as I am probably missing a few! :eyes:

TylerJDev:
Would love any and all reviews from the team! This PR is serving mainly as a proposal to the `NavList` API, to add [a new pattern/component](https://github.com/github/primer/issues/2637). :grin:

Cc: @primer/engineer-reviewers

packages/react/src/NavList/NavList.tsx
joshblack: There are two goals for me that come up for this kind of thing:

- Goal: colocate the effect (focus management) with the user interaction (onClick in this case)
- Sub-goal: avoid use of React.Children to promote composition 

The sub-goal isn't necessarily deal breaking since we already do this in the library, but if we could find ways to avoid this technique it will help us out in the long run. The core limitation of this technique is that it forces strict Parent > Child relationships and if people wrap components this composition will break the intended functionality.

First off, curious to hear what you think of these goals and if they align or not 👀 

With these in mind, a couple of approaches came to mind:

- useEffect that runs based on `expanded` (potential for focus-stealing behavior if we're not careful)
- Focus in `onClick` using `flushSync()`
- Focus in `onClick` using a helper that runs a callback in an effect
- Mutation observer on the outer list (probably not ideal since the code for focus is very removed from the actual item onClick)

The hope is that we could write something like:

```tsx
<button
  type="button"
  onClick={() => {
    setExpanded(true);

    // focus item
  }}
/>
```

Probably the biggest challenge with this is figuring out _which items are focus candidates_ since we don't have a common ancestor across elements.

One idea for this that avoids `React.Children` is to use a context provider at the `ShowMoreItems` level. For example:

```tsx
const ShowMoreItemContext = React.createContext<{
  groupId: string;
} | null>(null);
```

In each item, we could then add a data attribute, like in:

```tsx
function Item({ children }: React.PropsWithChildren<{}>) {
  const groupContext = React.useContext(ShowMoreItemContext);
  return (
    <li data-show-more-group={groupContext ? groupContext.groupId : undefined}>
      <button type="button">{children}</button>
    </li>
  );
}
```

Then in our handler we could ``querySelector(`[data-show-more-group="${groupId}"`)`` to get the first element in our group to focus.

```tsx
<button
  type="button"
  onClick={() => {
    setExpanded(true);
    const item = document.querySelector(`[data-show-more-group="${groupId}"`);
    item?.focus();
  }}
/>
```

But the challenge with this is that we have not rendered yet so these elements aren't in the document 😅 This is where things like `flushSync()` could come into play:

```tsx
<button
  type="button"
  onClick={() => {
    flushSync(() => {
      setExpanded(true);
    });
    const item = document.querySelector(`[data-show-more-group="${groupId}"`);
    item?.focus();
  }}
/>
```

Or the effect handler:

```tsx
React.useEffect(() => {
  if (expanded) {
    const item = document.querySelector(`[data-show-more-group="${groupId}"`);
    item?.focus();
  }
}, [expanded, groupId]);
```

_Note: this has a chance to steal focus accidentally due to it being re-run due to dependencies. It'd be great if we could fire this as a one-off and guard against it in the future_

Or some helper that allows us to run code with effects inline when the effects of a component run:

```tsx
<button
  type="button"
  onClick={() => {
    setExpanded(true);
    focus(() => {
      return document.querySelector(`[data-show-more-group="${groupId}"`);
    });
  }}
/>
```

---

Sorry that this ended up being so long 😂 You gave me a lot to chew on. I'd love to hear what you think and learn more about what you've tried out / what would work best here from your perspective 👀

TylerJDev: This is great stuff, thank you! 

I updated my PR to utilize context instead! I think this is a much better way to handle things, so thank you for the recommendation! I'm curious about the different routes we might have when it comes to focus management. I'm still utilizing `useEffect` to handle things, but I agree that there's a possibility that focus might be set unexpectedly based on certain conditions.

I'm wondering if utilizing `useRef` could be a good way to guard against dependencies causing `.focus()` being called again unexpectedly. The ref would only store `true` or `false` based on if the target has been focused or not, with the permanent value being `true` once focused. 

[I created a quick example of utilizing `useRef`](https://github.com/primer/react/pull/4686/commits/3c0c94259decd22eba88563430eae274031c46e8), and I'm wondering if there's still potential risk here 🤔

joshblack: @TylerJDev oh 100%, I think the `useRef` technique would work great here 👍 

I also opened up: https://github.com/primer/react/pull/4715 as a proposal for extending `useFocus()` to support this use-case too that I'd love to get your thoughts on.

TylerJDev:
> I'd love to try and come up with a descriptive name but it's definitely challenging 🤔 Maybe we could use `NavList.ShowMoreItem` to align with PVC? Which I think has the method: `with_show_more_item`

I think going with parity is a good idea! I renamed it to `ShowMoreItem` so that we're more in sync with the PVC `NavList`. Thank you for the suggestion! :grin:

packages/react/src/NavList/NavList.tsx
```diff
@@ -288,6 +290,54 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
   )
 }
 
+export type NavListShowMoreItemProps = {
+  children: React.ReactNode
+  label: string
```
joshblack, line 295: Could this have a default value of "Show more" or is there really not a good default for the label that is used most of the time?

packages/react/src/NavList/NavList.tsx
```diff
@@ -288,6 +290,54 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+          {...props}
+        >
+          {label}
+          <ActionList.TrailingVisual>
+            <PlusIcon />
+          </ActionList.TrailingVisual>
+        </ActionList.Item>
+      </Box>
+    ) : (
+      <ItemWithinGroup.Provider value={{groupId}}>{children}</ItemWithinGroup.Provider>
```
joshblack, line 336: Could we memoize this value since we're passing it to a context provider? Would help with unnecessary re-renders since this is an object defined inline.

packages/react/src/NavList/NavList.tsx
```diff
@@ -288,6 +290,54 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+const ShowMoreItem = React.forwardRef<HTMLButtonElement, NavListShowMoreItemProps>(
+  ({label, children, ...props}, forwardedRef) => {
+    const [expanded, setExpanded] = React.useState(false)
+    const targetFocused = React.useRef(false)
+
+    const groupId = useId()
+
+    React.useEffect(() => {
+      if (expanded && !targetFocused.current) {
+        const focusTarget: HTMLAnchorElement | null = document.querySelector(`[data-show-more-group-id="${groupId}"]`)
```
joshblack, line 309: Will this always be an `HTMLAnchorElement` or could it be another type? Couldn't quite tell from `NavList` if it forces `<a>` or not so wanted to double-check 👀

.changeset/many-rivers-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Adds new sub-component to `NavList`, `NavList.Expand`
```
joshblack, line 5:
```suggestion
Add `NavList.ShowMoreItem` component to support showing more content within a `NavList`.
```

Just an idea for the changelog, really just `NavList.Expand` -> `NavList.ShowMoreItem` is all that's needed.

joshblack (commented):
Looks great! Just left a couple of comments/questions. Super weird question, but we don't really have a way to mark this as "experimental" since it's a sub-component, do we? 😅

packages/react/src/NavList/NavList.tsx
TylerJDev: We definitely could! I feel like most use cases will have context for what the items in the `NavList` are for. So adding a default "Show more" shouldn't hurt, as we allow consumers to customize it further if more context is needed.

packages/react/src/NavList/NavList.tsx
TylerJDev: I believe so. We don't seem to force it as `NavList.Item as="button"` works, but it's not officially supported (e.g. styles will most likely be incorrect if it's anything but a link `<a>`). I _think_ it's a safe bet to rely on the `HTMLAnchorElement`, but we could make it more generic!

TylerJDev:
> Super weird question, but we don't really have a way to mark this as "experimental" since it's a sub-component, do we? 😅

I don't think we do, but maybe this is something we'd want to have for future sub-components? 🤔

TylerJDev:
Pinging @primer/engineer-reviewers for any additional reviews!

packages/react/src/NavList/NavList.tsx
```diff
@@ -288,6 +290,57 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+      if (expanded && !targetFocused.current) {
+        const focusTarget: HTMLAnchorElement | null = document.querySelector(
+          `[data-show-more-group-id="${groupId.id}"]`,
+        )
+
+        if (focusTarget) {
+          focusTarget.focus()
+          targetFocused.current = true
+        }
+      }
```
joshblack, line 318: Wanted to ask if this could run in the `onClick` handler instead of us needing to synchronize this with an effect 👀 The benefit of tying this to the handler would be that focus would shift naturally due to interacting with the button instead of potentially being stolen if one the dependencies of the effect change that is not tied to a user interaction

joshblack (commented):
This is looking great! 🔥 Just had one comment around the effect to see what you thought 👀

packages/react/src/NavList/NavList.tsx
TylerJDev: We could, but I think the only issue is the timing, as the node we want to focus once expanded doesn't exist yet 🤔. 

Something like the following works, but I'm not sure if it's counterintuitive or not 😅 

```tsx
onClick={() => {
  flushSync(() => {
    setExpanded(true)
  })
  expandItem()
}}
```

lesliecdubs:
👋🏻 @TylerJDev wanted to check in as I noticed this one has been sitting without action for a while. What's the next step needed here? Have we gotten the necessary alignment from Primer Design and the rest of the engineering team on the addition of the ["more" feature](https://github.com/github/primer/issues/2637)?

TylerJDev:
Hey @lesliecdubs! This is pretty much finish, I just haven't had time to come back to the [last suggested change](https://github.com/primer/react/pull/4686#discussion_r1771434035). As for alignment with design, I'm wondering what's the best way to go about that?

lesliecdubs:
Thanks for following up @tylerjdev. I'll leave you to the final change.

> As for alignment with design, I'm wondering what's the best way to go about that?

Typically, pattern proposal issues like https://github.com/github/primer/issues/2637 go to a Primer Patterns meeting for discussion and approval. Perhaps you could add yourself to the next meeting invite and agenda, or if that doesn't work perhaps the Primer Design FR could weigh in on the PR.

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,6 +415,104 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+    const [currentPage, setCurrentPage] = React.useState(0)
+    const targetFocused = React.useRef(currentPage)
+    const enabled = useFeatureFlag('primer_react_css_modules_team')
+
+    const id = useId()
+    const groupId = React.useMemo(() => ({id}), [id])
+    const childCount = React.Children.count(children)
+
+    React.useEffect(() => {
+      if (expanded && targetFocused.current !== currentPage) {
```
TylerJDev, line 438: I'm curious if this conditional is enough to ensure that focus won't be stolen upon re-renders, [as mentioned in this comment](https://github.com/primer/react/pull/4686#discussion_r1661384337). 

When a given "show more" button is expanded, we store the "page" that was expanded in the `useRef`. So if there's 2 pages, and a user expands once, the page placed in `targetFocused` would be 1, and the value in `currentPage` would also be 1. This means that if there's a re-render, the conditional (`targetFocused.current !== currentPage`) wouldn't apply, as both `targetFocused` and `currentPage` values are the same. 

We're essentially checking if the most recent page has been expanded, if it hasn't, then we'll focus the first element of the new "group", and store that page's value in the `targetFocused` useRef.

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,6 +415,104 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+
+          focusTarget[pages ? nextItemToFocus : focusTarget.length - childCount].focus()
+          targetFocused.current = currentPage
+        }
+      }
+    }, [expanded, groupId, currentPage, childCount, pages])
+
+    return (
+      <>
+        {expanded && (
```
joshblack, line 455: Small style suggestion, we could a ternary here instead of the short-circuit syntax to avoid rendering any unexpected values (reasoning from: https://eslint-react.xyz/docs/rules/no-leaked-conditional-rendering)

```suggestion
        {expanded ? (...) : null}
```

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,6 +415,104 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+        const focusTarget: HTMLElement[] = Array.from(
+          document.querySelectorAll(`[data-show-more-group-id="${groupId.id}"]`),
+        )
+
+        if (focusTarget.length) {
+          const itemsPerPage = Math.ceil(childCount / pages)
+          const nextItemToFocus = itemsPerPage * currentPage - itemsPerPage
+
+          focusTarget[pages ? nextItemToFocus : focusTarget.length - childCount].focus()
+          targetFocused.current = currentPage
```
joshblack, line 448: I'm a little hesitant with a programmatic `focus()` in a `useEffect` since it might not sequence with a user interaction (e.g. if some unrelated thing occurs, like child count, we would steal focus unintentionally). Would it be possible to have focus behavior occur during a user interaction?

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,6 +415,104 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
   )
 }
 
+export type NavListShowMoreItemProps = {
+  children: React.ReactNode
+  label?: string
+  pages?: number
+} & SxProp
```
joshblack, line 422: Feel free to just remove any `sx` stuff if you don't want to have to deal with toggling behind the flag!

```suggestion
}
```

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,6 +415,104 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+          targetFocused.current = currentPage
+        }
+      }
+    }, [expanded, groupId, currentPage, childCount, pages])
+
+    return (
+      <>
+        {expanded && (
+          <ItemWithinGroup.Provider value={groupId}>
+            {React.Children.toArray(children).filter((child, index, arr) => {
```
joshblack, line 457: Would reading from children in this way work if a fragment is used or if someone has a wrapper component that then renders multiple items? e.g.

```tsx
<NavList.ShowMoreItem label="Show more">
  <CustomNavItems /> {/* This might return three items */}
  <NavList.Item>...</NavList.Item>
  <NavList.Item>...</NavList.Item>
  <NavList.Item>...</NavList.Item>
</NavList.ShowMoreItem>
```

Or would this just be limited to direct descendant NavList.Item's?

packages/react/src/NavList/NavList.tsx
TylerJDev: Oh wow, TIL. Thank you for the link, will start using this pattern moving forward 🙏

packages/react/src/NavList/NavList.tsx
TylerJDev: You might've already read this, but curious on what you think about this: https://github.com/primer/react/pull/4686#discussion_r1932858352. Would this be enough to safeguard us from focus being stolen? I'm wondering if there are gaps I might not be seeing.

I can play around with the implementation more! The only thing I'm running into is getting the element after it is placed in the DOM, and not before 🤔 I tried other methods like `flushSync`, but am a bit hesitant on it since we don't use it elsewhere, and it didn't work exactly as I expected 😅

packages/react/src/NavList/NavList.tsx
TylerJDev: I believe it should still render! I went ahead and tested this, let me know if this isn't what you're envisioning!

```tsx
const CustomNavItems = () => {
  return (
    <>
      <NavList.Item href="#">Item 2D</NavList.Item>
      <NavList.Item href="#">Item 2E</NavList.Item>
      <NavList.Item href="#">Item 2F</NavList.Item>
    </>
  )
}

const NavListGroup = () => {
  return (
    <NavList>
      <NavList.Group title="Group 2">
        <NavList.Item href="#">Item 2A</NavList.Item>
        <NavList.Item href="#">Item 2B</NavList.Item>
        <NavList.Item href="#">Item 2C</NavList.Item>
        <NavList.ShowMoreItem pages={2} label="Show">
          <CustomNavItems />
          <NavList.Item>Item 2G</NavList.Item>
          <NavList.Item>Item 2H</NavList.Item>
          <NavList.Item>Item 2I</NavList.Item>
        </NavList.ShowMoreItem>
      </NavList.Group>
    </NavList>
  )
}
  ```
  
It seems to have correctly rendered all the children as you'd expect. The only issue I see now is that focus is set to "Item 2D", instead of "Item 2G" when expanded. Will take a look on why this is 👀

Comment (TylerJDev) [reply packages/react/src/NavList/NavList.tsx]: Ah, I misunderstood! I see the issue now. We calculate how many items to show at a given time based on the amount of children. If there are 100 children in a fragment then it will consider the fragment as one child, rather than 100.

I restricted the children to only being `NavList.Item`. I didn't really want to restrict children based on type, but I don't envision many cases where we'd want anything but `NavList.Item`. This also solves the fragment issue. We utilize a very similar pattern for `SubNav` in this component, so this isn't too different.

Curious on your thoughts here 😄

File: packages/react/src/NavList/NavList.tsx (hunk 7 above)
Comment (TylerJDev) [reply packages/react/src/NavList/NavList.tsx]: @joshblack, you mentioned if we could potentially set focus during user interaction, I'm curious on what type of interaction that could be? 🤔

Comment (joshblack) [reply packages/react/src/NavList/NavList.tsx]: @TylerJDev is it possible to do this on the `onClick` for expand or not really?

File: packages/react/src/NavList/NavList.tsx (hunk 11 above)
Comment (joshblack) [reply packages/react/src/NavList/NavList.tsx]: @TylerJDev sadly this is a pattern that we're trying to move away from since we're seeing folks downstream wrap the component in some way (either with the fragment case or just a wrapper around a specific component)

It's definitely a pattern that exists in PRC today though, like you said, but the more we add the harder it will be to dig ourselves out of this hole 😵

File: packages/react/src/NavList/NavList.tsx (hunk 7 above)
Comment (TylerJDev) [reply packages/react/src/NavList/NavList.tsx]: We _could_ utilize `setTimeout` here. Basically, using `setTimeout` with a timeout of 0 (or none at all) seems to query the elements after they are rendered in the DOM. 

This is a bit hacky though. I believe we already utilize this pattern [in other places such as `TreeView`](https://github.com/primer/react/blob/main/packages/react/src/TreeView/TreeView.tsx#L699), and [`LabelGroup`](https://github.com/primer/react/blob/main/packages/react/src/LabelGroup/LabelGroup.tsx#L239), so it _could_ work. Not ideal though.

Comment (TylerJDev) [reply packages/react/src/NavList/NavList.tsx]: Was able to utilize `flushSync` here! Seems to work as expected 😁

File: packages/react/src/NavList/NavList.tsx [hunk 12]
```diff
@@ -413,7 +414,141 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
-Group.displayName = 'NavList.Group'
+// ----------------------------------------------------------------------------
+// NavList.GroupExpand
+
+type GroupItems = {
+  text: string
+  trailingVisual?: Icon | string
+  leadingVisual?: Icon | string
+  trailingAction?: ActionListTrailingActionProps
+  'data-expand-focus-target'?: string
```
TylerJDev, line 425: I would love to not have to express this in `GroupItems`, as we *don't* want consumers adding this `data-*` attribute themselves. The only reason I have it here is so TS doesn't yell at me for adding it in `itemArr['data-expand-focus-target'] = focusTarget`.

TylerJDev:
@joshblack, should be ready for another review! I moved towards the data API that you suggested, and it seems to work great! Let me know if you have any suggestions or ways to improve it! 😁

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,7 +414,141 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+} & Omit<NavListItemProps, 'children'>
+
+export type NavListGroupExpandProps = {
+  label?: string
+  pages?: number
+  items: GroupItems[]
+  renderItem?: (item: {text: string}) => React.ReactNode
```
joshblack, line 432: For the signature here, do we need to extend it to include other things that might be passed along?

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,7 +414,141 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+            {items.map((itemArr, index) => {
+              const {
+                text,
+                trailingVisual: TrailingVisualIcon,
+                leadingVisual: LeadingVisualIcon,
+                trailingAction,
+                ...rest
+              } = itemArr
+              const {icon, label: actionLabel, ...actionProps} = trailingAction || {}
+              const focusTarget = index === focusTargetIndex ? groupId : 'false'
```
joshblack, line 464:
```suggestion
              const focusTarget = index === focusTargetIndex ? groupId : undefined
```

Does it make sense to have the value be `undefined` so that the data attribute is not set or do we want it to be a false-y value intentionally?

TylerJDev: I think either way is fine! I'm thinking that if it's not set to the group ID, then we don't really care about that item. We could have the data attribute persist, but I'm not sure if there's much of a difference to how it is now? 🤔

packages/react/src/NavList/NavList.tsx
```diff
@@ -413,7 +414,141 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+    const teamEnabled = useFeatureFlag('primer_react_css_modules_team')
+    const staffEnabled = useFeatureFlag('primer_react_css_modules_staff')
+
+    const {amountToShow, focusTargetIndex} = React.useMemo(() => {
+      const itemsPerPage = items.length / pages
+      const amountToShow = pages === 0 ? items.length : Math.ceil(itemsPerPage * currentPage)
+      const focusTargetIndex = currentPage === 1 ? 0 : amountToShow - Math.floor(itemsPerPage)
+
+      return {amountToShow, focusTargetIndex}
+    }, [pages, items, currentPage])
```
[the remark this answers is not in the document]
TylerJDev: This looks good! Went ahead and tested it and it performs the same!

packages/react/src/NavList/NavList.tsx
TylerJDev: Added `GroupItem`, as that's what we're passing to it either way.

packages/react/src/NavList/NavList.tsx
joshblack: @TylerJDev totally, really this was just to omit the data attribute if this value is `false` since we're not going to use it. To me this can be helpful (basically don't add things to the DOM that we don't need to help with clarity) but totally get if they are equivalent or if it doesn't seem that helpful 👍

joshblack (approved):
Congrats on putting this together! Just left a comment but approving since it's non-blocking 👍

packages/react/src/NavList/NavList.tsx
TylerJDev: AH, I had it confused. I assumed the `undefined` was how it was now. Not sure how I misread that 😆 Yeah, I think omitting the `data-*` is preferred!

PR #4687, opened 2024-06-19
Add CheckboxGroup.Choices and RadioGroup.Choices subcomponents which format their children as a list

JeffreyKozik:
Followup to https://github.com/github/github/pull/329358#discussion_r1647556804 and this discussion in [Slack](https://github.slack.com/archives/C01L618AEP9/p1718889234728939).

#### New

Created the `CheckboxGroup.Choices` and `RadioGroup.Choices` subcomponents. 

Currently, our Radio Group and Checkbox Groups aren't lists
Screenshot of the Checkbox Group component on Primer React's Storybook showing that it isn't a list

But according to the [ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox/), these groups should be lists:

> To enable assistive technology users to perceive the set of checkboxes as a list of four items, each div element that serves as a checkbox is contained within a li element contained by a ul element.

Additionally, one of [GitHub Accessibility's "trusted component libraries"](https://github.com/github/accessibility-playbook/blob/main/content/industry-references.md#patterns) Polaris, by Shopify, follows the APG's guidelines of making their [Checkbox Groups lists](https://polaris.shopify.com/components/selection-and-input/choice-list?example=choice-list-with-multi-choice): https://github.com/Shopify/polaris/blob/902db588f6a317f4d6a5a76f4814bdb1f80395db/polaris-react/src/components/ChoiceList/ChoiceList.tsx#L109-L146.

I thought of three different approaches for Primer

1. **Keep our current implementation** We currently put the burden on the developer to add `<ul>` and `<li>` elements to their code. The positive of this approach is it gives the developer maximal flexibility no matter how they're using `CheckboxGroup` or `RadioGroup`. However, the downside is that it means the developer has to manually add these elements nearly every single time, otherwise their code isn't fully accessible. A cursory search of GitHub's codebase shows that is almost never implemented.
2. **Automatically wrap each child in an `li` element** We could do it in the code here:
https://github.com/primer/react/blob/e957fef6785277b1c30684b146a0f591d8c6b3e2/packages/react/src/internal/components/CheckboxOrRadioGroup/CheckboxOrRadioGroup.tsx#L125-L134
This removes the onus from the developer to have to manually add the `<ul>` and `<li>` elements to their code. The positive of this approach is that the accessibility logic is completely abstracted away into the `CheckboxGroup` and `RadioGroup` components. However, the downside is that doing this would be a breaking change for some implementations. For example, this code wraps Radio buttons within an ActionList:
https://github.com/github/github/blob/b61537935e3cbdfa0e05a45656e854e24ae674d2/ui/packages/issue-viewer/components/RadioActionListItem.tsx#L19-L28
So automatically adding a `ul` element here would be wrong because an ActionList is already a list:
https://github.com/primer/react/blob/e957fef6785277b1c30684b146a0f591d8c6b3e2/packages/react/src/ActionList/List.tsx#L15
https://github.com/primer/react/blob/e957fef6785277b1c30684b146a0f591d8c6b3e2/packages/react/src/ActionList/List.tsx#L62-L70
3. The approach I took: **Create the `CheckboxGroup.Choices` and `RadioGroup.Choices` subcomponents** While this approach does still require the developer to explicitly use these subcomponents, it's considerably easier and more consistent than them having to use `ul` and `li` elements and apply custom styling. Following this approach, our Radio Group and Checkbox Groups are now lists
Screenshot of the Checkbox Group component on Primer React's Storybook showing that, with the new code from this PR, it is now a list

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


JeffreyKozik:
After further discussion on [Slack](https://github.slack.com/archives/C01L618AEP9/p1718889234728939), accessibility experts @ericwbailey and @patrickhlauke have mentioned that using `ul` and `li` for Radio groups and Checkbox groups is actually not needed and can be counterproductive.

Per  @ericwbailey 
> Lists are not required for radio groups and can circumstantially create unnecessary and unneeded verbosity. Screen readers have separate controls for traversing and enumerating radio groups, as well

PR #4688, opened 2024-06-20
docs(versioning): update guide with changes to event handlers

joshblack:
This came up in our release cycle this week and I wanted to document the behavior here to make sure we're aligned 👀 

This updates our versioning guide with a specific case around making event handlers broaden, specifically when changing the element type on which the event occurs.

#### Changed

- Add section to the versioning docs in the repo

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

This is a change to documentation

### DIFF
--- a/contributor-docs/versioning.md
+++ b/contributor-docs/versioning.md
@@ -16,6 +16,7 @@
   - [A component includes a landmark role](#a-component-includes-a-landmark-role)
   - [A component no longer includes a landmark role](#a-component-no-longer-includes-a-landmark-role)
   - [The element onto which props are spread is changed](#the-element-onto-which-props-are-spread-is-changed)
+  - [The element type in an event handler becomes broader](#the-element-type-in-an-event-handler-becomes-broader)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 <!-- prettier-ignore-end -->
@@ -56,6 +57,7 @@ For a full list of releases, visit our [releases](https://github.com/primer/reac
 |               | A prop is deprecated                                                                                                                          | `minor`             |
 |               | A prop is removed                                                                                                                             | `major`             |
 |               | [The element onto which props are spread is changed](#the-element-onto-which-props-are-spread-is-changed)                                     | potentially `major` |
+|               | [The element type in an event handler becomes broader](#the-element-type-in-an-event-handler-becomes-broader)                                 | `major`             |
 | Package       | A dependency is added                                                                                                                         | `minor`             |
 |               | A dependency is removed and it does not affect the public API of the package                                                                  | `minor`             |
 |               | A dependency is removed and it does affect the public API of the package                                                                      | `major`             |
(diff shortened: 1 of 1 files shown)


contributor-docs/versioning.md
```diff
@@ -246,3 +248,65 @@ function Component(props: Props) {
 ```
 
 </details>
+
+### The element type in an event handler becomes broader
+
+semver bump: **major**
```
broccolinisoup, line 254: Just to note that I am making this comment ActionList semantics issue in mind 🙂 

It feels to me that types, especially the event type changes in that PR, feels a bit blur to me in terms of semver. Because on the consumers side, the event types are `event: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>` and since it doesn't overlap with `event: React.MouseEvent<HTMLElement> | React.KeyboardEvent<HTMLElement>`, it fails. However if the types were less specific, or in other words, maybe not directly rely on the component details i.e. expecting the event to be triggered on the li element, that changes wouldn't break anything. Based on this, would this be more like "potentially breaking" ? 🤔 I would love to learn more about the best practises on the consumer side as well if you have thoughts on that! Is this a good practise to "hard code" the event types like currently or can they be inferred from what is defined in the component? 🤔

broccolinisoup (commented):
Thanks so much for pushing this PR! Types are hard, versioning types are harder 😆 Left a comment and questions, looking forward to hearing your thoughts 🙌🏻

contributor-docs/versioning.md
joshblack: > However if the types were less specific, or in other words, maybe not directly rely on the component details i.e. expecting the event to be triggered on the li element, that changes wouldn't break anything. Based on this, would this be more like "potentially breaking" ? 🤔

I think you're spot on pointing out that it 100% depends on what the caller is doing with the types. As you pointed out, this is definitely a case where if the caller had a less strict type or used the type from the component prop interfaces there wouldn't be any breakage that we would observe.

In general, I think broadening a type could be considered "potentially breaking". I think we have scenarios where it is okay like in:

```diff
type ComponentProps = {
-  variant?: 'a' | 'b',
+  variant?: 'a' | 'b' | 'c',
};
```

But then we have cases like this with an event handler where it will cause issues downstream:

```diff
-onClick: (gesture: 'a' | 'b') => void,
+onClick: (gesture: 'a' | 'b' | 'c') => void,
```

These seem like breaking changes to me since the caller cannot use the update without modifying their code. I think that the example for this with the generic would fall under this scenario since folks would have to update their code for it.

broccolinisoup: Thanks Josh, this is great! I love the comparison you made above between the two examples. 

>These seem like breaking changes to me since the caller cannot use the update without modifying their code.

I like how you frame it. I think most of the time, this is a great indication to use to assess the versioning unless the consumer code is misused in the first place but this is another story 😅 

Thanks for pushing this PR again, I appreciate talking about it and keeping it on the docs 🙏🏻

broccolinisoup (approved):

PR #4691, opened 2024-06-20
refactor(VisuallyHidden): update base element from div to span

joshblack:
Update our internal utility `VisuallyHidden` to use a `<span>` by default over a `<div>`. This is due to the fact that a `<div>` is not allowable in all contexts, for example a `<div>` cannot be rendered within a `<p>`. As a result, this helper may not work as intended if a component is within a specific context. Moving over to a `span` by default will help with this as it it allowed in all contexts (hopefully).

#### Changed

- Change base element from `div` to `span` for `VisuallyHidden`
- Update Spinner to no longer have to specify `span` in `as`

### Rollout strategy

- [x] Minor release

### DIFF
--- a/.changeset/thick-jars-count.md
+++ b/.changeset/thick-jars-count.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Update internal VisuallyHidden helper to use a `span` by default over a `div` to support more nesting scenarios by default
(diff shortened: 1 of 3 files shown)


TylerJDev (approved):

broccolinisoup (approved):

siddharthkp (approved):
LGTM!

Do you suspect any layout related changes for siblings when moving from a block element to span element?

joshblack:
@siddharthkp I'm not sure honestly 🤷 curious what you all think. Since this is hidden and position: absolute I would assume that it's not participating in layouts but I am very aware of the fact that I could be missing a scenario where that is true 😅

PR #4703, opened 2024-06-27
fix(FilteredActionList): label listbox by input

joshblack:
Closes https://github.com/github/primer/issues/3408

Fixes https://dequeuniversity.com/rules/axe/4.7/aria-input-field-name for `FilteredActionList` and `SelectPanel` by labeling the listbox by the input

#### Changed

- Add `id` to input for FilteredActionList
- Label listbox by input id

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- [ ] Verify that the axe violation for this no longer occurs


owenniblock (approved):
This looks good to me. Is it worth asking Tetralogical to do some screenreader testing before it gets released?

PR #4705, opened 2024-06-28
test(SelectPanel): update tests to account for more of API

joshblack:
As we're working more on the v1 of SelectPanel, I wanted to add some additional tests to the test suite.

This is not complete coverage but figured I'd add what I could this week and revisit another time for what was missed

#### Changed

- Update tests for Select Panel

### Rollout strategy

This is a change that adds tests to Select Panel

### DIFF
--- a/packages/react/src/SelectPanel/SelectPanel.test.tsx
+++ b/packages/react/src/SelectPanel/SelectPanel.test.tsx
@@ -4,13 +4,6 @@ import {SelectPanel, type SelectPanelProps} from '../SelectPanel'
 import {userEvent} from '@testing-library/user-event'
 import ThemeProvider from '../ThemeProvider'
 
-// actions
-// open
-// type / filter
-// toggleSelection
-// move to next option
-// move to previous option
-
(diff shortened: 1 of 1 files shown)


joshblack:
bump @siddharthkp when you get a sec 👀

broccolinisoup (approved):
Thanks so much for adding these tests 🙏🏻 

Question: I assume these are the current behaviours since all tests pass. Is there any behaviour that we know for sure that it will change after we implement the accessibility remediations? If there are some, we might want to add comments on them to help us refactor the tests later. 

Let me know if I am not making sense!

joshblack:
@broccolinisoup my understanding since we're keeping the combobox-related functionality is that it _should_ mostly be the same behavior, if I remember right 👀

broccolinisoup:
> @broccolinisoup my understanding since we're keeping the combobox-related functionality is that it _should_ mostly be the same behavior, if I remember right 👀

Oh this is great! Thanks for sharing that.

PR #4710, opened 2024-07-01
fix(Blankslate): update styles to prevent SSR hydration mismatch

joshblack:
Update Blankslate to address an SSR hydration mismatch issue that was occurring. When the styles were present as `children` of a `<style>` tag, they were being serialized in a way that transformed the `'` to escaped sequences leading to a hydration mismatch. Updating this to use `dangerouslySetInnerHTML` addresses this issue.

I also updated the styles to use fallbacks, where appropriate, so the component would render as expected without an explicit import to primitives.

#### Changed

- Update the `style` tag in `Blankslate` to better support SSR
- Add fallback values to CSS Custom Properties used in `Blankslate`.

### Rollout strategy

- [x] Patch release


joshblack:
bump @TylerJDev!

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -26,16 +26,17 @@ const StyledBlankslate = styled.div`
...
+
     display: grid;
     justify-items: center;
     padding: var(--blankslate-outer-padding-block) var(--blankslate-outer-padding-inline);
   }
 
   .Blankslate[data-spacious='true'] {
-    --blankslate-outer-padding-block: var(--base-size-80);
-    --blankslate-outer-padding-inline: var(--base-size-40);
+    --blankslate-outer-padding-block: var(--base-size-80, 5rem);
```
broccolinisoup, line 38: Nice!

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -128,7 +130,7 @@ function Blankslate({border, children, narrow, spacious}: BlankslateProps) {
         This is a workaround so we can use `@container` without upgrading `styled-components` to 6.x
         See [this comment](https://github.com/primer/react/pull/3869#discussion_r1392523030) for more info
       */}
-      <style type="text/css">{BlankslateContainerQuery}</style>
+      <style type="text/css" dangerouslySetInnerHTML={{__html: BlankslateContainerQuery}} />
```
broccolinisoup, line 133: This is very smart 🔥 

I also think it is obscure to cause hydration mismatch when it is passed as children 😄

broccolinisoup (approved):

PR #4711, opened 2024-07-01
Remove live region from `FormControl` validation

TylerJDev:
Unblocks: https://github.com/github/accessibility-audits/issues/6883, https://github.com/github/accessibility-audits/issues/6881, https://github.com/github/accessibility-audits/issues/5728

We've received [feedback from the Accessibility team](https://github.com/github/accessibility-audits/issues/6883#issuecomment-2115858770) that it may be less than ideal for the `FormControl` component to have a live region for validation, as the `aria-describedby` [should suffice in most scenarios](https://github.com/github/accessibility-audits/issues/6883#issuecomment-2200728486). Consumers of this component may still provide a live region, but it will no longer be included by default. 

#### Removed

* Removes `aria-live` from `FormControl.Validation`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/grumpy-masks-relax.md
+++ b/.changeset/grumpy-masks-relax.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Removes live region from `FormControl` validation
--- a/packages/react/src/internal/components/InputValidation.tsx
+++ b/packages/react/src/internal/components/InputValidation.tsx
@@ -48,7 +48,6 @@ const InputValidation: React.FC<React.PropsWithChildren<Props>> = ({children, id
         },
         ...sx,
       }}
-      aria-live="polite"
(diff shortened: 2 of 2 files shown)


TylerJDev:
This removes the live region from `FormControl.Validation` completely, but I think it might be worth in the future to revisit and see how, or even if we'd want to make this opt-in utilizing `@primer/live-region-element`. At the very least we can add an example of a custom implementation of a live region + validation if decide this is a pattern that _could_ be used accessibly.

joshblack (approved):

PR #4712, opened 2024-07-02
Adding color algo for hex colors to label component

lukasoppermann:
This PR adds the color algorithm to transform hex colors to something similar to variants to the new `IssueLabel` component.

@joshblack and anybody else who wants to chime in, please let me know anything that is not perfect and how I should improve it. Happy to learn.


packages/react/src/drafts/IssueLabel/IssueLabel.tsx
```diff
@@ -43,8 +45,17 @@ export interface IssueLabelProps extends React.PropsWithChildren {
...
+  console.log('theme', theme)
+
   return (
-    <StyledLabel {...rest} data-size={size} data-variant={fillColor ? undefined : variant}>
+    <StyledLabel
+      {...rest}
+      data-size={size}
+      data-variant={fillColor ? undefined : variant}
+      style={fillColor ? getColorsFromHex(fillColor, resolvedColorScheme, '#fff') : undefined}
+    >
```
lukasoppermann, line 58: @joshblack I need to get the default backgroundColor `bgColor-default` for the current selected theme. How would you suggest I bet retrieve it?

joshblack: @lukasoppermann most likely the best way is what you have right now which is `useTheme()`. I think that no matter what we'll need to use a context provider to pipe this through since we can't read this value from anywhere during SSR. One thing I'm not sure about is how teams are setting this so that the correct theme is set during SSR so that we can avoid a flash between themes.

lukasoppermann: Okay. One _issue_ I am seeing is that the color is using a css variable. At the moment we still have a fallback value, but I need a hex value to calculate this. 

Do you think I should create an object with the background colors in here, like this

```js
const bgColors = {
  light: "#ffffff",
  dark: "#0d1117"
}
```

The issue I see here is that it is easy to fall out of sync with primitives if they get changed.

Do you have any better idea? I guess I could use just the two dark and light values above for all themes and if they change it would probably not be that much of a problem.

lukasoppermann: @joshblack I implemented my idea from above. Is there anything else I should change?

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * transforms a hex color provided by the user into a color object with background and text colors
+ * @param colorHex — the hex color provided by the user
+ * @param colorScheme — the color scheme the user has currently set
+ * @param isSelected
+ * @param bgColor — the background color from the selected theme, needed to calc the contrast for the colors
+ * @returns
+ */
+export const getColorsFromHex = (
+  colorHex: string,
```
joshblack, line 13: Would it help to have a type like ``type Hex = `#${string}`;`` to help with inputs to the function? Could then remove the validation we're doing at the beginning and then the function could just return a value.

Definitely isn't true type safety but seems to mirror the rough checking we're doing below.

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,138 @@
...
+  colorScheme = 'light',
```
joshblack, line 14: Is there a union type for this anywhere? Would be great to use that (or construct our own) if possible

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,138 @@
...
+  bgColor: string,
+): React.CSSProperties | undefined => {
+  if (typeof colorHex !== 'string' || colorHex.startsWith('#') === false) return undefined
+  // start values for light mode
+  let bgLightness = 96
+  let lightnessIncrement = -1
```
joshblack, line 20:
```suggestion
  let lightnessIncrement: 1 | -1 = -1
```

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,138 @@
...
+  const hsluv = hexToHsluv(colorHex)
+  let {l: lightness} = hsluv
+  const {h, s} = hsluv
+  // change lightness until contrast is reached
+  while (getContrast(colorHex, bgHex) < contrastRatio && lightness > 0 && lightness < 100) {
+    lightness += increment
+    colorHex = hsluvToHex({h, s, l: lightness})
+  }
+  // return hex color and ligthness
+  return {colorHex, lightness}
```
joshblack, line 96: Instead of moving between color formats and spending time converting between each, could we parse colors into HSL at the beginning, mutate the channels in those colors, and then output hex? Or would that not work?

packages/react/src/drafts/IssueLabel/IssueLabel.tsx
joshblack: Looks great 👍 The only thought was that it would be great if we could pull the values themselves from `@primer/primitives` but that would be outside of the scope of this PR.

joshblack:
@lukasoppermann for the test failures at the moment with VRT and AAT, we just need to update the story ids in the e2e tests since we renamed the component to IssueLabel from Label 👍

packages/react/src/drafts/IssueLabel/IssueLabel.tsx
lukasoppermann: Yes, I'd love to do this. Realistically the difference should not be so big, so if this has a speed impact we should not do it.

packages/react/src/drafts/IssueLabel/IssueLabel.features.stories.tsx
```diff
@@ -51,7 +51,7 @@ export const SizeLarge = () => <IssueLabel size="large">Issue label</IssueLabel>
 
 export const SizeSmall = () => <IssueLabel size="small">Issue label</IssueLabel>
 
-export const Hex = hex => <IssueLabel fillColor={hex}>Issue label</IssueLabel>
+export const Hex = (hex: string) => <IssueLabel fillColor={hex}>Issue label</IssueLabel>
```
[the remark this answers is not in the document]
lukasoppermann: Yes, love it. I though you folks had named it `hex` on purpose, which is why I did not change it.

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
lukasoppermann: I don't think so, at least I did not find any in the ThemeProvider

packages/react/src/drafts/IssueLabel/IssueLabel.stories.tsx
```diff
@@ -15,6 +15,10 @@ export const Playground: StoryObj<typeof IssueLabel> = {
   render: args => <IssueLabel {...args}>Issue label</IssueLabel>,
   args: {},
   argTypes: {
+    fillColor: {
+      hex: undefined,
```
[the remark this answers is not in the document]
lukasoppermann: So if I remove this and try to add it to `[IssueLabel.features.stories.tsx](https://github.com/primer/react/pull/4712/files/dfd78d2001ae018d2317615c4b7a3cc166312c20#diff-7a37e6ad59ba286ccff6c8971cd721aa7409de27de2c120a2e864c39bf0f1d36)` I don't get the correct color control in storybook.

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
lukasoppermann: I think the issue is that `getContrast` does not work with `HSLuv`. We could look at if we can use a different color model witha library that offers a contrast check out of the box?

lukasoppermann: I may be wrong, but it seems like all libraries do conversions under the hood, e.g. https://github.com/color-js/color.js/blob/3e97aa382b7a861409c8f1bdaf04fa7a42b1f5ee/src/contrast/WCAG21.js#L13 uses `getLuminance` which also does a conversion. 

What I am not sure about is: Is this lib going to be loaded at runtime? Because we need to run the code with the user input, right? If so, the diff between the size of the library may also be important.

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @param colorScheme — the color scheme the user has currently set
+ * @param isSelected
+ * @param bgColor — the background color from the selected theme, needed to calc the contrast for the colors
+ * @returns
+ */
+export const getColorsFromHex = (
+  colorHex: `#{string}`,
+  colorScheme: string = 'light',
+  bgColor: string,
+): React.CSSProperties | undefined => {
```
joshblack, line 16: Some ideas for the types on this to make them more specific:

```suggestion

type Hex = `#${string}`;

/**
 * transforms a hex color provided by the user into a color object with background and text colors
 * @param colorHex — the hex color provided by the user
 * @param colorScheme — the color scheme the user has currently set
 * @param isSelected
 * @param bgColor — the background color from the selected theme, needed to calc the contrast for the colors
 * @returns
 */
export const getColorsFromHex = (
  colorHex: Hex,
  colorScheme: 'light' | 'dark' = 'light',
  bgColor: Hex,
): React.CSSProperties | undefined => {
```

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
...
+    s = 70
+  }
+  /**
+   * creating a background color from the provided hex color
+   */
+  const {colorHex: backgroundColor, lightness: currentBgLightness} = getColorWithContrast(
+    hsluvToHex({h, s, l: bgLightness}),
+    bgColor,
+    1.2,
+    lightnessIncrement as 1 | -1,
```
joshblack, line 43: Is this needed anymore now that the `lightnessIncrement` has the type above? 👀 

```suggestion
    lightnessIncrement,
```

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
...
+    s = 80
+  }
+  /**
+   * creating a text color from with a contrast ratio of at least 4.5 to the generated background color
+   */
+  const {colorHex: textColor} = getColorWithContrast(
+    hsluvToHex({h, s, l: 50}),
+    backgroundColor,
+    ratio,
+    lightnessIncrement as 1 | -1,
```
joshblack, line 57: If no longer needed anymore 🤞 

```suggestion
    lightnessIncrement,
```

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
...
+  return {
+    '--label-bgColor-rest': backgroundColor,
+    '--label-bgColor-hover': hsluvToHex({h, s, l: currentBgLightness + 4 * lightnessIncrement}),
+    '--label-bgColor-active': hsluvToHex({h, s, l: currentBgLightness + 8 * lightnessIncrement}),
+    '--label-fgColor': textColor,
+    '--label-fgColor-hover': textColor,
+    '--label-fgColor-active': textColor,
+  } as React.CSSProperties
+  // backgroundColorHover: hsluvToHex({h, s, l: currentBgLightness + 4 * lightnessIncrement}),
+  //   backgroundColorPressed: hsluvToHex({h, s, l: currentBgLightness + 8 * lightnessIncrement}),
```
joshblack, line 69: Are these needed anymore?

```suggestion
```

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
joshblack: Thanks for clarifying, that totally makes sense about contrast checks needing to do a conversion no matter what since the WCAG luminance spec works off of rgb 👍

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @param contrastRatio — the contrast ratio to reach
+ * @param increment — the direction to change the lightness depending on the color scheme
+ * @returns the new hex color
+ */
+const getColorWithContrast = (
+  colorHex: string,
+  bgHex: string,
+  contrastRatio: number,
+  increment: 1 | -1,
+): {colorHex: string; lightness: number} => {
```
joshblack, line 84: Some ideas to make these types more concrete if it's helpful:

```suggestion
const getColorWithContrast = (
  colorHex: Hex,
  bgHex: Hex,
  contrastRatio: number,
  increment: 1 | -1,
): {colorHex: Hex; lightness: number} => {
```

(Uses the `Hex` type from above)

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
...
+  return {colorHex, lightness}
+}
+
+/**
+ * Takes a hex value and returns the hue, saturation and lightness for HSLuv as an object
+ * @param hex user provided hex string
+ * @returns an object with the hue, saturation and lightness values
+ */
+const hexToHsluv = (
+  hex: string,
```
joshblack, line 104:
```suggestion
  hex: Hex,
```

joshblack (approved):
Looking great! Left a couple of comments around TypeScript stuff if they're helpful 👍

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
lukasoppermann: Hey @joshblack I removed `'light' | 'dark'` because it could be any of the color schemes, also `dark_colorblind`, etc.

packages/react/src/drafts/IssueLabel/IssueLabel.tsx
```diff
@@ -29,7 +31,7 @@ export interface IssueLabelProps extends React.PropsWithChildren {
   /**
    *
    */
-  fillColor?: string
+  fillColor?: `#{string}`
```
joshblack, line 34: I think this will help with some of the errors coming into `IssueLabel` from TS since it's missing the `$` for interpolation

```suggestion
  fillColor?: `#${string}`
```

packages/react/src/drafts/IssueLabel/IssueLabel.tsx
```diff
@@ -43,8 +45,22 @@ export interface IssueLabelProps extends React.PropsWithChildren {
 }
 
 export function IssueLabel({children, fillColor, size = 'small', variant = 'gray', ...rest}: IssueLabelProps) {
+  const {resolvedColorScheme} = useTheme()
+  const mode = resolvedColorScheme?.startsWith('dark') ? 'dark' : 'light'
+  // I define those bgColors as we can't easily get them from primitives,
```
joshblack, line 50: Would it be possible to get these values from primitives in the future? Would be awesome to be able to interact with it as a package to get the values 👀

packages/react/src/drafts/IssueLabel/IssueLabel.tsx
```diff
@@ -43,8 +45,22 @@ export interface IssueLabelProps extends React.PropsWithChildren {
...
+  // they are needed to calc contrast in getColorsFromHex()
+  const bgColors = {
+    light: '#ffffff',
+    dark: '#0d1117',
+  }
```
joshblack, line 55: I think with the hex color change that this will need to be updated to be something like:

```suggestion
  const bgColors: Record<string, Hex> = {
    light: '#ffffff',
    dark: '#0d1117',
  }
```

So that when we index into it below on line 62 it will infer it as a hex color

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
+import {getContrast} from 'color2k'
+import {Hsluv} from 'hsluv'
+
+type Hex = `#${string}`
```
joshblack, line 4: One idea I had for this was to pull this out into a dedicated file, maybe a `types.ts` or `color.ts` file in the folder, and export it so that we can use it in this file, in stories, and in the component without having to re-type out `#${string}` in make TS happy

packages/react/src/drafts/IssueLabel/getColorFromHex.ts
```diff
@@ -0,0 +1,137 @@
...
+  increment: 1 | -1,
+): {colorHex: Hex; lightness: number} => {
+  // deconstruct color
+  const hsluv = hexToHsluv(colorHex)
+  let {l: lightness} = hsluv
+  const {h, s} = hsluv
+  // change lightness until contrast is reached
+  while (getContrast(colorHex, bgHex) < contrastRatio && lightness > 0 && lightness < 100) {
+    lightness += increment
+    colorHex = hsluvToHex({h, s, l: lightness})
```
joshblack, line 92: For this TS error, I think we need to either cast it as a `Hex` here or later on in the return or loosen the signature so that `colorHex` is a `string` since right now we're running into a mismatch.

joshblack: One idea for this function was:

```tsx
/**
 * Changes the lightness of a hex color until the contrast ratio is reached and returns the new hex color
 * @param fgColor — initial hex color
 * @param bgColor — color to check the contrast against
 * @param contrastRatio — the contrast ratio to reach
 * @param increment — the direction to change the lightness depending on the color scheme
 * @returns the new hex color
 */
const getColorWithContrast = (
  fgColor: Hex,
  bgColor: Hex,
  contrastRatio: number,
  increment: 1 | -1,
): {colorHex: Hex; lightness: number} => {
  const hsluv = hexToHsluv(fgColor)
  let color: string = fgColor

  // change lightness until contrast is reached
  while (getContrast(color, bgColor) < contrastRatio && hsluv.l > 0 && hsluv.l < 100) {
    hsluv.l += increment
    color = hsluvToHex(hsluv)
  }

  return {
    colorHex: color as Hex,
    lightness: hsluv.l,
  }
}
```

joshblack (commented):
Hope this helps! I assumed that you were looking for some feedback on the TS errors that were coming up but let me know if I missed anything or if I can help out with anything else 👀 

Also totally happy to meet up this week if you want to chat about this!

lukasoppermann:
Closing in favor of #algo

PR #4715, opened 2024-07-02
feat(useFocus): add support for a focus target callback

joshblack:
This is an experiment alongside https://github.com/primer/react/pull/4686 to show a potential addition to the `useFocus` hook to colocate focus management and event handlers when the focus target is dependent on state updates.

In this scenario, `useFocus()` now supports a callback that can be used to find the focus target that will run after React has renderer

```tsx
const focus = useFocus();

// ...

<button onClick={() => {
  setExpanded(true);
  focus(() => {
    return document.querySelector(`[data-show-more-group-id="${groupId}"`);
  });
}} />
```

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/internal/hooks/useFocus.ts
+++ b/packages/react/src/internal/hooks/useFocus.ts
@@ -1,21 +1,28 @@
-import {type RefObject, useEffect, useState} from 'react'
+import {type RefObject, useCallback, useEffect, useState} from 'react'
+
+type FocusTarget = HTMLElement | RefObject<HTMLElement> | (() => HTMLElement | undefined)
 
 export function useFocus() {
-  const [focusTarget, setFocusTarget] = useState<HTMLElement | RefObject<HTMLElement> | null>(null)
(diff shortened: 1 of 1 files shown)


TylerJDev:
I did not realize we already had a `useFocus` hook already 🤯, I only looked in `src/hooks` and not `internal/hooks`. 

I really like this hook! I played around with it and it works great! I love the flexibility of it between passing a DOM node, or ref directly. I guess one question is, is there a reason why this is an `internal/` hook, versus the one we provide via `src/hooks`?

joshblack:
@TylerJDev no reason, really! Just was something that I used for an example with Banner and so if we want to pull it out and provide it as a part of Primer that'd be totally fine to me 👍

PR #4716, opened 2024-07-03
feat: Adding custom property fallback plugin to postcss css module processing

jonrohan:
Part of https://github.com/github/primer/issues/3159

This adds the [postcss-custom-properties-fallback](https://github.com/stipsan/postcss-custom-properties-fallback) plugin to the postcss config plugins. 

This plugin will add fallbacks to the compiled output for all primitive values used in the component CSS files. There are two major advantages to using this plugin.

1. This will ensure that the component styles are displayed correctly when primitives might not be included properly.
2. This will allow us to convert all styled system code to CSS modules without having to account for the styled dictionary lookup that is currently used. ie `var(--borderColor-default, ${get('colors.border.default')})`.

Here's an example of the compiled changes

```diff
.prc-Blankslate-Blankslate-7GPW- {
-  --blankslate-outer-padding-block: var(--base-size-32);
-  --blankslate-outer-padding-inline: var(--base-size-32);
+  --blankslate-outer-padding-block: var(--base-size-32,2rem);
+  --blankslate-outer-padding-inline: var(--base-size-32,2rem);
  display: grid;
  justify-items: center;
  padding: var(--blankslate-outer-padding-block) var(--blankslate-outer-padding-inline);
}
```

### DIFF
--- a/packages/react/rollup.config.mjs
+++ b/packages/react/rollup.config.mjs
@@ -133,25 +133,21 @@ const baseConfig = {
         customPropertiesFallback({
           importFrom: [
             () => {
-              const primitiveFallbacks = [
-                'color-fallbacks.json',
-                'base/size/size.json',
-                'base/typography/typography.json',
-                'functional/size/border.json',
-                'functional/size/breakpoints.json',
-                'functional/size/size-coarse.json',
-                'functional/size/size-fine.json',
-                'functional/size/size.json',
-                'functional/size/viewport.json',
-                'functional/typography/typography.json',
-              ]
               let customProperties = {}
-              for (const filePath of primitiveFallbacks) {
+              const filePaths = glob.sync(['**/*.json'], {
+                cwd: path.join(__dirname, '../../node_modules/@primer/primitives/dist/fallbacks/'),
+                ignore: ['color-fallbacks.json'],
+              })
+
+              for (const filePath of filePaths) {
                 const fileData = fs.readFileSync(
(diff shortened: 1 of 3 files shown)


packages/react/rollup.config.mjs
```diff
@@ -127,7 +129,36 @@ const baseConfig = {
     }),
     importCSS({
       modulesRoot: 'src',
-      postcssPlugins: [],
+      postcssPlugins: [
+        customPropertiesFallback({
+          importFrom: [
+            () => {
+              const primitiveFallbacks = [
```
jonrohan, line 136: I want to pull this in from a glob of the `node_modules/@primer/primitives/dist/fallbacks/` directory

packages/react/rollup.config.mjs
```diff
@@ -127,7 +129,32 @@ const baseConfig = {
...
       modulesRoot: 'src',
-      postcssPlugins: [],
+      postcssPlugins: [
+        customPropertiesFallback({
+          importFrom: [
+            () => {
+              let customProperties = {}
+              const filePaths = glob.sync(['**/*.json'], {
+                cwd: path.join(__dirname, '../../node_modules/@primer/primitives/dist/fallbacks/'),
+                ignore: ['color-fallbacks.json'],
```
jonrohan, line 139: I think we can ignore color fallbacks. But want to see what @langermank says

joshblack (approved):

PR #4718, opened 2024-07-03
feat(UnderlineNavItem): add support for icons as React.ReactElement

joshblack:
Update the `icon` prop in `UnderlineNavItem` to also support `React.ReactElement` in addition to `React.FC`.

This change will help us with Figma Code Connect files and also opens up the opportunity for the `icon` prop to be static as an element which can help with performance.

#### New

- Add support for `React.ReactElement` to the `icon` prop for `UnderlineNavItem`

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

- [ ] Verify that this functionality has been covered in a test case
- [ ] Verify that the icon still renders as expected in either branch


keithamus (approved):

broccolinisoup (approved):
Looks great 🎉 Thanks!!

PR #4729, opened 2024-07-10
Add support for initial focus ref to Dialog 2

broccolinisoup:
This reported internally and lack of being able to focus on a specific element when the dialog is open is problematic for some teams. This PR adds support for an `initialFocusRef` prop to the Dialog component. (As same as Dialog 1, although parity wasn't the motivation.)

> [!NOTE]  
> We thought about using the native `autoFocus` attribute however, there is [a known bug about using it in React](https://github.com/facebook/react/issues/23301). We also not sure about our timeline of switching to native dialog, we decided to stick with the current implementation. Let me know if you have any concerns 🙌🏻 

#### New

- Add `initialFocusRef` prop. This prop allows you to specify an element that should receive focus when the dialog opens.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/__tests__/Dialog.test.tsx
+++ b/packages/react/src/__tests__/Dialog.test.tsx
@@ -4,7 +4,7 @@ import {render as HTMLRender, fireEvent} from '@testing-library/react'
 import axe from 'axe-core'
 import {behavesAsComponent, checkExports} from '../utils/testing'
 
-/* Dialog Version 2 */
+/* Dialog Version 1*/
 
 const comp = (
   <Dialog isOpen onDismiss={() => null} aria-labelledby="header">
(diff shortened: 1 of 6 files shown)


packages/react/src/__tests__/Dialog.test.tsx
```diff
@@ -4,7 +4,7 @@ import {render as HTMLRender, fireEvent} from '@testing-library/react'
 import axe from 'axe-core'
 import {behavesAsComponent, checkExports} from '../utils/testing'
 
-/* Dialog Version 2 */
+/* Dialog Version 1*/
```
siddharthkp, line 7: if you say so 😅😅😅

siddharthkp (approved):

Issue #4696, opened 2024-06-25
Roll out tooltip in our primer/react components that uses IconButton under the hood


broccolinisoup:
This is a tracking issue to safely enable tooltips in icon buttons in primer/react components.

After we make the tooltips default in icon buttons, we will need to go through each component that uses icon button under the hood. test and verify that everything works as expected. 

- [x] [Dialog v1 close button](https://github.com/primer/react/blob/main/packages/react/src/Dialog.tsx)
- [x] [ActionBar menu anchor](https://github.com/primer/react/blob/main/packages/react/src/ActionBar/ActionBar.tsx#L246)
- [x] [ActionBar.IconButton](https://github.com/primer/react/blob/main/packages/react/src/ActionBar/ActionBar.tsx#L246)
- [x] [ActionList.TrailingAction](https://github.com/primer/react/blob/main/packages/react/src/ActionList/TrailingAction.tsx#L46)
- [x] [Banner dismiss button](https://github.com/primer/react/blob/main/packages/react/src/Banner/Banner.tsx#L155)
- [x] [Select Panel v2 back button](https://github.com/primer/react/blob/main/packages/react/src/drafts/SelectPanel2/SelectPanel.tsx#L405)
- [x] [Select Panel v2 clear selection button](https://github.com/primer/react/blob/main/packages/react/src/drafts/SelectPanel2/SelectPanel.tsx#L434)
- [x] [Select Panel v2 close button](https://github.com/primer/react/blob/main/packages/react/src/drafts/SelectPanel2/SelectPanel.tsx#L438)
- [x] [TextInputInnerAction](https://github.com/primer/react/blob/main/packages/react/src/internal/components/TextInputInnerAction.tsx#L100)
- [x] [LabelGroup close overlay button](https://github.com/primer/react/blob/main/packages/react/src/LabelGroup/LabelGroup.tsx#L136)

PR #4733, opened 2024-07-11
TextInput: Refactor `TextInputInnerAction` to use the default icon button tooltip

broccolinisoup:
Closes https://github.com/primer/react/issues/4696

#### Changed

- Updated `IconButton` in the `TextInputInnerAction` component to have the right a11y props (`aria-label` and `aria-labelledby`) and add `tooltipDirection` to specify the direction of the tooltip

#### Removed

- External Tooltip that wraps the `IconButton` in the `TextInputInnerAction` 

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/internal/components/TextInputInnerAction.tsx
+++ b/packages/react/src/internal/components/TextInputInnerAction.tsx
@@ -79,7 +79,16 @@ const ConditionalTooltip: React.FC<
 
 const TextInputAction = forwardRef<HTMLButtonElement, TextInputActionProps>(
   (
-    {'aria-label': ariaLabel, tooltipDirection, children, icon, sx: sxProp, variant = 'invisible', ...rest},
+    {
+      'aria-label': ariaLabel,
+      'aria-labelledby': ariaLabelledBy,
+      tooltipDirection,
+      children,
+      icon,
+      sx: sxProp,
+      variant = 'invisible',
+      ...rest
+    },
     forwardedRef,
   ) => {
     const sx =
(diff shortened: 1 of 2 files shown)


packages/react/src/internal/components/TextInputInnerAction.tsx
```diff
@@ -92,13 +101,28 @@ const TextInputAction = forwardRef<HTMLButtonElement, TextInputActionProps>(
...
       <Box as="span" className="TextInput-action" marginLeft={1} marginRight={1} lineHeight="0">
-        {icon && !children ? (
-          <Tooltip direction={tooltipDirection ?? 's'} text={ariaLabel ?? ''} type="label">
-            {/* @ts-ignore we intentionally do add aria-label to IconButton because Tooltip v2 adds an aria-labelledby instead. */}
-            <IconButton variant={variant} type="button" icon={icon} size="small" sx={sx} {...rest} ref={forwardedRef} />
-          </Tooltip>
+        {icon && !children && ariaLabel ? (
+          <IconButton
+            {...accessibleLabel}
+            tooltipDirection={tooltipDirection ?? 's'}
```
broccolinisoup, line 117: Sparkle @joshblack for helping me fix this obscure type issue ✨

packages/react/src/internal/components/TextInputInnerAction.tsx
```diff
@@ -92,13 +101,28 @@ const TextInputAction = forwardRef<HTMLButtonElement, TextInputActionProps>(
       console.warn('Use the `aria-label` prop to provide an accessible label for assistive technology')
     }
 
+    const accessibleLabel = ariaLabel
+      ? {'aria-label': ariaLabel}
```
siddharthkp, line 105: This gives priority to aria-label over aria-labelledby which is opposite of what browsers do. Is that intentional?

> aria-labelledby takes precedence over all other methods of providing an accessible name, including aria-label
Ref: https://arc.net/l/quote/ksytdmqp

siddharthkp (approved):
1 clarifying question, LGTM other than that

Approving in advance!

packages/react/src/internal/components/TextInputInnerAction.tsx
broccolinisoup: Yes, kind of. 

Because `aria-label` is a prop on the `TextInputInnerAction` component and it is used to provide tooltip text as well. 

If there is `aria-labelledby`, the tooltip wiring needs to be manually. So to my understanding, `aria-label` is encouraged and preferred. 

Let me know if you have any concerns or anything I am missing here.

siddharthkp: I see!

PR #4736, opened 2024-07-12
Add missing displayName to PageHeader.TitleArea and PageHeader.ParentLink

lumaxis:
#### Changed

Added `displayName` to `PageHeader.TitleArea` and `PageHeader.ParentLink` sub components.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/PageHeader/PageHeader.tsx
```diff
@@ -337,6 +337,7 @@ const TitleArea = React.forwardRef<HTMLDivElement, React.PropsWithChildren<Title
     )
   },
 ) as PolymorphicForwardRefComponent<'div', TitleAreaProps>
+TitleArea.displayName = 'TitleArea'
```
broccolinisoup, line 340: This is great, thanks for adding this! We initially had all sub components FC so I think React was able to infer their names but then we refactored a couple if them. 

Do you mind adding the same thing to Parent Link? 🙏🏻

.changeset/funny-dots-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
```
broccolinisoup, line 2: I don't think we need a changeset for it. We don't include changeset for docs and tests or changes that only impacts dev env. Do you have any concern with that?

If not, we will add `skip changeset` label and checks should be all green after that yeah let me know if you have any concern 🙌🏻

broccolinisoup (approved):
Thanks so much for pushing this PR 🙏🏻 Left two questions 👀  Approving in advance!

packages/react/src/PageHeader/PageHeader.tsx
lumaxis: Not at all, updated! Do you think we need it for `Root`, too?

.changeset/funny-dots-yawn.md
lumaxis: Makes sense, wasn't familiar with the conventions y'all use – removed the changeset file 🙂

packages/react/src/PageHeader/PageHeader.tsx
broccolinisoup: I think we have a displayName for Root

```
export const PageHeader = Object.assign(Root, {
  ContextArea,
  ParentLink,
  ContextBar,
  TitleArea,
  ContextAreaActions,
  LeadingAction,
  Breadcrumbs,
  LeadingVisual,
  Title,
  TrailingVisual,
  TrailingAction,
  Actions,
  Description,
  Navigation,
})

PageHeader.displayName = 'PageHeader'
```

PR #4744, opened 2024-07-16
Accessibility fixes for ToggleSwitch

TylerJDev:
Closes https://github.com/github/primer/issues/3463

Stemming from @camertron's work in https://github.com/primer/react/pull/3902 

[Example story](https://primer-4eb6b7399b-13348165.drafts.github.io/storybook/?path=/story/components-toggleswitch-features--loading-with-delay)

#### New

* Adds new props `loadingLabelDelay` and `loadingLabel`

#### Changed

* Makes `aria-labelledby` a required prop
* Utilizes `aria-disabled` instead of `disabled`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge
- [x] (GitHub staff only) [Integration tests](https://github.com/github/github/pull/333945) pass at github/github. ⚠️ [Requires small test change to pass](https://github.com/github/github/pull/333945/commits/ea7694d020f1ea53ef4d2352207902870bf3df5e).


packages/react/src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -253,13 +310,20 @@ const ToggleSwitch = React.forwardRef<HTMLButtonElement, React.PropsWithChildren
...
         {...rest}
       >
+        {isLoadingLabelVisible ? (
+          <VisuallyHidden>
+            <AriaStatus announceOnShow id={loadingLabelId}>
+              {loadingLabel}
+            </AriaStatus>
+          </VisuallyHidden>
+        ) : null}
         {loading ? <Spinner size="small" /> : null}
```
joshblack, line 320: One quick question for this, is the label that we're seeing as "Loading" that's already in the component coming from Spinner? If so, should we try to coordinate between the two so that one component "owns" the loading label?

packages/react/src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -223,27 +243,64 @@ const ToggleSwitch = React.forwardRef<HTMLButtonElement, React.PropsWithChildren
       onClick,
       size = 'medium',
       statusLabelPosition = 'start',
+      loadingLabelDelay = 2000,
```
joshblack, line 246: Could you share more about what folks would use the delay here for? Wanted to learn more 👀 I assume a delay is nice in case it resolves quickly so we didn't need to announce anything?

packages/react/src/ToggleSwitch/ToggleSwitch.tsx
TylerJDev: Good point! We'd only want to rely on the one a consumer can provide directly to `ToggleSwitch`. I adjusted `Spinner` so that it will no longer say "loading" anymore.

packages/react/src/ToggleSwitch/ToggleSwitch.tsx
TylerJDev: Definitely! [We received feedback some time ago on loading states](https://github.com/github/primer/issues/2520#issuecomment-1681405810), and it was mentioned that there may be cases where users quickly toggle between on/off and the loading state is quick enough so that a loading message wouldn't be needed (e.g. < 2 seconds).

The time we've added (2 seconds) is a bit arbitrary, but was a suggestion on how long we _could_ wait before adding the loading message.

packages/react/src/live-region/Announce.tsx
```diff
@@ -107,6 +107,7 @@ export function Announce({
 
     observer.observe(container, {
       subtree: true,
+      childList: true,
```
TylerJDev, line 110: I was running into the same issue I had with `FormControl`, where conditionally rendered text wasn't being added to the live region. I forgot that I added `childList` in one of my PRs before but I never merged it. I'm not too sure if we discussed adding anything else here, so I'd love your opinion on this @joshblack!

More context: https://github.com/primer/react/pull/4445/commits/f1e56144fa13b68b4c4d614c89ce03f863ee0c45#r1542149212

TylerJDev:
@joshblack - if you get the chance, I'd love a review from you on this PR, mainly because [I added onto `Announce.tsx`](https://github.com/primer/react/pull/4744/files#diff-e37b42a2d39b7fe3f86f55daeaf5b2bd3619385be184c9bed5e8b019d3236e34), and I want to make sure it's the right change :grin:

lesliecdubs:
Heads up @tylerjdev, Josh is out of office for a bit (check our team calendar for dates). Feel free to leave it until then, but wanted to let you know in case it made sense to request a different reviewer.

packages/react/src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -223,27 +243,64 @@ const ToggleSwitch = React.forwardRef<HTMLButtonElement, React.PropsWithChildren
...
+
+    if (loading) {
+      if (!loadingLabelTimeoutId) {
+        setLoadingLabelTimeoutId(
+          setTimeout(() => {
+            setLoadingLabelTimeoutId(null)
+            setIsLoadingLabelVisible(true)
+          }, loadingLabelDelay) as unknown as number,
+        )
+      }
```
joshblack, line 297: I think this could be considered a side-effect during render since we're creating a side-effect (the timeout) outside of an effect. I think we'll need this code to be wrapped in an effect (or triggered by an event handler) to prevent this

packages/react/src/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -253,13 +310,19 @@ const ToggleSwitch = React.forwardRef<HTMLButtonElement, React.PropsWithChildren
         sx={sxProp}
         {...rest}
       >
-        {loading ? <Spinner size="small" /> : null}
+        <VisuallyHidden>
+          <AriaStatus announceOnShow id={loadingLabelId}>
+            {isLoadingLabelVisible && loadingLabel}
+          </AriaStatus>
+        </VisuallyHidden>
```
joshblack, line 317: Would the logic for message showing up be similar to:

```suggestion
        {isLoadingLabelVisisble ? <VisuallyHidden>
          <AriaStatus announceOnShow id={loadingLabelId}>
            {loadingLabel}
          </AriaStatus>
        </VisuallyHidden> : null}
```

Or is this not the same? Only reason this came up was with `AriaStatus` we wouldn't necessarily want the false-y case (loading label not visible) be enqueued to announce

joshblack (commented):
Sorry for the delay on this! Left a couple of comments, let me know what you think 👀 Also happy to meet up 1:1 too if that would help get this in sooner 👍

packages/react/src/ToggleSwitch/ToggleSwitch.tsx
TylerJDev: Agreed! Most of the functionality should now be in an effect!

packages/react/src/ToggleSwitch/ToggleSwitch.tsx
TylerJDev: I guess the main difference is, with this change the live region isn't added to the DOM until after a user clicks the toggle and `isLoadingLabelVisisble` becomes `true`, compared to now where it's added during the initial render alongside the component.

joshblack (approved):

lesliecdubs:
👋🏻 @TylerJDev should we revisit this one to get it merged in?

TylerJDev:
@lesliecdubs, definitely! Will put this on my to-do for next week to ship! 🚢

TylerJDev:
To be unblocked by https://github.com/github/github/pull/371043. Once merged, this PR should be good to ship!

TylerJDev:
@joshblack, mind giving this one another quick review? Shouldn't include any additional changes since the last approval, but it's been a bit since then 😅

joshblack (approved):

PR #4747, opened 2024-07-18
Move `Stack` out of drafts

langermank:
Moving the stories out of drafts into the main nav


joshblack (approved):

broccolinisoup (commented):
Somewhat related to these changes;

Looking at the docs.json

```
  "id": "stack",
  "name": "Stack",
  "status": "alpha",
  "a11yReviewed": false,
  "stories": [],
  "importPath": "@primer/react/experimental",
```

I think importPath should be `@primer/react` considering the status is alpha, it is exported in the root bundle and now we are removing the drafts from stories. Let me know if I am mistaken!

PR #4748, opened 2024-07-18
feat(components-json): update parser to support StoryObj from storybook

joshblack:
Update our parser to support the `StoryObj` type from `storybook` so that stories are generated in docs that use this feature.

#### Changed

- Update built script to generate stories for StoryObj exports

### Rollout strategy

This is a change to our internal build script for generating story metadata


langermank (approved):
Yay! I'm curious now if this will fix other components in addition to Stack. I just only knew about Stack being broken.

PR #4750, opened 2024-07-18
Add `KeybindingHint` component

iansan5653:
- Closes https://github.com/github/primer/issues/3169

Adds a new component called `KeybindingHint`. This is upstreamed from our internal codebase, where it is used in production in several locations. As part of the upstreaming, I've taken the liberty of renaming the component from `KeyboardKey` to `KeybindingHint` to better indicate its purpose and flexibility. I've also refactored the code, splitting it into smaller files.

#### New

- New `KeybindingHint` component for indicating an available keybinding

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


packages/react/src/hooks/useIsMacOS.ts
```diff
@@ -0,0 +1,16 @@
...
+ * client turns out to be a MacOS device.
+ */
+export function useIsMacOS() {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  const [isMacOS, setIsMacOS] = useState(() => (window !== undefined ? ssrUnsafeIsMacOS() : false))
+
+  useEffect(() => setIsMacOS(ssrUnsafeIsMacOS()), [])
+
+  return isMacOS
+}
```
iansan5653, line 16: In the current usage of this component internally, we call `isMacOS` directly at the top level of the file. This is obviously not good for SSR-compatibility, so I've introduced this new hook:

- In non-SSR, it causes no extra renders and returns `isMacOS` directly
- In SSR, it initially returns `false`, then causes a re-render after hydration if the value would change to `true` (if the client is using a Mac)

I exposed this hook publicly so that we can also use it elsewhere when actually binding keyboard shortcuts.

packages/react/src/KeybindingHint/components/Chord.tsx
```diff
@@ -0,0 +1,70 @@
...
+/** Split and sort the chord keys in standard order. */
+const splitChord = (chord: string) =>
+  chord
+    .split('+')
+    .map(k => k.toLowerCase())
+    .sort(compareLowercaseKeys)
+
+export const Chord = ({keys, format = 'condensed', variant = 'normal'}: KeybindingHintProps) => (
+  <Text
+    sx={{
```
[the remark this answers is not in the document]
iansan5653: I'm definitely open to this. I didn't see any other CSS modules in the package so I figured I'd keep it consistent for now, but happy to make the change.

packages/react/src/KeybindingHint/KeybindingHint.tsx
```diff
@@ -0,0 +1,49 @@
+import React, {type ReactNode} from 'react'
```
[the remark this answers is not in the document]
iansan5653: I was considering this but I didn't think it made sense to go through the draft phase for a component that's already stable and live in production. The API has been stable for about a year now without any breaking changes.

That said I have no strong feelings either way, but I would like to know what promotion criteria would need to be resolved before moving it to alpha so that I can follow up. Is it just a standard time period, or is there more to it?

iansan5653:
Hmm, CI is [failing](https://github.com/primer/react/actions/runs/9996960112/job/27632504699?pr=4750#step:6:227) with:

```
error "window" is not available during server side rendering.

   9 | export function useIsMacOS() {
  10 |   // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
> 11 |   const [isMacOS, setIsMacOS] = useState(() => (window !== undefined ? ssrUnsafeIsMacOS() : false))
     |                                                 ^
  12 |
  13 |   useEffect(() => setIsMacOS(ssrUnsafeIsMacOS()), [])
  14 |
```

Which I guess is it trying to be helpful, but in this case I am just trying to detect if `window` is `undefined` _because_ of SSR. Is there a better way to do that?

packages/react/src/KeybindingHint/KeybindingHint.tsx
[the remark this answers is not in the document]
iansan5653: Alrighty, I'll move it

.vscode/settings.json
```diff
@@ -13,11 +13,11 @@
   "json.schemas": [
     {
       "fileMatch": ["*.docs.json"],
-      "url": "./script/components-json/component.schema.json"
+      "url": "./packages/react/script/components-json/component.schema.json"
     },
     {
       "fileMatch": ["generated/components.json"],
-      "url": "./script/components-json/output.schema.json"
+      "url": "./packages/react/script/components-json/output.schema.json"
```
iansan5653, line 20: Tiny unrelated change to fix these schema paths so we can get typechecking/documentation for component metadata JSON

lesliecdubs:
@broccolinisoup offered to give this a review from Primer this week, so tagging her in!

packages/react/src/KeybindingHint/components/Key.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/** Renders a single key with accessible alternative text. */
+export const Key = ({name, format}: KeyProps) => {
+  const isMacOS = useIsMacOS()
+
+  return (
+    <>
+      <VisuallyHidden>{accessibleKeyName(name, isMacOS)}</VisuallyHidden>
+      <span aria-hidden>{format === 'condensed' ? condensedKeyName(name, isMacOS) : fullKeyName(name, isMacOS)}</span>
```
broccolinisoup, line 19: I am curious if this needs to be customisable as well depending on where it is used? 🤔 

For example on IconButton and Tooltip, we would need to append the accessile name to the nae or the description. I wonder how this visually hidden element will play out here? 👀

packages/react/src/KeybindingHint/key-names.ts
```diff
@@ -0,0 +1,113 @@
...
+    capslock: 'Caps Lock',
+    printscreen: 'Print Screen',
+  })[key] ?? capitalize(key)
+
+/**
+ * Accessible key names intended to be read by a screen reader. This prevents screen
+ * readers from expressing punctuation in speech, ie, reading a long pause instead of the
+ * word "period".
+ */
+export const accessibleKeyName = (key: string, isMacOS: boolean) =>
```
broccolinisoup, line 62: How confident are we with the accessible names? Have we confirmed with the accessibility team? Just wanted to make sure there is some ground work for this.

packages/react/src/KeybindingHint/components/Sequence.tsx
```diff
@@ -0,0 +1,27 @@
...
+const splitSequence = (sequence: string) => sequence.split(' ')
+
+export const Sequence = ({keys, format = 'condensed', variant = 'normal'}: KeybindingHintProps) =>
+  splitSequence(keys).map((c, i) => (
+    <Fragment key={i}>
+      {
+        //  Since we audibly separate individual keys in chord with space, we need some other separator for chords in a sequence
+        i > 0 && (
+          <>
+            <VisuallyHidden>, then</VisuallyHidden>{' '}
```
broccolinisoup, line 15: As far as I know, the normal spacing gets stripped by screen readers. I am not sure how that might look like though. 

On UnderlineNav, we have a similar practise and we use non-breaking space to make sure screen reasers announce the spacing properly.

https://github.com/primer/react/blob/main/packages/react/src/UnderlineNav/UnderlineNav.tsx#L331

I might be missing some context though - let me know 🙌🏻

broccolinisoup (commented):
Hey @iansan5653 👋🏻 

This is a very thorough work!! Thanks so much for bringing this contribution to Primer, really appreciate it! 

I have done an initial review and left some comments and questions. Looking forward to hearing what you think.

packages/react/src/KeybindingHint/components/Key.tsx
iansan5653: I have verified that if we are generating the accessible name/description from DOM, this results in the correct output. For example, if you render this inside of a `Button` element, the `VisuallyHidden` part is what becomes part of the accessible name while the `aria-hidden` part is ignored. This is what we want since the `VisuallyHidden` part contains the readable string version of the key icons. 

For when we are appending the shortcut to a plain string used as `aria-label` or `aria-description`, we expose a `getAccessibleKeybindingHintString` that allows us to skip the whole DOM part and just get a usable string, so that case is covered as well.

packages/react/src/KeybindingHint/components/Sequence.tsx
iansan5653: In the `UnderlineNav` case, the space falls at the end of an element before the closing tag. In this case, there's a space inside the element between text, and there's a space between two elements. I am pretty sure (like 85% sure 😅) sure these will work as expected given their placement, but I'll add some stories using `KeyboardHint` as an `aria-labelledby` target to see what actually happens.

packages/react/src/KeybindingHint/key-names.ts
iansan5653: I'm pretty sure we discussed this with the accessibility team in Slack way back when we first wrote this component, but that was almost two years ago so unfortunately I don't have any record of it. I do feel pretty confident in this based on my own screen reader testing, but I am not the target audience so I can't say that with absolute certainty.

packages/react/src/index.ts
```diff
@@ -196,6 +196,8 @@ export type {UnderlineNavProps, UnderlineNavItemProps} from './UnderlineNav'
 export {ActionBar} from './ActionBar'
 export type {ActionBarProps} from './ActionBar'
 
+export * from './KeybindingHint'
```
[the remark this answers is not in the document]
iansan5653: Sure, I'll move it there.

packages/react/src/KeybindingHint/components/Sequence.tsx
iansan5653: I actually noticed an _extra_ space in the accessible label, rather than a missing one 😄 - the AT is inserting a space between elements. So we'd get the label "x , then y". I solved this by removing the comma, which in retrospect feels like an unnecessary pause anyway - it reads pretty naturally as "x then y".

packages/react/src/KeybindingHint/KeybindingHint.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+import type {Meta} from '@storybook/react'
+import {KeybindingHint} from './KeybindingHint'
+
+type KeybindingHintProps = React.ComponentProps<typeof KeybindingHint>
+
+const args = {
+  keys: 'Mod+Shift+K',
+} satisfies KeybindingHintProps
+
+const meta = {
```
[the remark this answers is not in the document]
iansan5653: Done! Pulled these out to `features` and also added several `examples` showing interplay with other components.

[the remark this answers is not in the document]
iansan5653: I don't think we should resolve this in this PR, but I did notice some accessibility issues with the example stories:

- In `ActionList.Item`, the `TrailingVisual` does not become part of the accessible label. This means that shortcut hints rendered in this manner are invisible to screen readers 😞 
- In `Button`, the `trailingVisual` is part of the label but with only a space as separation. This causes the button label to be read as, for example, "Submit command enter".
- In `TextInput`, the `trailingVisual` is read to screen readers as text after the input, not associated semantically with the input at all. 

I think in all these cases, the ideal solution would be to append the trailing visual to the semantic label with a comma as a separator. For example, "Submit, command enter" or "Search, forward slash".

packages/react/src/KeybindingHint/components/Chord.tsx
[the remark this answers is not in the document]
iansan5653: Do you have any prior art for using data attributes for this? I'm not sure I've seen that approach before. 

FWIW this should be pretty easy to migrate - we will just introduce an additional class like `onEmphasis` for the variant.

Maybe since this is a draft PR it would be a good proof of concept to just migrate it now? I'm happy to do so; it's pretty straightforward with [the migration tool](https://github.com/iansan5653/primer-sx-to-css-converter).

packages/react/src/KeybindingHint/KeybindingHint.tsx
```diff
@@ -0,0 +1,48 @@
...
+)
+
+/** Indicates the presence of an available keybinding. */
+// KeybindingHint is a good candidate for memoizing since props will rarely change
+export const KeybindingHint = memo((props: KeybindingHintProps) => (
+  <Kbd>
+    <Sequence {...props} />
+  </Kbd>
+))
+KeybindingHint.displayName = 'KeybindingHint'
```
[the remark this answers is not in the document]
iansan5653: The engine can infer a function name for simple  named declarations like `function Component() {}` and `const Component = ()=> {}`.

But in this case we are declaring an anonymous function without a name, then sending it through `memo` and assigning it to a named constant. There's no way for the engine to infer a name for the function so the component would be anonymous without a `displayName`. This is actually required by the linter.

packages/react/src/KeybindingHint/components/Key.tsx
broccolinisoup: Wonderful! I have played with it as well and it works great ✨ 

I have one comment though regarding using this in Tooltip:

I quickly implemented it to see how that would work and looks in Tooltip and I think we need to work through the colours if this is how we are going to use this component in Tooltip

Tooltip is open on a button and it has keyboard shortcuts on the tooltip. The text colour of the keybinding component doesn't create enough contrast on the tooltip background

tagging @mperrotti here for design 👀

broccolinisoup: Regarding using `getAccessibleKeybindingHintString` ,  is there a way to handle determining `isMacOS` under the hood so that we don't worry about it every time when we use this function? 

Let me know if there is anything I am missing here

packages/react/src/KeybindingHint/key-names.ts
broccolinisoup: Thank you! I'll share this with the team to see if we need to cross check this again with the Accessibility team. Thanks for your patience 🙏🏻

broccolinisoup (commented):
Thanks for addressing all these comments quickly 🙏🏻 Left a few more comments - looking forward to hearing your thoughts!!

packages/react/src/KeybindingHint/components/Key.tsx
iansan5653: This is actually one of the reasons the `onEmphasis` variant exists - we'd definitely want to use that here

iansan5653: >I quickly implemented it to see how that would work and looks in Tooltip and I think we need to work through the colours if this is how we are going to use this component in Tooltip

I think this should be resolvable with just `variant="onEmphasis"`, which should flip the text color to work here. 



> Regarding using `getAccessibleKeybindingHintString` , is there a way to handle determining `isMacOS` under the hood so that we don't worry about it every time when we use this function?

Not really, since the macOS logic is now inside a hook - we'd need to make this a hook as well. There are rare cases where we call this outside the react app, so we'd rather not have this as a hook. 

What we could do is introduce a new hook wrapper around this like `useAccessibleKeybindingHintString`

packages/react/src/__tests__/KeybindingHint.test.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react'
```
siddharthkp, line 1: (not blocking) You can move these inside `packages/react/src/KeyboardHint/__tests__`, that's the preferred location

docs/src/@primer/gatsby-theme-doctocat/nav.yml
```diff
@@ -82,6 +82,8 @@
       url: /Heading
     - title: IconButton
       url: /IconButton
+    - title: KeybindingHint
+      url: /KeybindingHint
```
siddharthkp, line 86: Should this be under drafts as well in the nav

(non blocking because these docs are going away soon anyways in favor of https://primer.style/components)

packages/react/src/KeybindingHint/components/Key.tsx
broccolinisoup: No worries, let's keep it as is 👍🏻

broccolinisoup (approved):
I assume we don't have any usages at dotcom since it is a new component and the overall PR and the API looks good to me.! 🚀 

We can always address later if any issues come up or if there is anything we want to improve.

iansan5653:
> I assume we don't have any usages at dotcom

To be clear, this is upstreaming a component that's in dotcom, so we do have a few existing usages (under the name `KeyboardKey`). I'll handle migrating them after this is shipped.

PR #4751, opened 2024-07-18
feat: add postcss-preset-primer

joshblack:
As more and more places are using/compiling CSS with Primer, it would be helpful to have a package for this that acts as a preset for the various plugins/strategies that we recommend.

This PR adds in a proof-of-concept package for this, `postcss-preset-primer`, that can be shared and reused for compiling CSS for Primer.

#### New

- Add `postcss-preset-primer` package

### Rollout strategy

This package is currently private


packages/postcss-preset-primer/src/index.js
```diff
@@ -0,0 +1,87 @@
...
+      ],
+      [
+        '@csstools/postcss-global-data',
+        {
+          files: globSync(`dist/css/**/*.css`, {cwd: primitivesPath}),
+        },
+      ],
+      ['postcss-nesting', {edition: '2024-02'}],
+      ['postcss-custom-media', {}],
+      ['postcss-mixins', {}],
```
jonrohan, line 58: This one is controversial we could leave off for now

langermank (approved):

adam-rocska:
Any hint maybe on when could this be available for install?

Just wanted to give it a spin according to the [getting started](https://github.com/primer/react/tree/main/packages/postcss-preset-primer#getting-started) ¯\\_(ツ)_/¯ but got this:
```zsh
rocskaadam@Rocska-MBP src.adam-rocska.github.io % pnpm install postcss-preset-primer --save-dev
 ERR_PNPM_FETCH_404  GET https://registry.npmjs.org/postcss-preset-primer: Not Found - 404

This error happened while installing a direct dependency of /Users/rocskaadam/src/adam-rocska/src/src.adam-rocska.github.io

postcss-preset-primer is not in the npm registry, or you have no permission to fetch it.

No authorization header was set for the request.
Progress: resolved 16, reused 16, downloaded 0, added 0
```

PR #4760, opened 2024-07-19
[Storybook] Start deprecation of tooltip v1

tbenning:
To keep things less confusing for consumers / users of Storybook, starting the deprecation cycle for tooltipV1 by removing it from the main components area of Storybook.

#### Changed

`Moved Components/Tooltip` -> `Deprecated/Components/TooltipV1`

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [X] None; if selected, include a brief description as to why

### Testing & Reviewing
Tests are still operating. E2E is updated.

### Merge checklist

- [X] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Tooltip/Tooltip.features.stories.tsx
+++ b/packages/react/src/Tooltip/Tooltip.features.stories.tsx
@@ -8,7 +8,7 @@ import {SearchIcon} from '@primer/octicons-react'
 /* Tooltip v1 */
 
 export default {
-  title: 'Deprecated/Components/TooltipV1/Features',
+  title: 'Deprecated/Components/Tooltip/Features',
   component: Tooltip,
 
   decorators: [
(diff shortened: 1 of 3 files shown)


packages/react/src/Tooltip/Tooltip.features.stories.tsx
```diff
@@ -8,7 +8,7 @@ import {SearchIcon} from '@primer/octicons-react'
 /* Tooltip v1 */
 
 export default {
-  title: 'Components/Tooltip/Features',
+  title: 'Deprecated/Components/TooltipV1/Features',
```
broccolinisoup, line 11: Sorry for making a comment on the draft PR, let me know if this is just for testing etc. - Just wanted to mentioned that  we probably don't want to update the name from `Tooltip` to `TooltipV1` because that might break the [story name generation](https://github.com/primer/react/blob/main/packages/react/script/components-json/build.ts#L62) that we end up displaying on the docs site.

tbenning: Ah very interesting, good point. I'll swap it back. We can revisit naming and organization when we properly deprecate it.

joshblack (approved):

PR #4764, opened 2024-07-22
StateLabel: Add labels to icon

siddharthkp:
- Fixes https://github.com/github/primer/issues/3496

> The `issueClosedNotPlanned` variant has an icon that differentiates it from `issueClosed ` and `pullClosed `. The icon has no text alternative and includes aria-hidden="true", so is ignored by screen readers. Screen reader users do not know what the labels 'Draft', 'Open' and 'Closed' refers to, unless context is being provided elsewhere on the page, for example, the component is on an Issue page.
> Screen shot of State Label variants for Issues and Pull requests, each consists of a single word, with a coloured background shaped like a pill, with an icon

Before: 

```html
<span class="StateLabel__StateLabelBase-sc-qthdln-0 eDFXzm">
  <svg aria-hidden="true" focusable="false">...</svg>
  Closed
</span>
```

After:

```html
<span class="StateLabel__StateLabelBase-sc-qthdln-0 eDFXzm">
  <svg role="img" aria-label="Issue" focusable="false">...</svg>
  Closed
</span>
```

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/StateLabel/StateLabel.tsx
+++ b/packages/react/src/StateLabel/StateLabel.tsx
@@ -4,6 +4,7 @@ import {
   GitPullRequestClosedIcon,
   GitPullRequestDraftIcon,
   IssueClosedIcon,
+  QuestionIcon,
   SkipIcon,
   IssueDraftIcon,
   IssueOpenedIcon,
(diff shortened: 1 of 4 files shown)


packages/react/src/StateLabel/StateLabel.tsx
```diff
@@ -33,6 +33,19 @@ const octiconMap = {
   unavailable: AlertIcon,
 }
 
+const labelMap: Record<keyof typeof octiconMap, 'Issue' | 'Pull request' | ''> = {
+  issueOpened: 'Issue',
+  pullOpened: 'Pull request',
+  issueClosed: 'Issue',
+  issueClosedNotPlanned: 'Issue',
```
TylerJDev, line 40: non-blocking, but would we want to add "not planned" here? (i.e. "Issue, not planned")

TylerJDev (approved):
Looks great! ✨

packages/react/src/StateLabel/StateLabel.tsx
siddharthkp: Great catch!

PR #4771, opened 2024-07-24
Autocomplete: Only open menu on click

TylerJDev:
Closes https://github.com/github/primer/issues/3387

Sets `openOnFocus` to `false` by default.

On focus or click, we no longer show the menu by default. You may still activate the menu by entering `Autocomplete.Input` and pressing the up/down arrow keys. Showing the menu instantly upon focus means that a screen reader user is notified of the menu instantly with no indication given that the user is on the input itself.

#### Changed

* Sets `openOnFocus` to `false`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [x] (GitHub staff only) [Integration tests pass at github/github](https://github.com/github/github/pull/334338) ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))


packages/react/src/Autocomplete/AutocompleteInput.tsx
```diff
@@ -52,14 +52,14 @@ const AutocompleteInput = React.forwardRef(
     const [highlightRemainingText, setHighlightRemainingText] = useState<boolean>(true)
     const {safeSetTimeout} = useSafeTimeout()
 
-    const handleInputFocus: FocusEventHandler<HTMLInputElement> = useCallback(
+    const handleInputClick: MouseEventHandler<HTMLInputElement> = useCallback(
```
joshblack, line 55: Feel free to ignore this since it's not really related to the changes proposed 👀 

I wanted to ask if this `useCallback` is doing much for us or if it would be something we could refactor to a default event handler / function.

joshblack (commented):
Would it be possible to use `openOnFocus` for this migration? This might be able to smooth things out and eventually remove this in v38 (let me know if you want me to add it to that list btw!) One approach could be to change the default to `false` and see how things go and eventually remove the prop altogether.

Curious what you think 👀

TylerJDev:
> Would it be possible to use openOnFocus for this migration?

So turning the prop to `false` by default so that both click and focus don't show the menu? We definitely could! I wasn't sure if there were instances where it was ideal for `AutocompleteMenu` to be visible upon click, but honestly I don't think it's the biggest risk.

packages/react/src/Autocomplete/AutocompleteInput.tsx
TylerJDev: Yeah we definitely can adjust this!

TylerJDev:
@joshblack, I adjusted it so that we instead change `openOnFocus` to `false` by default! This means neither click/focus will trigger the menu. The user will have to manually go into the input and trigger it themselves. Let me know if any adjustments are needed!

joshblack (approved):
LGTM! 🤞 it passed downstream 👀 

Would it be helpful to add a `@deprecated` tag with this to the props and `docs.json` file, too?

TylerJDev:
> Would it be helpful to add a @deprecated tag with this to the props and docs.json file, too?

Definitely! I added it to the prop & docs. I'll add it to the [v38 release issue](https://github.com/github/primer/issues/3541) too!

PR #4779, opened 2024-07-26
DialogV2: re-set focusTrap when footer content changes

TylerJDev:
Closes https://github.com/github/primer/issues/3358 https://github.com/github/primer/issues/3356 https://github.com/github/primer/issues/3357 https://github.com/github/primer/issues/2480

Adds `footer` as a dependency of `useFocusTrap`, ensures that if content is changed within either, focus trap will properly focus newly added focusable elements. Example story: [Repro Multistep Dialog With Conditional Footer](https://primer-a40bbbf7cd-13348165.drafts.github.io/storybook/?path=/story/components-dialog-features--repro-multistep-dialog-with-conditional-footer).

#### Changed

* Adds `footer` to `useFocusTrap` dependencies 
* Removes `autofocus` prop from stories

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
I had to utilize `querySelector` to search for the close button. I'm wondering if there'd be any benefit to letting consumers add a ref directly to it via a prop (e.g. `closeButtonRef`)? This fix is only required for a story, so it's low priority.

siddharthkp (commented):
Logic is sound. Just need to test for any performance implications

TylerJDev:
> Logic is sound. Just need to test for any performance implications

That is fair! Is there a good way to go about this outside of just testing existing dialogs? Another option is making this opt-in, but not too sure if that's the best route to take if we want this functionality by default 🤔

siddharthkp:
> That is fair! Is there a good way to go about this outside of just testing existing dialogs?

Can see how if the hook is called only when it should or also on every render or on unrelated changes inside the Dialog content.

> Another option is making this opt-in, but not too sure if that's the best route to take if we want this functionality by default 🤔

I think this should be default as well

TylerJDev:
> > That is fair! Is there a good way to go about this outside of just testing existing dialogs?
> 
> Can see how if the hook is called only when it should or also on every render or on unrelated changes inside the Dialog content. Dependency on a `const body` is looks doubtful to me
> 
> > Another option is making this opt-in, but not too sure if that's the best route to take if we want this functionality by default 🤔
> 
> I think this should be default as well

I removed `body` as a dependency as I don't think we actually need it. I believe that `focus-trap` automatically adds elements that are appended to the `container`, so we don't have to add it as a dependency, as it would only be redundant. If I had to guess, content inside of the `footer` is added after the last focused element when the focus trap is initialized, meaning it wont be included in the focus trap, only elements that are within the first focusable element and the last focusable elements are included. This is only an assumption looking at the [`focus-trap` behavior](https://github.com/primer/behaviors/blob/main/src/focus-trap.ts#L75) 😅 . 

I left `footer` as a dependency as we'd want to include content here if it's added after the fact, like in [the repro multistep story](https://primer.style/react/storybook/?path=/story/components-dialog-features--repro-multistep-dialog-with-conditional-footer). This is also less risky when it comes to performance, as the footer is less likely to have re-renders like `body` would (if any) after the initial render. We could also include `header` here, as it will most likely fall into the same group as the footer.

siddharthkp:
I added some state in the story to show the effect of dependency (committed the change to story), now if you add a `console.log` [to the line where focusTrap is called](https://github.com/primer/react/blob/main/packages/react/src/hooks/useFocusTrap.ts#L76), you see how many times it is called.

Video description:

Part 1, with dependencies:
1. When the Dialog opens, you see "setting focus trap" printed twice (let's ignore the second one for now)
2. With dependencies like footer or body added without "memo", you see "setting focus trap" printed once more. This is good! We want the focus trap to be set again for changed footer content.
3. Entering text into the input causes the component to re-render and all the variables inside it too. You see "setting focus trap" printed on every keystroke! This removing and attaching event listeners can cause performance regressions.

Part 2, after commenting out dependencies
1. When the Dialog opens, you see "setting focus trap" printed twice (same as before)
2. With no dependencies array, you don't see any more "setting focus trap" printed. Focus trap is not resetting when the footer content changed
3. Entering text into the input causes the component to re-render and all the variables inside it too. But, you don't see "setting focus trap" printed because we are not passing a dependencies to useFocusTrap


https://github.com/user-attachments/assets/d4faf9d1-9307-43f5-9320-1ba973a0a4e1

TylerJDev:
Thanks a ton for the notes and video @siddharthkp, it helps a lot to see where the problem is ✨  The current route is definitely not the best for performance as we're reinitializing the focus trap each render 😬 

Digging into this a bit more, I have a few alternatives without us utilizing dependencies, and I would love your opinion!

* Utilize a mutation observer in [`focus-trap` behavior](https://github.com/primer/behaviors/blob/main/src/focus-trap.ts)
  - We could use a mutation observer in our focus trap behavior to account for new focusable elements added. We do this [in `focus-zone`](https://github.com/primer/behaviors/blob/main/src/focus-zone.ts#L522) already, so we'd be able to take from an existing implementation. There is still some performance concern, but I don't think it'd be as bad if we're handling it correctly. This also means that on re-renders we wouldn't be initializing the focus trap once set.
* Adjust "sentinel" to always be at the start/end of the container
  - When the dialog footer is conditionally rendered, the ending sentinel is before the footer, meaning that it's not encompassing the entirety of the dialog's content. Technically, it _should_ encompass all elements within the container passed to `focusTrap`. We could adjust the behavior to always be at the start and end of the container so that any elements within are always accounted for. 
  
These are just two options, both requiring changes to the behavior, rather than the `useFocusTrap` hook. I'm still digging to see if there are any other routes we could take, or if it's possible to keep the dependency option, but only run the focus trap behavior when needed (i.e. footer/header has changed significantly from the previous render).

siddharthkp:
> We do this [in focus-zone](https://github.com/primer/behaviors/blob/main/src/focus-zone.ts#L522) already, so we'd be able to take from an existing implementation. There is still some performance concern, but I don't think it'd be as bad if we're handling it correctly. This also means that on re-renders we wouldn't be initializing the focus trap once set

Ooh i really like that idea! 

I wonder how _eager_ should the observer be. We haven't received any performance concerns about focus zone _yet_, but it's also really difficult to attribute it to that 😅

TylerJDev:
Quick update, working on a fix in primer/behaviors: https://github.com/primer/behaviors/pull/429. Basically adding a mutation observer that checks if the start or end sentinels are no longer at the start/end.

TylerJDev:
@siddharthkp, tested this with the updated behaviors version & very lightly in dotcom (though I couldn't find too many dialogs). Doesn't look like there's much performance hit, as the likelihood of content being added before/after sentinels is low. In cases where elements are added after the focus trap, they will only be moved once once it's asserted that either the first child or last child in the focus trap is a `.sentinel`.  

I'm curious if there's any additional testing we should do, or if we should ship this change (with the behaviors bump).

siddharthkp:
Great, thanks for making sure! I think it sounds good, let's ship it

siddharthkp (approved):

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -476,6 +469,13 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
           }),
         )
 
+  useFocusTrap({
```
siddharthkp, line 472: (non blocking) Do we still need to move this below?

TylerJDev: Nope! We can move this back.

PR #4784, opened 2024-07-26
refactor: remove stub folders for TypeScript resolution

joshblack:
Closes https://github.com/github/primer/issues/3665

#### Removed

- The `codemods` folder has been removed as it is no longer needed. If we would like to use this in the future, we can revisit where we would like this to live 👀 
- The stub folders for TypeScript resolution have been removed

### Rollout strategy

I believe the removal of the stub folders will trigger any consumer using Primer React and the `node` moduleResolution option may end up seeing an issue when importing from sub-paths. As a result, they should update their moduleResolution option to node16 or higher


siddharthkp (approved):
Very nice, very clean!

PR #4789, opened 2024-07-29
chore(changeset): enter prerelease mode for v37

joshblack:
Enters the project into "prerelease" mode for v37 based on: https://github.com/github/primer/discussions/3552. For more information about prereleases in changesets, checkout [this explainer](https://github.com/changesets/changesets/blob/main/docs/prereleases.md).

#### Changed

- Add `pre.json` config for changesets to enter project into prerelease mode with an `rc` tag
- Update canary release workflow to no longer use `--snapshots` as this is unnecessary in prerelease mode
- Update ignore config for changesets to exclude private packages

### Rollout strategy

This is a change to the repo and will not impact downstream consumers

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -6,5 +6,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["docs", "example-*", "codesandbox"]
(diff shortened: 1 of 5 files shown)


joshblack:
For the publish error, we might be running into: https://github.com/changesets/changesets/issues/835

jonrohan (approved):

PR #4795, opened 2024-07-30
ActionList: Enable focusZone for roles listbox and menu

siddharthkp:
- Fixes https://github.com/github/primer/issues/3475

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/ActionList/List.tsx
+++ b/packages/react/src/ActionList/List.tsx
@@ -51,6 +51,7 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
       disabled: !enableFocusZone,
       containerRef: listRef,
       bindKeys: FocusKeys.ArrowVertical | FocusKeys.HomeAndEnd | FocusKeys.PageUpDown,
+      focusOutBehavior: listRole === 'menu' ? 'wrap' : undefined,
     })
 
     return (
--- a/packages/react/src/ActionList/List.tsx
+++ b/packages/react/src/ActionList/List.tsx
@@ -33,15 +33,20 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
 
     /** if list is inside a Menu, it will get a role from the Menu */
     const {
-      listRole,
+      listRole: listRoleFromContainer,
       listLabelledBy,
       selectionVariant: containerSelectionVariant, // TODO: Remove after DropdownMenu2 deprecation
-      enableFocusZone,
+      enableFocusZone: enableFocusZoneFromContainer,
     } = React.useContext(ActionListContainerContext)
 
     const ariaLabelledBy = slots.heading ? slots.heading.props.id ?? headingId : listLabelledBy
-
+    const listRole = role || listRoleFromContainer
     const listRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLUListElement>)
+
+    let enableFocusZone = false
+    if (enableFocusZoneFromContainer !== undefined) enableFocusZone = enableFocusZoneFromContainer
+    else if (listRole) enableFocusZone = ['menu', 'menubar', 'listbox'].includes(listRole)
+
     useFocusZone({
       disabled: !enableFocusZone,
       containerRef: listRef,
@@ -54,14 +59,14 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
           variant,
           selectionVariant: selectionVariant || containerSelectionVariant,
           showDividers,
-          role: role || listRole,
+          role: listRole,
           headingId,
         }}
       >
         {slots.heading}
         <ListBox
           sx={merge(styles, sxProp as SxProp)}
-          role={role || listRole}
+          role={listRole}
           aria-labelledby={ariaLabelledBy}
           {...props}
           ref={listRef}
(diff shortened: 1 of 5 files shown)


siddharthkp:
Identified some of the failing tests in dotcom as coming from the release candidate. Will wait for next dotcom upgrade before going too deep in debugging

packages/react/src/ActionList/List.tsx
```diff
@@ -33,15 +33,20 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
...
     const ariaLabelledBy = slots.heading ? slots.heading.props.id ?? headingId : listLabelledBy
-
+    const listRole = role || listRoleFromContainer
     const listRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLUListElement>)
+
+    let enableFocusZone = false
+    if (enableFocusZoneFromContainer !== undefined) enableFocusZone = enableFocusZoneFromContainer
+    else if (listRole) enableFocusZone = ['menu', 'menubar', 'listbox'].includes(listRole)
+
     useFocusZone({
```
TylerJDev, line 50: I'm wondering if we'd want to add `focusOutBehavior: 'wrap'` by default, or allow consumers to add that themselves?

packages/react/src/ActionList/List.tsx
```diff
@@ -33,15 +33,20 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
...
     } = React.useContext(ActionListContainerContext)
 
     const ariaLabelledBy = slots.heading ? slots.heading.props.id ?? headingId : listLabelledBy
-
+    const listRole = role || listRoleFromContainer
     const listRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLUListElement>)
+
+    let enableFocusZone = false
+    if (enableFocusZoneFromContainer !== undefined) enableFocusZone = enableFocusZoneFromContainer
+    else if (listRole) enableFocusZone = ['menu', 'menubar', 'listbox'].includes(listRole)
```
TylerJDev, line 48: Non-blocking: I see we have have `menubar`, I am wondering if there'd be a case where `role=menubar` is used with ActionList 🤔

packages/react/src/ActionList/List.tsx
```diff
@@ -33,15 +33,20 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
...
-      enableFocusZone,
+      enableFocusZone: enableFocusZoneFromContainer,
     } = React.useContext(ActionListContainerContext)
 
     const ariaLabelledBy = slots.heading ? slots.heading.props.id ?? headingId : listLabelledBy
-
+    const listRole = role || listRoleFromContainer
     const listRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLUListElement>)
+
+    let enableFocusZone = false
```
TylerJDev, line 46: If there's an existing focus zone being used for an `ActionList`, would the one being used here replace it? I saw an instance where there was an `ActionList` inside of an `AnchoredOverlay` that had an implicit focus zone attached: 

```tsx
<AnchoredOverlay
  renderAnchor={props => <Button {...props}>Button</Button>}
  focusZoneSettings={{
    focusOutBehavior: 'wrap',
  }}
>
  <ActionList role="listbox">
    <ActionList.Item>Item 1</ActionList.Item>
    <ActionList.Item>Item 2</ActionList.Item>
  </ActionList>
</AnchoredOverlay>
```
I'm wondering if this would be an issue or not 🤔. I think the existing cases are rare, so doesn't seem like it would be.

TylerJDev (approved):
LGTM! Few non-blocking comments!

packages/react/src/ActionList/List.tsx
siddharthkp: We don't have one yet, but I think it's possible. The [issue linked to menu + menubar pattern](https://arc.net/l/quote/brrnmqyb), so I included it just in case

packages/react/src/ActionList/List.tsx
siddharthkp: focus zone can only be added, not removed. So, if the parent AnchoredOverlay has a focus zone, that would work on all it's children. Setting it to false here would not disable to parent. (ActionMenu, Autocomplete, etc. also use AnchoredOverlay)

packages/react/src/ActionList/List.tsx
siddharthkp: > I'm wondering if we'd want to add focusOutBehavior: 'wrap' by default

Honestly, not sure, what do you think?

> or allow consumers to add that themselves?

That's the part that's missing right now. I think I'd like to wait for feedback/requests before introducing `focusZoneSettings` like we have on other focus-zoning components

TylerJDev: For `role=menu`, we'd definitely want it to wrap by default. With `role=listbox` there may be cases where it doesn't need to wrap. I'd say if we can conditionally set it to `wrap` based on if it's a menu or not would be a safe bet, at least for now!

siddharthkp: Done!

siddharthkp:
Updated focusOutBehavior to wrap, will test with dotcom once again before merging

PR #4796, opened 2024-07-30
chore: add package version numbers

joshblack:
It seems like there is a bug in changesets that will fail in prerelease mode if there is no version available for a private workspace. This PR adds explicit version numbers to each workspace to prevent this from occurring.

#### Changed

- Add explicit version numbers to internal workspaces

### Rollout strategy

This is a change to internal private packages


mperrotti (approved):
I don't actually know what these changes do, but it looks like it's just a test so I'm approving.

PR #4799, opened 2024-07-31
Token: Add text alternative to close control

TylerJDev:
Closes https://github.com/github/primer/issues/3582 https://github.com/github/primer/issues/3443

#### New

* Add new prop `removeButtonLabel`

#### Changed
* Moved visually hidden text into token container

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Token/Token.tsx
```diff
@@ -40,6 +44,7 @@ const Token = forwardRef((props, forwardedRef) => {
     onRemove,
     id,
     leadingVisual: LeadingVisual,
+    removeButtonLabel = 'Remove token',
```
siddharthkp, line 47: Just wondering, is this a safe default? Where are Tokens used in dotcom and would it be obvious what "token" is referring to?

Open questions: 
- Is no default better than confusing default?
- Is there a better default like `Remove ${props.text}`?

packages/react/src/Token/Token.docs.json
```diff
@@ -51,6 +51,12 @@
     {
       "name": "sx",
       "type": "SystemStyleObject"
+    },
+    {
+      "name": "removeButtonLabel",
+      "type": "string",
+      "defaultValue": "'Remove token'",
+      "description": "Label to give to the remove button for assistive technologies, such as screen readers"
```
siddharthkp, line 59: I think this API is becoming cumbersome because it feels like it's an after thought

We have 3 props `onRemove`, `hideRemoveButton` and `removeButtonLabel` that all affect one button and clash with each other :(

I'd love to clean this up in a backward compatible way. Do you think we could/should address that as part of this work?

siddharthkp (commented):
Left a couple questions

packages/react/src/Token/Token.tsx
TylerJDev: Yup that's fair! I only kept it as `Remove token` since that was the default aria-label on it. We can definitely customize it a bit more, `Remove "${props.text}" token`.

> Is no default better than confusing default?

Yeah, making consumers add the label themselves is sometimes the better option. In this case, since it's already using an `aria-label="Remove token"`, I'm thinking the most backwards compatible way is to make it customizable while keeping the default? 🤔 

> Where are Tokens used in dotcom and would it be obvious what "token" is referring to?

Usage looks pretty limited [~ 3 instances](https://musical-adventure-wlr3n3k.pages.github.io/?name=token&attribute=onremove), which is good! I think by adding `${props.text}`, the "token" should be described in a way that users will know what it is.

packages/react/src/Token/Token.docs.json
TylerJDev: This is a good point! I had a lot more planned for this PR, but unfortunately this is so intertwined with [`TextInputWithTokens`](https://primer.style/components/text-input-with-tokens/react/alpha) it was hard to add meaningful changes without breaking something in that component 😕

I do think that we'd want to keep `removeButtonLabel` as there's no current way to change the "Remove token" `aria-label` attached on the token button now. I was planning on coming back to `Token` in https://github.com/github/primer/issues/3468, as this issue has the potential to require a lot of changes to both components.

We could remove `removeButtonLabel` and handle it later since that isn't part of any of the issues in https://github.com/github/primer/issues/3571, but an accessibility enhancement, or we could ship these changes and work towards making this component a bit cleaner and scope that into https://github.com/github/primer/issues/3468. Curious what you think though!

TylerJDev:
@siddharthkp, with the idea of us rescoping some of the work needed for `Token` in https://github.com/github/primer/issues/3468, I'm wondering if this PR is good to be shipped? Since [the two are intertwined](https://github.com/primer/react/pull/4799#discussion_r1715793406), I was going to work towards adding more changes to `Token` in that issue, as the additions in this PR wouldn't need to be changed.

TylerJDev:
@siddharthkp, removed all instances of the prop! Should only consist of the storybook changes now

packages/react/src/Token/Token.tsx
```diff
@@ -96,8 +96,11 @@ const Token = forwardRef((props, forwardedRef) => {
           <LeadingVisual />
         </LeadingVisualContainer>
       ) : null}
-      <TokenTextContainer {...(hasMultipleActionTargets ? interactiveTokenProps : {})}>{text}</TokenTextContainer>
-      {onRemove && <VisuallyHidden> (press backspace or delete to remove)</VisuallyHidden>}
+      <TokenTextContainer {...(hasMultipleActionTargets ? interactiveTokenProps : {})}>
+        {text}
+        {onRemove && <VisuallyHidden> (press backspace or delete to remove)</VisuallyHidden>}
+      </TokenTextContainer>
```
siddharthkp, line 102: Just confirming this change: What does bringing onRemove inside the TokenTextContainer give us?

siddharthkp: Ah, understood.

Let's run integration once just to make sure there are no jest tests that expect just the text to be present in the token

siddharthkp (approved):
Looks great!

Integration tests suggested, approving in advance

TylerJDev:
Tested in gh/gh, and integration tests look good! https://github.com/github/github/pull/341786

PR #4800, opened 2024-07-31
feat(react): enable support for CSS Modules and emitting *.css files in bundle

joshblack:
Closes https://github.com/github/primer/issues/3685

#### New

- Add support for emitting *.css files alongside *.js files in order to allow authoring styles in CSS Modules

### Rollout strategy

We've been testing out support for this incrementally in dotcom. As a result, we've landed some upstream PRs to enable support for this work. In addition, we will try out a feature flag approach to test out style-only changes before rolling them out more broadly.


langermank (approved):
YAY

PR #4804, opened 2024-07-31
chore: remove TabPanels component

joshblack:
Closes https://github.com/github/primer/issues/3692

#### Removed

- Remove TabPanels component from experimental, drafts entrypoints

### Rollout strategy

- [x] Minor release

Marking this as minor since this is an experimental component

### Testing & Reviewing

- [ ] Verify that TabPanels is no longer exported
- [ ] Verify that TabPanels does not show up in storybook
- [ ] Verify that TabPanels is not in our docs sub-site


joshblack:
bump @broccolinisoup @mperrotti !

.changeset/forty-ants-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Remove experimental TabPanels component in preference of UnderlinePanels
```
broccolinisoup, line 5: As far as I understand, we don't have an internal usage for this component but I am wondering in case for open source usages should we release this as "potentially major"?

broccolinisoup (approved):
Just left a comment about the changeset, let me know what you think!

.changeset/forty-ants-tell.md
joshblack: Good point, I'm honestly not sure 🤷 It'd be great if we wouldn't even need to do a semver bump for this since it's under experimental/drafts but it's clearly nice to know if one of these things get removed. I think because of that I'm torn since we definitely don't want these to be considered under semver but it's also nice to communicate this stuff 😕 

This is where the `minor` bump came from just to say that things are changing but happy to say `major` to make this simpler 👍

broccolinisoup: Oh this is draft!! Sorry I missed that. Yeah no worries. minor sounds good then.

joshblack:
bump @mperrotti when you have a sec!

langermank (approved):

PR #4809, opened 2024-08-01
test(e2e): add vrt for blankslate

joshblack:
Add VRT for Blankslate component before moving to CSS Modules

#### New

- Add e2e test for Blankslate for VRT and AAT

### Rollout strategy


langermank (approved):
Nice 👍

PR #4810, opened 2024-08-01
refactor(Blankslate): add support for css modules to Blankslate

joshblack:
Context https://github.com/github/primer/issues/3696

Refactor Blankslate to use CSS Modules behind a feature flag

#### Changed

- Update Blankslate component to use CSS Modules
- Add feature flag to Blankslate component to toggle between styled-components and CSS Modules
- Use feature flag for VRT in Blankslate
- Add fallback VRT tests for Blankslate
- Update script for generating export sizes to use noop plugin for CSS in entrypoints

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

- Verify that no regressions have occurred in VRT
- Verify that snapshots have been added for styled-component styles


joshblack:
FYI @siddharthkp here was one idea for feature flagging in the classes 👀

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -143,16 +158,31 @@ function Blankslate({border, children, narrow, spacious}: BlankslateProps) {
 export type VisualProps = React.PropsWithChildren
 
 function Visual({children}: VisualProps) {
-  return <span className="Blankslate-Visual">{children}</span>
+  const enabled = useFeatureFlag('primer_react_css_modules')
+  return (
+    <span
+      className={cx('Blankslate-Visual', {
+        [classes.Visual]: enabled,
```
siddharthkp, line 165: _Note for feature flags only_: This is cool! It's a little bit cheating because the component already had classNames 😅

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -124,6 +127,18 @@ const BlankslateContainerQuery = `
...
+
+  if (enabled) {
+    return (
+      <div className={classes.Container}>
+        <div className={classes.Blankslate} data-border={border} data-narrow={narrow} data-spacious={spacious}>
+          {children}
+        </div>
+      </div>
+    )
+  }
```
siddharthkp, line 140: _Note for feature flags only_: Idk how to feel about 2 completely different blocks. On one hand, it's very clear that one uses classNames without the Styled(component)Blankslate. On the other hand, if StyledBlankslate accepted sx, we wouldn't be able to do this and would need to merge them in some way.

I think this code looks good and we should ship it + we'll need to migrate a couple other components to see what our practices should be

joshblack: Agreed, this one really feels like a one off because of how simple it is. In reality I think we'll have more mix-and-match and toggling inline

langermank (approved):

packages/react/src/Blankslate/Blankslate.module.css
```diff
@@ -0,0 +1,109 @@
...
+.Heading,
+.Description {
+  margin: 0;
+  /* stylelint-disable-next-line primer/spacing */
+  margin-bottom: var(--stack-gap-condensed, var(--base-size-8));
+}
+
+.Heading {
+  font-size: var(--text-title-size-medium, 1.25rem);
+  font-weight: var(--text-title-weight-medium, 600);
```
jonrohan, line 39: We don't need to add fallbacks for the values here, the PostCSS parser inlines them for us.

PR #4816, opened 2024-08-06
export FilteredActionList

siddharthkp:
Closes #TODO

### Changelog

While work on SelectPanel is on-going, there is a gap between the design of SelectPanel and the implementation in PRC.

In order to fill the gap with a temporary solution, developers can opt to build a custom implementation. Exporting FilteredActionList, which lies under SelectPanel, would help.

More context on Slack: https://github.slack.com/archives/CSGAVNZ19/p1722543563922769?thread_ts=1722472095.221459&cid=CSGAVNZ19

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/index.ts
```diff
@@ -206,3 +206,6 @@ export type {PageHeaderProps} from './PageHeader'
 export {SSRProvider, useSSRSafeId} from './utils/ssr'
 export {default as sx, merge} from './sx'
 export type {SxProp} from './sx'
+
+export {FilteredActionList} from './FilteredActionList'
+export type {FilteredActionListProps} from './FilteredActionList'
```
siddharthkp, line 211: Question for reviewer: 

Should we export this from `experimental` instead? It isn't really experimental because it's been what powers SelectPanel from the start.

But if we want to treat this as a temporary solution, then it feels okay

joshblack: I'm down with experimental if it means we won't have to go through a deprecated flow 😂 But if we still have to go through that, then stable is fine 👍 

(Separate from this PR) this could be a good example for spots where an `unstable_*` convention could help out without having to couple things to entry points

broccolinisoup: My only comment is that it would be great to treat it as an experimental/unstable component so that we don't have to worry about semver in case if we need to make any changes for the SelectPanel remediation.

broccolinisoup (commented):
Left a comment! Let me know your thoughts!

packages/react/src/index.ts
siddharthkp: Valid!

joshblack (approved):

PR #4819, opened 2024-08-06
refactor(Heading): update component to CSS Modules

joshblack:
Part of https://github.com/github/primer/issues/3716

#### Changed

- Refactor `Heading` to use CSS Modules

### Rollout strategy

- [x] Minor release


langermank (approved):

packages/react/src/Heading/Heading.tsx
```diff
@@ -50,13 +57,32 @@ const Heading = forwardRef(({as: Component = 'h2', variant, ...props}, forwarded
     }, [innerRef])
   }
 
+  if (enabled) {
+    if (sx) {
```
jonrohan, line 61:
```suggestion
    if (props.sx) {
```

joshblack:
Looks like examples failure is the same thing you run into @jonrohan over in #4828 👀 Will wait for that to merge 👍

jonrohan (approved):

PR #4822, opened 2024-08-07
docs: remove docs sub-site

joshblack:
Closes https://github.com/github/primer/issues/3334

Removes our docs sub-site from PRC 🫡 With the latest redirects added to primer.style, we no longer need to publish this sub-site as all pages now live on primer/design.

This work is a part of our v37 release as our dependency on doctocat means we cannot remove certain types of entrypoints from Primer React.

This change will keep our GitHub Pages deploy previews but the root route of this temporary URL will now 404. In the future, we could consider adding a root file to help with navigating different deployment artifacts in these previews (like Playwright test reports!)

#### Removed

- Remove `docs` sub-site
- Remove references to `docs` workspace in changesets and npm

### Rollout strategy

This change will not impact the public API of Primer React


siddharthkp (commented):
Question: would removing the docs sub-site remove our ability to have permalinks to older versions?

joshblack:
@siddharthkp thanks for bringing this up! It definitely seems like we wouldn’t be able to use this one anymore 😞 I _think_ we would need to use the primer/design GH Pages deployments for this but I’m not sure what the best way would be to get snapshots of that like we do for this currently.

siddharthkp:
> I think we would need to use the primer/design GH Pages deployments for this but I’m not sure what the best way would be to get snapshots of that like we do for this currently.

Let's tackle that once we're on the new docs. Until then, let's keep our canary docs deployments to use as permalinks

siddharthkp:
@joshblack Sorry, catching up on PRs. Did we decide what we would like to do here?

siddharthkp (approved):
Approving to make progress

PR #4838, opened 2024-08-12
feat(Banner): add support for custom icons when variant="upsell"

joshblack:
Closes https://github.com/github/primer/issues/3784

Update Banner to support customizing icons with both `info` and `upsell`

#### New

- Add support for custom icons when Banner variant="upsell"

### Rollout strategy

- [x] Minor release


siddharthkp (approved):
LGTM!

PR #4855, opened 2024-08-16
simplify syncing updates for theme provider

mattcosta7:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ThemeProvider.tsx
```diff
@@ -101,22 +102,9 @@ export const ThemeProvider: React.FC<React.PropsWithChildren<ThemeProviderProps>
...
-  }, [props.colorMode, fallbackColorMode])
-
-  React.useEffect(() => {
-    setDayScheme(props.dayScheme ?? fallbackDayScheme ?? defaultDayScheme)
-  }, [props.dayScheme, fallbackDayScheme])
-
-  React.useEffect(() => {
-    setNightScheme(props.nightScheme ?? fallbackNightScheme ?? defaultNightScheme)
-  }, [props.nightScheme, fallbackNightScheme])
-
```
mattcosta7, line 119: these are all synced automatically in the `useSyncedState` calls and avoid `useEffect` so we get to skip a render pass every time one changes

packages/react/src/hooks/useSyncedState.tsx
```diff
@@ -0,0 +1,28 @@
...
+  {isPropUpdateDisabled = false, isEqual = Object.is} = {},
+) => {
+  const [state, setState] = useState(initialValue)
+  const [previous, setPrevious] = useState(initialValue)
+
+  const nextInitialValue = initialValue instanceof Function ? initialValue() : initialValue
+  if (!isPropUpdateDisabled && !isEqual(previous, nextInitialValue)) {
+    setPrevious(nextInitialValue)
+    setState(nextInitialValue)
+  }
```
mattcosta7, line 25: these `setState` calls in render are all good because they're done in the same component.  This will immediately queue an update and re-render instead of completing a render, then comparing in a useEffect, queuing an update and then rendering again then re-comparing. 

the main value is we save a full render from react on any change - a render we _know_ will immediately be discarded

packages/react/src/ThemeProvider.tsx
mattcosta7: also, I doubt these get hit often, but I saw this while looking at something related and just figured why not

joshblack (approved):
LGTM! Thanks for doing this 🥳

PR #4856, opened 2024-08-16
feat(postcss): add support for mixins to postcss-preset-primer

joshblack:
Add support for mixins to `postcss-preset-primer`. This change also includes a test setup so that we can verify that the preset is correctly transforming mixins, syntax, and more.

#### New

- Add support for mixins to postcss-preset-primer
- Add support for tests to postcss-preset-primer

### Rollout strategy

This does not impact the public API of Primer

### DIFF
--- a/packages/postcss-preset-primer/src/index.js
+++ b/packages/postcss-preset-primer/src/index.js
@@ -4,6 +4,7 @@ import {fileURLToPath} from 'node:url'
 import {globSync} from 'glob'
 import postcssGlobalData from '@csstools/postcss-global-data'
 import postcssPresetEnv from 'postcss-preset-env'
+import postcssMixins from 'postcss-mixins'
 // @ts-ignore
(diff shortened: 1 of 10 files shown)


langermank (approved):
What a gift! Thank you!

packages/postcss-preset-primer/src/index.js
```diff
@@ -64,6 +65,9 @@ const postcssPresetPrimer = () => {
         }),
       }),
       ...plugins,
+      postcssMixins({
+        mixinsDir: path.join(path.dirname(filepath), 'mixins'),
+      }),
```
jonrohan, line 70: I think this should come before `...plugins` so that the CSS that it generates also gets transformed by the preset plugins

joshblack: @jonrohan great point, done!

PR #4870, opened 2024-08-21
fix: Add :where selector to classes that land on nodes that have a sx prop

jonrohan:
This addresses a regression found in the last release of PRC https://github.com/github/github/pull/338276#issuecomment-2301557787

The problem is when the component has an `sx=` prop applied to it, the styled-components styles are being overridden by the CSS modules styles. This is because the two classes have equal specificity `0,1,0`. 

The solution here is to add the `:where` selector to the CSS modules classes whenever they might exist on the same dom node as a sx prop. This will give the CSS modules classes a lower specificity `0,0,0` and will prevent the styled-components styles from being overridden.

**Before**

CleanShot 2024-08-21 at 11 56 45@2x

**After**

CleanShot 2024-08-21 at 11 56 54@2x

#### Changed

Add `:where()` selector to classes that land on nodes that have a `sx` prop.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

I'm going to test this locally, then also on a review-lab instance.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/blue-stingrays-decide.md
+++ b/.changeset/blue-stingrays-decide.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Add `:where()` selector to classes that land on nodes that have a `sx` prop.
--- a/packages/react/src/Heading/Heading.module.css
+++ b/packages/react/src/Heading/Heading.module.css
@@ -1,4 +1,4 @@
-.Heading {
+:where(.Heading) {
   margin: 0;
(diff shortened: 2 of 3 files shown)


langermank (approved):
Awesome!

keithamus (approved):

PR #4876, opened 2024-08-23
ActionList: Add more checks for `ActionList.Item` when using button semantics

TylerJDev:
Part of https://github.com/github/primer/issues/3722

#### Changed

* Adds more conditional checks for `ActionList.Item`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/Item.tsx
```diff
@@ -112,7 +120,12 @@ export const Item = React.forwardRef<HTMLLIElement, ActionListItemProps>(
 
     const itemSelectionAttribute = selectionAttribute || inferredSelectionAttribute
     // Ensures ActionList.Item retains list item semantics if a valid ARIA role is applied, or if item is inactive
-    const listSemantics = listRole === 'listbox' || listRole === 'menu' || inactive || container === 'NavList'
+    const listItemSemantics =
+      role === 'option' || role === 'menuitem' || role === 'menuitemradio' || role === 'menuitemcheckbox'
+
+    const listRoleTypes = ['listbox', 'menu', 'list']
```
siddharthkp, line 126: Oooh, `list` is a new addition here. Can you say a bit more about why `list` was added here and what the remaining cases when we want buttonSemantics to be used?

siddharthkp (commented):
1 suggestion, couple questions

packages/react/src/ActionList/Item.tsx
```diff
@@ -24,6 +24,14 @@ import VisuallyHidden from '../_VisuallyHidden'
 
 const LiBox = styled.li<SxProp>(sx)
 
+const ButtonItemContainer = React.forwardRef(({as: Component = 'button', children, styles, ...props}, forwardedRef) => {
+  return (
+    <Box as={Component as React.ElementType} ref={forwardedRef} sx={styles} {...props}>
```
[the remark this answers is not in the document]
TylerJDev: I saw we were already handling this on [line #329](https://github.com/primer/react/blob/63816188a218645de10fe9fcb7773356fb9724c5/packages/react/src/ActionList/Item.tsx#L329C21-L329C67), so now we pass the same logic to `ButtonItemContainer`, but under the styles prop instead of handling it inside of the component.

packages/react/src/ActionList/Item.tsx
TylerJDev: We don't have any live cases of `role="list"` with `ActionList`, so this was mainly just an escape hatch. There are odd cases out there where the new semantics might not be as ideal, one I can think of is when ActionList is utilized as a way to load in items (e.g. [loading items story](https://ui.githubapp.com/storybook/?path=/story/others-listviewitemsissuesprs--list-loading-example)). The only issue is, they probably shouldn't be interactive if they're `role="list"`. 

Not really beholden to having it here, but thought it could be useful as a escape hatch for components that still need the old semantics, at least temporarily.

lesliecdubs:
👋🏻 Bumping this for priority. The associated issue to this PR https://github.com/github/primer/issues/3722 was listed as a short term repair item, which has an SLA for repair of 60 days and we are now 51 days _overdue_. Sid is out of office right now so we'll need to identify another reviewer. @camertron as the other engineer tagged for review, could you please add this to your queue to get reviewed this week or very early next week?

camertron (approved):
Makes sense to me 👍

PR #4878, opened 2024-08-23
ProgressBar: Adjust `ProgressBar.Item` for accessibility

TylerJDev:
Closes https://github.com/github/primer/issues/3359

#### Changed

* Moves `aria-*` attributes to `ProgressBar.Item`
* Marks `ProgressBar.Item` as `role="progressbar"`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ProgressBar/ProgressBar.tsx
```diff
@@ -57,36 +55,65 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
+      'aria-valuemax': 100,
     }
 
     warning(
-      children &&
+      !ariaAttributes['aria-valuenow'] &&
         typeof (rest as React.AriaAttributes)['aria-valuenow'] === 'undefined' &&
         typeof (rest as React.AriaAttributes)['aria-valuetext'] === 'undefined',
       'Expected `aria-valuenow` or `aria-valuetext` to be provided to <ProgressBar>. Provide one of these values so screen reader users can determine the current progress. This warning will become an error in the next major release.',
     )
```
TylerJDev, line 75: I'm curious if there's a reason why we didn't include `ariaAttributes['aria-valuenow']` as a value to include in the `warning` conditional? Since we automatically create the `aria-valuenow` based on `progressAsNumber` being true and a number, is that sufficient enough if consumers don't provide an `aria-valuenow` themselves? Also, I added `progressAsNumber >= 0` so we can include `aria-valuenow` values that start at 0.

Cc: @kendallgassner - would love your insight, but I realize that it's been a minute since we've introduced these changes 😅

kendallgassner: huh, your right that is odd and if a user passed it in it would override the value, we are setting via `...rest`. 🤔 I think it's safe to remove the warning all together.

packages/react/src/ProgressBar/ProgressBar.tsx
```diff
@@ -60,37 +57,76 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
-
-    const progressAsNumber = typeof progress === 'string' ? parseInt(progress, 10) : progress
-
-    const ariaAttributes = {
-      'aria-valuenow': progressAsNumber ? Math.round(progressAsNumber) : undefined,
-      'aria-valuemin': 0,
-      'aria-valuemax': 100,
-      'aria-valuetext': progressAsNumber ? `${Math.round(progressAsNumber)}%` : undefined,
-    }
+    const validChildren = React.Children.toArray(children).length
```
joshblack, line 114: Would https://react.dev/reference/react/Children#children-count work here as well for getting the length?

packages/react/src/ProgressBar/ProgressBar.tsx
```diff
@@ -60,37 +57,76 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
+    }
+
+    return (
+      <ProgressItem
+        role="progressbar"
+        aria-label={ariaLabel}
+        ref={forwardRef}
+        progress={progress}
+        {...ariaAttributes}
+        {...rest}
```
joshblack, line 84: One suggestion to prevent accidental overrides of earlier props if that'd be helpful. This showed up recently for `className` specifically so just wanted to share 👍 

```suggestion
        {...rest}
        role="progressbar"
        aria-label={ariaLabel}
        ref={forwardRef}
        progress={progress}
        {...ariaAttributes}
```

packages/react/src/ProgressBar/ProgressBar.tsx
```diff
@@ -60,37 +57,76 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
+export const Item = forwardRef<HTMLSpanElement, ProgressBarItems>(
+  (
+    {progress, 'aria-label': ariaLabel, 'aria-valuenow': ariaValueNow, 'aria-valuetext': ariaValueText, ...rest},
+    forwardRef,
+  ) => {
+    const progressAsNumber = typeof progress === 'string' ? parseInt(progress, 10) : progress
+
+    const ariaAttributes = {
+      'aria-valuenow':
+        ariaValueNow || (progressAsNumber !== undefined && progressAsNumber >= 0 ? Math.round(progressAsNumber) : 0),
```
joshblack, line 71: Small suggestion for `??` over `||` for fallback values, no real different though 👍 

```suggestion
        ariaValueNow ?? (progressAsNumber !== undefined && progressAsNumber >= 0 ? Math.round(progressAsNumber) : 0),
```

packages/react/src/ProgressBar/ProgressBar.tsx
```diff
@@ -60,37 +57,76 @@ const ProgressContainer = styled.span<StyledProgressContainerProps>`
...
-      'aria-valuetext': progressAsNumber ? `${Math.round(progressAsNumber)}%` : undefined,
-    }
+    const validChildren = React.Children.toArray(children).length
 
     return (
-      <ProgressContainer ref={forwardRef} role="progressbar" barSize={barSize} {...ariaAttributes} {...rest}>
-        {children ?? <Item data-animated={animated} progress={progress} bg={bg} />}
+      <ProgressContainer ref={forwardRef} barSize={barSize} {...rest}>
+        {validChildren ? (
+          children
```
joshblack, line 119: For this check, would it be enough to check for the presence of children and toggle or is the count important / used somewhere else?

packages/react/src/ProgressBar/ProgressBar.tsx
TylerJDev: I tried this but ran into an issue where non-components would be included in the count:

```tsx
<ProgressBar barSize="small" aria-label="Upload test.png">
  {false}
  {false}
  {false}
</ProgressBar>
  ```
  
This resulted in `React.Children.count(children)` to equal 3 😕 This isn't really a big deal, but I noticed this via a conditional in dotcom that was causing an error, and it was because the children were booleans instead of actual components 😅

File: packages/react/src/ProgressBar/ProgressBar.tsx (hunk 3 above)
Comment (TylerJDev) [reply packages/react/src/ProgressBar/ProgressBar.tsx:84]: Interesting! Thank you for the suggestion!

File: packages/react/src/ProgressBar/ProgressBar.tsx (hunk 4 above)
Comment (TylerJDev) [reply packages/react/src/ProgressBar/ProgressBar.tsx:71]: `??` looks cooler, so let's roll with it 😎

File: packages/react/src/ProgressBar/ProgressBar.tsx (hunk 5 above)
Comment (TylerJDev) [reply packages/react/src/ProgressBar/ProgressBar.tsx:119]: Right now, you can use the `ProgressBar` alone, or with children. If you use it alone, you can apply all the props directly to `<ProgressBar>`, but with children, you need to apply them to the individual items. 

I think I understand the ask, are you asking if we can switch up the current conditional? 🤔

File: packages/react/src/ProgressBar/ProgressBar.tsx (hunk 2 above)
Comment (joshblack) [reply packages/react/src/ProgressBar/ProgressBar.tsx:114]: Oh wow, TIL https://react.dev/reference/react/Children#children-toarray:~:text=Empty%20nodes%20(null%2C%20undefined%2C%20and%20Booleans)%20will%20be%20omitted%20in%20the%20returned%20array 🤯 

Would it be possible to add a comment about this? I had no idea and wanted to suggest this in case it would help others out, maybe something like:

```suggestion
    // Get the number of non-empty nodes passed as children, this will exclude
    // booleans, null, and undefined
    const validChildren = React.Children.toArray(children).length
```

joshblack (approved):
Looks great! Just left a small suggestion in case it was helpful 👍

packages/react/src/ProgressBar/ProgressBar.tsx
TylerJDev: Good idea, thank you for the suggestion and link 👀

PR #4889, opened 2024-08-27
Add @primer/engineer-reviewers to snapshots

jonrohan:
Let @primer/engineer-reviewers review the snapshots also.

I believe this will reduce friction in getting reviews for changes that effect both screenshots and code


langermank (approved):
I like this if true 😄

PR #4890, opened 2024-08-27
Add clarity to Timeline, TreeView, and Truncate stories

mperrotti:
- Makes Timeline `clipSidebar` example easier to understand
- Uses clearer name for TreeView Storybook story that shows skeleton loading nodes
- Adds missing Truncate Storybook stories for the following props
  - `maxWidth`
  - `expandable`
  - `inline`


langermank (approved):
LGTM 😄

PR #4891, opened 2024-08-27
`StackItem` responsive grow bug fix

langermank:
`StackItem` wasn't respecting the `grow` property for responsive values due to the boolean nature. This PR leaves the boolean in place, but returns a `true` or `false` string in the data-attribute instead.


packages/react/src/Stack/__tests__/StackItem.test.tsx
```diff
@@ -19,8 +19,8 @@ describe('StackItem', () => {
         <StackItem data-testid="grow-false" />
       </Stack>,
     )
-    expect(screen.getByTestId('grow-true')).toHaveAttribute('data-grow', '')
-    expect(screen.getByTestId('grow-false')).not.toHaveAttribute('data-grow')
+    expect(screen.getByTestId('grow-true')).toHaveAttribute('data-grow="true"')
```
langermank, line 22: I almost got it, but this is returning `null` 🤔

mperrotti (approved):

PR #4905, opened 2024-09-03
refactor(storybook): update storybook to use postcss-preset-primer

joshblack:
Update our storybook in `primer/react` to use our internal postcss-preset-primer package

#### Changed

- Update storybook to use our internal postcss-preset-primer-package
- Update the storybook.css file to no longer use CSS Modules syntax as it does not have that extension

### Rollout strategy

This is a change to our storybook environment and does not impact the public API of Primer React


jonrohan (approved):
🥳

jonrohan (approved):

PR #4917, opened 2024-09-04
fix(Text): update type cast to match styled-component type

joshblack:
Update the types for our Text component to match downstream expectations by using `StyledComponent`. This helps to address an issue blocking the current release

#### Changed

- Update type cast for `Text` from `PolymorphicForwardRef` to `StyledComponent`


jonrohan (approved):
Thanks for fixing this

PR #3399, opened 2023-06-09
Make styled-components types an optional peer peer dependency

mperrotti:
Closes https://github.com/primer/react/issues/2894

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated documentation
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #4920, opened 2024-09-05
Select Panel: Add built-in "No matches" item

siddharthkp:
| Without feature flag | With feature flag |
|--------|--------|
SelectPanel with an empty list | SelectPanel with 1 item in list that says "No matches" | 

- Only works for modern ActionList (feature flag on). See visual snapshot changes to see what it looks like [with flag](https://github.com/primer/react/pull/4920/files#diff-5ffaac10f839dcef19b74d856e886fce39eea54c6c9c2aea62077d3938a92e40) and [without flag](https://github.com/primer/react/pull/4920/files#diff-c27901ed2f6c2f41e1bc2ec41e5dfc8f0548ec36995d59be3af2698a2583f24a) (flag: [`primer_react_select_panel_with_modern_action_list `](https://devportal.githubapp.com/feature-flags/primer_react_select_panel_with_modern_action_list/overview))
- Visual changes only, screen reader announcements are tackled in https://github.com/primer/react/pull/4927, can be merged independently 
- The design matches what teams are using in production (more context in https://github.com/github/primer/issues/3409#issuecomment-2331799508). We might want to change the design later when we work on this epic: https://github.com/github/primer/issues/3399

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


camertron (changes requested):
Hey @siddharthkp, thanks for working on this 😄 However I'm not sure I understand the circumstances for using a "No matches" item. Is there a specific use-case or mockup you're working from?

siddharthkp:
I agree with you. I was under the impression https://github.com/github/primer/issues/3922 was not part of the current epic and we would only work on it next quarter with https://github.com/github/primer/issues/3399:) 

Happy to update the designs

> Since the empty state is officially part of the mockups design provided, I think we should avoid supporting an official "No matches" item and encourage teams to migrate.

I'm not sure if I agree on the last bit about encourage teams to migrate. It would help us to have a default no-results state that matches the new design with the option to customise the message.

camertron:
> I agree with you. I was under the impression https://github.com/github/primer/issues/3922 was not part of the current epic and we would only work on it next quarter with https://github.com/github/primer/issues/3399)

Ah yeah, I think this work was originally scoped under #3399 but I moved it into the group of issues due 9/27 because it's related pretty directly to accessibility.

> I'm not sure if I agree on the last bit about encourage teams to migrate. It would help us to have a default no-results state that matches the new design with the option to customise the message.

The good news is that "No matches" items will continue to work, so teams can either migrate and reap the accessibility rewards or choose to stick with what they already have and maintain existing functionality.

> On second thoughts, given how loading states (https://github.com/github/primer/issues/3921) is very similar design wise. I think we should address this after merging that.

Yes! That was my thinking as well.

PR #4928, opened 2024-09-06
Bug fix: `invisible` Button variant missing background color when disabled

langermank:
Bug fix: invisible button variant missing bg color when disabled


jonrohan (approved):

packages/react/src/Button/__tests__/__snapshots__/Button.test.tsx.snap
```diff
@@ -1680,6 +1680,7 @@ exports[`Button styles invisible button appropriately 1`] = `
   cursor: not-allowed;
   box-shadow: none;
   color: primer.fg.disabled;
+  background-color: var(--button-invisible-bgColor-disabled,undefined);
```
joshblack, line 1683: What should the fallback value here be? Seems to be showing up as `undefined` 🤔

langermank: Oh weird... lemme fix that

PR #4929, opened 2024-09-06
Select panel loading states

camertron:
Closes https://github.com/github/primer/issues/3921

#### Body spinner

Alt: a video of a button that, when clicked, opens a `SelectPanel`. There is a spinner where the list of items should be. After a few seconds, the spinner is replaced by the list of items.

https://github.com/user-attachments/assets/9124a657-0ee4-4618-bc88-6803706a80de

#### Body skeleton

Alt: a video of a button that, when clicked, opens a `SelectPanel`. There is a list of animated skeletons where the list of items should be. After a few seconds, the skeleton is replaced by the list of items.

https://github.com/user-attachments/assets/6be5e641-61cf-401b-b60e-e5575e93dabf

#### Input spinner

Alt: a video of an open `SelectPanel`. The letter "b" is entered into the filter input field. After a slight delay, a spinner appears inside the input field, replacing the magnifying glass icon that was there before. A few seconds later, the spinner is replaced the magnifying glass and the list of items is replaced by a filtered list of items. Each item contains the letter "b."

https://github.com/user-attachments/assets/27ca0751-e7e9-4995-8865-e0611cb4087f

#### Changed

In order to trigger loading the initial list of items, I had to manually call the `onFilterChange` callback, which is expected to be called with the filter input value (i.e. the text the user entered) and the corresponding `React.ChangeEvent` object. I was unable to figure out how to call it with a dummy `React.ChangeEvent` object, so I changed the type signature of the callback to allow `undefined` for the event object.

I used primer-query to identify all the existing usages of `SelectPanel` that pass a `onFilterChange` callback and have determined that nobody is actually using at second event arg. We could probably remove it entirely.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

I'm conflicted here. On the one hand, this is technically a breaking change and should therefore necessitate a major release. On the other hand, nobody in dotcom uses the second event parameter, so could be considered backwards-compatible if we consider dotcom to be our only consumer. Looking for feedback here.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TextInput/TextInput.tsx
```diff
@@ -137,6 +137,7 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(
           visualPosition="leading"
           showLoadingIndicator={showLeadingLoadingIndicator}
           hasLoadingIndicator={typeof loading === 'boolean'}
+          data-testid="text-input-leading-visual"
```
siddharthkp, line 140: Heads up: We avoid adding our own `data-testid`. For tests, using aria-label or placeholder is recommended

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-Height-Initial-with-Underflowing-Items-After-Fetch-dark-colorblind-linux.png
siddharthkp, line 1: This change is maybe unintional, because looks like default state changed from full spinner to spinner in text input?

packages/react/src/FilteredActionList/FilteredActionListLoaders.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function LoadingSkeleton({rows = 10, ...props}: {rows: number}): JSX.Element {
+  return (
+    <Box p={2}>
+      <Stack direction="vertical" justify="center" gap="condensed" {...props}>
+        {Array.from({length: rows}, (_, i) => (
+          <Stack key={i} direction="horizontal" gap="condensed" align="center">
+            <SkeletonBox width="16px" height="16px" />
+            <SkeletonBox height="10px" width={`${Math.random() * 60 + 20}%`} borderRadius={2} />
```
siddharthkp, line 51: Prefer sx over system props for things that are not part of the API (width and height)

packages/react/src/experimental/Skeleton/SkeletonBox.tsx
```diff
@@ -36,5 +39,15 @@ export const SkeletonBox = styled.div<SkeletonBoxProps>`
...
+  ${space};
+  ${color};
+  ${typography};
+  ${layout};
+  ${flexbox};
+  ${grid};
+  ${background};
+  ${border};
+  ${position};
+  ${shadow};
```
siddharthkp, line 51: We don't prefer system props in the API anymore. We recommend using sx or css modules

```suggestion
-  ${space};
-  ${color};
-  ${typography};
-  ${layout};
-  ${flexbox};
-  ${grid};
-  ${background};
-  ${border};
-  ${position};
-  ${shadow};
```

packages/react/src/experimental/Skeleton/SkeletonBox.tsx
```diff
@@ -18,7 +21,7 @@ const shimmer = keyframes`
 export const SkeletonBox = styled.div<SkeletonBoxProps>`
   animation: ${shimmer};
   display: block;
-  background-color: var(--bgColor-muted, ${get('colors.canvas.subtle')});
+  background-color: var(--skeletonLoader-bgColor, ${get('colors.canvas.subtle')});
```
siddharthkp, line 24: nice!

packages/react/src/experimental/Skeleton/SkeletonBox.tsx
```diff
@@ -2,13 +2,16 @@ import type React from 'react'
...
+import type {StyledBoxProps} from '../../Box/Box'
+import {background, border, color, flexbox, grid, layout, position, shadow, space, typography} from 'styled-system'
 
 type SkeletonBoxProps = {
   /** Height of the skeleton "box". Accepts any valid CSS `height` value. */
   height?: React.CSSProperties['height']
   /** Width of the skeleton "box". Accepts any valid CSS `width` value. */
   width?: React.CSSProperties['width']
-} & SxProp
+} & StyledBoxProps &
```
siddharthkp, line 13: Looks like a system prop addition, we don't prefer system props in the API anymore. We recommend using sx or css modules

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -231,6 +297,8 @@ export function SelectPanel({
             items={itemsToRender}
             textInputProps={extendedTextInputProps}
             inputRef={inputRef}
+            loading={isLoading}
+            loadingType={loadingType()}
```
siddharthkp, line 301: Aw, not ideal to have both loading and loadingType in the API but I guess we'll need to keep them for backward compatibility.

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -87,19 +93,69 @@ export function SelectPanel({
...
+      if (dataLoadedOnce.current) {
+        // If data has already been loaded once, delay the spinner a bit. This also helps
+        // not show and then immediately hide the spinner if items are loaded quickly, i.e.
+        // not async.
+
+        if (loadingDelayTimeoutId.current) {
+          safeClearTimeout(loadingDelayTimeoutId.current)
+        }
+
+        loadingDelayTimeoutId.current = safeSetTimeout(() => setIsLoading(true), 1000)
```
siddharthkp, line 119: Timeout matches what we have in [the docs](https://primer.style/ui-patterns/loading#adapting-to-different-wait-times) 👍

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -87,19 +93,69 @@ export function SelectPanel({
...
+
+    // Only load data when the panel is open
+    if (open) {
+      // Only trigger filter change event if there are no items
+      if (items.length === 0) {
+        // Trigger filter event to populate panel on first open
+        onFilterChange(filterValue, null)
+      }
+    }
+  }, [open, dataLoadedOnce, onFilterChange, filterValue, items])
```
siddharthkp, line 157: Can you explain a bit what we are doing here, I wasn't able to understand

Thinking out loud: when there is no data yet (`dataLoadedOnce.current = false` and `items.length = 0`)? why though?

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -87,19 +93,69 @@ export function SelectPanel({
...
   )
 
+  useEffect(() => {
+    if (isLoading) {
+      setIsLoading(false)
+      dataLoadedOnce.current = true
+    }
+    // Only fire this effect if items have changed
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [items])
```
siddharthkp, line 142: Note for myself: Come back to see if this block can be refactored to remove the eslint warning while avoiding an accidental infinite loop

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -42,6 +47,7 @@ interface SelectPanelBaseProps {
   inputLabel?: string
   overlayProps?: Partial<OverlayProps>
   footer?: string | React.ReactElement
+  initialLoadingType?: InitialLoadingType
```
siddharthkp, line 50: This is the only new API addition, correct?

(non blocker) Aw, kinda sad that we have both `loading` and `(initial)LoadingType` as props, but I think we have to keep them for backward compatibility. I was hoping we could get a `loading`/`loader` enum type, but not a big deal.

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-Height-Initial-with-Underflowing-Items-After-Fetch-dark-colorblind-linux.png
camertron: Ah good catch, I was going to address this and forgot 😅

packages/react/src/FilteredActionList/FilteredActionListLoaders.tsx
camertron: I wanted to use a value of `2` here to leverage our pre-defined spacing variables, but I can convert this to `sx`.

packages/react/src/SelectPanel/SelectPanel.tsx
camertron: Yessir!

packages/react/src/SelectPanel/SelectPanel.tsx
camertron: Yes! We do that to trigger loading items when the panel first opens. The way to trigger loading items is to tell the consumer that the filter has changed. If there are already items provided, then don't trigger loading items because they're presumably already loaded.

packages/react/src/SelectPanel/SelectPanel.tsx
camertron: Yeah, good call. I'm not 100% happy with this either, but it was the only approach I was able to get working. The idea here is that, if the component is in a loading state and items get updated, loading must be finished.

packages/react/src/SelectPanel/SelectPanel.tsx
camertron: Ugh I know 😞 

This is the only API addition. There is one additional change too - the second parameter to `onFilterChange` (the `Event` object) is now nullable.

packages/react/src/TextInput/TextInput.tsx
camertron: Oh interesting! I'm not sure I understand though, could you point me to an example? It seems strange to me to use an accessibility mechanism with specific semantics for this, especially when the component in question, `TextInputInnerVisualSlot`, is only used internally.

broccolinisoup: Sorry to barge in - wanted to share this tiny resource in case it is helpful! https://testing-library.com/docs/queries/about#priority I like how they prioritise the queries.

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -87,19 +93,69 @@ export function SelectPanel({
...
   overlayProps,
   sx,
+  loading,
+  initialLoadingType = 'spinner',
   ...listProps
 }: SelectPanelProps): JSX.Element {
+  const inputRef = React.useRef<HTMLInputElement>(null)
   const titleId = useId()
   const subtitleId = useId()
+  const dataLoadedOnce = useRef(false)
```
broccolinisoup, line 103: Could you tell me a little bit more about the reason of initialising `dataLoadedOnce` as a ref instead of a normal state variable?  Is it because to compare obj references?

packages/react/src/SelectPanel/SelectPanel.tsx
camertron: We could use an enum for this, but I chose a class + constructor because I wanted to also include the `appearsInBody` attribute so you can do eg. `FilteredActionListLoadingTypes.bodySpinner.appearsInBody`. However, this particular callsite _does_ use an enum, `InitialLoadingType`, which is defined as `'spinner' | 'skeleton'`.

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-Height-Initial-with-Underflowing-Items-After-Fetch-dark-colorblind-linux.png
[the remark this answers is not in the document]
camertron: Yes, good catch 👍

packages/react/src/SelectPanel/SelectPanel.tsx
camertron: Hahaha that's probably because I don't actually understand the difference between the two since I don't write React code every day 😅 I can switch it to use `useState`, that's a good idea 👍

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-Height-Initial-with-Underflowing-Items-After-Fetch-dark-colorblind-linux.png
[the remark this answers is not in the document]
camertron: On second thought, I think it's up to the consumer to set a `height` value, eg `height="medium"`.

broccolinisoup (approved):
Hey @camertron! Thanks for addressing the feedback and apologies for the delay. Let's merge this 🚀

PR #4939, opened 2024-09-09
TextInput: Add `LeadingVisual`/`TrailingVisual` to be part of the accessible description in `TextInput`

TylerJDev:
Closes https://github.com/github/primer/issues/3438 https://github.com/github/primer/issues/3465

#### New

* Adds `aria-describedby` for `LeadingVisual` and `TrailingVisual`
* Adds new prop `loaderText` to convey loading state to screen readers

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

(https://github.com/github/github/pull/341243, integration tests pass with minor changes, and some unrelated failures)


packages/react/src/TextInput/TextInput.tsx
```diff
@@ -24,6 +25,8 @@ export type TextInputNonPassthroughProps = {
    * 'trailing': at the end of the input
    **/
   loaderPosition?: 'auto' | 'leading' | 'trailing'
+  /** Text for screen readers to convey the loading state */
+  loaderText?: string
```
iansan5653, line 29: 🤔 I wonder if it's really necessary / a good idea for consumers to be able to customize this? It inherently means they'll be providing a different experience for sighted vs non-sighted users, because sighted users won't be able to access this text. Visually a loading indicator only indicates that something is loading, so "Loading" should be sufficient always.

Alternatively, we could make this visible by putting it in a tooltip. But that doesn't seem ideal since the loader is not focusable.

packages/react/src/TextInput/TextInput.tsx
```diff
@@ -149,12 +169,14 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(
...
           aria-invalid={validationStatus === 'error' ? 'true' : undefined}
           {...inputProps}
+          aria-describedby={inputDescribedBy}
           data-component="input"
         />
         <TextInputInnerVisualSlot
           visualPosition="trailing"
           showLoadingIndicator={showTrailingLoadingIndicator}
           hasLoadingIndicator={typeof loading === 'boolean'}
+          id={trailingVisualId}
```
iansan5653, line 179: Since the leading and trailing visuals are now accessible as part of the description, should they be hidden from screen readers as part of the normal document flow? Ie, should we add `aria-hidden`?

packages/react/src/TextInput/TextInput.tsx
```diff
@@ -96,6 +100,21 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(
...
+    const inputDescribedBy =
+      [
+        inputProps['aria-describedby'],
+        LeadingVisual && leadingVisualId,
+        TrailingVisual && trailingVisualId,
+        loading && loadingId,
+      ]
+        .filter(String)
+        .join(' ')
+        .trim() || undefined
```
iansan5653, line 116: We could use `clsx` for this. It was technically designed for classes but should work for any HTML attribute that has space-separated values.:

```suggestion
    const inputDescribedBy =
      clsx(
        inputProps['aria-describedby'],
        LeadingVisual && leadingVisualId,
        TrailingVisual && trailingVisualId,
        loading && loadingId,
      )
```

packages/react/src/TextInput/TextInput.tsx
TylerJDev: Good question, and good point! This mainly stems from https://github.com/github/primer/issues/3465. I believe the idea is, if there's context outside of the input that conveys _what_ is loading then it might be a good idea to add more descriptive text to indicate what is loading so users don't have to navigate outside of it to get that context.

I think this is a more specific use case that won't be used unless necessary, so in most cases having the default "Loading" is okay, which we could make more descriptive 🤔 

Cc: @patrickhlauke - curious if you have any additional thoughts here?

packages/react/src/TextInput/TextInput.tsx
TylerJDev: Yup, I think that's fair!

packages/react/src/TextInput/TextInput.tsx
iansan5653: > I believe the idea is, if there's context outside of the input that conveys _what_ is loading then it might be a good idea to add more descriptive text to indicate what is loading so users don't have to navigate outside of it to get that context.

If we do keep this prop, I think this (particularly the par about the context outside the input) seems like it would be useful to add to the props documentation. I think what may happen is consumers will see that this prop exists and think that they should fill it out similarly to how they fill out alt text in an image - they'll have good intentions but actually create a less accessible experience.

francinelucca (approved):
LGTM pending open conversations!

packages/react/src/TextInput/TextInput.tsx
TylerJDev: Yup, that's totally fair @iansan5653! I can add a note on this in the props docs to make sure consumers aren't haphazardly using it!

PR #4948, opened 2024-09-10
docs(examples): add theming example

joshblack:
Add a theming example to our `examples` directory. This example is to help us test out changes to `ThemeProvider` as we move over to CSS Modules.

#### New

- Add `examples/theming` example with Next.js App Router

### Rollout strategy

This is a local example that will not impact the public API of a Primer package


examples/theming/src/app/layout.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <html
+      lang="en"
+      suppressHydrationWarning
+      data-color-mode={colorMode}
+      data-light-theme="light"
+      data-dark-theme="dark"
+    >
+      <body>
+        <StyledComponentsRegistry>
+          <ThemeProvider colorMode={colorMode}>
```
jonrohan, line 28: Should we note that `<ThemeProvider` here is temporary?

jonrohan (approved):
Awesome 👍🏻

examples/theming/src/app/global.css
```diff
@@ -0,0 +1,7 @@
+@import '@primer/primitives/dist/css/functional/themes/light.css';
+@import '@primer/primitives/dist/css/functional/themes/dark.css';
```
jonrohan, line 2: We might want to include other primitives like spacing font etc. as far as color themes I think light/dark are the bare minimum and probably fine for this example

examples/theming/src/app/layout.tsx
joshblack: No preference on my end, happy to add it if it would be helpful 👍

examples/theming/src/app/layout.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    // Note: the focus-visible polyfill adds additional attributes to `html`
+    // that cause hydration mismatch errors
+    <html
+      lang="en"
+      suppressHydrationWarning
+      data-color-mode={colorMode}
+      data-light-theme="light"
+      data-dark-theme="dark"
```
jonrohan, line 24: Like the idea that this could one day be a helper function

```suggestion
      {...colorThemeAttributes(colorMode)}
```

examples/theming/src/app/global.css
langermank: You could pull in this index file for everything other than color https://unpkg.com/browse/@primer/primitives@8.2.2/dist/css/primitives.css

PR #4950, opened 2024-09-11
docs: add support for css to storybook

joshblack:
It seems like our existing CSS in place in Storybook is not working as intended and only style-loader is getting applied 😓 This PR updates the Storybook configuration to work correctly with CSS Modules and correctly apply our postcss config to it.

#### Changed

- Update storybook config to process CSS Modules

#### Removed

- Removed deprecated `@storybook/addon-styling` package

### Rollout strategy

This is a change to our storybook setup


jonrohan (approved):

packages/react/.storybook/main.ts
```diff
@@ -0,0 +1,59 @@
...
+    config.css = {
+      ...config.css,
+      modules: {
+        ...config.css?.modules,
+        generateScopedName: 'prc_[local]_[hash:base64:5]',
+      },
+      postcss: {
+        ...config.css?.postcss,
+        plugins: [postcssPresetPrimer()],
+      },
```
jonrohan, line 48: With this here now can we delete that postcss config file in the packages/react root?

joshblack: Only one way to find out 👀

jonrohan (approved):
beautiful

langermank (approved):

PR #3277, opened 2023-05-12
Use button instead of span for disabled prev/next controls

adrianababakanian:
Describe your changes here.

Per accessibility audit feedback, the current `span`s used for the prev/next links when disabled are not announcable by screen readers. Instead of `span`s, we should use `button`s so that these disabled elements are focusable.

I also updated the CSS to match the text styling of the buttons to the same styling used when the links aren't disabled.

### Screenshots

Please provide before/after screenshots for any visual changes

Before:
Screenshot 2023-05-12 at 5 00 01 PM

After:
Screenshot 2023-05-12 at 4 59 22 PM

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

Take a look at the [What we look for in reviews](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#what-we-look-for-in-reviews) section of the contributing guidelines for more information on how we review PRs.

PR #4951, opened 2024-09-11
fix(Pagination): a11y improvements

francinelucca:
Closes #https://github.com/github/primer/issues/3395

Makes the Previous/Next controls in `Pagination` non-focusable by rendering them as `a`s instead of `button`. Also adds `role="link" to all `<a/>'s in the component.

#### New

- Margin styles for disabled Next/Previous control to account for disparity between `button` borders and `a`s

#### Changed

- Render Next/Previous controls as `a` instead of `button` to make them non-focusable
- Adds `role="link"` to all `<a>`'s (Next, Previous and page numbers) in the component to improve accessibility

#### Removed

- Removes styles added for button on https://github.com/primer/react/pull/3277

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Pagination/Pagination.tsx
+++ b/packages/react/src/Pagination/Pagination.tsx
@@ -76,14 +76,11 @@ const Page = styled.a`
 
   &[aria-disabled],
   &[aria-disabled]:hover {
-    color: ${get('colors.primer.fg.disabled')}; // check
-    cursor: default;
-    background-color: transparent;
-    border-color: transparent;
-    font-size: inherit;
-    font-family: inherit;
-    padding-top: inherit;
-    padding-bottom: inherit;
+    margin: 0 2px;
+
+    &:first-child {
+      margin-right: 6px;
+    }
   }
 
   &[aria-disabled],
--- a/packages/react/src/Pagination/model.tsx
+++ b/packages/react/src/Pagination/model.tsx
@@ -148,13 +148,14 @@ export function buildComponentData(
       key = 'page-prev'
       content = 'Previous'
       if (page.disabled) {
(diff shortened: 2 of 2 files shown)


packages/react/src/Pagination/Pagination.tsx
```diff
@@ -76,14 +76,11 @@ const Page = styled.a`
...
-    border-color: transparent;
-    font-size: inherit;
-    font-family: inherit;
-    padding-top: inherit;
-    padding-bottom: inherit;
+    margin: 0 2px;
+
+    &:first-child {
+      margin-right: 6px;
+    }
```
francinelucca, line 83: I had to do this to enforce no visual regressions when changing from button to anchor to account for the 2px border all buttons have. Couldn't find a mapping in the theme spaces but happy to do this differently if there's a better alternative!

packages/react/src/Pagination/model.tsx
```diff
@@ -148,13 +148,14 @@ export function buildComponentData(
       key = 'page-prev'
       content = 'Previous'
       if (page.disabled) {
-        Object.assign(props, {as: 'button', 'aria-disabled': 'true'})
+        Object.assign(props, {'aria-disabled': 'true'})
```
TylerJDev, line 151: We can add `role="link"` here, since I believe we'd only need it when the link is disabled. This should allow us to remove it below in the else condition

TylerJDev (changes requested):
One quick comment, but otherwise looks good!

TylerJDev (approved):

PR #4954, opened 2024-09-12
fix(Banner): update alignment of actions when Banner is dismissible

joshblack:
Closes https://github.com/github/primer/issues/3957

#### Changed

- Update styling for Banner to work when the banner is dismissible, has a hidden title, and actions

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- [ ] Verify that the changes here address the issue above
- [ ] Verify that no regressions occur in layout 😅


joshblack:
Note: just realized with the snapshots that the buttons are not in the right order, looking at this now 👀

langermank (approved):

.changeset/spicy-eggs-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Update alignment of actions in Banner when it is dismissible and has a hidden title
```
siddharthkp, line 5:
```suggestion
Banner: Update alignment of actions when it is dismissible and has a hidden title
```

PR #4963, opened 2024-09-13
feat(rollup-plugin-import-css): add support for named exports

joshblack:
Update our rollup plugin to emit named imports so that we can better detect if a specific rule is unused in the future 👀 

#### Changed

- Update rollup plugin to emit named imports as part of transforming CSS Modules
- Update imports to CSS Module classes to use `import * as` (Rollup will correctly remove references we don't use and will automatically drop them from bundle)

### Rollout strategy

This is an internal change and should not impact the public API 👀


jonrohan (approved):
Cool

joshblack:
I'm going to close this one out since it seems like TypeScript is actually causing some challenges here since we tell TypeScript these modules are:

```
declare module '*.module.css' {
  const classes: {[key: string]: string}
  export default classes
}
```

And this will fail since the plugin we're using that allows for named imports (typescript-plugin-css-modules) will not run during build 😞

PR #4964, opened 2024-09-13
docs(storybook): add react plugin for fast refresh

joshblack:
With the change, it seems like Fast Refresh is not enabled by default in React Vite 🤔 This PR adds the react plugin to enable this (and get Fast Refresh for styles too)

#### Changed

- Update storybook to have Fast Refresh for React components

### Rollout strategy

This is a change to our internal storybook and will not impact the public API 

### Testing & Reviewing

- [ ] Pull down the PR, install dependencies, and very that making a change does not reload the page 👀

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1708,11 +1708,13 @@
       }
     },
     "node_modules/@babel/plugin-transform-react-jsx-self": {
-      "version": "7.24.1",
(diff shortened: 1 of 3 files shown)


jonrohan (approved):
⚡

PR #4969, opened 2024-09-16
TreeView: Add indication for "Loading" for screen readers

TylerJDev:
Closes https://github.com/github/primer/issues/3448

`TreeView` announces when content has loaded, but doesn't announce when content is loading. This PR adds "Loading" to TreeView's live region when content is actively loading.

#### Changed

* Adds "Loading" to live region in `TreeView` when loading asynchronous content

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
@ericwbailey, I added you as a reviewer as I'd love your opinion on this addition to `TreeView`. This is basically announcing "Loading" via the live region when content is actively loading. Wasn't sure if there's context I'm missing on TreeView only announcing when content has loaded, instead of both or just when it's loading.

Stems from https://github.com/github/primer/issues/3448

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -646,6 +646,11 @@ const SubTree: React.FC<TreeViewSubTreeProps> = ({count, state, children}) => {
 
         setLoadingFocused(false)
       }
+    } else if (state === 'loading') {
+      if (!parentElement) return
```
francinelucca, line 650: nit: could bubble up as well and return early on 619, since other if statement checks and returns for same condition on 6221

francinelucca (approved):
🚀

packages/react/src/TreeView/TreeView.tsx
TylerJDev: I see! So adding `if (!parentElement) return` at the top rather than in both conditionals. Thank you for the suggestion!

PR #4986, opened 2024-09-18
fix(Dialog): track mousedown event to prevent accidental closing

francinelucca:
Closes https://github.com/github/primer/issues/3477

Keeps a state of whether the last `mousedown` event originated on the backdrop to prevent closing the dialog when the backdrop is clicked accidentally.

#### New

- `lastMouseDownIsBackdrop` state variable to keep track of whether the last `mousedown` event originated on the backdrop.

#### Changed

- Modify `onBackdropClick` logic to only trigger the `onClose` callback if the `mousedown` event originates on the backdrop (i.e., `lastMouseDownIsBackdrop` is `true`).

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Follow repro instructions from [issue](https://github.com/github/primer/issues/3477) on [preview deployment](https://primer-417b07082e-13348165.drafts.github.io/storybook/?path=/story/components-dialog--default) and validate behavior is no longer experienced.

**Note:** I tried to add a unit test for this but couldn't get a combination of `mouseDown`/`mouseUp` to trigger a click on the backdrop so wasn't able to simulate this particular scenario. Open to any suggestions on how to do that.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Dialog/Dialog.tsx
```diff
@@ -419,14 +419,15 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
       footerButton.ref = autoFocusedFooterButtonRef
     }
   }
+  const [lastMouseDownIsBackdrop, setLastMouseDownIsBackdrop] = useState<boolean>(false)
   const defaultedProps = {...props, title, subtitle, role, dialogLabelId, dialogDescriptionId}
   const onBackdropClick = useCallback(
     (e: SyntheticEvent) => {
-      if (e.target === e.currentTarget) {
+      if (e.target === e.currentTarget && lastMouseDownIsBackdrop) {
```
francinelucca, line 426: for the test case (mouseDown on dialog, drag to backdrop, mouseUp), the e.target and e.currentTarget are both the same (the backdrop) since the click is generated at the backdrop on mouseUp. Circumventing this by explicitly checking that the mouseDown event that triggered the click was originated at the backdrop, but open to other ides on how to handle this if any recs!

keithamus:
Do you think it's viable to write a test for this?

keithamus:
Yes I guess any test that manually fired events would be too contrived. I would just like to find a way to preserve the functionality against refactors.

keithamus (approved):

packages/react/src/Dialog/Dialog.test.tsx
```diff
@@ -85,6 +85,25 @@ describe('Dialog', () => {
...
+    const backdrop = dialog.parentElement!
+
+    fireEvent.mouseDown(dialog)
+    fireEvent.mouseUp(backdrop)
+    // trigger the click on the backdrop, mouseUp doesn't do it for us
+    fireEvent.click(backdrop)
+
+    expect(onClose).not.toHaveBeenCalled()
+  })
+
```
francinelucca, line 106: how do you feel about this @keithamus ? I do have to manually trigger the `click` event but it will fail if the current implementation were to be refactored out 🤔

keithamus (approved):
LGTM.  Thanks for adding the test.

PR #4990, opened 2024-09-19
Overlay: Add `role="dialog"`, focus trap to `Overlay`

TylerJDev:
Closes https://github.com/github/primer/issues/3378

Usages of `Overlay` tend to lean more towards a dialog across GH. This PR aims to define it as a `dialog` in most cases, provided with a focus trap that's toggled via the newly added `focusTrap` prop.

#### New

* Add `role="dialog"` and `aria-modal="true"` to `Overlay` component

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Overlay/Overlay.docs.json
+++ b/packages/react/src/Overlay/Overlay.docs.json
@@ -113,6 +113,12 @@
       "description": "If defined, Overlays will be rendered in the named portal. See also `Portal`."
     },
     {
+      "name": "focusTrap",
+      "type": "string",
+      "defaultValue": "true",
+      "description": "Determines if the `Overlay` recieves a focus trap or not"
+    },
+    {
       "name": "sx",
       "type": "SystemStyleObject"
     }
--- a/packages/react/src/Overlay/Overlay.tsx
+++ b/packages/react/src/Overlay/Overlay.tsx
@@ -110,6 +110,7 @@ type BaseOverlayProps = {
   portalContainerName?: string
   preventFocusOnOpen?: boolean
   role?: AriaRole
+  focusTrap?: boolean
   children?: React.ReactNode
 }
 
@@ -134,6 +135,7 @@ type OwnOverlayProps = Merge<StyledOverlayProps, BaseOverlayProps>
  * @param bottom Optional. Vertical bottom position of the overlay, relative to its closest positioned ancestor (often its `Portal`).
  * @param position Optional. Sets how an element is positioned in a document. Defaults to `absolute` positioning.
  * @param portalContainerName Optional. The name of the portal container to render the Overlay into.
+ * @param focusTrap Optional. Determines if the `Overlay` recieves a focus trap or not. Defaults to `true`.
  */
 const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
   (
@@ -156,6 +158,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
       preventFocusOnOpen,
       position,
       style: styleFromProps = {},
+      focusTrap = true,
       ...rest
     },
     forwardedRef,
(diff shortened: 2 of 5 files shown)


TylerJDev:
This change essentially turns `Overlay` into a "Dialog" in the sense of `role="dialog"` and `aria-modal="true"` being added. A focus trap is also added by default, and is enabled if there are focusable elements inside of the overlay. This change will not affect overlays that already have a `role` attribute, as long as that `role` attribute isn't `role="dialog"`. 

Rollout plan: 
* Add `role="none"` to existing `Overlay` components that do not implicitly set `role` already
* Add `focusTrap={false}` to `Overlay` components that are used as Dialogs currently, but do not utilize a focus trap.

packages/react/src/Overlay/Overlay.tsx
```diff
@@ -109,6 +110,7 @@ type BaseOverlayProps = {
   portalContainerName?: string
   preventFocusOnOpen?: boolean
   role?: AriaRole
+  focusTrap?: boolean
```
TylerJDev, line 113: Added a new prop in case there's an implementation that wants to be a dialog, but doesn't want the focus trap applied. This is a very specific case, and ideally won't be necessary in more than a handful of instances.

TylerJDev:
@primer/engineer-reviewers - Hey team! Would love any additional eyes on this PR, mainly in regards [to my rollout plan](https://github.com/primer/react/pull/4990#issuecomment-2364685393).

packages/react/src/Overlay/Overlay.docs.json
```diff
@@ -112,6 +112,12 @@
       "defaultValue": "",
       "description": "If defined, Overlays will be rendered in the named portal. See also `Portal`."
     },
+    {
+      "name": "focusTrap",
+      "type": "string",
```
francinelucca, line 117: should this be a boolean instead?
```suggestion
      "type": "boolean",
```

francinelucca (commented):
LGTM! One minor suggestion and one question:
https://github.com/github/primer/issues/3378 suggests: 
> allow users to switch to/from the non-modal dialog using `Ctrl+F6` / `Command+F6`

Did we implement that? I couldn't get that to go in my tests

keithamus (changes requested):
If it is a `role=dialog` and has focus trapping, it should be a `<dialog>` element with `showModal()` called. 

We should steer away from `useFocusTrap` and use `showModal()` on `<dialog>` elements instead. It comes with many other small benefits such as retaining previously-focused-element.

TylerJDev:
> If it is a `role=dialog` and has focus trapping, it should be a `<dialog>` element with `showModal()` called.
> 
> We should steer away from `useFocusTrap` and use `showModal()` on `<dialog>` elements instead. It comes with many other small benefits such as retaining previously-focused-element.


That makes sense! I'm curious if we'd want this to be true for components that consume `Overlay` or `AnchoredOverlay` (e.g. `ActionMenu`, `SelectPanel`), as they rely on `Overlay` internally. I guess it would be worth making those components utilize `Dialog`, as we utilize it in the PVC `SelectPanel` already. But I'm curious what you think? This seems like it would require more effort and testing, but could be worth it in the long run.

Right now, this PR will only affect components that utilize `Overlay` in a standalone context. Components that utilize `Overlay` like `ActionMenu` would retain the same semantics/roles even though it technically uses it internally to render the menu. We could see about conditionally rendering a `<dialog>` vs the current `<div>` container but only in instances where it's being used as a standalone component, but I'm wondering if that's worth it or not 🤔

keithamus:
I think `Dialog` and components that use `<dialog>` can be discrete. I see little value in mapping our React library to individual HTML elements. 

As for conditionally rendering `<dialog>` vs `<div>`, that kind of seems fine to me but I’d actually sooner see us split out `ActionMenu` to _not_ use `Overlay` which is a case I’ve made before about the general reuse of these components; I don’t think there is enough overlap to consider reuse here. There can be multiple discrete types of floating UI.

PR #4991, opened 2024-09-19
Deprecate textinput contrast

### LINKED ISSUE #4355: Deprecate and remove the input `contrast` prop
The `contrast` prop, when applied to `Select`, `TextInput`, or `Textarea` (one would also expect it on `TextInputWithTokens` but it's not available there) renders an input with a subtle background, somewhere between the default and disabled styling:

screenshot of TextInput with normal, contrast, and disabled stylings

In dark theme it has no effect: 

the same screenshot in dark mode. All three inputs look the same

The documentation says this prop:

> Changes background color to a higher contrast color

But ironically it actually reduces the contrast between the text and background.

This prop is problematic because, as shown above, it can make active inputs look disabled. This can and has caused confusion for users, and so the recommendation and effort has been to avoid using it in GitHub UI (including in PVC components). We've been [considering](https://github.com/primer/react/issues/3487#issuecomment-1622579311) getting rid of it for a while, so  let's get it done.

CleverLemming1337:
Closes #4355 

This is my first contribution on this repository. I marked the `contrast` prop of TextInput and TextInputWithTokens as deprecated. I know that the issue says it should be removed, but for now I just deprecated it.

#### Changed

- Mark `contrast` of TextInput and TextInputWithTokens as deprecated.

### Rollout strategy

- Minor (versioning.md says it‘s a `minor` when a prop is deprecated)

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TextInputWithTokens/TextInputWithTokens.tsx
```diff
@@ -79,7 +79,7 @@ function TextInputWithTokensInnerComponent<TokenComponentType extends AnyReactCo
     trailingVisual: TrailingVisual,
     loading,
     loaderPosition = 'auto',
-    contrast,
+    contrast, // deprecated
```
francinelucca, line 82: I think we'd also want to add the `@deprecated` decorator to the `contrast` prop in the [StyledTextInputBaseWrapperProps](https://github.com/primer/react/blob/main/packages/react/src/internal/components/TextInputWrapper.tsx#L19) object.
Though I do see this is also used in `Select` in addition to `TextInput`,`TextArea` and `TextInputWithTokens`.
@mperrotti wondering if we're good to remove contrast for all of these or should we explore a different strategy? 🤔

francinelucca (commented):
Thanks for the contribution! Sorry in the delay in getting back with a review 🙏🏼
Mostly LGTM, just wanting to confirm proper deprecation with design (see comment below)

PR #5007, opened 2024-09-23
Add bg prop to ProgressBar.Item

lukasoppermann:
Adds a `bg` prop for on `ProgressBar.Item` as discussed in https://github.slack.com/archives/GACAW0NPM/p1726839001767409

## Follow-up task
- [x] **Once merged:** merge this one: https://github.com/primer/eslint-plugin-primer-react/pull/237

### Changelog
- Adds a `bg` prop for on `ProgressBar.Item`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/slow-walls-drum.md
+++ b/.changeset/slow-walls-drum.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+add bg prop to ProgressBar.Item
--- a/.changeset/slow-walls-drum.md
+++ b/.changeset/slow-walls-drum.md
@@ -2,4 +2,4 @@
 '@primer/react': minor
 ---
 
-add bg prop to ProgressBar.Item
+ProgressBar: Add `bg` prop to `ProgressBar.Item`
--- a/packages/react/src/ProgressBar/ProgressBar.features.stories.tsx
+++ b/packages/react/src/ProgressBar/ProgressBar.features.stories.tsx
@@ -20,7 +20,7 @@ export const Color = () => <ProgressBar progress="66" bg="done.emphasis" aria-la
 
 export const MultipleItems = () => (
   <ProgressBar aria-valuenow={70} aria-label="Upload test.png">
-    <ProgressBar.Item progress={33} bg={'accent.emphasis'} />
+    <ProgressBar.Item progress={33} sx={{bg: 'accent.emphasis'}} />
     <ProgressBar.Item progress={23} bg={'danger.emphasis'} />
     <ProgressBar.Item progress={14} bg={'severe.emphasis'} />
   </ProgressBar>
--- a/packages/react/src/ProgressBar/ProgressBar.features.stories.tsx
+++ b/packages/react/src/ProgressBar/ProgressBar.features.stories.tsx
@@ -20,9 +20,9 @@ export const Color = () => <ProgressBar progress="66" bg="done.emphasis" aria-la
 
 export const MultipleItems = () => (
   <ProgressBar aria-valuenow={70} aria-label="Upload test.png">
-    <ProgressBar.Item progress={33} />
-    <ProgressBar.Item progress={23} sx={{backgroundColor: 'danger.emphasis'}} />
-    <ProgressBar.Item progress={14} sx={{backgroundColor: 'severe.emphasis'}} />
+    <ProgressBar.Item progress={33} bg={'accent.emphasis'} />
+    <ProgressBar.Item progress={23} bg={'danger.emphasis'} />
+    <ProgressBar.Item progress={14} bg={'severe.emphasis'} />
   </ProgressBar>
 )
 
--- a/packages/react/src/ProgressBar/ProgressBar.stories.tsx
+++ b/packages/react/src/ProgressBar/ProgressBar.stories.tsx
@@ -1,8 +1,15 @@
-import React from 'react'
+import React, {useEffect} from 'react'
 import type {Meta} from '@storybook/react'
 import {ProgressBar, type ProgressBarProps} from '..'
 
-const sectionColors = ['success.emphasis', 'done.emphasis', 'severe.emphasis', 'danger.emphasis', 'attention.emphasis']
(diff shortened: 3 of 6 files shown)


.changeset/slow-walls-drum.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+add bg prop to ProgressBar.Item
```
siddharthkp, line 5:
```suggestion
ProgressBar: Add `bg` prop to `ProgressBar.Item`
```

packages/react/src/ProgressBar/ProgressBar.features.stories.tsx
```diff
@@ -20,9 +20,9 @@ export const Color = () => <ProgressBar progress="66" bg="done.emphasis" aria-la
 
 export const MultipleItems = () => (
   <ProgressBar aria-valuenow={70} aria-label="Upload test.png">
-    <ProgressBar.Item progress={33} />
-    <ProgressBar.Item progress={23} sx={{backgroundColor: 'danger.emphasis'}} />
-    <ProgressBar.Item progress={14} sx={{backgroundColor: 'severe.emphasis'}} />
+    <ProgressBar.Item progress={33} bg={'accent.emphasis'} />
+    <ProgressBar.Item progress={23} bg={'danger.emphasis'} />
+    <ProgressBar.Item progress={14} bg={'severe.emphasis'} />
```
siddharthkp, line 25: Can we add a story with `sx` prop as well to keep track of backward compat?

packages/react/src/ProgressBar/ProgressBar.tsx
```diff
@@ -7,16 +7,18 @@ import type {SxProp} from '../sx'
 import sx from '../sx'
 import {warning} from '../utils/warning'
 
-type ProgressProp = {progress?: string | number}
+type ProgressProp = {
+  progress?: string | number
```
siddharthkp, line 11: Should add `bg` as prop here as well. Bonus points if you type it based on `colors.*`

packages/react/src/__tests__/__snapshots__/ProgressBar.test.tsx.snap
```diff
@@ -4,7 +4,6 @@ exports[`ProgressBar respects the "progress" prop 1`] = `
 .c1 {
   width: 80%;
   background-color: var(--bgColor-success-emphasis,var(--color-success-emphasis,#1f883d));
-  background-color: var(--bgColor-success-emphasis,var(--color-success-emphasis,#1f883d));
```
siddharthkp, line 7: Cool diff! Looks like background-color was being set twice for no reason 😅

siddharthkp (changes requested):
Request some minor improvements

siddharthkp:
Would be super nice to have some VRT tests as well, I can add some on main so that we get a nice diff 👍

packages/react/src/ProgressBar/ProgressBar.tsx
lukasoppermann: Wouldn't this be a loooot fo colors? I am hoping we can soon get away from this and move to predefined values that map to css dsiplay colors.

Adding bg is a problem, because it is NOT optional for the item, but is optional for the progressBar. What to do? @siddharthkp

packages/react/src/ProgressBar/ProgressBar.features.stories.tsx
lukasoppermann: So with both bg and `sx`? Or do you think we should make `bg` optional?

siddharthkp: Just with sx so that we have a few examples of how it used to work.

> Or do you think we should make bg optional?

bg should definitely be optional, so that we don't break any existing instances

packages/react/src/ProgressBar/ProgressBar.tsx
siddharthkp: > Wouldn't this be a loooot fo colors? 

Yes, but same as what we support now. I'm not sure what the exact types would look like, something like `typeof keyof colors`, i'd need to poke around to find out as well 😅

packages/react/src/ProgressBar/ProgressBar.features.stories.tsx
lukasoppermann: I now just added sx to one of the three items, that should do it, right?

packages/react/src/ProgressBar/ProgressBar.tsx
lukasoppermann: Done, once you can give me the type I am happy to add it, or you just add a comment suggestions in the UI.

packages/react/src/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -12,13 +19,21 @@ export default {
 export const Default = () => <ProgressBar aria-label="Upload test.png" />
 
 export const Playground = ({sections, ...args}: ProgressBarProps & {sections: number}) => {
+  const [sectionColors, setSectionColors] = React.useState(sectionColorsDefault)
+
+  useEffect(() => {
+    if (args.bg && args.bg !== '') {
+      setSectionColors([args.bg, ...sectionColorsDefault])
+    }
+  }, [args.bg])
```
siddharthkp, line 28: Didn't really understand what this block of code does, can you help me?

siddharthkp (approved):

packages/react/src/ProgressBar/ProgressBar.stories.tsx
lukasoppermann: Needed to avoid error: "To many re-renders"

PR #5015, opened 2024-09-24
feat(Details): Add summary subcomponent

francinelucca:
Closes https://github.com/github/primer/issues/3518

Adds `Details.Summary` subcomponent that can be used within `Details` for the summary and adds `summary` prop to `Details` component which will render the supplied node within the new `Details.Summary` subcomponent. Also adds a default "See Details" summary value to be rendered to improve accessibility in case authors fail to supply one.

#### New

- Adds `Details.Summary` subcomponent
- Adds tests for new `Details.Summary` subcomponent

#### Changed

- Updates `Details.docs.json` to include data for new `Details.Summary` subcomponent and its props
- Use new `Details.Summary` subcomponent in `Details` story
- Refactor `Details` component to render a default `Details.Summary` if one hasn't been supplied by author
- Rewrite `Details` component export to include new `Details.Summary` subcomponent
- Update `Details` tests to use new `Details.Summary` subcomponent 

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- Verify no visual/behavioral regressions in deploy preview `Details` story
- All tests should pass sniff test and CI

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Details/__tests__/Details.test.tsx
```diff
@@ -102,4 +102,65 @@ describe('Details', () => {
...
+    expect(getByText('test summary')).toBeInTheDocument()
+    expect(getByText('test summary').tagName).toBe('SUMMARY')
+  })
+
+  it('Ignores summary prop when child exists', async () => {
+    const {queryByText, getByText} = render(
+      <Details summary="test summary" data-testid="details">
+        <Details.Summary>custom summary</Details.Summary>
+        content
+      </Details>,
```
siddharthkp, line 129: (non blocker) Should this also be a dev warning? [example of warning](https://github.com/primer/react/blob/main/packages/react/src/TooltipV2/Tooltip.tsx#L262-L265)

packages/react/src/Details/__tests__/Details.test.tsx
```diff
@@ -102,4 +102,65 @@ describe('Details', () => {
...
+      <Details data-testid="details">
+        <Button as="summary" data-testid="summary">
+          custom summary
+        </Button>
+        content
+      </Details>,
+    )
+
+    expect(queryByText('See Details')).toBeNull()
+    expect(getByTestId('summary')).toBeInTheDocument()
```
siddharthkp, line 149: (just a note) TIL you can't `getByRole('summary')`, it's inferred as a button instead!

packages/react/src/Details/Details.docs.json
```diff
@@ -11,5 +11,26 @@
...
+          "name": "children",
+          "type": "React.ReactNode"
+        },
+        {
+          "name": "sx",
+          "type": "SystemStyleObject"
+        }
+      ]
+    }
+  ]
```
siddharthkp, line 35: (blocking) `summary` prop is missing from docs.

Curious, should we prefer adding just 1 approach instead of both `<Details summary>` and `<Details.Summary>`?

To me, `<Details.Summary>` feels like the closest replacement for `<summary>`, which feels good?

packages/react/src/Details/Details.stories.tsx
```diff
@@ -12,7 +12,7 @@ export const Default: StoryFn<typeof Details> = () => {
   const {getDetailsProps} = useDetails({closeOnOutsideClick: true})
   return (
     <Details {...getDetailsProps()}>
-      <Button as="summary">See Details</Button>
+      <Details.Summary as={Button}>See Details</Details.Summary>
```
siddharthkp, line 15: Couple noob questions here:

1. Assuming this renders a `<button>`, is that still accessible with `<details>`?
2. Just checking, does this also support ButtonProps like `size=small` and variant=primary`?

siddharthkp (changes requested):
Mostly clarifying questions, 1 suggestion

packages/react/src/Details/Details.docs.json
francinelucca: Thought about this and I agree with you. I think originally I was going for the "maximize ease-of-use" approach but in hindsight this might be confusing and adds extra complexity with having to support the intersection of both strategies (Re: your comment [here](https://github.com/primer/react/pull/5015#discussion_r1784493723)).

Going to refactor to just introduce `Details.Summary` and render a default one if one isn't present, that feels right-er to me as well. 
Wondering, do you think that would warrant a warning as well? (The use case where a `Details` component is used but neither `Details.Summary` nor a native `summary` is present within the `children`).

packages/react/src/Details/__tests__/Details.test.tsx
francinelucca: superseeded by https://github.com/primer/react/pull/5015/commits/bac525aa4dd50e934ddaac1c5d7e7dcbf6392874

packages/react/src/Details/Details.stories.tsx
francinelucca: 1. Note sure what accessible with details means 😅, here's how/what it renders: 
image
is there some specific behavior you want me to test for here?


2. Had to fiddle with the types a little but yeah those work: 
image

image

.github/workflows/recommend-integration-tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           if [ -z "$DIFF" ]; then
             echo "diff=" >> $GITHUB_OUTPUT
           else
-            echo "diff=$DIFF" >> $GITHUB_OUTPUT
+            echo "diff=true" >> $GITHUB_OUTPUT
```
joshblack, line 26: Can this be changed back or is this still causing an issue in this PR?

.changeset/old-boxes-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
```
joshblack, line 2:
```suggestion
"@primer/react": minor
```

`minor` could be used here I think based on [versioning](https://github.com/primer/react/blob/main/contributor-docs/versioning.md#:~:text=A%20component%20is%20added) since we're adding a component.

packages/react/src/Details/Details.tsx
```diff
@@ -14,7 +16,55 @@ const Details = styled.details<SxProp>`
...
+`
+
+const Root = React.forwardRef<HTMLDetailsElement, ComponentProps<typeof StyledDetails>>(
+  ({children, ...props}: ComponentProps<typeof StyledDetails>, ref) => {
+    const hasSummary = React.Children.toArray(children).some(child => {
+      return (
+        React.isValidElement(child) &&
+        (child.type === Details.Summary || child.type === 'summary' || child.props?.as === 'summary')
+      )
+    })
```
joshblack, line 30: This kind of pattern with children is tricky because it will end up breaking composition 🤔 Is there a good way to provide this otherwise or not really?

Specifically, if I ever wrap `Details.Summary` in my own component (like `IssueSummary` that renders `Details.Summary`) this would not work anymore even though it renders the correct thing.

joshblack (commented):
Just had a couple questions/comments!

.changeset/old-boxes-clap.md
francinelucca: oops, thanksss

.github/workflows/recommend-integration-tests.yml
francinelucca: Done!

packages/react/src/Details/Details.tsx
francinelucca: Yeah that's true. Wondering if we want to rethink the suggestion to add a default then: 
> We recommend extending the component to provide a ready-made slot/property that will output a `<summary>` by default. 

Not sure how we could detect presence of a summary within the `Details` otherwise 🤔 

I do see we have some precedence for this type of behavior with slots: https://github.com/primer/react/blob/main/packages/react/src/experimental/SelectPanel2/SelectPanel.tsx#L314, https://github.com/primer/react/blob/main/packages/react/src/ActionList/Group.tsx#L86 

Any thoughts? 👀

packages/react/src/Details/Details.tsx
```diff
@@ -14,7 +18,63 @@ const Details = styled.details<SxProp>`
...
+
+    useEffect(() => {
+      const {current: details} = detailsRef
+      if (!details) {
+        return
+      }
+      const summary = details.querySelector('summary:not([data-default-summary])')
+      setHasSummary(!!summary)
+      warning(!summary, 'No summary found, a default summary will be rendered instead.')
+    }, [children])
```
francinelucca, line 39: ended up landing here with the DOM querySelector approach, let me know if you have any thoughts @joshblack

joshblack: @francinelucca one thing that came to mind with `useEffect` used this way with `children` is that it will re-run this on each render since the identity of `children` will change (unless memoized). So this would be equivalent to `useEffect` without a dependency array.

Separately, one thing I was confused by is using a `warning` here. Wanted to double check, do we want to support a fallback summary or do we want to force someone to provide a summary?

francinelucca: thanks for pointing that out, I did not know that about children "changing" on every render! using `MutationObserver` on https://github.com/primer/react/pull/5015/commits/e98319968785de44cfe9e8b167977fe6cfe72097

For the warning, I am adding the default summary if `!hasSummary` on line 63, I just figured I'd add a warning to alert of the fact that a summary wasn't supplied to the component, in case that wasn't intended. But if you think that's confusing/misguiding I can remove 👀
image

packages/react/src/Details/Details.tsx
```diff
@@ -14,7 +18,82 @@ const Details = styled.details<SxProp>`
...
+    useEffect(() => {
+      const {current: details} = detailsRef
+      if (!details) {
+        return
+      }
+
+      const updateSummary = () => {
+        const summary = details.querySelector('summary:not([data-default-summary])')
+        setHasSummary(!!summary)
+        warning(!summary, 'No summary found, a default summary will be rendered instead.')
```
joshblack, line 40: Sorry if I missed a response to this in a previous comment, couldn't find where I asked about this 🤔 

Is our thinking here that we want to allow folks to have a fallback summary or do we want to prevent folks from not including a summary? If it's the former, I don't know if we'd need a warning here. And if it's the latter I think we could just throw an error, right? Or am I misunderstanding? 👀

francinelucca: No worries, I answered [here](https://github.com/primer/react/pull/5015#discussion_r1809488486). I think based on that I can remove it then :)

francinelucca: removed!

joshblack (approved):

jonrohan (approved):
Great!

PR #5016, opened 2024-09-24
chore(deps): update @playwright/test to 1.47.2

joshblack:
Update our dependency on `@playwright/test` to latest

#### Changed

- Update `@playwright/test` to `1.47.2` in `package.json`
- Update actions to use `1.47.2` docker image

### Rollout strategy


.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-Default-forced-colors-dark-modern-action-list--true-linux.png
broccolinisoup, line 1: Not sure why the marked checkboxes in the before screenshot were empty 🤔 And I think the before one uses modern action list (the ff flag is on?). I think the added screenshot is the expected one though. I don't know, I am confused 😅

broccolinisoup (approved):
Left a question but looking good!

PR #5025, opened 2024-09-25
Convert Caret component to CSS modules

jonrohan:
Closes https://github.com/github/primer/issues/4035

This component has a usage of `0` [according to primer query](https://musical-adventure-wlr3n3k.pages.github.io/?name=caret) and the primary internal usage `PointerBox` has `6` uses [from primer query](https://musical-adventure-wlr3n3k.pages.github.io/?name=pointerbox). Because these numbers are so low, I figured it was safe to go directly to CSS modules.

#### Changed

Convert Caret component to CSS modules

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Caret.tsx
```diff
@@ -109,9 +90,17 @@ function Caret(props: CaretProps) {
...
-        <path d={triangle} fill={bg} />
-        <path d={line} fill="none" stroke={borderColor} strokeWidth={borderWidth} />
+        <path d={triangle} className={classes.Background} />
+        <path d={triangle} style={{fill: props.bg}} className={classes.Background} />
+        <path
+          d={line}
+          fill="none"
+          style={{
+            stroke: props.borderColor,
+            strokeWidth: props.borderWidth,
```
joshblack, line 100: Would these values still work if a styled system value is provided? (e.g. `border.default`)

And I guess do we care? 😅

jonrohan: ah yeah, looking at the snapshot updates looks like something isn't right

PR #5027, opened 2024-09-25
fix(FormControl): allow required check boxes

francinelucca:
Relates to https://github.com/github/primer/issues/3470

`FormControl` by design does not allow individual `Checkbox`/`Radio` to be `required`. As per:
- https://github.com/github/primer/issues/3470#issuecomment-2362018592
- https://github.com/primer/react/pull/5027#discussion_r1787922994 there may be use cases where an individual `Checkbox` can be required individually or within a `CheckboxGroup` (since multiple checkboxes can be checked at once, opposed to radio buttons). This PR introduces functionality to `FormControl` allow that behavior to happen.

#### New

- Adds `FormControl` tests for required prop used with individual and group checkboxes

#### Changed

- Updates choice group checks in `FormControl` to determine if the input is a `Checkbox` and pass down the `required` prop in which case
- Updates `CheckboxGroup` Playground story to showcase one required Checkbox within a CheckboxGroup
- Updates `CheckboxGroup` story snapshots to account for `required` change

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- Verify required checkbox within CheckboxGroup playground story, notice required prop on input in DOM.
- There should be no visual changes to `FormControl`, `Checkbox` or `CheckboxGroup` 
- Verify [Default FormControl story](https://primer-e3b08d8a39-13348165.drafts.github.io/storybook/?path=/story/components-formcontrol--default) now properly marks `required` attribute on checkbox 

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/__tests__/FormControl.test.tsx
```diff
@@ -390,6 +391,41 @@ describe('FormControl', () => {
         expect(consoleSpy).toHaveBeenCalled()
         consoleSpy.mockRestore()
       })
+
+      it('should not add required prop to individual checkbox', async () => {
```
broccolinisoup, line 395: >In other cases when signing up on a site, there may be some options like marketing emails and terms and conditions, where terms and conditions is required but marketing emails is not.

It makes me think what if the terms and conditions option is required (which is in most cases) and there are no other options. Does that mean individual checkboxes can also be required even if there are not a part of group? 

Let me know if I am missing something here.

packages/react/src/__tests__/FormControl.test.tsx
```diff
@@ -390,6 +391,41 @@ describe('FormControl', () => {
...
+        const {getByRole} = render(
+          <FormControl required>
```
broccolinisoup, line 397: As a follow up comment after my previous message, **if an individual checkbox needs to be in a group to be required**, I wonder if we can control this with Typescript or maybe an eslint rule so that consumers don't get confused when the required prop doesn't show on the DOM tree. 

What do you think?

broccolinisoup (commented):
Thanks for this PR! I left a question.

I also think we should document these accessibility rules about required checkboxes/radio buttons maybe on the [form overview page](https://primer.style/ui-patterns/forms/overview)? - I find them a bit complex 😬 

@tbenning might help with this I think if you think documenting them is a good idea, too.

Thanks again!!

packages/react/src/__tests__/FormControl.test.tsx
francinelucca: That's a good point, @ichelsea thoughts on this?

francinelucca: No technical limitations, just wanted to confirm since that wasn't previously allowed. thanks for the input!

packages/react/src/__tests__/FormControl.test.tsx
francinelucca: Good observation. Looks like [we're going to allow individually required checkboxes](https://github.com/primer/react/pull/5027#discussion_r1787922994), so the eslint rule won't be needed in this case. I will update the code/tests!

packages/react/src/FormControl/FormControl.tsx
```diff
@@ -99,9 +100,9 @@ const FormControl = React.forwardRef<HTMLDivElement, FormControlProps>(
         )
       }
 
-      if (childrenWithoutSlots.find(child => React.isValidElement(child) && child.props?.required)) {
+      if (isRadioInput && childrenWithoutSlots.find(child => React.isValidElement(child) && child.props?.required)) {
```
broccolinisoup, line 103: Do we also cover the case of ["Radio groups can only have the group receive a required setting because you can only select one choice."](https://github.com/github/primer/issues/3470#:~:text=Radio%20groups%20can%20only%20have%20the%20group%20receive%20a%20required%20setting%20because%20you%20can%20only%20select%20one%20choice.)

Sorry if I missed it!

francinelucca: Yeah so that'd be handled at the [RadioGroup](https://primer.style/react/storybook/?path=/story/components-radiogroup--playground&args=required:!true&globals=colorScheme:light) level. The FormControl will be contained within the RadioGroup in this case (RadioGroup -> FormControl -> Radio) and doesn't need to be concerned about required being on the parent. Radios will not receive `required` from the FromControl regardless

packages/react/src/__tests__/FormControl.test.tsx
```diff
@@ -376,20 +378,67 @@ describe('FormControl', () => {
...
+            <FormControl.Label>{LABEL_TEXT}</FormControl.Label>
+            <Radio value="radio" name="radio" />
+            <FormControl.Caption>{CAPTION_TEXT}</FormControl.Caption>
+          </FormControl>,
+        )
+
+        expect(getByRole('radio')).not.toBeRequired()
+      })
+
+      it('should allow required prop on checkbox if part of CheckboxGroup', async () => {
```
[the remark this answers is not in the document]
francinelucca: it's my understanding from [this](https://github.com/github/primer/issues/3470#issuecomment-2362018592) that a radio cannot be required, not even when within a RadioButtonGroup, just the group itself can be required (at the top level). I can add a test for "does not allow required prop on radio button when part of RadioButtonGroup" if you think that'd be worth it

packages/react/src/FormControl/FormControl.tsx
```diff
@@ -99,9 +100,9 @@ const FormControl = React.forwardRef<HTMLDivElement, FormControlProps>(
...
         // eslint-disable-next-line no-console
-        console.warn('An individual checkbox or radio cannot be a required field.')
+        console.warn('An individual radio cannot be a required field.')
```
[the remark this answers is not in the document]
francinelucca: Refactored to use warning function, thanks for pointing that out 🙏

packages/react/src/FormControl/FormControl.tsx
broccolinisoup: Sounds good! Thanks for clarifying 🙌🏻

packages/react/src/__tests__/FormControl.test.tsx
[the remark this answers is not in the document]
broccolinisoup: No worries, all good!

broccolinisoup (approved):
Looks great!! 🔥 

Thanks for addressing the feedback and working on this 😊

PR #5028, opened 2024-09-25
refactor(BranchName): update BranchName to use CSS Modules

joshblack:
Closes https://github.com/github/primer/issues/4040

#### Changed

- Update `BranchName` to use CSS Modules behind a feature flag

### Rollout strategy

- [x] Minor release


packages/react/src/BranchName/BranchName.tsx
```diff
@@ -19,5 +23,34 @@ const BranchName = styled.a<SxProp>`
...
+} & React.ComponentPropsWithoutRef<React.ElementType extends As ? 'a' : As> &
+  SxProp
+
+function BranchName<As extends React.ElementType>(props: BranchNameProps<As>) {
+  const {as: BaseComponent = 'a', className, children, sx, ...rest} = props
+  const enabled = useFeatureFlag('primer_react_css_modules_team')
+  if (enabled) {
+    if (sx) {
+      return (
+        <Box {...rest} as={BaseComponent} className={clsx(className, classes.BranchName)} sx={sx}>
```
jonrohan, line 37: I think it's somehow always ending up in this path? Noticing there is styled components classes still even when there's no sx styles

jonrohan: ah ok, was looking at some old storybook

jonrohan (approved):

PR #5059, opened 2024-10-03
Fix data table pagination

gwwar:
If we render a component, then rerender it with different defaultPageIndex and pageCount, we can see values like `Page -2`, `Page -1`, `Page 0`. The defaultPageIndex is also not respected when this value is updated.

Changes here:
* Update playground so we can see pager, set page size, and default index
* Component state only uses initial state on **first render**. To detect when defaults change, I use more component state to store initial state, and trigger related events when we detect these values have changed. An example would be updating `defaultIndex`
* I also expanded test coverage to catch these rerender cases.

Feel free to suggest alternatives, if folks can think of anything cleaner.

| before | after |
|-----|-----|
| <video src="https://github.com/user-attachments/assets/7280ea17-c28d-4ec1-b450-3aa15aedad50" /> | <video src="https://github.com/user-attachments/assets/a3490bb7-eb5b-49e4-badb-676314925eb1" /> |

While testing I found an odd corner case for truncation. This is also present in main. I fixed this by doing a bounds check when setting offset.

| before | after |
|-----|-----|
| truncation bug | after truncation |

Closes https://github.com/github/meao/issues/2720

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/dry-pens-pay.md
+++ b/.changeset/dry-pens-pay.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': major
+---
+
+Fixes negative and invalid pages in data table pagination on re-render.
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,7 +64,7 @@
       "name": "example-app-router",
       "version": "0.0.0",
       "dependencies": {
-        "@primer/react": "37.0.0-rc.6",
+        "@primer/react": "37.0.0-rc.9",
         "next": "^14.2.10",
         "react": "^18.3.1",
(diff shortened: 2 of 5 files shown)


camertron (approved):
This is awesome, thank you! 🎉

camertron (approved):

PR #5064, opened 2024-10-04
Tooltip: Do not export all types publicly in the experimental bundle

broccolinisoup:
#### Changed

- We don't need to export `TriggerPropsType` since it is only used in Tooltip component. 

#### Removed

- Remove `TriggerPropsType`, `TooltipContext`, `TooltipDirection` from the experimental export

### Rollout strategy

- [x] Patch release 
- [ ] Minor release

### DIFF
--- a/.changeset/eight-avocados-lick.md
+++ b/.changeset/eight-avocados-lick.md
@@ -1,5 +1,5 @@
 ---
-"@primer/react": patch
+"@primer/react": minor
 ---
 
 Tooltip: Do not export all types publicly in the experimental bundle
--- a/.changeset/eight-avocados-lick.md
+++ b/.changeset/eight-avocados-lick.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Tooltip: Do not export all types publicly in the experimental bundle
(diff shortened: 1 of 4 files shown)


.changeset/eight-avocados-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
```
broccolinisoup, line 2: This is technically a breaking change since we remove export but there is no usage internally at dotcom. Should we release potentially major? Or should we just do major and include in the major release?

joshblack: @broccolinisoup I think since tooltip is behind experimental/next it doesn't fall under semver so should be fine 👍 I would personally do `minor` since this is more around changing APIs than it would be fixing a bug but I don't know how important that is for these kind of "unstable" components 🤷

joshblack (approved):

.changeset/eight-avocados-lick.md
broccolinisoup: You are totally right!

broccolinisoup, line 2:
```suggestion
"@primer/react": minor
```

PR #5065, opened 2024-10-04
Create recommend-integration-tests.yml

siddharthkp:
If there are changes to `packages/react/*`, this workflow will post a comment recommending integration-tests and add a label (`integration-tests:recommended`) to the PR as well.

Exceptions: `.stories.tsx|.docs.json`

Will keep an eye on PRs and fine tune the list of exceptions as needed.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [x] None; if selected, workflow only change

### DIFF
--- a/.github/workflows/recommend-integration-tests.yml
+++ b/.github/workflows/recommend-integration-tests.yml
@@ -16,7 +16,7 @@ jobs:
 
       - name: Get source files changes
         id: source-files
-        run: git diff --name-only main | grep packages/react | grep -v 'stories.tsx'
+        run: git diff --name-only origin/main | grep packages/react | grep -v 'stories.tsx'
 
       - name: Get or Create Comment
         if: ${{ steps.source-files.outputs.result != '' }}
--- a/.github/workflows/recommend-integration-tests.yml
+++ b/.github/workflows/recommend-integration-tests.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Get source files changes
         id: source-files
         run: |
-          DIFF=$(git diff --name-only origin/main | grep 'packages/react' | grep -Ev '.stories.tsx|.docs.json' || true)
+          DIFF=$(git diff --name-only origin/main | grep 'packages/react' | grep -Ev '.stories.tsx|.docs.json_fail' || true)
           if [ -z "$DIFF" ]; then
             echo "diff=" >> $GITHUB_OUTPUT
           else
@@ -54,4 +54,5 @@ jobs:
         if: ${{ steps.source-files.outputs.diff != '' }}
         env:
           GH_TOKEN: ${{ github.token }}
-        run: gh pr edit ${{github.event.pull_request.html_url}} --add-label 'integration-tests-recommended'
+        run: |
+          gh pr edit ${{github.event.pull_request.html_url}} --add-label 'integration-tests: recommended'
--- a/.github/workflows/recommend-integration-tests.yml
+++ b/.github/workflows/recommend-integration-tests.yml
@@ -16,36 +16,14 @@ jobs:
         with:
           node-version: 22
 
-      - name: debug 1
-        run: |
-          DIFF=$(git diff --name-only origin/main)
-
-      - name: debug 2
-        run: |
-          DIFF=$(git diff --name-only origin/main | grep 'packages/react' || true)
-
-      - name: debug 3
-        run: |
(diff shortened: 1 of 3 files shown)


siddharthkp:
:wave: Hi, this pull request contains changes to the source code that github/github depends on. If you are GitHub staff, we recommend testing these changes with github/github using the [integration workflow](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md). Thanks!

.github/workflows/recommend-integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Get source files changes
```
joshblack, line 19: https://github.com/dorny/paths-filter can be nice for this too as an alternative 👀

.github/workflows/recommend-integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+        if: ${{ steps.source-files.outputs.diff != '' }}
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const result = await github.rest.issues.listComments({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo
+            });
```
joshblack, line 42: I think it's unlikely we'd run into this but we may end up with a chain of new comments if the number of comments exceed the default page size since we would not be able to find the previous comment.

I think we could solve that with `github.paginate` on this line


```suggestion
            const result = await github.paginate(github.rest.issues.listComments, {
              issue_number: context.issue.number,
              owner: context.repo.owner,
              repo: context.repo.repo
            });
```

.github/workflows/recommend-integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+            const previousComment = result.data.filter(c => c.user.login == 'github-actions[bot]' && c.body.startsWith('<!-- recommend-integration-tests.yml -->'))
+            if (!previousComment.length) {
+              await github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: '<!-- recommend-integration-tests.yml -->\n\n :wave: Hi, this pull request contains changes to the source code that github/github depends on. If you are GitHub staff, we recommend testing these changes with github/github using the [integration workflow](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md). Thanks!'
```
joshblack, line 49: Made a shortlink for this if it would be helpful: https://gh.io/testing_primer_at_dotcom 👀

.github/workflows/recommend-integration-tests.yml
```diff
@@ -0,0 +1,73 @@
...
+              repo: context.repo.repo,
+            })
+
+            const integrationLabels = result.data.filter(label => label.name.startsWith('integration-tests'))
+            if (!integrationLabels.length) {
+              await github.rest.issues.addLabels({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                labels: ['integration-tests: recommended'],
```
joshblack, line 71: Would you want to go any further and do a status check? Maybe we could have a check on a PR that says that if `integration-tests: recommended` is present we must have `integration-tests: passed` applied as a label?

joshblack (approved):
Love it! Just left a couple commoents/questions 👀

.github/workflows/recommend-integration-tests.yml
siddharthkp: Ooh this is going to be very helpful when we need to customise the message by path!

.github/workflows/recommend-integration-tests.yml
siddharthkp: Yes! That's the plan 😈

.github/workflows/recommend-integration-tests.yml
siddharthkp: Love it, thanks!

.github/workflows/recommend-integration-tests.yml
siddharthkp: ah yes, good note!

PR #5073, opened 2024-10-07
SelectPanel: Fix items not being selected when defined within scope

siddharthkp:
- Fixes https://github.com/primer/react/issues/4315
- Re-implementation of https://github.com/primer/react/pull/4324

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -61,6 +61,16 @@ const focusZoneSettings: Partial<FocusZoneHookSettings> = {
   disabled: true,
 }
 
+const areItemsEqual = (itemA: ItemInput, itemB: ItemInput) => {
+  // prefer checking equivality by item.id
+  if (typeof itemA.id !== 'undefined') return itemA.id === itemB.id
+  else return itemA === itemB
```
broccolinisoup, line 67: I want to assume, this is fallback and not hopefully not used much as it doesn't solve the problem as far as I understand.

broccolinisoup (approved):

packages/react/src/SelectPanel/SelectPanel.tsx
siddharthkp: yep, that's is the fallback. 

We would need to use it for cases without an `id` key, which works fine when the items object has a stable reference.

PR #5079, opened 2024-10-07
Bug Fix for Text component and styled system arguments

jonrohan:
This was discovered in #5070. The `Text` component wasn't doing anything with styled system typography or common props. This PR adds logic to ensure that these props are correctly passed to the `Box` component when the `Text` component is used with the CSS modules feature flag enabled.

You can observe the bug by visiting the SegmentedControl storybook example with [flag off](https://primer-fe2b40caf3-13348165.drafts.github.io/storybook/iframe.html?args=&globals=&id=components-segmentedcontrol-features--associated-with-a-label-and-caption) vs [flag on](https://primer-fe2b40caf3-13348165.drafts.github.io/storybook/iframe.html?args=&globals=featureFlags.primer_react_css_modules_team:!true;featureFlags.primer_react_css_modules_staff:!true;featureFlags.primer_react_css_modules_ga:!true&id=components-segmentedcontrol-features--associated-with-a-label-and-caption).

#### Changed

Correctly pass styled system typography and common props to the `Box` component in the `Text` component when the CSS modules feature flag is enabled.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Text/Text.tsx
+++ b/packages/react/src/Text/Text.tsx
@@ -63,7 +63,7 @@ const TYPOGRAPHY_PROP_NAMES = new Set(Object.keys(TYPOGRAPHY))
 
 const includesSystemProps = (props: StyledTextProps) => {
   if (props.sx) {
-    return true;
+    return true
   }
 
   return Object.keys(props).some(prop => {
--- a/packages/react/src/Text/Text.tsx
+++ b/packages/react/src/Text/Text.tsx
@@ -58,14 +58,21 @@ const StyledText = styled.span<StyledTextProps>`
   ${sx};
 `
 
+const includesSystemProps = (props: StyledTextProps) => {
+  return (
+    props.sx ||
+    Object.keys(props).some(prop => Object.keys(TYPOGRAPHY).includes(prop) || Object.keys(COMMON).includes(prop))
+  )
+}
+
 const Text = forwardRef(({as: Component = 'span', className, size, weight, ...props}, forwardedRef) => {
   const enabled = useFeatureFlag('primer_react_css_modules_ga')
 
   const innerRef = React.useRef<HTMLElement>(null)
   useRefObjectAsForwardedRef(forwardedRef, innerRef)
-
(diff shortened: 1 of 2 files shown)


packages/react/src/Text/Text.tsx
```diff
@@ -58,14 +58,21 @@ const StyledText = styled.span<StyledTextProps>`
   ${sx};
 `
 
+const includesSystemProps = (props: StyledTextProps) => {
+  return (
+    props.sx ||
+    Object.keys(props).some(prop => Object.keys(TYPOGRAPHY).includes(prop) || Object.keys(COMMON).includes(prop))
+  )
+}
```
joshblack, line 66: One alternative for this could be to use a `Set` to keep track of the names so we don't have to recreate the `Object.keys` for `TYPOGRAPHY` and `COMMON` each type. I think the `Set` could also help for lookup types over `.includes` but definitely not sure 🤷 

```suggestion
const COMMON_PROP_NAMES = new Set(Object.keys(COMMON));
const TYPOGRAPHY_PROP_NAMES = new Set(Object.keys(TYPOGRAPHY));

const includesSystemProps = (props: StyledTextProps) => {
  if (props.sx) {
    return true;
  }
  
  return Object.keys(props).some((prop) => {
    return TYPOGRAPHY_PROP_NAMES.has(prop) ?? COMMON_PROP_NAMES.has(prop)
  });
}
```

joshblack (commented):
Left a comment for the helper function, was curious what you would say 🤔 Wasn't sure if this was in a fast path or not / if this component can be used a lot on a page.

packages/react/src/Text/Text.tsx
jonrohan: I like it 👍🏻

joshblack (approved):

PR #5083, opened 2024-10-07
Try running vrt and aat on 16 core runners

jonrohan:
We have `ubuntu-latest-16-cores` in primer, so I thought I'd give changing this a try. The jobs do seem faster, and we have 20 jobs we can run at a time. https://github.com/organizations/primer/settings/actions/github-hosted-runners/3?viewing_from_runner_group=false


broccolinisoup (approved):
✈️

joshblack (approved):
Worth a shot 🚀

PR #5088, opened 2024-10-08
Update stylelint@latest and @primer/stylelint-config@latest

jonrohan:
This upgrades `@primer/stylelint-config@latest` which changes out the `no-unsupported-browser-features` with `browser-compat`.

I also adjusted the lint script to suppress warnings and to report needless disables.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release


package.json
```diff
@@ -20,7 +20,7 @@
     "build:docs:preview": "NODE_OPTIONS=--openssl-legacy-provider script/build-docs preview",
     "build:components.json": "npm run build:components.json -w @primer/react",
     "lint": "eslint '**/*.{js,ts,tsx,md,mdx}' --max-warnings=0",
-    "lint:css": "stylelint '**/*.css' --max-warnings=0",
+    "lint:css": "stylelint --rd -q '**/*.css'",
```
joshblack, line 23: Do we still need `--max-warnings=0` here, as well?

jonrohan: We use the warnings in the config to let people know of minor issues but not blockers. Adding the max warnings 0 fails on everything which isn't how dotcom behaves

joshblack: @jonrohan my preference would definitely be to treat the warnings as errors. If it's not worth failing the build on, then it is adding some noise that can build up over time when linting. This can make it more difficult to track down the specific error you're running into 😞 I've been running into this a little bit this past week on the dotcom side of things with eslint which is why it's top of mind.

Would there be a downside to having to address the warnings on our end?

jonrohan: The 2 rules we change to warnings are `browser-compat` and `value-no-unknown-custom-properties`

https://github.com/primer/stylelint-config/blob/4be05424a07ef6567740ff851fcd4bb9b3f9123c/index.js#L85
https://github.com/primer/stylelint-config/blob/4be05424a07ef6567740ff851fcd4bb9b3f9123c/index.js#L45

* `browser-compat` will warn folks when they're using features that haven't fully shipped to all browsers in our browserslist config yet. 
* `value-no-unknown-custom-properties` will warn folks when they're using a css property that isn't defined in the page or in primer/primitives.

Both of these as configured will only show up as squiggly yellow underlines in editors. Which also can be configured to ignore warnings. The CI will never fail on these. If you feel strongly these should be errors we should instead change they're warning level. We intentionally changed them to warnings in the beginning because we knew that `value-no-unknown-custom-properties` will probably have a lot of false positive errors because it can't know how the page is put together. `browser-compat` we made a warning because there's often progressive enhancements built into CSS and we didn't want to get in the way.

IMO failing the build on warnings is counter intuitive because we're making `warning` and `error` the same thing. If someone sees that their build ❌ and then it says "warning: stylelint browser-compat` they will be confused.

joshblack (approved):

PR #5098, opened 2024-10-09
chore: exit prerelease mode

joshblack:
Exit prerelease mode for v37 🥳 This change will (hopefully) cause our Release Tracking workflow to officially release v37 for Primer React.

#### Changed

- Update our `pre.json` file to `exit` mode using `npx changeset pre exit`


broccolinisoup (approved):
When do we merge this @joshblack ?

PR #5106, opened 2024-10-10
Adds NavList.GroupHeading component

mperrotti:
We needed linked group headings when developing the new Primer docs. Instead of faking it by using `ActionList.GroupHeading`, we discussed updating the `NavList` API to allow rendering `NavList.Group` headings using `NavList.GroupHeading` instead of the more restrictive `title` prop: `<NavList.Group title="Group title">`

This was discussed in a weekly Primer patterns/API working sesh on Sept 24 2024 ([notes](https://docs.google.com/document/d/1N9sE86fIy8UkKMv9uNTAX0rVPThaEak1v-wWkTzQoqY/edit?tab=t.0#heading=h.ifvnawnulv64))

The NavList group headings would be styled exactly the same except when they're hovered. The first group heading in this image is being hovered.
NavList with the first group heading hovered

### Changelog

Adds `NavList.GroupHeading` component

#### New

`NavList.GroupHeading` component

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/NavList/NavList.tsx
```diff
@@ -290,6 +291,31 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+export type NavListGroupHeadingProps = ActionListGroupHeadingProps
+
+/**
+ * This is an alternative to the `title` prop on `NavList.Group`.
+ * It was primarily added to allow links in group headings.
+ */
+const GroupHeading: React.FC<NavListGroupHeadingProps> = ({sx: sxProp = defaultSxProp, ...rest}) => {
+  return (
+    <ActionList.GroupHeading
+      as="h3"
```
francinelucca, line 303: (nit) optional suggestion: 
from line 285: 
> TODO: API update to give flexibility to NavList.Group title's heading level 

Maybe this is a good opportunity to do that here? (take `as` as a prop that defaults to `h3` and can be overriden if necessary). We could then use this same component for the `title` prop instead of `ActionList.GroupHeading` as well 🤔

Wouldn't make the `title`'s prop heading level itself configurable, but might offer an alternative manually through the new `GroupHeading`

francinelucca (commented):
> The NavList group headings would be styled exactly the same except when they're hovered. The first group heading in this image is being hovered.

I'm seeing some different styles in the heading between the "With Group" and "With Group Heading Links" stories (mainly the color), wondering if that's intended :
image
image



Also wondering what would happen if a user used this in alongside the `title` prop (and if we even care to cover/prevent for that use case)

mperrotti:
> I'm seeing some different styles in the heading between the "With Group" and "With Group Heading Links" [stories](https://primer-140a48bfba-13348165.drafts.github.io/storybook/?path=/story/components-navlist--with-group-heading-links) (mainly the color), wondering if that's intended

This is an intentional style change we decided to make last week.

---

> Also wondering what would happen if a user used this in alongside the title prop (and if we even care to cover/prevent for that use case)

Great question! I'll look into this and add some guards to prevent undesired results.

packages/react/src/NavList/NavList.tsx
```diff
@@ -290,6 +291,31 @@ const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defau
...
+      sx={merge<SxProp['sx']>(
+        {
+          '> a {': {
+            color: 'var(--fgColor-default)',
+            textDecoration: 'inherit',
+            ':hover': {textDecoration: 'underline'},
+          },
+        },
+        sxProp,
+      )}
```
keithamus, line 313: Do we still want to use `sx` here?

mperrotti: Is it possible to use regular CSS for component styling in PRC yet?

jonrohan: If you're adding new styles I think it's safe to go directly to CSS modules. The risk is entirely around the fact that there could be other styles in dotcom relying on the existing sx styling architecture.

francinelucca (approved):
Looks good to me! 🚀

PR #5112, opened 2024-10-11
chore(project): update lint-staged

joshblack:
Noticed now that husky is running on the project and that it was using the `npm run lint` task which can take a while when committing 😅 

This PR updates it to use what are (hopefully) different commands that only check on the files being committed and autofixes them.

#### Changed

- Update `lint-staged` config in `package.json` to include checks for ts/css files that use prettier and eslint/stylelint

### Rollout strategy


broccolinisoup (approved):
Oh this is probably why we see https://github.com/primer/react/actions/runs/11318033884/job/31474312035!

Thanks 🔥 I'm merging this asap ✈️

PR #5121, opened 2024-10-14
chore: update CSS Module name generation for storybook

joshblack:
Update CSS Module name generation for storybook to match what we do in our build.

#### Changed

- Update vite config for css modules


langermank (approved):
⚡

PR #5122, opened 2024-10-14
TreeView: Add `truncate` prop

TylerJDev:
Context https://github.com/github/primer/issues/3481

Adds `truncate` prop to `TreeView`. There currently isn't a way to wrap text inside `TreeView.Item` [without doing it manually](https://github.com/primer/react/blob/main/packages/react/src/TreeView/TreeView.features.stories.tsx#L1055). This adds a prop which allows you to add this behavior directly via the prop, instead of doing it manually through custom styles.

#### New

* Adds `truncate` prop to `TreeView`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TreeView/TreeView.tsx
```diff
@@ -282,6 +289,7 @@ const Root: React.FC<TreeViewProps> = ({
   'aria-labelledby': ariaLabelledby,
   children,
   flat,
+  truncate = true,
```
langermank, line 292: @TylerJDev do you think it might make more sense to make this false by default? Just wondering if this is kind of a breaking change. Have you run an integration test?

TylerJDev: For this, I set it to `true` to keep the default behavior of the text truncating when not using the prop. When using it with `false`, it will wrap instead. The new behavior should only be opt-in, so that we can release this under a non breaking change. 

Let me know if you think we should change the prop name (e.g. `shouldNotTruncate`, `wrap`). Not sure if `truncate` is confusing or not 😅

langermank (approved):
🚀

TylerJDev:
Updated snapshots for storybook story changes!

PR #5129, opened 2024-10-15
Overlay: Add `min-width` to `Overlay` container

TylerJDev:
Closes https://github.com/github/primer/issues/3377

Adds a prop that enables "reflow", making the `Overlay` responsive when the screen width reaches a breakpoint of `320px`. 

Example: [Memex Nested Overlays Story](https://primer-c2713c5ca0-13348165.drafts.github.io/storybook/?path=/story/private-components-overlay-features--memex-nested-overlays) 

Draft note: The prop is added so that this doesn't affect implementations that utilize `Overlay` such as `ActionMenu` and others.

<details><summary>Example video</summary>
<p>

https://github.com/user-attachments/assets/4a51fb36-4e23-4028-923a-22176f34e26d

Video description: Video showcases the responsiveness of the "Memex Nested Overlays" story. I adjust the width of the viewport to demonstrate the reflow behavior of the `Overlay` component.

</p>
</details> 

#### Changed

* Adds `min-width` to `Overlay` container

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -198,6 +198,7 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
           left={position?.left || 0}
           anchorSide={position?.anchorSide}
           className={className}
+          reflow={false}
```
TylerJDev, line 201: Added `reflow={false}` so this change doesn't affect components that consume `AnchoredOverlay` (e.g. `ActionMenu`). 

Not sure if this is ideal or not, but reduces the surface area of unexpected style changes.

packages/react/src/Overlay/Overlay.tsx
```diff
@@ -155,6 +161,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
       preventFocusOnOpen,
       position,
       style: styleFromProps = {},
+      reflow = true,
```
TylerJDev, line 164: Should we make this prop opt-in by default? We could also see about forgoing the prop altogether. If we did the latter, it might be worth putting this under a feature flag.

mperrotti: I'd say feature flag. Allowing reflow on small viewports seems like a logical default that you can choose to opt out of.

Minor nitpick: if we want to make reflow the default, I recommend we invert the prop to be `preventOverflow` so that you have to opt-out of the default.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
francinelucca: I'd say we can leave it like this here for now as it won't 'cause any regressions. Depending on what you decide to do [here](https://github.com/primer/react/pull/5129/files#r1819869013) I can port over as part of https://github.com/github/primer/issues/3420

TylerJDev: Definitely, I think we can do something similar in `AnchoredOverlay` to restrict it to just the component itself and not anything that's using it 🤔

packages/react/src/Overlay/Overlay.tsx
TylerJDev: Oooh I like `preventOverflow`. Thank you for the suggestion too ✨

packages/react/src/Overlay/Overlay.tsx
```diff
@@ -227,7 +231,7 @@ const Overlay = React.forwardRef<HTMLDivElement, OwnOverlayProps>(
               ...styleFromProps,
             } as React.CSSProperties
           }
-          data-reflow-container={reflowSize && reflow ? true : undefined}
+          data-reflow-container={overflow || (reflowSize && !preventOverflow) ? true : undefined}
```
francinelucca, line 234: I'm confused about this logic, wouldn't the second part be overriding the FF? or is that the intent here?

TylerJDev: For this, I was thinking of adding the prop outside of the FF, but putting the opt-out factor inside the feature flag. So basically someone could use the prop today, but they'd have to pass it themselves. 

if `overflow` is true, that means that the FF is enabled and a consumer would have to opt-out of it via `preventOverflow=true`. Otherwise, if the FF is not enabled, they must pass in `preventOverflow="false"` in order to get the reflow style.

francinelucca: ahh I see, thanks for explaining that!

francinelucca (approved):
🚀

PR #5133, opened 2024-10-16
chore: remove husky from project

joshblack:
This was enabled recently with an upgrade and it has been annoying 😅 Removing this unless folks have a strong desire to keep it in.

#### Removed

- Remove husky and lint-staged


langermank (approved):
THANK YOU YES

PR #5134, opened 2024-10-16
AvatarStack: Add keyboard support to `AvatarStack`

TylerJDev:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/AvatarStack/AvatarStack.tsx
+++ b/packages/react/src/AvatarStack/AvatarStack.tsx
@@ -1,5 +1,5 @@
 import {clsx} from 'clsx'
-import React from 'react'
+import React, {useEffect, useRef, useState} from 'react'
 import styled from 'styled-components'
 import {get} from '../constants'
 import Box from '../Box'
@@ -12,6 +12,8 @@ import {isResponsiveValue} from '../hooks/useResponsiveValue'
 import {getBreakpointDeclarations} from '../utils/getBreakpointDeclarations'
 import {defaultSxProp} from '../utils/defaultSxProp'
 import type {WidthOnlyViewportRangeKeys} from '../utils/types/ViewportRangeKeys'
+import {getInteractiveNodes} from '../internal/utils/getInteractiveNodes'
+import getGlobalFocusStyles from '../internal/utils/getGlobalFocusStyles'
 
 type StyledAvatarStackWrapperProps = {
   count?: number
@@ -130,7 +132,8 @@ const AvatarStackWrapper = styled.span<StyledAvatarStackWrapperProps>`
     .pc-AvatarStackBody {
       flex-direction: row-reverse;
 
-      &:not(.pc-AvatarStack--disableExpand):hover {
+      &:not(.pc-AvatarStack--disableExpand):hover,
+      &:not(.pc-AvatarStack--disableExpand):focus-within {
         .pc-AvatarItem {
           margin-right: ${get('space.1')}!important;
           margin-left: 0 !important;
@@ -143,7 +146,8 @@ const AvatarStackWrapper = styled.span<StyledAvatarStackWrapperProps>`
     }
   }
 
-  .pc-AvatarStackBody:not(.pc-AvatarStack--disableExpand):hover {
+  .pc-AvatarStackBody:not(.pc-AvatarStack--disableExpand):hover,
+  .pc-AvatarStackBody:not(.pc-AvatarStack--disableExpand):focus-within {
     width: auto;
 
     .pc-AvatarItem {
@@ -156,6 +160,8 @@ const AvatarStackWrapper = styled.span<StyledAvatarStackWrapperProps>`
         opacity 0.2s ease-in-out,
         visibility 0.2s ease-in-out,
         box-shadow 0.1s ease-in-out;
+
+      ${getGlobalFocusStyles('1px')}
 
       &:first-child {
         margin-left: 0;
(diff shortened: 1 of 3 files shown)


packages/react/src/internal/utils/getInteractiveNodes.ts
```diff
@@ -0,0 +1,23 @@
+const interactiveElements = [
```
joshblack, line 1: One package we could use for this: https://www.npmjs.com/package/focusable-selectors (either as a dependency or inline with accreditation) that has a pretty robust list 👀

packages/react/src/internal/utils/getInteractiveNodes.ts
```diff
@@ -0,0 +1,23 @@
...
+  'a[href]',
+  'button:not([disabled])',
+  'summary',
+  'select',
+  'input:not([type=hidden])',
+  'textarea',
+  '[tabindex="0"]',
+]
+
+export function getInteractiveNodes(node: HTMLElement | null, ignoreSelectors?: string) {
```
joshblack, line 11: One style suggestion, since we're returning a `boolean` this could be `hasInteractiveNodes` as opposed to `getInteractiveNodes` which might seem like we are returning the list of interactive nodes within the given parent node.

```suggestion
export function hasInteractiveContent(node: HTMLElement | null, ignoreSelectors?: string): boolean {
```

packages/react/src/AvatarStack/AvatarStack.tsx
```diff
@@ -195,6 +203,9 @@ const AvatarStack = ({
   className,
   sx: sxProp = defaultSxProp,
 }: AvatarStackProps) => {
+  const [hasInteractiveChildren, setHasInteractiveChildren] = useState<boolean | undefined>(false)
+  const AvatarStackContainer = useRef(null)
```
joshblack, line 207: style nit (and definitely non blocking lol) we could use lowercase for `useRef` here since we typically only use pascal case for component names or modules

```suggestion
  const avatarStackContainer = useRef(null)
```

packages/react/src/AvatarStack/AvatarStack.tsx
```diff
@@ -249,6 +260,11 @@ const AvatarStack = ({
     )
   }
 
+  useEffect(() => {
+    const hasInteractive = getInteractiveNodes(AvatarStackContainer.current, '[tabindex]')
+    setHasInteractiveChildren(hasInteractive)
+  }, [children])
```
joshblack, line 266: Would this be equivalent to:

```tsx
useEffect(() => {
  // ...
})
```

I wasn't sure if `children` would change each render or not if they are not memoized. If that's the case, would it be worth running as a mutation observer, potentially?

joshblack (commented):
Looking great! Left a couple comments/questions/suggestions 👀

packages/react/src/internal/utils/getInteractiveNodes.ts
TylerJDev: Definitely! Added some more selectors that could be handy.

packages/react/src/internal/utils/getInteractiveNodes.ts
TylerJDev: That's a good callout. I changed the names of the function + file!

packages/react/src/AvatarStack/AvatarStack.tsx
TylerJDev: Didn't realize I was using pascal case 🤦 Should be updated now!

packages/react/src/AvatarStack/AvatarStack.tsx
TylerJDev: We could! My assumption is that we'd want to utilize the mutation observer inside of `hasInteractiveNodes`?

TylerJDev: I see, this is great! Thank you for the example too. I'll rework this a bit.

TylerJDev: After implementing this, I realize that the need for checking the children for changes is probably very slim 😅 I removed the `children` dependency so it should only run on the initial render, as that's the only time we'd actually want to run this function/check.

TylerJDev:
Requires small changes in Dotcom for tests, https://github.com/github/github/pull/348018/commits/833ad355c270f37218f3bcc4071319846b0e552a. Removes empty space that `AvatarStack` previously had. More context at https://github.com/github/memex/issues/16409.

packages/react/src/AvatarStack/AvatarStack.tsx
```diff
@@ -249,6 +260,11 @@ const AvatarStack = ({
     )
   }
 
+  useEffect(() => {
+    const hasInteractive = hasInteractiveNodes(avatarStackContainer.current, 'div[tabindex]')
+    setHasInteractiveChildren(hasInteractive)
+  }, [])
```
joshblack, line 266: With the move to not running on `children` changing, wanted to ask if this is something that should be setup in a mutation observer for when contents change in case the "hasInteractiveNodes" check could change?

TylerJDev: We could! I was thinking that we'd know whether to make the container interactive or not based on the initial render as I don't think there are cases where interactive items are rendered in after the first render. But we could add a mutation observer just to be safe.

joshblack (approved):

PR #5139, opened 2024-10-17
[Select panel] Loading states re-submission

broccolinisoup:
This PR is a re-submission of #4929 

Closes https://github.com/github/primer/issues/3921

#### Body spinner

Alt: a video of a button that, when clicked, opens a `SelectPanel`. There is a spinner where the list of items should be. After a few seconds, the spinner is replaced by the list of items.

https://github.com/user-attachments/assets/9124a657-0ee4-4618-bc88-6803706a80de

#### Body skeleton

Alt: a video of a button that, when clicked, opens a `SelectPanel`. There is a list of animated skeletons where the list of items should be. After a few seconds, the skeleton is replaced by the list of items.

https://github.com/user-attachments/assets/6be5e641-61cf-401b-b60e-e5575e93dabf

#### Input spinner

Alt: a video of an open `SelectPanel`. The letter "b" is entered into the filter input field. After a slight delay, a spinner appears inside the input field, replacing the magnifying glass icon that was there before. A few seconds later, the spinner is replaced the magnifying glass and the list of items is replaced by a filtered list of items. Each item contains the letter "b."

https://github.com/user-attachments/assets/27ca0751-e7e9-4995-8865-e0611cb4087f

#### Changed

In order to trigger loading the initial list of items, I had to manually call the `onFilterChange` callback, which is expected to be called with the filter input value (i.e. the text the user entered) and the corresponding `React.ChangeEvent` object. I was unable to figure out how to call it with a dummy `React.ChangeEvent` object, so I changed the type signature of the callback to allow `undefined` for the event object.

I used primer-query to identify all the existing usages of `SelectPanel` that pass a `onFilterChange` callback and have determined that nobody is actually using at second event arg. We could probably remove it entirely.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

I'm conflicted here. On the one hand, this is technically a breaking change and should therefore necessitate a major release. On the other hand, nobody in dotcom uses the second event parameter, so could be considered backwards-compatible if we consider dotcom to be our only consumer. Looking for feedback here.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/FilteredActionList/FilteredActionListLoaders.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <Box p={3} flexGrow={1} sx={{alignContent: 'center', textAlign: 'center', height: '100%'}}>
+      <Spinner {...props} />
+    </Box>
+  )
+}
+
+function LoadingSkeleton({rows = 10, ...props}: {rows: number}): JSX.Element {
+  return (
+    <Box p={2} display="flex" flexGrow={1} flexDirection="column">
+      <Stack id="foobarbaz" direction="vertical" justify="center" gap="condensed" {...props}>
```
francinelucca, line 56: is this "foobarbaz" id production-ready?

packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
```diff
@@ -40,16 +45,19 @@ const StyledHeader = styled.div`
...
+  onListContainerRefChanged,
+  onInputRefChanged,
   items,
   textInputProps,
-  inputRef: providedInputRef,
   sx,
   className,
   ...listProps
 }: FilteredActionListProps): JSX.Element {
+  const inputRef = useRef<HTMLInputElement>(null)
```
francinelucca, line 60: does this mean consumers will no longer be able to provide their own refs?

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
```diff
@@ -46,16 +47,19 @@ export function FilteredActionList({
...
   items,
   textInputProps,
-  inputRef: providedInputRef,
   sx,
   groupMetadata,
   showItemDividers,
   className,
   ...listProps
 }: FilteredActionListProps): JSX.Element {
+  const inputRef = useRef<HTMLInputElement>(null)
```
francinelucca, line 62: same question here as before

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
```diff
@@ -174,13 +198,15 @@ export function FilteredActionList({
                         {group.header?.title ? group.header.title : `Group ${group.groupId}`}
                       </ActionList.GroupHeading>
                       {getItemListForEachGroup(group.groupId).map((item, index) => {
-                        return <MappedActionListItem key={index} {...item} renderItem={listProps.renderItem} />
+                        const key = ('key' in item ? item.key : undefined) ?? item.id?.toString() ?? index.toString()
```
francinelucca, line 201:
```suggestion
                        const key = item.key ?? item.id?.toString() ?? index.toString()
```

I think these are equivalent

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
```diff
@@ -174,13 +198,15 @@ export function FilteredActionList({
...
+                        return <MappedActionListItem key={key} {...item} renderItem={listProps.renderItem} />
                       })}
                     </ActionList.Group>
                   )
                 })
               : items.map((item, index) => {
-                  return <MappedActionListItem key={index} {...item} renderItem={listProps.renderItem} />
+                  const key = ('key' in item ? item.key : undefined) ?? item.id?.toString() ?? index.toString()
```
francinelucca, line 208:
```suggestion
                  const key = item.key ?? item.id?.toString() ?? index.toString()
```

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -17,11 +17,97 @@ import type {FocusZoneHookSettings} from '../hooks/useFocusZone'
...
+function announceLoading() {
+  const liveRegion = document.querySelector('live-region')
+
+  liveRegion?.clear() // clear previous announcements
+
+  announce('Loading.', {
+    delayMs,
+    from: liveRegion ? liveRegion : undefined, // announce will create a liveRegion if it doesn't find one
+  })
+}
```
francinelucca, line 84: these are pretty similar, could probably be combined into a single function that takes the announce text as paramenter

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -101,21 +190,169 @@ export function SelectPanel({
...
     (value, e) => {
+      if (loadingManagedInternally) {
+        if (dataLoadedOnce) {
+          // If data has already been loaded once, delay the spinner a bit. This also helps
+          // not show and then immediately hide the spinner if items are loaded quickly, i.e.
+          // not async.
+
+          if (loadingDelayTimeoutId.current) {
+            safeClearTimeout(loadingDelayTimeoutId.current)
+          }
```
francinelucca, line 242: nit: could be taken outside the if/else to not repeat on line 256

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
```diff
@@ -104,20 +119,22 @@ export function FilteredActionList({
...
-      scrollIntoView(activeDescendantRef.current, scrollContainerRef.current, {...menuScrollMargins, behavior: 'auto'})
+      scrollIntoView(activeDescendantRef.current, scrollContainerRef.current, {
+        ...menuScrollMargins,
+        behavior: 'auto',
+      })
     }
   }, [items])
 
   useScrollFlash(scrollContainerRef)
-  useAnnouncements(items, listContainerRef, inputRef)
```
francinelucca, line 120: could removing this have any impact on people using FilteredActionList directly without a SelectPanel or is that a non-issue?

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -485,3 +488,85 @@ export const WithLabelInternally = () => {
...
+      options: ['spinner', 'skeleton'],
+    },
+    height: {
+      control: 'select',
+      options: ['auto', 'xsmall', 'small', 'medium', 'large', 'xlarge'],
+    },
+    componentManagesLoading: {
+      control: {
+        type: 'boolean',
+      },
```
francinelucca, line 569: What does this control do? I thought it'd let me turn of the loading in the storybook but found I was seeing the spinners either way

francinelucca (commented):
It's looking great, some questions and minor suggestions

packages/react/src/FilteredActionList/FilteredActionListLoaders.tsx
camertron: Hahahaha nope, good catch 😆

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
camertron: This controls whether the component, i.e. `SelectPanel` manages the loading state or if it's controlled outside the component by the consumer. The spinner/skeleton will still show up in both cases, but that's because the consumer, i.e. the storybook story, is passing `loading={true/false}`. You'll notice that, in the case `componentManagesLoading` is `false`, the inline text field spinner appears immediately (indicating the consumer flipped `loading` to `true`), whereas when `componentManagesLoading` is `true`, the text field spinner appears after a short delay.

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
camertron: I was under the impression that `FilteredActionList` was only used internally since it's not documented on primer.style, etc. However I just did a [search](https://primer-query.githubapp.com/?name=%22FilteredActionList%22) for it and there is one usage in dotcom 🤦

francinelucca: it's also in our Storybook https://primer.style/react/storybook/?path=/story/components-filteredactionlist--default

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
francinelucca: ahh, that makes sense, got it!

packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
camertron: Oh man yeah, good catch. I meant to go back and clean this up!

camertron:
Dotcom integration tests are failing because of https://github.com/primer/react/pull/5559, so we should be able to merge safely.

packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
```diff
@@ -40,14 +48,18 @@ const StyledHeader = styled.div`
...
   filterValue: externalFilterValue,
   onFilterChange,
+  onListContainerRefChanged,
+  onInputRefChanged,
   items,
   textInputProps,
   inputRef: providedInputRef,
   sx,
   className,
+  announcementsEnabled: _announcementsEnabled = true,
```
francinelucca, line 62: are we adding this here just so the APIs match? I see no use in this file. If so that's fine, just checking

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -485,3 +488,85 @@ export const WithLabelInternally = () => {
...
+  argTypes: {
+    initialLoadingType: {
+      control: 'select',
+      options: ['spinner', 'skeleton'],
+    },
+    height: {
+      control: 'select',
+      options: ['auto', 'xsmall', 'small', 'medium', 'large', 'xlarge'],
+    },
+    componentManagesLoading: {
```
francinelucca, line 566: would adding a control for 'loading' here as well make the difference between the two evident? it's really hard to grasp the difference between the two off this control in the UI

francinelucca (approved):
LGTM, non-blocking comments :shipit:

packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
camertron: Yep, you got it 👍 Hopefully this will all go away in the near future when we get rid of `FilteredActionListWithDeprecatedActionList`.

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
camertron: I wonder if perhaps we don't need this option... I added it for testing purposes, but I can't really see it being useful to us in the future, or to outside contributors.

camertron:
Dotcom changes need to ship before this PR can be merged (internal-only link): https://github.com/github/github/pull/359822

PR #5114, opened 2024-10-11
revert: set global focus styles

joshblack:
Revert https://github.com/primer/react/pull/5029/files due to issues identified in the release: https://github.slack.com/archives/C02NUUQ9C30/p1728628747488159

PR #5140, opened 2024-10-17
Reopen "set global focus styles"

francinelucca:
Reverts primer/react#5114

### DIFF
--- a/.changeset/itchy-paws-bake.md
+++ b/.changeset/itchy-paws-bake.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+fix(SegmentedControl): set global focus styles
--- a/packages/react/src/SegmentedControl/SegmentedControlButton.tsx
+++ b/packages/react/src/SegmentedControl/SegmentedControlButton.tsx
@@ -8,6 +8,7 @@ import sx, {merge} from '../sx'
 import {getSegmentedControlButtonStyles, getSegmentedControlListItemStyles} from './getSegmentedControlStyles'
 import {defaultSxProp} from '../utils/defaultSxProp'
 import {isElement} from 'react-is'
+import getGlobalFocusStyles from '../internal/utils/getGlobalFocusStyles'
 
 export type SegmentedControlButtonProps = {
   /** The visible label rendered in the button */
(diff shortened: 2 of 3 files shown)


TylerJDev (approved):

broccolinisoup (approved):
Looks wonderful! 🌸

Should we add an e2e test for the focus state? Especially because we use a utility function for the focus styles which is great - just to be double sure that we don't introduce any visual regression if we ever make a change on the global utility focus styles. Let me know your thoughts 🙌🏻

francinelucca:
> Should we add an e2e test for the focus state? Especially because we use a utility function for the focus styles which is great - just to be double sure that we don't introduce any visual regression if we ever make a change on the global utility focus styles. Let me know your thoughts 🙌🏻

Yesss , great idea ✨

francinelucca:
> Should we add an e2e test for the focus state? 

@broccolinisoup done in [8756ea4](https://github.com/primer/react/pull/5140/commits/8756ea40d7a966eb96d3aa54a1ec9c4db67b2cab), [5736299](https://github.com/primer/react/pull/5140/commits/57362998d0b27cbf1d5d5f2382f204c5b15f87f5). LMK if I did that right 😅

broccolinisoup:
> > Should we add an e2e test for the focus state?
> 
> @broccolinisoup done in [8756ea4](https://github.com/primer/react/pull/5140/commits/8756ea40d7a966eb96d3aa54a1ec9c4db67b2cab), [5736299](https://github.com/primer/react/pull/5140/commits/57362998d0b27cbf1d5d5f2382f204c5b15f87f5). LMK if I did that right 😅

Yeah perfect! 🔥

Side note: Not sure if we need the focus state for all stories but I don't think we have any space issues so far; so maybe we should just keep them all until we can't 😅

francinelucca:
> Side note: Not sure if we need the focus state for all stories but I don't think we have any space issues so far; so maybe we should just keep them all until we can't 😅

Was wondering that myself haha, what would be "standard" to just do it on the default for button/iconButton? maybe that makes more sense... happy to fix!

broccolinisoup:
>what would be "standard" to just do it on the default for button/iconButton?

Yeah that makes sense to me!

Issue #4664, opened 2024-06-11
Inline `ActionList.Description` can result in horizontal overflow


khiga8:
### Description

When `<ActionList.Description variant="inline">` contains very long content and we're on a small viewport,  the content overflows and causes a horizontal scrollbar to appear. This is an accessibility [WCAG Reflow issue](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html).

Notably, `<ActionList.Description variant="inline">` currently wraps the content with `Truncate` which also poses accessibility concerns since keyboard users aren't able to access the truncated content.

One option that comes to mind is to not truncate at all, and have the content wrap. This would resolve the horizontal overflow concern and the truncate concern. This also aligns with what the Primer Rails ActionList currently does:

![Screenshot of very long inline description in Primer Rails ActionList, which wraps around](https://github.com/primer/react/assets/16447748/5adc1e00-b177-4f79-a116-cea5dca2a753)

**However**, this would be a visual change to Primer React ActionList, and may benefit from designer input. As of right now, there are around  ~40 instances of `ActionList.Description variant="inline"` according to primer query ([source 1](https://musical-adventure-wlr3n3k.pages.github.io/primer?name=actionlist.description&attribute=-variant) + [source 2](https://musical-adventure-wlr3n3k.pages.github.io/primer?variant=inline&name=actionlist.description&attribute=variant&value=inline))

### Steps to reproduce

See [storybook example](https://primer-06e9cd32fd-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--with-trailing-action) from https://github.com/primer/react/pull/4665!

I also tried [Code Sandbox](https://codesandbox.io/p/sandbox/purple-bash-qkdkfc) but I could not get the latest version of primer/react to run.

### Version

v36.21.0

### Browser

Chrome

joshblack:
Hey there! Going through this in backlog cleaning today for PRC. Wanted to include design here to double-check on the change for horizontal overflow to not wrapping 👀

broccolinisoup:
Hello! @mperrotti suggested that we can wrap the content as we do in PVC. Since this is a rare case, we are okay with that. 

@khiga8 Feel free to push a PR if you already have, otherwise we will move this in our backlog.

PR #5169, opened 2024-10-23
fix(ActionList): do not truncate description by default

francinelucca:
Closes https://github.com/github/primer/issues/3472, Closes https://github.com/primer/react/issues/4664

The current react implementation of `ActionList.Description` truncates the text by default which poses accessibility concerns due to the content not being able to be read by keyboard users. Additionally, the inline truncation can cause horizontal overflow which causes a WCAG reflow problem. This PR mimics the functionality of PVC's ActionList, which wraps the text instead of truncating by default. An additional `truncate` prop has been introduced to allow truncation functionality when applicable (documented [here](https://github.com/primer/design/pull/884))

**Before**
rendered ActionList component with different descriptions including inline truncation but not inline without truncation
**After**
rendered ActionList component with different descriptions including inline with and without truncation

#### New

- Tests for `ActionList.Description` in block, inline truncation and inline no-truncation variants
- New `truncate` prop on `ActionList.Description`

#### Changed

- Modify ActionList `TextWrapAndTruncation` story to include new truncation/no-truncation use cases
- Modify render behavior of `ActionList.Description` to not render inside a `Truncate` component if new `truncate` prop is false (false by default)
- `flex-basis` style change on `ActionList.Description` to grow automatically if `truncate` is false
- `word-break` is now set to `normal` instead of `break-word` if the `Actionlist.Item` contains an inline `Actionlist.Description`
- Updates snapshots for VRT tests

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Test new truncation behavior in `TextWrapAndTruncation` story

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


francinelucca:
Hi @langermank. Wanted to checkin about a slight difference I'm seeing here between PVC/PRC in relations to [this PR](https://github.com/primer/react/pull/4331) and the `word-break` property. Wondering if we're ok with this or if we want to rethink this change? See video below

https://github.com/user-attachments/assets/4afedebd-c86d-45de-ad71-b58c7d6467e9

francinelucca:
Will need [these changes](https://github.com/github/github/pull/350286/commits/44256266a8337fa11d50342b4bca700b0fb2614b) to be pulled into dotcom upon PRC version update

langermank (approved):
Lets go! 🚀

packages/react/src/ActionList/Description.tsx
```diff
@@ -37,11 +42,11 @@ export const Description: React.FC<React.PropsWithChildren<ActionListDescription
 
   const {blockDescriptionId, inlineDescriptionId} = React.useContext(ItemContext)
 
-  return variant === 'block' ? (
+  return variant === 'block' || !truncate ? (
     <Box
       as="span"
       sx={merge(styles, sx as SxProp)}
-      id={blockDescriptionId}
+      id={variant === 'block' ? blockDescriptionId : inlineDescriptionId}
```
broccolinisoup, line 49: I didn't know we have different IDs for inline and block 💭 good to know

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -648,4 +648,52 @@ describe('ActionList', () => {
...
+        </ActionList>,
+      )
+
+      const description = getByText('Item 1 description')
+      expect(description.tagName).toBe('DIV')
+      expect(description).toHaveAttribute('title', 'Item 1 description')
+      expect(description).toHaveStyleRule('flex-basis', '0')
+      expect(description).toHaveStyleRule('text-overflow', 'ellipsis')
+      expect(description).toHaveStyleRule('overflow', 'hidden')
+      expect(description).toHaveStyleRule('white-space', 'nowrap')
```
broccolinisoup, line 683: Very detail tests!! Love it 🔥 

I’m also curious to learn when you prefer to use Jest for visual style testing versus a tool like Playwright. 👀

broccolinisoup (approved):
so cool 🔥

packages/react/src/ActionList/Description.tsx
francinelucca: was surprised by this one as well hahaah

packages/react/src/ActionList/ActionList.test.tsx
francinelucca: That's a great question! I actually didn't take the fact that we already had a playwright test for the truncation story into account when writing these so this might be a bit "overkill" haha. I think I tend to go for Jest because it's what I'm most familiar with but love how "complete" vrt testing is in that it'll fail on the smallest visual change in the entire story so it feels like it covers more ground than a specific jest test. That being said I'm a big fan of redundancy.

francinelucca:
> Will need [these changes](https://github.com/github/github/pull/350286/commits/44256266a8337fa11d50342b4bca700b0fb2614b) to be pulled into dotcom upon PRC version update

Tagging release conductor as FYI @siddharthkp, @broccolinisoup  since I see you're next :)

PR #5175, opened 2024-10-25
Overlay: Add proper roles w/ keyboard expectations to stories

TylerJDev:
We utilize `Overlay` in many different ways. Our stories for `Overlay` largely represent a modal dialog pattern. This PR aims to add proper roles with the proper keyboard expectations for those roles to each story. 

#### Changed

* Adds roles with keyboard expectations to `Overlay` stories

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [x] None; if selected, include a brief description as to why (Only includes changes to Storybook examples)

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Overlay/Overlay.features.stories.tsx
+++ b/packages/react/src/Overlay/Overlay.features.stories.tsx
@@ -53,7 +53,7 @@ export default {
 } as Meta
 
 interface OverlayProps {
-  anchorSide: AnchorSide
+  anchorSide?: AnchorSide
   role?: AriaRole
   right?: boolean
 }
--- a/packages/react/src/Overlay/Overlay.features.stories.tsx
+++ b/packages/react/src/Overlay/Overlay.features.stories.tsx
@@ -1,6 +1,6 @@
 import React, {useState, useRef, useCallback} from 'react'
 import type {Meta} from '@storybook/react'
-import {TriangleDownIcon, PlusIcon, IssueDraftIcon} from '@primer/octicons-react'
+import {TriangleDownIcon, PlusIcon, IssueDraftIcon, XIcon} from '@primer/octicons-react'
 import {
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

francinelucca (commented):
Looking good! 
One question: For the positioned overlays story, I have to hit `esc` twice to get one of the overlays to close, is this intended? In prod I see we only have to click it once. 
I also noticed the default story is not showing up neither here nor in prod, do you know what that's about? I can open a separate issue for that 
Everything else looks good to me!

TylerJDev:
> One question: For the positioned overlays story, I have to hit esc twice to get one of the overlays to close, is this intended? In prod I see we only have to click it once.

Good catch, this seems to be because we're using an IconButton here with `Tooltip`. <kbd>Escape</kbd> will close the tooltip first before closing the dialog. I'm wondering if we should change this to a standard `<button>` with "close" as the text over an `IconButton`?

As for the default story, I believe it's because we only have a `.features` file for `Overlay`. I'm guessing this was a choice to only show features, but I'm not sure 🤔 We could create a separate issue for it and see if there are any other components w/o one.

francinelucca:
> Good catch, this seems to be because we're using an IconButton here with `Tooltip`. Escape will close the tooltip first before closing the dialog. I'm wondering if we should change this to a standard `<button>` with "close" as the text over an `IconButton`?

Ahh I didn't notice the first `esc` was closing the tooltip, that may be ok then. Your call

Opened issue for default story: https://github.com/primer/react/issues/5183

francinelucca (approved):
🚀

PR #5176, opened 2024-10-25
[WIP] fix(Timeline): render as unordered list

francinelucca:
Closes https://github.com/github/primer/issues/3354

Just explorations for now, but need the deployment

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Timeline/Timeline.tsx
+++ b/packages/react/src/Timeline/Timeline.tsx
@@ -7,7 +7,7 @@ import type {SxProp} from '../sx'
 import sx from '../sx'
 import type {ComponentProps} from '../utils/types'
 
-const Timeline = styled.div<{clipSidebar?: boolean} & SxProp>`
+const Timeline = styled.ul<{clipSidebar?: boolean} & SxProp>`
   display: flex;
   flex-direction: column;
   ${props =>
@@ -27,7 +27,7 @@ const Timeline = styled.div<{clipSidebar?: boolean} & SxProp>`
 
 type StyledTimelineItemProps = {condensed?: boolean} & SxProp
 
-const TimelineItem = styled.div.attrs<StyledTimelineItemProps>(props => ({
+const TimelineItem = styled.li.attrs<StyledTimelineItemProps>(props => ({
   className: clsx('Timeline-Item', props.className),
 }))<StyledTimelineItemProps>`
   display: flex;
@@ -108,7 +108,7 @@ const TimelineBody = styled.div<SxProp>`
   ${sx};
 `
 
-const TimelineBreak = styled.div<SxProp>`
+const TimelineBreak = styled.li<SxProp>`
   position: relative;
   z-index: 1;
   height: 24px;


packages/react/src/Timeline/Timeline.tsx
```diff
@@ -68,6 +78,21 @@ const TimelineItem = styled.div.attrs<StyledTimelineItemProps>(props => ({
...
 `
 
+export type TimelineItemsProps<As extends React.ElementType> = {
+  as?: As
+  condensed?: boolean
+} & React.ComponentPropsWithRef<React.ElementType extends As ? 'div' : As> &
+  SxProp
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function TimelineItem<As extends React.ElementType>(props: TimelineItemsProps<As>, ref: ForwardedRef<any>) {
```
francinelucca, line 88: Taking a page out of [Branchname](https://github.com/primer/react/blob/main/packages/react/src/BranchName/BranchName.tsx#L31) here, open to suggestion

joshblack (commented):
Sorry for the delay! Just wanted to ask, would it be possible for the behavior of `Timeline.Group` to live inside of `Timeline`? Basically have the flag turn Timeline into a ul and Timeline.Item into an `li`

If I remember right, when we last talked about this that would have a problem when we needed to make separate lists, is that right? If that's the case, could we recommend folks do multiple timelines?

My only hesitation with `Timeline.Group` is that it seems like it would always be necessary which made me think if we could have it baked into `Timeline` or not. Let me know what you think!

francinelucca:
> If I remember right, when we last talked about this that would have a problem when we needed to make separate lists, is that right? If that's the case, could we recommend folks do multiple timelines?

Yeah I think we can do that, the thing is we need `Timeline.Break` outside of the list, so to replicate something like [this](https://primer-571d3679a1-13348165.drafts.github.io/storybook/?path=/story/components-timeline-features--timeline-break), with that approach the code would look like this:

```
 <Timeline>
      <Timeline.Item>
        <Timeline.Badge sx={{
        bg: 'done.emphasis'
      }}>
          <Octicon icon={GitMergeIcon} color="fg.onEmphasis" aria-label="Merged" />
        </Timeline.Badge>
        <Timeline.Body>This is a message</Timeline.Body>
      </Timeline.Item>
</Timeline>
<Timeline.Break />
<Timeline>
    <Timeline.Item>
      <Timeline.Badge>
        <Octicon icon={GitBranchIcon} aria-label="Branch" />
      </Timeline.Badge>
      <Timeline.Body>This is a message</Timeline.Body>
    </Timeline.Item>
</Timeline>
```
I originally added the `Timeline.Group` because I just thought it looked more "neat" to have it all under one top-level element, but if that looks good to you I'm happy to implement it this way instead @joshblack

francinelucca:
Decided to close issue, closing PR

PR #5177, opened 2024-10-27
Allow primitives dep to be if version 9 OR 10 and add it to peerDep

lukasoppermann:
Allow primitives dep to be if version 9 OR 10 and add it to peerDep

Closes https://github.com/github/primer/issues/4322

#### Changed
dependency `@primer/primitives` to allow both major version `9` and `10` and move it from dependency to peer dendency

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -85,7 +85,7 @@
     "@primer/behaviors": "^1.7.2",
     "@primer/live-region-element": "^0.7.1",
     "@primer/octicons-react": "^19.9.0",
-    "@primer/primitives": "^9.0.3",
+    "@primer/primitives": "9.x || 10.x",
     "@styled-system/css": "^5.1.5",
     "@styled-system/props": "^5.1.5",
     "@styled-system/theme-get": "^5.1.2",
--- a/.changeset/spotty-rice-explain.md
+++ b/.changeset/spotty-rice-explain.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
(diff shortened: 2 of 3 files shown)


keithamus (approved):

packages/react/package.json
```diff
@@ -209,6 +208,7 @@
     "yaml": "2.3.4"
   },
   "peerDependencies": {
+    "@primer/primitives": "9.x || 10.x",
```
joshblack, line 211: I think a move to this being a peer dependency would be a breaking change on our end (since we would require downstream consumers to now download the package in order to use it)

Could we keep this as a dependency until then? We could add moving it to a peer dependency to our v38 roadmap if that would work

joshblack (changes requested):

packages/react/package.json
lukasoppermann: CC: @langermank This does sound logical to me, but does this break your strategy for the new release?

langermank: @joshblack ahh I didn't realize this part of the change is breaking. I think without it we will still be able to unblock dotcom bumps though 👍 

@lukasoppermann I can update the issue to just change the version now and later do a breaking `peerDep` change

lukasoppermann: @joshblack I updated it, can we merge it now?

PR #5192, opened 2024-10-29
feat(Header): Convert Header to CSS modules behind team feature flag

hussam-i-am:
Closes https://github.com/github/primer/issues/4138

#### Changed

Convert Header to CSS modules behind team feature flag

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


hussam-i-am:
Updated the unit and storybook tests to verify `className`, as well as added support for `as` prop for `Header` since it was failing in the integration build

packages/react/src/Header/Header.tsx
```diff
@@ -4,68 +4,131 @@ import {get} from '../constants'
...
+const HeaderItem = React.forwardRef<HTMLElement, StyledHeaderItemProps>(function HeaderItem(
+  {children, className, ...rest},
+  forwardRef,
+) {
+  const enabled = useFeatureFlag(CSS_MODULES_FEATURE_FLAG)
+  return (
+    <StyledHeaderItem
+      ref={forwardRef}
+      className={clsx(className, enabled && classes.HeaderItem, enabled && rest.full && classes.Full)}
+      {...rest}
```
langermank, line 82: For variant type props, we're using data-attributes for styling instead of classnames. You can read a little bit about that in the [CSS authoring guide](https://github.com/primer/react/blob/main/contributor-docs/authoring-css.md#component-prop-variants-as-data-attributes).

So for the `full` prop, instead of a classname it would look something like this:

```suggestion
      className={clsx(className, enabled && classes.HeaderItem}
      data-full={full}
      {...rest}
```

And then in CSS 

```
&:where([data-full]) {
  ....
}

If you check out the ButtonBase component you can see this in action 😄

Comment (hussam-i-am) [reply packages/react/src/Header/Header.tsx]: Nice, makes a lot of sense, updated!

**Review by jonrohan — APPROVED** [review]

File: packages/react/src/Header/Header.tsx [hunk 2]
```diff
@@ -4,68 +4,132 @@ import {get} from '../constants'
...
+  'a',
+  styled.a.attrs<StyledHeaderLinkProps>(({to}) => {
+    const isReactRouter = typeof to === 'string'
+    if (isReactRouter) {
+      // according to their docs, NavLink supports aria-current:
+      // https://reacttraining.com/react-router/web/api/NavLink/aria-current-string
+      return {'aria-current': 'page'}
+    } else {
+      return {}
+    }
```
joshblack, line 103: I think this logic will need to be ported over into the `HeaderLink` so that it works with the flag enabled/disabled. Right now, it will only be applied when the styled component is used if I understand right 👀

packages/react/src/Header/Header.tsx
```diff
@@ -4,68 +4,132 @@ import {get} from '../constants'
...
+  {children, className, ...rest},
+  forwardRef,
+) {
+  const enabled = useFeatureFlag(CSS_MODULES_FEATURE_FLAG)
+  return (
+    <StyledHeaderItem
+      ref={forwardRef}
+      className={clsx(className, enabled && classes.HeaderItem)}
+      data-full={rest.full}
+      {...rest}
```
joshblack, line 83: I think a challenge with this is that `full` will be an attribute when the flag is enabled and will give us an unexpected attribute error, e.g.

image

I think we should either update the styled component to use the data attribute pattern or only pass along `full` when the flag is enabled, if that makes sense.

joshblack (commented):
Sorry I was late on this! Just had a couple comments for this that I think would be good to cover

PR #5193, opened 2024-10-29
feat(VisuallyHidden): Convert VisuallyHidden to CSS module behind feature flag

randall-krauskopf:
Related To: https://github.com/github/primer/issues/4368

#### Changed
Convert VisuallyHidden component to css module behind feature flag

### Rollout strategy

- [ ] Patch release
- [x] Minor release


packages/react/src/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -12,17 +19,34 @@ import sx from '../sx'
  *
  * @see https://www.scottohara.me/blog/2023/03/21/visually-hidden-hack.html
  */
-export const VisuallyHidden = styled.span<SxProp>`
```
joshblack, line 15: Just wanted to share in case it was helpful (and feel free to ignore me if this is obvious 😂), when migrating styled components that are exported two helpful things I've found to double-check check are the `as` prop and `ref` support.

When we update these kinds of components these don't come along and they may be used downstream. It's hard to tell if these are breaking changes or not so I think we've been evaluating it by seeing if they are used this way in dotcom when updating.

Thankfully the integration tests tend to do a good job of double-checking this, also feel free to use Primer Query (e.g. [this link](https://primer-query.githubapp.com/?name=visuallyhidden), if you need to login then you'll have to click the link again since query params aren't persisted 😅)

joshblack (approved):
Looks great! Just left a comment in case it's helpful info. I think you're totally good to go though since integration checks passed 🔥

packages/react/src/VisuallyHidden/VisuallyHidden.tsx
randall-krauskopf: Thank you for the reminder. I'll definitely add this to my routine on components moving forward!

PR #5197, opened 2024-10-31
feat(Pagehead): Convert `Pagehead` component to CSS Modules behind feature flag

randall-krauskopf:
Closes https://github.com/github/primer/issues/4137

#### Changed
Migrated `Pagehead` component to CSS Modules behind feature flag.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### DIFF
--- a/packages/react/src/Pagehead/Pagehead.types.test.tsx
+++ b/packages/react/src/Pagehead/Pagehead.types.test.tsx
@@ -6,6 +6,5 @@ export function shouldAcceptCallWithNoProps() {
 }
 
 export function shouldNotAcceptSystemProps() {
-  // @ts-expect-error system props should not be accepted
   return <Pagehead backgroundColor="orchid" />
 }
--- a/.changeset/quiet-seahorses-yawn.md
+++ b/.changeset/quiet-seahorses-yawn.md
@@ -2,4 +2,4 @@
 "@primer/react": minor
 ---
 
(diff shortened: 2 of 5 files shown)


packages/react/src/Pagehead/Pagehead.types.test.tsx
```diff
@@ -6,6 +6,5 @@ export function shouldAcceptCallWithNoProps() {
 }
 
 export function shouldNotAcceptSystemProps() {
-  // @ts-expect-error system props should not be accepted
```
jonrohan, line 9: Torn on this change, cause it now will technically accept the system props we were blocking before. I think the change comes from the toggleStyledComponent somewhere. But this is temporary. When we remove the feature flag it will be an error again.

jonrohan (approved):

packages/react/src/Pagehead/Pagehead.types.test.tsx
randall-krauskopf: I can hold off on this and try and find a solution if you like? I don't mind, the temporary change could be in place for awhile depending on how things go?

jonrohan: Na, if we work fast enough on this, it won't make a difference. and I don't think folks are adding these attributes in dotcom much anymore. if at all

PR #5198, opened 2024-10-31
SelectPanel: Add variant=modal

siddharthkp:
- Built on top of https://github.com/primer/react/pull/5170
- Closes https://github.com/github/primer/issues/4106

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -369,3 +369,70 @@ export const WithGroups = () => {
...
+
+  return (
+    <>
+      <h1>Multi Select Panel as Modal with secondary action</h1>
+      <SelectPanel
+        variant="modal"
+        onCancel={() => setSelected(initialItems)}
+        // backward compatible API choice
+        // TODO: improve this API, rename it to a single secondaryAction or secondaryFooterSlotSomething
+        footer={<Button size="small">Edit labels</Button>}
```
siddharthkp, line 421: Todo 1: While footer is backward compatible, it's not obvious what it means. What we want here is a "secondary action".

Looking at usage ([3 instances](https://primer-query.githubapp.com/?name=selectpanel&attribute=footer)), we have
- Button
- LinkButton
- custom FooterActions component that has a spinner or an error message! (these cases are filling in for missing features and we should be able to simplify them with error messages and loading states)

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -355,103 +372,155 @@ export function SelectPanel({
...
+                        type="button"
+                        size="small"
+                        onClick={() => {
+                          if (typeof onCancel === 'function') onCancel()
+                          onClose('cancel-click')
+                        }}
+                      >
+                        Cancel
+                      </Button>
+                      {/* TODO: loading state for save? */}
```
siddharthkp, line 505: Todo 2: Once we have loading states, add one for "saving"

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -369,3 +369,70 @@ export const WithGroups = () => {
...
+  const filteredItems = items.filter(item => item.text.toLowerCase().startsWith(filter.toLowerCase()))
+  const [open, setOpen] = useState(false)
+  const buttonRef = useRef<HTMLButtonElement>(null)
+
+  return (
+    <>
+      <h1>Multi Select Panel as Modal</h1>
+      <SelectPanel
+        variant="modal"
+        onCancel={() => setSelected(initialItems)}
```
broccolinisoup, line 386: I wonder if it is doable/makes sense to handle the cancel in the component? 🤔 We can tae a call back in case there is anything else needs to be done after the cancel. Curious to hear what you think

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -31,23 +31,40 @@ interface SelectPanelMultiSelection {
...
   inputLabel?: string
   overlayProps?: Partial<OverlayProps>
   footer?: string | React.ReactElement
 }
 
+type SelectPanelVariantProps =
+  | {
+      // do we need keep the old variants for backward-compat?
+      // there isn't any usage in dotcom
+      variant: 'anchored' | FilteredActionListProps['variant']
```
broccolinisoup, line 53: >do we need keep the old variants for backward-compat? there isn't any usage in dotcom

I think we should. Although I am not sure if should merge the old variants and the new ones since they seem different variants to me. The new one is more of a type (modal, anchored) and the other (full and inset ) feels related to the stylings of the panel. 

One idea might be (For the sake of keeping things uncomplicated) we can merge them and immediately deprecated the old variants?

Curious to hear your thoughts

broccolinisoup (commented):
Thanks for working on this super quickly 🔥 Let a few questions. curious to hear your thoughts

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
siddharthkp: I would like that a lot. A combination of onSubmit and onCancel so that developers don't have to take care of transient state buuuuut that doesn't match what we have right now. `onSelectedChange` is fired immediately when an item is selected and the developer are expected to keep track of `selected` items and pass them back to SelectPanel 🤷 

I'd like to explore an API that abstracts the transient state but that is likely a breaking change, so only as a bonus at the end.

packages/react/src/SelectPanel/SelectPanel.tsx
siddharthkp: > Although I am not sure if should merge the old variants and the new ones since they seem different variants to me

That's very true.

I personally would like to get rid of the old ones because it's a design changing option that has _leaked out_ of deprecated ActionList all the way to SelectPanel. I don't think we want folks to use the full width with SelectPanel.

> One idea might be (For the sake of keeping things uncomplicated) we can merge them and immediately deprecated the old variants?

I like that

Luckily, there is no usage in dotcom and only [5 instances in internal status page](https://primer-query.githubapp.com/?name=selectpanel&attribute=variant&entrypoint=%22%40primer%2Freact%22), so it feels like we can remove them eventually.

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
broccolinisoup: >buuuuut that doesn't match what we have right now. onSelectedChange is fired immediately when an item is selected and the developer are expected to keep track of selected items and pass them back to SelectPanel 🤷

Yeah, I see! We don't have much control over the selected items, and it would probably violate the pattern if we try managing the state of the "selected" 

>I'd like to explore an API that abstracts the transient state but that is likely going to be a breaking change, so probably as a bonus task right at the end of the project

That would be interesting to explore! Yeah, no worries at all.

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-External-Anchor-light-modern-action-list--true-linux.png
siddharthkp, line 1: The padding on the is very off!

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-External-Anchor-light-linux.png
siddharthkp, line 1: I think the hover color is different?? It's really hard to tell

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-External-Anchor-light-modern-action-list--true-linux.png
siddharthkp: Resolved!

PR #5199, opened 2024-10-31
Remove root `<Select>` component

randall-krauskopf:
Related to: https://github.com/github/primer/issues/4299

#### Removed
Removed unused duplicate select component that was discovered while migrating the `src/Select/Select.tsx` file to use CSS modules.

### Rollout strategy

- [x] Minor release


randall-krauskopf:
@joshblack Follow up PR to remove the duplicate `<Select>` component we talked about yesterday. The integration test seems to be passing. I have this marked as a `minor` patch since it doesn't seem like anyone is using it anymore? I'm a bit unsure so lmk if I need to update it.

joshblack (approved):
LGTM if checks pass downstream! 🤞

joshblack:
@randall-krauskopf I think from a public API perspective this should be no change at all so feel free to skip a changeset for this one 👍

PR #5200, opened 2024-10-31
fix(ButtonGroup): add toolbar interactions for role toolbar

francinelucca:
Closes https://github.com/github/primer/issues/3423

Introduces an accessibility improvement to `ButtonGroup` by adding correct toolbar interactions (navigate with left/right arrow instead of tab) when supplied role="toolbar".

#### New

- Test file and tests for `ButtonGroup`
- Added `ButtonGroup` "As Toolbar" story
- Added e2e test for new `ButtonGroup` "As Toolbar" story
- Snapshots for new story

#### Changed

- Export `ButtonGroup` as `PolymorphicForwardRefComponent<'div', ButtonGroupProps>`
- Update `ButtonGroup` to use `focusZone` with left/right arrow when `role="toolbar"` is supplied.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Test deployed **As Toolbar** story, notice keyboard interaction. Compare to other ButtonGroup stories (these should have normal tab keyboard navigation)

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -73,24 +76,39 @@ const StyledButtonGroup = toggleStyledComponent(
...
         [classes.ButtonGroup]: enabled,
       })}
+      role={role}
       {...rest}
     >
       {children}
     </StyledButtonGroup>
   )
-})
+}) as PolymorphicForwardRefComponent<'div', ButtonGroupProps>
```
francinelucca, line 111: Had to add this because the tests where complaining about types when trying to spread (`{...props}`) of type `ButtonGroupProps` 🤷🏽‍♀️

francinelucca:
image
image
image

Looks good! closing

packages/react/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -73,24 +76,39 @@ const StyledButtonGroup = toggleStyledComponent(
   `,
 )
 
-export type ButtonGroupProps = ComponentProps<typeof StyledButtonGroup>
+export type ButtonGroupProps = ComponentProps<typeof StyledButtonGroup> & {
+  /**
+   * Settings to apply to the Focus Zone on the ButtonGroup container component. This is only used when role="toolbar" is supplied.
+   */
+  focusZoneSettings?: Partial<FocusZoneHookSettings>
```
joshblack, line 83: Would there be a way to capture relevant settings people might do as props on the component, instead? Just to avoid coupling this to `focusZoneSettings` if we want to refactor down the line.

TylerJDev: We could potentially forgo the `focusZoneSettings` prop since this is well scoped for only applying if `role="toolbar"` is true. Since the expectations are clear on how `role="toolbar"` should operate with keyboard, I think it's okay if consumers aren't able to pass their own settings.

I'm thinking if they truly needed to utilize their own settings, they could just apply their own `useFocusZone` outside of the component, as it _should_ replace the one that already exists, but I haven't tested so I'm not sure if it's true 😅

packages/react/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -73,24 +76,39 @@ const StyledButtonGroup = toggleStyledComponent(
...
+}
 const ButtonGroup = React.forwardRef<HTMLElement, ButtonGroupProps>(function ButtonGroup(
-  {children, className, ...rest},
+  {children, className, role, focusZoneSettings, ...rest},
   forwardRef,
 ) {
   const enabled = useFeatureFlag('primer_react_css_modules_team')
+  const buttonRef = useProvidedRefOrCreate(forwardRef as React.RefObject<HTMLDivElement>)
+
+  useFocusZone({
```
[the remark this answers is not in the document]
francinelucca: Done in [e8782e2](https://github.com/primer/react/pull/5200/commits/e8782e2267161836db8029160ba455157cdcae36)! Thanks so much for the suggestion 🙏🏼

packages/react/src/ButtonGroup/ButtonGroup.tsx
francinelucca: Did a quick test and Tyler's right, `useFocusZone` in outer component will override behavior, removed `focusZoneSettings` here [e8782e2](https://github.com/primer/react/pull/5200/commits/e8782e2267161836db8029160ba455157cdcae36). Thanks for the suggestion both!

siddharthkp:
Probably a noob question, but do we anticipate any overlap with [ActionBar](https://primer.style/react/storybook/?path=/story/experimental-components-actionbar--default), which is also a role="toolbar" with buttons

siddharthkp (approved):
Implementation looks good!

TylerJDev (approved):
Looks great ✨

francinelucca:
> Probably a noob question, but do we anticipate any overlap with [ActionBar](https://primer.style/react/storybook/?path=/story/experimental-components-actionbar--default), which is also a role="toolbar" with buttons and does that change our recommendation?
> 

I think it's fine to have both functionalities in the codebase, we're not making toolbar the default behavior for `ButtonGroup` but want to support accessibility-wise if that were the use case on the consumer end. `ActionBar` renders differently than `ButtonGroup` so I see how there could be valid use cases for both as toolbars depending on what the consumer is trying to do/convey. It also seems like the future of `ActionBar` on primer is uncertain (see https://github.slack.com/archives/GACAW0NPM/p1730231906392899) so I think we're good to move forward.

Tagging @primer/design-reviewers in case there's any thoughts

langermank (approved):

PR #5204, opened 2024-10-31
feat(InlineMessage): Convert `InlineMessage` component to CSS Modules behind feature flag

randall-krauskopf:
Closes https://github.com/github/primer/issues/4130

#### Changed
Migrated `InlineMessage` component to CSS Modules behind feature flag.

### Rollout strategy

- [ ] Patch release
- [x] Minor release


packages/react/src/InlineMessage/InlineMessage.tsx
```diff
@@ -1,78 +1,104 @@
...
+        className={clsx(classes.InlineMessage, className)}
+        {...rest}
+        data-size={size}
+        data-variant={variant}
+      >
+        {icon}
+        {children}
+      </StyledMessage>
+    )
+  }
```
joshblack, line 101: I bet we could keep using `StyledMessage` instead of having to branch here by toggling the class name on the flag, e.g.

```tsx
    <StyledMessage {...rest} className={clsx(className, { [classes.InlineMessage]: enabled })} data-size={size} data-variant={variant}>
      {icon}
      {children}
    </StyledMessage>
```

This could help to avoid any issues later on where it's easy to forget to add something to both branches with flagged code

packages/react/src/InlineMessage/InlineMessage.tsx
```diff
@@ -1,78 +1,104 @@
...
 
 const variantToSmallIcon: Record<MessageVariant, React.ReactNode> = {
-  warning: <AlertFillIcon className="InlineMessageIcon" size={12} />,
-  critical: <AlertFillIcon className="InlineMessageIcon" size={12} />,
-  success: <CheckCircleFillIcon className="InlineMessageIcon" size={12} />,
-  unavailable: <AlertFillIcon className="InlineMessageIcon" size={12} />,
+  warning: <AlertFillIcon className={classes.InlineMessageIcon} size={12} />,
+  critical: <AlertFillIcon className={classes.InlineMessageIcon} size={12} />,
+  success: <CheckCircleFillIcon className={classes.InlineMessageIcon} size={12} />,
+  unavailable: <AlertFillIcon className={classes.InlineMessageIcon} size={12} />,
```
joshblack, line 82: Should these also be behind a flag? Since the Styled component version will be looking for exactly "InlineMessageIcon" but with CSS Modules it will be hashed

Another option in the interim is that we could use the `.InlineMessageIcon` in CSS Modules so that these don't have to be flagged and we just use that class name in both flagged/unflagged scenarios

packages/react/src/InlineMessage/InlineMessage.dev.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+import {InlineMessage} from '.'
+
+const meta = {
+  title: 'Experimental/Components/InlineMessage/Dev',
+  component: InlineMessage,
+} satisfies Meta<typeof InlineMessage>
+
+export default meta
+
+export const Default = () => {
```
joshblack, line 12: One note with what Katie shared earlier, we should make sure these get added to our e2e tests too as we're adding in dev stories

packages/react/src/InlineMessage/InlineMessage.tsx
randall-krauskopf: yes, great callout!

randall-krauskopf: I'm not sure how that even worked before, pretty sure I tested on Storybook before I pushed these changes up

packages/react/src/InlineMessage/InlineMessage.tsx
randall-krauskopf: That's fair, I'm still not used to reading that `{ [classes.InlineMessage]: enabled }` syntax.

joshblack: Either works to me! 👍

joshblack (approved):

hussam-i-am (approved):

langermank (approved):

PR #5205, opened 2024-10-31
Update CSS Module contribution doc to include adding new VRT

langermank:
Adding some info about adding new VRT tests to the CSS Module contribution doc

### DIFF
--- a/contributor-docs/migrating-to-css-modules.md
+++ b/contributor-docs/migrating-to-css-modules.md
@@ -132,7 +132,7 @@ Validate that no visual regressions occur when the feature flag is enabled. The 
 
 **Example test file**
 
-```js
+```ts
 import {test, expect} from '@playwright/test'
 import {visit} from '../test-helpers/storybook'
 import {themes} from '../test-helpers/themes'
--- a/contributor-docs/migrating-to-css-modules.md
+++ b/contributor-docs/migrating-to-css-modules.md
@@ -21,7 +21,6 @@ This guide outlines the steps to follow when refactoring Primer React components
 
 - **Verify VRT (Visual Regression Testing) Coverage:**
   - Check for missing VRT coverage. We utilize the VRT tests to make sure we're matching styling in production with current expectations. Components should have a Storybook story for every "feature" or option available that impacts the UI for VRT to capture in a screenshot.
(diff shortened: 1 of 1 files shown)


contributor-docs/migrating-to-css-modules.md
```diff
@@ -121,11 +120,65 @@ This guide outlines the steps to follow when refactoring Primer React components
...
+#### Visual regression testing
+
+Validate that no visual regressions occur when the feature flag is enabled. The `vrt*` tests are setup to compare the feature flagged component with the original component and will fail if there is a mismatch.
+
+- Add `dev` stories for any edge cases spotted in production for the component (ie. `sx` prop, custom className, styled system attributes). `dev` stories may include things that we wouldn't normally recommend for the purpose of stress testing what happens when PRC components are overridden with custom styles.
+- Setup VRT tests for `dev` stories. Copy an existing test in the corresponding test file in the [e2e directory](https://github.com/primer/react/tree/main/e2e/components) and update the id to match the new `dev` story.
+
+**Example test file**
+
+```js
```
joshblack, line 135: Just a note for the syntax highlighting

```suggestion
```ts
```

joshblack (approved):

contributor-docs/migrating-to-css-modules.md
```diff
@@ -121,11 +120,65 @@ This guide outlines the steps to follow when refactoring Primer React components
...
     })
     ```
-- **Regression Testing:**
-  - Validate that no visual regressions occur when the feature flag is enabled. The `vrt*` tests are setup to compare the feature flagged component with the original component and will fail if there is a mismatch.
 - **Handling `sx` Prop:**
   - Confirm the `sx` prop behaves correctly with the feature flag enabled.
 
+#### Visual regression testing
+
+Validate that no visual regressions occur when the feature flag is enabled. The `vrt*` tests are setup to compare the feature flagged component with the original component and will fail if there is a mismatch.
```
randall-krauskopf, line 128: Seems like a good idea! Would we want to make sure we mark them to be removed once the CSS module migration is complete?

langermank: I think we want to keep the dev tests around because they are testing for `className` override support mainly. When the feature flags are all removed we will then remove the check for enabled vs. default, but we can do that all at once probably since each test uses the same configuration

jonrohan (approved):

hussam-i-am (approved):

PR #5207, opened 2024-11-01
try to repro github/primer#4027

siddharthkp:
- https://github.com/github/primer/issues/4027


packages/react/src/hooks/useResizeObserver.ts
```diff
@@ -31,5 +32,5 @@ export function useResizeObserver<T extends HTMLElement>(callback: ResizeObserve
     return () => {
       observer.disconnect()
     }
-  }, [target])
+  }, [targetEl])
```
siddharthkp, line 35: Note for reviewer: Adding a ref as dependency isn't a recommended pattern as changes to the value of the ref does not cause the effect to re-run. Replaced the ref with it's value instead

PR #5209, opened 2024-11-01
Fix HMR errors

dusave:
#### New

- Add optional chaining checks to prevent undefined references when hot-module reloading. This work is designed to support HMR in the dotcom monorepo specifically.

### Rollout strategy

- [x] Patch release
- [ ] Minor release


packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -260,7 +260,9 @@ const Overlay: React.FC<React.PropsWithChildren<MenuOverlayProps>> = ({
   // If the menu anchor is an icon button, we need to label the menu by tooltip that also labelled the anchor.
   const [anchorAriaLabelledby, setAnchorAriaLabelledby] = useState<null | string>(null)
   useEffect(() => {
-    if (anchorRef.current) {
+    // Necessary for HMR reloads
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (anchorRef?.current) {
```
joshblack, line 265: Just curious, do we know why this ref is undefined when doing our upstream work in dotcom? 🤔 Specifically for HMR

camertron: We _think_ it's because the DOM is being replaced, but honestly that's just a theory 😬

dusave: Like @camertron said, we think the DOM is being rebuilt and inserted, which temporarily seems to make the refs undefined. However, it's barely undefined long enough to matter. By the time the page renders, everything is rehydrated and all works fine. But for that split second while it's being swapped out, React throws a fit.

joshblack: I would have thought the ref would be stable since it ultimately is coming from `useRef` and the value returned from that wouldn't be `undefined` between renders  🤔 (although `current` totally would if the DOM is changing)

Could it be that the context value (`MenuContext`) is changing during HMR? It seems like we're explicitly casting this value as being defined in: https://github.com/primer/react/blob/f3a5ff22575f29dafc056d3d8a7edd7b3dac8c4b/packages/react/src/ActionMenu/ActionMenu.tsx#L255 but with the HMR stuff here that doesn't seem to be the case

camertron:
Based on our discussion during our sync meeting on Monday (11/18), let's merge this for now and conduct further research async 👍

camertron (approved):

PR #5229, opened 2024-11-04
fix(Blankslate): update styles to center description text

joshblack:
The text in Blankslate should be text-align: center instead of justified left 🔍 

#### Changed

- Set `text-align: center` on the Blankslate container

### Rollout strategy

- [x] Patch release


joshblack:
@jonrohan I noticed that in PVC this style is set at the container level and I wanted to ask if we should do it similarly or, since this component accepts children, if we should scope it to the description/heading? 

The only concern I would have is that anything rendered as children to Blankslate would also have text alignment this way (which could be intended but I wanted to double check).

jonrohan:
> since this component accepts children, if we should scope it to the description/heading?

Good point, I would probably scope to just the `.Description` class. Since scoping on the container will require children to have `text-align: left` to reset the style

PR #5230, opened 2024-11-05
SelectPanel: Prepare for non-anchored variants

siddharthkp:
- Based on https://github.com/primer/react/pull/5170

Replace AnchoredOverlay with Overlay + useAnchoredPosition to prepare for variants that would skip anchored position

need to check
- [x] are integration tests passing?
- [x] are the props okay? (were some props going to anchoredoverlay that need to be re-routed)
- [x] are the types okay? (same as props)
- [x] is the rendered dom the same?
   - button is missing tabIndex="0" and id, both of these are okay, as we don't need to do this forwarding anymore
   - check if data-has-active-descendant is still correct, it is
- [x] is the rendered a11y tree the same
- [x] check if footer focus works

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/SelectPanel/SelectPanel.tsx
+++ b/packages/react/src/SelectPanel/SelectPanel.tsx
@@ -1,7 +1,7 @@
 import {SearchIcon, TriangleDownIcon} from '@primer/octicons-react'
 import React, {useCallback, useMemo} from 'react'
 import type {AnchoredOverlayProps} from '../AnchoredOverlay'
-import {AnchoredOverlay} from '../AnchoredOverlay'
+import Overlay from '../Overlay'
 import type {AnchoredOverlayWrapperAnchorProps} from '../AnchoredOverlay/AnchoredOverlay'
 import Box from '../Box'
 import type {FilteredActionListProps} from '../FilteredActionList'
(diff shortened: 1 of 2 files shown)


siddharthkp:
@broccolinisoup Verified everything works well, ready for review

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -172,10 +219,13 @@ export function SelectPanel({
     })
   }, [onClose, onSelectedChange, items, selected])
 
-  const inputRef = React.useRef<HTMLInputElement>(null)
```
siddharthkp, line 175: Note for reviewer: inputRef still exists, it's just declared sooner

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -172,10 +219,13 @@ export function SelectPanel({
...
-  const focusTrapSettings = {
+  /** Focus trap */
+  useFocusTrap({
```
siddharthkp, line 223: Additional keys brought over from AnchoredOverlay

PR #5236, opened 2024-11-05
bugfix(Spinner): Spinner component size get's large when passing sx and size props

jonrohan:
Closes https://github.com/github/primer/issues/4391

This fixes an issue with the Spinner and CSS modules we discovered this morning. The problem is when `sx=` and `size=` is passed to the Spinner component, we render `<Box as={Spinner} {...props} />` but Box accepts a `size=` prop as well as `Spinner` so `Box` ends up creating some styles thinking that it's meant to render at different breakpoints.

To fix I figured we need to not pass `size` to Box. So to fix it, I ended up removing `Box` and using a `styled.div` component with nothing but `${sx}` inside to render any sx props passed along. 

This fixed the issue. We created a dev story to recreate the issue.

#### Changed

Spinner component size get's large when passing sx and size props

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/forty-rats-jog.md
+++ b/.changeset/forty-rats-jog.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Spinner component size get's large when passing sx and size props
--- a/.changeset/forty-rats-jog.md
+++ b/.changeset/forty-rats-jog.md
@@ -2,4 +2,4 @@
 "@primer/react": patch
 ---
 
-Spinner component size get's large when passing sx and size props
+Update Spinner component to correctly use the `size` prop when both `sx` and `size` are provided
--- a/packages/react/src/Spinner/Spinner.dev.stories.tsx
+++ b/packages/react/src/Spinner/Spinner.dev.stories.tsx
@@ -7,4 +7,4 @@ export default {
   component: Spinner,
 } as Meta<typeof Spinner>
 
-export const Default = () => <Spinner sx={{border: '1px solid red'}} />
+export const Default = () => <Spinner sx={{border: '1px solid red'}} size={'small'} />
--- a/packages/react/src/Spinner/Spinner.dev.stories.tsx
+++ b/packages/react/src/Spinner/Spinner.dev.stories.tsx
@@ -7,4 +7,4 @@ export default {
   component: Spinner,
 } as Meta<typeof Spinner>
 
-export const Default = () => <Spinner sx={{border: '1px solid red'}} size={'small'} />
+export const Default = () => <Spinner sx={{border: '1px solid red'}} size="small" />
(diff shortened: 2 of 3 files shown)


randall-krauskopf (approved):

.changeset/forty-rats-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Spinner component size get's large when passing sx and size props
```
joshblack, line 5: I think typo here for gets:

```suggestion
Spinner component size gets large when passing sx and size props
```

An alternative for the changelog too, not sure if it fully captures what is being changed though:

```suggestion
Update Spinner component to correctly use the `size` prop when both `sx` and `size` are provided
```

packages/react/src/Spinner/Spinner.dev.stories.tsx
```diff
@@ -7,4 +7,4 @@ export default {
   component: Spinner,
 } as Meta<typeof Spinner>
 
-export const Default = () => <Spinner sx={{border: '1px solid red'}} />
+export const Default = () => <Spinner sx={{border: '1px solid red'}} size={'small'} />
```
joshblack, line 10: For text-only values we can use this syntax instead:

```suggestion
export const Default = () => <Spinner sx={{border: '1px solid red'}} size="small" />
```

joshblack (changes requested):

packages/react/src/Spinner/Spinner.tsx
```diff
@@ -87,17 +87,21 @@ const StyledComponentSpinner = styled(Spinner)`
...
+function StyledSpinner({sx, className, ...props}: SpinnerProps) {
   const enabled = useFeatureFlag('primer_react_css_modules_team')
   if (enabled) {
     if (sx) {
-      return <Box sx={sx} as={Spinner} className={classes.SpinnerAnimation} {...props} />
+      return <StyledBaseSpinner sx={sx} as={Spinner} className={clsx(className, classes.SpinnerAnimation)} {...props} />
     }
 
-    return <Spinner className={classes.SpinnerAnimation} {...props} />
+    return <Spinner className={clsx(className, classes.SpinnerAnimation)} {...props} />
```
jonrohan, line 101: Not related to the bug, but was pointed out that className was overwriting the CSS modules classname when being passed in, so I added this

hussam-i-am (approved):

joshblack (approved):

PR #5257, opened 2024-11-08
feat(TextInput): Convert TextInput (Part 2) - Update CSS to use data attributes

hussam-i-am:
- Relates to https://github.com/github/primer/issues/4365

#### Changed

- As a precursor to https://github.com/primer/react/pull/5227, updated the styled components to use data attributes for styling and moved away from using the functions
- Replaced `get` calls with CSS variables and removed fallbacks
- Updated test helper to include the data attributes
- Updated snapshots to reflect reordered styles and CSS variables

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Not sure if it is a known issue but it seems that passing in `small` for `size` does not currently do anything, whereas with the changes made, we are getting smaller inputs

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/__tests__/Textarea.test.tsx
```diff
@@ -84,11 +84,11 @@ describe('Textarea', () => {
       width: '100%',
       display: 'flex',
     }
-    const defaultStyles = renderStyles(<Textarea block />)
-    const blockStyles = renderStyles(<Textarea />)
+    const defaultStyles = renderStyles(<Textarea />)
+    const blockStyles = renderStyles(<Textarea block />)
```
hussam-i-am, line 88: Updated so that variable names match component props

packages/react/src/utils/testing.tsx
```diff
@@ -87,15 +87,21 @@ export function percent(value: number | string): string {
...
-export function getComputedStyles(className: string) {
+export function getComputedStyles(className: string, restProps?: Record<string, string | undefined>) {
   const div = document.createElement('div')
   div.className = className
 
+  if (restProps) {
+    for (const [key, value] of Object.entries(restProps)) {
+      if (key.startsWith('data-') && value !== undefined) div.setAttribute(key, value)
+    }
+  }
```
hussam-i-am, line 103: Update to handle include data attributes for computed styles

packages/react/src/internal/components/TextInputWrapper.tsx
```diff
@@ -55,54 +14,34 @@ export type StyledBaseWrapperProps = {
...
-  `
-}
-
-export const TextInputBaseWrapper = styled.span<StyledBaseWrapperProps>`
-  font-size: ${get('fontSizes.1')};
-  line-height: 20px;
-  color: ${get('colors.fg.default')};
+export const StyledTextInputBaseWrapper = styled.span<StyledBaseWrapperProps>`
+  display: inline-flex;
+  min-height: var(--base-size-32);
```
joshblack, line 33: Typically we've still been using the `${get()}` helper for this but I'm curious what you would say @langermank if it's okay for us to just use custom properties now instead of the calls to the theme?

I think the only downside would be if someone is updating these values for the `ThemeProvider` then we aren't reading them anymore which I think would only impact folks outside of dotcom using Primer

packages/react/src/internal/components/TextInputWrapper.tsx
```diff
@@ -117,108 +56,208 @@ export const TextInputBaseWrapper = styled.span<StyledBaseWrapperProps>`
   textarea,
   select {
     &::placeholder {
-      color: var(---control-fgColor-placeholder, ${get('colors.fg.muted')});
+      color: var(---control-fgColor-placeholder, var(--fgColor-muted));
     }
   }
 
-  ${props => renderFocusStyles(Boolean(props.hasTrailingAction), Boolean(props.isInputFocused))}
+  &:where([data-trailing-action][data-focused]),
```
joshblack, line 63: Is `data-focused` something where we could use `:focus` or `:focus-visible` or is this something that's been providing through props or determined at runtime? Wasn't sure so wanted to double check 👀

packages/react/src/internal/components/TextInputWrapper.tsx
```diff
@@ -117,108 +56,208 @@ export const TextInputBaseWrapper = styled.span<StyledBaseWrapperProps>`
...
-  ${sizeVariants}
-  ${sx};
+    padding: var(--base-size-12);
+  }
+
+  &:where([data-contrast]) {
+    background-color: var(--bgColor-inset);
+  }
+
+  &:where([data-disabled]) {
```
joshblack, line 78: Is this something where we could use `:disabled` or do we need `data-disabled`?

packages/react/src/internal/components/TextInputWrapper.tsx
```diff
@@ -117,108 +56,208 @@ export const TextInputBaseWrapper = styled.span<StyledBaseWrapperProps>`
...
+  }
+
+  &:where([data-block]) {
+    display: flex;
+    width: 100%;
+    align-self: stretch;
+  }
+
+  /* Ensures inputs don' t zoom on mobile but are body-font size on desktop */
+  @media (min-width: var(--breakpoint-medium)) {
```
joshblack, line 117: Does this compile with stylus in styled-components v5? (Crazy if so, just wanted to double-check since I didn't think it would)

joshblack (commented):
Looking great! Honestly just wanted to say thanks for taking this on, such a big help 🙏 Left a couple of comments/questions, hope they make sense!

packages/react/src/internal/components/TextInputWrapper.tsx
hussam-i-am: It felt prudent to keep the logic matching to the existing styling. The contents of the wrapper could change and didn't want to rely on its state, but what is being passed in by the prop

packages/react/src/internal/components/TextInputWrapper.tsx
hussam-i-am: Looks like you're right and it does not, so reverting back to `get` in this case

packages/react/src/internal/components/TextInputWrapper.tsx
hussam-i-am: Similar to https://github.com/primer/react/pull/5257/files#r1841062688, being cautious and setting the attribute since the contents of the wrapper can vary

hussam-i-am:
Failing integration tests are the same owns failing in main https://github.com/primer/react/pull/5276

jonrohan (approved):

PR #4743, opened 2024-07-16
docs: Indicate the DataTable component has been reviewed for accessibility

smockle:
Closes https://github.com/github/primer/issues/1635 

This PR updates the frontmatter for the `DataTable` component to indicate the component has been reviewed for accessibility.

PR #5269, opened 2024-11-12
Update DataTable.docs.json

joshblack:
Update DataTable docs to indicate that the component has been reviewed for accessibility.

This was tackled in our old docs site here: https://github.com/primer/react/pull/4743 but was not included in the update.

#### Changed

- Update `a11yReviewed` to true for DataTable

### DIFF
--- a/packages/react/src/DataTable/DataTable.docs.json
+++ b/packages/react/src/DataTable/DataTable.docs.json
@@ -2,7 +2,7 @@
   "id": "data_table",
   "name": "DataTable",
   "status": "draft",
-  "a11yReviewed": false,
+  "a11yReviewed": true,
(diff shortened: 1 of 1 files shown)


packages/react/src/DataTable/DataTable.docs.json
```diff
@@ -2,7 +2,7 @@
   "id": "data_table",
   "name": "DataTable",
   "status": "draft",
```
lesliecdubs, line 4:
```suggestion
  "status": "alpha",
```

As per the checkmarks and discussion in https://github.com/github/primer/issues/1633#issuecomment-2471238498, it seems like this component has met Alpha criteria. Is that accurate?

PR #5221, opened 2024-11-04
feat(Avatar): Remove the CSS modules feature flag from Avatar

jonrohan:
#### Removed

Remove the CSS modules feature flag from Avatar

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #5271, opened 2024-11-13
feat(Token): Migrate to CSS modules behind feature flag Pt 2

randall-krauskopf:
Related To: https://github.com/github/primer/issues/4367

This PR is part 2 of breaking apart a [larger PR](https://github.com/primer/react/pull/5237) into grokkable parts.

Apologies for the larger PR but there was an odd interplay with the specificity across the various components that made it hard to break this PR into smaller chunks


randall-krauskopf:
Integration tests are failing due to the regression that should be fixed soon.

packages/react/src/Token/AvatarToken.module.css
```diff
@@ -0,0 +1,37 @@
+:root {
+  --spacing: calc(var(--base-size-4) * 2);
+}
+
+.AvatarContainer {
+  display: block;
+}
+
+/* TODO: Remove this once the avatar component is converted to CSS modules */
```
jonrohan, line 9: Avatar is converted https://github.com/primer/react/pull/5221

packages/react/src/Token/AvatarToken.module.css
```diff
@@ -0,0 +1,37 @@
...
+.Avatar {
+  width: 100% !important;
+  height: 100% !important;
+}
+
+.Token {
+  padding-left: var(--base-size-4) !important;
+}
+
+.AvatarContainer[data-size='small'] {
```
jonrohan, line 19:
```suggestion
.AvatarContainer:where([data-size='small']) {
```

packages/react/src/Token/AvatarToken.module.css
```diff
@@ -0,0 +1,37 @@
...
+  width: calc(16px - var(--spacing));
+  height: calc(16px - var(--spacing));
+}
+
+.AvatarContainer[data-size='medium'] {
```
jonrohan, line 24:
```suggestion
.AvatarContainer:where([data-size='medium']) {
```

packages/react/src/Token/AvatarToken.module.css
randall-krauskopf: Hi @jonrohan, I'll get these changes in. Quick question though, why is `:where` preferred over the bare attribute selector?

jonrohan: Where will reduce the specificity on the attribute selector and make it easier to override the styles in dotcom. We try and add it whenever the attribute lands on the same dom node where we add the passed in className

jonrohan (approved):

hussam-i-am (approved):

PR #5282, opened 2024-11-13
chore(Banner): Remove the CSS modules feature flag from Banner

jonrohan:
#### Removed

Remove the CSS modules feature flag from Banner

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Banner/Banner.test.tsx
```diff
@@ -115,9 +115,9 @@ describe('Banner', () => {
       />,
     )
 
-    expect(screen.getByRole('button', {name: 'test primary action'})).toBeInTheDocument()
+    expect(screen.queryAllByText('test primary action')[0]).toBeInTheDocument()
```
francinelucca, line 118: is this because the tests where failing? If so maybe we should run integration tests to make sure it doesn't happen in dotcom

jonrohan: For responsiveness we're rendering 2 elements now. I couldn't figure out if I could only `getByRole` for visible ones so I ended updating to get the first result

jonrohan: I think the jest tests couldn't differentiate between seen and not seen 🤔 will most likely need to fix tests in dotcom when this one lands

francinelucca: Is it possible to make the change in dotcom now so it doesn't break the release? We've been trying to do that if possible. If not, can we have the changes ready to go in a commit so that they can be cherry-picked into the release branch?

jonrohan:
I'm going to land this PR before this releases https://github.com/github/github/pull/353901

packages/react/src/Banner/Banner.test.tsx
```diff
@@ -148,8 +148,8 @@ describe('Banner', () => {
       />,
     )
 
-    expect(screen.getByRole('button', {name: 'test primary action'})).toBeInTheDocument()
-    expect(screen.getByRole('button', {name: 'test secondary action'})).toBeInTheDocument()
+    expect(screen.queryAllByRole('button', {name: 'test primary action', hidden: true}).length).toBe(2)
+    expect(screen.queryAllByRole('button', {name: 'test secondary action', hidden: true}).length).toBe(2)
```
francinelucca, line 152: can you walk me through this? we have to buttons now and they're both hidden?

jonrohan: There was always 2 buttons, but I think jest doesn't have access to the CSS modules styles like it did before, so it thinks they're both visible. I only really added the `hidden` for backwards compatibility but I think it's unnecessary here.

francinelucca (approved):

PR #5319, opened 2024-11-20
fix(Link): Do not pass through hoverColor prop in Link

hussam-i-am:
Fixes error in [integration tests](https://github.com/github/github/actions/runs/11937592991/job/33274055190?pr=351887):

```
      If the error is expected, test for it explicitly by mocking it out using jest.spyOn(console, 'error').mockImplementation() and test that the warning occurs.
  
      Warning: React does not recognize the `hoverColor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `hovercolor` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

#### Changed

- Converts the `hoverColor` prop to a `data-hover-color` attribute in the `Link` component.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Link/Link.tsx
+++ b/packages/react/src/Link/Link.tsx
@@ -17,7 +17,7 @@ type StyledLinkProps = {
   inline?: boolean
 } & SxProp
 
-const Link = forwardRef(({as: Component = 'a', className, inline, underline, ...props}, forwardedRef) => {
+const Link = forwardRef(({as: Component = 'a', className, inline, underline, hoverColor, ...props}, forwardedRef) => {
   const innerRef = React.useRef<HTMLAnchorElement>(null)
   useRefObjectAsForwardedRef(forwardedRef, innerRef)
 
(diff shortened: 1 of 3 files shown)


packages/react/src/Link/Link.tsx
```diff
@@ -53,6 +53,7 @@ const Link = forwardRef(({as: Component = 'a', className, inline, underline, ...
         data-muted={props.muted}
         data-inline={inline}
         data-underline={underline}
+        data-hover-color={hoverColor}
```
francinelucca, line 56: does this need a css counterpart or does it just work? 🤔

hussam-i-am: I don't believe it's actually being used any longer since it is deperecated, this just ensures that we don't pass in an invalid attribute into the dom

francinelucca: I see, thanks for explaining that!
Do we gain anything by adding the `data-hover-holor` then? could we just ignore the prop?

jonrohan (approved):

packages/react/src/Link/Link.tsx
siddharthkp: Hello! I'm Sid from the future. I ran into this prop. Looks like this is a remnant of styled-system, we can simply remove this prop.

PR #5321, opened 2024-11-20
Update lit labs imports

mattcosta7:
Updates `@lit-labs/react` to `@lit/react`

### Changelog

Just updates a dependency

### Rollout strategy

- [x] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #5323, opened 2024-11-21
Revert "Update lit labs imports"

hussam-i-am:
Reverts primer/react#5321


joshblack:
Sharing the last time we looked at this for reference: https://github.com/primer/react/pull/3708

I believe this will be addressed when PRC goes ESM only but things might have changed since then, not sure 👀

joshblack (approved):

PR #5342, opened 2024-12-02
refactor(FormControl): update existing code for sub-components to prep for migration

joshblack:
Update `FormControl` sub-components to be easier to migrate to CSS modules.

#### Changed

- Update the following `FormControl` sub-components to a different styled component style:
  - Label
  - Caption
  - LeadingVisual
  - Validation

### Rollout strategy

- [x] Minor release


package.json
```diff
@@ -93,5 +93,6 @@
       "webpack": false,
       "running": false
     }
-  ]
+  ],
+  "packageManager": "npm@10.9.1+sha512.c89530d37c4baa38afd43e76a077a84b9aa63840b986426584fd5c5a54ab0a0b21bb1595c851042b733784b0b43706d36a494b4d8ae1a086a762cb8d3f95942a"
```
jonrohan, line 97: Don't know if this was unintended?

packages/react/src/internal/components/InputLabel.tsx
```diff
@@ -38,37 +38,62 @@ const InputLabel: React.FC<React.PropsWithChildren<Props>> = ({
...
-        ...sx,
-      }}
+      sx={sx}
       {...props}
     >
       {required || requiredText ? (
-        <Box display="flex" as="span">
-          <Box mr={1}>{children}</Box>
+        <StyledRequiredText>
+          <div>{children}</div>
```
jonrohan, line 55: nit, but since this in inside a span, could it also be `<span>`?

```suggestion
          <span>{children}</span>
```

jonrohan (approved):

package.json
joshblack: @jonrohan happy to move it to another PR, this just makes it so that (if you have corepack enabled from Node.js) that it will this version. Mine kept auto-adding a field here since I have corepack on so I figured I could just commit it lol

packages/react/src/internal/components/InputLabel.tsx
joshblack: @jonrohan I think I got `span` for `StyledRequiredText` which should map to `<Box as="span"` but for the inner wrapper around `children` I think `Box` will default to `div`, or is that not the case? 👀

jonrohan: Think that's the case. And what you have matches existing. I get old school ick vibes having a block element inside an inline. I don't think it matters anymore plus these are resolving to flex anyways so it doesn't really matter 🤷🏻

joshblack: @jonrohan ah, sorry I totally misunderstood you earlier. Happy to update it to be span! What you said makes a lot of sense to me 👍

PR #5350, opened 2024-12-04
fix(TooltipV2): delay tooltip opening time by 50ms

francinelucca:
Closes https://github.com/github/primer/issues/4432

Adds a 50ms delay to show the tooltip when triggered via the mouse to prevent showing tooltips when the user is just "passing by"

#### New

- Set Tooltip event after mouseEnter under a timeout to prevent opening when the mouse is just "passing by"

#### Changed

- use `fireEvent` in overlay test instead of `userEvent` to trigger click.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Overlay/Overlay.test.tsx
```diff
@@ -128,7 +128,7 @@ describe('Overlay', () => {
     expect(container.getByText('Add to list')).toBeInTheDocument()
 
     // open second menu
-    await user.click(container.getByText('Create list'))
+    fireEvent.click(container.getByText('Create list'))
```
francinelucca, line 131: I tried for *minute* to get this to work with user.click, I replicated this test manually in storybook and it works as expected. I noticed userEvent was triggering different events than manual testing did 🤷🏽‍♀️

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -265,6 +267,8 @@ export const Tooltip = React.forwardRef(
         ) {
           tooltipElRef.current.hidePopover()
           setIsPopoverOpen(false)
+        } else {
+          setIsPopoverOpen(false)
```
francinelucca, line 271: Fixes bug where overlay would stay "stuck" and not able to close via keyboard due to popOver state being open but it not actually being in the DOM

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -213,6 +213,8 @@ export const Tooltip = React.forwardRef(
 
     const [isPopoverOpen, setIsPopoverOpen] = useState(false)
 
+    const [timeoutId, setTimeoutId] = useState<NodeJS.Timeout | null>(null)
```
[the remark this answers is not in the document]
francinelucca: Done in https://github.com/primer/react/pull/5350/commits/7bc59c47d877c5a9c6d501eac1dd13c7bb886e70!

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -213,6 +214,10 @@ export const Tooltip = React.forwardRef(
 
     const [isPopoverOpen, setIsPopoverOpen] = useState(false)
 
+    const timeoutRef = React.useRef(0)
```
joshblack, line 217: I didn't really realize how awkward this was with cancellation since you still need to keep track of the timeout 🤔 I think the type for this should be `number | null` technically and then we set it to `null` after clearing it but at that point the hook isn't doing much for us is it 😅 (really just has the effect clean-up)

joshblack (approved):
LGTM, only comment is around the timeoutId ref stuff and having it be `number | null`

PR #5370, opened 2024-12-05
refactor(FormControl): update to CSS Modules behind feature flag

joshblack:
Closes https://github.com/github/primer/issues/3812

#### Changed

- Update FormControl and shared internal components to use form control behind feature flag

### Rollout strategy

- [x] Minor release


packages/react/src/internal/components/InputLabel.module.css
```diff
@@ -0,0 +1,31 @@
+.Label {
+  align-self: flex-start;
+  display: block;
+  color: var(--fgColor-default);
+  cursor: pointer;
+  font-weight: 600;
```
langermank, line 6:
```suggestion
  font-weight: var(--base-text-weight-semibold);
```

packages/react/src/internal/components/InputLabel.module.css
```diff
@@ -0,0 +1,31 @@
...
+  font-size: var(--text-body-size-medium);
+
+  &:where([data-control-disabled]) {
+    color: var(--fgColor-muted);
```
langermank, line 10:
```suggestion
    color: var(--control-fgColor-disabled);
```

packages/react/src/internal/components/InputValidation.module.css
```diff
@@ -0,0 +1,30 @@
+.InputValidation {
+  color: var(--inputValidation-fgColor);
+  display: flex;
+  font-size: var(--text-body-size-small);
+  font-weight: 600;
```
langermank, line 5:
```suggestion
  font-weight: var(--base-text-weight-semibold);
```

jonrohan (approved):
Woo, here we go 🎉

PR #5409, opened 2024-12-10
Fix `border-color` on ActionList Group header

langermank:
Prepping for ActionList CSS modules and updating snapshots ✨


packages/react/src/ActionList/Group.tsx
```diff
@@ -97,7 +97,7 @@ export const Group: React.FC<React.PropsWithChildren<ActionListGroupProps>> = ({
           // because the heading is hidden from the accessibility tree and only used for presentation role.
           // We will instead use aria-label to label the list. See a line below.
           aria-labelledby={listRole ? undefined : groupHeadingId}
-          aria-label={listRole ? (title ?? (slots.groupHeading?.props.children as string)) : undefined}
+          aria-label={listRole ? title ?? (slots.groupHeading?.props.children as string) : undefined}
```
langermank, line 100:
```suggestion
          aria-label={listRole ? (title ?? (slots.groupHeading?.props.children as string)) : undefined}
```

francinelucca (approved):
✨

PR #5422, opened 2024-12-12
chore(deps): update @primer/octicons-react to 19.14.0

joshblack:
Update `@primer/octicons-react` to latest

#### Changed

- Update `@primer/octicons-react` to 19.14.0
- Update how `Octicon` forwards along the `sx` prop since this will now be spread on icon components

### Rollout strategy

- [ ] Patch release
- [x] Minor release

Selected `minor` here since this is a notable change to `@primer/octicons-react` that we might need to coordinate around.


packages/react/src/NavList/__snapshots__/NavList.test.tsx.snap
```diff
@@ -1897,11 +1900,14 @@ exports[`NavList.Item with NavList.SubNav has active styles if SubNav contains t
...
-                  class="c9"
+                  class="octicon octicon-chevron-down c9"
+                  display="inline-block"
                   fill="currentColor"
                   focusable="false"
                   height="16"
-                  style="display: inline-block; user-select: none; vertical-align: text-bottom; overflow: visible;"
+                  overflow="visible"
+                  style="vertical-align: text-bottom;"
+                  sx="[object Object]"
```
jonrohan, line 1910: Should this sx be inlining like this?

lukasoppermann: I am sure not, but how do we fix it?

langermank:
I think we should hold off on merging this until we can address the strange `sx` issue. I bumped PRC to 19.13.0 so we're a bit more up to date in the meantime.

packages/react/src/NavList/__snapshots__/NavList.test.tsx.snap
joshblack: I bet we have something in NavList passing `sx` to an icon when it should not have. With the recent changes, this means that the icon will spread this on the outermost node instead of it being dropped.

jonrohan (approved):

PR #5432, opened 2024-12-13
chore(NavList) add e2e tests

langermank:
- Add e2e snapshots for NavList
- Fix markup bug causes a11y failures

Adding snapshots caused an accessibility failure due to some incompatible markup. I added a new prop to try and address that. There are cases where this heading _should_ be a `div` (I think most cases?) but within NavList it should sometimes be an `li`.

before | after |
| -- | -- |
| image | image |


TylerJDev (approved):
LGTM! ✨

Issue #4734, opened 2024-07-11
Tooltip is misplaced when it is rendered on a dialog


broccolinisoup:
### Description

When using Dialog 2, if the first interactive item that receives focus uses Tooltip (v2), the tooltip appears in a misplaced. It is reported [internally on Slack](https://github.slack.com/archives/C0FSWLQ0Y/p1720018968831849).

Tooltip is corectly placed if it made it appeared again. it only happens when the dialog first opened.

It happens on Safari when the dialog is open with a mouse or keyboard. Other browsers, it only happends when the dialog is opened with using keyboard.

### Steps to reproduce

1. Go to issues
2. Navigate to the "New Issue" button with a keyboard
3. Enter

Observe that the tooltip is misplaced. 

### Version

v35.24.0

Amk-001:
> ### Description
> When using Dialog 2, if the first interactive item that receives focus uses Tooltip (v2), the tooltip appears in a misplaced. It is reported [internally on Slack](https://github.slack.com/archives/C0FSWLQ0Y/p1720018968831849).
> 
> Tooltip is corectly placed if it made it appeared again. it only happens when the dialog first opened.
> 
> It happens on Safari when the dialog is open with a mouse or keyboard. Other browsers, it only happends when the dialog is opened with using keyboard.
> 
> ### Steps to reproduce
> 1. Go to issues
> 2. Navigate to the "New Issue" button with a keyboard
> 3. Enter
> 
> Observe that the tooltip is misplaced.
> 
> ### Version
> v35.24.0
> 
> ### Browser
> _No response_

Amk-001:
fix

PR #5436, opened 2024-12-13
Dialog: Show `Tooltip` on "close" button

TylerJDev:
Part of https://github.com/github/primer/issues/3726

Removes `unsafeDisableTooltip` from `IconButton` in `Dialog`, and removes eslint disable along with it.

Dialog component showcasing the dialog's close button with a tooltip. The tooltip is positioned above the button, with the text 'Close'.

#### Removed

- Removed unsafeDisableTooltip

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
Some of the snapshots look a bit odd due to the tooltip placement, but I believe this is just something with how the snapshots were taken (maybe timing?) and not an actual issue with the tooltip, as it cannot be produced locally.

langermank (approved):
@TylerJDev those snapshots are kind of scary! Could we maybe adjust the test to add a short delay? Do you think that's the issue? 🤔

broccolinisoup:
@TylerJDev @langermank sorry to barge in 🙃  I saw the Slack notification and wanted to mention that we have one open issue regarding Tooltip misplacement on Dialogs that we probably need to fix it first 😬 https://github.com/primer/react/issues/4734

Amk-001 (changes requested):
Reconnect

packages/react/src/Dialog/Dialog.dev.stories.tsx
```diff
@@ -137,7 +137,7 @@ export const WithSx = ({width, height, subtitle}: DialogStoryProps) => {
             {buttonType: 'danger', content: 'Delete the universe', onClick: onDialogClose},
             {buttonType: 'primary', content: 'Proceed'},
           ]}
-          sx={{borderColor: 'accent.emphasis', borderWidth: '1px', borderStyle: 'solid'}}
+          sx={{borderColor: 'accent.emphasis', borderWidth: '1px', borderStyle: 'solid', animation: 'none !important'}}
```
TylerJDev, line 140: There was a very small difference between the `dev` component snapshots (0.02 pixel difference). It seems like this was caused by the animation, but only [affected the colorblind theme](https://github.com/primer/react/pull/5791/files#diff-b524b7368a3a947aab78b8c4429e5ee36858a6673007b4e1ed8a7dad12559de7). 

It would sometimes pass/fail, so I marked it off as being flaky via the custom CSS. It never affected the non-dev stories.

langermank (approved):
Excellent work solving this very weird, annoying problem!

PR #5437, opened 2024-12-13
test(e2e): add snapshots for DataTable stories

joshblack:
Add VRT for DataTable, including a mask for relative time elements since those will always be different.

#### Changed

- Update e2e tests for DataTable to include VRT tests

### Rollout strategy

This is adding snapshots for VRT for DataTable and does change our public API


e2e/components/DataTable.test.ts
```diff
@@ -2,202 +2,89 @@ import {test, expect} from '@playwright/test'
...
+              id: story.id,
+              globals: {
+                colorScheme: theme,
+              },
+            })
+
+            // Default state
+            expect(
+              await page.screenshot({
+                mask: await page
```
langermank, line 68: COOL!

langermank (approved):
🚀

PR #5443, opened 2024-12-13
Deprecate use of sx prop by deprecating the SxProp type definition

jonrohan:
Closes https://github.com/github/primer/issues/4571

This pull deprecates the `SxProp` type definition. This will inform anyone using the `sx` prop on these components that it is deprecated and should be CSS modules instead. The result is non-blocking but looks like this in the editor:

CleanShot 2024-12-13 at 13 46 43@2x

#### New

Depreacte use of sx prop by switching to a deprecated type definition

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/few-coins-kick.md
+++ b/.changeset/few-coins-kick.md
@@ -2,4 +2,4 @@
 "@primer/react": minor
 ---
 
-Depreacte use of sx prop by switching to a deprecated type definition
+Deprecate use of sx prop by switching to a deprecated type definition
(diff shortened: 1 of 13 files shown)


hussam-i-am (approved):

jonrohan:
Thinking on this more, I think only the converted components are using `SxProp` so might be better to just deprecate that type rather than creating a new one

PR #5448, opened 2024-12-16
Corrects styling on `Button`s that only have a `count` and `icon`

mperrotti:
Closes https://github.com/github/primer/issues/4501

Screenshot 2024-12-18 at 4 06 40 PM

#### New

- corrects padding on `<Button>` where there is only an icon and a counter label to closer align with `<IconButton>` sizing

#### Changed

Nothing

#### Removed

Nothing

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Confirm that this change doesn't cause any regressions with existing IconButton behavior.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


langermank (dismissed):
Discussed on Slack and summarizing my thoughts here!

I would suggest putting this logic in `Button` to avoid adding a wrapper span inside `IconButton`. In order to utilize Button's layout wrapper `.ButtonContent` we're adding it to `IconButton` which feels like its adding unnecessary complexity given that this already exists in `Button` 😄 

I would suggest...

- Add a data-attribute to `ButtonBase` when `count` is true `data-has-count` for example
- Add a CSS selector `&:where([data-has-count])` and check if `.Label` is not there (not has?)
- Remove padding if has count is true, `.Label` is not present. You could also check that a leading visual is present but I'm not sure that's necessary

If this doesn't work out I would feel better about that span being conditional in IconButton!

packages/react/src/Button/ButtonBase.module.css
```diff
@@ -505,6 +505,20 @@
...
+    padding-inline: var(--control-medium-paddingInline-condensed);
+
+    &:where([data-size='small']) {
+      padding-inline: var(--control-xsmall-paddingInline-condensed);
+    }
+
+    &:where([data-size='large']) {
+      padding-inline: var(--control-large-paddingInline-normal);
+    }
+  }
```
langermank, line 521:
```suggestion
  &:where([data-has-count]):has([data-component='leadingVisual']):not(:has([data-component='text'])) {
  /* stylelint-disable-next-line primer/spacing */
    padding-inline: var(--control-medium-paddingInline-condensed);

    &:where([data-size='small']) {
    /* stylelint-disable-next-line primer/spacing */
      padding-inline: var(--control-xsmall-paddingInline-condensed);
    }

    &:where([data-size='large']) {
    /* stylelint-disable-next-line primer/spacing */
      padding-inline: var(--control-large-paddingInline-normal);
    }
  }
```

I think this will work for lint but the formatting might be off

packages/react/src/Button/Button.features.stories.tsx
```diff
@@ -61,9 +61,7 @@ export const TrailingCounter = () => {
   )
 }
 
-export const TrailingCounterWithNoText = () => {
-  return <Button leadingVisual={CommentIcon} count={3} />
-}
+export const TrailingCounterWithNoText = () => <Button leadingVisual={CommentIcon} count={3} />
```
langermank, line 64: Should we add an aria-label or a Tooltip to this example?

langermank:
You'll need to remove the snapshots that were added, and if you want add a new VRT test for the feature story 👍

packages/react/src/Button/Button.features.stories.tsx
mperrotti: Hmmm I'm torn on this. I'd prefer not to encourage tooltips, but it's common to use them for icon-only buttons.

Maybe add a second example in the same story?

langermank (approved):
🚀

PR #5492, opened 2025-01-06
feat(react): update range to include support for 19.x

joshblack:
Update package to include support for React 19.x

#### Changed

- Update `peerDependencies` for react packages and types to include support for `19.x` along with `18.x`

### Rollout strategy

- [x] Minor release


packages/react/package.json
```diff
@@ -209,11 +209,11 @@
     "yaml": "2.3.4"
   },
   "peerDependencies": {
-    "@types/react": "^18.0.0",
```
francinelucca, line 212: so the chevron up doesn't include 19.x? I thought it would accept everything above that 🤔

francinelucca (approved):

packages/react/package.json
joshblack: Nope 🙃 haha, at least as far as I understand. `^` would be only for `minor` or `patch`. An alternative operator could be `>=` which I think works exactly how you described

ammunk:
Hi,

Is there a timeline for getting this merged?

Thanks!

Mr-Python-in-China:
Hi! Is there a plan for this work? When might it be merged?

PR #5507, opened 2025-01-08
chore(deps): update storybook dependencies together

joshblack:
Update storybook dependencies together. This was causing an issue in local development.

#### Changed

- Update all storybook dependencies to 8.4.7
- Update dependabot config to use a group for storybook dependencies


francinelucca (approved):
✨

PR #5518, opened 2025-01-09
Update dependabot.yml

joshblack:
Update storybook book to be included in the default `npm` config. This should keep our current behavior (dependabot makes a PR per package) but will now group updates that match the pattern into a storybook group (see: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--:~:text=When%20groups%20is,individual%20pull%20requests.)


francinelucca (approved):
:shipit:

PR #5527, opened 2025-01-10
Improve programmatic state management of UnderlinePanels

ddoyle2017:
Closes [#4575](https://github.com/github/primer/issues/4575)

#### New

- Adds an optional `onSelect()` prop to both `UnderlinePanels` and `UnderlinePanels.Tab`. This allows users to define a function that gets called when a tab is selected.

#### Changed

- Updated `UnderlinePanels` to handle updates to tabs and panels within a `useEffect()` hook. This ensures that the selected tab is updated when the `aria-selected` attribute changes, as well as keeping prop data fresh.

#### Removed

- N/A

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- [WIP] Are there other ways to test these changes outside of unit testing?
- Added unit tests verifying that the `onSelect()` function is called when a tab is clicked + the selected tab can be managed programmatically via updating the `aria-selected` prop

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


packages/react/src/experimental/UnderlinePanels/UnderlinePanels.tsx
```diff
@@ -34,6 +43,10 @@ export type UnderlinePanelsProps = {
    * ID of the element containing the name for the tab list
    */
   'aria-labelledby'?: React.AriaAttributes['aria-labelledby']
+  /**
+   * Callback that will trigger both on click selection and keyboard selection.
+   */
+  onSelect?: (event: React.KeyboardEvent<HTMLButtonElement> | React.MouseEvent<HTMLButtonElement>) => void
```
cmbrose, line 49: Should this just be on `TabProps`?

ddoyle2017: 🤔 I added this at first since it was on the recommended [path](https://github.com/github/primer/issues/4575#issuecomment-2576142196). But you have a point, we're not using it + I'm not sure having a whole panel click handler would be useful. I also looked over [UnderlineNav](https://github.com/primer/react/blob/main/packages/react/src/UnderlineNav/UnderlineNav.tsx) and `onSelect` is only on the tabs/items.

I'm remove it in the next commit 👍

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.tsx
```diff
@@ -199,25 +221,37 @@ const UnderlinePanels: FC<UnderlinePanelsProps> = ({
...
+
+  return (
+    <UnderlineItem
+      as="button"
+      role="tab"
+      tabIndex={ariaSelected ? 0 : -1}
+      aria-selected={ariaSelected}
+      sx={sxProp}
+      type="button"
+      onClick={clickHandler}
```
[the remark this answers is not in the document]
ddoyle2017: Good catch 😅 I checked [UnderlineItem] (https://github.com/primer/react/blob/497e22d2f3af85c6b9b46ff63f44d8df77b0a089/packages/react/src/internal/components/UnderlineTabbedInterface.tsx#L281) and there's no keyboard event handler, so we'll have to add it here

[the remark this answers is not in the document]
ddoyle2017: Added in [35a98c0](https://github.com/primer/react/pull/5527/commits/35a98c08e5974e7cc4ff1f45fc1bf44fbc1d9df8)

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.tsx
ddoyle2017: Fixed in [35a98c0](https://github.com/primer/react/pull/5527/commits/35a98c08e5974e7cc4ff1f45fc1bf44fbc1d9df8)

francinelucca (approved):
Looks great, thanks! Just one small comment on the UnderlinePanels.Tab story

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.Tab.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
```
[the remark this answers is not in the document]
ddoyle2017: It was for testing! I wanted to verify that changing the value of `aria-selected` would visually select/deselect the tab. Prior to this PR, changing the value for the prop wouldn't update the visuals the tabs. 

That prop is only exposed on the parent component, which is why I added a story for the subcomponent 😅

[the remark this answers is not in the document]
francinelucca: Ah I see, perhaps we could make it a .dev story then if we want to keep it around? that way it will show up locally but not in production. Example: https://github.com/primer/react/blob/main/packages/react/src/ActionList/ActionList.dev.stories.tsx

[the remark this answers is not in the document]
ddoyle2017: Makes sense! I'll add that change

mperrotti (approved):
This is great. Thank you!

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.Tab.stories.tsx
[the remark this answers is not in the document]
ddoyle2017: Added in [5f503fc](https://github.com/primer/react/pull/5527/commits/5f503fc12b9a17ef0a199ae5c1b2ce6d424c5f9f)

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.dev.stories.tsx
```diff
@@ -18,3 +19,41 @@ export const Default = () => (
...
+export const NoTabsSelected = () => (
+  <UnderlinePanels aria-label="Select a tab">
+    <UnderlinePanels.Tab aria-selected={false}>Tab 1</UnderlinePanels.Tab>
+    <UnderlinePanels.Tab aria-selected={false}>Tab 2</UnderlinePanels.Tab>
+    <UnderlinePanels.Tab aria-selected={false}>Tab 3</UnderlinePanels.Tab>
+    <UnderlinePanels.Panel>Panel 1</UnderlinePanels.Panel>
+    <UnderlinePanels.Panel>Panel 2</UnderlinePanels.Panel>
+    <UnderlinePanels.Panel>Panel 3</UnderlinePanels.Panel>
+  </UnderlinePanels>
+)
```
ddoyle2017, line 32: I found a bug while adding updating this story that I could use some help with 🤔 So I would expect `NoTabsSelected` to render something like this:
Screenshot 2025-01-22 at 3 49 08 PM

But it seems like `UnderlinePanels` always defaults to underlining the first tab when it first renders, even when `aria-selected` is false for every tab. I'm not sure where in the code this behavior is coming from though

francinelucca: This is coming from [@github/tab-container-element](https://github.com/github/tab-container-element/blob/main/src/tab-container-element.ts#L356-L358). We use it under the hood and looks like it's designed to set tab 0 as default if none is selected. If we want to support this we're going to have to do some hacking to programmatically reset the `aria-selected` value I'm thinking.

But I'm wondering if this is even a case we need to support... @mperrotti do we foresee the need for "no tabs selected"?

ddoyle2017: No tabs selected is something we wanted to support for Copilot Workspace 😅 For example, in [these designs](https://github.com/github/copilot-workspace/issues/64), the initial state of a Workspace has an `UnderlinePanels` at the bottom that's collapsed + has no tabs selected

cmbrose: Honestly I'm not really all that concerned with having this behavior. We can double check with our designer, but I partly think that it makes sense that we have an initially selected tab for the collapsed state so that it's not surprising that that tab becomes active if the user clicks the expand button rather than clicking a tab.

PR #5537, opened 2025-01-13
Deploy preview with VRT/AAT reports (Take Two)

hussam-i-am:
Reverts primer/react#5536

#### Changed

Giving this another shot 🤞🏼 

Added [checks](https://github.com/primer/react/pull/5537/commits/3760650ee77eacb84ee35196f931f30a1298967a) for `deploy_preview_with_reports.yml` to only run if `ci.yml` was triggered by a `pull_request` event. I suspect that it was deploying to production because it was being triggered by the `merge_group` and `push` events.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [x] None; if selected, include a brief description as to why

Only contains workflow changes

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #5543, opened 2025-01-14
Revert "Deploy preview with VRT/AAT reports (Take Two)"

hussam-i-am:
Reverts primer/react#5537

Looks like it is still deploying to production :/


langermank (approved):
Aw man!

PR #5562, opened 2025-01-17
fix(SelectPanel): Correctly recalculate position on overflow

francinelucca:
Closes https://github.com/github/primer/issues/4027

Continuation of https://github.com/primer/react/pull/5207. 

When the selectPanel grows after opening (example: open in loading state, data is fetched, items are rendered), the position is not recalculated and may cause content to go off screen (See for more context https://github.com/github/primer/issues/4027). This PR corrects that undesired behavior by recalculating the position of the overlay on size change. Additionally, bake in logic so that the overlay retains it's size on filtering when anchored to the top (to prevent the panel from repositioning itself too many times).

**before**

https://github.com/user-attachments/assets/a7e2e2b9-e271-40c2-b67c-502b8d8eb1c9

https://github.com/user-attachments/assets/a1e7ce7d-7ab3-4850-89d4-6eb0480707b7

**after**

https://github.com/user-attachments/assets/5a6a7995-6726-433e-b89c-55a5c6144c02

https://github.com/user-attachments/assets/2fd0d76a-e1ab-4177-81e0-a0530a1ba020

#### New

- AnchoredOverlay dev stories for content resizing within page and dialog.
- Support for `pinPosition` prop in `AnchoredOverlay` to prevent position shifting when sitting at the top of the anchor.
- Support 'anchorSide' prop through 'overlayProps' in AnchoredOverlay
- Adds `max-height: 100vh` default to AnchoredOverlay.
- SelectPanel example stories for content resizing within page and dialog.
- Support for `width` overlay prop in `SelectPanel`
- Set `pinPosition` to true in SelectPanel if height hasn't been set by user
- Added resizeObserver to `useAnchoredPosition` to watch for changes within the floating element (i.e., the overlay)

#### Changed

- Add logic to `useAnchoredPosition`'s `updatePosition` function to prevent anchored element to unpin from the top when possible (dependent on `pinPosition`) prop
- Slight refactor in `useResizeObserver` hook
- VRT updates (couldn't determine the source of these)

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Test new functionality in https://primer-88a2e13e51-13348165.drafts.github.io/storybook/?path=/story/components-selectpanel-examples--reposition-after-loading, https://primer-88a2e13e51-13348165.drafts.github.io/storybook/?path=/story/components-selectpanel-examples--select-panel-reposition-inside-dialog

Ensure default functionality still works as expected: https://primer-88a2e13e51-13348165.drafts.github.io/storybook/?path=/story/components-selectpanel--default

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -8,6 +8,7 @@ import useLayoutEffect from '../utils/useIsomorphicLayoutEffect'
 export interface AnchoredPositionHookSettings extends Partial<PositionSettings> {
   floatingElementRef?: React.RefObject<Element>
   anchorElementRef?: React.RefObject<Element>
+  pinPosition?: Boolean
```
francinelucca, line 11: Added this flag so that I'm isolating the new behavior to selectPanel only, I believe we discussed not extending to overlays for now. CC: @camertron

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,49 @@ export function useAnchoredPosition(
   const floatingElementRef = useProvidedRefOrCreate(settings?.floatingElementRef)
   const anchorElementRef = useProvidedRefOrCreate(settings?.anchorElementRef)
   const [position, setPosition] = React.useState<AnchorPosition | undefined>(undefined)
+  const [_, setPrevHeight] = React.useState<number | undefined>(undefined)
```
francinelucca, line 34: used to force the height of the selectPanel when filtered and it wants to shrink but it's anchored at the top (we want it to stay at the top and lock the height)

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,49 @@ export function useAnchoredPosition(
...
     () => {
       if (floatingElementRef.current instanceof Element && anchorElementRef.current instanceof Element) {
-        setPosition(getAnchoredPosition(floatingElementRef.current, anchorElementRef.current, settings))
+        const newPosition = getAnchoredPosition(floatingElementRef.current, anchorElementRef.current, settings)
+        setPosition(prev => {
+          if (
+            settings?.pinPosition &&
+            prev &&
+            ['outside-top', 'inside-top'].includes(prev.anchorSide) &&
+            (prev.anchorSide !== newPosition.anchorSide || prev.top < newPosition.top)
```
francinelucca, line 45: if the selectPanel is anchored to the top and it shrinked in size or wants to change position...

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,49 @@ export function useAnchoredPosition(
...
+          ) {
+            const anchorTop = anchorElementRef.current?.getBoundingClientRect().top ?? 0
+            if (anchorTop > (floatingElementRef.current?.clientHeight ?? 0)) {
```
francinelucca, line 48: if the anchorTop is bigger than the selectPanel's height, that means there is plenty of space for the selectPanel to stay at the top

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,49 @@ export function useAnchoredPosition(
...
+              setPrevHeight(prevHeight => {
+                if (prevHeight && prevHeight > (floatingElementRef.current?.clientHeight ?? 0)) {
+                  requestAnimationFrame(() => {
```
francinelucca, line 51: using requestAnimationFrame here to prevent a resizeObserver loop here due to us changing the element's height, but the observer listening for changes in the element...

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,49 @@ export function useAnchoredPosition(
...
+          if (
+            settings?.pinPosition &&
+            prev &&
+            ['outside-top', 'inside-top'].includes(prev.anchorSide) &&
+            (prev.anchorSide !== newPosition.anchorSide || prev.top < newPosition.top)
+          ) {
+            const anchorTop = anchorElementRef.current?.getBoundingClientRect().top ?? 0
+            if (anchorTop > (floatingElementRef.current?.clientHeight ?? 0)) {
+              setPrevHeight(prevHeight => {
+                if (prevHeight && prevHeight > (floatingElementRef.current?.clientHeight ?? 0)) {
```
francinelucca, line 50: re-establish the height if it used to be bigger and is now attempting to shrink

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,49 @@ export function useAnchoredPosition(
...
+                  requestAnimationFrame(() => {
+                    ;(floatingElementRef.current as HTMLElement).style.height = `${prevHeight}px`
+                  })
+                } else {
+                  prev = newPosition
```
francinelucca, line 55: edgecase where the selectPanel is actually growing instead of shrinking, in this case we do want to allow it to recalculate/reposition

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -8,6 +8,7 @@ import useLayoutEffect from '../utils/useIsomorphicLayoutEffect'
 export interface AnchoredPositionHookSettings extends Partial<PositionSettings> {
   floatingElementRef?: React.RefObject<Element>
   anchorElementRef?: React.RefObject<Element>
+  pinPosition?: boolean
```
francinelucca, line 11: Added this flag so that I'm isolating the new behavior to selectPanel only, I believe we discussed not extending to overlays for now. CC: @camertron

packages/react/src/Overlay/Overlay.module.css
```diff
@@ -30,6 +30,7 @@
 
   &:where([data-reflow-container='true']) {
     max-width: calc(100vw - 2rem);
+    max-height: 100vh;
```
francinelucca, line 33: would love a check from you @TylerJDev over here 🙏

packages/react/src/Overlay/Overlay.tsx
```diff
@@ -75,7 +75,7 @@ const StyledOverlay = toggleStyledComponent(
     min-width: 192px;
     max-width: ${props => props.maxWidth && widthMap[props.maxWidth]};
     height: ${props => heightMap[props.height || 'auto']};
-    max-height: ${props => props.maxHeight && heightMap[props.maxHeight]};
+    max-height: ${props => (props.maxHeight ? heightMap[props.maxHeight] : '100vh')};
```
francinelucca, line 78: would love a check from you @TylerJDev over here 🙏, since I know you've worked quite a bit on overlay sizing issues

TylerJDev: Just checking, is the intention here so that it's responsive, or is there something else I should be looking at? Want to make sure I'm looking at the right thing 🤔

francinelucca: Ideally we want people to use the `height` and `width` parameters to limit the size of the panel. But I just wanted to add a safeguard that at most it'll grow to be the screen size. So a scroll bar is added otherwise, right now it's just jumping out of the screen in those cases. I guess my question is, do you foresee any issues as a result of adding this?

.playwright/snapshots/components/Overlay.test.ts-snapshots/Overlay-SX-Props-dark-high-contrast-linux.png
francinelucca, line 1: No idea what these VRT changes are about. From the pictures it looks like the focus outline color is changing but I tried to reproduce locally and have been unable to (this is a dev only story). It also looks like the new outline more closely matches production 🤷‍♀️

francinelucca: I really don't know, it was failing on both flags (on and off)

packages/react/src/hooks/useResizeObserver.ts
```diff
@@ -59,5 +59,5 @@ export function useResizeObserver<T extends HTMLElement>(
     }
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [target, ...depsArray])
+  }, [target?.current, ...depsArray])
```
francinelucca, line 62: I know you originally moved line 25 outside of the hook and put the dependency on the `targetEl` but that was generating SSR issues due to document not being defined, was wondering how you felt about this alternative instead `@siddharthkp`

.playwright/snapshots/components/Overlay.test.ts-snapshots/Overlay-SX-Props-dark-high-contrast-linux.png
francinelucca, line 1: RE: https://github.com/primer/react/pull/5562#discussion_r1943056791

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,50 @@ export function useAnchoredPosition(
...
     () => {
       if (floatingElementRef.current instanceof Element && anchorElementRef.current instanceof Element) {
-        setPosition(getAnchoredPosition(floatingElementRef.current, anchorElementRef.current, settings))
+        const newPosition = getAnchoredPosition(floatingElementRef.current, anchorElementRef.current, settings)
+        setPosition(prev => {
+          if (
+            settings?.pinPosition &&
+            prev &&
+            ['outside-top', 'inside-top'].includes(prev.anchorSide) &&
+            (prev.anchorSide !== newPosition.anchorSide || prev.top < newPosition.top)
```
camertron, line 46: Could we wrap this in a function and give it some comments so we know what's going on when we come back to the code in a few months?

packages/react/src/hooks/useAnchoredPosition.ts
```diff
@@ -30,22 +31,50 @@ export function useAnchoredPosition(
...
+          ) {
+            const anchorTop = anchorElementRef.current?.getBoundingClientRect().top ?? 0
+            if (anchorTop > (floatingElementRef.current?.clientHeight ?? 0)) {
```
camertron, line 49: Same question: maybe wrap in a function with a nice, intention-revealing name and some comments?

camertron (changes requested):
Wow, this is really good stuff, nice work! 🎉

I noticed a flicker in one of the after videos where the panel shrinks but then immediately grows again. What do you think it would take to remove the flicker?

emilybrick (approved):
Reviewing on behalf on design here - 👍

packages/react/src/Overlay/Overlay.tsx
TylerJDev: I think this should be good! I didn't see any issues with it when I tested.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -112,11 +116,12 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
   overlayProps,
   focusTrapSettings,
   focusZoneSettings,
-  side = 'outside-bottom',
+  side = overlayProps?.['anchorSide'] || 'outside-bottom',
   align = 'start',
   alignmentOffset,
   anchorOffset,
   className,
+  pinPosition,
```
TylerJDev, line 124: Should we add this to `AnchoredOverlay.docs.json`?

francinelucca: Yes! Done!!

packages/react/src/hooks/useAnchoredPosition.ts
francinelucca: Done!

packages/react/src/hooks/useAnchoredPosition.ts
francinelucca: Done!

francinelucca:
@camertron cleaned up the `useAnchoredPosition` logic and split into functions, lmk how that looks.
As for the flashing, I tried a couple things with the visibility like you suggested but couldn't get it to bypass it. I think visibility won't help in this case because the flashing happens when the SelectPanel is already in view, user filter so SelectPanel shrinks, and then we re-establish the height when the `useAnchoredPosition` hooks runs again as a response to the element resizing

camertron (approved):

PR #5559, opened 2025-01-16
[Accessibility][Storybook] Add aria-labels to the multiple item progress bar + enforce aria-label in progressbar.item

kendallgassner:
Closes https://github.com/github/accessibility-audits/issues/10134

In this PR I add an aria-label to each role='progressbar'. I also have updated the ProgressBar.item props to enforce that aria-labels are always added

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

I am updating a storybook story.

### Testing & Reviewing

Tested with a screen reader.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [X] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #5580, opened 2025-01-21
Run storybook axe scans on CI builds

kendallgassner:
We noticed in https://github.com/primer/react/pull/5559 that our axe-scans do not run on every storybook story. This PR aims to fix that.

Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack:
Hi here, @kendallgassner! 👋 

Just providing some context, storybook axe scans currently run through playwright ([for example](https://github.com/primer/react/blob/main/e2e/components/Banner.test.ts#L92)) instead of in Jest so that we can get contrast checks reported accurately. We still have some old test suites that use this technique but we can get rid of them (I think 😅)

kendallgassner:
👀 Thanks @joshblack! Is there a reason we don't do this for every story?

joshblack:
@kendallgassner I don't think so, definitely would make sense to run it on every story. It would be worth it to add a dedicated test for this so that each component e2e test doesn't have to add it explicitly which I think is likely the root cause if something is not being checked.

kendallgassner:
@joshblack cool! I'll convert this PR to aim to do that

packages/react/src/Button/ButtonBase.tsx
```diff
@@ -72,7 +72,8 @@ const ButtonBase = forwardRef(
         if (
           innerRef.current &&
           !(innerRef.current instanceof HTMLButtonElement) &&
-          !((innerRef.current as unknown) instanceof HTMLAnchorElement)
+          !((innerRef.current as unknown) instanceof HTMLAnchorElement) &&
+          !((innerRef.current as HTMLElement).tagName === 'SUMMARY')
```
kendallgassner, line 76: I don't know how I feel about this but the [details story](https://primer.style/react/storybook/?path=/story/components-details--default) feels like it's incorrectly erroring here.

kendallgassner:
Closing for now while I play around with this locally.

PR #5581, opened 2025-01-21
cleanup: remove deprecated underline attribute

khiga8:
Small change to clean up deprecated `underline` attributes!

I was getting bothered in storybook files that this `underline` attribute is deprecated, but it's used in many examples.

Let's clean this up so developers don't add undesirable usage in their code when copying examples from Primer's storybook.

#### Removed

* Removed undesirable `underline` attribute usage in stories.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/AnchoredOverlay/AnchoredOverlay.features.stories.tsx
+++ b/packages/react/src/AnchoredOverlay/AnchoredOverlay.features.stories.tsx
@@ -26,7 +26,7 @@ const hoverCard = (
     <Stack direction="horizontal" gap="none">
       <Text weight="medium">monalisa</Text>
       <Text color={'var(--fgColor-muted)'} ml={1}>
-        <Link inline underline muted href="#">
+        <Link inline muted href="#">
           Monalisa Octocat
         </Link>
       </Text>
--- a/packages/react/src/AnchoredOverlay/AnchoredOverlay.stories.tsx
+++ b/packages/react/src/AnchoredOverlay/AnchoredOverlay.stories.tsx
@@ -23,7 +23,7 @@ const hoverCard = (
     <Stack direction="horizontal" gap="none">
       <Text weight="medium">monalisa</Text>
       <Text color={'var(--fgColor-muted)'} ml={1}>
-        <Link inline underline muted href="#">
+        <Link inline muted href="#">
           Monalisa Octocat
         </Link>
       </Text>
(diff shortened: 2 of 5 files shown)


TylerJDev (commented):
Since we're using `inline`, do we need to add `data-a11y-link-underlines="true"` to the containers of the links so that the underlines will show? I did [something similar in another PR](https://github.com/primer/react/pull/5577/files#diff-7baae9a5c38b80141097370d1bb92ffe2a872ead56446fafb2ef087d4a8f3564R25), but now I'm wondering if there's some sort of base Storybook component we could add this to 🤔 

Edit: We might be able to add the `data-*` attribute to [the `<div>` in `preview.jsx`](https://github.com/primer/react/blob/main/packages/react/.storybook/preview.jsx#L308), if we did go that route!

khiga8:
> Edit: We might be able to add the data-* attribute to [the <div> in preview.jsx](https://github.com/primer/react/blob/main/packages/react/.storybook/preview.jsx#L308), if we did go that route!

@TylerJDev I think this route makes sense! We have something [similar in dotcom](https://github.com/github/github/blob/995c5177f4a0c582c504d249f26dabad0adc88a2/app/views/layouts/component_preview.html.erb#L1)!

packages/react/.storybook/preview.jsx
```diff
@@ -271,6 +270,14 @@ export const decorators = [
       document.body.setAttribute('data-dark-theme', darkTheme)
     }, [colorScheme])
 
+    // Set data-a11y-link-underlines=true to enable underlines in all stories except the Link DevOnly Inline Story.
```
khiga8, line 273: The Link Devonly inline story has examples to show what impact this underline data attribute has. 

Adding this conditional to avoid ending up with this markup on that particular story, because it results in all links being underlined.=:

```.tsx
<div data-a11y-link-underlines="true">
  ... 
  ...
  <div data-a11y-link-underlines="false">
     Learn more <Link href="#">about GitHub</Link>
  </div>
```

TylerJDev (approved):
Looks great! Thank you for all the work on this ✨

PR #5583, opened 2025-01-21
Banner allows critical variant to use ondismiss

### LINKED ISSUE #5582: Allow Banner to use `onDismiss`
### Description

Currently, the Banner component doesn't allow onDismiss to be used if the critical variant is used.

We would like to be able to use the onDismiss button when using the critical variant.

### Steps to reproduce

Add a Banner component and configure variant as `critical` and onDismiss as non null.

### Version

v37.9.1

### Browser

_No response_

brphelps:
Closes #5582

#### Changed

Banner now supports onDismiss used when using variant `critical`. 

### Rollout strategy

- [X] Patch release
- [ ] Minor release

### Testing & Reviewing

I added a unit test and updated the example to use onDismiss.

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.changeset/blue-actors-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
```
joshblack, line 2:
```suggestion
"@primer/react": minor
```

Suggesting `minor` here since we're adding in functionality

packages/react/src/Banner/Banner.features.stories.tsx
```diff
@@ -25,6 +25,7 @@ export const Critical = () => {
           to enable two-factor authentication as an additional security measure.
         </>
       }
+      onDismiss={action('onDismiss')}
```
joshblack, line 28: From a docs perspective we might not want this to be visible in the default story for critical, I think the idea was that it would be rare for critical banners to be dismissible since you must address them but I'm curious what @lukasoppermann would say 👀

joshblack (commented):
Thanks so much for putting this together! Just left a couple of comments 👍

packages/react/src/Banner/Banner.features.stories.tsx
lukasoppermann: Agree, @joshblack and I discussed it and we would like to leave the the critical story as is, without the dismiss. Users can still get to it in the playground.

If you could to this change @brphelps, we can merge this. 

Thank you for this PR. 🙏

brphelps: @joshblack Alright, took care of that. Thanks!

joshblack (approved):

PR #5584, opened 2025-01-21
refactor(ButtonBase): update selector to have specificty 0,1,0

joshblack:
Closes https://github.com/github/primer/issues/4703

#### Changed

- Update `.ButtonBase` to have specificity of at least `0,1,0` to no longer get overridden by reset styles or equivalent selectors with `0,1,0` specificity

### Rollout strategy

- [x] Patch release


packages/react/src/Button/ButtonBase.module.css
```diff
@@ -98,7 +98,7 @@
 
   /* Button layout */
 
-  & .ButtonContent {
+  & :where(.ButtonContent) {
```
langermank, line 101: I suspect `.ButtonContent` and `.Label` don't need `where` 🤔 just a variant of `ButtonBase` right?

joshblack: @langermank honestly just did this to keep the specificity the same but have no idea if it's needed 🤷 

Specifically right now this would be: `:where(.ButtonBase) .ButtonContent` at `0,1,0` and without the `:where` it would be `0,2,0` with `.ButtonBase .ButtonContent`.

Would it be helpful if I did not include `:where()` for descendant selectors unless we see a problem during testing?

langermank: Ahh okay, no, I think that's fair. Carry on!

mperrotti (approved):
Used the preview release to test in `primer-docs`, and this works as expected 🎉

langermank (approved):
🚀

PR #5590, opened 2025-01-23
Reintroduce Pagination algorithm enhancements

hectahertz:
Reverts primer/react#5560

The original cause of the failing tests was that changing `aria-disabled` with `aria-hidden` triggered the contrast rules.

Here we add `aria-disabled` back.


hectahertz:
Depends on https://github.com/github/github/pull/359662

francinelucca (approved):
✨

PR #5592, opened 2025-01-23
[Axe] Run axe across all stories

kendallgassner:
In follow up PR:
- [ ] Remove old `@aat` tests

This PR aims to run axe ci testing on all Storybook stories. I noticed in a previous PR that we had uncaught axe violations because we have to manually add a Playwright test for each Storybook story. Because of this manual effort many of our stories do not have axe testing setup.

In addition to writing a script to run axe against all stories I made fixes to all pre-existing stories with axe failures.  

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -217,9 +217,8 @@ const Anchor = React.forwardRef<HTMLElement, ActionMenuAnchorProps>(({children:
 })
 
 /** this component is syntactical sugar 🍭 */
-export type ActionMenuButtonProps = Omit<ButtonProps, 'children'> & {
-  children: React.ReactNode
-}
+export type ActionMenuButtonProps = ButtonProps
```
kendallgassner, line 220: Fixes: `Buttons must have discernible text [button-name]`

[for components-buttongroup-features--dropdown-split ](https://primer.style/react/storybook/?path=/story/components-buttongroup-features--dropdown-split)

packages/react/src/Button/ButtonBase.tsx
```diff
@@ -72,7 +72,8 @@ const ButtonBase = forwardRef(
         if (
           innerRef.current &&
           !(innerRef.current instanceof HTMLButtonElement) &&
-          !((innerRef.current as unknown) instanceof HTMLAnchorElement)
+          !((innerRef.current as unknown) instanceof HTMLAnchorElement) &&
+          !((innerRef.current as HTMLElement).tagName === 'SUMMARY')
```
kendallgassner, line 76: Removes console.warn for [components-details--default](https://primer.style/react/storybook/?path=/story/components-details--default)

.github/workflows/ci.yml
```diff
@@ -188,7 +222,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 22
-          cache: 'npm'
+          cache: "npm"
       - name: Install dependencies
         run: npm ci
       - name: Build storybook
```
joshblack, line 228: Since we're building storybook in this step, I believe you can use `packages/react/storybook-static/index.json` to get a list of all stories (if I remember the path right 😅) so that we don't have to depend on `build-components-json`. This also gives access to some meta information about the story (in particular tags) if that would be helpful for changing behavior of the axe test (like making sure to run access both before and after opening a menu)

packages/react/src/ActionMenu/ActionMenu.examples.stories.tsx
```diff
@@ -410,15 +410,15 @@ export const MultipleSections = () => {
       </ActionMenu.Anchor>
       <ActionMenu.Overlay width="small">
         <ActionList>
-          <ActionList.Group selectionVariant="multiple">
+          <ActionList.Group>
```
kendallgassner, line 413: Fixes: `Required ARIA attributes must be provided` for [Multiple Sections story](https://primer.style/react/storybook/?path=/story/components-actionmenu-examples--multiple-sections)

packages/react/src/ButtonGroup/ButtonGroup.features.stories.tsx
```diff
@@ -73,7 +73,7 @@ export const InactiveButtonsGroup = () => {
   )
 
   const secondaryButton = (
-    <IconButton aria-label="Secondary Button Aria Label" aria-disabled={true} inactive={true} icon={DashIcon} />
+    <IconButton aria-label="Secondary Button" aria-disabled={true} inactive={true} icon={DashIcon} />
```
kendallgassner, line 76: Cleans up bad `aria-label` and ensures that either `aria-label` or `aria-labelledby` is provided on the IconButton (not both).

packages/react/src/FilteredActionList/FilteredActionList.stories.tsx
```diff
@@ -57,7 +58,7 @@ export function Default(): JSX.Element {
   const filteredItems = items.filter(item => item.text.toLowerCase().startsWith(filter.toLowerCase()))
 
   return (
-    <>
+    <FeatureFlags flags={{primer_react_select_panel_with_modern_action_list: true}}>
```
kendallgassner, line 61: Flagging in this feature cleans up [2 axe violations](https://primer.style/react/storybook/?path=/story/components-filteredactionlist--default):
- ARIA input fields must have an accessible name
- Certain ARIA roles must contain particular children

packages/react/src/Header/Header.dev.module.css
```diff
@@ -1,11 +1,12 @@
 .HeaderDev {
-  background-color: var(--label-olive-bgColor-active);
+  background-color: var(--display-pine-bgColor-emphasis);
```
kendallgassner, line 2: Cleans up inaccessible color contrast

packages/react/src/Header/Header.dev.stories.tsx
```diff
@@ -22,7 +22,7 @@ export const WithCss = () => (
       primer_react_css_modules_ga: true,
     }}
   >
-    <Header as="summary" className={classes.HeaderDev}>
+    <Header className={classes.HeaderDev}>
```
kendallgassner, line 25: Cannot have nested interactive elements.

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.stories.tsx
```diff
@@ -16,11 +16,6 @@ const meta: Meta<typeof UnderlinePanels> = {
         name: 'string',
       },
     },
```
kendallgassner, line 18: We do not need `aria-labelledby` and `aria-label` controls

packages/react/src/stories/deprecated/ActionMenu.stories.tsx
```diff
@@ -233,7 +231,7 @@ export function ComplexListStory(): JSX.Element {
 ComplexListStory.storyName = 'Complex List'
 
 export function CustomTrigger(): JSX.Element {
-  const customAnchor = (props: LinkProps) => <Link {...props} sx={{cursor: 'pointer'}} />
+  const customAnchor = (props: ButtonProps) => <Button {...props} sx={{cursor: 'pointer'}} />
```
kendallgassner, line 234: No nested interactive elements.

kendallgassner, line 234: Elements must only use supported ARIA attributes. Cannot have `aria-expanded` on link.

packages/react/src/Banner/Banner.stories.tsx
```diff
@@ -49,6 +50,7 @@ export const Playground: StoryObj<typeof Banner> = {
 
         <PageLayout.Content>
           <Banner
+            aria-label="Content level banner"
```
kendallgassner, line 53: Unique aria-labels!

kendallgassner:
In follow up PRs:
- work to remove individual aat tests.
- consider automating VRT tests.

kendallgassner:
✅ Integration failures seem to be unrelated to this PR. 

TODO:

-[ ] Once this [revert](https://github.com/primer/react/pull/5612) is merged add an aria-label to the `MultipleItems` ProgressBar story as this will fail axe tests.

e2e/components/Axe.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  'components-treeview-features--contain-intrinsic-size',
+]
+
+type Component = {
+  name: string
+}
+
+const {entries} = componentsConfig
+
+test.describe('@aat', () => {
```
joshblack, line 22:
```suggestion
test.describe('aat', () => {
```

I would just annotate this as `aat`, or remove it, since we're using the `@aat` tag at the test-level to filter things out currently (although we could totally change this in the future.

e2e/components/Axe.test.ts
```diff
@@ -0,0 +1,49 @@
...
+                  colorScheme: themes[0],
+                },
+              })
+              await expect(page).toHaveNoViolations()
+            }
+          })
+        })
+      }
+    })
+  }
```
joshblack, line 48: Just a couple of suggestions, let me know what you think!

- Skip tests at the outermost loop level instead of having empty tests
- Update `colorScheme` to point  to `theme`
- Add `@aat` tag at the test-level

```suggestion
  for (const [id, entry] of Object.entries(entries as Record<string, Component>)) {
    if (SKIPPED_TESTS.includes(id)) {
      continue;
    }

    const {name} = entry
    // remove parentheses and slashes from the name to avoid playwright file issues
    // eslint-disable-next-line no-useless-escape
    const cleanedName = name.replaceAll(/[\(\)\/]/g, '')

    test.describe(id, () => {
      for (const theme of themes) {
        test.describe(theme, () => {
          test(`${cleanedName} @aat`, async ({page}) => {
            await visit(page, {
              id,
              globals: {
                colorScheme: theme,
              },
            })
            await expect(page).toHaveNoViolations()
          })
        })
      }
    })
  }
```

joshblack (commented):
Looks great! Just left a couple of comments 👀

e2e/components/Axe.test.ts
kendallgassner: My thought was to remove all `@aat` tests after this is merged since we don't need to generate them ourselves

kendallgassner: @joshblack do you feel like we need the other @aat tests now?

joshblack: @kendallgassner I totally bet we could, especially the automated/boilerplate ones.

I think it can still be helpful to have this tag since there are stories where we would like to put the component in a specific state and then run axe, ideally. Like with menus to make sure we test things when it's open and closed or if we want to use the `matrix` test helper to make sure we run axe against prop combinations of things.

kendallgassner: If we keep the '@aat' in this file, doesn't it break up the tests to run concurrently in [aat-report](https://github.com/primer/react/actions/runs/13019158341/job/36315594252?pr=5592). I assumed aat-runner 1-4 were different threads.

kendallgassner:
A legit color-contrast failure: [flash banner with a link in it](https://primer-191c780ed4-13348165.drafts.github.io/storybook/?path=/story/components-flash-features--with-icon-action-dismiss&globals=colorScheme:light_high_contrast).

packages/react/src/FilteredActionList/FilteredActionList.stories.tsx
khiga8: Since the default storybook reflects what's currently GA-ed (with the option to toggle on the feature flag in the storybook toolbar), I don't think we should add this feature flag here.

Maybe we can exclude this violation for now with a note that the fix is behind a feature flag?

TylerJDev:
Random question; I see the flow isn't part of the workflow `ci.yml` anymore, and was wondering if this was intentional? Just trying to see where the tests are being ran in the CI 👀

kendallgassner:
@TylerJDev I didn't have to change the CI because @joshblack pointed out that I could use the [storybook-static/index.json](https://github.com/primer/react/pull/5592/files#diff-f6520f6c13c3e6344a4bad43e3687433ad97c4a7b6aac74d5213200846fec292R3) file to pull storybook ids ✨ . This file is generated during [npx storybook build](https://github.com/primer/react/blob/main/.github/workflows/aat-reports.yml#L31C14-L31C33) which is already run during the aat report!

The aat report will run these specific tests because of the `@aat` tag.

TylerJDev (approved):

e2e/components/Axe.test.ts
joshblack: @kendallgassner I think we should definitely keep `@aat` in the file for the reasons you mentioned, just was saying that we could add it at the test-level instead of in the `describe` (which seems like the case on line 38? let me know if I'm misreading 👀)

If we have that in place then I think we can remove the outer `describe` block or replace it with just `describe('aat')`

PR #5593, opened 2025-01-24
[SelectPanel] Fix experimental SelectPanel anchoring behavior

camertron:
Closes #

#### Changed

The experimental `SelectPanel` component draws itself off the screen in certain scenarios, i.e. when the anchor element is on the right-hand side of the viewport.

Before:

Alt: a video of a `SelectPanel` story demonstrating how, when opened, the panel causes the viewport to grow off the screen.

https://github.com/user-attachments/assets/a42999cc-c064-4b73-ad21-9013bf1f756c

After:

Alt: a video of a `SelectPanel` story demonstrating that the panel now re-orients itself into the available space as expected.

https://github.com/user-attachments/assets/6cd11c98-df91-4208-ac5a-f9475fa5bb6d

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

~- [ ] Added/updated tests~
~- [ ] Added/updated previews (Storybook)~
~- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)~
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/experimental/SelectPanel2/SelectPanel.tsx
```diff
@@ -309,8 +312,13 @@ const Panel: React.FC<SelectPanelProps> = ({
                 '--max-height': maxHeightValue,
                 '--position-top': `${position?.top ?? 0}px`,
                 '--position-left': `${position?.left ?? 0}px`,
+                visibility: internalOpen ? 'visible' : 'hidden',
+                display: 'flex',
```
francinelucca, line 316: if we could find a way to put this in the SelectPanel.module.css file I think that'd be preferred. Could default to 'visible' and add a [data-visibility-hidden] that hides it

packages/react/src/experimental/SelectPanel2/SelectPanel.tsx
```diff
@@ -309,8 +312,13 @@ const Panel: React.FC<SelectPanelProps> = ({
...
               } as React.CSSProperties)
-            : undefined
+            : {
+                visibility: internalOpen ? 'visible' : 'hidden',
+                display: 'flex',
+              }
```
francinelucca, line 321: and these I think could go inside the sx?

francinelucca (approved):
Looking good! some non-blocking suggestions

packages/react/src/experimental/SelectPanel2/SelectPanel.tsx
camertron: I tried that, but the specificity wasn't high enough to override the default browser stylesheet, which sets `display` to `none`. We basically have no choice but to set this in the `style` attribute 😬

packages/react/src/experimental/SelectPanel2/SelectPanel.tsx
camertron: Unfortunately they have to be set as the `style` attribute to override the browser's default stylesheet 😢

PR #5595, opened 2025-01-25
fix(useResizeObserver): undefined variable alternative

francinelucca:
Closes https://github.com/github/primer/issues/4685

We were seeing a small number of errors for undefined `ResizeObserver`, introducing a check for when `ResizeObserver` is not defined and using a manual solution instead.

#### Changed

- Detect when ResideObserver is not defined and use manual solution instead

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


camertron (approved):
LGTM 🎉 Looks like there might be a few cases where we don't need to check for `null`/`undefined` tho?

PR #5606, opened 2025-01-27
fix(BaseStyles): remove unknown props from fallback when feature flag is enabled

joshblack:
Update how we pass along system props in BaseStyles when the CSS Modules flag is enabled. Currently, when the flag is enabled we are still passing along `color`, `fontFamily`, and `lineHeight` to the underlying element. This change updates the component to use CSS Custom Properties as a bridge between what is provided as props and what is applied to the element in CSS.

#### Changed

- Update BaseStyles to set common styles using CSS Custom Properties instead of as attributes on the element
- Update `FeatureFlags` to merge scope with parent context value

### Rollout strategy

- [x] Patch release


jonrohan (approved):

francinelucca (approved):
Noted the weight of the item titles changed in the VRT updates, intended?

joshblack:
@francinelucca thanks for calling that out! Went to double-check and it seems like yes 👍 The current snapshots do not seem to be using the modern action list and it's probably because of the same bug with feature flags that I ran into 🤔

PR #5626, opened 2025-01-30
feat: add keybindingDescription exploration

joshblack:
@ericwbailey observed that in IconButton's that use a tooltip and a keybinding hint that the computed label was including extra information. This seems due to the fact that we're constructing the keybinding hint using `aria-hidden` but this is not respected when `aria-labelledby` or `aria-describedby` is used.

This PR is an exploration into what it would look like to add a way to provide an appropriate description through `keybindingDescription` but it would be worth seeing if we can make this work with `KeybindingHint` directly.

#### Changed

- Add conditional to when the loading describedby is provided in `ButtonBase`
- Update Tooltip so that only the inner context is pointed to by `aria-labelledby` or `aria-describedby`
- Add `keybindingDescription` to see what this would look like to provide an appropriate description for the keybindingHint

### DIFF
--- a/packages/react/src/Button/ButtonBase.tsx
+++ b/packages/react/src/Button/ButtonBase.tsx
@@ -196,11 +196,11 @@ const ButtonBase = forwardRef(
               </>
             )}
           </Box>
-          {loading && (
+          {loading ? (
             <VisuallyHidden>
               <AriaStatus id={loadingAnnouncementID}>{loadingAnnouncement}</AriaStatus>
             </VisuallyHidden>
-          )}
+          ) : null}
         </ConditionalWrapper>
       )
     }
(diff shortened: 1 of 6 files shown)


ericwbailey:
A little more context, if it's helpful. This work is in response to https://github.com/github/accessibility/issues/8018, which was discovered with [a usability testing session with TetraLogical](https://github.com/github/accessibility/issues/7868).

There are two user-facing concerns we're trying to address:

- The lack of a pause between the button's accessible name/action (ex: "Save") and its companion description (ex: "Control S").
- Assistive technology will read glyphs literally, so content like `⇧` will be announced as "Upwards White Arrow Commercial".

This work addresses those concerns by allowing for visually maintaining the design, while also providing a mechanism for an author to override the auto-generated string with something that makes more sense for a user to hear.

iansan5653:
Wrote up some findings and suggestions + more context in https://github.com/github/accessibility/issues/8018#issuecomment-2751656621

PR #5629, opened 2025-01-30
Make Dialog footer scrollable on very short viewports

hectahertz:
Part of the work to improve reflow of the component. This helps to make all the footer buttons accessible on small screens and gives some extra breathing room to the body.

#### Changed

Dialog footer becomes scrollable on very short viewports.

Before

https://github.com/user-attachments/assets/4e34557c-b346-48dd-bf44-d05284029d2b

After

https://github.com/user-attachments/assets/409f99e0-e209-4cf2-8c3c-97a81bfd389b

### Rollout strategy

- [X] Patch release
- [ ] Minor release

### Testing & Reviewing

The CSS changes seem a bit esoteric, but are due to `overflow: scroll` not being compatible with `justify-content: flex-end`. `flex-direction: row-reverse; flex-wrap: wrap-reverse;` plus reversing the items is visually equivalent but compatible. The focus order keeps being the same.

### Merge checklist

- [ ] Added/updated tests
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack (approved):

packages/react/src/Dialog/Dialog.module.css
```diff
@@ -244,10 +244,15 @@
   z-index: 1;
   display: flex;
   padding: var(--base-size-16);
-  /* stylelint-disable-next-line primer/box-shadow */
+  /* stylelint-disable-next-line primer/box-shadow*/
   box-shadow: 0 -1px 0 var(--borderColor-default);
-  flex-flow: wrap;
-  justify-content: flex-end;
+  flex-direction: row-reverse;
+  flex-wrap: wrap-reverse;
```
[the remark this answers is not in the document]
hectahertz: They shouldn't be! Let's review this together

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -626,12 +626,16 @@ const StyledFooter = toggleStyledComponent(
...
     display: flex;
-    flex-flow: wrap;
-    justify-content: flex-end;
+    flex-direction: row-reverse;
+    flex-wrap: wrap-reverse;
     gap: ${get('space.2')};
     z-index: 1;
     flex-shrink: 0;
 
+    @media (max-height: 256px) {
```
TylerJDev, line 635: I'm wondering if we might want to increase this a bit, maybe by 50-100px? I noticed that on some screen sizes, the body isn't visible. I think increasing it a bit would fix it. 

The only potential issue is if a consumer adds even more buttons, which could mean that even an increase of the `max-height` wouldn't help. I don't think this is an issue the component should solve, as we wouldn't want consumers adding dozens of buttons in a footer 😅

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -626,12 +626,16 @@ const StyledFooter = toggleStyledComponent(
...
+      flex-wrap: nowrap;
+      overflow-x: scroll;
+    }
```
TylerJDev, line 638:
```suggestion
    flex-flow: wrap;
    justify-content: flex-end;
    gap: ${get('space.2')};
    z-index: 1;
    flex-shrink: 0;

    @media (max-height: 356px) {
      flex-wrap: nowrap;
      overflow-x: scroll;

      flex-direction: row;
      justify-content: unset;
    }
```

Curious what you think about this change? I think this also means we don't have to utilize `reverse()` on the footer buttons.

TylerJDev (commented):
Left a few more comments, curious on what you think! Looks almost good to go 🚀

lesliecdubs:
Ping @hectahertz, would be awesome to get this one over the line before you fully switch focus to the next epic 🙇🏻‍♀️

packages/react/src/Dialog/Dialog.module.css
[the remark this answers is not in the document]
hectahertz: Yeah I'm not a fan of changing that one, might introduce bugs on other usages too. I'll follow the other approach you suggested instead!

packages/react/src/Dialog/Dialog.tsx
hectahertz: I like it, it doesn't wrap the same way but I think it's valid and it is simpler!

PR #5662, opened 2025-02-05
[Accessibility][Autocomplete] Announce "no selectable options"

kendallgassner:
Closes #

Closes: https://github.com/github/accessibility-audits/issues/10123

Use the `Announce` component to announce "no selectable options" when a user filters options and there are no results in the Autocomplete component. 

### Rollout strategy

- [X] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [X] Tested in Chrome (NVDA)
- [ ] Tested in Firefox
- [X] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Autocomplete/AutocompleteMenu.tsx
+++ b/packages/react/src/Autocomplete/AutocompleteMenu.tsx
@@ -1,4 +1,5 @@
 import React, {useContext, useEffect, useMemo, useRef, useState} from 'react'
+import {Announce} from '../live-region'
 import {scrollIntoView} from '@primer/behaviors'
 import type {ScrollIntoViewOptions} from '@primer/behaviors'
 import type {ActionListItemProps} from '../ActionList'
@@ -381,9 +382,9 @@ function AutocompleteMenu<T extends AutocompleteItemProps>(props: AutocompleteMe
             </ActionList>
           ) : emptyStateText !== false && emptyStateText !== null ? (
             enabled ? (
-              <Box className={classes.EmptyStateWrapper}>{emptyStateText}</Box>
+              <Announce className={classes.EmptyStateWrapper}>{emptyStateText}</Announce>
             ) : (
-              <Box p={3}>{emptyStateText}</Box>
+              <Announce p={3}>{emptyStateText}</Announce>
             )
           ) : null}
         </div>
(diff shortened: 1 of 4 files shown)


packages/react/src/Autocomplete/AutocompleteMenu.tsx
```diff
@@ -381,9 +382,9 @@ function AutocompleteMenu<T extends AutocompleteItemProps>(props: AutocompleteMe
             </ActionList>
           ) : emptyStateText !== false && emptyStateText !== null ? (
             enabled ? (
-              <Box className={classes.EmptyStateWrapper}>{emptyStateText}</Box>
+              <Announce className={classes.EmptyStateWrapper}>{emptyStateText}</Announce>
```
kendallgassner, line 385: Need to use `LiveRegion` or `Debounce` because this does not provide feedback if a user continues to type.

packages/react/src/Autocomplete/AutocompleteMenu.tsx
```diff
@@ -267,6 +277,12 @@ function AutocompleteMenu<T extends AutocompleteItemProps>(props: AutocompleteMe
     allItemsToRenderRef.current = allItemsToRender
   })
 
+  React.useEffect(() => {
+    if (allItemsToRender.length === 0) {
+      debounceAnnouncement(emptyStateText as string)
+    }
+  }, [allItemsToRender])
+
```
kendallgassner, line 285: ### Why not `Announce`
I chose not to use the `Announce` component because it does not re-render (aka re-announce) if TextContent does not change. 

#### Why does this matter?

Imagine my autocomplete contains 
`Banana`, `Apple`, `Pear`

if a user types `aa`, `no selectable options` should be announced. Then if a user types another `a` making the input read `aaa`, `no selectable options` should be announced again. 

If a user types `bb`, `no selectable options` should be announced. Then if the user presses `CMD + A` to highlight all text and then `CMD + C` to paste new text input we should hear `no selectable options` announced.

kendallgassner: ### Why Debounce

I had to pair this functionality with debounce so that when a user types rapidly, we do not hear too much feedback

camertron (approved):
Seems ok to me!

PR #5666, opened 2025-02-06
ActionBar: Add `disabled` support to `ActionBar`

TylerJDev:
https://github.com/github/primer/issues/4613

#### Changed

* Improves `disabled` support on `ActionBar.IconButton`
* Ensures that the overflow menu respects the `disabled` state

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/ActionBar/ActionBar.tsx
+++ b/packages/react/src/ActionBar/ActionBar.tsx
@@ -252,22 +252,6 @@ export const ActionBar: React.FC<React.PropsWithChildren<ActionBarProps>> = prop
   )
 }
 
-interface RepoPickerProps {
-  id: string
-  name: string
-  ownerLogin: string
-  visibility: 'public' | 'private' | 'internal'
-  selected?: boolean
-}
-
-// {
-//     id: string, required: false
-//     name: string, required: true
-//     ownerLogin: string, required: true
-//     visibility: 'public' | 'private' | 'internal', required: true
-//     selected: boolean, required: false
-// }
-
 export const ActionBarIconButton = forwardRef(
   ({disabled, onClick, onKeyDown, ...props}: ActionBarIconButtonProps, forwardedRef) => {
     const backupRef = useRef<HTMLElement>(null)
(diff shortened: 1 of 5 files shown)


TylerJDev:
@andrejusk, quick implementation question! Are you expecting the IconButton not to be actionable when disabled, or are you handling this in your implementation? Right now we're using `aria-disabled`, which is more accessible and follows the same pattern we use in the overflow menu. This means that users can still click/trigger the button even when disabled. We could block `onClick` events when `aria-disabled` is true, but wanted to hear about your use case first.

andrejusk:
👋 @TylerJDev our current implementation always sets the `onClick` prop. I think what you propose makes sense, where the user can interact with the button and see that it is disabled, but triggering it does nothing.

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -245,17 +252,39 @@ export const ActionBar: React.FC<React.PropsWithChildren<ActionBarProps>> = prop
...
+
+    const clickHandler = useCallback(
+      (event: React.MouseEvent<HTMLButtonElement>) => {
+        if (disabled) {
+          return
+        }
+        onClick?.(event)
+      },
+      [disabled, onClick],
+    )
```
TylerJDev, line 274: Added this to block the `onClick` that a consumer might attach to the button. The button could still be activated on an `onKeyDown` event, but I don't think that's as likely to be added here. 

Curious on any thoughts on this method, and if it just be adjusted or not 👀

francinelucca: Should we still add a keydown eventhandler then? 🤔

francinelucca (approved):

TylerJDev:
> Should we still add a keydown eventhandler then? 🤔

Doesn't hurt! I can go ahead and add it just in case

packages/react/src/ActionBar/ActionBar.tsx
TylerJDev: After doing a bit more testing, I realize we probably don't need to add one for keydown. `onClick` will handle most cases that we'd expect when utilizing a keyboard. We also probably don't want consumers using `onKeyDown` instead of `onClick`.

PR #5673, opened 2025-02-08
fix: apply specificity so that style actually applies to checked radio

### LINKED ISSUE #5672: Checked radio does not display properly
### Description

A radio button is expected to be visually differentiated when checked. 
It is not the case with @primer/react v37.11.2

### Steps to reproduce

```js
import { Radio, RadioGroup } from "@primer/react";

export default function App() {
  return (<RadioGroup name="type">
    <Radio value="foo" checked />
    <Radio value="bar" />
  </RadioGroup>)
}
```
[CodeSanbox](https://codesandbox.io/p/sandbox/primer-react-qyyepc?file=%2Fsrc%2FApp.tsx%3A1%2C1-8%2C2)

### Version

v37.11.2

### Browser

Edge

swiing:
`:where()` has 0 specificity, the consequence of which is that style does not apply, as it is overridden by other css rules. => use `:is()` instead

Closes #5672

#### Changed

replace `:where()` css by `:is()`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Radio/Radio.module.css
+++ b/packages/react/src/Radio/Radio.module.css
@@ -4,7 +4,7 @@
     background-color,
     border-color 80ms cubic-bezier(0.33, 1, 0.68, 1); /* checked -> unchecked - add 120ms delay to fully see animation-out */
 
-  &:where(:checked) {
+  &:is(:checked) {
     /* stylelint-disable-next-line primer/colors */
     background-color: var(--control-checked-fgColor-rest);
 


joshblack:
Hi there @swiing! 👋 Thanks for taking the time to open up a PR 🙏 

When looking at this it seems like the selector has a specificity of `0,1,0` since the `:where()` is nested within a class. The issue seems to be related to ordering of classes when looking at the compiled CSS:

An image of devtools showing that a CSS class with the same specificity is winning out over the intended CSS class

We could also totally increase the specificity like in this PR (which I think would go to `0,2,0`) but I think with how these two interact it would make more sense to change the ordering. I opened up: https://github.com/primer/react/pull/5765 to hopefully address this 🤞 

Hope that makes sense, let me know if you have any questions!

swiing:
Hi @joshblack , lgtm

PR #5685, opened 2025-02-10
RelativeTime: Remove `title` from Storybook stories

TylerJDev:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/RelativeTime/RelativeTime.examples.stories.tsx
+++ b/packages/react/src/RelativeTime/RelativeTime.examples.stories.tsx
@@ -2,13 +2,54 @@ import React from 'react'
 import RelativeTime from './RelativeTime'
 import Link from '../Link'
 import {Tooltip} from '../TooltipV2'
+import {Button} from '../Button'
+import {Stack} from '../Stack'
 
 export default {
   title: 'Components/RelativeTime/Examples',
   component: RelativeTime,
 }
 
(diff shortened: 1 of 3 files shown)


khiga8 (approved):
Looks good to me! It's cool that the `relative-time` already had the `noTitle` prop!

I think it's worth documenting the existence of the `noTitle` prop somewhere (whether the story or the docs).

kendallgassner:
Is there ever a time we would want a title? Why don't we hard code `no-title=true` in the component?

TylerJDev:
> Is there ever a time we would want a title? Why don't we hard code no-title=true in the component?

To be honest, not really. The goal is to eventually deprecate the prop and make not having a `title` the default, regardless of the prop. The only reason why I'm not doing this just yet is because I'm not entirely sure how often it's being used by users today. I'm hoping that by adding a lint rule, we can steer teams to more accessible patterns before removing it completely.

kendallgassner:
>  The only reason why I'm not doing this just yet is because I'm not entirely sure how often it's being used by users today. I'm hoping that by adding a lint rule, we can steer teams to more accessible patterns before removing it completely.

Can Primer Query only query github repos? Or can we use it to query for this across GitHub

kendallgassner:
Seems like a quick win could be adding `noTitle` to all use cases though!

PR #5687, opened 2025-02-11
Blankslate: Add button support to `PrimaryAction`

TylerJDev:
Closes https://github.com/github/primer/issues/4767

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Blankslate/Blankslate.tsx
+++ b/packages/react/src/Blankslate/Blankslate.tsx
@@ -59,13 +59,13 @@ function Description({children}: DescriptionProps) {
 }
 
 export type PrimaryActionProps = React.PropsWithChildren<{
-  href: string
+  href?: string
 }>
 
 function PrimaryAction({children, href}: PrimaryActionProps) {
   return (
     <div className={clsx('Blankslate-Action', classes.Action)}>
-      <Button as="a" href={href} variant="primary">
+      <Button as={href ? 'a' : 'button'} href={href} variant="primary">
         {children}
       </Button>
     </div>
--- a/packages/react/src/Blankslate/Blankslate.docs.json
+++ b/packages/react/src/Blankslate/Blankslate.docs.json
@@ -11,7 +11,10 @@
       "id": "experimental-components-blankslate-features--with-visual"
     },
     {
-      "id": "experimental-components-blankslate-features--with-primary-action"
+      "id": "experimental-components-blankslate-features--with-primary-action-as-link"
+    },
+    {
+      "id": "experimental-components-blankslate-features--with-primary-action-as-button"
     },
     {
       "id": "experimental-components-blankslate-features--with-secondary-action"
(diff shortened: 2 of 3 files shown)


joshblack (approved):

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -58,14 +58,19 @@ function Description({children}: DescriptionProps) {
...
 }
 
-export type PrimaryActionProps = React.PropsWithChildren<{
-  href: string
-}>
+export type PrimaryActionProps =
+  | (React.PropsWithChildren<{
+      href?: never
+    }> &
+      ButtonProps)
```
joshblack, line 65: With the change to include a button, do we want to leak everything from `ButtonProps` or do we want to narrow it down? I'm not sure if all the prop combinations that `Button` allows would make sense so just wanted to double-check.

One option could be to just do `React.ComponentPropsWithoutRef<'button'>` instead which should be safe.

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -58,14 +58,19 @@ function Description({children}: DescriptionProps) {
...
+  | React.PropsWithChildren<{
+      href: string
+    }>
 
-function PrimaryAction({children, href}: PrimaryActionProps) {
+function PrimaryAction({children, href, ...props}: PrimaryActionProps) {
   return (
     <div className={clsx('Blankslate-Action', classes.Action)}>
-      <Button as="a" href={href} variant="primary">
+      <Button as={href ? 'a' : 'button'} href={href} variant="primary" {...props}>
```
joshblack, line 73:
```suggestion
      <Button {...props} as={href ? 'a' : 'button'} href={href} variant="primary">
```

One small suggestion, when spreading props I think it's helpful that it's always placed beforehand otherwise we could end up in a spot where someone unintentionally is overriding a prop that we set (e.g. they could override variant when we don't want them to)

packages/react/src/Blankslate/Blankslate.tsx
TylerJDev: I wasn't too sure to be honest 🤔 Maybe it's good to restrict that a bit, at least for the time being. Thank you for the suggestion 🙏

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -1,6 +1,6 @@
 import {clsx} from 'clsx'
 import React from 'react'
-import {Button} from '../Button'
+import {Button, type ButtonProps} from '../Button'
```
TylerJDev, line 3:
```suggestion
import {Button} from '../Button'
```

PR #5705, opened 2025-02-18
Wrap `AvatarStack` avatars in `isolation: isolate` to create a new stacking context

iansan5653:
The stacking order of the avatars inside `AvatarStack` is controlled via `z-index` because it is reversed from the DOM order (earlier items display on top). This works fine within the component, but the effects of `z-index` can easily leak outside the component and cause unexpected bugs wherever any other component overlays `AvatarStack`. An example of this is https://github.com/github/copilot-productivity/issues/3985 (internal issue link) where the avatars appear on top of a dialog that should cover them.

The stacking order here only matters relative to the other avatars within the same component, so we can easily prevent this type of bug by isolating the `z-index` effects via a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context). The most explicit way to create a new stacking context is via the [`isolation`](https://developer.mozilla.org/en-US/docs/Web/CSS/isolation) property:

> The `isolation` CSS property determines whether an element must create a new stacking context.

This PR simply adds `isolation: isolate` to the `AvatarStack` container. 

There is prior art for this approach in `ButtonGroup`; see https://github.com/primer/react/pull/2910.

As a followup, it might be worth considering an audit of all uses of `z-index` within Primer to ensure they are properly isolated. In fact I would even consider an ESLint rule banning the use of `z-index` so that we have to explicitly override the rule and thus verify that we have considered the side effects + stacking context.

#### Changed

Fixed stacking bugs relating to `AvatarStack`.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/strange-pets-stare.md
+++ b/.changeset/strange-pets-stare.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Fix `AvatarStack` stacking `z-index` bugs by creating a new stacking context around the component
--- a/packages/react/src/AvatarStack/AvatarStack.module.css
+++ b/packages/react/src/AvatarStack/AvatarStack.module.css
@@ -9,6 +9,7 @@
   display: flex;
   min-width: var(--avatar-stack-size);
   height: var(--avatar-stack-size);
+  isolation: isolate;
 
   &:where([data-responsive]) {
     @media screen and (--viewportRange-narrow) {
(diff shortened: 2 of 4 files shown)


camertron (approved):
Wow, TIL about stacking contexts 😎 Any way to write tests for this?

camertron (approved):

mperrotti:
I have another branch I'm working on that uses `mask` styles to achieve something similar, and it seems like a slightly nicer approach

mperrotti:
Here's my alternative: https://github.com/primer/react/pull/5714

It's not as straightforward as this solution, but it has some nice benefits. I'll leave it up to the reviewers to decide which direction we want to go.

mperrotti (approved):

PR #5709, opened 2025-02-19
chore: add reset script for development

joshblack:
Add a reset script to `script/reset` that can be used to reset primer/react when the project may be in an invalid state. This is helpful for contributors who are trying to re-create something in CI that they are not seeing locally.

#### Changed

- Add `script/reset` as an alias for resetting the project

### Rollout strategy


jonrohan (approved):
Good idea, should you also create a `npm run reset`?

PR #5716, opened 2025-02-20
chore: Ensure class names conform to PascalCase

hussam-i-am:
Closes https://github.com/github/primer/issues/4250

#### Changed

- Adds `selector-class-pattern` rule to check casing of class names
- Updates classes that did not conform to casing
- Wraps `focus-visible` and `:popover-open` classes with `:global` and make them exceptions to the rule

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/ActionList.module.css
```diff
@@ -453,12 +453,6 @@
     & .ExpandIcon {
       transform: scaleY(-1);
     }
-
-    &.ActionListContent--hasActiveSubItem {
-      & > .ItemLabel {
-        font-weight: var(--base-text-weight-semibold);
-      }
-    }
```
hussam-i-am, line 461: @langermank I did not see any usage of this class anywhere, so I removed it. Can you confirm?

packages/react/src/BaseStyles.module.css
```diff
@@ -37,17 +37,17 @@ table {
...
+summary:focus:not(:focus-visible):not(:global(.focus-visible)),
+a:focus:not(:focus-visible):not(:global(.focus-visible)) {
   outline: none;
   box-shadow: none;
 }
 
-[tabindex='0']:focus:not(:focus-visible):not(.focus-visible),
-details-dialog:focus:not(:focus-visible):not(.focus-visible) {
+[tabindex='0']:focus:not(:focus-visible):not(:global(.focus-visible)),
+details-dialog:focus:not(:focus-visible):not(:global(.focus-visible)) {
```
hussam-i-am, line 50: @jonrohan I noticed that these did not have the `:global` wrapper, which I assume are needed for it to behave as expected.

packages/react/src/TooltipV2/Tooltip.module.css
```diff
@@ -109,8 +110,8 @@
   }
 
   /* Animation styles */
-  &.\\:popover-open,
-  &.\\:popover-open::before {
+  /* stylelint-disable-next-line selector-class-pattern */
+  &.\\:popover-open, &:global(.\\:popover-open)::before {
```
hussam-i-am, line 114: @jonrohan similar to https://github.com/primer/react/pull/5716/files#r1963911493, wrapped these in `:global`

packages/react/src/ActionList/ActionList.module.css
langermank: Oh weird, it is used in `Item` 

AL item

packages/react/src/DataTable/Table.tsx
```diff
@@ -394,15 +394,15 @@ function TableSortHeader({align, children, direction, onToggleSort, ...rest}: Ta
...
-              [classes['TableSortIcon--ascending']]: enabled,
+              [classes['TableSortIcon-ascending']]: enabled,
             })}
           />
         ) : null}
         {direction === SortDirection.DESC ? (
           <SortDescIcon
             className={clsx('TableSortIcon', 'TableSortIcon--descending', {
               [classes.TableSortIcon]: enabled,
-              [classes['TableSortIcon--descending']]: enabled,
```
joshblack, line 405: For this style I think it'd be helpful to preserve the original convention from BEM until we can update it to use our preferred approach with data attributes. Having `-descending` ends up as a weird hybrid between the two and I think it can lose the modifier meaning because of it.

joshblack (commented):
Thanks for taking the time to put this together! Just left one comment but otherwise 100% down to give this a try 👍

packages/react/src/DataTable/Table.tsx
hussam-i-am: Makes sense, updated

packages/react/src/ActionList/ActionList.module.css
langermank: Ah! Yes, that doesn't appear to be used.

joshblack (approved):
Thanks for doing this!

jonrohan (approved):

langermank (approved):

hussam-i-am:
> Would be cool to setup something that would check for unused CSS

Yeah, I thought we already had something that checks, but I guess not in primer

hussam-i-am:
@jonrohan I added the package [typed-css-modules](https://github.com/primer/react/pull/5716/commits/e1ae9b5c491503c2e12396bfc4970bbfa7f9f599), which will generate type definition files and at least ensure we are using class names that do exist. Couldn't find a clear solution for ensuring that any classes within the css file are being used.

packages/react/src/BaseStyles.module.css
francinelucca: This is going in on the next release, anything specific that needs to be validated UI-wise before landing this? @jonrohan @hussam-i-am

jonrohan: Not sure how to recreate this selector in the dom. Maybe drop a focus-visible class on a details-dialog element

jonrohan: Double checked, doesn't look like it breaks anything. We have some duplicate styles for this in primer/css that's already being imported

Issue #5723, opened 2025-02-27
Bug: AriaAlert does not announce RelativeTime content.


khiga8:
### Description

`AriaAlert` fails to announce `RelativeTime` content.

Related: https://github.com/github/accessibility-audits/issues/10539

Questions: 
* Should AriaAlert be able to wrap and announce any content? Or should this only be used to wrap simple text content?
  * It may add a lot of complexity to try to get `AriaAlert` to handle custom elements like `relative-time`, or handle ARIA semantics correctly.
  * For example, if `AriaAlert` wraps content containing `<p aria-hidden="true">Don't announce this</p>`, should this aria-hidden content be announced?
  * I think whichever we land on is fine, as long as we document limitations/expectations!

### Steps to reproduce

1. Checkout [my branch](https://github.com/primer/react/pull/new/kh-aria-alert-relative-time).
2. Start up storybook.
3. Go to the AriaAlert relative time storybook example.
4. Navigate example with screen reader (e.g. VoiceOver, NVDA) and observe that relative time content is skipped in the screen reader announcement.

### Version

v.35.2.0

joshblack:
Great point @khiga8, I guess with elements that have a Shadow DOM the utility we're using won't pull out their text content 🤔 

I feel like we could go down the route of trying to add accessible name computation to the utility and make it more robust (I'm assuming that would mean these steps? https://www.w3.org/TR/accname-1.2/#computation-steps). We could also offer up a "lightweight" version of this that works in most situations (but not all)

I opened up a PR at: https://github.com/primer/react/pull/5724 to explore this, would love to hear your thoughts!

PR #5724, opened 2025-02-27
feat(Announce): add support for computing text equivalence

joshblack:
Context https://github.com/primer/react/issues/5723

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/live-region/Announce.features.stories.tsx
```diff
@@ -36,3 +37,11 @@ export const WithDelay = () => {
 
   return <Announce delayMs={1000}>{message}</Announce>
 }
+
+export const WithCustomElement = () => {
+  return (
+    <Announce>
+      <RelativeTime date={new Date('2020-01-01T00:00:00Z')} noTitle={true} />
```
joshblack, line 44: Note, the announcement in this story will still be the default date (and not the "on ..." message) since when the story renders and gets announced the custom element hasn't updated yet.

packages/react/src/live-region/Announce.tsx
```diff
@@ -132,12 +132,110 @@ export function Announce({
...
+  allowHidden: false,
+}
+
+/**
+ * Simplified version of the algorithm to compute the text equivalent of an
+ * element.
+ *
+ * @see https://www.w3.org/TR/accname-1.2/#computation-steps
+ */
+function computeTextEquivalent(
```
khiga8, line 149: This seems like a good start - I think we can refine this as we identify different usecases (alongside stories/tests).

packages/react/src/live-region/Announce.tsx
```diff
@@ -132,12 +132,110 @@ export function Announce({
...
+        return null
+      })
+      .filter(Boolean)
+      .join(' ')
+  }
+
+  const role = elementOrText.getAttribute('role')
+
+  if (role === 'combobox' || role === 'listbox') {
+    const options = elementOrText.querySelectorAll('option[aria-selected="true"]')
```
khiga8, line 201: I wonder how much if/how much support we need for complex interactive controls.

Ideally, consumers aren't wrapping entire interactive components with a live region, since semantics won't properly be conveyed 🤔

khiga8 (commented):
Wow, thank you so much for following up so quickly! I think this is a good starting point! ✨ 

I wonder if the announcement should be somewhat close to the output we'd get if we added `role='alert'` directly on an element.
* I'm actually not sure how `aria-hidden` elements are handled by `role="alert"` , so I think it'd be good to test that with a screen reader, and potentially align the output here! 
* Developers _shouldn't_ be adding live regions on anything and everything (e.g. complex interactive widgets) since semantics aren't conveyed through live regions. I wonder how much we need to support it 🤔

PR #5725, opened 2025-02-28
Change button to accept string in count

lukasoppermann:
For the RepoPicker we use a button that shows the count of the selected items. This can easily be more than `1000`. SO we would like to use the _"human format"_, like `3.2k` instead of `3200`.

This PR allows this by changing the type of count to `number | string` as suggested by @mperrotti 

#### Changed

Type of `count` changed to `number | string`

### Rollout strategy

- [ ] Patch release
- x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/yellow-trainers-arrive.md
+++ b/.changeset/yellow-trainers-arrive.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": minor
+---
+
+Allowed button count prop to accept a number or a string, to allow for human format, like 3.2k
--- a/packages/react/src/Button/Button.features.stories.tsx
+++ b/packages/react/src/Button/Button.features.stories.tsx
@@ -63,6 +63,10 @@ export const TrailingCounter = () => {
 
 export const TrailingCounterWithNoText = () => <Button aria-label="Comments" leadingVisual={CommentIcon} count={3} />
 
+export const TrailingCounterWithHumanFormat = () => (
+  <Button aria-label="Comments" leadingVisual={CommentIcon} count={'3.2k'} />
+)
+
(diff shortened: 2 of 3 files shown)


packages/react/src/Button/Button.features.stories.tsx
```diff
@@ -63,6 +63,10 @@ export const TrailingCounter = () => {
 
 export const TrailingCounterWithNoText = () => <Button aria-label="Comments" leadingVisual={CommentIcon} count={3} />
 
+export const TrailingCounterWithHumanFormat = () => (
+  <Button aria-label="Comments" leadingVisual={CommentIcon} count={'3.2k'} />
```
joshblack, line 67:
```suggestion
  <Button aria-label="Comments" leadingVisual={CommentIcon} count="3.2k" />
```

Small suggestion since this is a string and we don't need to the JSX expression syntax for this

.changeset/yellow-trainers-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": minor
+---
+
+Allowed button count prop to accept a number or a string, to allow for human format, like 3.2k
```
joshblack, line 5:
```suggestion
Allow button count prop to accept a number or a string, to allow for human format, like 3.2k
```

Small suggestion for present tense

joshblack (approved):
Looks great! Just left a couple of suggestions

PR #5741, opened 2025-03-06
Add prop to PageHeader to allow a bottom border

mperrotti:
Relates to https://github.com/github/primer/issues/2837

Screenshot 2025-03-06 at 4 11 06 PM

#### New

- `hasBorder` prop on `PageHeader`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/PageHeader/PageHeader.tsx
```diff
@@ -176,6 +183,7 @@ const Root = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageHeader
         ref={rootRef}
         as={as}
         className={clsx(enabled && classes.PageHeader, className)}
+        data-has-border={hasBorder}
```
joshblack, line 186:
```suggestion
        data-has-border={hasBorder ? '' : undefined}
```

Small suggestion so that in the HTML it shows up as `data-has-border` when true and then we can target it with just `[data-has-border]`. When it's not in the DOM it's removed.

I also asked this question in our CSS Modules group to see if there was a style preference here 🤔 Not sure what folks prefer and this is definitely just a style thing

packages/react/src/PageHeader/PageHeader.tsx
```diff
@@ -754,6 +762,7 @@ const Navigation: React.FC<React.PropsWithChildren<NavigationProps>> = ({
       aria-label={as === 'nav' ? ariaLabel : undefined}
       aria-labelledby={as === 'nav' ? ariaLabelledBy : undefined}
       className={clsx(enabled && classes.Navigation, className)}
+      data-component="PH_Navigation"
```
joshblack, line 765:
```suggestion
      data-component="PageHeader.Navigation"
```

One suggestion that could make it more clear what the component is when coming back to this later

lukasoppermann:
@mperrotti somehow I can't add this comment to the code, but we need this: `hasBorder={args.hasBorder}` on the Playground story: https://github.com/primer/react/pull/5741/files#diff-38be7109ab257cb99d6f1950aa251d530e55a99c8705aaf94ff68f2d98c82897R195

lukasoppermann:
Also it is buggy for those two cases:
CleanShot 2025-03-07 at 10 21 54@2x
CleanShot 2025-03-07 at 10 22 25@2x

I wonder if we should disable those cases, so that lines can not be added when a Navigation or description is enabled?

But for a normal text description it could be useful.

packages/react/src/PageHeader/PageHeader.tsx
mperrotti: `PH_{Subcomponent}` is the naming pattern we use for the other subcomponents. Do you still think we should change it?

mperrotti:
@lukasoppermann - I thought I handled those cases already, but I must have broke something. I'll take another look.

packages/react/src/PageHeader/PageHeader.tsx
langermank: I think we're kind of trying to move away from `data-component` actually.. what about just giving it a classname instead?

mperrotti:
Updating snapshots since we added a new PageHeader story

emilybrick (approved):
Snapshots look good here & in storybook 
Screenshot 2025-03-19 at 1 01 56 PM


I'm not sure if the [naming pattern discussion](https://github.com/primer/react/pull/5741/files#r2001931485) is resolved, but story + snapshots lgtm 👍

packages/react/src/PageHeader/PageHeader.tsx
TylerJDev: I agree, `PageHeader.Navigation` is more clear, but to keep it consistent I think `PH_{Subcomponent}` is fine too! Would be curious about the className route too 👀

TylerJDev (approved):
LGTM!

PR #5756, opened 2025-03-10
feat: Add workflow to toggle release lock status

jonrohan:
This workflow toggles rulesets for the release conductor to block further merges into `main` while allowing the conductor to still merge. This first step makes it a workflow_dispatch that we can run from the actions tab, but I could see in the future more automation around doing this via labeling. 

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

This has to do with internal actions workflows and nothing to add to releases

### DIFF
--- a/.github/workflows/lock-release.yml
+++ b/.github/workflows/lock-release.yml
@@ -1,0 +1,59 @@
+name: Lock Release Pull Request
+
+on:
+  workflow_dispatch:
+    inputs:
+      action:
+        description: 'Lock or unlock the release?'
+        required: true
+        type: choice
+        options:
+          - lock
+          - unlock
+
+jobs:
+  lock:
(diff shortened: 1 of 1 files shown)


.github/workflows/lock-release.yml
```diff
@@ -0,0 +1,59 @@
...
+    name: Toggle Release Lock
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get App Token
+        uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.PRIMER_APP_ID_SHARED }}
+          owner: primer
+          repositories: react
```
joshblack, line 25: Hello this is a test 😄

joshblack (approved):

PR #5760, opened 2025-03-11
Radio: Remove `aria-required` and `aria-invalid`

TylerJDev:
Context https://github.com/github/primer/issues/3726

Removes `aria-required` and `aria-invalid` from the `Radio` component, as they are not required on the component. This change allows us to remove a lint disable for `a11y-role-supports-aria-props`.

#### Removed

* Removes `aria-invalid` and `aria-required` from `Radio`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Radio/Radio.tsx
+++ b/packages/react/src/Radio/Radio.tsx
@@ -77,7 +77,6 @@ const Radio = React.forwardRef<HTMLInputElement, RadioProps>(
 
     if (sxProp !== defaultSxProp) {
       return (
-        // eslint-disable-next-line github/a11y-role-supports-aria-props
         <Box
           as="input"
           sx={sxProp}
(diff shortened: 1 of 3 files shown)


packages/react/src/Radio/Radio.tsx
```diff
@@ -35,10 +34,6 @@ export type RadioProps = {
    * Indicates whether the radio button must be checked before the form can be submitted
    */
   required?: boolean
-  /**
-   * Only used to inform ARIA attributes. Individual radio inputs do not have validation styles.
-   */
-  validationStatus?: FormValidationStatus
```
TylerJDev, line 41: We don't advertise this prop in our docs, and it doesn't have any usage outside of the `aria-*` props which are removed in this PR. The prop itself didn't add any functionality and I don't think this would need to be a major release. There's only one place (outside of Dotcom) that appears to utilize this prop.

joshblack: @TylerJDev if you feel like we can safely roll this out then I'm okay with that 👍 I think it would fall under a breaking change but I get if it doesn't feel worth it to migrate something we don't advertise and doesn't have usage.

TylerJDev: Definitely! I think this is one of those things that didn't really do anything, and was never adopted anywhere, so we can probably get away with just removing it 😄

joshblack (approved):

PR #5761, opened 2025-03-11
SelectPanel: Make panel full screen on narrow screens with Save and optional Cancel button

siddharthkp:
> [!NOTE]  
> - Companion PR in gh/gh is merged: https://github.com/github/github/pull/367564 to keep main green

---

- Closes https://github.com/github/primer/issues/4109
- Implements `Save` button on narrow screens which calls `onClose('click-outside')` to simulate wide screen behavior
- Accept `onCancel` prop that renders a `Cancel` button on narrow which calls `onClose('escape')`

https://github.com/user-attachments/assets/7d4f7927-9356-47b3-a2c7-e9b87cf1584b

With `onCancel` prop:

https://github.com/user-attachments/assets/6815d98c-07bc-411c-b443-a81846d4b505

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


francinelucca (commented):
Looking good! Wondering if we are still going with [ option 2](https://github.com/github/primer/issues/4109#issuecomment-2454634188) described here as well for the close/cancel button? 👀

siddharthkp:
> Looking good! Wondering if we are still going with https://github.com/github/primer/issues/4109#issuecomment-2454634188 described here as well for the close/cancel button? 👀

👋 Still making my mind up about it. I like option 2 but my confidence with a hacky cancel isn't super high. I think we should try it but we would run into scenarios where it doesn't clean up side effects that it doesn't know about, does that make sense?

I'm wondering if we should split this work into 2 tasks:
- the first task implements option 1 which "unblocks" accessibility fixes for teams to implement `onCancel`
- the second is to experiment with a hacky cancel to help provide a fallback for teams who haven't added one yet (priority 2?)

I think this is also relevant for https://github.com/github/primer/issues/4106 and what we decide to do there (implicit cancel vs explicit onCancel like experimental/SelectPanel)

francinelucca:
@siddharthkp's comment above makes sense to me. I'm fine with making the cancel/close button opt-in for non-modal version, that way the [45 existing instances](https://primer-query.githubapp.com/?name=%22SelectPanel%22&entrypoint=%22%40primer%2Freact%22) of SelectPanel v1 in dotcom don't need to be compromised by this change. We can also add a lint rule to flag the missing prop functions for onCancel/onClose if we want people to slowly migrate into it on their own time in lieu.
This is not a concern for modal version since this is a new feature in SelectPanel v1 and we can make those props required, guaranteeing that the buttons are always available.

Wondering if this approach would be ok with design @emilybrick? 👀  Happy to talk through further if more context if required
Also tagging @TylerJDev in case there's any accessibility concerns here you can think of.

emilybrick:
I don't think the end user experience would be affected as a result of going w/ the 'hacky cancel' (option 2?) – is that correct?

> the first task implements option 1 which "unblocks" accessibility fixes for teams to implement onCancel
the second is to experiment with a hacky cancel to help provide a fallback for teams who haven't added one yet (maybe in priority 2 batch?)

If not, this SGTM 👍 

Dumb question – both the modal + nonmodal desktop versions would be the same exact experience on small screens/mobile (full screen), correct?

siddharthkp:
> Dumb question – both the modal + nonmodal desktop versions would result in the same mobile experience (full-screen), correct?

Good question! Yes, it would be the same full-screen.

More context: experimental/SelectPanel(2) did have an additional [`bottom-sheet` variant](https://primer.style/react/storybook/?path=/story/experimental-components-selectpanel-features--responsive-variants) for narrow screens as well (resize your window). But, I think we should consider that out of scope for this epic.

siddharthkp:
Before reading on, try this: Go to an issue or pull request, add a label then hit `Escape`, should the label have been applied or discarded?

Side bonus: Once we have `onCancel`, we can also look into _cancelling_ updates on `Escape`

emilybrick:
> But, I guess we should consider that out of scope for this epic.

Yes @siddharthkp ! It's added to this (WIP) issue: https://github.com/github/primer/issues/4862#issue-2908674393

siddharthkp:
@francinelucca @langermank Ready for your review!

langermank (approved):
VRT 👍

packages/react/src/SelectPanel/SelectPanel.docs.json
```diff
@@ -127,6 +127,12 @@
       "description": "Callback when the search input changes",
       "defaultValue": ""
     },
+    {
+      "name": "onCancel",
+      "type": "() => void",
+      "description": "(Narrow screens) Callback when the user hits cancel or close",
```
francinelucca, line 133: do we expect this to be used in modal as well? If so maybe we need to correct this verbiage

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -530,6 +562,45 @@ export function SelectPanel({
...
+                      display: ['flex', 'flex', 'none', 'none'],
+                      gap: 'var(--stack-gap-condensed)',
+                      justifyContent: 'right',
+                      padding: 3,
+                      // .Footer
+                      borderTop: 'var(--borderWidth-thin) solid',
+                      borderTopColor: 'var(--borderColor-default)',
+                    }
+              }
+              className={enabled ? clsx(classes.Footer, classes.ResponsiveFooter) : undefined}
```
francinelucca, line 581: if this is net new code can we code it with css modules from the get-go? 🙏

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -560,3 +560,49 @@ export const AsyncFetch: StoryObj<SelectPanelProps> = {
...
+          <Button trailingAction={TriangleDownIcon} {...anchorProps} aria-haspopup="dialog">
+            {children}
+          </Button>
+        )}
+        open={open}
+        onOpenChange={setOpen}
+        items={selectedItemsSortedFirst}
+        selected={selected}
+        onSelectedChange={setSelected}
+        onCancel={() => setSelected(intialSelection)}
```
francinelucca, line 602: could we show a more "real-life" implementation of this? Right now it always goes back to the initial items on cancel rather than what was actually selected when the panel opened:

https://github.com/user-attachments/assets/10627930-ffe7-43f8-bde3-2d9fde61213b

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -530,6 +562,45 @@ export function SelectPanel({
             >
               {footer}
             </Box>
+          ) : (
+            <Box
```
francinelucca, line 566: should we only show this when the variant is multi? feels weird to have this save button here when the selectpanel closes on selection anyways 🤔 
CC: @emilybrick 

https://github.com/user-attachments/assets/89689204-5d08-442e-afed-5a2674f40be8

francinelucca (changes requested):
Looking great! Have a few comments, suggestions.
Also can we make sure to merge https://github.com/github/github/pull/367564 before this as to not break main? 🙏

siddharthkp:
> Also can we make sure to merge https://github.com/github/github/pull/367564 before this as to not break main? 🙏

Absolutely. Idk why I thought we can merge them in any order, I'm rusty from my break 😅

packages/react/src/SelectPanel/SelectPanel.tsx
siddharthkp: Yesss, good point. I'll hide the footer here but keep the close button in the header 👍

packages/react/src/SelectPanel/SelectPanel.tsx
siddharthkp: Oh we get to that now? I thought we can't yet because it's still behind a feature flag that we'd want the ability to roll back?

packages/react/src/SelectPanel/SelectPanel.docs.json
siddharthkp: We do, but we don't have that yet. Can update once we have modal.

It's going to be a strange description: `Narrow screens or variant=modal` 😅

packages/react/src/SelectPanel/SelectPanel.tsx
jonrohan: Should be able to go all in with CSS for new features. The only precaution with having the flag fall back to styled components is that the new styles don't match the old ones. In this case there is nothing to compare the new styles to

packages/react/src/SelectPanel/SelectPanel.tsx
siddharthkp: Fixed!

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
siddharthkp: Fixed!

francinelucca (approved):
✨  yayy, looks great!
smaaaaall nit: if we could add onCancel to our singleSelect story so that the close button shows I think that'd be great. Just realized there's no way to close the panel without changes 🙏 
image

PR #5770, opened 2025-03-13
Update ActionListGroupProps to include `className`

joshblack:
Update the props for `ActionListGroupProps` to include `className` and other props that are spread onto the outermost element.

#### New

- Add support for `className` to the TypeScript types for `ActionListGroupProps`

### Rollout strategy

- [x] Minor release


jonrohan (approved):

hectahertz (approved):
👍🏻

PR #5788, opened 2025-03-18
chore(Table): Remove the CSS modules feature flag from the Table component

jonrohan:
This PR removes the CSS modules feature flag from the `Table` component. The component [Table](https://primer-query.githubapp.com/?name=table&package=%22%40primer%2Freact%22&repo=%22github%2Fgithub%22) and sub components are used `121` times in dotcom.

Closes https://github.com/github/primer/issues/4307

#### Removed

Removes the CSS modules feature flag from the `Table` components. 

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Using Integration testing.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/DataTable/Table.tsx
```diff
@@ -572,33 +261,13 @@ export type TableTitleProps = React.PropsWithChildren<{
...
   return (
-    <StyledTableTitle
-      as={as}
-      className={clsx('TableTitle', {
-        [classes.TableTitle]: enabled,
-      })}
-      id={id}
-      ref={ref}
-    >
+    <Box as={as} className={clsx('TableTitle', classes.TableTitle)} id={id} ref={ref}>
```
jonrohan, line 265: It seems like there should be a simpler way of doing this? Can I avoid `Box as=` when not using `sx`? @joshblack What do I need to change here?

joshblack: I bet you could just render whatever `as` is directly, specifically:

```tsx
const TableTitle = React.forwardRef<HTMLElement, TableTitleProps>(function TableTitle({as: BaseComponent = 'h2', children, id}, ref) {

  return (
    <BaseComponent className={clsx('TableTitle', classes.TableTitle)} id={id} ref={ref}>
```

jonrohan: Typescript not happy about that

```
src/DataTable/Table.tsx:268:6 - error TS2322: Type '{ children: ReactNode; className: string; id: string; ref: ForwardedRef<HTMLElement>; }' is not assignable to type 'SVGProps<SVGSymbolElement>'.
  Types of property 'ref' are incompatible.
    Type 'ForwardedRef<HTMLElement>' is not assignable to type 'LegacyRef<SVGSymbolElement> | undefined'.
      Type '(instance: HTMLElement | null) => void' is not assignable to type 'LegacyRef<SVGSymbolElement> | undefined'.
        Type '(instance: HTMLElement | null) => void' is not assignable to type '(instance: SVGSymbolElement | null) => void'.
          Types of parameters 'instance' and 'instance' are incompatible.
            Type 'SVGSymbolElement | null' is not assignable to type 'HTMLElement | null'.
              Type 'SVGSymbolElement' is missing the following properties from type 'HTMLElement': accessKey, accessKeyLabel, autocapitalize, dir, and 25 more.

268     <BaseComponent className={clsx('TableTitle', classes.TableTitle)} id={id} ref={ref}>
         ~~~~~~~~~~~~~

src/DataTable/Table.tsx:268:20 - error TS2590: Expression produces a union type that is too complex to represent.

268     <BaseComponent className={clsx('TableTitle', classes.TableTitle)} id={id} ref={ref}>
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

joshblack: Ah, I think `as?: keyof JSX.IntrinsicElements | React.ComponentType` is including SVG but the `forwardRef` is saying `HTMLElement`

We could extend that for now if we want:

```
const TableTitle = React.forwardRef<HTMLElement | SVGSymbolElement, TableTitleProps>
```

Or we could restrict the `as?` prop to omit things. Let me know what you think 👀

francinelucca (approved):

joshblack (approved):

PR #5789, opened 2025-03-18
Fix axe violation: scrollable-region-focusable

kendallgassner:
Closes #

Fixes: Scrollable-region-focusable axe violations in the primer-docs site. 

The primer-docs site uses the `Table` component which contains a ScrollableRegion. The ScrollableRegion has `overflow: auto` without always adding `tab-index:0`

#### Changed

Only add `overflow: auto` if there is overflow. 

### Rollout strategy

- [X] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/icy-pianos-enjoy.md
+++ b/.changeset/icy-pianos-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+[ScrollableRegion] fix axe violation: scrollable-region-focusable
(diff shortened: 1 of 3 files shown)


joshblack (commented):
Oh wow, interesting 👀  Does this fail the axe check even if the region does not have overflow?

One thing I think removing the default styles does is cause a flash when the component hydrates since this no longer is applied by default and will be applied when the JS executes.

https://github.com/user-attachments/assets/3955678d-fc0c-4f61-84dc-cc626c0248d9

This also seems to cause some visual regressions in Dialog when used in this way (just based off of CI)

kendallgassner:
@joshblack sorry I was in a meeting and didn't have a chance to look at the vrts.

Sad.. I assumed this would be an easy bug to fix. 

Honestly, this might just be a false positive. When the section actually becomes scrollable than the `tab-index` is added. The rule just breaks because `overflow: auto` is always include in the css.

kendallgassner:
I'll work to disable the test in primer-docs.

PR #5797, opened 2025-03-19
SubNav: Update selected styles to only apply to `data-selected='true'`

broccolinisoup:
When the SunNav component is used in an environment with practical data, it is high likely that the `selected` attribute will be used in each SubNav children and their value will be conditionally rendered to true or false. Therefore the generated HTML will be 

```html
<a data-selected="false" aria-current="false" href="#nav1">Nav 1</a>
<a data-selected="true" aria-current="true" href="#nav2">Nav 2</a>
<a data-selected="false" aria-current="false" href="#nav3">Nav 3</a>
```

Currently the way we apply the selected styles are causing bug because the styles are applied to all links because all have `data-selected` property.

Here is an example from RL that shows the bug

sub nav had two items and both are highlighted with the selected color

There is currently no usage of SubNav at dotcom - this is why we probably didn't noticed the issue so far.

#### Changed

- Instead of selecting `data-selected` attribute to apply selected styles, I updated the code to select `data-selected="true"`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

If you could edit the Selected story so that each children have selected prop. You can repro the issue and with the change in this PR, you can verify the fix.

```js
export const Selected = () => {
  const [selectedItem, setSelectedItem] = React.useState('home')

  return (
    <SubNav aria-label="Main">
      <SubNav.Links>
        <SubNav.Link href="#home" selected={selectedItem === 'home'}>
          Home
        </SubNav.Link>
        <SubNav.Link href="#documentation" selected={selectedItem === 'documentation'}>
          Documentation
        </SubNav.Link>
        <SubNav.Link href="#support" selected={selectedItem === 'support'}>
          Support
        </SubNav.Link>
      </SubNav.Links>
    </SubNav>
  )
}
```

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/strong-lemons-rescue.md
+++ b/.changeset/strong-lemons-rescue.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+SubNav: Update selected styles to only apply to `data-selected='true'`
--- a/packages/react/src/SubNav/SubNav.module.css
+++ b/packages/react/src/SubNav/SubNav.module.css
@@ -60,7 +60,7 @@
     transition: background-color 0.2s ease;
   }
 
-  &:is([data-selected]) {
+  &:is([data-selected='true']) {
     color: var(--fgColor-onEmphasis);
     background-color: var(--bgColor-accent-emphasis);
     /* stylelint-disable-next-line primer/colors */


siddharthkp (approved):
Thank you!

PR #5801, opened 2025-03-20
SelectPanel: Remove `aria-activedescendant` from SelectPanel 

TylerJDev:
Closes https://github.com/github/primer/issues/4740

Replaces `aria-activedescendant` pattern with roving tabindex. This functionality will exist only under the `primer_react_select_panel_with_modern_action_list` feature flag

#### Changed
* Replaced `aria-activedescendant` pattern with roving tabindex under FF

#### Removed
* Removed some `useAnnouncement` logic

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/List.tsx
```diff
@@ -55,7 +56,7 @@ export const List = React.forwardRef<HTMLUListElement, ActionListProps>(
       disabled: !enableFocusZone,
       containerRef: listRef,
       bindKeys: FocusKeys.ArrowVertical | FocusKeys.HomeAndEnd | FocusKeys.PageUpDown,
-      focusOutBehavior: listRole === 'menu' ? 'wrap' : undefined,
+      focusOutBehavior: listRole === 'menu' || container === 'SelectPanel' ? 'wrap' : undefined,
```
TylerJDev, line 59: This will also affect `SelectPanelV2`, not too sure if that's a concern 🤔 

Essentially, it allows focus to "wrap", so that if you continue to hold the arrow up or down, focus will continue to be set.

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -307,7 +307,7 @@ export function SelectPanel({
     if (inputRef?.current) {
       const ref = inputRef.current
       const listener = () => {
-        announceFilterFocused()
+        // announceFilterFocused()
```
TylerJDev, line 310: Not sure if we need this, so I commented it out for now 😅 

It announces 'Focus on filter text box and list of items', which I think we already have covered with the 'Items will be filtered as you type' announcement. This is mainly to reduce some verbosity.

packages/react/src/FilteredActionList/useAnnouncements.tsx
```diff
@@ -17,26 +16,6 @@ const useFirstRender = () => {
   return firstRender.current
 }
 
-const getItemWithActiveDescendant = (
```
TylerJDev, line 20: Since we're no longer using `aria-activedescendant` in the modern ActionList FF, I removed this. 

We also don't seem to utilize `useAnnouncements` in the deprecated version, so I don't think there's a need to keep it.

packages/react/src/FilteredActionList/useAnnouncements.tsx
```diff
@@ -57,22 +40,9 @@ export const useAnnouncements = (
...
-          const announcementText = [
-            `Focus on filter text box and list of items`,
-            `Focused item: ${text}`,
-            `${selected ? 'selected' : 'not selected'}`,
-            `${index + 1} of ${items.length}`,
-          ].join(', ')
-          announce(announcementText, {
-            delayMs,
-            from: liveRegion ? liveRegion : undefined, // announce will create a liveRegion if it doesn't find one
-          })
```
TylerJDev, line 75: Same as https://github.com/primer/react/pull/5801/commits/5ef984ee805c09457148e91b8d7fcd0dcfb12b66#r2007612765, a lot of this should be provided for free now that we're using roving tabindex, as the user must navigate through the list. This will notify them of the element that is focused/selected.

packages/react/src/ActionList/List.tsx
francinelucca: > This will also affect SelectPanelV2, not too sure if that's a concern 🤔

So the only impact here is that SelectPanelV2 will now focus wrap? I think this should be fine, especially since V2 is on its way out anyways... CC: @emilybrick

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
```diff
@@ -189,33 +177,36 @@ export function FilteredActionList({
...
-            id={listId}
-            sx={{flexGrow: 1}}
+          <ActionListContainerContext.Provider
+            value={{
+              container: 'SelectPanel',
+              listRole: 'listbox',
+              selectionAttribute: 'aria-selected',
+              selectionVariant,
+              enableFocusZone: true,
+            }}
```
francinelucca, line 187: can you add more context (pun intended) around this? I'm wondering if this should live in SelectPanel.tsx instead? seems weird to me that we're supplying container: 'SelectPanel' here, since the FilteredActionList can be used independent of the SelectPanel

francinelucca (changes requested):
Mostly LGTM, just one tiny thing: I don't think it's announcing "No matching items" for FilteredActionList story:



https://github.com/user-attachments/assets/4a1b25b8-57a3-47ba-b5b6-485ee4245056

packages/react/src/SelectPanel/SelectPanel.tsx
francinelucca: Makes sense to me, let's make sure to remove the line then before merging this 🙏

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
TylerJDev: Yup, good point!

I added this so that I could create a [conditional for the "wrapping" behavior of the focus zone](https://github.com/primer/react/pull/5801/files#diff-62a4339dd17d9c37cec92eefa74dcc28806909bd4b3c7c8eea281b97813a4a15R59). This also gives us the benefit of not rendering things like [`TrailingAction` in the SelectPanel](https://github.com/primer/react/blob/main/packages/react/src/ActionList/Item.tsx#L604). 

I think the name of the container is a good callout. I'm wondering if we should instead rename it instead (e.g. `container: 'FilteredActionList'`)? The benefit of that is this would ensure things like `TrailingAction` wouldn't be rendered in `FilteredActionList`.

We could also add this to `SelectPanel`, but will need to put it behind the FF so it doesn't render when the modern action list flag isn't on. My preference would be changing the name, mainly so it still handles the `TrailingAction` component, but curious on what you think 🤔

TylerJDev:
> Mostly LGTM, just one tiny thing: I don't think it's announcing "No matching items" for FilteredActionList story:

I think this is because we're announcing the "No matching items" in SelectPanel still. I'll follow up with adding it into `useAnnouncement` instead!

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
francinelucca: changing the name sounds good to me!

emilybrick (approved):
Snapshots look good here. Will update [these docs](https://primer.style/product/components/select-panel/guidelines/#interactions) to reflect new changes.

francinelucca (approved):
LGTM now ✨ . Let's wait until integration CI runs green before merging 🙏

francinelucca (approved):
:shipit:

francinelucca (approved):

emilybrick (approved):
LGTM 

[Chatted in Slack for reference](https://github.slack.com/archives/C087S9D3W87/p1742927773451449?thread_ts=1742846536.006249&cid=C087S9D3W87)

TylerJDev:
> 🔴 [golden-jobs](https://github.com/github/github/actions/runs/14067664804) completed with status `failure`.

Failures are unrelated to the PR, and should be resolved in the next release!

PR #5805, opened 2025-03-21
docs: add api guidelines doc

joshblack:


joshblack:
Couple of ideas:

- Prefer `interface` to `type` to avoid `never` from intersections

    ```tsx
    type A = { value: number };
    type B = A & { value: string };
    // B['value'] == never

    interface C { value: number }
    interface D extends C { value: string } // <- Types of property 'value' are incompatible
    ```

- Prefer `React.HTMLAttributes<HTMLElement>` over `React.ComponentPropsWithoutRef` unless element type is a part of public API
  - Allows generic spread on outermost element so that you can change element type over time

contributor-docs/api.md
```diff
@@ -0,0 +1,194 @@
...
+- When naming "opt-in" variants, consider the default value and what it should
+  be. Name the prop according to what they will "opt-in" to.
+
+#### Naming variant props
+
+- Any specific naming conventions?
+- Should they always have a default value? (Maybe the one we expect to be most
+  used?)
+
+#### `useCallback` and `useMemo`
```
hectahertz, line 111: This won't be really necessary after the React Compiler, so maybe  we can skip it

contributor-docs/api.md
```diff
@@ -0,0 +1,194 @@
...
+  synchronizing the value with the effect
+
+### Focus management
+
+- Prefer managing focus in event handlers over effects
+- Technique: how to manage focus when visibility is managed by state
+  - Use `flushSync()`
+  - Use custom hook for this (`useFocus`)
+
+### When to provide default props
```
hectahertz, line 179: Also I'd like for us to talk about prop spread: it gets really convoluted and hard to know what we actually surface. [SelectPanel is a good example].(https://github.com/primer/react/blob/703b69dc06db5701a0b09dd51f46b6ceb9ec2e25/packages/react/src/SelectPanel/SelectPanel.tsx#L100)

joshblack: Great point 👍

joshblack:
Closing in favor of: https://github.com/primer/react/pull/5987

PR #5759, opened 2025-03-11
AnchoredOverlay: Add responsive variant prop

siddharthkp:
> [!NOTE]  
> - Parallel PR in gh/gh: https://github.com/github/github/pull/366815, can be merged in any order as long as gh/gh PR is merged before we upgrade primer/react

---

- Prerequisite for https://github.com/primer/react/pull/5761
- Adds optional `variant` prop to AnchoredOverlay. It follows the same API style as other responsive variant props.
- No user changes expected
- Bonus: Add vrt for select panel with narrow screen

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [x] Added/updated previews (Storybook)
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #5817, opened 2025-03-24
SelectPanel: Add variant=modal

siddharthkp:
> [!NOTE] 
> Current status: `main` is failing integration tests, so need to run integration tests again once that is cleared up
> 
> Note: This PR does not include radio icons for single selection in a modal

- Closes https://github.com/github/primer/issues/4106

#### Features:

Adding `variant=modal` does 3 things
1. Changes the position to be in the center of the screen
2. Makes `onCancel` a required prop
3. Adds a backdrop, when you click the backdrop (click-outside), we call `onCancel`

#### Shortcut

We are still using `AnchoredOverlay` underneath. But, for `variant=modal`, we override its position by directly passing `top`, `left`, etc. to `overlayProps` going down to the Overlay

The ideal super clean solution would have been to refactor SelectPanel and conditionally use `useAnchoredPosition` instead of AnchoredOverlay for variant=anchored. Something I did attempt in #5230 earlier but had to revert.

This shortcut is easier to implement and touches less surface area.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Overlay/Overlay.tsx
```diff
@@ -127,6 +127,7 @@ type BaseOverlayProps = {
   role?: AriaRole
   children?: React.ReactNode
   className?: string
+  'data-variant'?: 'anchored' | 'fullscreen' | 'modal'
```
siddharthkp, line 130: hmm maybe this should be `variant` 🤔

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -459,6 +463,7 @@ export function SelectPanel({
           role: 'dialog',
           'aria-labelledby': titleId,
           'aria-describedby': subtitle ? subtitleId : undefined,
+          ...(variant === 'modal' ? {'data-variant': 'modal' /* override AnchoredOverlay */} : {}),
```
siddharthkp, line 466: this is a bit icky, we can't pass variant=anchored here because it overrides AnchoredOverlay :/

packages/react/src/Overlay/Overlay.docs.json
```diff
@@ -143,6 +143,13 @@
     {
       "name": "sx",
       "type": "SystemStyleObject"
+    },
+    {
+      "name": "data-responsive",
```
siddharthkp, line 148: maybe call this responsive variant instead?

packages/react/src/Overlay/Overlay.tsx
```diff
@@ -103,15 +103,6 @@ const StyledOverlay = toggleStyledComponent(
       max-width: calc(100vw - 2rem);
     }
 
-    &:where([data-variant='fullscreen']) {
```
siddharthkp, line 106: we have this in css now with ga, dont need to have it here as well

packages/react/src/Overlay/Overlay.module.css
```diff
@@ -157,12 +157,14 @@
...
 
-  &:where([data-variant='fullscreen']) {
-    top: 0;
-    left: 0;
-    width: 100vw;
-    height: 100vh;
-    margin: 0;
-    border-radius: unset;
+  &:where([data-responsive='fullscreen']) {
+    @media screen and (--viewportRange-narrow) {
```
siddharthkp, line 161: Note for reviewer: I changed full screen to only be a responsive thing instead of variant you can apply whenever

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -131,14 +131,17 @@ interface SelectPanelBaseProps {
     text: string | React.ReactElement
     variant: 'info' | 'warning' | 'error'
   }
-  onCancel?: () => void
 }
 
+// onCancel is optional with variant=anchored, but required with variant=modal
+type SelectPanelVariantProps = {variant?: 'anchored'; onCancel?: () => void} | {variant: 'modal'; onCancel: () => void}
```
siddharthkp, line 137: Note for reviewer: Moved the variant + onCancel prop combo here. You can always pass an onCancel but you especially must pass it with variant=modal

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -459,6 +464,16 @@ export function SelectPanel({
           role: 'dialog',
           'aria-labelledby': titleId,
           'aria-describedby': subtitle ? subtitleId : undefined,
+          ...(variant === 'modal'
+            ? {
+                /* override AnchoredOverlay position */
```
siddharthkp, line 469: Note for reviewer: This is where all the shame lies. We are still using an AnchoredOverlay, but then we override it's position by directly passing top, left, etc to overlayProps going down

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-Default-responsive-width-light-modern-action-list--true-linux.png
siddharthkp, line 1: Oooops, this feels wrong. BRB

siddharthkp, line 1: Note for reviewer: I don't know what changed here 😅 I ran it through a diff highlighter, it shows nothing but the second file is 26 bytes bigger.

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -132,10 +132,11 @@ interface SelectPanelBaseProps {
     variant: 'info' | 'warning' | 'error'
   }
   onCancel?: () => void
+  variant?: 'anchored' | 'modal'
 }
 
 export type SelectPanelProps = SelectPanelBaseProps &
-  Omit<FilteredActionListProps, 'selectionVariant'> &
+  Omit<FilteredActionListProps, 'selectionVariant' | 'variant'> &
```
siddharthkp, line 139: Note for reviewer: There used to be a `variant` prop in FilteredActionList but that's not used at all with SelectPanel. [react query with no results ↗](https://primer-query.githubapp.com/?rattribute=variant&reattribute=variant&repattribute=variant&repoattribute=variant&name=selectpanel&repo=github%2Fgithub&attribute=variant&entrypoint=%22%40primer%2Freact%22)

packages/react/src/SelectPanel/SelectPanel.tsx
siddharthkp: Tried to be smart, backfired in dotcom of course 😓

francinelucca:
@siddharthkp FYI these CI failures have nothing to do with this PR , https://github.com/primer/react/pull/5802#issuecomment-2744125066

e2e/components/SelectPanel.test.ts
```diff
@@ -112,7 +113,9 @@ test.describe('SelectPanel', () => {
   test(`Default @vrt responsive width .modern-action-list--true`, async ({page}) => {
     await visit(page, {
       id: 'components-selectpanel--default',
-      globals: {featureFlags: {primer_react_select_panel_with_modern_action_list: true}},
+      globals: {
+        featureFlags: {primer_react_select_panel_with_modern_action_list: true, primer_react_css_modules_ga: true},
+      },
```
francinelucca, line 118: is there a css modulers related regression here? 👀 , why is this one needed?

packages/react/src/Overlay/Overlay.tsx
francinelucca: I think let's keep this for now just in case, in theory the ga FF *could* still be turned off at any moment? The team can remove this when they remove the FF entirely from this component 🙏

packages/react/src/Overlay/Overlay.docs.json
francinelucca: I don't feel strongly either way, I don't see why not make it public though 👀

packages/react/src/SelectPanel/SelectPanel.tsx
francinelucca: fair

francinelucca (changes requested):
Mostly LGTM, just some questions/suggestions.

Was also wondering if you could elaborate on the feasibility of https://github.com/primer/react/pull/5230? I see from the comments it was reverted due to failing integration checks, wondering how significant those were to warrant going this route. I'm ok with this solution, was just curious

siddharthkp:
> Mostly LGTM, just some questions/suggestions.

Was also wondering if you could elaborate on the feasibility of https://github.com/primer/react/pull/5230? I see from the comments it was reverted due to failing integration checks, wondering how significant those were to warrant going this route. I'm ok with this solution, was just curious

It's definitely still feasible, but doesn't feel worth the effort anymore 🤔  I don't mind if the current solution stays long term too

e2e/components/SelectPanel.test.ts
siddharthkp: It's needed because i removed the styles from sx (your other comment), so it doesn't work without css modules enabled :)

packages/react/src/Overlay/Overlay.tsx
siddharthkp: I thought we weren't adding adding css for new features? Based on https://github.com/primer/react/pull/5761#discussion_r2000030071

I recently added this css in https://github.com/primer/react/pull/5759, so thought I'll clean it up as well. yay/nay?

francinelucca: yeah I think the issue here is the styles won't be added at all because if FF is not enabled the Overlay classname will not be added so these styles will never make it in (why you need that FF on that other test), the correct way of adding new features without sx styling is to always have it go to css modules, for example if you created a separate class and added it to both (enabled, not enabled) implementations

siddharthkp:
@emilybrick, Oops i only added the close icon button to the css module (feature flagged), not without it. Added it to both now!

On to radios

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -411,159 +414,176 @@ export function SelectPanel({
...
-      width={width}
-      anchorId={id}
-      variant={usingFullScreenOnNarrow ? {regular: 'anchored', narrow: 'fullscreen'} : undefined}
-      pinPosition={!height}
-      className={classes.Overlay}
-    >
-      <Box
-        sx={enabled ? undefined : {display: 'flex', flexDirection: 'column', height: 'inherit', maxHeight: 'inherit'}}
-        className={enabled ? classes.Wrapper : undefined}
+    <>
```
siddharthkp, line 417: Note for reviewer: Recommend to review this with `hide whitespace`

e2e/components/SelectPanel.test.ts
siddharthkp: Fixed :)

packages/react/src/SelectPanel/SelectPanel.tsx
hectahertz: I'd rather us have branching on the container component instead of doing this, as it semantically is much harder to implement. I think it's fine for now, but definitely worth improving.

hectahertz:
This should be complete now, refer to my commit history for all the changes and features added.

packages/react/src/SelectPanel/SelectPanel.tsx
francinelucca: ^ @hectahertz do we have an idea on the effort to do branching off the get go? I agree with you and we do have a bit of leeway time-wise with this, wondering if we can get it in 👀

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -857,3 +857,80 @@ export const WithOnCancel = () => {
...
+
+  const [selected, setSelected] = React.useState<ItemInput[]>(intialSelection)
+  const [filter, setFilter] = React.useState('')
+  const filteredItems = items.filter(
+    item =>
+      // design guidelines say to always show selected items in the list
+      selected.some(selectedItem => selectedItem.text === item.text) ||
+      // then filter the rest
+      item.text.toLowerCase().startsWith(filter.toLowerCase()),
+  )
```
francinelucca, line 872:
```suggestion
  const filteredItems = items.filter(
    item => item.text.toLowerCase().startsWith(filter.toLowerCase()),
  )
```

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -330,7 +340,11 @@ export function SelectPanel({
 
   const itemsToRender = useMemo(() => {
     return items.map(item => {
-      const isItemSelected = isMultiSelectVariant(selected) ? doesItemsIncludeItem(selected, item) : selected === item
+      const isItemSelected = isMultiSelectVariant(selected)
+        ? doesItemsIncludeItem(selected, item)
+        : isSingleSelectModal
+          ? intermediateSelected?.id === item.id
+          : selected?.id === item.id
```
francinelucca, line 347: can we offload this into a function? 🙏  this is hard to read

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -410,160 +428,186 @@ export function SelectPanel({
...
+            >
+              {footer}
+            </Box>
+          ) : showCancelSaveButtons ? (
+            <div className={clsx(classes.Footer, classes.ResponsiveFooter)}>
               <Button
                 size="medium"
                 onClick={() => {
-                  onCancel()
+                  onCancel && onCancel()
```
francinelucca, line 585:
```suggestion
                  onCancel?.()
```

packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -857,3 +857,80 @@ export const WithOnCancel = () => {
     </FormControl>
   )
 }
+
+export const AsMultiSelectModal = () => {
```
francinelucca, line 861: can we add these stories to vrt testing? 🙏

francinelucca (commented):
Some small suggestion/requests on the codebase. I wasn't able to test the UI since there's not a deployment available yet. I think you need to solve the merge conflicts @hectahertz

hectahertz:
Continued in #5883

PR #5818, opened 2025-03-25
Add workaround for Chrome bug where `KeybindingHint` in combination with `aria-labelledby` results in incorrect label

iansan5653:
`KeybindingHint` renders the key symbols as `aria-hidden` spans and the key names as `VisuallyHidden` spans to provide a good experience for both sighted and screen reader users. 

However, there is a bug in Chrome where `aria-hidden` text is included in a label generated using `aria-labelledby`. More context from https://github.com/github/accessibility/issues/8018#issuecomment-2751656621: 

> `KeybindingHint` (the Primer component we are using under the hood here) does try to account for this by rendering the symbol in an `aria-hidden` span and rendering the name of that symbol in an `sr-only` span next to it. However, there seems to be a bug in Chrome where the accessibility tree is incorrectly accounting for `aria-hidden` text in the target of an `aria-labelledby`. So we get the following (abridged) HTML:
> 
> <button class="IconButton" type="button" aria-labelledby=":rg:">
>   <svg>...</svg>
> </button>
> <span class="TooltipV2" aria-hidden="true" id=":rg:" popover="auto">
>   Add attachment
>   <span class="TooltipV2">
>     <span class="VisuallyHidden">(</span>
>     <kbd class="KeybindingHint" data-testid="keybinding-hint">
>       <span>
>         <span class="VisuallyHidden">command</span>
>         <span aria-hidden="true">⌘</span>
>         <span class="VisuallyHidden">shift</span>
>         <span aria-hidden="true">⇧</span>
>         <span class="VisuallyHidden">at</span>
>         <span aria-hidden="true">@</span>
>       </span>
>     </kbd>
>     <span class="VisuallyHidden">)</span>
>   </span>
> </span>
> This results in an incorrectly calculated accessible label of "Add attachment ( command ⌘ shift ⇧ at @ )".
> 
> If the label was being calculated correctly by the browser, we'd get a label of "Add attachment (command shift at)" (which looks wrong at first glance but sounds correct when read aloud by a screen reader - you don't say "command plus shift plus at" out loud).
> 
> Ideally we really want to fix this bug in the component so that consumers don't have to manually work around, so the approach used in [primer/react#5626](https://github.com/primer/react/pull/5626) works but it isn't ideal - it relies on consumers doing things correctly.
> 
> I don't think the way `KeybindingHint` itself works can really be improved much. If we were to update it to render the hidden keys and label next to each other, rather than mixed in, it would slightly improve what is being read but the data would still be duplicated. `KeybindingHint`'s approach does work well in cases that don't involve `aria-labelledby` - when read as part of the document flow the output is correct.
> 
> I think the solution probably has to be in `IconButton` or `TooltipV2`. `KeybindingHint` does export a helper function that gives you the plain accessible string representation of a shortcut, so we could build the `aria-label` string ourselves rather than using `aria-labelledby` in `IconButton`. Or, a more robust approach might be to update `TooltipV2` to move the `id` prop into a child element so that the tooltip could better control exactly what makes it into the label; then we could move `KeybindingHint` outside that element and add an `sr-only` element inside with just the string representation of the shortcut.

To work around this bug, I've taken the approach proposed in the quote to update `TooltipV2` to exclude the keybinding hint from the id-tagged container. Then I added the keybinding hint text as a plain hidden string inside that container.

Before in VoiceOver/Chrome:

VoiceOver text readout reading 'link, Learn More( shift ⇧ question mark ?)

After:

VoiceOver text readout reading 'link, Learn More (shift question mark)

Note this isn't a perfect example because obviously "question mark" is combined with "shift", but it's pulled from storybook so I just kept it. Intuitively you might think the label should be "shift + question mark", but when read by a screen reader this can result in the awkward label "shift plus question mark" - it's better to just exclude the punctuation here.

#### Changed

Fixed a Chrome bug impacting accessible labels rendered by `TooltipV2` with the `keybindingHint` prop.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Button/__tests__/Button.test.tsx
+++ b/packages/react/src/Button/__tests__/Button.test.tsx
@@ -338,7 +338,7 @@ describe('Button', () => {
   })
   it('should append the keyshortcuts to the tooltip text that labels the icon button when keyshortcuts prop is passed', () => {
     const {getByRole} = render(<IconButton icon={HeartIcon} aria-label="Heart" keyshortcuts="Command+H" />)
-    const triggerEl = getByRole('button', {name: 'Heart ( command h )'})
+    const triggerEl = getByRole('button', {name: 'Heart (command h)'})
     expect(triggerEl).toBeInTheDocument()
   })
   it('should render aria-keyshortcuts on an icon button when keyshortcuts prop is passed (Description Type)', () => {
@@ -353,6 +353,6 @@ describe('Button', () => {
       <IconButton icon={HeartIcon} aria-label="Heart" description="Love is all around" keyshortcuts="Command+H" />,
     )
     const triggerEl = getByRole('button', {name: 'Heart'})
-    expect(triggerEl).toHaveAccessibleDescription('Love is all around ( command h )')
+    expect(triggerEl).toHaveAccessibleDescription('Love is all around (command h)')
   })
 })
--- a/.changeset/curvy-ties-breathe.md
+++ b/.changeset/curvy-ties-breathe.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Add workaround for Chrome bug where `KeybindingHint` in combination with `aria-labelledby` results in incorrect label
--- a/packages/react/src/KeybindingHint/KeybindingHint.tsx
+++ b/packages/react/src/KeybindingHint/KeybindingHint.tsx
@@ -26,8 +26,7 @@ export const KeybindingHint = memo(({className, ...props}: KeybindingHintProps) 
 KeybindingHint.displayName = 'KeybindingHint'
 
 /**
- * AVOID: `KeybindingHint` is nearly always sufficient for providing both visible and accessible keyboard hints, and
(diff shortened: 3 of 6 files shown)


joshblack (approved):

packages/react/src/Button/__tests__/Button.test.tsx
```diff
@@ -353,6 +353,6 @@ describe('Button', () => {
       <IconButton icon={HeartIcon} aria-label="Heart" description="Love is all around" keyshortcuts="Command+H" />,
     )
     const triggerEl = getByRole('button', {name: 'Heart'})
-    expect(triggerEl).toHaveAccessibleDescription('Love is all around ( command h )')
+    expect(triggerEl).toHaveAccessibleDescription('Love is all around (command h)')
```
iansan5653, line 356: Note: This is indicative of the one potential breaking change I see for dotcom - we may need to update some test selectors to remove spaces around the keybinding hint in the labels. This is an intentional change.

PR #5142, opened 2024-10-17
SelectPanel: Implement empty state 

broccolinisoup:
Closes https://github.com/github/primer/issues/3922, https://github.com/github/primer/issues/4104

Implementing empty state for SelectPanel. Default message and the ability to customise them when there is no item to show either in the first load or when filtering (Under `primer_react_select_panel_with_modern_action_list` FF).

#### New

- SelectPanel.Message child component
- When there is no item to show after the initial load, we now show a message. There is a default message and ability to customise it.

```jsx
<SelectPanel .....>
  <SelectPanel.Message variant="empty" title={`No label found`}>
    Start creating some labels
  </SelectPanel.Message>
</SelectPanel>
```

- When there is no item to show after user filters the result, we now show a message. There is a default message and ability to customise it.

```jsx
<SelectPanel .....>
  <SelectPanel.Message variant="noResults" title={`No language found for `}>
    Adjust your search term to find other languages
  </SelectPanel.Message>
</SelectPanel>
```

##### No initial Item

Screenshot of the select panel that doesn't have any items to show. There is a message says You haven't created any projects yet. Start your first project to organise your issues.

##### No filtered Item demo

Video description:

https://github.com/user-attachments/assets/ed7e408c-960f-46da-9adb-7647bb5561aa

#### Changed

- SelectPanel accepts children now

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Testing & Reviewing

[...]

PR #5826, opened 2025-03-26
Feat/select panel empty state

francinelucca:
Closes https://github.com/github/primer/issues/3922, https://github.com/github/primer/issues/4104

Implementing empty state for SelectPanel. Default message and the ability to customise them when there is no item to show either in the first load or when filtering (Under `primer_react_select_panel_with_modern_action_list` FF).

Resubmission of https://github.com/primer/react/pull/5142 that had to be reverted

#### New

- SelectPanel now accepts a message prop to display error, warning and empty states through a new (internal) SelectPanelMessage component
- FilteredActionList (modern) now accepts a message prop to render instead of the list for empty states
- Stories demonstrating new empty, warning and error states functionality
- Tests for new empty states functionality
- Container css module class for FilteredActionList

##### No initial Item

Screenshot of the select panel that doesn't have any items to show. There is a message says You haven't created any projects yet. Start your first project to organise your issues.

##### No filtered Item demo

Video description:

https://github.com/user-attachments/assets/ed7e408c-960f-46da-9adb-7647bb5561aa

#### Changed

- Offloaded a lot of unnecessary announce logic for ModernActionList from SelectPanel to FilteredActionListWithDeprecatedActionList 
- SelectPanel now uses de-facto ModernActionList announcement logic through useAnnouncements
- Adjusted SelectPanel stories to add empty states
- Corrected some tests and added fakeTimers were needed due to new announcement logic
- Renamed incorrect story "Height Variantions and Scroll" -> "Height Variations and Scroll"
- VRT updates to account for corrected story name
- Fixed loading bug where loading spinner was being displayed when no results where available on the panel (items had already been loaded but filter applied with no matching items)

#### Removed

- Removed usage of `LiveRegion` component that seemed unnecessary as announcements are already being handled programmatically

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- [This](https://primer-64c6e73b6f-13348165.drafts.github.io/storybook/?path=/story/components-selectpanel-features--customised-no-initial-items&globals=featureFlags.primer_react_select_panel_with_modern_action_list:!true) and [This](https://primer-64c6e73b6f-13348165.drafts.github.io/storybook/?path=/story/components-selectpanel-features--customised-no-results&globals=featureFlags.primer_react_select_panel_with_modern_action_list:!true) are a good stories to test the behaviour.  Please note `primer_react_select_panel_with_modern_action_list` FF must be on

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
```diff
@@ -46,6 +48,68 @@ const StyledHeader = styled.div`
...
+      from: liveRegion ? liveRegion : undefined, // announce will create a liveRegion if it doesn't find one
+    })
+  },
+  250,
+)
+
+async function announceFilterFocused() {
+  await announceText('Focus on filter text box and list of items')
+}
+
```
francinelucca, line 112: moved all this logic that was making the SelectPanel component bulky down to the FilteredActionList so we can get rid of it when we finally remove it 💃

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -277,12 +238,11 @@ export function SelectPanel({
   useEffect(() => {
     if (open) {
       if (items.length === 0) {
-        announceNoItems()
-      } else {
-        if (listContainerElement) {
-          announceItemsChanged(items, {current: listContainerElement})
+        // we need to wait for the listContainerElement to disappear before announcing no items, otherwise it will be interrupted
+        if (!listContainerElement || !usingModernActionList) {
+          announceNoItems(message?.title)
```
francinelucca, line 243: this is the best I got 😅 , there's some type of race condition in between when the list disappears and the empty state is rendered, this is what I came up with to prevent the announcement from getting killed. Waiting until the list disappears before announcing

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -295,7 +255,7 @@ export function SelectPanel({
       return
     }
 
-    if (isLoading) {
+    if (isLoading || items.length > 0) {
```
francinelucca, line 258: noticed a bug in prod where it was setting the loading screen on empty when items had already been loaded but there were no results after filtering

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -311,24 +271,13 @@ export function SelectPanel({
...
       const ref = inputRef.current
-      const listener = () => {
-        announceFilterFocused()
-      }
-
-      if (document.activeElement === ref) {
-        listener()
-      }
-
-      ref.addEventListener('focus', listener)
```
francinelucca, line 322: useAnnouncements() is doing this for us now

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -448,167 +398,172 @@ export function SelectPanel({
...
-        {usingModernActionList ? null : (
-          <Message
-            value={
-              filterValue === ''
-                ? 'Showing all items'
-                : items.length <= 0
-                  ? 'No matching items'
-                  : `${items.length} matching ${items.length === 1 ? 'item' : 'items'}`
-            }
-          />
```
francinelucca, line 483: Removing this in favor of the regular announcements, not sure why he had this on top of the announcements, I tested without it and all looks good, seems to me that we were over-announcing on deprecated

packages/react/src/SelectPanel/SelectPanel.tsx
TylerJDev: I think this works! Maybe we can follow up with a centralized place to store announcements? I think once we remove the feature flag stuff, it'll be much easier to tackle this since we don't have to rely on as many conditionals.

PR #5837, opened 2025-03-28
Toggling who can bypass the merge queue with the deployment lock

jonrohan:
This updates the lock_release.yml workflow to toggle bypass restrictions on the merge queue.

Removing the ability to bypass the merge queue when the release is unlocked will remove the bypass check that will always be present. Helping the conductor only be able to bypass when a lock is in place.

I also added another event to unlock. When a release is `released` it will automatically unlock the queue.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,6 +8,9 @@
   "features": {
     "ghcr.io/devcontainers/features/sshd:1": {
       "version": "latest"
+    },
+    "ghcr.io/devcontainers/features/github-cli": {
+      "version": "latest"
     }
   },
   "hostRequirements": {
--- a/.github/workflows/lock-release.yml
+++ b/.github/workflows/lock-release.yml
@@ -35,25 +35,9 @@ jobs:
             -f "enforcement=${enforcement}"
         env:
           GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
-      # - name: Toggle Rule > Update Before Merging
-      #   run: |
-      #     enforcement=$([ "${{ github.event.inputs.action }}" == "lock" ] && echo "active" || echo "disabled")
-      #     gh api \
-      #       --method PUT \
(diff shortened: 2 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -8,6 +8,9 @@
   "features": {
     "ghcr.io/devcontainers/features/sshd:1": {
       "version": "latest"
+    },
+    "ghcr.io/devcontainers/features/github-cli": {
+      "version": "latest"
```
jonrohan, line 13: I wanted to add `gh` cli command to the codespace

.github/workflows/lock-release.yml
```diff
@@ -1,6 +1,8 @@
-name: Lock Release Pull Request
+name: Toggle Release Lock
 
 on:
+  release:
+    types: [released]
```
jonrohan, line 5: I want to unlock when a release is merged

.github/workflows/lock-release.yml
```diff
@@ -24,36 +27,52 @@ jobs:
...
-          enforcement=$([ "${{ github.event.inputs.action }}" == "lock" ] && echo "active" || echo "disabled")
+          # Allow react-release-conductor to bypass merge queue
+          gh api \
+            --method PUT \
+            -H "Accept: application/vnd.github+json" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            /repos/primer/react/rulesets/4089335 \
+            -F "bypass_actors[][actor_id]=12276524" \
+            -f "bypass_actors[][actor_type]=Team" \
+            -f "bypass_actors[][bypass_mode]=always"
```
jonrohan, line 40: This lets react-release-conductor bypass the merge queue when lock is in effect. We have to do this because if we leave the merge queue on the checkbox won't showup to bypass unless the conductor can also bypass the queue.

francinelucca (approved):

PR #5867, opened 2025-04-03
fix(SelectPanel): selected items should appear at the top

francinelucca:
Closes https://github.com/github/primer/issues/2409

Adds logic to SelectPanel to automatically sort items, displaying the selected items first. The sorted list gets rearranged on:
- Filter value is cleared
- Items have changed
- Panel is reopened

Also adds capabilities for custom sorting, providing a sorting key, sort by ascending or descending and opt-out of "selected on top" functionality.

by default, selected items will be displayed at the top when primer_react_select_panel_order_selected_at_top FF is on

#### New

- Added sorting logic to SelectPanel
- sortKey, sortFn, sortDirection and orderSelectedFirst props in SelectPanel
- New Jest tests for sorting
- New stories for sorting
- Added "sortKey" to all SelectPanel stories so that they're sorted alphabetically

#### Changed

- Update VRT tests
- Update Jest tests to account for new sorting logic
- Update ItemProps interface to extend from HtmlElement

#### Removed
- Custom sorting logic from SelectPanel stories as it is no longer needed

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

[SelectPanel stories
](https://primer-341ecc4c34-13348165.drafts.github.io/storybook/?path=/story/components-selectpanel--default)
* Ensure to turn on modern_action_list FF, as well with primer_react_select_panel_fullscreen_on_narrow and primer_react_select_panel_order_selected_at_top FF

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -328,40 +340,106 @@ export function SelectPanel({
...
+      .sort((itemA, itemB) => {
+        // itemA is selected (for sorting purposes) if an object in selectedOnSort matches every property of itemA, except for the selected property
+        const itemASelected = selectedOnSort.some(item =>
+          Object.entries(item).every(([key, value]) => {
+            if (key === 'selected') {
+              return true
+            }
+            return itemA[key as keyof ItemProps] === value
+          }),
+        )
```
francinelucca, line 414: this seems expensive :(

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -328,40 +340,106 @@ export function SelectPanel({
...
+        } else if (itemASelected < itemBSelected) {
+          return 1
+        }
+
+        // Then order by text
+        if ((itemA.text ?? '') < (itemB.text ?? '')) {
+          return -1
+        } else if ((itemA.text ?? '') > (itemB.text ?? '')) {
+          return 1
+        }
```
francinelucca, line 438: I baked in alphabetically ordering through text but now that I'm looking at the [issue](https://github.com/github/primer/issues/2409) it doesn't really look like a requirement 🤔 
I can remove and do selection-at-top only if this is not wanted/needed 👀 
CC: @emilybrick

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -328,40 +340,106 @@ export function SelectPanel({
...
-    })
-  }, [onClose, onSelectedChange, items, selected])
+            // single select
+            const singleSelectOnChange = onSelectedChange as SelectPanelSingleSelection['onSelectedChange']
+            singleSelectOnChange(item === selected ? undefined : item)
+            onClose('selection')
+          },
+        } as ItemProps
+      })
+      .sort((itemA, itemB) => {
```
francinelucca, line 405: I'm coding this as the default behavior. Wondering if we should allow consumer to opt-out of this behavior or provide their custom sort function 👀 
cc: @emilybrick

.playwright/snapshots/components/SelectPanel.test.ts-snapshots/SelectPanel-Above-Tall-Body-dark-colorblind-linux.png
francinelucca, line 1: visual updates are due to items being sorted now

packages/react/src/SelectPanel/SelectPanel.tsx
mperrotti: I feel strongly that we should let consumers opt-out by providing their own sort function. I think DataTable has a nice API for handling this: https://github.com/primer/react/blob/main/packages/react/src/DataTable/column.ts#L59

packages/react/src/SelectPanel/SelectPanel.tsx
mperrotti: Consumers could want to sub-sort based on something besides alphabetical order. For example: sort most frequently used labels to the top.

I think we should remove this default alphanumeric sort.

emilybrick:
This looks good visually - though I think we should remove the border that appears under the selected items. The items being selected differentiates them enough from the normal list. 
Screenshot 2025-04-08 at 12 40 11 PM

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -213,6 +213,22 @@
       }
     }
 
+    &:where([data-last-selected]:not(:last-of-type)) {
```
mperrotti, line 216: I like this CSS-only solution for visually separating selected and unselected items, but where did this design change come from? We didn't do this before.

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -213,6 +213,22 @@
...
+      &::before {
+        position: absolute;
+        bottom: 0;
+        width: calc(100% + var(--base-size-16));
+        margin-left: calc(-1 * var(--base-size-8));
+        content: '';
+        /* stylelint-disable-next-line primer/borders */
+        border-bottom: 1px solid var(--borderColor-default);
+        border-bottom-right-radius: 0;
+        border-bottom-left-radius: 0;
```
mperrotti, line 228: A shorter way to do this would be:

```suggestion
        position: absolute;
        content: '';
        inset: 100% calc(-1 * var(--base-size-8));
        background-color: var(--borderColor-default);
        height: var(--borderWidth-default);
```

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -213,6 +213,22 @@
       }
     }
 
+    &:where([data-last-selected]:not(:last-of-type)) {
+      padding-bottom: var(--base-size-4);
```
mperrotti, line 217: Why do we reduce the padding for these items?

mperrotti (commented):
I think it would feel better if the order of the button children matched the order of the selected items. How can we help consumers keep the ordering in sync?

Screenshot 2025-04-08 at 12 20 00 PM

packages/react/src/SelectPanel/SelectPanel.tsx
joshblack: Wanted to ask if it would be possible for the sort order to be maintained as a stateful value and compute it `onClose` or `onOpen` instead of having to run the sort every render?

It seemed like the logic that we have for this is something like:

- When open, keep the ordering of items constant
- When closing (or when re-opening), sort the items by:
  - selected items come first
    - sort selected items by group, if applicable
    - otherwise, sort by alphanumeric
  - otherwise, sort by alphanumeric

Does that track with what you're seeing?

packages/react/src/SelectPanel/SelectPanel.tsx
joshblack: Feel free to steal the alphanumeric stuff from data table btw if it's helpful! We have some sort helpers over there

joshblack (commented):
Just had some initial comments/questions, let me know if they make sense lol. Also happy to pair on this if you want!

packages/react/src/ActionList/ActionList.module.css
francinelucca: here's the issue https://github.com/github/primer/issues/2409!

packages/react/src/ActionList/ActionList.module.css
francinelucca: I'm actually adding padding to separate from the divider line, this is only added to the last selected item for such purposes

francinelucca:
> Wanted to ask if it would be possible for the sort order to be maintained as a stateful value and compute it onClose or onOpen instead of having to run the sort every render?
> It seemed like the logic that we have for this is something like:

>When open, keep the ordering of items constant
When closing (or when re-opening), sort the items by:
selected items come first
sort selected items by group, if applicable
otherwise, sort by alphanumeric
otherwise, sort by alphanumeric
Does that track with what you're seeing?

A big one is we want to reset the sorting when the filter value is cleared, we also do want to sort on the first open, so I don't think it's as simple as onClose/onOpen. There's definitely some cleanup we might be able to do if we keep a state of the mapped items though 👀

francinelucca:
> https://github.com/primer/react/pull/5867#issuecomment-2787059146

@emilybrick double-checking, you're saying remove the line separation entirely, just keep the ordering on top functionality?

packages/react/src/ActionList/ActionList.module.css
mperrotti: I still don't fully understand why it's 4px instead of 6px 🙈 

Maybe we you can screen-share and explain?

packages/react/src/ActionList/ActionList.module.css
francinelucca: Done, thanks!

emilybrick:
> > [#5867 (comment)](https://github.com/primer/react/pull/5867#issuecomment-2787059146)
> 
> @emilybrick double-checking, you're saying remove the line separation entirely, just keep the ordering on top functionality?

Yes. Keep the ordering at top functionality, remove the line entirely. I don't think it's doing much. /cc @mperrotti for a gut check but I don't think we need it.

emilybrick (approved):
This looks good to me, selected items move to top/appear at top + in the same order as they appear in the button. WIth border removed. Thank you!

.changeset/itchy-bears-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
```
joshblack, line 2:
```suggestion
"@primer/react": minor
```

With what we're adding into SelectPanel what do you think about having this as a `minor` since we're adding in functionality?

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -347,13 +406,32 @@ export function SelectPanel({
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [open])
+
+  useEffect(() => {
+    if (prevItems.length === 0 && items.length > 0) {
+      resetSort()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [items, prevItems])
```
joshblack, line 431: Instead of synchronizing with an effect one idea is to:

```suggestion
  const [previousItems, setPreviousItems] = useState(items);
  
  if (previousItems !== items) {
    setPreviousItems(items);
    if (previousItems.length === 0 && items.length > 0) {
      resetSort()
    }
  }
```

Based on: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -347,13 +406,32 @@ export function SelectPanel({
...
+  }, [placeholder, renderAnchor, selected, sortDirection, sortFn, sortKey])
 
-  const itemsToRender = useMemo(() => {
-    return items.map(item => {
+  useEffect(() => {
+    if (open) {
+      resetSort()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [open])
```
joshblack, line 424: We could potentially handled this this like in: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes to avoid having to use an effect here. Another idea is to call `resetSort()` when `onClose()` is called (or whatever appropriate prop it would be for when it open/closes)

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -385,8 +463,56 @@ export function SelectPanel({
         },
       } as ItemProps
     })
+
+    setItemsToRender(itemsToRender)
```
joshblack, line 467: Could this and the effect below potentially move into just the render function instead of effects or are we experiencing some perf problems when that happens?

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -93,6 +93,10 @@ interface SelectPanelBaseProps {
     variant: 'empty' | 'error' | 'warning'
   }
   onCancel?: () => void
+  orderSelectedFirst?: boolean
```
joshblack, line 96:
```suggestion
  showSelectedOptionsFirst?: boolean
```

Just a naming idea, totally get if it doesn't work as well as `orderSelectedFirst` 👍

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -93,6 +93,10 @@ interface SelectPanelBaseProps {
...
+  sortKey?: keyof ItemInput
+  sortDirection?: 'asc' | 'desc'
+  sortFn?: (a: ItemInput, b: ItemInput) => number
```
joshblack, line 99: Could you share more about baking these directly into select panel? Put another way, folks could just sort the array however they wanted to before but now they would need to hook into SelectPanel's API to do that. 

This might be a preference for folks downstream or on the team, totally understand if so 👍 It wasn't immediately clear to me why we would do this and just wanted to ask.

joshblack (commented):
Left some comments around the effect stuff, feel free to ignore me if this is getting annoying 😅 lol

packages/react/src/SelectPanel/SelectPanel.tsx
francinelucca: done!

packages/react/src/SelectPanel/SelectPanel.tsx
francinelucca: my logic was we're already sorting in order to display selected at the top so I was thinking instead of doing (potentially) a double sort, one at consumer-side and one on our side, we could just do it all together, so that's where `sortKey` and `sortDirection` come in `sortFn` comes from @mperrotti [suggestion](https://github.com/primer/react/pull/5867#discussion_r2033552691) that we should allow for custom sort functions.
I also thought providing these could offload work from the consumer and make it "easier", I could be wrong though 👀 
>  Put another way, folks could just sort the array however they wanted to before but now they would need to hook into SelectPanel's API to do that

these are all opt-in, folks could totally still do their own sorting before hand and everything should work as expected.

All that being said, these three props can totally be removed and this PRs purpose will still be served.

packages/react/src/SelectPanel/SelectPanel.tsx
francinelucca: Yeah we can, I put it in a useEffect as a response to your comment here https://github.com/primer/react/pull/5867#discussion_r2033979938, (also see https://github.com/primer/react/pull/5867#issuecomment-2787771215) so we didn't have to sort on every render 👀

francinelucca:
closing in favor of https://github.com/primer/react/pull/5971

PR #5899, opened 2025-04-07
feat: add support for size to Blankslate

joshblack:
Closes https://github.com/github/primer/issues/4863

#### New

- Add support for `size` to `Blankslate`

#### Changed

- Update overall styling to align with intended usage in Figma

### Rollout strategy

- [x] Minor release


packages/react/src/Blankslate/Blankslate.stories.tsx
```diff
@@ -50,4 +50,12 @@ Playground.args = {
   border: false,
   narrow: false,
   spacious: false,
+  variant: 'default',
+}
+Playground.argTypes = {
+  variant: {
+    controls: {
+      options: ['default', 'large', 'small'],
```
lukasoppermann, line 58: Let's change `default` to `medium`

packages/react/src/Blankslate/Blankslate.tsx
```diff
@@ -20,45 +25,61 @@ export type BlankslateProps = React.PropsWithChildren<{
    */
   spacious?: boolean
 
-  className?: string
-}>
+  /**
+   * Specify the variant of this component
+   */
+  variant?: 'small' | 'default' | 'large'
```
lukasoppermann, line 31:
```suggestion
  variant?: 'small' | 'medium' | 'large'
```

We would need to change this everywhere.

lukasoppermann (approved):
👍 LGTM

Is there any way we can get it out of experimental? I mean, is it really experimental?

lukasoppermann (changes requested):
Actually I noticed that the icon size for `small` should be `24px`, but it is 16 at the moment. Can we change this?

francinelucca (approved):
LGTM as long as the visual changes are expected 👀

lukasoppermann:
Some more updates were added to the Figma version, specifically to the small variant: https://www.figma.com/design/GCvY3Qv8czRgZgvl1dG6lp/Primer-Web?node-id=4398-2383&t=T3iXg81O8v9mxzrZ-4

joshblack:
@lukasoppermann just made the two updates! (Icon and small variant, not spacious padding)

lukasoppermann:
Hey @joshblack,

great. Some minor points:
1. Could you add a border to the stories with narrow and spacious, as this makes it easier to see the difference.
2. The `small` variant should use a `small` button variant as well. 
3. Maybe we can make primary and secondary action togglable in the playground story?

lukasoppermann (changes requested):
Great changes, all that is left is to make sure that in the small variant, the actions use the small variant as well.

joshblack:
Thanks @lukasoppermann! Made the following changes:

- [x] Add border to stories with narrow and spacious
- [x] Add support for `small` button action in `small` blankslate
- [x] Add `primaryAction` and `secondaryAction` knob to playground to toggle primary/secondary actions

lukasoppermann (approved):

PR #5913, opened 2025-04-10
Update data table pagination pageEnd to handle zero based index

### LINKED ISSUE #5909: Paginated data table pagination range bug
### Description

Reported by @AAugustine:

> Hi, I've noticed something while using the [Paginated Data Table ](https://primer.style/product/components/data-table/#with-paginated-rows)that seems very bug like. When the number of rows is one more than the page size, the count in the lower left hand corner is off by one. 

### Steps to reproduce

> To reproduce it in the example page, you just update the for loop to return only 11 rows instead of 1000 and the count in the lower left will show "1-11" instead of "1-10". It only seems to do it when it it one over the page size, because when you change it to 12, the count is correct and shows "1-10" on the first page.

Image
Image

### Version

37.18.0

### Browser

Chrome

AAugustine:
Closes #5909 

#### New

A test that specifically tests the failing case on when the total count is 1 + pageSize

#### Changed

Changed how pageEnd was being calculated. It now makes pageEnd represent the actual display rather than a zero-based index.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/weak-spiders-love.md
+++ b/.changeset/weak-spiders-love.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Update data table pagination pageEnd to handle zero based index
--- a/packages/react/src/DataTable/Pagination.tsx
+++ b/packages/react/src/DataTable/Pagination.tsx
@@ -367,7 +367,7 @@ type RangeProps = {
 
 function Range({pageStart, pageEnd, totalCount}: RangeProps) {
   const start = pageStart + 1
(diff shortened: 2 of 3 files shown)


hectahertz (approved):
Looks good! TYSM for the contribution!

francinelucca (approved):

PR #5916, opened 2025-04-11
docs: add css layers adr

joshblack:
Add ADR talking about our usage of `:where()` and how we could use CSS layers to accomplish a similar goal.

Curious to hear what you all think! Let me know if you have any comments/questions 👀 

#### New

- Add ADR for CSS layers


contributor-docs/adrs/adr-021-css-layers.md
```diff
@@ -0,0 +1,104 @@
...
+
+Using a CSS layer in this way allows us to author our CSS without having to
+consider whether or not to use `:where()`. This is due to the nature of CSS
+layers where if two layers are targetting the same element, the last layer
+targetting that element will take precedence.
+
+Since we are moving our CSS into a CSS layer, this means that later layers (in
+particularly the anonymous layer) will take precedence when two rules are
+matching the same element. For example:
+
```
francinelucca, line 90: I think the example is missing here?

francinelucca (approved):
Love the idea! Just one comment

I'd love to be able to just code the css without having to worry about the `where()`

jonrohan (commented):
Where does this fit with the current `@primer-react` [layer in dotcom](https://github.com/github/github/blob/26587cf2bd3798ffd5f8bfa382452502bfde3a02/ui/packages/client-build-plugins/primer-react-css-layer.js#L15)?

Could we do something like this, so we could [define in dotcom](https://github.com/github/github/blob/26587cf2bd3798ffd5f8bfa382452502bfde3a02/app/assets/stylesheets/bundles/primer-primitives/layers.css#L2) the priority of the layers for all primer? `@layer primer-css-base, primer, recipes;`

```css
@layer primer {
  @layer components {
     @layer ComponentName {
          ....
      }
   }
}
```

joshblack:
@jonrohan I bet we could remove the layer upstream now that our stuff could be reliably in a layer. Thankfully I think even if we do no change it should still be okay.

> Could we do something like this, so we could [define in dotcom](https://github.com/github/github/blob/26587cf2bd3798ffd5f8bfa382452502bfde3a02/app/assets/stylesheets/bundles/primer-primitives/layers.css#L2) the priority of the layers for all primer?

And yeah! I think so. That's probably one of the nicest parts about this is that we could also have our own internal logic too for ordering of components (like links always come after breadcrumbs)

contributor-docs/adrs/adr-021-css-layers.md
joshblack: Thanks 😅 Updated!

jonrohan (approved):

TylerJDev (approved):

PR #5918, opened 2025-04-11
test(e2e): update Banner to take screenshots off of the storybook root

joshblack:
Update our e2e test setup for Banner to only screenshot the component instead of the overall page. This helps to reduce our overall screenshot sizes and also makes it easier to detect differences in the UI for reviewing.

#### Changed

- Update our e2e tests for Banner to only snapshot the storybook root
- Update padding in our global storybook styles so that padding is added on the storybook root

### Rollout strategy

This is a change to how we snapshot in an e2e test and does not impact the public API


hectahertz (approved):
Oh, lovely! Now we have to do all the others lol

PR #5929, opened 2025-04-15
Add the framework required to run component stress tests

hectahertz:
#### New

This PR adds the framework required to run component stress tests and a couple of examples.

The tests run on Storybook and measure the time it takes for a component to react to a state change.

This is how one of them looks like:

https://github.com/user-attachments/assets/c29c7ad0-ed02-4468-96b5-ab1f6c4634e4

I tried to make the code self explanatory but I'll give a brief overview of how the pieces fit together here.

`StressTest.tsx` is a helper component to be used in Storybook, that takes care of all the performance measuring part of the stress tests. It has a very simple API so that defining the tests is as simple as possible.
It uses `afterframe` and a React state update to do this. We trigger a state update, which the component will react to and rerender appropriately according to the test defined. The afterFrame library calls the function when the next frame starts and measures the elapsed time. The reasoning behind this approach is well explained [in this article](https://3perf.com/blog/react-monitoring/#why-use-afterframe).

`[Component].stress.stories.tsx` contain the stress tests for a few components. The components will react to the state change by interacting with the `count` prop provided in the render function. This reaction is what the stress test will measure.

`stress-tests.yml` defines the workflow that runs the stress tests. It runs the Playwright tests against the Storybook server and stores the results in a file. This file is stored in the cache and compared to the previous results. Each commit will generate a new result in the cache.

If the result of any commit is slower than the alert threshold defined there, the workflow will fail and a comment will be added to the commit with the results.

Screenshot 2025-04-15 at 20 46 39

`StressTests.test.ts` is where the Playwright tests are defined. The tests just run the Storybook story and grab the mean result from them.

`stress-test-reporter.ts` gets the results from the Playwright tests and stores them in a file, in a format that `benchmark-action/github-action-benchmark` expects. 

#### How much of a stable environment do you need for this to be reliable?

So far, defining `runs-on: ubuntu-latest-4-cores` on the workflow has been very reliable. Running the tests locally is also very consistent from run to run but, as you can imagine, it will only be useful to compare numbers between runs on the same hardware.

#### How accurate is it at detecting regressions?

Very accurate according to my tests. Here's an example of enabling and disabling CSS modules on ActionList:

https://github.com/user-attachments/assets/d67231a3-9a04-4b05-98d7-a942a6da912b

### Rollout strategy

- [ ] Patch release
- [X] Minor release

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/ActionList.stress.stories.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
+
+import type {Meta} from '@storybook/react'
+import type {ComponentProps} from '../utils/types'
+import {StressTest} from '../utils/StressTest'
+import {TableIcon} from '@primer/octicons-react'
+import {ActionList} from '.'
+
+export default {
+  title: 'StressTests/Components/ActionList',
```
francinelucca, line 10: is there a way to make the StressTests folder not appear in production? deployments similar to the Dev stories? or is there a reason we'd want that in production?

.github/workflows/stress-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+          # Where the previous data file is stored
+          external-data-json-path: ./cache/stress-tests-benchmark-data.json
+          # Workflow will fail when an alert happens
+          fail-on-alert: true
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          alert-threshold: '150%'
+          # Enable alert commit comment
+          comment-on-alert: true
+          # Mention @rhysd in the commit comment
+          alert-comment-cc-users: '@hectahertz'
```
francinelucca, line 55: are you sure you want to be CC'd in every comment? 😅

francinelucca (approved):
Very excited for this! Just two non-blocking questions

.github/workflows/stress-tests.yml
hectahertz: Hahaha, yes for now, The plan is to remove it once we've battle tested this enough

packages/react/src/ActionList/ActionList.stress.stories.tsx
hectahertz: Oh let me take a look!

I think they are fine either way though. I always thought transparency is a good policy there and if I were a developer looking into the design system I'd appreciate having them.

joshblack (commented):
Thanks for putting this together! Really the only question I had with this was around StressTest and afterframe and what differences there are between setting up a harness with this versus using something like `Profiler` with React.

PR #5981, opened 2025-04-28
chore(BaseStyles): Remove the CSS modules feature flag from the BaseStyles component

jonrohan:
Closes https://github.com/github/primer/issues/4358

#### Removed

Remove the CSS modules feature flag from the BaseStyles component

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


results.json
francinelucca, line 1: IDK what's up with adding this file when VRT runs :(, can we remove?

packages/react/src/__tests__/BaseStyles.test.tsx
```diff
@@ -20,7 +20,7 @@ describe('BaseStyles', () => {
     expect(container).toMatchSnapshot()
   })
 
-  it('respects styling props', () => {
+  it.skip('respects styling props', () => {
```
francinelucca, line 23: why do we need to skip these now? 🤔

packages/react/src/BaseStyles.tsx
```diff
@@ -52,77 +19,29 @@ export type BaseStylesProps = PropsWithChildren & {
...
-    const newClassName = clsx(classes.BaseStyles, className)
-    const baseStyles = {
-      ['--BaseStyles-fgColor']: color,
-      ['--BaseStyles-fontFamily']: fontFamily,
-      ['--BaseStyles-lineHeight']: lineHeight,
-    }
-
-    // If props includes TYPOGRAPHY or COMMON props, pass them to the Box component
-    if (includesSystemProps(props)) {
-      const systemProps = getTypographyAndCommonProps(props)
```
francinelucca, line 70: Can you clarify why this was needed before but not anymore? I'm confused since this seems to be the old-enabled version 😅

francinelucca (commented):
some questionsss

packages/react/src/BaseStyles.tsx
jonrohan: This was added to check if system props were coming in, but we've abstracted that same functionality to the `BoxWithFallback` component so no need to double check now https://github.com/primer/react/blob/36f100706a32d369d1f464caeaa60ff38b576d30/packages/react/src/internal/components/BoxWithFallback.tsx#L11

packages/react/src/__tests__/BaseStyles.test.tsx
jonrohan: The `.toHaveStyle` check doesn't work anymore because of the CSS modules not being included in the test suite. I think eventually we'll figure out something and un-skip all these, but I'm not sure what we can do here in the meantime so I've been skipping them.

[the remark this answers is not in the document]
hectahertz: Seems like a good idea to add this

hectahertz (approved):

PR #5995, opened 2025-05-02
fix(TextArea): only add class name to outermost element

joshblack:
Update `TextArea` to only add class name to outermost element

#### Changed

- Update `TextArea` to only add class name to outermost element

### Rollout strategy

- [x] Patch release


francinelucca (approved):
⚡

PR #5997, opened 2025-05-04
Remove the CSS modules feature flag from Pagination component 

jamieshark:
- Closes https://github.com/github/primer/issues/4301

#### Removed

CSS modules feature flag from Pagination component

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Pagination/Pagination.tsx
```diff
@@ -158,7 +34,6 @@ export type PageProps = {
 } & Omit<PageDataProps['props'], 'as' | 'role'>
 
 type UsePaginationPagesParameters = {
-  theme?: Record<string, unknown> // set to theme type once /src/theme.js is converted
```
jamieshark, line 161: wasn't totally sure what to do here, it didn't seem like it was being used by our CSS modules, so I removed it.

packages/react/src/Pagination/Pagination.tsx
```diff
@@ -196,53 +68,22 @@ function usePaginationPages({
       }
 
       return (
-        <Page {...props} key={key} theme={theme} className={clsx(enabled && classes.Page)}>
+        <span {...props} key={key} className={clsx(classes.Page)}>
```
jamieshark, line 71: I thought this should have been an `a` tag based on the previous code, but that caused a test to fail with the expected number of `a` tags to render. Not sure if the test had to be updated, or if this is the correct default element.

joshblack (approved):

PR #6007, opened 2025-05-05
refactor(Flash): update sx usage to CSS Modules

joshblack:
Update the `Flash` component from sx to CSS Modules. This change is not behind a feature flag since we're in the process of removing our GA flags.

#### Changed

- Update the `sx` usage of Flash internally to use CSS Modules instead

### Rollout strategy

- [x] Minor release


jonrohan (approved):
Seems good, just needs the CSS lint fixes

PR #5869, opened 2025-04-03
Fix extra padding on SelectPanel with groups

hectahertz:
Closes https://github.com/github/primer/issues/4830

#### Changed

Changes the `ActionList` mode to `full` when groups are present so we don't have that extra padding

### Rollout strategy

- [X] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Testing & Reviewing

(with modern action list FF ONLY)

Before
Screenshot 2025-04-03 at 19 14 41

After
Screenshot 2025-04-03 at 21 20 09

### Merge checklist

- [X] Added/updated tests
- [X] Added/updated documentation
- [X] Added/updated previews (Storybook)
- [X] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [X] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #6009, opened 2025-05-05
Fix extra padding on SelectPanel with groups

francinelucca:
Re-adds https://github.com/primer/react/pull/5869 with some fixes

Closes https://github.com/github/primer/issues/4830

#### Changed

Changes the `ActionList` mode to `full` when groups are present so we don't have that extra padding

### Rollout strategy

- [X] Patch release
- [ ] Minor release

### Testing & Reviewing

(with modern action list FF ONLY)

Before
Screenshot 2025-04-03 at 19 14 41

After
Screenshot 2025-04-03 at 21 20 09

### Merge checklist

- [X] Added/updated tests
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [X] Tested in Safari
- [ ] Tested in Edge


hectahertz (approved):

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -686,6 +686,7 @@ function Panel({
             onInputRefChanged={onInputRefChanged}
             placeholderText={placeholderText}
             {...listProps}
+            variant={listProps.groupMetadata?.length ? 'horizontal-inset' : 'inset'}
```
hectahertz, line 689: Should we maybe just do this in list? and have the logic be internal?

francinelucca: the list would have to override it's own variant prop and that feels a bit weird to me 🤔, there'd be no way for the consumer to override it

PR #6050, opened 2025-05-15
Switch to a CSS based solution for the PageLayout sticky pane

hectahertz:
Closes https://github.com/github/primer/issues/5174

This change removes the `useResizeObserver` and `useEffect` hooks previously used to calculate and set the `--sticky-pane-height` CSS custom property.

**Problem:**
The constant recalculation and application of this style on scroll and resize events were triggering frequent and costly style recalculations (observed at 60ms+ per recalc). This significantly impacted page performance during scrolling near the top of the page. This is not even counting the React component re-render time.

**Solution:**
By replacing this JS code with a CSS based solution we eliminate the continuous style updates, thereby preventing the performance bottleneck caused by excessive style recalculations. The underlying layout or component manages the sticky pane's height without requiring this dynamic JavaScript-driven height adjustment.

This is a small compromise in UX that gets us big performance gains.

More context on the original implementation:
https://share.cleanshot.com/O9uSpc

#### Changed

The only difference in behaviour happens on a very specific scenario:
- The pane doesn't fit in the viewport
- The pane is sticky
- The user hasn't scrolled past the header yet

Existing behavior:
- We can directly scroll to the bottom of the pane (This is what the JS code compensates for)

https://github.com/user-attachments/assets/6ca3c933-4fd5-462b-966b-ab9f079047ef

New behavior:
- To scroll to the bottom of the pane, we have to scroll past the header or attempt to scroll through the pane again, then the scroll interaction bubbles to the parent and we scroll past the header (see video)

https://github.com/user-attachments/assets/3be8aa02-e133-4794-82a2-fe7e8b8cdc03

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


joshblack (approved):
Changes LGTM! Just would want design to 👍 the UX changes. Is there a good way for us to test this in dotcom? Changes layout-wise terrify me lol

hectahertz:
@joshblack Thanks for reviewing! Yes, we'll be review-labing this :)

Issue #6045, opened 2025-05-13
`Skeleton` is invisible inside `Overlay` in dark mode


iansan5653:
### Description

The `SkeletonText` (and other skeleton components) cannot be seen in `Overlay`s because it is invisible against the `--overlay-bgColor` token:

<img width="288" alt="screenshot of apparently empty overlay" src="https://github.com/user-attachments/assets/31e2c0cd-7d99-4708-9861-d2e2e1ffd176" />

In light mode, this is the same as the default background color and looks fine:

<img width="288" alt="screenshot of overlay with three skeleton text lines in it" src="https://github.com/user-attachments/assets/22f02a5a-13ed-468b-a86b-40139f691e1d" />

### Steps to reproduce

```jsx
<Overlay>
  <SkeletonText />
</Overlay>
```

### Version

v37.22.0

### Browser

Chrome

PR #6077, opened 2025-05-20
Fix `Skeleton` background color

langermank:
Closes https://github.com/primer/react/issues/6045

Adds the correct color var


francinelucca (approved):
✨

PR #6083, opened 2025-05-20
chore(_VisuallyHidden): Convert internal component _VisuallyHidden to CSS module styling

jonrohan:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/_VisuallyHidden.tsx
+++ b/packages/react/src/_VisuallyHidden.tsx
@@ -1,29 +1,16 @@
-import styled from 'styled-components'
-import type {SxProp} from './sx'
-import sx from './sx'
+import classes from './_VisuallyHidden.module.css'
+import {clsx} from 'clsx'
 
 interface Props {
   isVisible?: boolean
 }
 
-const VisuallyHidden = styled.span<Props & SxProp>`
-  ${({isVisible = false}) => {
-    if (isVisible) {
-      return sx
-    }
-
-    return `
-        position: absolute;
-        width: 1px;
-        height: 1px;
-        padding: 0;
-        margin: -1px;
-        overflow: hidden;
-        clip: rect(0, 0, 0, 0);
-        white-space: nowrap;
-        border-width: 0;
-    `
-  }}
-`
+function VisuallyHidden({isVisible, children, ...rest}: Props & React.HTMLAttributes<HTMLSpanElement>) {
+  return (
+    <span className={clsx({[classes.InternalVisuallyHidden]: !isVisible})} {...rest}>
+      {children}
+    </span>
+  )
+}
 
 export default VisuallyHidden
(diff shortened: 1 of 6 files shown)


packages/react/src/_VisuallyHidden.tsx
```diff
@@ -1,29 +1,25 @@
...
-    `
-  }}
-`
+function VisuallyHidden({
+  isVisible,
+  children,
+  as = 'span',
+  className,
+  ...rest
+}: Props & React.HTMLAttributes<HTMLSpanElement> & SxProp) {
```
joshblack, line 17: If `as` is offered then the given `React.HTMLAttributes` type here could be incorrect if they pass in something other than a `span`. I think this would need to be something like:

```tsx
```suggestion
function VisuallyHidden<As extends React.ElementType>({
  isVisible,
  children,
  as = 'span',
  className,
  ...rest
}: Props & { as?: As } & React.ComponentPropsWithoutRef<ElementType extends As ? As : 'span'> & SxProp) {
```
```

**Review by joshblack — CHANGES_REQUESTED** [review]

File: packages/react/src/_VisuallyHidden.tsx [hunk 2]
```diff
@@ -1,29 +1,25 @@
...
+import classes from './_VisuallyHidden.module.css'
+import {clsx} from 'clsx'
+import {BoxWithFallback} from './internal/components/BoxWithFallback'
 import type {SxProp} from './sx'
-import sx from './sx'
 
 interface Props {
   isVisible?: boolean
+  as?: React.ElementType
 }
```
joshblack, line 9: I think if we want to have `as?` defined on `Props` then we would need get rid of the intersection below with `{ as?: As }` and make Props generic like:

```suggestion
interface Props<As extends React.ElementType> {
  isVisible?: boolean
  as?: As
}
```

joshblack (changes requested):

packages/react/src/_VisuallyHidden.tsx
```diff
@@ -1,29 +1,19 @@
...
+import classes from './_VisuallyHidden.module.css'
+import {clsx} from 'clsx'
+import {BoxWithFallback} from './internal/components/BoxWithFallback'
 import type {SxProp} from './sx'
-import sx from './sx'
 
-interface Props {
+interface Props extends React.ComponentPropsWithoutRef<'span'> {
   isVisible?: boolean
+  as?: React.ElementType
```
jonrohan, line 8: This component is only used internally, and I verified the only instances pass `as="h2"`

joshblack (approved):

PR #5829, opened 2025-03-27
feat(UnderlineNav): Support `inset` and `flush` variants

maraisr:
__*Variant: flush*__

image

__*Variant: inset (default)*__

image

#### New

Added `inset` and `full` as variant props to `UnderlineNav` to aid in removing horizontal spacing.

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

Since the current behaviour is with inset spacing, the default for this new prop is also `inset`. So fully backwards compatible, a testing strategy is that all existing vrt tests pass.

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [x] Added/updated previews (Storybook)
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [x] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #6086, opened 2025-05-20
Add variant no-horizontal-padding to UnderlineNav

hectahertz:
Closes https://github.com/github/primer/issues/4607

Continued from https://github.com/primer/react/pull/5829v

This pull request introduces a new `variant` prop to the `UnderlineNav` component, allowing removal of the horizontal padding for tabs.

https://github.com/user-attachments/assets/810fb91e-fa0e-41d6-b14b-ae257cf93a60

##### Component Enhancements:
* Added a new `variant` prop to `UnderlineNav` with options `'default'` and `'no-horizontal-padding'` to control horizontal padding.

#####  Storybook Updates:
* Added the `variant` prop to the Storybook controls for `UnderlineNav`, allowing users to toggle between the two variants in the Storybook UI.
* Added an extra story to demonstrate the new prop value.

### Rollout strategy

- [ ] Patch release
- [X] Minor release

### Merge checklist

- [ ] Added/updated tests
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [ ] Tested in Edge


joshblack (commented):
One question I had was if we want the no padding case to be the default in the future or no 🤔 Put another way, is the default case that people should be removing this padding or not.

This came up when using components like Stack because it doesn't seem like most of the time you wouldn't want that padding in there. At a higher level, it's almost like we need a bleed concept aka https://github.com/github/primer/issues/786 for these types of moments

hectahertz:
> One question I had was if we want the no padding case to be the default in the future or no 🤔 Put another way, is the default case that people should be removing this padding or not.
> 
> This came up when using components like Stack because it doesn't seem like most of the time you wouldn't want that padding in there. At a higher level, it's almost like we need a bleed concept aka [github/primer#786](https://github.com/github/primer/issues/786) for these types of moments

100%; I'm gonna change the variant 'default' value to indicate that it has padding, so we avoid future refactors no matter what we end up going for.

hectahertz:
Addressed in https://github.com/primer/react/pull/5829

Issue #5446, opened 2024-12-16
`Stack` does not properly forward ref


iansan5653:
`Stack` appears to accept a `ref` prop based on the TypeScript types, but this ref is not actually bound to the underlying element:

```tss
export function Example() {
  const ref = useRef<HTMLDivElement>(null)

  useEffect(() => console.log(ref.current))

  return <Stack {...props} ref={ref}>contents</Stack>
}
```

If `Stack` correctly forwarded its ref, the console would log an `HTMLDivElement` instance. However, it instead always logs `null`.

Jaykhanderay:
import React, { forwardRef } from "react";

const Stack = forwardRef((props, ref) => {
  return (
    <div {...props} ref={ref}>
      {props.children}
    </div>
  );
});

export default Stack;

Rakesh31-syc:
In React, when a ref is passed to a component, that component must forward the ref to a DOM element using React.forwardRef. If the Stack component does not implement forwardRef, the ref will not be attached to the underlying DOM element, resulting in ref.current being null or undefined.

import React, { useRef, useEffect } from 'react';

export function Example(props) {
  const ref = useRef<HTMLDivElement>(null);

  useEffect(() => {
    console.log(ref.current);
  }, []);

  return <Stack {...props} ref={ref}>contents</Stack>;
}

lesliecdubs:
Primer First Responder, we'd love your help fixing this `ref`!

joshblack:
@iansan5653 I believe the types might be too loose for Stack which might be part of the problem here 🤔 The component does not currently support `ref` but the types seem to be allowing basically anything to be passed in so it has the appearance of being supported.

We've been trying to find a good way to do polymorphic `as` types and this seems to be a limitation with the approach we have 😅 Let me see if we can address that separately.

For `Stack` specifically, would it still be desirable for it to accept a `ref`? We can treat this as an enhancement then and get it added.

nita939:
Q gak retiii

junautogroup:
> `Stack` appears to accept a `ref` prop based on the TypeScript types, but this ref is not actually bound to the underlying element:
> 
> ```
> export function Example() {
>   const ref = useRef<HTMLDivElement>(null)
> 
>   useEffect(() => console.log(ref.current))
> 
>   return <Stack {...props} ref={ref}>contents</Stack>
> }
> ```
> 
[...]

PR #6111, opened 2025-05-22
fix(Stack): correctly forward a Ref

francinelucca:
Closes https://github.com/primer/react/issues/5446

#### New

- Wrap `Stack` and `StackItem` in forwardRef call

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Stack/Stack.tsx
+++ b/packages/react/src/Stack/Stack.tsx
@@ -1,5 +1,5 @@
 import type React from 'react'
-import {type ElementType} from 'react'
+import {forwardRef, type ElementType} from 'react'
 import type {ResponsiveValue} from '../hooks/useResponsiveValue'
 import {getResponsiveAttributes} from '../internal/utils/getResponsiveAttributes'
 import classes from './Stack.module.css'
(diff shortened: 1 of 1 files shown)


packages/react/src/Stack/Stack.tsx
```diff
@@ -68,34 +68,40 @@ type StackProps<As> = React.PropsWithChildren<{
...
+      align = 'stretch',
+      direction = 'vertical',
+      gap,
+      justify = 'start',
+      padding = 'none',
+      wrap = 'nowrap',
+      className,
+      ...rest
+    }: StackProps<As> & React.ComponentPropsWithRef<ElementType extends As ? As : 'div'>,
+    forwardedRef: React.Ref<HTMLDivElement> | undefined,
```
joshblack, line 85: I think the type for this will get tricky because the `forwardRef` type should match whatever `As` is and I believe with the setup now it would always be `HTMLDivElement`.

The way this is typed also seems to be making the props inferred as `any` so they won't show up in autocomplete / won't fail if passed the wrong value (let me know if you're not seeing this locally though, could be something on my end)

```tsx
 <Stack
  as="span"
  ref={node => {
    // `node` is `HTMLDivElement` instead of `HTMLSpanElement`
  }}
  // I don't think this will fail TS now the way the types are inferred
  gap="nonsense"
/>
```

joshblack (changes requested):

packages/react/src/Stack/Stack.tsx
francinelucca: Good catch! I didn't notice 😅. I **think** I fixed it now, lmk what you think!

packages/react/src/Stack/Stack.tsx
```diff
@@ -112,24 +119,30 @@ type StackItemProps<As> = React.PropsWithChildren<{
...
-      {...rest}
-      className={clsx(className, classes.StackItem)}
-      {...getResponsiveAttributes('grow', grow)}
-    >
-      {children}
-    </StackItemBaseComponent>
-  )
-}
+const StackItem = forwardRef(
+  <As extends ElementType>(
```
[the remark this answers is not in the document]
francinelucca: ah yeah my bad, fixed!

joshblack (approved):

PR #6126, opened 2025-05-28
chore(deps): update eslint and related deps to v9

joshblack:
Closes https://github.com/github/primer/issues/4155

This is take 2 for updating the project to use ESLint v9. The previous attempt ended up having a larger than expected diff so I updated some plugins and then opened this up again so that it was easier to merge.

#### Changed

- Update eslint to v9
- Migrate our eslint config to the new flat config
- Address new eslint errors, mostly from removed/replaced rules


francinelucca:
@copilot can you fix the broken CI here?

Issue #6168, opened 2025-06-05
ProgressBar.Item does not pass through style prop


hussam-i-am:
Noticed that the `style` prop gets overridden for the component `ProgressBar.Item` and is replaced by custom style within the component.

PR #6169, opened 2025-06-05
fix(ProgressBar): pass through style prop

hussam-i-am:
- Closes https://github.com/primer/react/issues/6168

#### Changed

Appends the passed in style props to the existing styles object

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/@types/react/index.d.ts
+++ b/packages/react/@types/react/index.d.ts
@@ -1,0 +1,8 @@
+// eslint-disable-next-line no-restricted-imports
+export * from 'react'
+
+declare module 'react' {
+  interface CSSProperties {
+    [key: `--${string}`]: string | number | undefined
+  }
+}
--- a/.changeset/small-baboons-press.md
+++ b/.changeset/small-baboons-press.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+fix(ProgressBar): passthrough style props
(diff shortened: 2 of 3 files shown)


packages/react/@types/react/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line no-restricted-imports
+export * from 'react'
+
+declare module 'react' {
+  interface CSSProperties {
```
francinelucca: can you elaborate as to why this is needed? 🙏

hussam-i-am: Oh, forgot to mention it in the description, this updates the definition of CSSProperties to allow for CSS variables, so that we can just used `React.CSS properties` as the type in https://github.com/primer/react/pull/6169/files#diff-c755511099c2739a4276dc3b714e3ad5083385129f4fb3a461115179c50cbcbcR52, instead of `{[key: string]: string}` which is a bit too generic.

hussam-i-am: Actually looks like its causing the build to fail, so reverting that part of the change

PR #6170, opened 2025-06-05
Llastflowers/5118/select panel unhide footer

llastflowers:
Closes https://github.com/github/primer/issues/5118

#### New

- Detect mobile keyboard visibility by detecting viewport size changes when keyboard appears
- Calculate keyboard height with original viewport height - reduced viewport height when the keyboard is visible
- Adjust panel height based on keyboard height to keep footer above keyboard and always visible

(This approach made more sense to me vs. trying to reposition the footer itself to stick to the keyboard, because it seemed a lot simpler to implement, generally less issue-prone, easier to understand the code, required fewer changes, seems more likely to work across different browsers and devices, and the UI interactions are basically exactly the same, if not a bit cleaner. That said, I'm open to feedback ofc!)

#### Changed

Panel now automatically resizes when mobile keyboard is visible so that content including the footer is not hidden behind the keyboard.

Safari on iOS:

Safari screen before interaction&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Safari screen when keyboard appears&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Safari screen after keyboard is dismissed

Chrome on iOS:

Chrome screen before interaction&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Chrome screen when keyboard appears&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Chome screen after keyboard is dismissed

#### Removed

Removed one line from the corresponding CSS module that I noticed was making the footer button slightly off-center. If this was intentional lmk and I can revert

Button spacing before&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Button spacing after

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- Test in Storybook on mobile with the following feature flags enabled (or maybe just try [this link](https://primer-2f59907f39-13348165.drafts.github.io/storybook/?path=/story/components-selectpanel--default&globals=featureFlags.primer_react_css_modules_ga:!true;featureFlags.primer_react_select_panel_with_modern_action_list:!true;featureFlags.primer_react_select_panel_fullscreen_on_narrow:!true;featureFlags.primer_react_select_panel_order_selected_at_top:!true)?), and tap "Open canvas in new tab" so that Storybook UI doesn't interfere with the functionality

Feature flags screenshot from Storybook
- `@tylerjdev` offered to test on his Android mobile phone for good measure (ty!)
- Please let me know if the repositioning movement of the footer when the keyboard opens appears too abrupt/messy in testing. I can probably add a delay or transition to make it look a bit smoother if needed 😅 

### References
https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API
https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


packages/react/src/SelectPanel/SelectPanel.module.css
```diff
@@ -140,7 +140,6 @@
   display: none;
   align-items: center;
   padding: var(--base-size-8);
-  gap: var(--stack-gap-condensed);
```
francinelucca, line 143: @hectahertz tagging you since I know you worked on a lot of these, any idea why/if this is needed or are we safe to remove? 👀

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -141,7 +141,56 @@ const doesItemsIncludeItem = (items: ItemInput[], item: ItemInput) => {
...
+  })
+
+  const initialHeightRef = useRef(0)
+
+  useEffect(() => {
+    if (window.visualViewport) {
+      initialHeightRef.current = window.visualViewport.height
+    }
+
+    const handleViewportChange = () => {
```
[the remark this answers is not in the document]
francinelucca: I think this is a good idea!

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -141,7 +141,56 @@ const doesItemsIncludeItem = (items: ItemInput[], item: ItemInput) => {
...
```
francinelucca: @joshblack can we get an SSR-safe gut check here? not sure if this falls in the same categories as useMediaQuery, useResponsiveValues situation....

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -141,7 +141,56 @@ const doesItemsIncludeItem = (items: ItemInput[], item: ItemInput) => {
...
+        setKeyboardState({
+          height: keyboardHeight,
+          isVisible: keyboardHeight > 10,
+        })
+      }
+    }
+    if (window.visualViewport) {
+      // Using visualViewport to more reliably detect viewport changes across different browsers, which specifically requires these listeners
+      // eslint-disable-next-line github/prefer-observers
+      window.visualViewport.addEventListener('resize', handleViewportChange)
```
francinelucca: any reason we can't use `useResizeObserver` here? 👀

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -153,7 +202,7 @@ const defaultRendorAnchor: NonNullable<SelectPanelProps['renderAnchor']> = props
 function Panel({
   open,
   onOpenChange,
-  renderAnchor = defaultRendorAnchor,
+  renderAnchor = defaultRenderAnchor,
```
francinelucca, line 156: ✨

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -647,6 +710,13 @@ function Panel({
             '--max-height': overlayProps?.maxHeight ? heightMap[overlayProps.maxHeight] : heightMap['large'],
             /* override AnchoredOverlay position */
             transform: variant === 'modal' ? 'translate(-50%, -50%)' : undefined,
+            // set maxHeight based on calculated availablePanelHeight when keyboard is visible
+            ...(keyboardState.isVisible
+              ? {
+                  maxHeight:
+                    typeof availablePanelHeight === 'number' ? `${availablePanelHeight}px` : availablePanelHeight,
+                }
+              : {}),
```
francinelucca:
```suggestion
            // set maxHeight based on calculated availablePanelHeight when keyboard is visible
             maxHeight:  availablePanelHeight ? `${availablePanelHeight}px` : 'auto'
```


if you set the `availablePanelHeight` type to `number | undefined` , can also eliminate else statement in line 439 in favor of `setAvailablePanelHeight(window.visualViewport?.height)`

francinelucca (commented):
some preliminary comments ✨

.changeset/dull-pots-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Llastflowers/5118/select panel unhide footer
```
francinelucca: probably change this text here since this is what shows up in the changelog 👀

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -141,7 +141,56 @@ const doesItemsIncludeItem = (items: ItemInput[], item: ItemInput) => {
...
-const defaultRendorAnchor: NonNullable<SelectPanelProps['renderAnchor']> = props => {
+interface KeyboardState {
+  height: number
+  isVisible: boolean
+}
+
+// calculate keyboard visibility state by detecting starting viewport height and subtracting viewport height when keyboard is visible
+function useKeyboardState(): KeyboardState {
+  const [keyboardState, setKeyboardState] = useState<KeyboardState>({
+    height: 0,
```
francinelucca: do we need height in the state? it doesn't look like we're using it anywhere on the consumer side 🤔 
Wondering if this hook can be removed in favor of a scroll/resize listener on the component that checks for `window.visualViewport.height > 10` to update the `availableHeight`

packages/react/src/SelectPanel/SelectPanel.tsx
llastflowers: My understanding is that useResizeObserver looks for element size changes, which I had some concerns about not working reliably across different browsers and devices. Apparently some browsers/devices don't resize the viewport when the mobile keyboard pops up but instead just overlay the keyboard, making it difficult to detect and measure the keyboard. 

Visual Viewport specifically focuses on the visible viewport excluding things like mobile keyboards, and seemed more likely to work consistently. Because testing this mobile keyboard behavior on all kinds of different devices and browsers to confirm it always works seems pretty difficult, I focused on the approach that seemed the most reliable according to what I found in research. 

That said, I don't know a ton about working with mobile stuff (especially with native keyboards, etc.), so if the team prefers a different approach that may be better, and has high confidence in its reliability across browsers and devices or a good method to test and confirm, I'm definitely open to changing this!

francinelucca: Makes sense, thanks for explaining! 🙏

packages/react/src/SelectPanel/SelectPanel.tsx
joshblack: I bet it's good since the select panel opens up only after an interaction, right?  So should be stable during SSR and only show up once client JS hydrates 👀

francinelucca: yep, sounds good. Thanks!

packages/react/src/SelectPanel/SelectPanel.tsx
llastflowers: Thanks for the suggestion! I do think this is cleaner. I have implemented these changes with the addition of an explicit `availablePanelHeight !== undefined` check instead of just the truthy check because it more precisely aligns with the number | undefined type definition and makes the code's intention clearer.

packages/react/src/SelectPanel/SelectPanel.tsx
llastflowers: Thanks for the suggestions; I agree! Some of this is actually left over from how I was testing this stuff on mobile, but it doesn't need to be that way anymore 😅

llastflowers:
Just wanted to document this in case anyone else notices in testing: Tyler and I noticed an issue in mobile testing where if you initiate the scroll swipe with your finger ON the footer while the keyboard is up, you're able to scroll beneath the footer, where there shouldn't be any content. I did some testing by turning the storybook preview wrapper background color to blue, and am pretty sure this confirms that that weird space you can see in the above condition is just part of storybook's UI, and would not be an issue in an implementation of SelectPanel.

blue background no keyboard&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;blue background with keyboard&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;blue background shown below keyboard

packages/react/src/SelectPanel/SelectPanel.tsx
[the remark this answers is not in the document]
llastflowers: accepted

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -378,6 +382,41 @@ function Panel({
...
+      if (window.visualViewport) {
+        window.visualViewport.removeEventListener('resize', handleViewportChange)
+        window.visualViewport.removeEventListener('scroll', handleViewportChange)
+      }
+    }
+  }, [])
+
+  // Update available panel height when keyboard is visible
+  useEffect(() => {
+    setAvailablePanelHeight(isKeyboardVisible ? window.visualViewport?.height : undefined)
```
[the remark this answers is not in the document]
llastflowers: accepted

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -378,6 +382,37 @@ function Panel({
     }
   }, [open, dataLoadedOnce, onFilterChange, filterValue, items, loadingManagedExternally, listContainerElement])
 
+  useEffect(() => {
+    if (window.visualViewport) {
+      initialHeightRef.current = window.visualViewport.height
+    }
```
TylerJDev: I'm wondering if we should also check if the SelectPanel is open or not? This way, `handleViewportChange` will only run when it is open, and will clean up when closed. Could be something like this:

```suggestion
    if (!window.visualViewport || !open) {
      return
    }
    
    initialHeightRef.current = window.visualViewport.height
```

We'd need to add `open` as a dependency though. Curious what you think 🤔

francinelucca (commented):
This is looking pretty good. Noticed some weird behavior on zoom. Wondering if there's anything we can do about that 👀 . Also I like @TylerJDev suggestion about only doing recalculations when the panel is open.

https://github.com/user-attachments/assets/55286872-9ce9-422a-92d2-b2912a53ce35

packages/react/src/SelectPanel/SelectPanel.tsx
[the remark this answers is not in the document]
llastflowers: accepted

packages/react/src/SelectPanel/SelectPanel.tsx
llastflowers: Great suggestion, thank you! I'll make this change 👍

llastflowers:
> This is looking pretty good. Noticed some weird behavior on zoom. Wondering if there's anything we can do about that 👀 . Also I like @TylerJDev suggestion about only doing recalculations when the panel is open.
> 
>  ScreenRecording_06-09-2025.11-40-42_1.mov

Thanks for calling this out! I'm adding an additional check for viewport scale changes, so this resizing/repositioning stuff won't happen on scale change (like when a user zooms), only height change.

francinelucca (approved):
🧑‍🍳💋

packages/react/src/SelectPanel/SelectPanel.module.css
hectahertz: I think it's a leftover from a previous implementation; if the snapshots look fine I can't think of any issues with removing it!

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -378,6 +383,46 @@ function Panel({
     }
   }, [open, dataLoadedOnce, onFilterChange, filterValue, items, loadingManagedExternally, listContainerElement])
 
+  useEffect(() => {
+    if (!window.visualViewport || !open) {
```
TylerJDev: Nit: I'm wondering if it would also be good to check if `isNarrowScreenSize` is true in addition to `open`? This would reduce calls within the function unless the panel is fullscreen 🤔 

If we make the narrow screen opt-in, I'm wondering if we'll need to add a conditional to the `useEffect` too?

cc: @francinelucca

TylerJDev (approved):
Looks good! Nice work ✨ 

Left one comment, but non-blocking!

packages/react/src/SelectPanel/SelectPanel.tsx
llastflowers: I'm not sure how this works in practice, but could this potentially cause issues for iPad users?

TylerJDev: I think for this case specifically, `isNarrowScreenSize` would only be `true` when we'd need to calculate the size for the keyboard. Though let me know if I'm wrong 😅

llastflowers: Got it; since the footer is only visible on narrow screen, I think it does make sense to do this!

francinelucca: Yes! great idea. in case its not already in the SelectPanel code, the check for isNarrowFullScreen would look something like this:

```
const isNarrow = useResponsiveValue({narrow: true}, false)
const isNarrowFullScreen = isNarrow && usingFullScreenOnNarrow
```

PR #6178, opened 2025-06-08
remove deprecated react-test-renderer dependency

nazarifard:


francinelucca:
Thanks for proposing this change. We have reviewed and decided not to move forward with the library change, closing

PR #6179, opened 2025-06-08
remove @lit-labs/react dependency

nazarifard:
createComponent re-developed without using of @lit-lab/react.
an important step to reduce unnecessary dependencies. It can be more useful if we want to migrate from "React" to "Preact"


joshblack:
Hi there! 👋 Thanks for taking the time to make this PR 🙏 

We talked about it real quick as a team and I think we're okay with keeping on this dependency (and moving to `@lit/react` in the future when our package is pure ESM). Hope that makes sense!

I'm going to close this PR because of that, let me know if you have any questions!

PR #6182, opened 2025-06-09
Textarea: Add `minHeight` and `maxHeight` as props

TylerJDev:
Closes https://github.com/github/primer/issues/5296

Adds new props `minHeight` and `maxHeight` to the `Textarea`. This allows you to customize the min/max height of the textarea directly.

#### New

* Added `minHeight` and `maxHeight` props to `Textarea`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Textarea/Textarea.tsx
+++ b/packages/react/src/Textarea/Textarea.tsx
@@ -34,6 +34,14 @@ export type TextareaProps = {
    * The className to apply to the wrapper element
    */
   className?: string
+  /**
+   * The minimum height of the Textarea
+   */
+  minHeight?: number
+  /**
+   * The maximum height of the Textarea
+   */
+  maxHeight?: number
 } & TextareaHTMLAttributes<HTMLTextAreaElement> &
   SxProp
 
@@ -55,6 +63,8 @@ const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
       block,
       contrast,
       className,
+      minHeight,
+      maxHeight,
       ...rest
     }: TextareaProps,
     ref,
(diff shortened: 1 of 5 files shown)


packages/react/src/Textarea/Textarea.tsx
```diff
@@ -34,6 +34,14 @@ export type TextareaProps = {
    * The className to apply to the wrapper element
    */
   className?: string
+  /**
+   * The minimum height of the Textarea
+   */
+  minHeight?: number
```
francinelucca, line 40: should we use small/medium/large/xlarge instead? 🤔 CC: @langermank

TylerJDev:
I noticed that we directly expose `style` on the `textarea`, as the `...rest` allows for pretty much anything (`style` included). I'm wondering if this change is still worth it? 

e.g. we could do `<Textarea style={{minHeight: 100}} />` to get the same effect. Curious what y'all think 

(cc: @langermank, @francinelucca)

francinelucca:
if we're not going to "Standardize" the sizes then I feel better with opting for just the existing style prop for these type of changes, i.e. no changes needed. But will defer to @langermank for final call

packages/react/src/Textarea/Textarea.tsx
langermank: Nah, I feel like with a field its just going to be sooo specific to the use case

langermank:
I feel like it makes sense here as a prop since its a common attribute for text fields. Might as well make it easier to adjust at the prop level.

PR #6190, opened 2025-06-10
fix Banner spacing

llastflowers:
Closes https://github.com/github/primer/issues/3720

#### Changed

Styling update addressing banner spacing issues--see vertical alignment on single line banner below (before / after):

Screenshot 2025-06-10 at 10 41 12
Screenshot 2025-06-10 at 10 41 30

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Banner/Banner.module.css
+++ b/packages/react/src/Banner/Banner.module.css
@@ -48,7 +48,7 @@
 
 .BannerContainer {
   font-size: var(--text-body-size-medium);
-  align-items: start;
+  align-items: center;
   line-height: var(--text-body-lineHeight-medium);
   row-gap: var(--base-size-4);
   column-gap: var(--base-size-4);
--- a/packages/react/src/Banner/Banner.stories.tsx
+++ b/packages/react/src/Banner/Banner.stories.tsx
@@ -63,8 +63,7 @@ export const Playground: StoryObj<typeof Banner> = {
   },
   args: {
     title: 'Banner title',
-    description:
-      'GitHub users are now required to en able two-factor authentication as an additional security measure.',
+    description: 'GitHub users are now required to enable two-factor authentication as an additional security measure.',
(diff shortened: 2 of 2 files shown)


lukasoppermann:
The primary button is perfect.

However, now (not sure if this was already like this before) the secondary button alone has bad spacing.

CleanShot 2025-06-11 at 13 04 56@2x


This would be a good spacing token, however, it can only be used of the primary button is not set: `--control-large-paddingInline-normal`

llastflowers:
> The primary button is perfect.
> 
> However, now (not sure if this was already like this before) the secondary button alone has bad spacing.
> 
> secondary action spacing--before
> 
> This would be a good spacing token, however, it can only be used of the primary button is not set: `--control-large-paddingInline-normal`

Thanks for surfacing this! It looks like this issue was already present before, but I think it's fine to lump in these changes with this PR. Here's what it looks like now:

updated secondary action spacing

llastflowers:
Documenting some context for the newer code changes based on slack DMs with `@lukasoppermann`:

The change that addressed this original issue (before/after):

banner with buttons without dismiss zoomed in before

banner with buttons without dismiss zoomed in after

was causing this view to look wonky because the `x` was not vertically aligned, and the buttons now are:

banner with button with dismiss zoomed in before

So I have conditionally applied a matching style to the dismiss `x` to align it with buttons when present:

banner with button with dismiss zoomed in after

but the `x` will not move if there are no buttons, to prevent it from moving from its current top-right default position:

banner with position of dismiss x

Margins adjusted with buttons vs. default:

moved x showing margins
unmoved x showing margins

jonrohan (approved):

PR #5345, opened 2024-12-03
docs: fix *.docs.json stories

francinelucca:
Closes https://github.com/github/primer/issues/4433

Some story IDs are listed incorrectly in their respective *.docs.json, this PR fixes them so that they can be pulled into primer.style storybook embed

#### Changed

- Update story ids in some components' *.docs.json files

### Rollout strategy

- [x] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Testing & Reviewing

Test deployment: https://primer-e313264adc-26441320.drafts.github.io/components/data-table/react/alpha

### Merge checklist

- [ ] Added/updated tests
- [x] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #6210, opened 2025-06-16
[SelectPanel] Add notice announcements to SelectPanel

kelsey-myers:
Closes [#11748](https://github.com/github/accessibility-audits/issues/11748) and [#5345](https://github.com/github/primer/issues/5345)

### Changelog

image

image

image

#### New

- Added an announcement functionality for SelectPanel notices

### Rollout strategy

- [X] Patch release
- [ ] Minor release

### Merge checklist

- [X] Added/updated tests
- [ ] Added/updated documentation (N/A)
- [X] Added/updated previews (Storybook) (auto)
- [X] Tested in Chrome
- [X] Tested in Firefox
- [X] Tested in Safari
- [ ] Tested in Edge
- [X] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

### DIFF
--- a/packages/react/src/SelectPanel/SelectPanel.tsx
+++ b/packages/react/src/SelectPanel/SelectPanel.tsx
@@ -757,7 +757,7 @@ function Panel({
             ) : null}
           </div>
           {notice && (
-            <div ref={noticeRef} aria-live="polite" data-variant={notice.variant} className={classes.Notice}>
+            <div ref={noticeRef} data-variant={notice.variant} className={classes.Notice}>
               {iconForNoticeVariant[notice.variant]}
               <div>{notice.text}</div>
             </div>


kelsey-myers:
> 🔴 [golden-jobs](https://github.com/github/github/actions/runs/15691643375) completed with status `failure`.

Failing tests are unrelated 🐸

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -425,6 +426,23 @@ function Panel({
     }
   }, [open, isNarrowScreenSize])
 
+  useEffect(() => {
+    const announceNotice = async () => {
+      if (!noticeRef.current) return
+      const liveRegion = document.querySelector('live-region')
```
TylerJDev, line 432: Would it be possible to utilize the [`announceText` function](https://github.com/primer/react/blob/main/packages/react/src/SelectPanel/SelectPanel.tsx#L41) here instead?

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -739,7 +757,7 @@ function Panel({
             ) : null}
           </div>
           {notice && (
-            <div aria-live="polite" data-variant={notice.variant} className={classes.Notice}>
+            <div ref={noticeRef} aria-live="polite" data-variant={notice.variant} className={classes.Notice}>
```
TylerJDev: Since we're utilizing `announce`, we can most likely remove the `aria-live` here, as it is redundant.

packages/react/src/SelectPanel/SelectPanel.tsx
kelsey-myers: @TylerJDev I did it this way because`notice.text` can either be a string or a react element, so I figured `announceFromElement` covers both cases? (unless you know of a way we can extract the text from the element)

packages/react/src/SelectPanel/SelectPanel.tsx
kelsey-myers: Good point, updated :)

packages/react/src/SelectPanel/SelectPanel.tsx
TylerJDev: Makes sense! We can leave it as-is then 😁

TylerJDev (approved):
LGTM! ✨

PR #6211, opened 2025-06-16
feat: add support for react compiler

joshblack:
Context:
https://github.com/github/primer/issues/4742
https://github.com/github/primer/issues/4798

Add support for React Compiler by following the guide in: https://react.dev/learn/react-compiler. We use the approach where we are opting out files from being processed by React Compiler (as recommended by the guide). Files that we are not processing with React Compiler are captured in our migration-status workflow.

#### New

- Add support for React Compiler through babel

#### Changed

- Update config for vite in Storybook and vitest to include React Compiler settings

### Rollout strategy

- [x] Minor release


hectahertz (approved):
Love this approach, let's do it! 🚀

PR #6223, opened 2025-06-19
[ActionList] Prevent title="[object Object]" when Description has non-string children

liuliu-dev:
Closes #4400

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/ActionList/ActionList.test.tsx
+++ b/packages/react/src/ActionList/ActionList.test.tsx
@@ -174,4 +174,39 @@ describe('ActionList', () => {
     expect(container.querySelector('li[aria-disabled="true"]')?.nextElementSibling).toHaveTextContent('Option 4')
     expect(container.querySelector('li[aria-disabled="true"]')?.nextElementSibling).toHaveAttribute('tabindex', '0')
   })
+
+  it('sets title correctly for Description component', () => {
+    const {container} = HTMLRender(
+      <ActionList>
+        <ActionList.Item>
+          Option 1<ActionList.Description truncate>Simple string description</ActionList.Description>
+        </ActionList.Item>
+        <ActionList.Item>
+          Option 2
+          <ActionList.Description truncate>
+            <span>Complex</span> content
+          </ActionList.Description>
+        </ActionList.Item>
+        <ActionList.Item>
+          Option 3
+          <ActionList.Description truncate title="Custom title">
+            <span>Complex</span> content
+          </ActionList.Description>
+        </ActionList.Item>
+        <ActionList.Item>
+          Option 4
+          <ActionList.Description>
+            <span>Non-truncated</span> content
+          </ActionList.Description>
+        </ActionList.Item>
+      </ActionList>,
+    )
+
+    const descriptions = container.querySelectorAll('[data-component="ActionList.Description"]')
+
+    expect(descriptions[0]).toHaveAttribute('title', 'Simple string description')
+    expect(descriptions[1]).not.toHaveAttribute('title')
+    expect(descriptions[2]).toHaveAttribute('title', 'Custom title')
+    expect(descriptions[3]).not.toHaveAttribute('title')
+  })
 })
--- a/packages/react/src/ActionList/Description.tsx
+++ b/packages/react/src/ActionList/Description.tsx
@@ -20,6 +20,11 @@ export type ActionListDescriptionProps = {
    * Whether the inline description should truncate the text on overflow.
    */
   truncate?: boolean
+  /**
+   * The title attribute for the truncated text tooltip.
+   * If not provided and children is a string, it will be set automatically.
+   */
+  title?: string
 } & SxProp
 
 export const Description: React.FC<React.PropsWithChildren<ActionListDescriptionProps>> = ({
@@ -27,9 +32,11 @@ export const Description: React.FC<React.PropsWithChildren<ActionListDescription
   sx = defaultSxProp,
   className,
   truncate,
+  title,
   ...props
 }) => {
   const {blockDescriptionId, inlineDescriptionId} = React.useContext(ItemContext)
+  const effectiveTitle = title || (typeof props.children === 'string' ? props.children : undefined)
 
   if (variant === 'block' || !truncate) {
     return (
@@ -49,7 +56,7 @@ export const Description: React.FC<React.PropsWithChildren<ActionListDescription
         id={inlineDescriptionId}
         className={clsx(className, classes.Description)}
         sx={sx}
-        title={props.children as string}
+        title={effectiveTitle}
         inline={true}
         maxWidth="100%"
         data-component="ActionList.Description"
--- a/.changeset/brown-rules-taste.md
+++ b/.changeset/brown-rules-taste.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Fix ActionList.Description title attribute for non-string children with truncate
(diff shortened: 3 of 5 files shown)


packages/react/src/ActionList/Description.tsx
```diff
@@ -20,16 +20,23 @@ export type ActionListDescriptionProps = {
    * Whether the inline description should truncate the text on overflow.
    */
   truncate?: boolean
+  /**
+   * The title attribute for the truncated text tooltip.
+   * If not provided and children is a string, it will be set automatically.
+   */
```
TylerJDev: Can we add an additional note here for accessibility? The `title` attribute is a tricky one since it's not really accessible, but since we're utilizing it anyways, we can try suggest only utilizing it when it's truly needed 🤔 


```suggestion
  /**
   * The title attribute for the truncated text tooltip.
   * If not provided and children is a string, it will be set automatically.
   * 
   * `title` should be used sparingly, as it may be inaccessible to some users.
   */
```

packages/react/src/ActionList/Description.tsx
```diff
@@ -20,16 +20,23 @@ export type ActionListDescriptionProps = {
...
 export const Description: React.FC<React.PropsWithChildren<ActionListDescriptionProps>> = ({
   variant = 'inline',
   sx = defaultSxProp,
   className,
   truncate,
+  title,
   ...props
 }) => {
   const {blockDescriptionId, inlineDescriptionId} = React.useContext(ItemContext)
+  const effectiveTitle = title || (typeof props.children === 'string' ? props.children : '')
```
TylerJDev: Nit: Should we add additional types so that `title` can only be used when `variant !== 'block'` or `truncate === true`? Something like this: 

```tsx
type VariantTypes =
  | {
      variant: 'block'
      title: never
    }
  | {
      /**
       * Secondary text style variations.
       *
       * - `"inline"` - Secondary text is positioned beside primary text.
       * - `"block"` - Secondary text is positioned below primary text.
       */
      variant?: 'inline'
      /**
       * The title attribute for the truncated text tooltip.
       * If not provided and children is a string, it will be set automatically.
       *
       * `title` should be used sparingly, as it may be inaccessible to some users.
       */
      title?: string
    }

export type ActionListDescriptionProps = VariantTypes & { ... }
```

Not a blocker, just an idea! 😁 Let me know what you think!

liuliu-dev: Great suggestion! 🎉 I've added the `DescriptionVariantTypes`.

packages/react/src/ActionList/ActionList.test.tsx
```diff
@@ -171,4 +171,39 @@ describe('ActionList', () => {
...
+          Option 2
+          <ActionList.Description truncate>
+            <span>Complex</span> content
+          </ActionList.Description>
+        </ActionList.Item>
+        <ActionList.Item>
+          Option 3
+          <ActionList.Description truncate title="Custom title">
+            <span>Complex</span> content
+          </ActionList.Description>
```
joshblack: @TylerJDev would this be considered failing: https://dequeuniversity.com/rules/axe/4.7/label-content-name-mismatch ? 🤔 I don't think axe flags it but was curious if that was the case.

If so, it would be great to derive `title` from `props.children` instead of having it be explicit just to avoid the mismatch here when using downstream. In the `props.children` is not a string case, maybe we could use the text content from the rendered content? Just thinking out loud, curious what you think @liuliu-dev and @TylerJDev 👀

TylerJDev: I think as long as the text is provided, `title` will default to the description, rather than accessible name. If we were able to utilize `Tooltip` instead of rely on the `title` "tooltip", it would be best, but that'd be harder to implement 😅 

> In the props.children is not a string case, maybe we could use the text content from the rendered content?

For this, would we need to grab the text after it was rendered? (e.g. accessing the DOM element's text content within the description container)

liuliu-dev: Great points! For extracting text from complex children, I think we can probably do this before DOM rendering using a recursive function that traverses `props.children`. Does this approach make sense?

joshblack: > If we were able to utilize Tooltip instead of rely on the title "tooltip", it would be best, but that'd be harder to implement 😅

Totally makes sense 👍 

> For this, would we need to grab the text after it was rendered?

That was definitely what came to mind for me. Could even be baked into `Truncate` unless there is a reason to have `title` not match the contents 🤔 (I'm just not sure at all lol). But definitely get if this seems tedious/requires some setup. One idea could be update `Truncate` to compute the `title` if an explicit one is not provided like:

<details>
<summary>View example</summary>

```tsx
const Truncate = React.forwardRef(function Truncate(
  {as, expandable = false, inline = false, maxWidth = 125, title, ...rest},
  forwardRef,
) {
  const containerRef = useRef<ElementRef<'div'>>(null)
  const ref = useMergedRefs(containerRef, forwardRef)
  const [computedTitle, setComputedTitle] = useState(title)

  useEffect(() => {
    if (title) {
      return
    }

    const {current: container} = containerRef
    if (!container) {
      return
    }

    const observer = new MutationObserver(() => {
      if (container.textContent) {
        setComputedTitle(container.textContent.trim())
      }
    })

    observer.observe(container, {
      childList: true,
      subtree: true,
    })

    return () => {
      observer.disconnect()
    }
  }, [title])

  return (
    <StyledTruncate
      ref={ref}
      as={as}
      expandable={expandable}
      inline={inline}
      maxWidth={maxWidth}
      title={computedTitle}
      {...rest}
    />
  )
}) as PolymorphicForwardRefComponent<'div', TruncateProps>
```
</details>

@liuliu-dev mentioned another alternative with `children` which could definitely work, I'm just personally not sure how to derive the rendered DOM from the element tree in `children` in order to get the text content.

---

Separately, I'm assuming with this component that ideally `ActionList.Description` would work like:

```tsx
<ActionList.Description truncate>
  This is a description
</ActionList.Description>
```

But since `children` may not always be a `string`, we want to add `title` as a way to workaround this. Is that the case? Just wanted to double-check 👀

TylerJDev: > That was definitely what came to mind for me. Could even be baked into Truncate unless there is a reason to have title not match the contents 🤔 (I'm just not sure at all lol). But definitely get if this seems tedious/requires some setup.

Yup I agree! I don't think there'd be many cases where a custom implementation would be required. If we could take directly from the text contents then I think that would be preferred. 

I feel like even if we did access the DOM directly after render it wouldn't be as bad since the `title` tooltip won't be rendered until some action is taken (hover), so it's not required instantaneously.

> But since children may not always be a string, we want to add title as a way to workaround this. Is that the case? Just wanted to double-check 👀

Yeah! I think that was the main goal, for cases like: 

```tsx
<ActionList.Description truncate>
  <>This is a description</>
  <ExampleComponentWithExtraText />
</ActionList.Description>
```

I'm thinking that if we could get the text directly regardless of what is passed, then we could potentially omit the prop, unless we want to give more control 🤔

liuliu-dev: I agree with both of you, @joshblack and @TylerJDev. Taking the title directly from the text content is the best approach - traversing React nodes to construct the text title is not as reliable. I also agree with @TylerJDev that we might not need the `title` prop since we can get the text directly from the rendered content.
Thank you all for your inputs!

TylerJDev (approved):
LGTM! ✨

PR #6236, opened 2025-06-25
Update vitest config and TreeView tests to Vitest

joshblack:
Updates our TreeView tests from Jest to Vitest. This PR also includes some changes I needed to land in order to convert the tests, namely:

- The vitest browser environment is now set up so that our styles should match how they appear in storybook
- Our test suite will now fail in CI if a `console.*` method is logged (helps to prevent `act()` warnings from sneaking in)

#### Changed

- Update TreeView tests from Jest to Vitest
- Update vitest setup files
  - Split browser-config into specific setup file
  - Use shared setup config across Node.js and browser tests
- Use postcss preset for transforming CSS files in vitest

### Rollout strategy

This is a change to our internal test setup and does not impact our public API.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31892,6 +31892,33 @@
         }
       }
     },
+    "node_modules/vitest-fail-on-console": {
+      "version": "0.7.1",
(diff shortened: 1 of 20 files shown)


TylerJDev (approved):
LGTM!

PR #6239, opened 2025-06-25
Add loading support to ActionList.TrailingAction component

liuliu-dev:
Closes [5346](https://github.com/github/primer/issues/5346)

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/ActionList/ActionList.module.css
+++ b/packages/react/src/ActionList/ActionList.module.css
@@ -628,7 +628,7 @@ default block */
   /* remove after FF ships */
   /* stylelint-disable-next-line primer/typography */
   line-height: 16px;
-  color: var(--fgColor-default);
+  color: var(--fgColor-muted);
 }
 
 /* helper for grid alignment with multi-line content


packages/react/src/ActionList/TrailingAction.tsx
```diff
@@ -44,6 +49,7 @@ export const TrailingAction = forwardRef(
             variant="invisible"
             as={as}
             href={href}
+            loading={loading}
```
TylerJDev, line 55: I noticed when using `loading` without an `Icon` (e.g. `ActionList.TrailingAction as="a" href="#" label="Some action 1" loading />`) the `TrailingAction` takes the full width of the inner contents when loading. I'm wondering if we should add some sort of conditional or style so that the width remains consistent when `loading` is true. 


Example: 

Trailing action within ActionList, that is currently in a loading state. The action itself has a large width, due to the inner text being only visibly hidden via `visibility: hidden` rather than `display: none`

What we might want: 

Trailing Action with a smaller width due to the inner contents being removed.

packages/react/src/ActionList/TrailingAction.tsx
```diff
@@ -44,6 +49,7 @@ export const TrailingAction = forwardRef(
             variant="invisible"
             as={as}
```
TylerJDev, line 53: Non-blocking: I'm wondering if we'd want to convert this to a button when loading state is true (e.g. `as={loading ? undefined : as}`). Mainly because we can't disable a link (`aria-disabled` is added to the loading states of `IconButton`). There are some accessibility considerations, such as if it's confusing to go from a button to a link once the loading state is finished. We'd also need to ensure focus remained on the element once the loading state resolves. An alternative is to only allow loading states for button trailing actions.

I'm indifferent on this though. Curious what you and @joshblack think!

packages/react/src/ActionList/TrailingAction.tsx
```diff
@@ -19,10 +19,14 @@ export type ActionListTrailingActionProps = ElementProps & {
   icon?: React.ElementType
   label: string
   className?: string
+  /**
+   * Specify whether the action is in a loading state
+   */
+  loading?: boolean
```
TylerJDev: I think we should be able to [uncomment the existing `WithTrailingAction` story](https://github.com/primer/react/blob/main/packages/react/src/ActionList/ActionList.features.stories.tsx#L830-L831). We can adding the `loading` prop to one of the `TrailingAction` implementations in that story 😁

packages/react/src/ActionList/TrailingAction.tsx
liuliu-dev: Good catch! I've actually addressed this by aligning the spinner to the right when loading, rather than changing the button width, so the button maintains its natural width and the layout doesn't shift during the loading transition. Let me know what you think 👀

packages/react/src/ActionList/TrailingAction.tsx
liuliu-dev: I prefer the type restriction approach - only allowing loading states for button trailing actions. The element type remains consistent, no changing from a button to a link.

PR #6264, opened 2025-07-01
Fix/update navlist group divider

joshblack:
Closes https://github.com/github/primer/issues/5391

Update the `NavList.Group` component to have the same result when both `sx` and no `sx` prop is provided. This change has an unintended side-effect which is that in the `sx` case no divider was being rendered 😅 

#### Changed

- Consolidate `sx` and no-`sx` scenarios into single return for `NavList.Group`

### Rollout strategy

- [x] Patch release

### DIFF
--- a/packages/react/src/NavList/NavList.tsx
+++ b/packages/react/src/NavList/NavList.tsx
@@ -285,20 +285,11 @@ export type NavListGroupProps = React.HTMLAttributes<HTMLLIElement> & {
   title?: string
 } & SxProp
 
-const defaultSx = {}
-const Group: React.FC<NavListGroupProps> = ({title, children, sx: sxProp = defaultSx, ...props}) => {
-  if (sxProp !== defaultSx) {
-    return (
-      <Box sx={sxProp} as="li" data-component="ActionList.Group">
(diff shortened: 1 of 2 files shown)


packages/react/src/NavList/NavList.tsx
```diff
@@ -285,31 +285,23 @@ export type NavListGroupProps = React.HTMLAttributes<HTMLLIElement> & {
...
-        {children}
-      </Box>
-    )
-  }
+function Group({title, children, sx, ...rest}: NavListGroupProps) {
   return (
     <>
       <ActionList.Divider />
-      <ActionList.Group {...props}>
+      <BoxWithFallback as={ActionList.Group} {...rest} sx={sx}>
```
jonrohan: I was thinking we probably don't need boxwithfallback here since ActionList.Group would have BoxWithFallback?

joshblack: Ooo great point, will update 👍

jonrohan (approved):

joshblack:
I think I need to coordinate this with: https://github.com/github/primer/issues/5407 in order to land in a release and not disrupt memex 👀

PR #6266, opened 2025-07-01
bugfix(Dialog): Fixing issue where page shifts when dialog is open

jonrohan:
Closes https://github.com/github/primer/issues/5376

This PR changes how we turn on disable scrolling when the dialog is open. The strategy was taken from what we do in PVC. https://github.com/primer/view_components/blob/2677324d35cac1ce5ebf9aede851fa4267f7c6e2/app/components/primer/dialog_helper.ts#L81

We essentially set a variable on the body element that equals the width of the scrollbar. And turns padding on when that scrollbar is hidden.

I also used this as an opportunity to cleanup the old inline styles we were doing for hidding, and migrated the test to vitest in the process for the toHaveStyle assertion.

#### Changed

Dialog adds padding to page when scrollbar is hidden.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Dialog/Dialog.module.css
+++ b/packages/react/src/Dialog/Dialog.module.css
@@ -1,3 +1,16 @@
+/* The --dialog-scrollgutter property is used only on the body element to
+ * simulate scrollbar-gutter:stable. This property is not and should not
+ * be used elsewhere in the DOM. There is a performance penalty to
+ * setting inherited properties which can cause a large style recalc to
+ * occur, so it benefits us to prevent inheritance for this property.
+ * See https://web.dev/blog/at-property-performance
+ */
+@property --dialog-scrollgutter {
+  initial-value: 0;
+  inherits: false;
+  syntax: '<length>';
+}
+
 @keyframes dialog-backdrop-appear {
   0% {
     opacity: 0;
@@ -214,6 +227,11 @@
       }
     }
   }
+}
+
+body:has(.Dialog.DisableScroll) {
+  /* stylelint-disable-next-line primer/spacing */
+  padding-right: var(--dialog-scrollgutter) !important;
 }
 
 .DialogOverflowWrapper {
--- a/packages/react/src/Dialog/Dialog.module.css
+++ b/packages/react/src/Dialog/Dialog.module.css
@@ -232,6 +232,7 @@
 body:has(.Dialog.DisableScroll) {
   /* stylelint-disable-next-line primer/spacing */
   padding-right: var(--dialog-scrollgutter) !important;
+  overflow: hidden !important;
 }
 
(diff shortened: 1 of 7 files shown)


packages/react/src/Dialog/Dialog.tsx
```diff
@@ -288,19 +288,7 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
...
-    if (bodyOverflowStyle === 'hidden') {
-      return
-    }
-
-    document.body.style.overflow = 'hidden'
-
-    return () => {
-      document.body.style.overflow = bodyOverflowStyle
-    }
+    document.body.style.setProperty('--dialog-scrollgutter', `${window.innerWidth - document.body.clientWidth}px`)
```
jonrohan, line 291: I cleared up this overflow style setting since we can include it in the class. The property I'm setting here is identical to what we have in PVC https://github.com/primer/view_components/blob/2677324d35cac1ce5ebf9aede851fa4267f7c6e2/app/components/primer/dialog_helper.ts#L81

But I want to know if `useEffect` is the correct callback to have this in?

packages/react/src/Dialog/Dialog.tsx
```diff
@@ -339,7 +327,7 @@ const _Dialog = React.forwardRef<HTMLDivElement, React.PropsWithChildren<DialogP
             data-width={width}
             data-height={height}
             sx={sx}
-            className={clsx(className, classes.Dialog)}
+            className={clsx(className, classes.Dialog, classes.DisableScroll)}
```
jonrohan, line 330: Opted to add a separate class here in case we wanted to make this a prop in the future, which is what we do in PVC

packages/react/src/Dialog/Dialog.module.css
```diff
@@ -1,3 +1,16 @@
...
+ * setting inherited properties which can cause a large style recalc to
+ * occur, so it benefits us to prevent inheritance for this property.
+ * See https://web.dev/blog/at-property-performance
+ */
+@property --dialog-scrollgutter {
+  initial-value: 0;
+  inherits: false;
+  syntax: '<length>';
+}
+
```
jonrohan, line 13: Most of this implementation was taken from what we did in PVC https://github.com/primer/view_components/pull/2542

packages/react/src/Dialog/Dialog.tsx
joshblack: @jonrohan I think that's the right place to have it 👍

PR #6329, opened 2025-07-16
SelectPanel `Select all` feature

llastflowers:
Closes https://github.com/github/primer/issues/5348

#### New

- Added `Select all` section with checkbox in SelectPanel
    - `Select all` checkbox selects all filtered items
    - Manually selecting all items automatically checks `Select all` box
    - `Select all` label changes to `Deselect all` when all items are selected
    - `Deselect all` checkbox deselects all filtered items
    - Manually checking some items but not all results in `indeterminate` checkbox state
    - Clicking `Select all` checkbox in `indeterminate` state selects all remaining filtered items
- Added `With Select All` story for SelectPanel

#### Changed

N/A

#### Removed

N/A

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

This feature is designed to visually match the ModernActionList version of SelectPanel, so please toggle that feature flag on in Storybook when testing. (It still works without the FF, but just looks a bit goofy with the old SelectPanel design).

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/FilteredActionList/FilteredActionList.module.css
+++ b/packages/react/src/FilteredActionList/FilteredActionList.module.css
@@ -13,3 +13,22 @@
     }
   }
 }
+
+.SelectAllContainer {
+  display: flex;
+  align-items: center;
+  padding-block: var(--base-size-12);
+  padding-inline: var(--base-size-16);
+  background: var(--bgColor-muted);
+  border-bottom: var(--borderWidth-thin) solid var(--borderColor-default);
+}
+
+.SelectAllCheckbox {
+  margin-right: var(--base-size-8);
+  margin-top: 0;
+}
+
+.SelectAllLabel {
+  font-size: var(--text-body-size-medium);
+  color: var(--fgColor-muted);
+}
--- a/packages/react/src/FilteredActionList/FilteredActionList.module.css
+++ b/packages/react/src/FilteredActionList/FilteredActionList.module.css
@@ -24,8 +24,7 @@
 }
 
 .SelectAllCheckbox {
-  margin-right: var(--base-size-8);
-  margin-top: 0;
+  margin: var(--base-size-4) var(--base-size-8) var(--base-size-4) 0;
 }
 
 .SelectAllLabel {
--- a/packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
+++ b/packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
@@ -45,6 +45,10 @@ export interface FilteredActionListProps
   className?: string
   announcementsEnabled?: boolean
   showSelectAll?: boolean
+  selectAllChecked?: boolean
+  selectAllIndeterminate?: boolean
+  onSelectAllChange?: (checked: boolean) => void
+  selectAllLabel?: string
 }
 
 const StyledHeader = styled.div`
@@ -129,6 +133,10 @@ export function FilteredActionList({
   className,
   announcementsEnabled = false,
   showSelectAll = false,
+  selectAllChecked = false,
+  selectAllIndeterminate = false,
+  onSelectAllChange,
+  selectAllLabel,
   ...listProps
 }: FilteredActionListProps): JSX.Element {
   const [filterValue, setInternalFilterValue] = useProvidedStateOrCreate(externalFilterValue, undefined, '')
@@ -148,8 +156,7 @@ export function FilteredActionList({
   const activeDescendantRef = useRef<HTMLElement>()
   const listId = useId()
   const inputDescriptionTextId = useId()
-  const [selectAllChecked, setSelectAllChecked] = useState(false)
-  const [selectAllIndeterminate, setSelectAllIndeterminate] = useState(false)
+  const selectAllLabelText = selectAllLabel || (selectAllChecked ? 'Deselect all' : 'Select all')
   const onInputKeyPress: KeyboardEventHandler = useCallback(
     event => {
       if (event.key === 'Enter' && activeDescendantRef.current) {
@@ -225,6 +232,15 @@ export function FilteredActionList({
 
   useScrollFlash(scrollContainerRef)
 
+  const handleSelectAllChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
+      if (onSelectAllChange) {
+        onSelectAllChange(e.target.checked)
+      }
+    },
+    [onSelectAllChange],
+  )
+
   return (
     <Box
       display="flex"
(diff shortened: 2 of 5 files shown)


packages/react/src/SelectPanel/SelectPanel.features.stories.tsx
```diff
@@ -884,3 +884,44 @@ export const WithInactiveItems = () => {
     </FormControl>
   )
 }
+
+export const WithSelectAll = () => {
```
francinelucca, line 892: 👀  just dropping by, let's make sure to add this new story to e2e 🙏

packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
```diff
@@ -249,6 +271,20 @@ export function FilteredActionList({
         />
       </StyledHeader>
       <VisuallyHidden id={inputDescriptionTextId}>Items will be filtered as you type</VisuallyHidden>
+      {showSelectAll && (
```
francinelucca: this doesn't hurt but for future reference we don't need to support new features in deprecated

llastflowers: Makes sense! I kinda just thought it'd look weird in Storybook without it in this case, and TypeScript got mad when the props were missing from the deprecated version 😅

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
```diff
@@ -70,6 +76,11 @@ export function FilteredActionList({
   className,
   announcementsEnabled = true,
   fullScreenOnNarrow,
+  showSelectAll = false,
+  selectAllChecked = false,
+  selectAllIndeterminate = false,
+  onSelectAllChange,
+  selectAllLabel,
```
francinelucca: the component has access to the items, is there a world where all these props can just be consolidated to `onSelectAllChange`, where if it is present then the selectAll is showed, if all items are selected=true, then it is selected, and if at least one item is selected=true, then it is indeterminate? 👀

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -107,6 +107,7 @@ interface SelectPanelBaseProps {
    * @default undefined (uses feature flag default)
    */
   disableFullscreenOnNarrow?: boolean
+  showSelectAll?: boolean
```
francinelucca, line 110: let's make sure to document in the selectpanel.docs.json

packages/react/src/FilteredActionList/FilteredActionListWithModernActionList.tsx
francinelucca: Also am I wrong or is the selectAlllabel not being used (i.e., not being passed into the FilteredActionList from SelectPanel) ? I honestly think we can do without it as well unless/until requirements change

llastflowers: 1. My understanding is that all of the selection state logic is owned by SelectPanel currently, and the FAL is just rendering the list as determined by SelectPanel. So I think there would need to be more significant refactoring to give FAL all the context it needs to manage the list selection state internally. If I'm off base with this, lmk!

2. Sure, I can remove it!

francinelucca: the items determine what is/isn't selected which ultimately is controlled by the consumer 🤔. Can you expand on what refactor would be needed to support this on FilteredActionList instead? as I understand it it'd be more or less just moving the (new) logic that is now in SelectPanel down to FilteredActionList, but happy to chat/pair otherwise!

llastflowers: I think where I'm unsure of how this would work is when the list of items is filtered--doesn't FAL only "see" the subset of filtered items, and not have context about the overall selection state of all of the available items?

This might actually expose a potential a lack of clarity around how the `Select all` feature should function when used with filters. From a UX standpoint, I think the most intuitive functionality would be for the `Select all` checkbox to only control (and visually represent) the visible items, but not overwrite the global selection state when used while the items are filtered. 

But I guess that depends on how this is actually going to be used, and I don't have as much context around that. Do we want to be able to maintain the global selection state while filtering and using `Select all`? If so, this is where I think issues would arise with FAL not having access to all of the information it needs to do this internally.

SelectPanel and FAL also appear to have a pretty clear separation of concerns regarding control of selection logic and overall selection state management vs. reporting user interactions and displaying what is visibly selected, so it was also my intent to follow this pattern in my implementation.

Let me know if this makes sense! I'm happy to discuss further if needed.

TylerJDev:
> is tab to get to the checkbox all right or should arrow down take you there?

@francinelucca, I think tab is okay. I could also see the case for it being part of the items list, but I think for now being to access it via tab works.

llastflowers:
> > is tab to get to the checkbox all right or should arrow down take you there?
> 
> @francinelucca, I think tab is okay. I could also see the case for it being part of the items list, but I think for now being to access it via tab works.

@TylerJDev FYI, in [this comment](https://github.com/github/accessibility/issues/9094#issuecomment-3069959551), there were some potential issues raised against having it be part of the items list 👀

francinelucca (approved):

packages/react/src/FilteredActionList/FilteredActionList.module.css
```diff
@@ -13,3 +13,21 @@
     }
   }
 }
+
+.SelectAllContainer {
+  display: flex;
+  align-items: center;
+  padding-block: var(--base-size-12);
```
langermank:
```suggestion
  padding-block: var(--base-size-4);
```

This is a bit nitpicky and I'm not sure if you're working with a design for this, but I think the padding is a little chonky. I wonder if we could swap to 4px. This could be a followup PR for sure!

Padding

langermank (approved):
Left a non blocking suggestion but could easily be a followup!

packages/react/src/FilteredActionList/FilteredActionList.module.css
llastflowers: Oh yeah, I was going from the designs [here](https://github.com/github/primer/issues/5348) and [here](https://www.figma.com/design/lIPxmElkJmh4zKRnCMoxG2/Teams--Org---Enterprise-?node-id=4-36830&t=h2js9qvrVoxEbh1u-0)--they are both a little different. What do you think based on that? I could definitely make it a bit smaller!

PR #6346, opened 2025-07-21
SelectPanel: Add default empty message to announcement

TylerJDev:
**Addresses multiple audit issues:** https://github.com/github/accessibility-audits/issues/12437, https://github.com/github/accessibility-audits/issues/11114

Ensures that the announcement for no items takes from `DefaultEmptyMessage` instead

#### Changed

* Adds default message for empty announcements in `FilteredActionList` and `SelectPanel`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


francinelucca (commented):
Tagged @llastflowers for review since this might intersect with the work she's doing as part of https://github.com/github/primer/issues/5377

packages/react/src/SelectPanel/SelectPanel.stories.tsx
```diff
@@ -88,15 +88,6 @@ export const Default = () => {
...
         width="medium"
-        message={
-          filteredItems.length === 0
-            ? {
-                variant: 'empty',
-                title: `No language found for \`${filter}\``,
-                body: 'Adjust your search term to find other languages',
-              }
-            : undefined
-        }
```
llastflowers: Please correct me if I'm wrong, but I don't think we want to get rid of this, because then when a user tries to filter existing list items but no results are found for their search terms, the message displayed is:

```
You haven’t created any items yet
Please add or create new items to populate the list.
```

We still want to display something like this in that case, right?

```
No language found for `hello`
Adjust your search term to find other languages
```

I think we can either keep this block or replace it with:

```
message={filteredItems.length === 0 ? NoResultsMessage(filter) : undefined}
```

and then also add at the top (same as what is done in `SelectPanel.features.stories.tsx`):

```
const NoResultsMessage = (filter: string): {variant: 'empty'; title: string; body: string} => {
  return {
    variant: 'empty',
    title: `No language found for \`${filter}\``,
    body: 'Adjust your search term to find other languages',
  }
}
```

Just keeping the original block might be simpler since it doesn't seem like we need the more complex conditionals for the type of message displayed in this story like we do for `CustomisedNoInitialItems` in `SelectPanel.features.stories.tsx`.

llastflowers (commented):
I left one comment, and the rest of the changes look fine to me! I'm wondering if the original message mismatch with the announcement had to do with the two different types of 'empty' messages being conflated? 🤔 

Regarding my work on https://github.com/github/primer/issues/5377, I think I can build those changes on top of this without any issues!

packages/react/src/SelectPanel/SelectPanel.stories.tsx
TylerJDev: Ah woops, I didn't mean to delete this 😅  I was testing on my local what a screen reader would announce if one wasn't provided it, but thought I added it back after testing 😆 Thanks for the callout!

llastflowers: Ahh makes sense 😹

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -30,10 +30,14 @@ import type {ButtonProps, LinkButtonProps} from '../Button/types'
 // we add a delay so that it does not interrupt default screen reader announcement and queues after it
 const SHORT_DELAY_MS = 500
 const LONG_DELAY_MS = 1000
+const EMPTY_MESSAGE = {
+  title: 'You haven’t created any items yet',
+  description: 'Please add or create new items to populate the list.',
+}
```
TylerJDev: Added a const to hold our default messages. If we ever decide not to provide a default, we'll need to remove this.

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -228,12 +232,12 @@ function Panel({
   const onListContainerRefChanged: FilteredActionListProps['onListContainerRefChanged'] = useCallback(
     (node: HTMLElement | null) => {
       setListContainerElement(node)
-      if (!node && needsNoItemsAnnouncement) {
+      if (!node && needsNoItemsAnnouncement && !usingModernActionList) {
```
TylerJDev: I don't think we need to announce through SelectPanel if it's not using the modern action list. I was getting multiple of the same announcements, some through `useAnnouncements` and others through this file.

packages/react/src/FilteredActionList/FilteredActionListWithDeprecatedActionList.tsx
```diff
@@ -42,6 +42,10 @@ export interface FilteredActionListProps
   inputRef?: React.RefObject<HTMLInputElement>
   className?: string
   announcementsEnabled?: boolean
+  messageText?: {
+    title: string
+    description: string
+  }
```
TylerJDev, line 50: This is just so TypeScript doesn't yell. Won't be needed once we remove the FF.

joshblack (approved):

PR #6363, opened 2025-07-22
SelectPanel: Remove `FormControl.Label` link with SelectPanel

TylerJDev:
Closes https://github.com/github/primer/issues/4977, https://github.com/github/primer/issues/4976

Removes link between `SelectPanel` and `FormControl.Label`, and instead suggests to utilize `aria-labelledby`

#### Added

* `isReferenced` prop in `FormControl.Label` to exclude `for` attribute when used with `SelectPanel`

#### Changed

* Added examples of `aria-labelledby` with `SelectPanel` and `FormControl.Label`

#### Removed

* Removed `SelectPanel` from `FormControl.Label`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
My original idea was to reference the label from the anchor in SelectPanel: e.g. 

```html
<label id="label-example">Label</label>
<button id="btn-example" aria-labelledby="label-example btn-example">SelectPanel Trigger</button>
```

This required more work in `FormControl.Label` and `SelectPanel`, as it requires us to create a new ID for both the trigger and the label, and makes us need to utilize the `FormControl` context within `SelectPanel`. Wasn't sure if it was worth it, as the alternative is for the consumer to provide the attributes directly. Ideally, we'd provide the label from the component, as this would make the link between the 2 easier to handle, but for now I think this will at least solve the issue and provide examples for how it could be done. Curios what y'all think, or if there might be an alternative path here!

cc: @francinelucca, @hectahertz

hectahertz:
>  Ideally, we'd provide the label from the component, as this would make the link between the 2 easier to handle, but for now I think this will at least solve the issue and provide examples for how it could be done.

I agree, LGTM!

PR #6371, opened 2025-07-23
Update package.json to modify react-is dependency

mattcosta7:
This pull request updates the dependencies in the `packages/react/package.json` file, primarily focusing on upgrading `react`-related packages and ensuring compatibility with new versions. The most significant changes include the addition and upgrade of `react-is` and its associated type definitions, as well as adjustments to peer dependencies.

Primer fails to install in a react 19 project because the dependency on react-is conflicts, by making it a peer we avoid that conflcit

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/package.json
```diff
@@ -218,9 +216,11 @@
   "peerDependencies": {
     "@types/react": "18.x || 19.x",
     "@types/react-dom": "18.x || 19.x",
+    "@types/react-is": "18.x || 19.x",
```
mattcosta7, line 221: optional peer

packages/react/package.json
```diff
@@ -107,7 +106,6 @@
     "lodash.isobject": "^3.0.2",
     "react-compiler-runtime": "^19.1.0-rc.2",
     "react-intersection-observer": "^9.16.0",
-    "react-is": "^18.2.0",
```
hectahertz, line 110: Why did you move this to devdependencies? It is imported and used in the components code

hectahertz (changes requested):

hectahertz (approved):

joshblack:
For context on removing from the queue, just was waiting for the integration PR to pass 👀 

The only thing I'm worried about is adding a new `peerDependency` here in a minor release since they are not auto-installed so we will need to update install instructions to include this now.

Could we instead update the range on the dependency? e.g. `"react-is": "18.x || 19.x"` or would that not work as expected?

mattcosta7:
> For context on removing from the queue, just was waiting for the integration PR to pass 👀
> 
> The only thing I'm worried about is adding a new `peerDependency` here in a minor release since they are not auto-installed so we will need to update install instructions to include this now.
> 
> Could we instead update the range on the dependency? e.g. `"react-is": "18.x || 19.x"` or would that not work as expected?

NPM started installing peerDependencies by default in version 7 https://github.blog/news-insights/product-news/npm-7-is-now-generally-available/#peer-dependencies

> Automatically installing peer dependencies is an exciting new feature introduced in npm 7

I don't think we should be installing a version of it for consumers and 2 versions installed could cause some conflcits I think? (although they're generally backwards compat).  With that said, a broader dependency might be ok - but is probably not ideal?

joshblack:
Oh nice! Didn't realize that was added (thanks for sharing)

> I don't think we should be installing a version of it for consumers and 2 versions installed could cause some conflcits I think?

What scenarios would you expect to have two versions installed? Was trying to think through it:

| Type | Primer | Consumer | Installed |
| :--- | :------ | :------- | :-------- |
| dependencies | v18 | v19 | v18, v19 |
| peerDependencies | v18 | v19 | Would this be an error with the new behavior in npm v7? |
| dependencies | v18 \|\| v19 | v19 | v19 |
| peerDependencies | v18 \|\| v19 | v19 | v19 |

Are there any that I'm missing? 👀

mattcosta7:
> Oh nice! Didn't realize that was added (thanks for sharing)
> 
> > I don't think we should be installing a version of it for consumers and 2 versions installed could cause some conflcits I think?
> 
> What scenarios would you expect to have two versions installed? Was trying to think through it:
> 
> Type	Primer	Consumer	Installed
> dependencies	v18	v19	v18, v19
> peerDependencies	v18	v19	Would this be an error with the new behavior in npm v7?
> dependencies	v18 || v19	v19	v19
> peerDependencies	v18 || v19	v19	v19
> Are there any that I'm missing? 👀

I think they generally _should_ resolve to a single version, but I could see npm nesting one inside the pacakge instead when upgrading - I don't think it's a _common_ case where that would happen instead of hosting, but maybe not impossible?

joshblack:
Sounds good 👍 Definitely down to try this out. Thanks for making the PR!

mattcosta7:
> Sounds good 👍 Definitely down to try this out. Thanks for making the PR!

no problem!  thank you!

PR #6417, opened 2025-07-30
feat: add @primer/styled-react package

joshblack:
Closes https://github.com/github/primer/issues/5584

Add support for temporary package to help bridge between styled-components usage and `@primer/react`

#### New

- Add new package, `@primer/styled-react` that will be the adapter between styled-components usage and `@primer/react` when we remove support for styled-components


.changeset/spotty-coats-post.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/styled-react': minor
```
jonrohan, line 2: Just double checking we're defaulting minor for all releases of this project?

packages/styled-react/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+```bash
+npm install -S @primer/styled-react
+```
+
+## Usage
+
+This is an interim package designed to make it easy to interop between Primer
+and existing styled-components usage. Over time, this component will stop
+receiving updates.
```
jonrohan: Wanting some wording in here about

* This is pre-major and will have breaking changes
* The components in here are temporary until we have determined they've been successfully migrated away.

Not married to the wording so whatever you want to fix here is 👍🏻 with me.

```suggestion
receiving updates. And the components in here will be removed when we determine they are no longer in use without a major release.
```

packages/styled-react/src/__tests__/exports.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  expect(Object.keys(StyledReact)).toMatchSnapshot()
+})
+
+test('@primer/styled-react/deprecated exports', () => {
+  expect(Object.keys(StyledReactDeprecated)).toMatchSnapshot()
+})
+
+test('@primer/styled-react/experimental exports', () => {
+  expect(Object.keys(StyledReactExperimental)).toMatchSnapshot()
+})
```
jonrohan, line 19: Not sure if there's some kinda test we could write that would verify the component is the same? Not a blocker, and might be overkill.

.changeset/spotty-coats-post.md
joshblack: That was my default but no preference on my end!

packages/styled-react/src/__tests__/exports.test.ts
joshblack: We could do a type test if that would be helpful, like make sure that every type matches the one from `@primer/react`. That wouldn't really be behavior/visual though if that's more so what you're looking for 👀

packages/styled-react/README.md
joshblack: Just pushed an update, let me know what you think!

jonrohan (approved):
Looks good :shipit:

packages/styled-react/src/index.ts
```diff
@@ -0,0 +1,18 @@
+export {
```
francinelucca, line 1: when does the wrapping happen? 🤔 or as we remove the sx from @primer/react we'd come here and do that wrapping? 👀

joshblack: > as we remove the sx from @primer/react we'd come here and do that wrapping?

Exactly 💯

francinelucca (approved):

PR #6425, opened 2025-07-31
feat: only publish esm bundles from @primer/react

joshblack:
Part of https://github.com/github/primer/issues/5541

Update `@primer/react` to be ESM-only

#### Changed

- Update our package's `"type"` field to be `"module"`
- Update our package's `"exports"` configuration so that:
  - Each entrypoint has a `types` condition
  - Each entrypoint has a `default` condition (which will not be ESM)
- Add a `lint:publish` step that checks that our package is setup correctly with its types
- Update scripts that failed now that we're running in an ESM context instead of a CommonJS context by default

#### Removed

- Removed the CommonJS and UMD builds from rollup

### Rollout strategy

We are considering this a major release since the package is no longer publishing ESM bundles. With recent releases of Node.js, we should not need to handle migration (since you can now import ESM from a CommonJS context) but we will test against dotcom to confirm.


francinelucca (approved):
LGTM pending failing IC

PR #6426, opened 2025-07-31
feat: add support for turbo to project

joshblack:
As we add more packages to the project, it's helpful to have a tool that will:

- Run tasks in the correct order (and in parallel, when possible)
- Has support for caching output of common tasks (like build)

In this PR, we're adding `turbo` to the project to support both of these scenarios. Initially, it will be used for `build`, `build:storybook`, and `type-check` tasks across the project. This PR also includes support for a build cache in GitHub Actions to reduce the time we spend re-building packages that have not changed.

#### New

- Add support for `turbo` to the project

#### Changed

- Update GitHub Actions workflows to use the turbo cache and appropriate commands, where applicable

### DIFF
--- a/.github/workflows/aat-reports.yml
+++ b/.github/workflows/aat-reports.yml
@@ -34,7 +34,6 @@ jobs:
         run: npm ci
       - name: Build storybook
         run: npx turbo build:storybook
-        working-directory: packages/react
       - name: Run storybook
         id: storybook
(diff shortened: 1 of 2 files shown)


jonrohan (approved):
**lgtmrnamt** _(looks good to me right now and maybe tomorrow)_

PR #6156, opened 2025-06-02
TreeView: Add `TreeView.TrailingAction` subcomponent

TylerJDev:
Closes #

Adds new sub-component `TreeView.TrailingAction`, which allows you to add supplemental actions that are visually appended to the end of `TreeView.Item`.

#### New

* Added new sub component `TreeView.TrailingAction`

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [x] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #6456, opened 2025-08-04
TreeView: Add TreeView.TrailingAction accessible via prop

TylerJDev:
Related to https://github.com/primer/react/pull/6156

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/TreeView/TreeView.module.css
+++ b/packages/react/src/TreeView/TreeView.module.css
@@ -48,7 +48,7 @@
     cursor: pointer;
     border-radius: var(--borderRadius-medium);
     grid-template-columns: var(--spacer-width) var(--leading-action-width) var(--toggle-width) 1fr;
-    grid-template-areas: 'spacer leadingAction toggle content';
+    grid-template-areas: 'spacer leadingAction toggle content trailingAction';
 
     --leading-action-width: calc(var(--has-leading-action, 0) * 1.5rem);
     --spacer-width: calc(calc(var(--level) - 1) * (var(--toggle-width) / 2));
@@ -176,6 +176,16 @@
     display: flex;
     color: var(--fgColor-muted);
     grid-area: leadingAction;
+
+    & > button {
+      flex-shrink: 1;
+    }
+  }
+
+  .TreeViewItemTrailingAction {
+    display: flex;
+    color: var(--fgColor-muted);
+    grid-area: trailingAction;
 
     & > button {
       flex-shrink: 1;
--- a/packages/react/src/TreeView/TreeView.test.tsx
+++ b/packages/react/src/TreeView/TreeView.test.tsx
@@ -4,6 +4,8 @@ import {beforeEach, afterEach, describe, it, expect, vi} from 'vitest'
 import React from 'react'
 import type {SubTreeState} from './TreeView'
 import {TreeView} from './TreeView'
+import {IconButton} from '../Button'
+import {GearIcon} from '@primer/octicons-react'
 
 // TODO: Move this function into a shared location
 function renderWithTheme(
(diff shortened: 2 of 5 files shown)


packages/react/src/TreeView/TreeView.tsx
```diff
@@ -250,17 +269,36 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
...
+            if (!(event.shiftKey && (event.metaKey || event.ctrlKey))) return
+
+            if (secondaryActions.length > 1) {
+              // If there are multiple secondary actions, open the action dialog
+              // as this allows users to select the action they want to interact with.
+              setActionCommandPressed(true)
+            } else {
+              // If there is only one secondary action, trigger it directly
+              const action = secondaryActions[0].onClick
+              action()
```
TylerJDev: We could avoid utilizing the dialog if there's only 1 action. In order to enhance this, we might need to allow the shortcut text to be customizable to describe what that action does. For pointer devices, this can be conveyed with the tooltip.

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -250,17 +269,36 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
...
-      [onSelect, setIsExpandedWithCache, toggle],
+      [onSelect, setIsExpandedWithCache, toggle, secondaryActions],
     )
 
     const ariaDescribedByIds = [
       slots.leadingVisual ? leadingVisualId : null,
       slots.trailingVisual ? trailingVisualId : null,
     ].filter(Boolean)
 
+    const shortcut = `Shift+${isMacOS ? 'Meta' : 'Control'}+U`
```
TylerJDev, line 302: `Meta` will be converted to `Command` via `getAccessibleKeybindingHintString`

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -250,17 +269,36 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
...
+    const trailingActionShortcutText = `Press (${getAccessibleKeybindingHintString(shortcut, isMacOS)}) for more actions.`
```
TylerJDev, line 303: Following my previous comment, should we make part of this customizable depending on the action(s)?

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -670,6 +722,102 @@ const LeadingAction: React.FC<TreeViewVisualProps> = props => {
...
+      onKeyDown={event => {
+        if (['Backspace', 'ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown', 'Enter'].includes(event.key)) {
+          // Prevent keyboard events from bubbling up to the TreeView
+          // and interfering with keyboard navigation
+          event.stopPropagation()
+        }
+      }}
+    >
+      <Dialog
+        title="Supplemental actions"
```
TylerJDev, line 805: Should the title dialog be customizable?

packages/react/src/TreeView/TreeView.module.css
```diff
@@ -182,6 +182,16 @@
     }
   }
 
+  .TreeViewItemTrailingAction {
+    display: flex;
+    color: var(--fgColor-muted);
+    grid-area: trailingAction;
+
+    & > button {
```
joshblack: Would it be possible to have a class on this `button` to avoid targeting specific elements or does this selector need to be generic?

```suggestion
    & > button {
```

packages/react/src/TreeView/TreeView.test.tsx
```diff
@@ -1678,6 +1679,84 @@ describe('Asynchronous loading', () => {
   })
 })
 
+it('should render `TrailingAction`', async () => {
+  const {getByRole} = renderWithTheme(
```
joshblack: Random, does `getByRole` need to be used from `renderWithTheme` or does `screen.getByRole` work? Wasn't sure if this one behave different so just wanted to double-check

```suggestion
  renderWithTheme(
```

packages/react/src/TreeView/TreeView.test.tsx
joshblack, line 1: Would it be possible to add a test for activating the secondary actions / making assertions on the dialog that shows up?

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -250,17 +269,36 @@ const Item = React.forwardRef<HTMLElement, TreeViewItemProps>(
...
+              // If there is only one secondary action, trigger it directly
+              const action = secondaryActions[0].onClick
+              action()
+            }
+
             break
         }
       },
-      [onSelect, setIsExpandedWithCache, toggle],
+      [onSelect, setIsExpandedWithCache, toggle, secondaryActions],
```
joshblack: I'm not sure if this memoization is worth it since this `useCallback()` will be re-created each render if `secondaryActions` is an array (am I understanding that right?). Would it be worth just dropping down to a regular function and let react compiler optimize this?

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -670,6 +721,114 @@ const LeadingAction: React.FC<TreeViewVisualProps> = props => {
...
+  const {trailingActionId, itemId} = React.useContext(ItemContext)
+  const {items, shortcutText} = props
+
+  return (
+    <>
+      <div id={trailingActionId} className={clsx('PRIVATE_VisuallyHidden', classes.TreeViewVisuallyHidden)}>
+        ; {shortcutText}
+      </div>
+      <div
+        className={clsx('PRIVATE_TreeView-item-trailing-action', classes.TreeViewItemTrailingAction)}
```
joshblack: Just wanted to ask, since this seems to be net-new do we need the `PRIVATE_` classes?

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -670,6 +721,114 @@ const LeadingAction: React.FC<TreeViewVisualProps> = props => {
...
 LeadingAction.displayName = 'TreeView.LeadingAction'
+// ----------------------------------------------------------------------------
+// TreeView.TrailingAction
+
+export type TreeViewTrailingAction = {
+  items: TreeViewSecondaryActions[]
+  shortcutText: string
+}
+
+const TrailingAction: React.FC<TreeViewTrailingAction> = props => {
```
joshblack:
```suggestion
const TrailingAction = (props: TreeViewTrailingAction) => {
```

Generally I think we don't need to use `React.FC` anymore if you just want to annotate the props and treat this as a normal function

packages/react/src/TreeView/TreeView.module.css
TylerJDev: Definitely! Added a class instead.

packages/react/src/TreeView/TreeView.test.tsx
TylerJDev: It looks like `screen.getByRole` works! I replaced `renderWithTheme` with that.

packages/react/src/TreeView/TreeView.tsx
TylerJDev: I replaced it with a regular class!

packages/react/src/TreeView/TreeView.tsx
TylerJDev: Should be addressed now.

packages/react/src/TreeView/TreeView.tsx
TylerJDev: Does this mean we'd create a function solely for the `secondaryActions` case in the switch, and use that within the switch case? e.g. 

```tsx
const exampleFn = () => {
    if (secondaryActions.length > 1) {
      setActionCommandPressed(true)
    } else {
      const action = secondaryActions[0].onClick
      action()
    }
}
```

Let me know if I'm understanding this wrong! 😁

TylerJDev:
@joshblack, don't worry about re-reviewing this PR if you don't have time! I nearly forgot I implemented some of the feedback but forgot to actually request another review 😅 I've assigned another reviewer to check it out 😁

joshblack (approved):

PR #6478, opened 2025-08-06
Update Select stories to no longer use styled-components

llastflowers:
Closes https://github.com/github/primer/issues/5625
Closes https://github.com/github/primer/issues/5626
Closes https://github.com/github/primer/issues/5627

#### New

Added 2 new CSS modules

#### Changed

Changed `Select.features.stories.tsx` and `Select.stories.tsx` and `Select.dev.stories.tsx` to no longer use styled-components

Changed the className prop to be on TextInputWrapper instead of Select, so that using className on Select components styles the wrapper correctly (as was already happening when using the sx prop, which was correctly passed to TextInputWrapper).

#### Removed

Removed anything associated with styled-components

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Select/Select.dev.stories.tsx
```diff
@@ -1,17 +1,18 @@
...
   component: Select,
 } as Meta
 
 export const Default = () => (
-  <Box as="form">
+  <form>
     <FormControl>
       <FormControl.Label>Default label</FormControl.Label>
-      <Select sx={{color: 'danger.fg'}}>
+      <Select wrapperClassName={classes.DangerText}>
```
pksjce: Is this a naming convention? Why not `classes`?

packages/react/src/Select/Select.tsx
```diff
@@ -57,16 +59,17 @@ const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
         disabled={disabled}
         size={size}
         validationStatus={validationStatus}
-        className={classes.TextInputWrapper}
+        className={clsx(classes.TextInputWrapper, className)}
         sx={sx}
+        data-testid={dataTestId}
```
jonrohan: There's [one spot in dotcom](https://primer-query.githubapp.com/?query=name%3A%22Select%22+attribute%3A%22data-testid%22) that look like they would not expect this change https://github.com/github/github/blob/1d45981872f2facdfabb585eda8027e7e52dc762/ui/packages/github-hosted-runners-settings/views/components/shared/ImageUploadValuesSelector.tsx#L37

llastflowers: Okay, I updated the test to instead look for the test className on the outer <span> wrapper of Select, which is the TextInputWrapper. This should verify the correct passing of className from Select to TextInputWrapper without causing any potential issues with the testid prop.

jonrohan (approved):
👍🏻

PR #6484, opened 2025-08-06
feat(mcp): add server export to package

joshblack:
Exports our MCP server from the `@primer/mcp` package in case we want to play around with hosting it at primer.style 👀 

This should allow us to do the following:

```tsx
// src/app/mcp/route.ts

import {server} from '@primer/mcp';
import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";

export async function POST(req, res) {
  // In stateless mode, create a new instance of transport and server for each request
  // to ensure complete isolation. A single instance would cause request ID collisions
  // when multiple clients connect concurrently.
  
  try {
    const transport: StreamableHTTPServerTransport = new StreamableHTTPServerTransport({
      sessionIdGenerator: undefined,
    });
    res.on('close', () => {
      console.log('Request closed');
      transport.close();
      server.close();
    });
    await server.connect(transport);
    await transport.handleRequest(req, res, req.body);
  } catch (error) {
    console.error('Error handling MCP request:', error);
    if (!res.headersSent) {
      res.status(500).json({
        jsonrpc: '2.0',
        error: {
          code: -32603,
          message: 'Internal server error',
        },
        id: null,
      });
    }
  }
}
```

Reference: https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#streamable-http

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -6,5 +6,8 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["example-*", "codesandbox"]
+  "ignore": ["example-*", "codesandbox"],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "onlyUpdatePeerDependentsWhenOutOfRange": true
+  }
 }
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -6,7 +6,7 @@
 
 ## Getting started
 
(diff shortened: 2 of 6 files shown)


jonrohan (approved):
Awesome

pksjce:
So I can give prompts to build ui with primer react components? How? where?

joshblack:
@pksjce just an idea we're playing around with! Hopefully it works out 😅 You can install it in VSCode here: https://github.com/primer/react/tree/main/packages/mcp#getting-started using the badge. Then just make to start the MCP server after installing it

PR #6509, opened 2025-08-07
fix(Banner): update layout to only create explicit row when actions are defined

joshblack:
Closes https://github.com/github/primer/issues/5652

There was a space being generated in smaller viewports when no actions were present in a Banner. This was due to us creating two explicit rows and we had a row-gap defined. This change updates our rule to only create explicit rows when we have actions defined

#### Changed

- Update layout logic within Banner to accommodate no actions

### Rollout strategy

- [x] Patch release


pksjce:
Seems like there's a Banner snapshot that's failing inside TreeView. Updating should fix it.

PR #5739, opened 2025-03-06
Working on actionList code connect

lukasoppermann:
This PR adds the codeConnect file for ActionList that helps show up code in Figma.

CleanShot 2025-03-11 at 08 36 06@2x

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [x] None; if selected, include a brief description as to why
No change to react, the standard process for those changes is to not increase the version.

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #6546, opened 2025-08-08
chore: remove type-check from dependencies

Uzlopak:
This removes the dependency type-check from @primer/react, which was introduced in #5739

I think, it was accidently introduced by @lukasoppermann, when he was running mistakingly `npm i type-check` instead of `npm run type-check`.


hectahertz:
Thanks for spotting this!

PR #6548, opened 2025-08-08
chore: remove react-compiler-runtime

Uzlopak:
This package was introduced in #6211 but is not used anywhere in the codebase.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24082,15 +24082,6 @@
         "node": ">=0.10.0"
       }
     },
-    "node_modules/react-compiler-runtime": {
-      "version": "19.1.0-rc.2",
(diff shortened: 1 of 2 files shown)


joshblack:
Hi there! 👋 This package is used as a part of the react compiler transformation here: https://github.com/primer/react/blob/main/packages/react/babel.config.js#L10 from what I understand.

Hope that makes sense! Let me know if you have any questions.

Uzlopak:
@joshblack 

Thats odd. @primer/react uses react 18, so it needs this shim. But does it need it to be a production dependency?

joshblack:
I think so since it's included in our generated files, e.g.: https://app.unpkg.com/@primer/react@37.30.0/files/lib-esm/Skeleton/SkeletonBox.js

Uzlopak:
Thank you for this perspective. Closing this PR ;)

PR #6561, opened 2025-08-11
Breadcrumbs : Add overflow menu for responsive behavior

pksjce:
Closes #https://github.com/github/primer/issues/5122

### Changelog

Breadcrumbs now show an overflow menu with the following features
- If no of items in the breadcrumbs is >5 then we show the menu
- If the screen width can't accomodate the items then we show the menu
- if `hideRoot` is false then we show the root/overflow menu/leaf 

https://www.loom.com/share/8902798416b741c9959fc7bb501185a5

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,29 +1,255 @@
...
+          variant="invisible"
+          size="small"
+          trailingAction={null}
+          style={{display: 'inline-flex'}}
+          {...rest}
+        >
+          …
+        </ActionMenu.Button>
+        <ActionMenu.Overlay width="auto">
+          <ActionList role="menu">
```
joshblack, line 52: Just want to confirm the semantics for this since I wasn't sure, should we use something more like: https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/ ? Having this as a `role="menu"` feels off to me based on stuff like:

- https://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html
- `https://github.com/w3c/aria-practices/issues/353`
- `https://github.com/openui/open-ui/issues/1193`
- `https://github.com/openui/open-ui/issues/1193#issuecomment-2811377928`

(Using code instead of links so that this PR doesn't get linked to those issues 😅)

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,29 +1,255 @@
...
+            {items.map((item, index) => {
+              const href = item.props.href
+              const children = item.props.children
+              const selected = item.props.selected
```
joshblack, line 56: Since we're reading the props from the children, does it mean this wouldn't work for wrapped components? e.g. 

```tsx
<Breadcrumbs>
  <WrappedBreadcrumbItem>Example</WrappedBreadcrumbItem>
</Breadcrumbs>
```

This can come up sometimes in projects like Next.js where you want to wrap the component from Primer so you don't have to repeat stuff like `<Breadcrumb.Item as={NextLink}>` so the links participate in the router

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,29 +1,255 @@
...
+
+      const result = calculateOverflow(containerWidth)
+      setVisibleItems(result.visibleItems)
+      setMenuItems(result.menuItems)
+      setEffectiveHideRoot(result.effectiveHideRoot)
+    }
+  }, [childArray, overflow, containerWidth, hideRoot, itemWidths])
+
+  // Determine final children to render
+  const finalChildren = React.useMemo(() => {
```
joshblack, line 222: Since we're using `visibleItems` here, and that's only set with `useEffect()`, I think that this component will end up not rendering anything when SSR'd. Here's a quick video from the Next.js example when I was looking into this:


https://github.com/user-attachments/assets/5986361e-1d6e-4731-9916-d37142cbfd1e

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,29 +1,255 @@
...
+    if (overflow === 'wrap' || menuItems.length === 0) {
+      return visibleItems.map(child => (
+        <li className={classes.ItemWrapper} key={child.key}>
```
joshblack: One thing that came to mind since we're using `Children.toArray()` is that the keys could act weird based on: https://react.dev/reference/react/Children#children-toarray:~:text=The%20returned%20elements%E2%80%99%20keys%20will%20be%20calculated%20from%20the%20original%20elements%E2%80%99%20keys%20and%20their%20level%20of%20nesting%20and%20position, specifically that they may not be consistent based on that note.

One thing that could be related is that the items in the breadcrumbs seem to retain focus on resize but not the items in the menu for some reason 🤔 (even with emulate keyboard turned on)

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,29 +1,255 @@
...
+      if (overflow === 'wrap') {
+        setVisibleItems(childArray)
+        setMenuItems([])
+        setEffectiveHideRoot(hideRoot)
+        return
+      }
+
+      // For 'menu' overflow mode
+      // Helper function to calculate visible items and menu items with progressive hiding
+      const calculateOverflow = (availableWidth: number) => {
```
joshblack: One odd thing I noticed is that as items change it seems like the widths might be cached and so the layout ends up being incorrect

https://github.com/user-attachments/assets/d82a2686-f8dd-4f06-8a78-242edd49ae11

Would it be helpful to move more of this logic directly into the resize observer so that more values are computed instead of cached in state variables? Just an idea to avoid stuff like this.

This kind of stuff seems so rough to do the state <-> effect dance since effects aren't necessarily run right after 😞

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,29 +1,255 @@
...
+        const listElement = containerRef.current?.querySelector('ol')
+        if (listElement && listElement.children.length > 0 && itemWidths.length === 0) {
+          const widths = Array.from(listElement.children).map(child => (child as HTMLElement).offsetWidth)
+          setItemWidths(widths)
+        }
+        const MENU_BUTTON_WIDTH = 50 // Approximate width of "..." button
```
joshblack: Would it be possible to measure the trigger button itself instead of needing to estimate it?

joshblack (commented):
Looking great! Just left a couple comments. Also noticed some behavior stuff:

- It seems like the last item is clipped on the edge (noticed it with the focus outline)
- It seems like the last item content can get clipped when resizing down
- When the menu is at its smallest and the item it truncated, it seems like in the original issue we want to add a tooltip to it, am I understanding that right? If so I wasn't able to bring that up

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Hey this video wont show for me. Cant download it either. I get what you mean by the review. But would be great to look and verify the specific issue in the video where I'm presuming wrong layout comes up

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: I added a story with this and all the wrapper stuff gets stripped off. The conditional stuff too :( How should it look when it gets inserted into the menu? I need to test it properly in the nextjs app.

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: wow! I didnt know about these. I've used role='menu' in UnderlineNav and I was simply going for the same pattern. What do you think it should change to?

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,29 +1,255 @@
...
+type BreadcrumbsMenuItemProps = {
+  items: React.ReactElement[]
+  'aria-label'?: string
+}
+
+const BreadcrumbsMenuItem = React.forwardRef<HTMLButtonElement, BreadcrumbsMenuItemProps>(
+  ({items, 'aria-label': ariaLabel, ...rest}, ref) => {
+    return (
+      <ActionMenu>
+        <ActionMenu.Button
```
langermank: @pksjce could you update this to use the `IconButton` with the `kebab` icon?

pksjce: I am struggling with the vertical align of this button now.
Screenshot 2025-08-15 at 7 31 52 pm

pksjce (commented):
@copilot can you delete the snapshot changes in this PR?

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Will implement as part of accessibility improvements. https://github.com/github/primer/issues/5761

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: I verified that it works with Nextjs links

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Verified that SSR now renders basic breadcrumbs

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Fixed up the keys

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: https://github.com/github/primer/issues/5762

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Done!

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Done!

pksjce:
I've messed up this PR beyond repair. I will cherrypick my commits and open a new one

PR #6599, opened 2025-08-13
feat(mcp): add better primitives output, add coding guidelines tool

joshblack:
Update how we grab primitives from `@primer/primitives` and add a coding guidelines tool to see if it can help projects that don't have guidelines for working with Primer.


package-lock.json
```diff
@@ -30307,7 +30307,7 @@
         "@primer/behaviors": "^1.8.0",
         "@primer/live-region-element": "^0.7.1",
         "@primer/octicons-react": "^19.13.0",
-        "@primer/primitives": "11.1.0",
+        "@primer/primitives": "10.x || 11.x",
```
jonrohan: This change seems suspicious like a bad merge?

joshblack: I'll double check! I'm pretty sure that's the range for `packages/react` so not sure why it's saying that it is currently 11.1.0 🤔 It doesn't seem that way upstream: https://github.com/primer/react/blob/main/package-lock.json#L25382

joshblack: Just pulled from latest @jonrohan and it seems like this is no longer included in the diff

PR #4826, opened 2024-08-07
document "Banner.hideTitle" prop

ktravers:
The Banner "overview" docs mention the ability to hide the title visibly (see https://primer.style/components/banner#title), but how to actually implement this in the React component isn't documented. This change adds the correct prop, `hideTitle`, to the documented props for the Banner React component.

### Changelog

Adds documentation for `Banner.hideTitle` prop.

#### New

Documentation for `Banner.hideTitle` prop.

#### Changed

React `Banner` component documentation.

#### Removed

Nothing.

### Rollout strategy

- [x] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Testing & Reviewing

N/A

### Merge checklist

- [x] Added/updated tests
- [x] Added/updated documentation
- [x] Added/updated previews (Storybook) (already exists: https://primer.style/react/storybook/?path=/story/drafts-components-banner-features--with-hidden-title&globals=colorScheme:light)
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge
- [x] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #6627, opened 2025-08-19
Remove `sx` prop support from `ToggleSwitch`

liuliu-dev:
Migrates the `ToggleSwitch` component away from `Box` components and `sx` prop support to use CSS modules, closes [#4826](https://github.com/github/primer/issues/4826)

## What Changed

- Removed `Box` wrapper and `sx` prop support from the main ToggleSwitch component
- Replaced `sx` prop styling with CSS modules and data attributes for state-based styling
- Added ToggleSwitch to the `styled-react` package for [repos still using `sx` props](https://github.com/github/moda-app-homepage/blob/43c24d664570de1430c56c5fe827d82877ad6863/src/PageComponents/ContainerLogs/Containers/LogsView/LogsView.tsx#L39)

- **Note**: `SwitchButton` and `ToggleKnob` still use styled-components (`styled.button` and `styled.div`) and will be migrated in a future iteration

### Rollout strategy


packages/styled-react/src/__tests__/__snapshots__/exports.test.ts.snap
```diff
@@ -2,6 +2,7 @@
 
 exports[`@primer/styled-react exports 1`] = `
 [
+  "ToggleSwitch",
```
francinelucca, line 5: looks like this was previously following alphabetical order, can we fix back ?

francinelucca (approved):

packages/styled-react/src/__tests__/__snapshots__/exports.test.ts.snap
liuliu-dev: `exports.test.ts.snap` is auto-generated by the test runner.
I guess the reason `ToggleSwitch` appears first is that it's handled differently from other components, it's wrapped with `createStyledComponent` and exported separately.

liuliu-dev:
integration checks manually ran here: https://github.com/github/github-ui/pull/1177

PR #6630, opened 2025-08-20
[WIP] Investigating how we might add test coverage to Primer's MCP

kendallgassner:
## Intro

I am playing around with two different types of testing... 

⚠️ This is all explorative. Please be kind when it comes to my file names (😆). Also, I recognize that this can be written in a more reusable way but I am curious to hear initial feedback.
 
#### What 

 Very basic jest testing. These types of tests can check if the tool was successfully added into the MCP and can check what the MCP returns if a specific tool has been called. 
 
#### What 

This is more comprehensive. This file runs three parallel tests, sending a prompt to OpenAI setup to have access to a local instance of Primer’s MCP Server. For each run, OpenAI's output is compared against an expected result using embeddings and cosine similarity. The test also records which MCP tools were used in each run for analysis.

Pros:
- Realistic and reusable testing  -> this type of testing actually uses AI to show us how and if AI will access our MCP server when prompted with a Primer question. 
- Output comparisons -> using embeddings and cosine similarity means we are not just comparing exact strings but can still check if AI's output is meaningful. 
- tool usage analysis. -> gathering metrics on which MCP tools are used allows us to fine tune tool descriptions and to increase how often our MCP is used. 

Cons: 
- resource intensive -> the more test cases we add the more compute resources we use and the slower the evaluation will run
- embedding limitations -> subtle bugs and logic errors may be missed


packages/mcp/src/__tests__/evalWithOpenAI.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+const GPT_MODEL = 'gpt-5'
+
+export class PrimerMcpClient {
+  private static instance: PrimerMcpClient | null = null
+  private transport: StdioClientTransport
+  private client: Client
+
+  private constructor() {
+    this.transport = new StdioClientTransport({
```
kendallgassner: Loads up the local Primer MCP

packages/mcp/src/__tests__/evalWithOpenAI.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      description: tool.description,
+      parameters: tool.inputSchema,
+    },
+  }))
+
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  const messages: any[] = [{role: 'user', content: prompt}]
+
+  // Initial OpenAI completion
+  const completion = await openai.chat.completions.create({
```
kendallgassner: Feeds OpenAI a prompt and the Primer MCP

packages/mcp/src/__tests__/evalWithOpenAI.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    model: GPT_MODEL,
+    messages,
+    tools: openAITools,
+    // eslint-disable-next-line camelcase
+    tool_choice: 'auto',
+  })
+
+  // Handle tool calls if present
+  const choice = completion.choices[0]
```
kendallgassner: If openAI stopped due to to `tool_calls` (Primer MCP calls). It feeds the MCP calls to OpenAI.

packages/mcp/src/__tests__/evalWithOpenAI.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    aiResponse = finalResponse.choices[0].message.content
+  } else {
+    // Handle non-tool call response
+    aiResponse = choice.message.content
+  }
+
+  return {
+    toolUsage,
+    aiResponse,
```
kendallgassner: Returns the number of times a Primer MCP tool was called and the AIResponse to the prompt.

packages/mcp/src/__tests__/evalWithOpenAI.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    toolUsagePercentages,
+  }
+}
+
+const prompt = `Only output code. Create a Primer Button that will be used to confirm deleting content.`
+const expectedOutput = `<Button variant="danger">Cancel</Button>`
+
+const stats = await getStats(prompt, expectedOutput)
+
+console.log(stats)
```
kendallgassner: Output example:

`meetsSimilarityThreshold` means that on avg the AIResponse was at least `80%` similar to our expected response.

```
{
  meetsSimilarityThreshold: true,
  toolUsagePercentages: [
    { init: '0.00%' },
    { list_components: '0.00%' },
    { get_component: '0.00%' },
    { get_component_examples: '0.00%' },
    { get_component_usage_guidelines: '0.00%' },
    { get_component_accessibility_guidelines: '0.00%' },
    { list_patterns: '0.00%' },
    { get_pattern: '0.00%' },
    { list_tokens: '0.00%' },
    { get_color_usage: '0.00%' },
    { get_typography_usage: '0.00%' },
    { list_icons: '0.00%' },
    { get_icon: '0.00%' }
  ]
}
```

packages/mcp/jest.config.js
joshblack, line 1: Feel free to throw in a `vitest.config.ts` file in here and it should auto pick up the tests based on our project config 👀 I think th e one in `styled-react` should have a node config too that can be used.

kendallgassner: 🤯 One thing I want to discuss in our sync is _if_ we like the idea of "embedding tests" we might want to run them nightly instead of at the CI stage since they are so costly.

kendallgassner:
Closing this issue in favor of tracking data in azure.

At this time I am not convinced that the Embedding-Based AI Evaluation will provide us information that Azure cannot. Additionally, I am weary of adding resource intensive test cases into our CI or workflows.

Moving forward, my focus will be on collecting analytics related to tool usage rather than AI output.

Understanding how tools are used will help us refine their descriptions to improve frequency and relevance in actual usage.

PR #6633, opened 2025-08-21
SelectPanel story to load more items on scroll

siddharthkp:


packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -765,6 +765,7 @@ function Panel({
         onOpen={onOpen}
         onClose={onClose}
         overlayProps={{
+          id: id ? `${id}--dialog` : undefined,
```
cheshire137: Did you need to make this change to SelectPanel? Wondering if you could've passed an `id` from your story in `overlayProps` instead.

siddharthkp: You're right. Can do with just `overlayProps`. Updated!

cheshire137: Nice, thanks! Was wondering because if we wanted to go this route of loading more items on scroll, not having to modify SelectPanel here would mean we could do that today, without needing a primer/react update first.

siddharthkp: perfect! there is however another bug here :( 

Scroll resets to top when the items change. Works well when `primer_react_select_panel_remove_active_descendant` feature flag is on. I've asked what's the rollout plan [on slack](https://github.slack.com/archives/C087S9D3W87/p1755848768159079)

pksjce: You should assume that FF to be on for this use case

packages/react/src/SelectPanel/SelectPanel.examples.stories.tsx
```diff
@@ -474,3 +476,115 @@ export const WithDefaultMessage = () => {
...
+        // has scrolled to bottom
+        setNumberOfItemsInSubset(numberOfItemsInSubset + 50)
+      }
+    }
+
+    // eslint-disable-next-line github/prefer-observers
+    if (scrollContainer) scrollContainer.addEventListener('scroll', handler)
+    return () => scrollContainer?.removeEventListener('scroll', handler)
+  })
+
```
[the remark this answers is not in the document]
siddharthkp: No thanks 😇

packages/react/src/SelectPanel/SelectPanel.examples.stories.tsx
```diff
@@ -474,3 +476,115 @@ export const WithDefaultMessage = () => {
...
+  const [renderSubset, setRenderSubset] = React.useState(true)
+
+  const [filter, setFilter] = useState('')
+  const filteredItems = lotsOfItems.filter(item => item.text.toLowerCase().startsWith(filter.toLowerCase()))
+
+  const [numberOfItemsInSubset, setNumberOfItemsInSubset] = useState(50)
+  const subsetOfFiltereredItemsToRender = filteredItems.slice(0, renderSubset ? numberOfItemsInSubset : NUMBER_OF_ITEMS)
+
+  useEffect(function loadMoreItemsOnScrollEnd() {
+    const scrollContainer = document.querySelector('#select-labels-panel-dialog [role="listbox"]')?.parentElement
```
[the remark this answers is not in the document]
siddharthkp: Yeah, but you can't really pass a ref to an inner container so this is one possible way

pksjce (approved):

PR #6641, opened 2025-08-21
Derive component prop meta-data from TypeScript

adierkens:
This change builds upon some of the foundations from https://github.com/primer/react/pull/6047 to enable us to progressively move towards more colocated documentation for our prop types inline with the TS source-code. 

As a first-step, the list of properties to expose to documentation will still remain in the `*.docs.json` file, but each now have the option of using the `"derive": true` flag to gather any pertinent information from the TS types, including `description` (JSDoc), `defaultValue`, `type`, and `required`. Fields specified in the `*.docs.json` will override any from TS, allowing for customization as more is moved into each component. 

This reduces the amount of overhead for maintaining the separate `*.docs.json` per Component and also has the added benefit of exposing better documentation for users within their IDEs (which already pull from JSDoc comments and tags like `@default`)

---

## Context

Documentation for https://primer.style/ is generated via summaries defined in `*.docs.json` files for components (e.g. [Button.docs.json](packages/react/src/Button/Button.docs.json)). This package facilitates ensuring TypeScript typings & the documentation site don't diverge.

### Generation

As part of the [build](packages/react/script/components-json/build.ts) script for compiling the `docs.json` summaries, this `doc-gen` module is run as a pre-processor for the content. Any props marked as `derived` within their `Component.docs.json` will have their meta-data pulled & merged from it's TS source and validated using the existing JSONSchema. Errors during this phase will fail the build.

### Validation

In order to validate that the properties taken from TypeScript and the authored `.docs.json` are aligned, there is a `validate` command to run:

```bash
npx doc-gen validate ./packages/react/src/Button
```

This will display a table summary of all props found in `.docs.json` and ensure they are also present in TypeScript.

#### Deriving Props from TS

In order to derive a prop for a component from it's TS source, use the `"derive": true` flag within the given property. The `name` is the only other required property in order to map the appropriate property. Any property marked as _derived_ that are not found within the TS source will throw an error.

<table>
<tr><td>Before</td><td>After</td></tr>
<tr>
<td>

In the component's `*.docs.json`

```json
{
  "props": [
    {
      "name": "alignContent",
      "type": "'start' | 'center'",
      "default": "start",
      "description": "Content alignment for when visuals are present.",
      "required": false
    }
  ]
}
```

</td>

<td>

In the component's `*.docs.json`

```json
{
  "props": [
    {
      "name": "alignContent",
      "derive": true
    }
  ]
}
```

In the component's `*.tsx`

```tsx
interface Props {
  /**
   * Content alignment for when visuals are present.
   *
   * @default 'start'
   */
  alignContent?: 'start' | 'center'
}
```

</td>

</tr>
</table>

To override any value of a derived property, set the value within the `*.docs.json` as usual:

```json
{
  "props": [
    {
      "name": "alignContent",
      "description": "Override the default value",
      "derive": true
    }
  ]
}
```

### Auto-Fix

To make the transition easier for the migration, a `--fix` option is available:

```bash
npx doc-gen validate ./packages/react/src/Button --fix
```

For each property in the `*.docs.json`, it will:

- Write any `description`, `required`, and `default` values for that property into the TS source.
- Mark the `prop` in the `.docs.json` as _derived_
- Remove any attributes from the `.docs.json` that duplicate the values from the TS source.


packages/react/src/ActionBar/ActionBar.docs.json
```diff
@@ -34,16 +30,11 @@
     {
       "name": "size",
       "type": "'small' | 'large' | 'medium'",
-      "required": false,
-      "description": "Size of the action bar",
-      "defaultValue": ""
+      "derive": true
     },
```
adierkens, line 34: Example result of the derived content: https://app.unpkg.com/@primer/react@0.0.0-20250930171152/files/generated/components.json#L38-42

```json
        {
          "name": "size",
          "type": "'small' | 'large' | 'medium'",
          "required": false,
          "defaultValue": "'medium'",
          "description": "Size of the action bar"
        },
```

.github/workflows/migration-status.yml
```diff
@@ -34,3 +34,18 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'npm'
+      - name: Install dependencies
+        run: npm ci
+      - name: Run migration script
+        run: |
+          node --experimental-strip-types script/ts-docs-status.mts >> $GITHUB_STEP_SUMMARY
```
adierkens, line 51: Example status: https://gist.github.com/adierkens/320f4fce1cd1f3c42b3b7e0ee41ffea7#file-component-status-md

jonrohan (approved):
A lot to verify. I like the approach and I checked major entry points for the new package. Things look good enough to me that we could iterate if necessary 👍🏻

pksjce (approved):

PR #6664, opened 2025-08-25
Breadcrumbs : Add overflow menu for responsive behavior

pksjce:
Closes https://github.com/github/primer/issues/5122

### Changelog

Breadcrumbs now show an overflow menu with the following features
- If no of items in the breadcrumbs is >5 then we show the menu
- If the screen width can't accomodate the items then we show the menu
- if `overflow==="menu-with-root"` show the root item.

https://www.loom.com/share/8902798416b741c9959fc7bb501185a5

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


strackoverflow (approved):
This is looking great!

One edge case issue I uncovered (not necessarily blocking), is that when I have the root item shown and resize the window to a narrower viewport, at certain widths, the component seems to struggle with the root item getting hidden (which may be triggering another resize event), creating an infinite loop of hiding and showing the root crumb:

https://github.com/user-attachments/assets/35d989f3-68b1-47ed-a679-9830a06169ff

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+
+  useEffect(() => {
+    const listElement = containerRef.current?.querySelector('ol')
+    if (listElement && listElement.children.length > 0) {
+      const listElementArray = Array.from(listElement.children) as HTMLElement[]
+      const widths = listElementArray.map(child => child.offsetWidth)
+      setChildArrayWidths(widths)
+      setRootItemWidth(listElementArray[0].offsetWidth)
+    }
+  }, [childArray.length])
```
joshblack: I don't think this will synchronize correctly if it depends on the length of the items instead of the items themselves changing.

Put another way, if the items are updated to be different items but of the same length then this code won't re-run and set up the state values we have for item and root width.

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+            aria-expanded="false"
+            variant="invisible"
+            size="small"
+            icon={KebabHorizontalIcon}
+            className={classes.MenuButton}
+            {...rest}
+          />
+        </ActionMenu.Anchor>
+        <ActionMenu.Overlay width="auto">
+          <ActionList role="menu">
```
francinelucca: does this have the a11y enhancements in? if not, is the plan to follow up this PR with those enhancements? if so:
- I think it makes sense for this PR *not* to close https://github.com/github/primer/issues/5122
- would it be a breaking change to get this in and then switch the underlying html structure? 🤔

.changeset/good-cougars-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Spike for breadcrumbs overflow
```
francinelucca: I think this needs to be updated as it'll show up in the changelog

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,296 @@
...
+  const [menuItems, setMenuItems] = useState<React.ReactElement[]>([])
+  const [rootItemWidth, setRootItemWidth] = useState<number>(0)
+
+  // Menu button width with fallback
+  const MENU_BUTTON_FALLBACK_WIDTH = 32 // Design system small IconButton
+  const [menuButtonWidth, setMenuButtonWidth] = useState(MENU_BUTTON_FALLBACK_WIDTH)
+
+  // if (typeof window !== 'undefined') {
+  //   effectiveOverflow = overflow
+  // }
```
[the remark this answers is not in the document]
francinelucca: 👀

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+      return React.Children.map(children, child => <li className={classes.ItemWrapper}>{child}</li>)
+    }
+
+    let effectiveMenuItems = [...menuItems]
+    if (!effectiveHideRoot) {
+      effectiveMenuItems = [...menuItems.slice(1)]
+    }
+    const menuElement = (
+      <li className={classes.BreadcrumbsItem} key="breadcrumbs-menu">
+        <BreadcrumbsMenuItem
```
francinelucca: from https://github.com/github/primer/issues/5122
> Next Steps
Merge Breadcrumbs as currently implemented under a feature flag

But I don't see a FF here 👀

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+
+    // Position menu based on effective hideRoot setting and visible items
+    if (effectiveHideRoot) {
+      // Show: [overflow menu, leaf breadcrumb]
+      return [menuElement, ...visibleElements]
+    } else {
+      // Show: [root breadcrumb, overflow menu, leaf breadcrumb]
+      return [rootElement, menuElement, ...visibleElements]
+    }
+  }, [overflow, menuItems, effectiveHideRoot, visibleItems, rootItem, children])
```
joshblack: If I'm understanding right, I think on every resize this value will be memoized is that right? Even if the items aren't changing. Would it make sense in `handleResize` to compare the two lists (visible items, menu items) to see if they have changed before updating?

I'm also curious if, with React compiler, we need to use memo for this 🤔 Curious what you've run into for that 👀

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
-  const wrappedChildren = React.Children.map(children, child => <li className={classes.ItemWrapper}>{child}</li>)
+type BreadcrumbsMenuItemProps = {
+  items: React.ReactElement[]
+  'aria-label'?: string
+}
+
+const BreadcrumbsMenuItem = React.forwardRef<HTMLButtonElement, BreadcrumbsMenuItemProps>(
+  ({items, 'aria-label': ariaLabel, ...rest}, ref) => {
+    return (
+      <ActionMenu>
```
joshblack: I think my personal preference would be to have this use the disclosure pattern before getting this in since using a menu for navigation is not preferred.

I assume with this work that if we went down that route we would have to do more from scratch since that's not supported in `ActionMenu` at the moment?

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+import {useResizeObserver} from '../hooks/useResizeObserver'
+import type {ResizeObserverEntry} from '../hooks/useResizeObserver'
 
 const SELECTED_CLASS = 'selected'
 
 export type BreadcrumbsProps = React.PropsWithChildren<
   {
     className?: string
+    overflow?: 'wrap' | 'menu'
+    hideRoot?: boolean
```
joshblack: Definitely not sure what the right names would be, but what would you think of something like:

```suggestion
    overflow?: 'wrap' | 'hide' | 'hide-root'
```

So that we wouldn't have the case of `hideRoot` only making sense with `overflow="menu"`. It would also avoid the confusion here with it not being a `menu` since this would be a popup with a list of links versus `role="menu"`

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+
+  // Measure actual menu button width when it exists
+  useEffect(() => {
+    if (menuButtonRef.current) {
+      const measuredWidth = menuButtonRef.current.offsetWidth
+      if (measuredWidth > 0) {
+        setMenuButtonWidth(measuredWidth)
+      }
+    }
+  }, [menuItems.length]) // Re-measure when menu button appears/disappears
```
joshblack: One idea for measuring the menu button could be to measure it in a ref callback, curious if that would make sense or not 🤔 Would avoid the need for us to have an effect for it. Hopefully this doesn't violate the rules of hooks too 😅 lol

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+        if (currentMenuItems.length > 0) {
+          visibleItemsWidthTotal += MENU_BUTTON_WIDTH
+        }
+        while (
+          overflow === 'menu' &&
+          (visibleItemsWidthTotal > availableWidth || currentVisibleItems.length > MIN_VISIBLE_ITEMS)
+        ) {
+          // Remove the last visible item
+          const itemToHide = currentVisibleItems.slice(0)[0]
+          const itemToHideWidth = currentVisibleItemWidths.slice(0)[0]
```
joshblack: What's the call to `.slice(0)[0]` for here? Is it to do a deep copy on the item in the array or or something else?

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+          currentMenuItems = [...currentMenuItems, itemToHide]
+          currentMenuItemsWidths = [...currentMenuItemsWidths, itemToHideWidth]
+          currentVisibleItems = [...currentVisibleItems.slice(1)]
+          currentVisibleItemWidths = [...currentVisibleItemWidths.slice(1)]
```
joshblack: One idea, I think `.slice(1)` would create a new array here so we could avoid spreading it into a new array

```suggestion
          currentVisibleItems = currentVisibleItems.slice(1)
          currentVisibleItemWidths = currentVisibleItemWidths.slice(1)
```

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+            eHideRoot = true
+            break
+          } else {
+            eHideRoot = hideRoot
+          }
+        }
+      }
+      return {
+        visibleItems: [...currentVisibleItems],
+        menuItems: [...currentMenuItems],
```
joshblack: One idea with the spread:

```suggestion
        visibleItems: currentVisibleItems,
        menuItems: currentMenuItems,
```

Let me know if these spread suggestions aren't applicable here btw!

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,292 @@
...
+    let effectiveMenuItems = [...menuItems]
+    if (!effectiveHideRoot) {
+      effectiveMenuItems = [...menuItems.slice(1)]
+    }
+    const menuElement = (
+      <li className={classes.BreadcrumbsItem} key="breadcrumbs-menu">
+        <BreadcrumbsMenuItem
+          ref={menuButtonRef}
+          items={effectiveMenuItems}
+          aria-label={`${effectiveMenuItems.length} more breadcrumb items`}
```
joshblack: This could be a good use for the `…` character specifically on toggle element (got the idea from spectrum: https://react-spectrum.adobe.com/react-spectrum/Breadcrumbs.html) and that might mean we don't need a tooltip on this

joshblack (commented):
Looking great! Just left some comments/questions/ideas.

One odd thing I noticed when working with the component is that when I saved the breadcrumb file, the layout would break 🤔 Curious if you ran into that too

https://github.com/user-attachments/assets/67a49831-db9c-4b9d-9c2f-570f1db5ef39

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: I know! I debated it but I decided since its only addition of a prop and only that triggers the new feature, we don't need it. WDYT?

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: I created subissues on that issue to tackle the ally issues!

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Yes, We need to use a pattern involving `Details`. I tried it once on Friday and it was not straightforward. Will give it another go, now that I have a clean slate

.changeset/good-cougars-hug.md
pksjce: 🤔 I thought i did. I think its fine now. I think something with my cherry-pick messed it

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: I remember Katie specifically mentioned that its good to use the iconbutton for the toggle element. Is that what you mean here? Also the tooltip is the same as they used in context-region

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Oh this may be an improvement, but do you think it makes the api more convoluted? `menu` is what it practically looks like. `hide` and `hide-root` as values dont semantically make sense in this context

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
francinelucca: yeah I just mean feels weird to close the parent issue if there's still open subissues

.changeset/good-cougars-hug.md
francinelucca: Now there's two 😅 , can you delete this one?

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
francinelucca: I think the good thing about a FF is we can get away with changing more things within it to refine stuff, but if you don't foresee making any major changes (outside the pre-release window), then a prop is fine 👍

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Yes yes, I won't be closing the parent issue. It's just an old description. I just want it linked to the issue

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: I'm still undecided. There is definite change in the html structure and css to the original. I want to have a look at the review-lab here and decide

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
siddharthkp: Am I correct in saying that hideRoot would only be relevant with overflow: `menu`? If yes, how do you feel about `overflow: 'menu' | 'menu-with-hidden-root'`? A little long but keeping related options together?

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Not really, I was slicing a larger part in a prev implementation and ended up with this staler code.

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: I think this is resolved too.

pksjce:
**Summary of changes since last review**

- @strackoverflow - Nice catch on the looping behavior. This is now resolved. 
- @francinelucca 
   - I have added the ally recommendations.
   - Also added the feature flag - primer_react_breadcrumbs_overflow_menu
   - Removed the extra changeset file
- @joshblack 
   - I have fixed the issue of wrong update on saving the file
   - Worked on the memoized resize calculation
   - Used `Details` to implement disclosure pattern
   - Changed the `overflow` property to have values `menu` and `menu-with-root` eliminating `hideRoot` boolean. cc- Thanks @siddharthkp 
   - Removed unnecessary `slice`
  
  
  **Notes on Accessibility improvements**
- `ActionMenu` has been replaced by an implementation using `Details` and `Details.Summary`. 
- `aria-expanded` property moves from `false` to `true` as required.
- There are no more `menuitem` roles. All are `li` tags with the inbuilt role
- The tooltip has been positioned on the right.
- Esc and outside click will work to close the menu
- Tabbing is continous inside and outside the menu without any focus traps.

**Advice needed**
- Tabbing into the menubutton will not open it automatically nor Tabbing out will close it automatically. This was recommended in the session but did not seem standard to disclosure implementations eg UnderlineNav.

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+}
+
+.MenuDetails summary::-webkit-details-marker {
+  display: none;
+}
+
+.MenuOverlay {
+  position: absolute;
+  z-index: 1;
+  box-shadow: var(--shadow-resting-medium);
```
langermank:
```suggestion
  box-shadow: var(--shadow-resting-small);
```

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,356 @@
...
+
+    useOnOutsideClick({
+      onClickOutside: closeOverlay,
+      containerRef: menuContainerRef,
+      ignoreClickRefs: [iconButtonRef],
+    })
+
+    return (
+      <Details ref={detailsRefCallback} className={classes.MenuDetails}>
+        <Details.Summary className={classes.MenuSummary} tabIndex={-1}>
```
langermank: I wonder if you could actually give the `IconButton` an `as="summary"` so that all the button logic just gets used as the summary element? Rather than wrapping the button in a summary.

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+  cursor: pointer;
+  outline: none;
+}
+
+.MenuDetails summary::-webkit-details-marker {
+  display: none;
+}
+
+.MenuOverlay {
+  position: absolute;
```
langermank: We need to have an offset between the trigger and the menu, but I'm not sure how you want to implement that with this custom overlay. The variable for this is `var(--overlay-offset)` which is 4px, so you might need to use a `calc` to achieve that.

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+}
+
+.BreadcrumbsItem {
+  display: inline-grid;
+  grid-auto-flow: column;
+  align-items: center;
+  flex: 0 99999 auto;
+  min-width: auto;
+  font-size: var(--text-body-size-medium);
+  white-space: nowrap;
```
langermank: This `white-space: nowrap` is bleeding into the items in the overflow menu from this declaration, but also from another spot in here (I think). Something nested.. but anyway, we don't want this on the items inside the menu. They should be able to wrap line.

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+    outline: 2px solid var(--focus-outlineColor, var(--color-accent-fg));
+    outline-offset: -2px;
+  }
+}
+
+.MenuSummary {
+  list-style: none;
+  display: inline-block;
+  outline: none;
+  height: 0.5rem;
```
langermank: I think if you're able to use the `as` in IconButton a lot of this (maybe all?) can be removed

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+}
+
+.MenuOverlay {
+  position: absolute;
+  z-index: 1;
+  box-shadow: var(--shadow-resting-medium);
+  border-radius: var(--borderRadius-large);
+  background-color: var(--overlay-bgColor);
+  list-style: none;
+  width: max-content;
```
langermank:
```suggestion
  min-width: var(--overlay-width-xsmall);
  max-height: 100vh;
  max-width: var(--overlay-width-small);
  overflow: hidden;
```

Just aligning with Overlay

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,356 @@
...
+  ) : (
     <BoxWithFallback as="nav" className={clsx(className, classes.BreadcrumbsBase)} aria-label="Breadcrumbs" sx={sxProp}>
       <BreadcrumbsList>{wrappedChildren}</BreadcrumbsList>
     </BoxWithFallback>
   )
 }
 
+const ItemSeparator = () => {
+  return (
+    <span className={classes.ItemSeparator}>
```
langermank, line 364: You might need some labeling on this for a11y support (just a hunch, would look it up)

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+}
+
```
langermank:
```suggestion
@media (prefers-reduced-motion: no-preference) {
  .MenuOverlay {
    animation: overlay-appear 200ms cubic-bezier(0.33, 1, 0.68, 1);
  }
}
```

Messy but you can include this somewhere

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+      display: none;
+    }
+  }
+}
+
+[data-overflow='menu'] .Item,
+[data-overflow='menu-with-root'] .Item {
+  display: inline-block;
+  font-size: var(--text-body-size-medium);
+  color: var(--fgColor-link);
```
langermank:
```suggestion
  color: var(--fgColor-default);
```

I wasn't sure if we should include this styling change, but since we're adding the hover state we might as well.. it seems like the teams building their custom breadcrumb experiences are using the default text color anyway or overriding our link color.

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+  text-decoration: none;
+  padding-inline: var(--base-size-6);
+  padding-block: var(--base-size-4);
+  border-radius: var(--borderRadius-medium);
+
+  &:hover {
+    background: var(--control-transparent-bgColor-hover);
```
langermank:
```suggestion
    background: var(--control-transparent-bgColor-hover);
    text-decoration: none;
```

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -9,6 +10,100 @@
...
+  }
+
+  &:focus {
+    box-shadow: none;
+    outline: 2px solid var(--focus-outlineColor, var(--color-accent-fg));
+    outline-offset: -2px;
+  }
```
langermank:
```suggestion
  &:focus-visible {
    @mixin focusOutline;
  }
```

We have a mixin now!

langermank (commented):
This is purely a CSS review! Will take a closer look at behavior next.

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: oh! that works thanks! I can remove that `tabIndex` now

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: It's not focusable. I'm not sure if the label is required. What would you expect in the label?
cc - @janmaarten-a11y

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,358 @@
...
+import {useOnEscapePress} from '../hooks/useOnEscapePress'
+import {useOnOutsideClick} from '../hooks/useOnOutsideClick'
+import {useFeatureFlag} from '../FeatureFlags'
 
 const SELECTED_CLASS = 'selected'
 
 export type BreadcrumbsProps = React.PropsWithChildren<
   {
     className?: string
+    overflow?: 'wrap' | 'menu' | 'menu-with-root'
```
adierkens, line 32: 💅 thanks for adding the descriptions into the docs file. We should also consider adding it to here as a jsdoc tag

packages/react/src/Breadcrumbs/Breadcrumbs.features.stories.tsx
```diff
@@ -0,0 +1,207 @@
...
+            border: '1px solid var(--borderColor-accent-emphasis)',
+          },
+        }}
+        aria-label="Edit document name"
+      />
+    </Breadcrumbs.Item>
+  </Breadcrumbs>
+)
+
+export const DynamicChildren = () => {
```
joshblack: One potential thing we could do with this story is to put it into dev stories (`Breadcrumbs.dev.stories.tsx`) and that way we can see it but it won't show up in our production storybook

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,33 +1,358 @@
...
+        // Prevent the button click from bubbling up and interfering with details toggle
+        event.preventDefault()
+        // Manually toggle the details element
+        if (detailsRef.current) {
+          const newOpenState = !detailsRef.current.open
+          detailsRef.current.open = newOpenState
+          setIsOpen(newOpenState)
+        }
+      },
+      [detailsRef],
```
joshblack:
```suggestion
      [],
```

I think since this comes from `useRef` the rules of hooks eslint rule shouldn't flag this 👀

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
janmaarten-a11y: If this is just the `/` separator, it shouldn't need any kind of announcement. The distinction between them is handled by the list semantics themselves.

packages/react/src/Breadcrumbs/Breadcrumbs.features.stories.tsx
pksjce: Yes this has been done

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
pksjce: Ah you're right! thanks!

pksjce:
@joshblack - I have created this issue for the Esc button bug. It emerged because I made the `IconButton` of as `summary`.
Thanks for catching it. Will fix it separately

strackoverflow (approved):
Looks amazing!

PR #6666, opened 2025-08-25
Remove styled-system usage - TextInputWrapper and related

mperrotti:
Relates to https://github.com/github/primer/issues/5422

#### Removed

> [!NOTE]
> **Reviewers:** I misunderstood the scope of https://github.com/github/primer/issues/5422 when I started on it and also removed `sx` props. If we want to preserve those for now, I can revert those changes.

- `sx` prop from
  - `Select`
  - `Textarea`
  - `TextInput`
  - `TextInputWithTokens`
  - `TextInputWrapper`
- `width`, `minWidth`, and `maxWidth` props from `TextInputWrapper`
- `styled-system` dependencies

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Breadcrumbs/Breadcrumbs.dev.stories.module.css
+++ b/packages/react/src/Breadcrumbs/Breadcrumbs.dev.stories.module.css
@@ -1,0 +1,10 @@
+.EditableTextInput {
+  min-width: 120px;
+  max-width: 180px;
+  font-size: inherit;
+  border: var(--borderWidth-thin) dashed var(--borderColor-muted);
+}
+
+.EditableTextInput:focus {
+  border: var(--borderWidth-thin) solid var(--borderColor-accent-emphasis);
+}
--- a/packages/react/src/Breadcrumbs/Breadcrumbs.dev.stories.tsx
+++ b/packages/react/src/Breadcrumbs/Breadcrumbs.dev.stories.tsx
@@ -1,6 +1,7 @@
 import {useState} from 'react'
 import Breadcrumbs from '.'
 import TextInput from '../TextInput'
+import classes from './Breadcrumbs.dev.stories.module.css'
 
 export default {
   title: 'Components/Breadcrumbs/Dev',
@@ -146,15 +147,7 @@ export const WithEditableNameInput = () => (
       <TextInput
         defaultValue="Untitled Document"
         size="small"
-        sx={{
-          minWidth: '120px',
-          maxWidth: '180px',
-          fontSize: 'inherit',
-          border: '1px dashed var(--borderColor-muted)',
-          '&:focus': {
-            border: '1px solid var(--borderColor-accent-emphasis)',
-          },
-        }}
+        className={classes.EditableTextInput}
         aria-label="Edit document name"
       />
     </Breadcrumbs.Item>


packages/react/src/Textarea/Textarea.tsx
```diff
@@ -58,7 +56,6 @@ const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
     {
       value,
       disabled,
-      sx: sxProp,
```
francinelucca, line 61: I see [usage of this](https://primer-query.githubapp.com/?query=name%3A%22Textarea%22+attribute%3A%22sx%22) in dotcom, are we planning to migrate?

packages/react/src/internal/components/TextInputWrapper.tsx
```diff
@@ -51,9 +42,6 @@ export const TextInputBaseWrapper = React.forwardRef<HTMLElement, StyledTextInpu
       contrast,
       monospace,
       block,
-      width,
-      minWidth,
-      maxWidth,
```
jonrohan, line 56: should we have left these props? They're getting added to the style attribute.

packages/react/src/Overlay/Overlay.features.stories.tsx
```diff
@@ -387,17 +390,17 @@ export const NestedOverlays = ({role, open}: Args) => {
               aria-label={role === 'dialog' ? 'Create a list' : undefined}
               ref={secondaryContainer}
             >
-              <Box as="form" sx={{display: 'flex', flexDirection: 'column', p: 3}} aria-label="Set Duration Form">
-                <Text color="fg.muted" sx={{fontSize: 1, mb: 3}}>
+              <Stack as="form" gap="condensed" padding="normal" aria-label="Set Duration Form">
+                <Text color="fg.muted" size="medium">
```
jonrohan: Can you convert the color here?

packages/react/src/internal/components/TextInputWrapper.tsx
mperrotti: We totally could, but they're deprecated anyway. A major release seemed like a good opportunity to get rid of them.

Happy to add them back if you or other reviewers feel strongly.

packages/react/src/Textarea/Textarea.tsx
mperrotti: I assume we would, but we're also planning to get rid of `sx` props in this upcoming release. Is there a plan for how to handle this? I know we have `react-styled`, but I don't _really_ know what it is or how to use it.

francinelucca: ah yes I think react-styled is the way to go here. @joshblack would it be part of  https://github.com/github/primer/issues/5422 as well to go in dotcom and replace the textarea + sx usage with react-styled instead or what's the pipeline here?

joshblack: Exactly, yeah. I would make sure in github-ui to update usage there first (the migration in this case is just find the places that use styled-system for these components and update the imports to pull the component from `@primer/styled-react`) and then merge here so that we don't have to deal with any challenges in the release.

Hope that makes sense, let me know if it'd be helpful to walk through this or do a loom or something since this is a new process 👀

PR #6667, opened 2025-08-25
Remove `sx` props and `BoxWithFallback` from `TooltipV2` and `Tooltip`

liuliu-dev:
Closes [#5769](https://github.com/github/primer/issues/5769)

#### Removed

Removed `sx` props from deprecated `Tooltip` and `TooltipV2`

### Rollout strategy

### DIFF
--- a/.changeset/pink-rockets-win.md
+++ b/.changeset/pink-rockets-win.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': major
+---
+
+Update Tooltip component to no longer support sx.
--- a/packages/react/src/Tooltip/Tooltip.docs.json
+++ b/packages/react/src/Tooltip/Tooltip.docs.json
@@ -31,11 +31,6 @@
       "name": "wrap",
       "type": "boolean",
       "description": "Use `true` to allow text within tooltip to wrap."
-    },
-    {
-      "name": "sx",
-      "type": "SystemStyleObject",
-      "deprecated": true
     }
   ],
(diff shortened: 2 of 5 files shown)


packages/styled-react/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return <Box as={PrimerTooltip} ref={ref} {...props} />
+})
+
+export {Tooltip, type TooltipProps}
+
+type DeprecatedTooltipProps = PrimerDeprecatedTooltipProps & SxProp
+
+function DeprecatedTooltip(props: DeprecatedTooltipProps) {
+  return <Box as={PrimerDeprecatedTooltip} {...props} />
+}
```
francinelucca, line 24: should we do the new `as` hack just in case? I think in theory it affects every component

francinelucca (approved):
✨

packages/styled-react/src/components/Tooltip.tsx
liuliu-dev: I think this only affects `polymorphic` components that take an `as` prop, `Tooltip` isn’t one of them, so it should be fine.
I tested it in `primer-react.browser.test.tsx` and confirmed that the props and children are passed correctly for `Tooltip`.

francinelucca: ah got it, thanks! ❤️

PR #6668, opened 2025-08-25
Remove support for `sx` and `Box` usage from the `UnderlineNav` component

liuliu-dev:
Closes [#5770](https://github.com/github/primer/issues/5770)

#### Removed

Removed `sx` props and `Box` from `UnderlineNav`.

### Rollout strategy

 
### DIFF
--- a/packages/react/src/UnderlineNav/UnderlineNav.tsx
+++ b/packages/react/src/UnderlineNav/UnderlineNav.tsx
@@ -1,8 +1,5 @@
 import type {MutableRefObject, RefObject} from 'react'
 import React, {useRef, forwardRef, useCallback, useState, useEffect} from 'react'
-import Box from '../Box'
-import type {SxProp} from '../sx'
-import sx from '../sx'
 import {UnderlineNavContext} from './UnderlineNavContext'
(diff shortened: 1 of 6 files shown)


packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -45,10 +41,8 @@ export const MORE_BTN_WIDTH = 86
 // The height is needed to make sure we don't have a layout shift when the more button is the only item in the nav.
 const MORE_BTN_HEIGHT = 45
 
-// Needed this because passing a ref using HTMLULListElement to `Box` causes a type error
-export const NavigationList = styled.ul`
-  ${sx};
-`
+// Needed for typing the ref with HTMLULListElement
+export const NavigationList = styled.ul``
```
joshblack: Super random, do we need this export? I couldn't find where it is used 🤔

joshblack (approved):
Looks great! just left one question

packages/react/src/UnderlineNav/UnderlineNav.tsx
liuliu-dev: Removed 🧹

PR #6674, opened 2025-08-26
Remove `CircleBadge` component

TylerJDev:
Closes https://github.com/github/primer/issues/5535

Removes `CircleBadge` from PRC.

https://github.com/github/github-ui/pull/1113 needs to be merged before this PR.

#### Removed

* Removes `CircleOcticon` component

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp:
Hi 👋 

Should we move this to deprecated in case someone outside of github/github-ui is using it?

TylerJDev:
@siddharthkp, I think we've just been removing them outright from PRC and skipping the deprecation step, at least for the smaller components with low usage. We could include a migration guide, which would advise consumers to utilize the styles we used for the component, but I'm not sure if that'd be helpful or not 🤔 Let me know what you think!

siddharthkp:
(I feel 50% strongly about this, not 100%)

Looking at [primer-query](https://primer-query.githubapp.com/?query=name%3Acircleocticon+-repo%3Agithub%2Fgithub-ui), there are a few repos that use it and these are the kind of repos that get stuck on old primer versions because we don't help them upgrade 😅. The feedback we get from these teams is that primer is really hard to upgrade, so I have some guilt around it.

I'd recommend moving them to deprecated for at least one major, that's what we usually promise (and hopefully it's not a huge time commitment). But I'd also not want to block your progress, so I'll let you choose either direction. 😇

siddharthkp (approved):
Left some direction-ey feedback, approving in advance :)

TylerJDev:
> (I feel 50% strongly about this, not 100%)
> 
> Looking at [primer-query](https://primer-query.githubapp.com/?query=name%3Acircleocticon+-repo%3Agithub%2Fgithub-ui), there are a few repos that use it and these are the kind of repos that get stuck on old primer versions because we don't help them upgrade 😅. The feedback we get from these teams is that primer is really hard to upgrade, so I have some guilt around it.
> 
> I'd recommend moving them to deprecated for at least one major, that's what we usually promise (and hopefully it's not a huge time commitment). But I'd also not want to block your progress, so I'll let you choose either direction. 😇

Definitely makes sense. In retrospect, I think we should've deprecated the components as soon as we decided to "bonfire" them, as it would've given more time for folks to move away from those components. I'm indifferent to the approach we take, but might lean more towards removing it, only because we've already removed some for v38 😅. 

I do think we should follow the path of deprecating first then waiting for the next major. I can follow up with either a migration guide, or a lint rule that could help with existing usage outside of Dotcom if we end up removing this one.

cc: @joshblack, curious what your take would be on this!

joshblack:
I think my ideal would be:

- Deprecate a component or functionality in the existing major when we decide to remove it (my preference would be just adding the deprecated tag instead of also moving it to the entrypoint)
- Remove it in the next major

For this one, I'm down either way. I don't think there's any harm in keeping these components around really.

Also wanted to share if it helps with decision making, I recently added support for querying by version in Primer Query. Here's a link to projects on the current major of Primer React that are using this component: https://primer-query.githubapp.com/?query=name%3Acircleocticon+-repo%3Agithub%2Fgithub-ui+-repo%3Agithub%2Fprimer-docs+package%3A%22%40primer%2Freact%22+version%3A%3E%3D37.x

If we're thinking about this release specifically, if we can help teams migrate that'd be great. If we don't have a clear migration story though then it might be worth us revisiting this to see what they should use instead and reassessing if there is a good path moving forward (I'm not caught up on context so let me know if that's already the case 👀)

TylerJDev:
Yup that's fair. I think for now I'll go the route of deprecating the component and we can revisit it in our next major release. Thanks y'all!

PR #6690, opened 2025-08-28
Remove sx prop support from the `Token` component 

liuliu-dev:
**Wait until this PR is merged in `github-ui`: https://github.com/github/github-ui/pull/3366**

Closes [#5813](https://github.com/github/primer/issues/5813)

Need to be careful about the `PolymorphicForwardRefComponent` type change.

Need to do changes in `github-ui` since it's[ still using `Token` component with `sx`](https://primer-query.githubapp.com/?query=attribute%3A%22sx%22+name%3Atoken+package%3A%22%40primer%2Freact%22+version%3A%3E%3D37.x), tracking here: https://github.com/github/primer/issues/5827
sx usage

### Rollout strategy


packages/react/src/TextInputWithTokens/__snapshots__/TextInputWithTokens.test.tsx.snap
```diff
@@ -60,7 +60,7 @@ exports[`TextInputWithTokens > renders a leadingVisual and trailingVisual 1`] =
             tabindex="-1"
             text="zero"
           >
-            <div
```
francinelucca, line 63: do we know why they all changed from div to span? 🤔

liuliu-dev: I think it's because of the changes in `TokenBase` from `BoxWithFallback` to `Component`:

```tsx
// Before:
<BoxWithFallback as={as} /> // where `as` defaulted to 'span' in the TokenBase props, but BoxWithFallback defaulted back to 'div'

// After: 
<Component /> // where Component = as || 'span', so it defaults to 'span'
```

PR #6720, opened 2025-09-04
v38: Move Banner to main export

pksjce:
Closes #https://github.com/github/primer/issues/5628

In this PR, we will move `Banner` component to the main export. I think it is well suited to recommend for migration from `Flash` component. 
Any advise on how to measure it in terms of our new readyness scale? @adierkens @siddharthkp ?

As a companion to this I will create a path-to-green commit on `github-ui` where I will move all the imports from experimental to the main one.


siddharthkp:
> Any advise on how to measure it in terms of our new readyness scale? @adierkens @siddharthkp ?

The new scale is still a proposal, so we're using alpha/beta/stable for now. 

Buuuut, if you want [here are the 17 items in the current criteria](https://arc.net/l/quote/jojddmud) today, each criterion is worth 1 point.

PR #6721, opened 2025-09-04
Add `count` to SegmentedControlButton

cheshire137:
This adds a new `count` prop to `SegmentedControl.Button`, modeled after the `counter` that `UnderlineNav.Item` has and the `count` that `Button` has.

screenshot of Storybook showing With Counter Labels story for SegmentedControl

cc @francinelucca with whom I talked about adding such a prop

#### New

Adds the `count` prop to `SegmentedControl.Button` to allow displaying a count label on the right side of the button.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

The SegmentedControl > Features > With Counter Labels story and SegmentedControl > SegmentedControl.Button > Playground story have been updated to show off the new prop.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.changeset/unlucky-hotels-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
```
joshblack:
```suggestion
"@primer/react": minor
```

Small suggestion based on our [versioning guidelines](https://github.com/primer/react/blob/main/contributor-docs/versioning.md#changes) 👀

packages/react/src/SegmentedControl/SegmentedControl.docs.json
```diff
@@ -111,6 +111,12 @@
           "defaultValue": "",
           "description": "Whether the segment is selected. This is used for uncontrolled SegmentedControls to pick one SegmentedControlButton that is selected on the initial render."
         },
+        {
+          "name": "counter",
```
joshblack: Random question for you @langermank, do you know why for `Button` we call this `count`. but for `UnderlineNav.Item` it's `counter`? 🤔

packages/react/src/SegmentedControl/SegmentedControl.module.css
```diff
@@ -65,6 +65,12 @@
     /* stylelint-disable-next-line primer/spacing */
     margin-right: -1px;
   }
+
+  [data-component='counter'] {
```
joshblack: Would it be possible for this to be a class name itself instead of a data attribute? If I remember right, the `data-component` pattern only shows up when targeting a child component. It seems like in this case we are rendering the `span` directly but let me know if that's not the case!

joshblack (commented):
Thanks so much for taking the time to submit a PR! Appreciate it a ton 🙏 

Just left a couple of comments/suggestions, hope they make sense. I also tagged in @langermank on the design side to review 🔍

packages/react/src/SegmentedControl/SegmentedControl.module.css
cheshire137: That suits me fine! I didn't know why the data-component pattern might be used, just that UnderlineNav.Item used it. I can use an explicit class here.

packages/react/src/SegmentedControl/SegmentedControl.docs.json
langermank: I don't think there's any logical reason other than they were probably added at different times and by different people, and it would be nice if they matched 😄

langermank (commented):
Some squishyness happening, I haven't taken a closer look to find a solution but happy to if needed! 

Squishy

@cheshire137 if you wouldn't mind, could you add the new feature story you added to [this file](https://github.com/primer/react/blob/e11a508300c9bcd2750008f518a5e85258ee926e/e2e/components/SegmentedControl.test.ts) so it will create a VRT snapshot? No worries if not, we can add it in a followup!

packages/react/src/SegmentedControl/SegmentedControl.docs.json
joshblack: @langermank do you have a preference for `count` or `counter`? Mine would be `count`, I think

langermank: Agree on "count"!

cheshire137: I can update!

joshblack (approved):

langermank (approved):
I'm going to open a followup PR to fix the overflow problem 👍

cheshire137:
Oh heck, I forgot all about that comment, sorry about that and thank you @langermank! If you'd prefer, we could get it fixed in this branch.

langermank:
@cheshire137 no worries at all. I dug into it and its not really tied to the count addition, it's pre-existing. So I think it's probably better to fix it in a separate PR anyway!

cheshire137:
I'll get conflicts resolved.

PR #6728, opened 2025-09-04
Breadcrumb overflow styling

langermank:
- Adds a variant prop to handle the two styles
- CSS cleanup and consolidation
- More feature stories


pksjce:
Thanks for this!
I made the following tweaks
- Got the stories back to state where I can flick the FF from the top menu
- Made variant to be applicable only with new ff. The old one can remain as is.

pksjce:
@copilot fix the lint error

pksjce (approved):

packages/react/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -365,10 +371,7 @@ const BreadcrumbsItem = React.forwardRef(({selected, className, ...rest}, ref) =
   return (
     <BoxWithFallback
       as="a"
-      className={clsx(className, classes.Item, {
-        [SELECTED_CLASS]: selected,
-        [classes.ItemSelected]: selected,
-      })}
+      className={clsx(className, classes.Item, selected && 'selected')}
```
langermank, line 374: @pksjce I just moved all selected logic into `aria-current` in CSS, so we don't need that ItemSelected class. No idea what `selected` is doing, probably something in dotcom

packages/react/src/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -153,6 +136,7 @@
     & summary {
       list-style: none;
       cursor: pointer;
+      outline: none;
```
pksjce, line 139: This makes the focus outline around the menu button go away.  Since button + summary is same element. When you tab along, there is no focus outline.

PR #6739, opened 2025-09-05
Revert "v38: Move Banner to main export (#6720)"

joshblack:
This reverts commit 0466819e4573155a347847bb6850d46c03525d41.

When going through the release, it seems like this change has usage in dotcom that needs to be updated before we can remove these components from the `@primer/react/experimental` entrypoint 👀


llastflowers (approved):
✅

PR #6740, opened 2025-09-05
chore(project): update to Node.js v24 LTS

joshblack:
Update project to Node.js v24. It also updates our GitHub Actions to reference our local `nvmrc` file and our devcontainer to the latest image.

#### Changed

- Update Node.js version in nvmrc to v24
- Update GitHub Actions to use the nvmrc file
- Update our devcontainer to use the latest image


.github/workflows/vrt-reports.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Use Node.js 20.x
```
joshblack:
```suggestion
      - name: Set up Node.js
```

.github/workflows/aat-reports.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Use Node.js 20.x
```
joshblack:
```suggestion
      - name: Set up Node.js
```

PR #6836, opened 2025-09-10
Add browser test script to styled-react package

liuliu-dev:
Adds a `browser-test` npm script to the `@primer/styled-react` package to run browser tests

## Changes

- Added `"browser-test": "vitest --config vitest.config.browser.ts src/__tests__/primer-react*.browser.test.tsx --run"` script to `package.json`


joshblack:
If it's helpful, I think with: https://github.com/primer/react/pull/6818 it will be brought into the full vitest setup so you can filter by the test name after running `npm test` as another option!

PR #6843, opened 2025-09-12
Allow changing initially focused button in `ConfirmationDialog`

camchenry:
- Related to https://github.com/github/memex/issues/20813

Adds the ability to change the button that is initially focused when using `ConfirmationDialog` or `useConfirm`. The majority of the time the current default behavior is desirable: focus the confirm button unless the action is dangerous. However, there are some cases where we want to automatically focus the dangerous action button to remove friction from actions that are done repeatedly. Otherwise, the user needs to press  <kbd>RightArrow</kbd> or <kbd>Tab</kbd> before every action which is tedious. This behavior should rarely be overridden, but the API should support changing this to make some actions easier for keyboard users.

https://github.com/user-attachments/assets/d4647a8e-b9e5-4711-a5ad-9d62db60da51

#### New

- Adds a `overrideButtonFocus` prop to `ConfirmationDialog`/`useConfirm` to change the button that is initially focused, overriding the default focus behavior

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

Nothing in particular: I've added an automated test for this behavior, and I've also manually validated the behavior in Storybook.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


packages/react/src/ConfirmationDialog/useConfirm.hookDocs.json
```diff
@@ -46,6 +46,11 @@
           "type": "\"normal\" | \"primary\" | \"danger\"",
           "defaultValue": "normal",
           "description": "The type of button to use for the confirm button."
+        },
+        {
+          "name": "initialFocusButton",
```
pksjce: Like you said, this prop could be dangerous.. Maybe you want something like `overrideButtonFocus` so its more clear?

camchenry: Probably a good idea, should hopefully make it more clear to any code reviewers on PRs using this prop as well.

camchenry: Updated in https://github.com/primer/react/pull/6843/commits/f33ebba7ae45c6b6fb655eb1950181e992e4d427

Issue #6752, opened 2025-09-08
IconButton does not setup the ref with the element value.


pksjce:
### Description

When you use the following

```
const buttonRef = useRef(null)
return <IconButton ... ref={buttonRef}/>
```

`buttonRef.current` is always null

### Steps to reproduce

TBD

### Version

latest

### Browser

Chrome

hellojanehere:
@pksjce just checking in on how you came by this bug? :) was it something you encountered as a blocker, did a team highlight it to you, or was it something you noticed incidentally?

pksjce:
@hellojanehere - I noticed this happening in my breadcrumbs work. I wrote some focus behavior that depends on IconButton providing a correct ref. But it failed. Since the fix needs to happen in a different component, I made an issue for it.

PR #6854, opened 2025-09-15
Breadcrumbs: Fix esc button not being able to focus on the menubuttonRef

pksjce:
Closes #https://github.com/primer/react/issues/6752

### Changelog

The details component seems to swallow the ref value here for some reason. The button ref really doesn't get set here. There seems no issue with the IconButton ref as I initially assumed. I had to workaround to find the right `summary` element for this feature to work well.

Before 

https://github.com/user-attachments/assets/e87e828f-3ebb-4a54-b53c-22e486f43e17

After

https://github.com/user-attachments/assets/ebf8b300-910d-4211-86f6-50a30fa6d208

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


pksjce:
Hi, So this is the route I first went through to fix this issue.
1. IconButton does [include the forwardedRef](https://github.com/primer/react/blob/main/packages/react/src/Button/IconButton.tsx#L58). With standalone `IconButton` I had zero issues with ref. Even with the `as="summary"`.
2. The issue here is how `Details` component works. The component renders [a different summary component](https://github.com/primer/react/blob/main/packages/react/src/Details/Details.tsx#L43) before it recognises the one that I've passed in.

Maybe we could pair on this tomorrow and find a different solution?

PR #6861, opened 2025-09-16
refactor(styled-react): update structure to reduce merge conflicts

joshblack:
Update the structure of `packages/styled-react` to reduce merge conflicts (potentially) by:

- Splitting out components into their own files instead of in the main `index.tsx` files
- Moving common dependencies (Box, SxProp, sx, SystemProps, etc) into files in `packages/styled-react` that will be where these will end up to reduce churn later on

It also exports the types from these components to make migration easier


francinelucca (approved):
👏

jonrohan (approved):

PR #6870, opened 2025-09-17
docs(styled-react): update architecture for polymorphic components

joshblack:
Updates our architecture docs for polymorphic components based on our mob pair today 👀 

This helps to address a couple of problems:

- When using the `as` prop, the base component will not get overriden
- For `sx` being applied to nested components, this approach should correctly transform `sx` into `className` and pass it along
- This will also prevent `sx` from being passed to the wrapped component


TylerJDev (approved):

francinelucca (approved):
❤️

mperrotti (approved):

PR #6881, opened 2025-09-19
Implement forced colors for progress bar 

langermank:
Closes https://github.com/github/primer/issues/5820


packages/react/src/ProgressBar/ProgressBar.module.css
```diff
@@ -59,3 +59,15 @@
     width: 100%;
   }
 }
+
+@media (forced-colors: active) {
+  :root {
+    --progress-bg: LinkText;
+    --progressBar-track-bgColor: CanvasText;
```
hectahertz, line 61: I trust your judgment here but it looks like this might be problematic in the future if we start changing values

hectahertz (approved):

packages/react/src/ProgressBar/ProgressBar.module.css
langermank: Like change the CSS vars? That's a good point. I suppose if we changed the naming we would see these in the same file and update them.. I'll merge for now but we could always just move it to define the CSS instead!

PR #6889, opened 2025-09-22
Add delay functionality to tooltips

broccolinisoup:
- Closes https://github.com/github/primer/issues/3313

We need tooltip delay functionality on icon buttons for mouse pointers. [Slack reference (Internal access only)](https://github.slack.com/archives/CSGAVNZ19/p1757997198312459)

We are adding a flexibility to delay the tooltip. 

#### New

I added a `delay` prop with an enum: `instant` (50ms) as default , `medium` (400ms), `long` (1200ms) 

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

You can view the stories below: 

For medium delay: https://primer-83c142dcae-13348165.drafts.github.io/storybook/?path=/story/components-tooltipv2-features--with-medium-delay

For long delay: https://primer-83c142dcae-13348165.drafts.github.io/storybook/?path=/story/components-tooltipv2-features--with-long-delay

You can also view this story to observe delay functionality for Icon Button: https://primer-83c142dcae-13348165.drafts.github.io/storybook/?path=/story/components-iconbutton-features--long-delayed-tooltip

Additional fun story: https://primer-83c142dcae-13348165.drafts.github.io/storybook/?path=/story/components-tooltipv2-examples--emoji-picker

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Button/IconButton.features.stories.tsx
```diff
@@ -95,3 +95,9 @@ export const KeybindingHintOnDescription = () => (
 )
 
 export const KeybindingHint = () => <IconButton icon={BoldIcon} aria-label="Bold" keybindingHint="Mod+B" />
+
+export const DelayedTooltip = () => (
+  <Tooltip text="This is a tooltip with 600ms delay" delay={600}>
+    <IconButton icon={HeartIcon} aria-label="HeartIcon" />
```
broccolinisoup, line 102: Instead of adding another prop to the IconButton, I thought we can use external tooltip if we want to delay the tooltip, as it feels like it is not as common as changing directions where we have a dedicated prop on icon button -- but totally open to feedback 🙌🏻

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -20,6 +20,11 @@ export type TooltipProps = React.PropsWithChildren<
     text: string
     type?: 'label' | 'description'
     keybindingHint?: KeybindingHintProps['keys']
+    /**
+     * Delay in milliseconds before showing the tooltip
+     * @default 50
+     */
+    delay?: number
```
siddharthkp: I really think this is a risky thing to do.

Here's my preference (in order), we should get primer designers to weigh in
1) If we can, have the delay baked into the component instead of instance
2) If we do have to add a delay on an instance, it's one fixed number, not something you have to decide. (`delayed: boolean` instead of `delay: number`)

siddharthkp (changes requested):

packages/react/src/TooltipV2/Tooltip.tsx
broccolinisoup: Yeah totally fair. I am not comfortable to have this as a free text field either, just wanted to push something as we are not super clear about the direction 😄

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -22,9 +22,11 @@ export type TooltipProps = React.PropsWithChildren<
     keybindingHint?: KeybindingHintProps['keys']
     /**
      * Delay in milliseconds before showing the tooltip
-     * @default 50
+     * @default instant 50ms
+     * medium 400ms
+     * long 1200ms
      */
-    delay?: number
+    delay?: 'instant' | 'medium' | 'long'
```
siddharthkp: nit: instant delay sounds strange? how about short delay, goes well with `short | medium | long`

packages/react/src/TooltipV2/Tooltip.features.stories.tsx
```diff
@@ -204,76 +203,18 @@ export const KeybindingHint = () => (
...
+export const WithMediumDelay = () => (
   <div className={classes.PaddedContainer}>
-    <Tooltip text="Tooltip is delayed by 600ms" delay={600}>
+    <Tooltip text="Tooltip is delayed by 600ms" delay="medium">
       <Button>With delay</Button>
     </Tooltip>
   </div>
 )
 
-export const OcticonPicker = {
```
siddharthkp: I don't mind keeping this story in Tooltip.examples.stories, especially with a comment of which instance in dotcom does this represent

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -22,9 +22,11 @@ export type TooltipProps = React.PropsWithChildren<
     keybindingHint?: KeybindingHintProps['keys']
     /**
      * Delay in milliseconds before showing the tooltip
-     * @default 50
+     * @default instant 50ms
+     * medium 400ms
+     * long 1200ms
```
siddharthkp: Can we leave a comment on how we picked these numbers? 

They feel arbitrary in code so it would be nice to have the context (or a link to the context) so that, in the future when someone wants to change these, they have more information

packages/react/src/TooltipV2/Tooltip.tsx
rotev: Is `50ms` considered instant? in this case I'm pro 'short'. Otherwise, if it's exactly 0, then maybe 'none' would be most appropriate?

broccolinisoup: This is a good point. Since we have `50ms` as default, `none` or `instant` would be misleading. I'll update this to be `short`

packages/react/src/TooltipV2/Tooltip.features.stories.tsx
broccolinisoup: I was tempted to leave this but wasn't sure if that was very helpful. I'll tidy up and push it back 🚀

packages/react/src/TooltipV2/Tooltip.tsx
broccolinisoup: Yes, good idea! I'll add some context.

packages/react/src/TooltipV2/Tooltip.features.stories.tsx
broccolinisoup: I made this a fun story 🎉 Added comments as well

Screenshot 2025-10-01 at 11 38 28 am

packages/react/src/TooltipV2/Tooltip.tsx
broccolinisoup: https://github.com/github/primer/issues/3313#issuecomment-3336696699 - Added this link in the code comments as well

PR #6902, opened 2025-09-23
Update `ActionBar` to use support deep child trees using pure context

iansan5653:
Currently `ActionBar` only supports direct children because it uses `Children` APIs to find the props to determine what to render in the menu. By switching to a pure context-based approach using a registry that stores all the necessary data, `ActionBar` can be updated to support deeply nested children, aliased children, fragments, etc. 

The children are registered with the same layout effect approach for determining their measurements, but they are now stored in a `Map` by unique ID rather than a simple array. The map allows them to update themselves if their props change and de-register themselves when they unmount. 

The challenge with this approach is preserving the order of the items in the registry vs in the React tree. We need the items in the registry to always be in the same order as the tree so that the `ActionBar` knows which ones are last (to hide first) and can preserve their order in the overflow menu. For this we rely on `Map` preserving the insertion order. Since React will always render the tree from start to end, top to bottom, we know we'll receive the elements in the right order on the first render. 

However, effect cleanup still presents a challenge. If we don't clean up after ourselves, the `ActionBar` will never know if an item got removed from the bar. But if we remove items from the map on cleanup, they will get re-added at the end of the map when they update. So instead, I've set the values to `null` on cleanup and added logic to ignore `null` values in the component. This preserves the insertion order for the lifetime of the registry.

**Caveats**: Unfortunately this approach is not perfect. While it's robust to an item getting removed from the list, or new items getting added to the end of the list, it will cause unexpected behavior if a new item is added in the middle or beginning of the list. This is because these items would be added at the end of the registry, rather than in the correct location. I haven't been able to come up with a solution for this -- we'd have to somehow trigger the registry to be rebuilt from scratch when an item is added, but I don't think that's possible without triggering infinite render loops, since items have to be added to build the registry in the first place.

**TLDR**: This change would allow consumers to wrap `ActionBar` item components, use aliased components, and otherwise provide indirect children. But the drawback is that it would cause unexpected behavior if new items are added to the beginning or middle after the initial render. 

FWIW, I think the drawback may be a reasonable tradeoff. I don't think that adding new items to the middle would be a very common scenario anyway.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -74,104 +78,92 @@ const calculatePossibleItems = (childWidthArray: ChildWidthArray, navWidth: numb
...
     const numberOfListItems =
       numberOfItemsInMenu === 1 ? numberOfItemsPossibleWithMoreMenu - 1 : numberOfItemsPossibleWithMoreMenu
-    for (const [index, child] of childArray.entries()) {
+    for (const [index, [id, child]] of registryEntries.entries()) {
       if (index < numberOfListItems) {
-        items.push(child)
+        continue
         //if the last item is a divider
-      } else if (childWidthArray[index].text === 'divider') {
+      } else if (child.type === 'divider') {
```
pksjce, line 135: Does this fix https://github.com/github/primer/issues/5239 ?

pksjce (approved):

pksjce:
Since no one else is using this component and it is experimental, we don't need an integration test passing. I'm happy for this to be merged. The drawbacks as @iansan5653 suggested are minimal. Thanks for this change!

PR #6910, opened 2025-09-26
chore: use forwardedAs prop in styled-react

francinelucca:
wraps all custom components in @primer/styled-react with forwardedAs prop to prevent issue caused by https://github.com/styled-components/styled-components/issues/1981

#### Changed

- Modify all components in styled-react/src/components/* to use `forwardedAs` prop

### Rollout strategy

- [x] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #6911, opened 2025-09-26
Add browser tests for polymorphic `as` prop behavior in styled-react components

liuliu-dev:
Adds browser tests to verify that `styled-react` components correctly handle the polymorphic `as` prop. Ensures styled-components wrappers maintain component behavior when rendering and provides test coverage for component-specific props preservation.

## Current State
Tests currently fail for several components, reproducing the existing `as` prop forwarding issues in styled-react wrappers. PR #6910 should fix these test failures.

### DIFF
--- a/packages/styled-react/src/__tests__/primer-react.browser.test.tsx
+++ b/packages/styled-react/src/__tests__/primer-react.browser.test.tsx
@@ -47,8 +47,9 @@ import {
 
 describe('@primer/react', () => {
   test('ActionList supports `sx` prop', () => {
-    render(<ActionList data-testid="component" sx={{background: 'red'}} />)
+    render(<ActionList as="ul" data-testid="component" sx={{background: 'red'}} variant="inset" />)
     expect(window.getComputedStyle(screen.getByTestId('component')).backgroundColor).toBe('rgb(255, 0, 0)')
+    expect(screen.getByTestId('component')).toHaveAttribute('data-variant', 'inset')
   })
 
   test('ActionMenu.Button supports `sx` prop', () => {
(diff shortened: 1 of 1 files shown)


packages/styled-react/src/__tests__/primer-react.browser.test.tsx
```diff
@@ -47,8 +47,9 @@ import {
 
 describe('@primer/react', () => {
   test('ActionList supports `sx` prop', () => {
-    render(<ActionList data-testid="component" sx={{background: 'red'}} />)
+    render(<ActionList as="ul" data-testid="component" sx={{background: 'red'}} variant="inset" />)
```
siddharthkp: This caught my eye, why do we need to explicitly set ActionList as="ul"?

liuliu-dev: Hi @siddharthkp, thanks for the review! Do you mean that we should use `as="div"` instead of the default `ul` for `ActionList`?

siddharthkp: No sorry, I meant why do we need to pass it any `as` prop? Does it no longer render a `<ul>` by default?

liuliu-dev: The `as` prop in this test is specifically to catch the `polymorphic` prop forwarding issue that the pattern in PR [#6910](https://github.com/primer/react/pull/6910) fixes, when you wrap a component with `styled()`, the `as` prop can get consumed by `styled-components` instead of being forwarded to the wrapped component, which causes component specific props to fail to be consumed correctly. That's why we add both `as` and `variant` here to test that `variant` is correctly passed.

siddharthkp: Oh lol, my bad! That is literally what you are testing 😅 

> Do you mean that we should use as="div" instead of the default ul for ActionList?

This is a very good idea in that case!

francinelucca (approved):
✨

PR #6935, opened 2025-10-01
[Action Bar] Fix warning with menu items

JelloBagel:
Related to https://github.com/github/github-ui/pull/3638#discussion_r2395195779

Fixed `Each child in a list should have a unique "key" prop` warning when mapping the menu items

Console log showing the warning

#### Added

- Added a header right to the ActionBar's Comment Box story 

#### Changed

- Update `key` to use `ariaLabel` instead of undefined `menuItemChildren`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -217,16 +217,10 @@ export const ActionBar: React.FC<React.PropsWithChildren<ActionBarProps>> = prop
...
-                        onClick,
-                        icon: Icon,
-                        'aria-label': ariaLabel,
-                        disabled,
-                      } = menuItem.props
+                      const {onClick, icon: Icon, 'aria-label': ariaLabel, disabled} = menuItem.props
                       return (
                         <ActionList.Item
-                          key={menuItemChildren}
+                          key={ariaLabel}
```
pksjce: Why not make it related to `index` ? Would index be changing between renders somehow?

JelloBagel: The index could change if the window is resized. For example, if the toolbar overflowed at the "Insert Link" button, then it would be the first index in the menu and the key would be "0". When the window is resized smaller to overflow the next button, then the "Insert Code" button would be the first index now with a key of "0". While each key is still unique, this can cause odd issues if the keys match different components when rerendering in React

PR #6942, opened 2025-10-03
Remove Box usage and `sx` prop from PageLayout

francinelucca:
Reintroduces https://github.com/primer/react/pull/6872 but without PageHeader and PageFooter styled wrappers (not needed and doesn't play well with slots)


francinelucca:
Needs https://github.com/github/github-ui/pull/3762

PR #6973, opened 2025-10-09
fix(AnchoredOverlay): Omit aria-label and aria-labelledby from renderAnchor props

jonrohan:
Closes https://github.com/github/primer/issues/5180

> Using the AnchoredOverlay component with its renderAnchor prop rendering an IconButton with the prop spread and aria-label="whatever" results in some TypeScript error. Adding the prop aria-labelledby={undefined} is a workaround that fixes this error.

The error was caused by a type conflict with the IconButton component's accessibility props. The IconButton component has a strict type constraint requiring either aria-label OR aria-labelledby (but not both). When the AnchoredOverlay's renderAnchor function passes props that might include aria-labelledby, and we also explicitly provide aria-label="Info", TypeScript detected a potential violation of this constraint.

#### Changed

AnchoredOverlay, omit aria-label and aria-labelledby from renderAnchor props

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


pksjce:
Do you think the requirement is more like having either one of `aria-label` or `aria-labelledby` ? In this case, `renderAnchor` wouldn't accept `aria-label` as a prop right?

jonrohan:
> Do you think the requirement is more like having either one of `aria-label` or `aria-labelledby` ? In this case, `renderAnchor` wouldn't accept `aria-label` as a prop right?

From what I could tell we weren't using these labels in the renderAnchor props, they were inherited by declaring all html attributes as the type. It is the IconButton's requirements that we use one or the other. And these are controlled by the user so I think omitting the type from the spread is valid fix here.

PR #6976, opened 2025-10-10
feat: support custom slots

francinelucca:
Extends useSlot functionality to support comparing by `__SLOT__` prop, not just component type. This allows consumers to create wrapper for subcomponents while still maintaining support with @primer/react components that use slots

#### New

- tests for useSlots and FormControl to account for new slot functionality

#### Changed

- update useSlots config and functionality to support comparing by both component types and slot names
- updated components and subcomponents that use useSlots or child.type === comparisons to account for slot names

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/styled-react/src/components/ActionList.tsx
```diff
@@ -113,6 +113,21 @@ const ActionListTrailingVisual: React.ComponentType<ActionListTrailingVisualProp
   ${sx}
 `
 
+// @ts-ignore - TS doesn't know about the __SLOT__ property
+ActionListItem.__SLOT__ = Symbol('ActionList.Item')
```
francinelucca: open to suggestions about this @ts-ignore here 
CC: @adierkens

adierkens: Didn't want to push to your branch, but this should fix the TS types: https://github.com/primer/react/commit/60799a373acd2e03fb51679e224d42b2e81b8ced

packages/react/src/Flash/Flash.tsx
```diff
@@ -28,4 +28,6 @@ if (__DEV__) {
   Flash.displayName = 'Flash'
 }
 
+Flash.__SLOT__ = Symbol('Flash')
```
siddharthkp: Curious, do we use Flash somewhere else as a valid slot?

francinelucca: condensed down to just the components used as slots as per our conv!

packages/styled-react/src/components/ActionList.tsx
```diff
@@ -136,3 +137,15 @@ export const ActionList: typeof ActionListImpl & {
   Heading: PrimerActionList.Heading,
   TrailingAction: ActionListTrailingAction,
 })
+
+// Assign slot markers after component definitions
+// @ts-ignore can't seem to fix this one
+ActionList.__SLOT__ = PrimerActionList.__SLOT__
```
adierkens: I don't think this is needed since `ActionList` isn't a sub-component or slot that I'm aware of.

francinelucca: removed!

Issue #4938, opened 2024-09-09
ActionMenu with overflow doesn’t contain scrollbars within its rounded border


dylanatsmith:
### Description

Scrollbar contained within rounded corners of ActionList

![image](https://github.com/user-attachments/assets/354bc902-dc03-4731-95ba-7ba456670f36)

### Steps to reproduce

1. Put a whole bunch of items in an ActionMenu too short to contain them
2. You may need to force enable scrollbars at OS level

### Version

Whatever’s in dotcom

siddharthkp:
Passerby notes: `overflow: scroll` should be on the ActionList inside the Overlay. not on the overflow itself, that way we can ensure the border radius of the Overlay stays intact

lesliecdubs:
Wondering if someone in Primer Design might be able to take a look and weigh in on the CSS change needed here?

tallys:
@camertron can you follow up on this if it's React?

lesliecdubs:
@tallys confirmed via Storybook that this appears to be happening in React.

francinelucca:
Tested this suggestion: https://github.com/primer/react/issues/4938#issuecomment-2340216670. Setting the overflow inside the ActionList doesn't solve the problem by itself because the ActionList is unaware of the max-height set by the overlay due to the overlay being absolutely positioned. In order for the overflow to work in the ActionList, it would have to be given a max-height as well. Throwing this back in the backlog as a solution for this will require further discussion

sunnyi101:
Thanks for submitting this issue! We’ll review it during monthly planning preparation to consider picking up based on current priorities.

francisfuzz:
> Throwing this back in the backlog as a solution for this will require further discussion

@francinelucca - Hi! I was looking at your issue comment in https://github.com/primer/react/issues/4938#issuecomment-2492089164 and came across this last line as a part of our internal work efforts. I am wondering where the discussion for the solution would normally happen: in this issue, in `github`'s Slack channels, or elsewhere?

francinelucca:
@francisfuzz  Usually in our internal triage meetings that happen on Mondays. That's where @sunnyi101's comment above came from. Next steps here would be to scope this work as part of a work stream which @sunnyi101/ @hellojanehere might have more context on. I see it has been included in the Paper Cuts board so there's a chance it gets worked on this week 🤞

pksjce:
Picked this up as my next papercut

[...]

PR #6978, opened 2025-10-10
Fix: ActionMenu with overflow doesn’t contain scrollbars within its rounded border

pksjce:
Closes #https://github.com/primer/react/issues/4938?reload=1

### Changelog

The fix is to have an element inside the Overlay that has overflow and max-height set to it.
This kind of works here. Not sure if there is a better way.
Screenshot 2025-10-10 at 10 35 43 pm

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


pksjce:
With the same fix on copilot chat, it looks like this

**Before**

https://github.com/user-attachments/assets/3d57b929-6554-4698-a130-e387a79b48d0

**After**

https://github.com/user-attachments/assets/e6535de7-2193-4199-943c-4ff2b288f9c1

.changeset/serious-schools-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Fix: ActionMenu with overflow doesn’t contain scrollbars within its rounded border
```
siddharthkp: Can we change this to what is the state now

packages/react/src/ActionMenu/ActionMenu.module.css
```diff
@@ -2,4 +2,46 @@
...
+  &:where([data-max-height-medium]) {
+    max-height: 320px;
+  }
+
+  &:where([data-max-height-large]) {
+    max-height: 432px;
+  }
+
+  &:where([data-max-height-xlarge]) {
+    max-height: 600px;
```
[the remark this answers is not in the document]
siddharthkp: Good suggestion!

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -316,7 +316,13 @@ const Overlay: FCWithSlotMarker<React.PropsWithChildren<MenuOverlayProps>> = ({
       onPositionChange={onPositionChange}
       variant={variant}
     >
-      <div ref={containerRef} className={styles.ActionMenuContainer} data-variant={responsiveVariant}>
+      <div
+        ref={containerRef}
+        className={styles.ActionMenuContainer}
+        data-variant={responsiveVariant}
+        {...(overlayProps.overflow ? {[`data-overflow-${overlayProps.overflow}`]: ''} : {})}
```
siddharthkp, line 323: Do we want a default value or not?

pksjce: not sure :( Don't want to add unnecessary `hidden`. I just want it to reflect its parent.

siddharthkp: Valid, what's the default right now? Maybe we should keep that (is it auto?)

pksjce: It came with no value(maybe default is `visible` from [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow)). And I add it only if parent has some value.

pksjce:
The integration PR seems to have conflicts even though I rebuilt this PR and it is up to date. I'll create one manually.

siddharthkp:
👋 @pksjce Let me know whenever you'd like a review on this again

pksjce:
Closing this in favor of https://github.com/primer/react/pull/7046

PR #6979, opened 2025-10-10
ActionBar: Add support for groups

TylerJDev:
Closes https://github.com/github/primer/issues/5928

Adds sub component `ActionBar.Group`. 

* Introduces submenus inside of the overflow menu when `ActionBar.Group` is used
* Adds items to overflow menu by group instead of individual item

#### New

* Adds `ActionBar.Group` sub component

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


dusave:
> Introduces submenus inside of the overflow menu when ActionBar.Group is used

@TylerJDev @pksjce I just want to make sure this is captured correctly. The use of `ActionBar.Group` shouldn't be related to submenus in the overflow.

For Groups, I was pitching that it made sense for groups of items to move into the menu in certain cases (like when you have bold, italics, and underline) because leaving only one or two of the options didn't really make sense.

For sub-menus, this was meant for a continuation of the action menu components in the action bar itself. For instance, the Heading button would have a dropdown with headings 1-6. When that overflows into the menu, that item would then need to have a sub-menu. 

Hope that clarifies things 😄

pksjce (commented):
Image

Is there extra space between the groups? Is that intentional?

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -248,11 +276,11 @@ export const ActionBar: React.FC<React.PropsWithChildren<ActionBarProps>> = prop
 
                     if (menuItem.type === 'divider') {
                       return <ActionList.Divider key={id} />
-                    } else {
+                    } else if (menuItem.type === 'action') {
                       const {onClick, icon: Icon, label, disabled} = menuItem
                       return (
                         <ActionList.Item
-                          key={label}
+                          key={id}
```
pksjce: This was flagged earlier that it could change. Hence we'd kept it as `label`

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -26,8 +26,11 @@ type ChildProps =
       icon: ActionBarIconButtonProps['icon']
       onClick: MouseEventHandler
       width: number
+      groupId?: string
+      groupLabel?: string
```
pksjce: Do we need this `groupLabel`? I'm not sure if its used anywhere

packages/react/src/ActionBar/ActionBar.tsx
TylerJDev: Gotcha! Added it back!

packages/react/src/ActionBar/ActionBar.tsx
TylerJDev: Good catch, we don't, I removed it!

TylerJDev:
> Is there extra space between the groups? Is that intentional?

It isn't. I fixed this by letting the group containers inherit the `ActionBar` container style.

PR #6999, opened 2025-10-13
feat(ActionList): Combine Item and LinkItem into unified Item component

liuliu-dev:
### **Summary**
Unifies `ActionList.Item` and `ActionList.LinkItem`  into a single `ActionList.Item` component that handles both regular items and link items. Users can create link items by simply adding `href`  or other anchor attributes to `ActionList.Item`, without needing to choose between two separate components.

All existing `ActionList.LinkItem` usage continues to work unchanged.

Closes https://github.com/github/primer/issues/5956

### **Implementation**

**Link Container**
Created `LinkItemContainerNoBox` following the pattern of `ButtonItemContainerNoBox` /`DivItemContainerNoBox`, used as `ItemWrapper` when Item should be rendered as a link .

The structure is similar to `_PrivateItemWrapper`  used in the `LinkItem`, but with some props changes : passing `userOnClick` and `inactiveText `which are no longer in scope.

**Props Merging**
The challenge is `ActionListItemProps` now includes both previous `ActionListItemProps` and `LinkProps`.

To avoid breaking current usage and support migration from `ActionList.LinkItem` to `ActionList.Item` with just the component name change (no props change), defined `INTERACTIVE_ELEMENT_PROPS` const in `shared.ts`. This const is used to strip LinkItem-specific props and only pass `containerOnlyProps` to container, avoiding props like `href` showing up in `<li>`.

Props are split into `interactiveProps` and `containerOnlyProps`.

**Migration**
Change component name, props stay the same:

```ts
// Before
<ActionList.LinkItem href="/path">Link</ActionList.LinkItem>

// After
<ActionList.Item href="/path">Link</ActionList.Item>
```

### Breaking Changes

- `ActionList.LinkItem` deprecated (still works but will show deprecation warnings)
- `ActionListItemProps` now includes link props. And removed [unused `type` prop](https://primer-query.githubapp.com/?query=ActionList.LinkItem+attribute%3A%22type%22) from `LinkProps`, it caused type conflicts(https://github.com/github/github-ui/blob/b6bf6d94ab0a339243ebdcad5e6eeefce2b93f48/packages/copilot-chat/components/autocomplete/FileSuggestion.tsx#L8) when extending the interface.

### Notes

- The ESLint rule `primer-react/prefer-action-list-item-onselect` will need to be updated to allow `onClick` when link props are present.

- Are these conditions sufficient to detect link items? Should any other props trigger link behavior?

```ts
  const isLinkItem = Boolean(
    props.href ||
      props.download ||
      props.target ||
      props.rel ||
      props.hrefLang ||
      props.to ||
      (typeof props.as === 'string' && props.as.toLowerCase() === 'a') ||
      role === 'link',
  )
```

### Rollout strategy

 

.changeset/shiny-pianos-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': major
+---
+
+ActionList.Item now supports link functionality through href and other anchor attributes, deprecating the separate ActionList.LinkItem component.
```
siddharthkp, line 5: Hi! 

What does the deprecation show up as? Does it show up as an error in CI for github-ui or only a soft suggestion in vscode?

liuliu-dev: It would only be a suggestion in vscode like this:

Screenshot 2025-10-16 at 9 26 10 AM

packages/react/src/NavList/NavList.tsx
```diff
@@ -91,15 +91,16 @@ const Item = React.forwardRef<HTMLAnchorElement, NavListItemProps>(
     }
 
     return (
-      <ActionList.LinkItem
+      <ActionList.Item
         ref={ref}
+        as="a"
```
siddharthkp: This has `as="a"` but does not have an `href` or `to`, should we throw a typescript error when that happens?

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -386,9 +386,10 @@ export const UnderlineNav = forwardRef(
                     }
 
                     return (
-                      <ActionList.LinkItem
+                      <ActionList.Item
                         key={menuItemChildren}
                         style={menuItemStyles}
+                        /* eslint-disable-next-line primer-react/prefer-action-list-item-onselect */
```
siddharthkp, line 392: (non blocker) worth asking @pksjce or @broccolinisoup if there is a reason to prefer onClick over onSelect or if we can use onSelect here

packages/react/src/ActionList/ActionList.dev.stories.tsx
```diff
@@ -257,19 +256,19 @@ export const ItemLabelStylesWithMixedDescriptions = () => (
...
-          </ActionList.LinkItem>
-          <ActionList.LinkItem>Item with no description</ActionList.LinkItem>
-          <ActionList.LinkItem>
+          </ActionList.Item>
+          <ActionList.Item>Item with no description</ActionList.Item>
+          <ActionList.Item>
             Item with inline description
             <ActionList.Description variant="block">Block description</ActionList.Description>
-          </ActionList.LinkItem>
+          </ActionList.Item>
```
siddharthkp, line 271: I think we should keep one story for LinkItem in ActionList.dev.stories so that we don't accidentally break it before it's time to remove ActionList.LinkItem.

packages/react/src/ActionList/Item.tsx
```diff
@@ -115,6 +144,18 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
+  /** Detect if this should be a link item based on props - needed early for role inference */
+  const isLinkItem = Boolean(
+    props.href ||
+      props.download ||
+      props.target ||
+      props.rel ||
+      props.hrefLang ||
+      props.to ||
+      (typeof props.as === 'string' && props.as.toLowerCase() === 'a') ||
+      role === 'link',
```
siddharthkp: can we reduce this list to the main things that make it anchor tag

For example, having hrefLang without href does not make it a link. I imagine the list is just href, as and to.

packages/react/src/ActionList/shared.ts
```diff
@@ -63,12 +63,54 @@ export type ActionListItemProps<As extends React.ElementType = 'li'> = ExcludeSe
...
+ * Props that should only appear on the interactive element (link), not the container (<li>).
+ * This includes:
+ * - Link-specific props from LinkProps
+ * - HTMLAnchorElement attributes (navigation and link behavior)
+ * - Semantic/interactive ARIA attributes
+ * - Styling props
+ * - Test identifiers
+ */
+export const INTERACTIVE_ELEMENT_PROPS = [
+  // Link-specific props from LinkProps
```
siddharthkp: How can we make sure this list is exhaustive?

packages/react/src/ActionList/shared.ts
```diff
@@ -63,12 +63,54 @@ export type ActionListItemProps<As extends React.ElementType = 'li'> = ExcludeSe
...
+  'referrerPolicy',
+  'to',
+  // Polymorphic component prop
+  'as',
+  // ARIA attributes that belong on interactive elements
+  'aria-current',
+  'aria-keyshortcuts',
+  // Styling props (should be on the interactive element)
+  'style',
+  'sx',
```
siddharthkp: just checking, do we still want sx here?

packages/react/src/ActionList/Item.tsx
```diff
@@ -219,18 +260,46 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
+
+  // Create a Set for lookup of interactive-only props
+  const interactivePropsSet = new Set<string>(INTERACTIVE_ELEMENT_PROPS)
+
+  // Split props into container and interactive element props
+  const containerOnlyProps: Record<string, unknown> = {}
+  const interactiveProps: Record<string, unknown> = {}
+
+  for (const [key, value] of Object.entries(props)) {
+    if (interactivePropsSet.has(key)) {
```
siddharthkp: Sorry, just curious, is there a benefit of converting the array into a set to use Set.has instead of Array.includes? Took me a second to follow the trail 😅

packages/react/src/ActionList/Item.tsx
```diff
@@ -219,18 +260,46 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
-    ? {role: itemRole ? 'none' : undefined, ...props}
-    : // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-      (listSemantics && {...menuItemProps, ...props, ref: forwardedRef}) || {}
+  // Props distribution depends on which component pattern is being used:
+  // 1. _PrivateItemWrapper - used by LinkItem
+  // 2. isLinkItem - direct Item usage with link props
+  // 3. default - regular button/div behavior
+
+  // Create a Set for lookup of interactive-only props
+  const interactivePropsSet = new Set<string>(INTERACTIVE_ELEMENT_PROPS)
```
siddharthkp: This is a longer list than what we checked before, are there any props that were earlier going to the `<li>` that will now end up on `<a>`? Especially in instances where ActionList.Item was used to act like a Link

For example: `ActionList.Item as="a" href=""` or `ActionList.Item onClick={...}`

packages/react/src/ActionList/Item.tsx
```diff
@@ -219,18 +260,46 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
   const wrapperProps = _PrivateItemWrapper
     ? menuItemProps
-    : !listSemantics && {
-        ...menuItemProps,
-        ...props,
-        ref: forwardedRef,
-      }
+    : isLinkItem
+      ? {
+          ...props,
```
siddharthkp: just checking, do we want to spread all the props here? Or only some of the props?

packages/react/src/ActionList/Item.tsx
```diff
@@ -219,18 +260,46 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
+          ...menuItemProps,
+          inactiveText,
+          userOnClick: interactiveProps.onClick as ((event: React.MouseEvent<HTMLAnchorElement>) => void) | undefined,
```
siddharthkp: Are we also using the other props in interactiveProps somewhere else? I could only spot onClick?

siddharthkp (changes requested):
Big PR with tricky changes, well done!

Left some questions to understand the changes better

siddharthkp:
> [ci](https://github.com/github/github-ui/actions/runs/18668106903) completed with status success.

Heads up! There seem to be more tests that are failing in the integration PR that the comment does not know about: https://github.com/github/github-ui/pull/5128. I have not checked if they are related to changes in this PR, may or may not be.

packages/react/src/NavList/NavList.tsx
liuliu-dev: I think it makes sense to validate for `href` or `to` when `as="a"` is used 

I added validation in `Item.tsx`:
```ts
  if (typeof props.as === 'string' && props.as.toLowerCase() === 'a' && !inactiveText) {
    invariant(
      props.href || props.to,
      'ActionList.Item with as="a" must have an href or to prop for proper link semantics and accessibility.',
    )
  }
```

However, this will break existing usages in github-ui where `NavList.Item` is used without `href` (like [this example](https://github.com/github/github-ui/blob/e246bbe0a532c5c5af652bdd2269bd348f52ce45/packages/global-navigation/menu/dynamic-content/PinnableDynamicSidePanelItems.tsx#L94C1-L95C1)).

PR #7012, opened 2025-10-15
chore: refactor IssueLabelToken styles

francinelucca:
Refactor IssueLabelToken to use global data-color-mode attribute for css selector instead of custom local component one from theme

#### Changed

- Refactor IssueLabelToken to use global data-color-mode attribute for css selector instead of custom local component one from theme

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Token/IssueLabelToken.module.css
+++ b/packages/react/src/Token/IssueLabelToken.module.css
@@ -1,67 +1,61 @@
-
 /* stylelint-disable primer/colors */
 @define-mixin lightThemeIssueLabel {
-      --lightness-threshold: 0.453;
-    --border-threshold: 0.96;
-    --border-alpha: max(0, min(calc((var(--perceived-lightness) - var(--border-threshold)) * 100), 1));
+  --lightness-threshold: 0.453;
+  --border-threshold: 0.96;
+  --border-alpha: max(0, min(calc((var(--perceived-lightness) - var(--border-threshold)) * 100), 1));
 
-    background: rgb(var(--label-r), var(--label-g), var(--label-b));
-    color: hsl(0deg, 0%, calc(var(--lightness-switch) * 100%));
-    border-color: hsla(
-      var(--label-h),
-      calc(var(--label-s) * 1%),
-      calc((var(--label-l) - 25) * 1%),
-      var(--border-alpha)
-    );
+  background: rgb(var(--label-r), var(--label-g), var(--label-b));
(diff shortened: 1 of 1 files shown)


.changeset/hungry-beds-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+chore: refactor IssueLabelToken styles
```
siddharthkp: Seems like an implementation detail, does this need a changelog?

packages/react/src/Token/IssueLabelToken.module.css
```diff
@@ -51,10 +52,8 @@
 }
 
 /* Selected state */
-.IssueLabel:where(
-    [data-selected='true'][data-in-color-mode*='light'],
-    [data-selected='true'][data-in-color-mode='auto']
-  ) {
+[data-color-mode*='light'] .IssueLabel:where([data-selected='true']),
+[data-color-mode='auto'] .IssueLabel:where([data-selected='true']) {
```
siddharthkp: auto should be handled like this:

https://github.com/primer/react/blob/a15faefa58c20742557de7e95bfe2813794e93a7/packages/react/src/BaseStyles.module.css#L28-L38

packages/react/src/Token/IssueLabelToken.module.css
```diff
@@ -16,46 +17,160 @@
   border-style: solid;
 }
 
+@media (prefers-color-scheme: light) {
```
francinelucca: open to suggestions on how to make this not so huge

siddharthkp: (optional) Looks like all the css inside is the same, you can put it in a mixin but it's not a big deal: https://github.com/primer/react/blob/34e7fa8cd88684c16d17e997aa7ef69d40b50340/packages/postcss-preset-primer/src/mixins/activeIndicatorLine.css#L1

packages/react/src/Token/IssueLabelToken.module.css
francinelucca: implemented! thankssss

packages/react/src/Token/IssueLabelToken.module.css
francinelucca: implemented! thanksss

PR #7038, opened 2025-10-20
feat: add live editor to Avatar stories

francinelucca:
Closes https://github.com/github/primer/issues/6162

This pull request introduces the `storybook-addon-code-editor` to the Avatar component's Storybook stories, enabling live code editing in the documentation. It also refactors the Avatar stories to support this feature and updates related tests to accommodate the new story structure.

#### New

- Added `storybook-addon-code-editor` as a dependency in `package.json` and included it in the Storybook configuration (`main.ts`)

#### Changed

- Refactored `Avatar.stories.tsx` to conditionally import and use the live editor functionality, allowing users to edit and preview the `Default` and `Playground` stories interactively
- Modified the Storybook tests to allow `argTypes` on the `Default` story for the Avatar component, accommodating the new story structure required by the live editor integration

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Avatar/Avatar.stories.tsx
+++ b/packages/react/src/Avatar/Avatar.stories.tsx
@@ -5,7 +5,7 @@ import Avatar, {type AvatarProps, DEFAULT_AVATAR_SIZE} from './Avatar'
 import type {Meta, StoryObj} from '@storybook/react-vite'
 // eslint-disable-next-line import/no-namespace
 import * as PrimerReactLibrary from '../index'
-import { makeLiveEditStory } from 'storybook-addon-code-editor';
+import {makeLiveEditStory} from 'storybook-addon-code-editor'
 
 // Conditionally import live editor with error handling
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -92,13 +92,13 @@ export const Default: StoryObj = {
 
 // Enable live editor if available
 // if (makeLiveEditStory) {
-  makeLiveEditStory(Default, {
-    availableImports: {'../../index': PrimerReactLibrary},
-    code: defaultStoryCode,
-  })
+makeLiveEditStory(Default, {
+  availableImports: {'../../index': PrimerReactLibrary},
+  code: defaultStoryCode,
+})
 
-  makeLiveEditStory(Playground, {
-    availableImports: {'../../index': PrimerReactLibrary, '../storyHelpers': {parseSizeFromArgs}},
-    code: playgroundStoryCode,
-  })
+makeLiveEditStory(Playground, {
+  availableImports: {'../../index': PrimerReactLibrary, '../storyHelpers': {parseSizeFromArgs}},
+  code: playgroundStoryCode,
+})
 // }


packages/react/src/Avatar/Avatar.stories.tsx
```diff
@@ -1,7 +1,22 @@
...
+
+// Conditionally import live editor with error handling
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let makeLiveEditStory: any = null
+try {
+  const liveEditor = await import('storybook-addon-code-editor')
+  makeLiveEditStory = liveEditor.makeLiveEditStory
+} catch {
+  // Live editor not available or broken - continue without it
+}
```
francinelucca: the tests get really mad unless we do this, some mismatch with ES imports. this was copilot's solution, open to suggestions

joshblack:
I'm definitely missing context on this (just taking a look this morning) but some things stood out to me:

- This dependency seems to bring in Monaco in order to support the live code editor which leads to a decent sized chunk (it seems like 680kB when comparing the preview and our site) 
- The dependency itself seems to be from an individual versus something supported from storybook itself, do we have a good plan for what we would do if they stop maintaining it?
- In order to support this, we would have to make changes to how our `*.stories.tsx` files are written (at least ones we want the editor in) for the format this dependency wants us to use

I tried looking up through the issue thread to see the original issue but couldn't find what the original issue was. We have access to a sandbox over in the docs site, is the idea that this isn't sufficient since it:

- Is not colocated on storybook, and
- It does not provide TypeScript/intellisense to help with discovery

Or is there more to the request?

francinelucca:
this is the original issue @joshblack https://github.com/github/primer/issues/5993. 

Deferring to @lukasoppermann for more context on the need for co-located editable stories, but I think you're on the nose with the issue being co-location.

As for the drawbacks you listed they're all true and unfortunately I don't see a way around them 😅. I tried to format this in a way where the stories don't end up being too far off from than what we have now and it'll only be the playground/default stories, but yeah it's definitely different.

joshblack:
@francinelucca ah sorry, when I said original issue I meant the problem we were having. It seemed like we had a research task but I wasn't sure where it came up, how pervasive the problem is, etc.

And sounds good! Interested to see what Lukas says 👀

lukasoppermann:
Hey,

the issue is that we are often missing in storybook stories and I was looking for a way to easily allow users to test our components better. Yes, we can do this in primer.style, but often people are in storybook and it is more convenient to work there. We also have support for all color themes in storybook, etc.

This is all the context. I only found this one solution / plugin, but I am not set on this. I also did not anticipate this being so problematic. If this has severe drawbacks we don't have to add it. I am very surprised storybook does not natively support something like this. 

Please let me know if I can provide any more helpful info.

packages/react/src/Avatar/Avatar.stories.tsx
```diff
@@ -1,7 +1,22 @@
-import type {Meta, StoryFn} from '@storybook/react-vite'
-import type {AvatarProps} from './Avatar'
-import Avatar, {DEFAULT_AVATAR_SIZE} from './Avatar'
 import {parseSizeFromArgs} from './storyHelpers'
+// @ts-expect-error -- this is fine
+import playgroundStoryCode from './StorySources/Avatar.playground.source.tsx?raw'
+// @ts-expect-error -- this is fine
```
liuliu-dev: We might not need `@ts-expect-error` if we add module declaration:

```ts
declare module '*?raw' {
  const content: string
  export default content
}
```

joshblack:
@lukasoppermann I think the reason it adds a lot is that we are needing to re-create the editor experience for this use-case. Are there examples of other storybooks that incorporate this? Tried looking around real quick but couldn't find anything significant.

For this use-case:

> I was looking for a way to easily allow users to test our components better

Is the idea that we could use our stories as a playground for folks to edit and see the results of the changes without having to spin up a dev environment?

lukasoppermann:
Hey @joshblack

Makes sense. I don't know of a storybook that uses this.

When we originally discussed it the idea was that we could use copilot to just add it. If it now turns out to be a huge time sink, feel free to skip this and just not add the editor.

> Is the idea that we could use our stories as a playground for folks to edit and see the results of the changes without having to spin up a dev environment?

Well, either this, or to replicate your use case for the component and see if it works.

PR #7039, opened 2025-10-20
Remove styled-components from dependencies for primer/react

siddharthkp:
- Remove styled-components from primer/react
- Did not remove styled-system because src/constants still uses it (https://github.com/primer/react/pull/7043)


francinelucca (approved):
✨

PR #7046, opened 2025-10-21
Fix: ActionMenu with overflow doesn’t contain scrollbars within its rounded border

pksjce:
Closes #https://github.com/primer/react/issues/4938?reload=1

### Changelog

The fix is to have an element inside the Overlay that has overflow and max-height set to it.
This kind of works here. Not sure if there is a better way.
Screenshot 2025-10-10 at 10 35 43 pm

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/purple-pigs-check.md
+++ b/.changeset/purple-pigs-check.md
@@ -1,0 +1,7 @@
+---
+"@primer/react": patch
+---
+
+Fix: ActionMenu with overflow contains scrollbars within its rounded border
+
+


pksjce:
With the same fix on copilot chat, it looks like this

**Before**

https://github.com/user-attachments/assets/3d57b929-6554-4698-a130-e387a79b48d0

**After**

https://github.com/user-attachments/assets/e6535de7-2193-4199-943c-4ff2b288f9c1

pksjce:
@siddharthkp - These are the same changes as the previous PR. Somehow it wasn't creating an appropriate integration PR. But this one seems to work fine 👍

.changeset/purple-pigs-check.md
```diff
@@ -0,0 +1,7 @@
+---
+"@primer/react": patch
+---
+
+Fix: ActionMenu with overflow contains scrollbars within its rounded border
```
siddharthkp, line 5: nit: we should say what the new state is

for example: 


```suggestion
ActionMenu: Scrollbars no longer overflow outside the menu's rounded border
```

PR #7052, opened 2025-10-22
Cleanup: Replace theme usage with css variables or their hardcoded values

siddharthkp:
Remove Leftover usage of `theme` in components

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/utils/layout.ts
```diff
@@ -1,23 +0,0 @@
-import theme from '../theme'
```
siddharthkp, line 1: was only used in .storybook/preview.jsx. Moved the hardcoded values to .storybook/preview

packages/react/.storybook/preview.jsx
```diff
@@ -10,6 +9,17 @@ import './storybook.css'
 import './primitives-v8.css'
 import {Profiler} from 'react'
 
+// TODO: Update the hard-coded values when the primitives are ready
```
siddharthkp, line 12: carrying this TODO from the past forward instead of solving it 😇

siddharthkp:
>  should this point to main instead of bye-bye 👀

maybe, but i branched this off from bye-bye, so it shows a bigger diff from main. If it's all the same, I'll just keep it in bye-bye and it will get merged together

PR #7094, opened 2025-10-28
Update examples/nextjs to use primer/react without styled-components

siddharthkp:
### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [x] None; only updated examples

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


examples/nextjs/src/app/global.css
```diff
@@ -1,8 +1,3 @@
 @import '@primer/primitives/dist/css/primitives.css';
 @import '@primer/primitives/dist/css/functional/themes/light.css';
 @import '@primer/primitives/dist/css/functional/themes/dark.css';
-
-body {
-  color: var(--fgColor-default);
-  background-color: var(--bgColor-default);
-}
```
joshblack, line 8: I think we still need these on `body` since `BaseStyles` only adds them to the root of that component, right? Or is that not the case anymore 👀

siddharthkp: Pretty sure these styles never worked because the variables are not defined if `data-color-mode` is not present on `<html>` or `<body>` 🙃 

> BaseStyles only adds them to the root of that component

That is still correct

Eventually, it would make sense for us to offer a way to have ThemeProvider set `data-color-mode` on `<html>` instead on the div rendered by ThemeProvider or BaseStyles

joshblack: Fair 😂 lol, those attributes definitely need to be there 😰

PR #7100, opened 2025-10-30
Update release workflows for NPM OIDC trusted publishing

jonrohan:
Part of https://github.com/github/primer/issues/6015

I also have a companion PR that will need to ship after this change to fix the integration workflows https://github.com/github/github-ui/pull/5972

This updates our workflows to allow for the NPM trusted publishing. With this we will no longer need to use tokens to publish the packages. https://docs.npmjs.com/trusted-publishers

One caveat though is that the configuration only allows for one workflow to publish. Because of this I needed to combine release.yml, release_candidate.yml, release_canary.yml into 1 file.

### Testing & Reviewing

I tested the canary release obviously here, I'm not sure we can test the main and candidate scenario without merging and fixing any issues that might exist. So in review, take a closer look at the `release-main:` and `release-candidate:` jobs.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,8 +30,8 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444
         with:
-          node-version-file: ".nvmrc"
-          cache: "npm"
+          node-version-file: '.nvmrc'
+          cache: 'npm'
 
       - name: Install dependencies
         run: npm ci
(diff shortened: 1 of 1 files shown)


siddharthkp (approved):
Looks good to me!

Only one allowed workflow is a bummer, but it will have to do!

PR #7109, opened 2025-10-30
Add support to `ActionList` and `ActionList.Item` for `tablist` and `tab` roles

adierkens:
This change stems from an effort to create a vertical version of [UnderlinePanels](https://primer.style/product/components/underline-panels/) leveraging an ActionList. The semantic checks in `ActionList` prevented the `role` from showing on the right item. 

#### Changed

- Updates the handling for checking list semantics of `ActionList` to include `tablist` and `tab` items in order to work with the `tab-container` element. 

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- Check for `ActionList` selection bugs. 

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/early-ghosts-enter.md
+++ b/.changeset/early-ghosts-enter.md
@@ -1,5 +1,5 @@
 ---
-'@primer/react': patch
+'@primer/react': minor
 ---
 
 Add support to ActionList for 'tablist' and 'tab' roles


packages/react/src/ActionList/Item.tsx
```diff
@@ -142,7 +144,11 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
   const itemSelectionAttribute = selectionAttribute || inferredSelectionAttribute
   // Ensures ActionList.Item retains list item semantics if a valid ARIA role is applied, or if item is inactive
   const listItemSemantics =
-    role === 'option' || role === 'menuitem' || role === 'menuitemradio' || role === 'menuitemcheckbox'
+    itemRole === 'option' ||
```
adierkens, line 147: This just seemed like a bug. We weren't using the `inferredItemRole` when checking if the item is a list item

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -186,7 +186,8 @@
     }
 
     /* active state [aria-current] */
-    &:where([data-active]) {
+    &:where([data-active]),
+    &:where([aria-selected='true']) {
```
adierkens: The implications of this may stem further than just the small change I want, but is a result of mixing React & non-React together (e.g. web-components)

In my case, the `aria-selected` role is added by the `tab-container` element so it never makes it back into the React Component. I could also add a handler to my own code to track selection changes and update the props accordingly, but I'd rather avoid that if I don't have to (a CSS only solution is much more performant). if we decide the risk of adding this is too high or it breaks to many things, I can undo this part of the change.

packages/react/src/ActionList/Item.tsx
adierkens, line 57: We may want to also derive this based on the aria-attributes (i.e. active if `aria-selected` == true)

packages/react/src/ActionList/Item.tsx
```diff
@@ -249,6 +255,7 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
         ref={listSemantics ? forwardedRef : null}
         data-variant={variant === 'danger' ? variant : undefined}
         data-active={active ? true : undefined}
+        data-selectable={includeSelectionAttribute ? true : undefined}
```
adierkens: Using this to limit the new CSS selector. Could also gate this behavior behind a FF or prop to limit exposure if needed.

.changeset/early-ghosts-enter.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
```
joshblack:
```suggestion
'@primer/react': minor
```

Small suggestion based on the versioning guidelines we have (specifically that we're adding functionality)

adierkens: 👍 I was considering it a bug (hence the `patch`), but happy to update

PR #7123, opened 2025-11-03
Add `Tabs` component

adierkens:
Adds a new experimental, headless `Tabs` utility component for managing tabbed interfaces. This is a React-first replacement for [tab-container-element](https://github.com/github/tab-container-element) enabling users to provide their own visuals and not worry about accessibility

New Tabs component with an ActionList

#### New

* Creates a new `Tabs` component and associated hooks (`useTab`, `useTabList`, `usePanel`) as an internal utility for creating tabbed interfaces

### Rollout strategy

- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/experimental/Tabs/Tabs.tsx
+++ b/packages/react/src/experimental/Tabs/Tabs.tsx
@@ -128,12 +128,16 @@ function useTabList<T extends HTMLElement>(
       return
     }
 
-    event.preventDefault()
     const tabs = getFocusableTabs(tablist)
 
     const isVertical = ariaOrientation === 'vertical'
     const nextKey = isVertical ? 'ArrowDown' : 'ArrowRight'
     const prevKey = isVertical ? 'ArrowUp' : 'ArrowLeft'
+
+    if (event.key === nextKey || event.key === prevKey || event.key === 'Home' || event.key === 'End') {
+      event.preventDefault()
+      event.stopPropagation()
+    }
 
     if (event.key === nextKey) {
       const selectedTabIndex = tabs.findIndex(tab => {


packages/react/src/internal/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+  useIsomorphicLayoutEffect(() => {
+    savedOnValueChange.current = onValueChange
+  }, [onValueChange])
+
+  useEffect(() => {
+    if (value !== undefined) {
+      setSelectedValue(value)
+    }
+  }, [value])
```
joshblack: I think this will error on the `react-hooks/set-state-in-effect` rule, maybe we could do:

```suggestion
  if (selectedValue !== value) {
    setSelectedValue(value)
  }
```

Or could use the the `useControllableState` hook as another option.

packages/react/src/internal/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,303 @@
...
+): Pick<React.HTMLProps<HTMLElement>, 'aria-label' | 'aria-labelledby'> &
+  Required<Pick<React.HTMLProps<HTMLElement>, 'aria-orientation' | 'onKeyDown'>> & {
+    ref: React.RefObject<T>
+    role: 'tablist'
+  } {
+  const {'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledby, 'aria-orientation': ariaOrientation} = props
+
+  const ref = useRef<T>(null)
+
+  const onKeyDown = useCallback(
```
joshblack: I think we can omit this now with React Compiler which is cool 😎 (specifically using `useCallback`)

packages/react/src/internal/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,303 @@
...
+  )
+
+  return {
+    ref,
+    'aria-label': ariaLabel,
+    'aria-labelledby': ariaLabelledby,
+    'aria-orientation': ariaOrientation ?? 'horizontal',
+    role: 'tablist',
+    onKeyDown,
+  }
```
joshblack: What would you say about this convention? I've seen it in a couple other places and was curious if we'd benefit from it or not. My assumption is that it makes the return result easier to extend (like if we needed to expose state fields). 

```suggestion
  return {
    tablistProps: {
      ref,
      'aria-label': ariaLabel,
      'aria-labelledby': ariaLabelledby,
      'aria-orientation': ariaOrientation ?? 'horizontal',
      role: 'tablist',
      onKeyDown,
    },
  }
```

packages/react/src/internal/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,303 @@
...
+        if (tabs.length > 0) {
+          tabs[tabs.length - 1].focus()
+        }
+      }
+    },
+    [ref, ariaOrientation],
+  )
+
+  return {
+    ref,
```
joshblack: Right now we have this guidance for `refs` and hooks: https://github.com/primer/react/blob/main/contributor-docs/style.md#prefer-authoring-hooks-that-accept-a-ref-instead-of-returning-a-ref-to-apply if it would be possible to have this work for `useTabList`, as well. Let me know if not or if it'd be too awkward 👀

packages/react/src/internal/components/Tabs/Tabs.tsx
adierkens: I like that -- I was also considering using a tuple as well: `[tablistProps, otherConfig] = useTabList()`
Do you have a preference?

packages/react/src/internal/components/Tabs/Tabs.tsx
adierkens: That shouldn't be an issue to update to use that. I can pull in `useProvidedRefOrCreate()` to make the passed in ref optional

e2e/components/Axe.test.ts
```diff
@@ -18,13 +18,14 @@ const SKIPPED_TESTS = [
...
   name: string
+  type: 'story' | 'docs'
 }
 
 const {entries} = componentsConfig
 
 test.describe('Axe tests', () => {
   for (const [id, entry] of Object.entries(entries as Record<string, Component>)) {
-    if (SKIPPED_TESTS.includes(id)) {
+    if (SKIPPED_TESTS.includes(id) || entry.type !== 'story') {
```
adierkens, line 28: This was trying to axe scan the README.mdx file, so added the additional check to avoid it.

PR #7130, opened 2025-11-04
fix(AnchoredOverlay): do not overwrite ref from prop spread

joshblack:
Closes https://github.com/github/primer/issues/6090

#### Changed

- Update `AnchoredOverlay` to no longer overwrite the `ref` prop when `overlayProps.ref` is present (in React 19, it is more common for this value to be `null` due to `forwardRef` usage)

### Rollout strategy

- [x] Patch release


joshblack:
@copilot add a test case to make sure the `ref` is not accidentally overwritten in the future

PR #7131, opened 2025-11-04
ci: add status workflow for overriding github-ci checks

joshblack:
Closes https://github.com/github/primer/issues/6045

#### New

- Add a new workflow for interacting with status checks that allows contributors to skip `github-ui` status checks when they apply the `skip github ci` label

### Rollout strategy

This is a CI change to our dev process


siddharthkp (approved):
That's cool!

Additional things to think about:

- We also have the label `integration-tests: skipped manually`. Should re re-use that one or remove that one? Used in [check-for-integration-result-comment.yml](https://github.com/primer/react/blob/main/.github/workflows/check-for-integration-result-comment.yml#L18) and [recommend-integration-tests.yml](https://github.com/primer/react/blob/main/.github/workflows/recommend-integration-tests.yml#L40)
- We should add this command to the [comment which suggests you to run integration tests](https://github.com/primer/react/blob/main/.github/workflows/recommend-integration-tests.yml#L61C277-L61C315)

joshblack:
@siddharthkp great point, I'll change from the skip github label to the one we have 👍 

For the workflows, since the checks will now be required do you want any changes to these workflows? I pushed up some messaging in the recommend one, let me know what you think 👀

PR #7142, opened 2025-11-05
TooltipV2: Hide `Tooltip` when menu is active

TylerJDev:
Closes https://github.com/github/primer/issues/6003.

<details><summary>Video example</summary>
<p>

<video src="https://github.com/user-attachments/assets/63c09dda-758b-4499-83d3-a01b2ff5a5b3" />

</p>
</details> 

Hides tooltip based on ARIA props passed to `IconButton`. This solution mainly targets `ActionMenu` usage with `IconButton`, but may be applicable elsewhere.

#### New

* Tooltip is now hidden when a menu is open.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/TooltipV2/Tooltip.tsx
+++ b/packages/react/src/TooltipV2/Tooltip.tsx
@@ -26,7 +26,7 @@ export type TooltipProps = React.PropsWithChildren<{
    */
   delay?: 'short' | 'medium' | 'long'
   /**
-   * Prevents the tooltip from opening if `true`.
+   * Private API for use internally only. Prevents the tooltip from opening if `true`.
    *
    * Accessibility note: This prop should be used with caution. Only use when needing to
    * programmatically close the tooltip in response to a specific user action, such as
@@ -34,7 +34,7 @@ export type TooltipProps = React.PropsWithChildren<{
    *
    * @default false
    */
-  closeTooltip?: boolean
+  _privateDisableTooltip?: boolean
 }> &
   React.HTMLAttributes<HTMLElement>
 
@@ -117,7 +117,7 @@ export const Tooltip: ForwardRefExoticComponent<
       className,
       keybindingHint,
       delay = 'short',
-      closeTooltip: hideTooltip = false,
+      _privateDisableTooltip = false,
       ...rest
     }: TooltipProps,
     forwardedRef,
@@ -142,7 +142,7 @@ export const Tooltip: ForwardRefExoticComponent<
           triggerRef.current &&
           tooltipElRef.current.hasAttribute('popover') &&
           !tooltipElRef.current.matches(':popover-open') &&
-          !hideTooltip
+          !_privateDisableTooltip
         ) {
           const tooltip = tooltipElRef.current
           const trigger = triggerRef.current
--- a/packages/react/src/Button/IconButton.tsx
+++ b/packages/react/src/Button/IconButton.tsx
@@ -62,7 +62,7 @@ const IconButton = forwardRef(
           type={description ? undefined : 'label'}
           direction={tooltipDirection}
           keybindingHint={keybindingHint ?? keyshortcuts}
-          closeTooltip={hasActivePopup}
+          _privateDisableTooltip={hasActivePopup}
         >
           <ButtonBase
             icon={Icon}


packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -25,6 +25,16 @@ export type TooltipProps = React.PropsWithChildren<{
...
+  /**
+   * Prevents the tooltip from opening if `true`.
+   *
+   * Accessibility note: This prop should be used with caution. Only use when needing to
+   * programmatically close the tooltip in response to a specific user action, such as
+   * opening a menu, or content where the tooltip could overlap with interactive content.
+   *
+   * @default false
+   */
+  closeTooltip?: boolean
```
siddharthkp: Can we make this an internal mechanism instead of part of the API? Don't want to accidentally introduce/promote a way of disabling tooltips unless that is the explicit intention. 

If it exists, it will get used; we still have 57 instances of [unsafeDisableTooltip](https://github.com/search?q=repo%3Agithub%2Fgithub-ui%20unsafeDisableTooltip&type=code) that have not been cleaned up 😅 

Possible solutions: 
1. Using TooltipContext that IconButton is already aware of
2. Using css to hide the tooltip, would something like `:has(button[aria-expanded=true])` work?
3. `_privateDisableTooltip` prop that looks private and we don't document (we have a [similar prop](https://github.com/primer/react/blob/b83328cfedea43db3e9227901154da54946629ea/packages/react/src/ActionList/LinkItem.tsx#L38) in ActionList)

TylerJDev:
> Random, since we already are attaching event listeners to the child could we attach something like `onClick` that will trigger hiding the tooltip? Or is that too broad of an event?
> 
> It seems like we want cases where `button:active` is true to dismiss the tooltip if I'm understanding right, let me know if not 👀

Just to confirm, is the child in this case the one within `Tooltip` (via [Tooltip.tsx#L269](https://github.com/primer/react/blob/main/packages/react/src/TooltipV2/Tooltip.tsx#L269))? Does this mean instead of applying the logic through `IconButton`, apply it within `Tooltip` through an event that checks if `button:active`?

packages/react/src/TooltipV2/Tooltip.tsx
TylerJDev: Definitely can! I'm thinking option 3 might be the best, but let me know what you think.

> Using TooltipContext that IconButton is already aware of

This could work, we are exposed to `TooltipContext`, which currently provides the ID. We're mostly relying on what ActionMenu (through `AnchoredOverlay`) provides to `IconButton` to determine if it should be hidden or not (if `aria-expanded="true"` and `aria-haspopup="true"` exist on the IconButton)

I'm not too sure what we could pass to context for Tooltip here 🤔 We could check within Tooltip for if it's a menu, but that might add more logic to handle. Let me know what you think though 👀 

> Using css to hide the tooltip, would something like :has(button[aria-expanded=true]) work?

This would definitely work! The only concern is how we'd scope it to only apply if it's both an `IconButton` and that button triggers a menu. 

e.g. `:has(button[aria-expanded=true][aria-haspopup=true])`

```html
<!-- We'd only want the CSS to apply to the second button (IconButton) -->
<button aria-haspopup="true" aria-expanded="true">Not an IconButton</button>
<button className="IconButton" aria-haspopup="true" aria-expanded="true">Is an IconButton</button>
```
There doesn't' seem to be a good identifier between the buttons 🤔 

> _privateDisableTooltip prop that we don't document (we have a [similar prop](https://github.com/primer/react/blob/b83328cfedea43db3e9227901154da54946629ea/packages/react/src/ActionList/LinkItem.tsx#L38) in ActionList)

This works! Ideally we wouldn't have to use a prop to begin with, but I figured that this approach is less likely to interfere with any existing `Tooltip` or `IconButton` instances.

siddharthkp: > There doesn't' seem to be a good identifier between the buttons 🤔

`data-component="IconButton"`?

> I'm not too sure what we could pass to context for Tooltip here 🤔 We could check within Tooltip for if it's a menu, but that might add more logic to handle. Let me know what you think though 👀

The provider is in the tooltip which wraps the button, so would have to pass a function down in context for button to call. Not my favorite.

> Definitely can! I'm thinking option 3 might be the best

Good backup if the css other does not work reliably

TylerJDev: I tried the CSS way, but since the Tooltip is a sibling of the IconButton it's not as feasible of a solution 😭 

**Example 1**: `&[popover]:not(:has(> [data-component="IconButton"][aria-haspopup='true'][aria-expanded='true'])):popover-open` - this would apply if the tooltip was wrapped around the IconButton, but it's a sibling. 

**Example 2**: `&[popover]:not(:has(~ [data-component="IconButton"][aria-haspopup='true'][aria-expanded='true'])):popover-open` - this only targets siblings after the Tooltip in the DOM, components using Tooltip will always come before, not after.

I might be missing another way, but I think the prop _might_ be the option to run with.

siddharthkp: Ouch, fair enough!

TylerJDev:
@joshblack, I think if we did utilize a click event, we'd need to check the following: 

* If the cloned control has `aria-expanded="true"` and `aria-haspopup="true"`
* If the cloned control is an `IconButton` 

e.g (mocked example). 

```ts
onClick: (event: React.MouseEvent) => {
  const {'aria-expanded': ariaExpanded, 'aria-haspopup': ariaHasPopup} = child.props

  if (ariaExpanded === true || ariaHasPopup === true && isComponentIconButton) {
    closeTooltip()
  }

  child.props.onClick?.(event)
}
```

Not sure if this is better or not 🤔 but accomplishes the same thing.

TylerJDev:
I'm thinking we could merge this one now, but happy to iterate on this! It's a new "private" prop, so we can definitely change the approach without risking a major breaking change.

PR #7143, opened 2025-11-06
Style `KeybindingHint` within Button context

langermank:
Closes https://github.com/github/primer/issues/6070

Part of https://github.com/github/github-ui/issues/6414

### Changelog

Buckle up!

Today, `KeybindingHint` is used as a `trailingVisual` for buttons. The component has it's own set of variants like `onPrimary` which is only used once in dotcom and is only ever intended to be used in Button. I think this is generally problematic but my goal is to find a solution to the visual problem without having to do much reconstruction for now 😄 

Since button has state the current implementation is sometimes broken, especially in danger buttons. I added new tokens specifically to handle this and rather than adding it to the `Chord` component I'm nesting it within the button context where it belongs. Now, when someone passes it into `Button` they don't need to set a variant (and if they do nothing will happen.)

I specifically added a data attribute instead of CSS class so that I could impact existing implementations and not add a new slot prop to `Button`.

Buttons with keybinding hints in all themes

#### New

- Visual adjustments to how `KeybindingHint` renders inside a button

#### Changed

- Added a data attribute to the `Chord` component
- New CSS targeting nested KeybindingHints

### Rollout strategy

- [ ] Patch release
- [x ] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.changeset/lovely-files-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": minor
+---
+
+Adjust styling for button's that have nested keybinding hints
```
siddharthkp, line 5: nit: recommend starting the changelog with the component's name for skimming/scanning

```suggestion
Button: Adjust styling for buttons that have a nested keybinding hint
```

packages/react/src/KeybindingHint/components/Chord.tsx
```diff
@@ -33,6 +33,7 @@ const splitChord = (chord: string) =>
 
 export const Chord = ({keys, format = 'condensed', variant = 'normal', size = 'normal'}: KeybindingHintProps) => (
   <Text
+    data-kbd-chord
```
siddharthkp, line 36: nit: Prefer `data-component="kbd-chord"` to keep it consistent with the pattern followed [in other components](https://github.com/search?q=repo%3Aprimer%2Freact+data-component&type=code)

siddharthkp (approved):
Overall looks good!

Couple of minor nits for following code patterns. Approving in advance.

--- 

Follow up requests:

> The component has it's own set of variants like onPrimary which is only used once in dotcom and is only ever intended to be used in Button. 
> Now, when someone passes it into Button they don't need to set a variant (and if they do nothing will happen.)

Once this PR is in dotcom, lets remove them and deprecate variant prop?

> Since button has state the current implementation is sometimes broken, especially in danger buttons. I added new tokens specifically to handle this and rather than adding it to the Chord component

CSS question for my curiosity: for this pattern, should the styles live in Button.css (button affects child components) or Chord.css (chord changes based on its parent/context)

PR #7149, opened 2025-11-06
StateLabel: Add size prop and deprecate variant prop

lukasoppermann:
Closes https://github.com/github/primer/issues/5328

### Changelog

Replaced StateLabel variant prop with size prop.

- removed: variant property
- added size property
- changed property value normal to medium

This is so that the component is inline with how other component apis work.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/styled-react-peer-deps.md
+++ b/.changeset/styled-react-peer-deps.md
@@ -1,0 +1,5 @@
+---
+'@primer/styled-react': patch
+---
+
+Allow any `@primer/react` version as peer dependency.
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -51,7 +51,7 @@ jobs:
     name: Production
     needs: [guard]
     if: ${{ needs.guard.outputs.should_deploy == 'true' }}
-    uses: primer/.github/.github/workflows/deploy.yml@5d641c4a200cecaddcb067822ea54dd17ae0a821
+    uses: primer/.github/.github/workflows/deploy.yml@2bbbbf004e8281856e08f6607eb46a7ee59e65ed
     secrets:
       gh_token: ${{ secrets.GITHUB_TOKEN }}
     with:
--- a/package-lock.json
+++ b/package-lock.json
@@ -27880,7 +27880,7 @@
         "typescript": "^5.9.2"
       },
       "peerDependencies": {
-        "@primer/react": "^38.0.0",
+        "@primer/react": "*",
         "@types/react": "18.x || 19.x",
         "@types/react-dom": "18.x || 19.x",
         "@types/react-is": "18.x || 19.x",
--- a/packages/styled-react/package.json
+++ b/packages/styled-react/package.json
@@ -60,7 +60,7 @@
     "typescript": "^5.9.2"
   },
   "peerDependencies": {
-    "@primer/react": "^38.0.0",
+    "@primer/react": "*",
     "@types/react": "18.x || 19.x",
     "@types/react-dom": "18.x || 19.x",
     "@types/react-is": "18.x || 19.x",


lukasoppermann:
I assume this must be a **major** change. There are 7 cases where the `variant` prop is used on GitHub: https://primer-query.githubapp.com/?query=name%3AStateLabel++attribute%3Avariant

I am not sure how we would best roll this out. I would need support here to define that.

llastflowers:
Hi! Looks like the main issue with the snapshots was actually the StateLabel e2e tests which needed to be updated to expect the new props from the component. I did this locally with `npx vitest -u` and pushed up the results. I also refreshed the VRT snapshots manually, and CI looks good now! 

I did notice that the icon size change in the updated VRT snapshots looked like it could have been unintentional, so wanted to call that out just in case, but I'll leave it to you as the designer!

lukasoppermann:
@llastflowers icons size in the small labels you mean?

.changeset/thin-planets-taste.md
```diff
@@ -0,0 +1,10 @@
+---
+'@primer/react': major
```
siddharthkp: This is no longer major

```suggestion
'@primer/react': minor
```

.changeset/thin-planets-taste.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+Replaced StateLabel variant prop with size prop.
+- removed: variant property
+- added size property
+- changed property value normal to medium
+
+This is so that the component is inline with how other component apis work.
```
siddharthkp: love seeing a good changelog

packages/react/src/StateLabel/__tests__/StateLabel.test.tsx
```diff
@@ -11,11 +11,16 @@ describe('StateLabel', () => {
     expect(HTMLRender(<StateLabel status="pullQueued" />).container).toMatchSnapshot()
   })
 
-  it('respects the variant prop', () => {
+  it('respects the deprecated variant prop', () => {
```
siddharthkp, line 14: very good test!

packages/react/src/StateLabel/__tests__/StateLabel.test.tsx
```diff
@@ -11,11 +11,16 @@ describe('StateLabel', () => {
...
     expect(HTMLRender(<StateLabel variant="small" status="issueOpened" />).container).toMatchSnapshot()
     expect(HTMLRender(<StateLabel variant="normal" status="issueOpened" />).container).toMatchSnapshot()
   })
 
+  it('respects the size prop', () => {
+    expect(HTMLRender(<StateLabel size="small" status="issueOpened" />).container).toMatchSnapshot()
+    expect(HTMLRender(<StateLabel size="medium" status="issueOpened" />).container).toMatchSnapshot()
+  })
+
```
siddharthkp, line 23: Maybe a good idea to add this test as well:

```suggestion

  it('prefers the size prop over deprecated variant prop', () => {
    expect(HTMLRender(<StateLabel size="small" variant="medium"status="issueOpened" />).container).toMatchSnapshot()
  })

```

packages/react/src/StateLabel/StateLabel.tsx
```diff
@@ -47,26 +47,36 @@ const labelMap: Record<keyof typeof octiconMap, 'Issue' | 'Issue, not planned' |
...
     // Open and closed statuses, we don't want to show an icon
     const noIconStatus = status === 'open' || status === 'closed'
 
+    // For backwards compatibility, we map variant to size
+    if (variantProp === 'small') {
+      sizeProp = 'small'
+    }
+    if (variantProp === 'normal') {
+      sizeProp = 'medium'
+    }
```
siddharthkp: Important: Usually a bad practice to overwrite variable from props. Instead assign the value to a variable and then use that variable in data-size below

bonus points if we can keep the original names from props (size and variant)

```suggestion
    // Prefer size, but maintain backwards compatibility for variant
    const inferredSize = size || variant === 'normal'  ? 'medium': 'small'
```

.changeset/thin-planets-taste.md
siddharthkp: love seeing a good changelog!

even better if:

```suggestion
StateLabel: Add size prop and deprecate variant prop to align with APIs in other components.
```

siddharthkp (changes requested):
Requested some changes

PR #7157, opened 2025-11-06
feat(Spinner): add support for synchronized animations

joshblack:
Closes https://github.com/github/primer/issues/6105

Uses a technique from Spectrum to drive our Spinner animations so that they are always in sync: https://github.com/adobe/react-spectrum/blob/ab5e6f3dba4235dafab9f81f8b5c506ce5f11230/packages/%40react-spectrum/s2/src/Skeleton.tsx#L21

We build on top of this technique since we want spinners to also work during SSR -> hydration. As a result, this PR adds the following:

- Use CSS Animations so that spinners work when first rendered to the page
- When hydrating, JavaScript will find this animation, pause it, and use the start time of that animation to synchronize all existing, and new, spinners to the same cycle

We use the Web Animations API to synchronize these since we cannot do this with CSS alone without introducing jank during the transition between CSS -> JS (at least none that I could find 😅)

Quick video with before/after with these changes:

https://github.com/user-attachments/assets/5da93c77-01b3-4342-8fc1-64bb965da82e

#### New

- Add story for synchronized spinners demo

#### Changed

- Add a feature flag to control a new behavior where spinners are synchronized

### Rollout strategy

- [x] Minor release

We will roll this out behind a feature flag first and then stabilize as-needed

### DIFF
--- a/.changeset/afraid-buckets-build.md
+++ b/.changeset/afraid-buckets-build.md
@@ -2,4 +2,4 @@
 '@primer/react': minor
 ---
 
-Update Spinner animations to be synchronized
+Add feature flag to control whether Spinner animations are synchronized


siddharthkp:
@joshblack Did you see the css first implementation with little javascript in https://github.com/github/github-ui/pull/6120, it uses negative animation-delay which I thought was very smart

joshblack:
@siddharthkp yeah! I didn't really know if it would synchronize accurately 🤔 When I tried it out and slowed it down they seemed to be off a bit:

https://github.com/user-attachments/assets/d4b33aa1-ae33-41b0-b27f-073a00b48036

This could be related to test setup though. Definitely the faster the animation the less I can see it (if at all). I also wasn't sure what would happen with the difference between SSR / client hydration and if the spinner would jump if the delay changes between what is sent in the HTML payload versus hydration

siddharthkp:
> When I tried it out and slowed it down they seemed to be off a bit

Seems super odd!

> Definitely the faster the animation the less I can see it (if at all)

Valid, not sure if it matters in the case of spinners.

> I also wasn't sure what would happen with the difference between SSR / client hydration and if the spinner would jump if the delay changes between what is sent in the HTML payload versus hydration

My (untested) guess is that it might start out of sync and then hydrate to be in sync. That would make a really cool demo!

joshblack:
@siddharthkp let me know if there is anything else you'd like to see here 👀 The MDN docs seem to point out this limitation too:

> Note that this is only possible with the Web Animation API: it is impossible to sync two separate animations with CSS animations.
> [Link](https://developer.mozilla.org/en-US/docs/Web/API/Animation/startTime#syncing_different_animations:~:text=Note%20that%20this%20is%20only%20possible%20with%20the%20Web%20Animation%20API%3A%20it%20is%20impossible%20to%20sync%20two%20separate%20animations%20with%20CSS%20animations.)

packages/react/src/Spinner/Spinner.tsx
```diff
@@ -38,14 +40,15 @@ function Spinner({
     /* inline-flex removes the extra line height */
     <span className={classes.Box}>
       <svg
+        ref={animation}
```
siddharthkp: nit: should this be `animatingRef`? I was confused why an animation was assigned to ref 😅

siddharthkp:
> @siddharthkp let me know if there is anything else you'd like to see here

Sorry just a question before I approve: Does this work with SSR or does it wait for the client bundle to start animation? If it's the later, then we're making the component worse and better at the same time. 😅 This should ideally act like a progressive enhancement

joshblack:
Sharing this over from Slack in case folks are curious 👀  _Hopefully_ this does the trick :fingers_crossed:

It gets so weird to not cause jank when going from CSS -> JS 😅 I ended up doing something like:

- Use CSS for the animation on the server side so spinners keep animating
- When JS hydrates, pause the existing CSS animation
- Use the start time of that CSS animation as the start time for all spinners (so they are always in sync)

PR #7162, opened 2025-11-07
refactor: apply react-element-default-any-props codemod

joshblack:
Closes https://github.com/github/primer/issues/6109

Applies the React 19 codemod for: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#changes-to-the-reactelement-typescript-type

#### Changed

- Apply [react-element-default-any-props](https://github.com/eps1lon/types-react-codemod#react-element-default-any-props) codemod

### Rollout strategy

I believe this is a change that won't have a public impact and will only be an internal type change.


llastflowers (approved):
✅

PR #7200, opened 2025-11-17
feat: add leadingVisual to UnderlineNav.Item

joshblack:
Closes https://github.com/github/primer/issues/5289

Add support for `leadingVisual` to `UnderlineNav.Item` and update existing references for `icon` to `leadingVisual`.

#### New

- Add support for `leadingVisual` to `UnderlineNav.Item` with the `React.ReactElement` type

#### Changed

- Deprecate the `icon` prop in favor of `leadingVisual`

### Rollout strategy

- [x] Minor release


joshblack:
@siddharthkp good point, I made an issue for this over at: https://github.com/github/primer/issues/6144 to see if we could add support for repurposing the `no-deprecated-props` rule 👀

PR #7203, opened 2025-11-17
ci: add issue comment support for status checks

joshblack:
Unfortunately our labeling workflow won't work if the PR is one from a fork 😞 Trying out one related to issue_comment so that we could have CI pass on PRs from forks.


joshblack:
.skip-github-ui

PR #7207, opened 2025-11-17
ci: allow status checks job to run from forks

joshblack:
It seems like `github/command` (understandably) does not allow to run on forks unless the option is explicitly provided. Since we need to apply this on forks, this PR enables this option.


joshblack:
.skip-integration-checks

jonrohan (approved):

PR #7208, opened 2025-11-17
refactor: apply refobjects-default codemod

joshblack:
Applies the codemod required for this change from React 19: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument

#### Changed

- Apply the codemod for the changes to `React.RefObject`
- Use `patch-package` to override the internal types for `@types/react` due to incompatibilities with `LegacyRef`. These types match what the Ref type is in React 19

### Rollout strategy

- [x] Minor release

This change loosens the types for props that include `React.RefObject` from `React.RefObject<T>` to `React.RefObject<T | null>` to mirror changes in React 19

### DIFF
--- a/.github/workflows/status-checks.yml
+++ b/.github/workflows/status-checks.yml
@@ -85,7 +85,7 @@ jobs:
           command: '.skip-integration-checks'
           allowed_contexts: pull_request
           # Note: this permission step is _critical_ to make sure only maintainers can run the command
-          permissions: write
+          permissions: 'write,admin'
       - name: Override status checks for issue comment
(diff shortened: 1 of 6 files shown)


script/setup-react-19.mts
```diff
@@ -43,4 +43,14 @@ for (const packageJsonPath of packageJsonPaths) {
...
+console.log('Removing patches for @types/react...')
+execSync('rm ./patches/@types+react+18.3.11.patch', {
+  stdio: 'inherit',
+})
+
+// eslint-disable-next-line no-console
+console.log('Installing updated dependencies...')
+execSync('npm install', {
+  stdio: 'inherit',
+})
```
pksjce, line 56: Nice!

PR #7209, opened 2025-11-17
test: update test suites for React 19

joshblack:
Closes https://github.com/github/primer/issues/6140

#### Changed

- Update or refactor tests to work in React 18/19

### Rollout strategy

This is a change to our internal test suite


francinelucca (changes requested):
Mostly LGTM, could you elaborate on why snapshots need to be removed/don't work anymore?

joshblack:
@francinelucca the big thing with the snapshots were that the ids collided (in React 19 the way `useId()` generates ids has changed). They also weren't that helpful so I either rewrote them if it made sense or removed them if they already had a visual alternative for the test

PR #7210, opened 2025-11-18
Add virtualization to FilteredActionList

kendallgassner:
Closes #

#### New

This pull request introduces virtualization support to the `FilteredActionList` component, significantly improving performance when rendering large lists. II used `@tanstack/react-virtual` library to handle the virtualization. I only add virtualizing if a developer opts-in via the `isVirtualized` prop.  

To avoid conflating the item size estimation I only virtualize non-group ActionLists

### Rollout strategy

- [ ] Patch release
- [X] Minor release

### Merge checklist

- [ ] Added/updated tests
- [X] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


kendallgassner:
Keyboard navigation concerns: 

- Home/End will not work as expected
- Arrow up on first item (expectation is to go to bottom)
- Arrow down on last item (expectation is to go to top)


✅ `focusOutBehavior: isVirtualized ? 'stop' : 'wrap',`

packages/react/src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -44,6 +46,19 @@ export interface FilteredActionListProps extends Partial<Omit<GroupedListProps,
...
+   *
+   * When to use: Enable this option when rendering lists with many items to avoid performance bottlenecks.
+   *
+   * Limitations:
+   * - Virtualization is only supported for non-grouped ActionLists.
+   * - For best results, items should have consistent heights.
+   *
+   * @default false
+   */
+  isVirtualized?: boolean
```
adierkens, line 61: Is there any way to structure this so it's more of a "bolted-on" addition rather than embedded directly within `FilteredActionList`? 

Nothing against the code you have here (and not a blocker for you to consider), I'm just a little hesitant when adding a features like this that require fundamental changes to the core component, as it's typically a sign that we aren't exposing the right primitive/modules to enable building out more complex interactions. Would love any insights from @joshblack or others as well

kendallgassner: I was thinking add the prop would be the safest way to test this functionality out -- I am hesitant to add this to ALL `FilteredActionList`'s since we have a few use cases attempting to self resolve this issue.

adierkens: 💯 I'm definitely in favor of isolating the changes to avoid breaking anything unintentionally. I was more so thinking of the potential to do it as an independant utility: 

```tsx
const { filteredActionListProps } = useVirtualizedActionList({
   ...
   ...
})

return <FilteredActionList {...filteredActionListProps} />
``` 

but that may not be possible with the props or primitives exposed from FilteredActionList today

kendallgassner:
I believe these aat errors are false positives. I am seeing `Scrollable region must have keyboard access [scrollable-region-focusable]` which should be fine since we use the input field to manage focus and aria-activedescendents.

packages/react/src/SelectPanel/SelectPanel.examples.stories.tsx
```diff
@@ -583,3 +583,83 @@ export const RenderMoreOnScroll = () => {
...
+          open={open}
+          onOpenChange={onOpenChange}
+          items={lotsOfItems}
+          selected={selected}
+          onSelectedChange={setSelected}
+          onFilterChange={setFilter}
+          width="medium"
+          height="large"
+          message={filteredItems.length === 0 ? NoResultsMessage(filter) : undefined}
+          isVirtualized={renderSubset}
```
siddharthkp, line 657: Are we ready to add this prop to `SelectPanel.docs.json` to show up in the docs?

.changeset/brown-mirrors-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Adding isVirtualized prop to allow virtualization for non-grouped FilteredActionLists
```
siddharthkp, line 5: nit 


```suggestion
FilteredActionList & SelectPanel: Add `isVirtualized` prop to allow virtualization for non-grouped lists
```

packages/react/src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -109,6 +112,13 @@ export function FilteredActionList({
   const [listContainerElement, setListContainerElement] = useState<HTMLUListElement | null>(null)
   const activeDescendantRef = useRef<HTMLElement>()
 
+  const virtualizer = useVirtualizer({
+    count: items.length,
+    getScrollElement: () => scrollContainerRef.current,
+    estimateSize: () => 35,
+    overscan: 10,
```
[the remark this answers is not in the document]
siddharthkp: @kendallgassner do you have thoughts on this suggestion, useful?

packages/react/package.json
```diff
@@ -83,6 +83,7 @@
     "@primer/live-region-element": "^0.7.1",
     "@primer/octicons-react": "^19.13.0",
     "@primer/primitives": "10.x || 11.x",
+    "@tanstack/react-virtual": "^3.13.12",
```
siddharthkp, line 86: What is the cost of this dependency?

siddharthkp (approved):
Left a few comments, Approving in advance

packages/react/src/FilteredActionList/FilteredActionList.tsx
joshblack: Just wanted to +1 the idea of layering (if possible). What comes to mind with the prop-based approach is that we would not be able to tree-shake the tanstack dependency in areas where virtualization is not used. If we were able to layer it (either through hooks or a component on top of FilteredActionList) then we could still have that separation.

But FilteredActionList is...FilteredActionList so not sure how feasible that is 😅

PR #7214, opened 2025-11-18
Remove `variant` prop from KeybindingHint Chord component 

langermank:
Closes https://github.com/github/primer/issues/6142

Removes the `variant` prop and replaces it with Tooltip specific CSS. Primary was [handled previously](https://github.com/primer/react/pull/7143). 

Depends on https://github.com/github/github-ui/pull/7169 merging to unblock test failures.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/grumpy-cats-protect.md
+++ b/.changeset/grumpy-cats-protect.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": major
+---
+
+Remove `variant` prop from KeybindingHint Chord component 


packages/react/src/KeybindingHint/components/Chord.tsx
```diff
@@ -31,13 +31,10 @@ const splitChord = (chord: string) =>
     .map(k => k.toLowerCase())
     .sort(compareLowercaseKeys)
 
-export const Chord = ({keys, format = 'condensed', variant = 'normal', size = 'normal'}: KeybindingHintProps) => (
+export const Chord = ({keys, format = 'condensed', size = 'normal'}: KeybindingHintProps) => (
```
francinelucca, line 34: Does this mean the "variant" prop in [KeyBindingHint](https://github.com/primer/react/blob/ececdb914eb0d0c5bd9699681386da744a470a76/packages/react/src/KeybindingHint/props.ts#L29) will be a no-op? Not sure we should remove this without a major 🤔

langermank: Yeah, I did add a major changeset as it is breaking. FWIW I'm removing calls from dotcom in this PR but probably safe to put this in the next major anyway! https://github.com/github/github-ui/pull/7169

PR #7218, opened 2025-11-19
Use only repo name in status-checks

joshblack:
It seems like `repositories` set this way does not work even though the app is installed on this repo 🤔 https://github.com/organizations/primer/settings/installations/51066279

Trying out just `react` to see if that works.


joshblack:
Ooo okay seems like this work based on the label flow 😌

PR #7221, opened 2025-11-19
ci: remove styled-components migration report

joshblack:
Remove the report for styled-components now that it's complete! 🥳


joshblack:
Confirmed that the workflow works by running `workflow_dispatch` which created this run: https://github.com/primer/react/actions/runs/19519131782

PR #7227, opened 2025-11-21
fix: update release workflow to create an rc for styled-react

pksjce:


.changeset/short-ladybugs-fly.md
joshblack, line 1: Should be good to omit this since we're not updating the public API or anything 👀

joshblack (approved):
Just had one comment, otherwise looks good! Thanks for doing this 🙏

PR #7230, opened 2025-11-21
Replace `aria-live` and `LiveRegion` usage

mperrotti:
Closes https://github.com/github/primer/issues/4892

#### Changed

- Replaces 'aria-live' usage with newer helpers

#### Removed

- Internal `LiveRegion` component

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/TreeView/TreeView.test.tsx
+++ b/packages/react/src/TreeView/TreeView.test.tsx
@@ -1392,11 +1392,14 @@ describe('State', () => {
 })
 
 describe('Asynchronous loading', () => {
-  it('updates aria live region when loading is done', () => {
-    // Set up live region for this test
-    const liveRegionEl = document.createElement('live-region')
-    document.body.appendChild(liveRegionEl)
+  afterEach(() => {
+    const liveRegion = document.querySelector('live-region')
+    if (liveRegion) {
+      document.body.removeChild(liveRegion)
+    }
+  })
 
+  it('updates aria live region when loading is done', async () => {
     function TestTree() {
       const [state, setState] = React.useState<SubTreeState>('initial')
 
@@ -1428,6 +1431,7 @@ describe('Asynchronous loading', () => {
         </div>
       )
     }
+    const user = userEvent.setup()
     const {getByRole} = renderWithTheme(<TestTree />)
 
     const doneButton = getByRole('button', {name: 'Load'})
@@ -1437,12 +1441,16 @@ describe('Asynchronous loading', () => {
     expect(liveRegion.getMessage('polite')).toBe('')
 
     // Click load button to mimic async loading
-    fireEvent.click(doneButton)
+    await act(async () => {
+      await user.click(doneButton)
+    })
 
     expect(liveRegion.getMessage('polite')).toBe('Parent content loading')
 
     // Click done button to mimic the completion of async loading
-    fireEvent.click(doneButton)
+    await act(async () => {
+      await user.click(doneButton)
+    })
 
     act(() => {
       vi.runAllTimers()
@@ -1450,9 +1458,6 @@ describe('Asynchronous loading', () => {
 
     // Live region should be updated
     expect(liveRegion.getMessage('polite')).toBe('Parent content loaded')
-
-    // Clean up live region
-    document.body.removeChild(liveRegionEl)
   })
 
   it('moves focus from loading item to first child', async () => {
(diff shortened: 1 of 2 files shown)


packages/react/src/DataTable/Pagination.tsx
```diff
@@ -179,7 +176,11 @@ function Range({pageStart, pageEnd, totalCount}: RangeProps) {
   const end = pageEnd
   return (
     <>
-      <Message value={`Showing ${start} through ${end} of ${totalCount}`} />
+      <VisuallyHidden>
+        <AriaStatus aria-atomic="true">
```
joshblack:
```suggestion
        <AriaStatus>
```

You should be okay on this, by default the `live-region` element sets up the polite region like this:

```html
<div id="polite" aria-live="polite" aria-atomic="true">...</div>
```

packages/react/src/TreeView/TreeView.test.tsx
```diff
@@ -1392,6 +1393,10 @@ describe('State', () => {
 
 describe('Asynchronous loading', () => {
   it('updates aria live region when loading is done', () => {
+    // Set up live region for this test
+    const liveRegionEl = document.createElement('live-region')
+    document.body.appendChild(liveRegionEl)
```
joshblack: The live region utilities should be doing this automatically, let me know if not 👀 

I think test-wise we just have to have a clean up for them, e.g.

```tsx
afterEach(() => {
  const liveRegion = document.querySelector('live-region');
  document.body.removeChild(liveRegion);
});
```

packages/react/src/TreeView/TreeView.tsx
```diff
@@ -144,8 +145,11 @@ const Root: React.FC<TreeViewProps> = ({
       }}
     >
       <>
-        <VisuallyHidden role="status" aria-live="polite" aria-atomic="true">
-          {ariaLiveMessage}
+        <VisuallyHidden>
+          {/* Message fail without passing a `key` */}
```
joshblack: Ooo what issue were you running into without `key`? 👀

packages/react/src/TreeView/TreeView.test.tsx
mperrotti: That's what I thought too, but they weren't.

packages/react/src/TreeView/TreeView.tsx
mperrotti: It just wasn't picking up the content. Here's the test failure when I don't pass a `key`:

```
 FAIL   @primer/react (chromium)  src/TreeView/TreeView.test.tsx > Asynchronous loading > updates aria live region when loading is done
AssertionError: expected '' to be 'Parent content loading' // Object.is equality

- Expected
+ Received

- Parent content loading

 ❯ toBe src/TreeView/TreeView.test.tsx:1442:44
    1440|     fireEvent.click(doneButton)
    1441| 
    1442|     expect(liveRegion.getMessage('polite')).toBe('Parent content loading')
       |                                            ^
    1443| 
    1444|     // Click done button to mimic the completion of async loading
```

packages/react/src/TreeView/TreeView.test.tsx
joshblack: @mperrotti if I remove these lines from the test it still seems to pass, are you seeing that too or no?

packages/react/src/TreeView/TreeView.tsx
joshblack: @mperrotti it seems like the test is complaining about updates not being wrapped in `act()` when debugging locally, are you seeing that as well? When updating to the following it seemed to work without `key`:

```tsx
    // Click load button to mimic async loading
    await act(async () => {
      await user.click(doneButton)
    })
```

mperrotti: Ahh ok I see! Will update. Thank you!

packages/react/src/TreeView/TreeView.test.tsx
mperrotti: Huh, that's odd. The failures may have been caused by something else I changed. Will remove.

packages/react/src/TreeView/TreeView.tsx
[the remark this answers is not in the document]
mperrotti: Nope, we're getting rid of the key. Thanks though.

PR #7231, opened 2025-11-22
chore: add className testing util

francinelucca:
Closes https://github.com/github/primer/issues/5255

This pull request adds a reusable test utility to verify that components correctly handle `className` props and applies it to the components' tests. The main focus is on improving test coverage and consistency for CSS class name behavior.

#### New

- Added the `implementsClassName` function to `utils/testing.tsx`, which provides a standardized way to test that components render both their default and custom `className` props.
- Integrated `implementsClassName` into existing components test suite, ensuring that the components correctly apply both the base and custom class names.
- Adds CI to enforce the testing util is used in every component test file

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
Test-only update

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/script/check-classname-tests.mjs
+++ b/script/check-classname-tests.mjs
@@ -12,7 +12,6 @@ const IGNORED_FILES = [
   // Add file paths relative to project root here, e.g.:
   // 'packages/react/src/SomeComponent/SomeComponent.test.tsx',
   'packages/react/src/Portal/Portal.test.tsx',
-  'packages/react/src/Autocomplete/AutocompleteOverlay.tsx',
   'packages/react/src/ActionMenu/ActionMenu.test.tsx',
   'packages/react/src/Pagination/PaginationModel.test.tsx',
   'packages/react/src/DataTable/__tests__/DataTable.test.tsx',
@@ -20,8 +19,6 @@ const IGNORED_FILES = [
   'packages/react/src/DataTable/__tests__/Pagination.test.tsx',
   'packages/react/src/FeatureFlags/__tests__/FeatureFlags.test.tsx',
   'packages/react/src/FormControl/__tests__/useFormControlForwardedProps.test.tsx',
-  'packages/react/src/internal/components/Caret.tsx',
-  'packages/react/src/ThemeProvider.tsx',
   'packages/react/src/experimental/SelectPanel2/__tests__/SelectPanelLoading.test.tsx',
   'packages/react/src/__tests__/ThemeProvider.test.tsx',
   'packages/react/src/__tests__/deprecated/ActionMenu.test.tsx',


packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
```diff
@@ -50,6 +54,12 @@ const AnchoredOverlayTestComponent = ({
 }
 
 describe('AnchoredOverlay', () => {
+  implementsClassNameBehavior(
+    AnchoredOverlay,
+    'prc-Overlay-Overlay-ViJgm',
+    component => component.container.firstChild!.childNodes[1].firstChild?.firstChild as HTMLElement,
```
francinelucca: 🫠

siddharthkp: what's going on here? 😅 I don't understand

Is this the way to get the baseHtmlElement? + Is this so weird because every component puts the className in a different spot and not just the first child?

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
```diff
@@ -50,6 +54,12 @@ const AnchoredOverlayTestComponent = ({
 }
 
 describe('AnchoredOverlay', () => {
+  implementsClassNameBehavior(
+    AnchoredOverlay,
+    'prc-Overlay-Overlay-ViJgm',
```
siddharthkp: Hmm, feels off putting the hashed className here, can we import this from .module.css file instead, does that work?

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
```diff
@@ -50,6 +54,12 @@ const AnchoredOverlayTestComponent = ({
...
+    component => component.container.firstChild!.childNodes[1].firstChild?.firstChild as HTMLElement,
+    props => <AnchoredOverlayTestComponent initiallyOpen={true} {...props} />,
+  )
```
siddharthkp: my dream test function call looks like this:

```tsx
implementsClassName(ActionList)

// or if we need to,

implementsClassName(props => <AnchoredOverlayFixture {...props} />)
```

`implementsClassName` should
1. render without custom className to figure out the base class and base element
2. render with custom className, make sure it has both classes on the base element

packages/react/src/utils/testing.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {it, expect} from 'vitest'
+
+export function implementsClassNameBehavior<TProps extends {className?: string}>(
+  Component: React.ComponentType<TProps>,
+  baseClassName?: string,
+  getClassNameElement: (component: RenderResult) => HTMLElement = component =>
+    component.container.firstChild as HTMLElement,
+  renderComponent: (props: {className?: string}) => React.JSX.Element = props => <Component {...(props as TProps)} />,
+) {
+  it('renders with the base className', () => {
```
siddharthkp: idk why but this feels like a strange test to test for. The custom className test is 10/10

packages/react/src/utils/testing.tsx
```diff
@@ -0,0 +1,24 @@
+import {type RenderResult, render as HTMLRender} from '@testing-library/react'
+import {it, expect} from 'vitest'
+
+export function implementsClassNameBehavior<TProps extends {className?: string}>(
```
siddharthkp: nit:

```suggestion
export function implementsClassName<TProps extends {className?: string}>(
```

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
francinelucca: > render without custom className to figure out the base class and base element

can you elaborate on this one? I'm confused on how it would do that 🤔

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
francinelucca: > Is this the way to get the baseHtmlElement?

Yeah, I'm trying to tell the helper how to find the element with the className

> Is this so weird because every component puts the className in a different spot and not just the first child?

Correct! a lot of them will be the first child, which I've made the default, but for some of them I expect some "I need to tell you what the styled element is" which is what this is

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
siddharthkp: (i'm literally making this up so it might not work, i should invest some time doing a proof of concept) Is there a pattern we follow for base classes? Does it always have the component name like in `prc-Overlay-Overlay-ViJgm`, is it safe to find the first `prc-*` class we find, that's probably the base class?

francinelucca: yeah we could, the main problem I see is we don't have (as far as I can tell) "getByClassName" or something of the sorts util, so even knowing the full className, I don't know how to find the element is my main problem.

Another potential roadblock I see is some components are composed of others so will definitely have multiple "prc-*" classes

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
francinelucca: Done!

packages/react/src/utils/testing.tsx
francinelucca: Removed the baseclass test

packages/react/src/utils/testing.tsx
francinelucca: Done!

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
francinelucca: Found a way around this!

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
francinelucca: Update: yes we can look for the className, cleaned it up to where we're just passing the render fixture and the className, I'm not sure about trying to infer the className just by looking for prc-* because of what I said ^ regarding composed components, let me know what you think!

script/check-classname-tests
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
siddharthkp: non blocker question: just curious, do you prefer writing this script in bash or node? 

Seems like we have a [mix of formats in scripts](https://github.com/primer/react/tree/main/script), so you can choose whichever feels easier to write/update

francinelucca: copilot wrote it for me I just went with it 😳 I can have it switch it to node if you think that'd be easier to read/maintain

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
```diff
@@ -50,6 +57,7 @@ const AnchoredOverlayTestComponent = ({
 }
 
 describe('AnchoredOverlay', () => {
+  implementsClassName(props => <AnchoredOverlayTestComponent initiallyOpen={true} {...props} />, overlayClasses.Overlay)
```
siddharthkp, line 60: so pretty!

script/check-classname-tests
siddharthkp: up to you, i don't have a strong preference

packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
siddharthkp: (thinking out loud, non blocking, ignore for now)

> so even knowing the full className, I don't know how to find the element is my main problem

Is it possible to run query selectors like `document.querySelector('[class^=prc]')` (class starts with prc)  and the first result _should_ be the element with base class?

francinelucca: > Is it possible to run query selectors like document.querySelector('[class^=prc]') (class starts with prc) and the first result should be the element with base class?

Here's an example where this doesn't prove true: 
The [ActionMenu](https://primer.style/react/storybook/?path=/story/components-actionmenu--default) is wrapped within an Overlay, so the first PRC class is not the one we're looking for here

image

siddharthkp: Ohhh, righto! I don't know why I understand this earlier.

> so even knowing the full className, I don't know how to find the element is my main problem

Is this still valid?

francinelucca: > Is this still valid?

No, I figured [how to find classNames](https://github.com/primer/react/pull/7231#discussion_r2587350040) out, but I still think we need to supply the className we're looking for as a parameter due to https://github.com/primer/react/pull/7231#discussion_r2593672585 👀

siddharthkp (approved):
Amazing!

PR #7241, opened 2025-11-25
fix(LabelGroup): add role and aria-label to hidden items overlay

francinelucca:
Closes https://github.com/github/primer/issues/5207

This pull request makes a small accessibility improvement to the `LabelGroup` component. The change adds proper ARIA attributes to the overlay for better screen reader support. 

#### New

- Added `role="dialog"` and a dynamic `aria-label` to the overlay via the `overlayProps` property, improving screen reader support for overlays displaying hidden labels.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/LabelGroup/LabelGroup.tsx
```diff
@@ -91,6 +91,7 @@ const OverlayToggle: React.FC<
         </Button>
       )}
       focusZoneSettings={{disabled: true}}
+      overlayProps={{role: 'dialog', 'aria-label': `All ${hiddenItemIds.length} labels`}}
     >
```
francinelucca, line 96: @TylerJDev wanted to get your a11y-eyes on this change since I did see that AnchoredOverlay has a role="none" set 🤔 https://github.com/primer/react/blob/main/packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx#L250, wanted to make sure it was ok to override in this case...

packages/react/src/LabelGroup/LabelGroup.tsx
```diff
@@ -91,6 +91,7 @@ const OverlayToggle: React.FC<
         </Button>
       )}
       focusZoneSettings={{disabled: true}}
+      overlayProps={{role: 'dialog', 'aria-label': `All ${hiddenItemIds.length} labels`}}
```
TylerJDev: I noticed on the example (https://primer-eb2abbd43f-13348165.drafts.github.io/storybook/?path=/story/components-labelgroup-features--truncate-auto-with-interactive-tokens) it'll say "all 4 labels" but there are more than 4 labels in the dialog 🤔 

Maybe we go with something more generic like "Additional labels"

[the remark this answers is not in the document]
TylerJDev: @francinelucca, I think this is a fair add

packages/react/src/LabelGroup/LabelGroup.tsx
TylerJDev: Should be good! I believe we override it elsewhere 👀

TylerJDev (approved):

packages/react/src/LabelGroup/LabelGroup.tsx
francinelucca: oops, thanks! I corrected it to the totalLength instead since it's actually showing all of them. I feel like "additional labels" is misleading as well

francinelucca:
Seeing same integration errors on [main branch](https://github.com/github/github-ui/actions/runs/19846744086/job/56865614306?pr=7825), merging

PR #7246, opened 2025-11-26
chore - Fix type compatibility with react 19 refs

pksjce:
This makes only type related changes.
- We convert reftypes from `RefObject<T>` to `RefObject<T|null>`
- In other places where this is not compatible we've added a `@ts-expect-error`


examples/nextjs/src/app/products/page.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
```
francinelucca: what is this? 👀 is this necessary for the React 19 changes?

pksjce: Let me remove it and see if I'm still getting the errors I was

PR #7248, opened 2025-11-26
Update page layoutdrag

mattcosta7:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/PageLayout/PageLayout.tsx
+++ b/packages/react/src/PageLayout/PageLayout.tsx
@@ -194,13 +194,17 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
 
   React.useEffect(() => {
     stableOnDrag.current = onDrag
-  }, [onDrag])
+  })
 
   React.useEffect(() => {
     stableOnDragEnd.current = onDragEnd
-  }, [onDragEnd])
+  })
 
   React.useEffect(() => {
+    if (!isDragging) {
+      return
+    }
+
     function handleDrag(event: MouseEvent) {
       stableOnDrag.current?.(event.movementX, false)
       event.preventDefault()
@@ -210,6 +214,20 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
       setIsDragging(false)
       stableOnDragEnd.current?.()
       event.preventDefault()
+    }
+
+    window.addEventListener('mousemove', handleDrag)
+    window.addEventListener('mouseup', handleDragEnd)
+
+    return () => {
+      window.removeEventListener('mousemove', handleDrag)
+      window.removeEventListener('mouseup', handleDragEnd)
+    }
+  }, [isDragging])
+
+  React.useEffect(() => {
+    if (!isKeyboardDrag) {
+      return
     }
 
     function handleKeyDrag(event: KeyboardEvent) {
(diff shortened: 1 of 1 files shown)


packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -194,13 +194,17 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
 
   React.useEffect(() => {
     stableOnDrag.current = onDrag
-  }, [onDrag])
+  })
 
   React.useEffect(() => {
     stableOnDragEnd.current = onDragEnd
-  }, [onDragEnd])
```
mattcosta7, line 201: assignment is cheaper than comparison for these

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -857,3 +910,75 @@ Header.__SLOT__ = Symbol('PageLayout.Header')
...
+}
+
+type RafAccumulatorControls = {
+  enqueue: (delta: number, options?: RafAccumulatorOptions) => void
+  flush: () => void
+  cancel: () => void
+}
+
+/**
+ * Batches numeric updates so that only a single callback runs per animation frame.
```
manuelpuyol, line 925: ❤️

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -857,3 +910,75 @@ Header.__SLOT__ = Symbol('PageLayout.Header')
...
+ * Falls back to synchronous updates when the DOM is not available (SSR/tests).
+ */
+function useRafAccumulator(applyDelta: (delta: number) => void): RafAccumulatorControls {
```
manuelpuyol, line 928: is this something we could extract from this component and maybe use as a pattern in other cases similar to this one?

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -232,32 +250,38 @@ const VerticalDivider: React.FC<React.PropsWithChildren<DividerProps & Draggable
       stableOnDragEnd.current?.()
       event.preventDefault()
     }
-    // TODO: Support touch events
```
manuelpuyol, line 235: is this todo still relevant?

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -623,15 +647,38 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
...
-      }
-    }
+    const updatePaneWidth = React.useCallback(
+      (nextWidth: NextPaneWidth, {persist = false}: {persist?: boolean} = {}) => {
+        setPaneWidth(previous => {
+          const resolvedWidth = typeof nextWidth === 'function' ? nextWidth(previous) : nextWidth
+
+          if (persist) {
+            try {
+              localStorage.setItem(widthStorageKey, resolvedWidth.toString())
```
manuelpuyol, line 659: nit: could we look into IndexedDB here? I recall @JasonMore saying that it would be faster :)

packages/react/src/PageLayout/PageLayout.tsx
mattcosta7: yeah - I think it will need to be moved though!

packages/react/src/PageLayout/PageLayout.tsx
mattcosta7: yep - leaving here to minimize the files I needed to copy for now!

mattcosta7:
@copilot can we further improve performance here by using element.SetPoinyerCapture /release pointer on start/stop for dragging by mouse.  If we do this we can probably avoid a user-select none override on. Body * and just handle that here since the drag element will own the pointer, right?

Can we also set a transform instead of width updates while dragging and only commit the width at the end of the drag to avoid document reflowing

mattcosta7:
moving this PR to https://github.com/primer/react/pull/7251

PR #7259, opened 2025-12-02
AvatarStack: Prevent border styles from applying to non-avatar components

TylerJDev:
Fixes multiple minor issues with the `AvatarStack` component.

Additional context: [Slack conversation](https://github.slack.com/archives/C0948TXP4U

### With single `Avatar`
**Before**
AvatarStack component with single circular Avatar, with an incorrect square border

**After**
AvatarStack component with single circular Avatar, with a circular border

### With multiple `Avatar`(s)
**Before**
AvatarStack component with multiple circular Avatars, with cutoff borders in dark theme

AvatarStack component with multiple circular Avatars, with cutoff borders in light theme

**After**
AvatarStack component with multiple circular Avatars, with correct borders in dark theme

AvatarStack component with multiple circular Avatars, with correct borders in light theme

#### Changed

* Fixes issue where square border applies when using non-avatar children within `AvatarStack`
* Fixes issue where using `square={true}` did not apply on individual `Avatar` components within `AvatarStack`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


.changeset/bitter-cougars-deny.md
```diff
@@ -0,0 +1,8 @@
+---
+'@primer/react': patch
+---
+
+AvatarStack:
+
+* Border was incorrectly applying to elements that were not `Avatar` children.
+* The square prop was not applying to individual `Avatar` components.
```
joshblack: Just a thought, the way this shows up in our changelog it could be helpful to split this up into two changesets

packages/react/src/AvatarStack/AvatarStack.module.css
```diff
@@ -36,7 +36,7 @@
   }
 
   &:where([data-avatar-count='1'][data-shape='circle']) {
-    .AvatarItem {
+    .AvatarItem:where([data-component='Avatar']) {
```
joshblack, line 39: It seemed like we were running into a case like `AvatarStack -> div.AvatarItem -> Avatar` and we were incorrectly targeting `div.AvatarItem`, is that the case?

If so, does this selector need to target descendent `data-component='Avatar'` or is it okay like this?

.changeset/bitter-cougars-deny.md
TylerJDev: That's fair! I added an additional changeset 😁

packages/react/src/AvatarStack/AvatarStack.module.css
```diff
@@ -143,6 +143,10 @@
     box-shadow: 0 0 0 var(--avatar-border-width) transparent;
   }
 
+  &:not([data-component='Avatar']) {
+    border-radius: 50%;
+  }
```
TylerJDev: Fixes issue where border would be clipped due to the non-Avatar components not respecting the border radius.

packages/react/src/AvatarStack/AvatarStack.module.css
TylerJDev: Yeah exactly! For this issue, we were applying styles that should only affect `Avatar` onto non-Avatar components: 

For example: 

```
- AvatarStack
  - Link (Border style would be applied here instead of `Avatar`)
    - Avatar
```

vs when only using `Avatar`

```
- AvatarStack
  - Avatar (Border style is applied correctly to `Avatar`)
```

The fix in this PR will ensure that the style will be added to only `Avatar`(s) via `[data-component='Avatar']`, which should always target the descendant that is the `Avatar`.

joshblack (approved):

Issue #6171, opened 2025-06-05
Migrate bundling to rolldown-based setup


TheAlexLichter:
At the moment, Primer React is [bundled via `rollup`](https://github.com/primer/react/blob/100c9cf5f9e8ef44de0e7af89ad9c7d438403e8e/packages/react/rollup.config.mjs#L4). To improve build speed, switching to [Rolldown](https://rolldown.rs/), either plain or via a bundler like [tsdown](https://github.com/rolldown/tsdown/) or [obuild](https://github.com/unjs/obuild) might be worth a shot.

joshblack:
Hi there, @TheAlexLichter! 👋 Thanks for reaching out about this!

I actually tried out rolldown recently to see if it worked in our existing setup (specifically with `rolldown -c rollup.config.mjs` in `packages/react` but sadly ran into some errors with `@rollup/plugin-commonjs` and then our custom `'preserve-directives'` plugin for making sure that client directives (like `'use client'`) appear in specific bundles we ship.

Curious if you have any thoughts on how to fix these or if there was anything you'd recommend doing 👀

TheAlexLichter:
Hey @joshblack 👋 

* `@rollup/plugin-commonjs` should be supported out of the box, so you could try removing it in the config. Which issues did you encounter?
* Same for the custom rollup plugin - any issue you've seen.

joshblack:
Sure thing @TheAlexLichter, I opened up a PR here to make it easier to share/talk about: https://github.com/primer/react/pull/6175

I removed the commonjs plugin and then got this error:

```
[plugin preserve-directives]
RangeError: Maximum call stack size exceeded
    at get preserveModules (file:///Users/joshblack/gh/primer/react/node_modules/rolldown/dist/shared/src-IinHd_xa.mjs:3267:16)
    at get preserveModules (file:///Users/joshblack/gh/primer/react/node_modules/rolldown/dist/shared/src-IinHd_xa.mjs:3267:16)
    at get preserveModules (file:///Users/joshblack/gh/primer/react/node_modules/rolldown/dist/shared/src-IinHd_xa.mjs:3267:16)
    at get preserveModules (file:///Users/joshblack/gh/primer/react/node_modules/rolldown/dist/shared/src-IinHd_xa.mjs:3267:16)
    at get preserveModules (file:///Users/joshblack/gh/primer/react/node_modules/rolldown/dist/shared/src-IinHd_xa.mjs:3267:16)
[...]

PR #7263, opened 2025-12-02
refactor(project): update project to use rolldown over rollup

joshblack:
Closes https://github.com/primer/react/issues/6171

Update project to use rolldown over rollup. Rolldown is a part of voidzero and is powered by a Rust-based toolchain through projects like oxc. It features a rollup-compatible API so (ideally) speeds up our builds without many changes 😄 

It also allows us to remove several rollup plugins that are baked into the library, namely ones for commonjs, json, resolution, and more. We can also remove some of our rollup custom plugins, like the one we have for preserving directives.

#### Changed

- Update `rollup-plugin-import-css` to `rolldown-plugin-import-css`
  - Update how the plugin matches files by using rolldown's `filter` option which allows for more efficient bridging between Rust <-> JS
  - Update our logic for creating the `*.module.css.js` asset since rolldown's behavior is different than rollups. Now, the asset extension no longer gets a `.js` automatically so we needed to add that in and update the logic accordingly
  - Remove build step since Node.js natively supports `*.ts` extensions now 😄 
- Update the following packages to use rolldown:
  - `doc-gen`
  - `mcp`
  - `react`
  - `styled-react`
- Add external deps to `styled-react` that were missing (`clsx`, `deepmerge`, and `focus-visible`)

### Rollout strategy

This is a change to our internal build and should have no user-facing impact

### DIFF
--- a/.changeset/brown-bats-run.md
+++ b/.changeset/brown-bats-run.md
@@ -1,5 +1,0 @@
----
-'@primer/react': patch
----
-
-"Update logic for how hex codes are applied"
--- a/.changeset/dry-pets-appear.md
+++ b/.changeset/dry-pets-appear.md
@@ -1,5 +1,0 @@
----
-'@primer/react': minor
----
-
-Make MappedActionListItem a forwardRef component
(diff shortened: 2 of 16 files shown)


joshblack:
Breaking out into: https://github.com/primer/react/pull/7298

PR #7265, opened 2025-12-03
Delete .github/workflows/add_to_inbox.yml

jonrohan:
Deleting this workflow because it is redundant now. This workflow in the project will auto add the issues to the project https://github.com/orgs/github/projects/4503/workflows/68218328


francinelucca (approved):
🧹

PR #7294, opened 2025-12-09
feat(rollup-plugin): add support for css files to plugin

joshblack:
Update our rollup plugin to handle plain `*.css` files in addition to `*.module.css` files. This will allow us to sidecar files to our entrypoints, like a `layer.css` file, in the future if we would like to (like for our CSS Layers ADR).

This change also coincides with an update we made for the rolldown experiment. Instead of using rollup's default behavior of a `*.module.css` file becoming `*.module.css.js` we explicitly create this file to make it easier to port to rolldown in the future.

In addition, we force the intermediate `*.css.js` files to not be tree-shaken. This is due to how `external` files in `resolveId` are treated. Typically, we generate the CSS file alongside the `*.module.css.js` file and can refer to it like: `./file-name.css`. However, when the intermediate file is tree-shaken this path is **not** updated and is incorrect. This was a challenge for entrypoints since the path to the common file would not resolve correctly 😅 

When using the JS file instead, all paths work out as expected. If someone can find a better way to do this definitely give it a shot as this is the best I could come up with 😂


siddharthkp (approved):
Trust you with this!

PR #7297, opened 2025-12-10
test: add storybook vrt setup for styled-react

joshblack:
Closes https://github.com/github/primer/issues/6074

#### New

- Add storybook support to `styled-react`
- Add initial button stories for VRT in `styled-react`
- Add new utility, `getStories`, to our e2e test-helpers to help with more easily getting stories from a storybook source file

#### Changed

- Update our VRT setup to use two storybook URLs, one for react and one for styled-react

### Rollout strategy

This is a change to our test strategy for `styled-react`

### DIFF
--- a/packages/styled-react/tsconfig.build.json
+++ b/packages/styled-react/tsconfig.build.json
@@ -5,5 +5,5 @@
     "declaration": true,
     "emitDeclarationOnly": true
   },
-  "exclude": ["vitest.config.ts", "vitest.config.browser.ts", "config", "**/*.test.ts", "**/*.test.tsx", ".storybook"]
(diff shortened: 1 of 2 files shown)


packages/styled-react/src/components/Button.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import {ButtonComponent as Button} from './Button'
+
+export default {
+  title: 'Components/Button',
+  component: Button,
+}
+
+export const WithColor = () => <Button sx={{color: 'rebeccapurple'}}>Click me</Button>
```
francinelucca, line 8: do we have a plan for adding stories for the rest of the components? 👀

francinelucca (approved):

packages/styled-react/src/components/Button.stories.tsx
joshblack: @francinelucca I figured if folks were okay with this approach what I could do is make an umbrella issue and start fanning out from there if that works 👀

PR #7329, opened 2025-12-15
perf(Dialog): Replace body:has() with direct class

mattcosta7:
## Summary

Performance optimizations for Dialog to improve INP via a feature flag (`primer_react_css_has_selector_perf`).

### Changes

1. **Feature flag for body:has() optimization** - Add `primer_react_css_has_selector_perf` flag to toggle between legacy `:has()` selector and direct data attribute on body
2. **Dual CSS path** - Legacy path uses `body:not([data-dialog-scroll-optimized]):has(.Dialog.DisableScroll)`, optimized path uses `body[data-dialog-scroll-disabled]`
3. **Provider-level optimization marker** - `data-dialog-scroll-optimized` set at `FeatureFlags` provider level to short-circuit `:has()` evaluation with O(1) negation check
4. **Dialog-level scroll disabled marker** - `data-dialog-scroll-disabled` set on body when a dialog mounts (optimized path only)

### Expected INP Impact

| **Worst case** (Dialog open on large page, 10k+ nodes) | ~100-200ms style recalc | <5ms | **95-98% reduction** |
| **Average case** (Dialog on typical page) | ~30-60ms style recalc | <3ms | **90-95% reduction** |
| **Best case** (Dialog on small page) | ~10-20ms style recalc | <2ms | **80-90% reduction** |

### Why this matters

`body:has(.Dialog.DisableScroll)` is **extremely expensive**:
- Fires on **every style recalculation** across the entire page
- Forces full DOM traversal from body element
- Impacts all interactions while dialog is open, not just dialog interactions

By using `body[data-dialog-scroll-disabled]` (a direct attribute selector), we eliminate all traversal cost.

### Why we didn't scope :has() to portal roots

An alternative considered was scoping the `:has()` selector to portal root containers (e.g., `[data-portal-root]:has(.Dialog.DisableScroll)`). This was rejected because:

- **Still O(n) complexity** - `registerPortalRoot` sets `data-portal-root` on portal containers, but scoping `:has()` to these elements still requires traversing all descendants to find matching elements
- **No browser optimization** - Browsers cannot optimize `:has()` traversal regardless of the scoping ancestor; the selector must evaluate all descendants
- **Direct attribute is always O(1)** - Using `body[data-dialog-scroll-disabled]` is a simple attribute lookup with constant time complexity, regardless of DOM size

The only way to achieve true O(1) performance is to avoid `:has()` entirely and use a direct attribute on body.

### Why we kept :has() for footer detection

The `.Dialog:has(.Footer)` selector was considered for optimization but left unchanged because:

- **Dialogs are small** - The `:has()` scope is limited to the dialog element itself, which contains a small, bounded number of children
- **Acceptable cost** - Traversing dialog children is fast enough that optimization provides negligible benefit
- **Simpler implementation** - Avoiding additional data attributes for footer detection keeps the code cleaner

### Implementation Details

- **Flag OFF (default)**: Uses legacy `body:not([data-dialog-scroll-optimized]):has(.Dialog.DisableScroll)` selector
- **Flag ON**: Uses `body[data-dialog-scroll-disabled]` direct attribute selector
- **Short-circuit mechanism**: When flag is ON, `data-dialog-scroll-optimized` is set on body by the `FeatureFlags` provider, causing the `:not()` check to fail first (O(1)), preventing the expensive `:has()` evaluation
- **Ref counting**: Multiple dialogs and providers are handled via ref counting to ensure proper cleanup
.

### DIFF
--- a/.changeset/perf-basestyles-has-selector.md
+++ b/.changeset/perf-basestyles-has-selector.md
@@ -1,0 +1,8 @@
+---
+'@primer/react': patch
+---
+
+perf(BaseStyles): Remove expensive :has([data-color-mode]) selectors
+
+Remove `:has([data-color-mode])` selectors that scanned the entire DOM on every style recalculation.
+Input color-scheme is already handled by global selectors in the codebase.
--- a/packages/react/src/BaseStyles.module.css
+++ b/packages/react/src/BaseStyles.module.css
@@ -60,19 +60,12 @@ details-dialog:focus:not(:focus-visible):not(:global(.focus-visible)) {
   /* stylelint-disable-next-line primer/colors */
   color: var(--BaseStyles-fgColor, var(--fgColor-default));
 
-  /* Global styles for light mode */
-  &:has([data-color-mode='light']) {
-    input & {
-      color-scheme: light;
-    }
-  }
-
-  /* Global styles for dark mode */
-  &:has([data-color-mode='dark']) {
-    input & {
-      color-scheme: dark;
-    }
-  }
+  /*
+   * PERFORMANCE: Removed :has([data-color-mode]) selectors that scanned entire DOM.
+   * Input color-scheme is already handled by global selectors above:
+   *   [data-color-mode='light'] input { color-scheme: light; }
+   *   [data-color-mode='dark'] input { color-scheme: dark; }
+   */
 
   /* Low-specificity default link styling */
   :where(a:not([class*='prc-']):not([class*='PRC-']):not([class*='Primer_Brand__'])) {


packages/react/src/Dialog/Dialog.module.css
```diff
@@ -244,8 +249,9 @@ Add a border between the body and footer if:
 - the dialog has a footer
 - the dialog has a body that can scroll
 - the browser supports the `animation-timeline` property and its `scroll()` function
+PERFORMANCE: Footer is a direct child of Dialog, scope with > for O(1) lookup
 */
-.Dialog:has(.Footer) {
+.Dialog:has(> .Footer) {
```
mattcosta7: I belive this is a stable direct child access - but if not we can flip this back and update the docs.  Or maybe find a way to add an attribute to a dialog that has a footer to avoid the has lookups

packages/react/src/Dialog/Dialog.module.css
```diff
@@ -229,7 +229,12 @@
   }
 }
 
-body:has(.Dialog.DisableScroll) {
```
mattcosta7, line 232: I believe this should be ok - but if Dialogs can stack, we might want to modify the logic for adding/removing the classes to follow the stack?

mattcosta7: maybe a data-attribute that we maintain as a list or something with ids?

mattcosta7:
> This seems good on the surface, but because it's visual changes, can you please verify this has no accidental side effects in dotcom?
> 
> Orrrr, we can feature flag the change to be safe

@siddharthkp totally onboard with flagging! Is there a good pattern for that when the flag relies on changes to global styles and an eager matching selector?

if the old `:has` stype exists in the document at all, we take the cost of matching it - but i'm not sure how we'd actively flag that here!  Any thoughts?

Otherwise i'll pick this up in Jan and do some thorough testing too

siddharthkp:
> Is there a good pattern for that when the flag relies on changes to global styles...

Not yet. But adding a data-attribute is probably the way to go. However, in this case, because you are adding the className targeted by the global style styles, you can check for the feature flag in javascript.

> ...and an eager matching selector?
> if the old :has stype exists in the document at all, we take the cost of matching it - but i'm not sure how we'd actively flag that here! Any thoughts?

Yeah. That's tricky.

Does doing something like this help? I'm guessing if the data-optimised doesn't match, it's not expensive?

```css
/* we need to add both of these */

/* equivalent to old selector with default value of flag */
.Dialog[data-optimised="false"]:has(.Footer) {

/* optimised behavior when feature flag is true */
.Dialog[data-optimised="true"]:has(> .Footer) {
```

packages/react/src/Dialog/Dialog.module.css
siddharthkp: I think this is true most of the time. Unless there's something funny happening in renderFooter because it let's you wrap the footer in a custom wrapper

mattcosta7:
> > Is there a good pattern for that when the flag relies on changes to global styles...
> 
> Not yet. But adding a data-attribute is probably the way to go. However, in this case, because you are adding the className targeted by the global style styles, you can check for the feature flag in javascript.
> 
> > ...and an eager matching selector?
> > if the old :has stype exists in the document at all, we take the cost of matching it - but i'm not sure how we'd actively flag that here! Any thoughts?
> 
> Yeah. That's tricky.
> 
> Does doing something like this help? I'm guessing if the data-optimised doesn't match, it's not expensive?
> 
> ```css
> /* we need to add both of these */
> 
> /* equivalent to old selector with default value of flag */
> .Dialog[data-optimised="false"]:has(.Footer) {
> 
> /* optimised behavior when feature flag is true */
> .Dialog[data-optimised="true"]:has(> .Footer) {
> ```
> 

Ok, I think this handles it

The flagging is a bit rough, since we need to add an attribute to the body when the flag is enabled (and increment/decrement a count when doing so), but I think that handles it well enough - and expect we should be quickly able to remove that afterwards 

I removed the footer has query changes for now as well, the actual dialog content should be small enough that has is well enough scoped there as is - really just the body unbounded has is the main issue

PR #7337, opened 2025-12-15
perf(Autocomplete): Split context to reduce unnecessary re-renders

mattcosta7:
## Summary

Performance optimizations for Autocomplete to improve INP by reducing unnecessary re-renders.

### Changes

Split AutocompleteContext into separate contexts:
- **AutocompleteStaticContext** - For values that don't change (id)
- **AutocompleteSettersContext** - For setter functions (stable references)
- **AutocompleteContext** - For dynamic state that changes frequently

Components now subscribe only to the context slices they need.

### Expected INP Impact

| **Worst case** (rapid typing, many suggestions) | ~40-80ms per keystroke (full re-render) | <15ms (targeted re-render) | **60-80% reduction** |
| **Average case** (normal typing speed) | ~20-40ms per keystroke | <10ms | **50-75% reduction** |
| **Best case** (slow typing, few suggestions) | ~10-20ms per keystroke | <5ms | **50-75% reduction** |

### Why this matters

When all context values are in a single context object, **any change causes all consumers to re-render**:
- Input component re-renders when menu state changes (unnecessary)
- Menu component re-renders when input state changes (unnecessary)
- Every keystroke triggered cascading re-renders

By splitting contexts, each component only re-renders when its specific data changes.

---


llastflowers:
Code changes LGTM! Is there any particular way you'd like me to test this before approving?

mattcosta7:
> Code changes LGTM! Is there any particular way you'd like me to test this before approving?

I've mostly gone through the stories and validated they work as before

Then i've gone through the performance/profilers and validated that re-renders are more scoped than before.  I think this is a very low risk change since none of these contexts are [public api](https://github.com/primer/react/blob/e732a02175978436dc3c6fc39faccb75a0ed9804/packages/react/src/index.ts#L72-L73).

llastflowers (approved):
✨

PR #7342, opened 2025-12-15
perf(hasInteractiveNodes): Optimize with combined selector and early attribute checks

mattcosta7:
## Summary

Performance optimizations for hasInteractiveNodes utility function to improve INP.

### Changes

1. **Combined selector** - Use querySelectorAll with combined selector instead of recursive traversal
2. **Early attribute checks** - Check disabled, hidden, inert attributes before expensive getComputedStyle
3. **Deferred getComputedStyle** - Only call getComputedStyle when CSS-based visibility check is absolutely needed

### Expected INP Impact

| **Worst case** (large DOM subtree, 500+ nodes) | ~50-100ms (recursive traversal + getComputedStyle per node) | <10ms | **80-90% reduction** |
| **Average case** (medium subtree, 50-200 nodes) | ~15-40ms | <5ms | **67-88% reduction** |
| **Best case** (small subtree, <50 nodes) | ~5-15ms | <2ms | **60-87% reduction** |

### Why this matters

hasInteractiveNodes is called by:
- **AvatarStack** - To detect if avatars are clickable
- **ActionList** - To detect interactive items
- Various other components that need to detect focusable children

Previous implementation problems:
1. **Recursive traversal** - Visited every node in subtree individually
2. **getComputedStyle per node** - Called expensive style calculation for every single node
3. **No early termination optimization**

New implementation:
1. **Single querySelectorAll** - Browser-optimized selector matching
2. **Attribute-first checks** - Fast O(1) attribute lookup before style calc
3. **Early return** - Stops at first interactive node found

---

### DIFF
--- a/.changeset/perf-autocomplete-context-split.md
+++ b/.changeset/perf-autocomplete-context-split.md
@@ -1,0 +1,8 @@
+---
+'@primer/react': patch
+---
+
+perf(Autocomplete): Split context to reduce unnecessary re-renders
+
+Split AutocompleteContext into separate contexts for static values, setters, and dynamic state.
+Components now subscribe only to the context slices they need, reducing re-renders.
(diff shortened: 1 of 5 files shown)


francinelucca:
@copilot can you add tests for hasInteractiveNodes

francinelucca (approved):
✨

PR #7347, opened 2025-12-16
Only add the loading id when loading is shown

owenniblock:
Closes https://github.com/github/copilot-issues-xp/issues/309

#### Changed

Only include the id for the loading state element when the component is in the loading state.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Testing with storybook

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Button/__tests__/Button.test.tsx
```diff
@@ -177,6 +177,8 @@ describe('Button', () => {
     )
     const buttonNode = container.getByRole('button')
 
+    fireEvent.click(buttonNode)
```
owenniblock, line 173: The test component shows the loading element after the button has been clicked.

siddharthkp (approved):
Sounds sensible!

I'd recommend [running the integration tests](https://github.com/primer/react/pull/7347#issuecomment-3659866213) just to make sure there isn't a test in dotcom that expects aria-describedby to always exist

owenniblock:
@siddharthkp looks like there's an issue with this test: https://github.com/github/github-ui/blob/main/packages/mergebox/components/sections/__tests__/ConflictsSection.test.tsx#L600 - although looking at the logic, I can't see why it's failing 🤔

siddharthkp:
> although looking at the logic, I can't see why it's failing 🤔

Same, reading the test, it shouldn't fail because of this PR. I'd recommend running it once in codespaces as well for the branch and for main, just to be sure it's related.

owenniblock:
OK - it's related. It looks like the waitFor behaves differently when the `id` is set by the element referenced by the `id` isn't in the DOM 🤯 basically it seems to wait for a bit to see if the DOM element appears.

Whereas the check when there is _no_ `aria-describedby` instantly fails and goes around the loop for another try 😅 

The fix is to change the assertion so it's not doing a count (which seems pretty likely to be flaky anyway).

I see that this has already been done [in another test in the same file](https://github.com/github/github-ui/blob/prc-integration-test-for-7347/packages/mergebox/components/sections/__tests__/ConflictsSection.test.tsx#L549)

Testing [with a change on the validation PR](https://github.com/github/github-ui/pull/8932/commits/a42432e300b2179d1c1e21380ed71cde9c37b3b5).

How would we normally handle changes like this? I could cherrypick the fix into github before we merge my original primer react PR since it'll still pass for the old version of Primer?

owenniblock:
@siddharthkp there were some legitimate test failures as well which I've added fixes for on that PR. Assuming the PR is green I'll create a clean dotcom PR so we can get this PR moving 😄

siddharthkp:
> How would we normally handle changes like this? I could cherrypick the fix into github before we merge my original primer react PR since it'll still pass for the old version of Primer?

We don't have any rules around that (yet?). I prefer to do exactly what you said, prepare github for the change and then make the primer change.

owenniblock:
@siddharthkp [integration test results are all green](https://github.com/github/github-ui/pull/9150) 🎉 Think this is ready to go 😄

siddharthkp:
> @siddharthkp [integration test results are all green](https://github.com/github/github-ui/pull/9150) 🎉 Think this is ready to go 😄

Amazing! I see is also merged https://github.com/github/github-ui/pull/9052. I've already approved the PR, so you can merge whenever you'd like

--

github-ui checks are still waiting

I see the required checks are still waiting and won't let you merge. That's because the integration PR in github-ui does not have the latest commit from this primer/react PR, which is a bit annoying. I'll run it again so that it can pull in the latest commit and let you through

siddharthkp:
> 'll run it again so that it can pull in the latest commit and let you through

Done! Merge away!

owenniblock:
Yay, thanks @siddharthkp - merging 🎉

PR #7348, opened 2025-12-16
Add `onResizeEnd` and `currentWidth` props to PageLayout.Pane for controlled resizable width

### LINKED ISSUE #7311: PageLayout resizable is not SSR Safe
### Description

TLDR: 

When a resizable PageLayout is used in SSR (and has a nondefault width stored), is always causes a react hydration error and potential tearing, due to the read of localStorage in `useState` initialization.

---

When a resizable PageLayout is used the current size is set in localStorage

https://github.com/primer/react/blob/335e9e8eedc23de3041656acffb4873cd007f698/packages/react/src/PageLayout/PageLayout.tsx#L626-L634

On initial render that is read and used to seed the state

https://github.com/primer/react/blob/335e9e8eedc23de3041656acffb4873cd007f698/packages/react/src/PageLayout/PageLayout.tsx#L610-L624

In CSR scenarios. this is ok, albeit not ideal.  React should avoid interacting with DOM apis during the render cycle - even at state initialization, instead syncing them in useLayoutEffect or useEffect.

This would look like:

```js
const [width, setWidth] = useState(defaultValue)
useLayoutEffect(() => {
   const initial = localStorage.getItem(...)
   // validate
   setWidth(initial)
}, [])

<div style={{ --pane-width: width }}>...</div>
```

In this scenario, we'd reinitialize this value before the browser paints, and generally avoid flashing invalid sizes (as long as it's a layout effect)

However, in SSR scenarios, because the server doesn't have a way to really generate the default width here, we intiially render with the server default width, then on client hydration react receives different values for the width here - since it gets initialized differnelty from the server. 

This causes react to warn about hydration errors, and act non-deterministically in what the user sees.
- if the component doesn't re-render itself, users will see the server applied width only - because react won't sync this
- if the componet does re-render, react will 'correct' the invalid server value, possibly leading to CLS or a flash change of location
- in suspense boundaries, react may 'correct' this already, but outside of suspense it won't
  

You can see this on any page in GitHub that uses a resizable panel and SSRs  - the wrong resizable size is shown on reloads, even though react seems to know what it is - and then the resizer (today) will jump, since position is based on width + cursor-diff (https://github.com/primer/react/pull/7307 will change that to align with cursor position so it won't be an issue). However, it can also be seen as a jump resize much less frequently if something forces a PageLayout to render unexpectedly.

---

We should consider how resizable panes work, store values and seed them to enhance SSR safety and ensure that clients display the proper values.

Maybe this is - let's offload all of the localStorage work for this to the consumers, who can set/get widths however they choose, instead of relying on the internal implementation of this?

In this case we might still need to read some width value from a layout effect or something else to sync them properly?

### Steps to reproduce

SSR and hydrateRoot a resizable pane
Resize it
Reload the page
See hydration errors and incorrect width OR CLS

### Version

v35.x

### Browser

Chrome

mattcosta7:
## Summary

Adds `onResizeEnd` and `currentWidth` props to `PageLayout.Pane` for controlled resizable width, addressing #7311.

### Problem

The current `resizable={true}` implementation always persists to localStorage. Users have no way to:
1. Use custom persistence (server-side, IndexedDB, sessionStorage, etc.)
2. Control the pane width from React state
3. Avoid the SSR hydration mismatch caused by localStorage reads during render

### Solution

Add two new props (only available when `resizable={true}`):

| Prop | Type | Description |
|------|------|-------------|
| `onResizeEnd` | `(width: number) => void` | **Required for controlled mode.** Callback fired when resize ends (pointer release or keyboard key up). Replaces localStorage persistence. |
| `currentWidth` | `number` | Current displayed width in pixels. Overrides internal state. The `width` prop still defines the default used on reset (double-click). Requires `onResizeEnd`. |

### API Design

**Discriminated union types** enforce valid prop combinations at the type level:

```tsx
// ✅ Default behavior — localStorage persistence (unchanged)
<PageLayout.Pane resizable />

// ✅ Fully controlled — currentWidth + onResizeEnd
<PageLayout.Pane resizable currentWidth={width} onResizeEnd={setWidth} />

// ❌ Type error — currentWidth requires onResizeEnd
<PageLayout.Pane resizable currentWidth={width} />

// ❌ Type error — these props require resizable={true}
<PageLayout.Pane currentWidth={width} onResizeEnd={setWidth} />
```

**Three union branches** on `PageLayoutPaneProps`:

| Branch | `resizable` | `onResizeEnd` | `currentWidth` |
|--------|-------------|---------------|----------------|
| Default persistence | `true` | `never` | `never` |
| Custom callback | `true` | required | optional |
| Not resizable | `false` / omitted | `never` | `never` |

### Key behaviors

- **`onResizeEnd` without `currentWidth`**: The hook manages width internally but calls `onResizeEnd` instead of writing to localStorage. Good for fire-and-forget persistence (e.g., save to server).
- **`currentWidth` + `onResizeEnd`**: Fully controlled. The consumer owns the width state. `width` prop still defines constraints (min/max) and the reset value for double-click.
- **Values are rounded**: `saveWidth` rounds to integer pixels before persisting or calling `onResizeEnd` — sub-pixel floats from pointer drags are meaningless for persistence.
- **SSR**: Default `resizable={true}` uses localStorage which can cause a hydration mismatch (`suppressHydrationWarning` is only applied in this case). Using `onResizeEnd` avoids localStorage reads entirely, so no mismatch occurs and suppression is not needed.

### New export

- `defaultPaneWidth` — `{small: 256, medium: 296, large: 320}` — useful for initializing controlled width state.

### Example: Custom persistence (avoids SSR CLS)

```tsx
import {PageLayout, defaultPaneWidth} from '@primer/react'

function MyPage() {
  const [width, setWidth] = useState(defaultPaneWidth.medium)

  const handleResizeEnd = (newWidth: number) => {
    setWidth(newWidth)
    myServerStorage.save('pane-width', newWidth)
  }

  return (
    <PageLayout>
      <PageLayout.Pane
        resizable
        currentWidth={width}
        onResizeEnd={handleResizeEnd}
        aria-label="Side pane"
      >
        {/* ... */}
      </PageLayout.Pane>
      <PageLayout.Content>{/* ... */}</PageLayout.Content>
    </PageLayout>
  )
}
```

### Files changed

| File | Change |
|------|--------|
| `PageLayout.tsx` | Discriminated union types, pass new props to `usePaneWidth`, scoped `suppressHydrationWarning` |
| `usePaneWidth.ts` | `onResizeEnd`/`currentWidth` params, inline state sync, pixel rounding, `localStoragePersister` extraction |
| `usePaneWidth.test.ts` | 54 tests covering controlled mode, callback, fallbacks |
| `PageLayout.features.stories.tsx` | 4 new stories (without persistence, custom persistence, number width, controlled width) |
| `PageLayout.docs.json` | New props + story IDs, SSR note on `resizable` |
| `SplitPageLayout.docs.json` | New props (shares `PageLayoutPaneProps`) |
| `index.ts` | Exports `defaultPaneWidth` |
| Changeset | Minor bump for `@primer/react` |

### Testing

- **54 unit tests** for `usePaneWidth` — initialization, controlled mode, callback, localStorage, resize listener
- **15 PageLayout component tests** — all pass
- TypeScript compilation clean
- ESLint + Prettier clean

Fixes #7311

### DIFF
--- a/packages/react/src/PageLayout/PageLayout.features.stories.tsx
+++ b/packages/react/src/PageLayout/PageLayout.features.stories.tsx
@@ -361,22 +361,26 @@ export const WithCustomPaneHeading: StoryFn = () => (
   </PageLayout>
 )
 
-export const ResizablePaneWithoutPersistence: StoryFn = () => (
-  <PageLayout>
-    <PageLayout.Header>
-      <Placeholder height={64} label="Header" />
-    </PageLayout.Header>
-    <PageLayout.Pane resizable aria-label="Side pane">
-      <Placeholder height={320} label="Pane (resizable, not persisted)" />
-    </PageLayout.Pane>
-    <PageLayout.Content>
-      <Placeholder height={640} label="Content" />
-    </PageLayout.Content>
-    <PageLayout.Footer>
-      <Placeholder height={64} label="Footer" />
-    </PageLayout.Footer>
-  </PageLayout>
-)
+export const ResizablePaneWithoutPersistence: StoryFn = () => {
+  const [currentWidth, setCurrentWidth] = React.useState<number>(defaultPaneWidth.medium)
+
+  return (
+    <PageLayout>
+      <PageLayout.Header>
+        <Placeholder height={64} label="Header" />
+      </PageLayout.Header>
+      <PageLayout.Pane resizable currentWidth={currentWidth} onResizeEnd={setCurrentWidth} aria-label="Side pane">
+        <Placeholder height={320} label={`Pane (resizable, not persisted, width: ${currentWidth}px)`} />
+      </PageLayout.Pane>
+      <PageLayout.Content>
+        <Placeholder height={640} label="Content" />
+      </PageLayout.Content>
+      <PageLayout.Footer>
+        <Placeholder height={64} label="Footer" />
+      </PageLayout.Footer>
+    </PageLayout>
+  )
+}
 ResizablePaneWithoutPersistence.storyName = 'Resizable pane without persistence'
 
 export const ResizablePaneWithCustomPersistence: StoryFn = () => {
--- a/packages/react/src/PageLayout/usePaneWidth.ts
+++ b/packages/react/src/PageLayout/usePaneWidth.ts
@@ -419,7 +419,7 @@ export function usePaneWidth({
       endResizeOptimizations()
       window.removeEventListener('resize', handleResize)
     }
-  }, [resizable, customMaxWidth, minPaneWidth, paneRef, handleRef])
+  }, [resizable, customMaxWidth, minPaneWidth, paneRef, handleRef, contentWrapperRef])
 
   return {
     currentWidth,


mattcosta7:
@copilot

We received some feedback on this 

tiny API note: how do you feel about the idea of persist: false | 'localStorage' | fn where the default is localStorage for now

Let's open a new pr based on this one that tweaks the resizable API in this way

mattcosta7:
@copilot update changeset and pr description based on the latest changes

mattcosta7:
@copilot let's also open a separate pr for a breaking change option that removes the local storage read during render and streamlines configuration too. 

## Breaking API Changes PR

### Before → After

| Current | Proposed |
|---------|----------|
| `width="medium"` | `defaultWidth="medium"` or `defaultWidth={296}` |
| `width={{min: '256px', default: '296px', max: '600px'}}` | `minWidth={256}` `maxWidth={600}` `defaultWidth={296}` |
| `resizable={{persist: fn, width: currentWidth}}` | `width={currentWidth}` `onWidthChange={fn}` |
| `resizable={{persist: 'localStorage', widthStorageKey: '...', width}}` | `width={width}` `onWidthChange={...}` (use hook) |
| `resizable={{persist: false}}` | `resizable` (boolean, just enables drag) |
| `widthStorageKey` prop | Removed - use `useLocalStoragePaneWidth` hook |

### New API

```tsx
// Simple resizable with no persistence
<PageLayout.Pane 
  resizable 
  defaultWidth="medium"
/>

// Controlled width with custom persistence
const [width, setWidth] = useState(defaultPaneWidth.medium)
<PageLayout.Pane
  resizable
  width={width}
  defaultWidth="medium"
  minWidth={256}
  maxWidth={600}
  onWidthChange={setWidth}
/>

// localStorage persistence via hook
const [width, setWidth] = useLocalStoragePaneWidth('my-pane-key', {
  defaultWidth: defaultPaneWidth.medium,
  minWidth: 256,
})
<PageLayout.Pane
  resizable
  width={width}
  onWidthChange={setWidth}
/>
```

### New Exports

```tsx
// Hook for localStorage persistence (SSR-safe)
export function useLocalStoragePaneWidth(
  key: string,
  options: {
    defaultWidth: number
    minWidth?: number
    maxWidth?: number
  }
): [number, (width: number) => void]

// Preset width values (already exported)
export const defaultPaneWidth: Record<'small' | 'medium' | 'large', number>
```

### Benefits

1. **Standard React patterns** - `width` + `onWidthChange` is idiomatic controlled component API
2. **Simpler types** - No union types for width prop, separate concerns into separate props
3. **Composable** - `useLocalStoragePaneWidth` can be used independently or with other storage
4. **SSR-safe by default** - Hook handles hydration internally
5. **Clearer mental model** - `resizable` is just a boolean, persistence is separate

### Migration Path

```tsx
// Before
<PageLayout.Pane 
  width={{min: '256px', default: '296px', max: '600px'}}
  resizable={{persist: 'localStorage', widthStorageKey: 'my-key', width: currentWidth}}
/>

// After
const [width, setWidth] = useLocalStoragePaneWidth('my-key', {
  defaultWidth: 296,
  minWidth: 256,
  maxWidth: 600,
})
<PageLayout.Pane
  resizable
  width={width}
  defaultWidth={296}
  minWidth={256}
  maxWidth={600}
  onWidthChange={setWidth}
/>
```

mattcosta7:
@copilot

instead of making resizable configurable, we're going to just add a new prop `onWidthChange`.
When not resizable this is not called
When resizable, if one is passed we call onWidthChange instead of save to localStorage.`false` could opt out of saving/reading from localstorage entirely. undefined maintains the local storage access we have today.

we also need to add a `currentWidth` to the api - either for `width` whn custom or when given a known width.

Where are good places to add this so consumers can handle initialization

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -783,7 +846,7 @@ const Pane = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayout
                 }
           }
           // If pane is resizable, the divider should be draggable
-          draggable={resizable}
+          draggable={!!resizable}
```
mattcosta7:
```suggestion
          draggable={resizable}
```

siddharthkp:
Because me and Matt discussed this change before this PR, going to request a review from @francinelucca for an additional set of eyes

While the component could definitely use a round of improvements on the API, we thought this was a good way of solving the problem without blowing up scope.

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -624,6 +639,50 @@ export type PageLayoutPaneProps = {
...
+         * Uses localStorage persistence by default unless `onResizeEnd` is provided.
+         *
+         * Note: With default localStorage persistence in SSR, the server-rendered
+         * width may differ from the stored client width, causing a brief layout
+         * shift on hydration. Use `onResizeEnd` with server-aware storage to avoid this.
+         */
+        resizable: true
+        currentWidth?: never
+        onResizeEnd?: never
+      }
```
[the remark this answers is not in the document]
mattcosta7: fixed in later commits

packages/react/src/PageLayout/PageLayout.docs.json
```diff
@@ -172,13 +184,23 @@
...
+          "name": "onResizeEnd",
+          "type": "(width: number) => void",
+          "description": "Callback fired when a resize operation ends (drag release or keyboard key up). When provided, replaces localStorage persistence. Requires `currentWidth`."
         },
         {
           "name": "widthStorageKey",
           "type": "string",
           "defaultValue": "'paneWidth'",
-          "description": "Provide a key used by localStorage to persist the size of the pane on the client."
+          "description": "localStorage key used to persist the pane width across sessions. Only applies when `resizable` is `true` and no `onResizeEnd` callback is provided."
```
francinelucca, line 203: is the plan still to get rid of localStorage next major? if so I think we can deprecate this prop

packages/react/src/PageLayout/PageLayout.features.stories.tsx
```diff
@@ -358,3 +360,171 @@ export const WithCustomPaneHeading: StoryFn = () => (
     </PageLayout.Footer>
   </PageLayout>
 )
+
+export const ResizablePaneWithoutPersistence: StoryFn = () => (
+  <PageLayout>
+    <PageLayout.Header>
+      <Placeholder height={64} label="Header" />
+    </PageLayout.Header>
+    <PageLayout.Pane resizable aria-label="Side pane">
```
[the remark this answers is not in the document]
francinelucca: I think copilot is right here 👀

packages/react/src/PageLayout/PageLayout.docs.json
mattcosta7: @siddharthkp  do y'all intend to get rid fo that default, or maybe control it differnetly?

I'm not sure we need to get rid of it, but I imagine the api might be opt in with some server rendering logic somewhere?

mattcosta7: I'll probably avoid deprecating that here, and in the future it might be something that gets handled?

siddharthkp: I'd like to eventually and instead export a `persistInLocalStorage` function that you can pass to `onResizeEnd` but it doesn't have to be now

A tracking issue would be very helpful though, so we can get it on the backlog

PR #7391, opened 2025-12-29
Add existing "All combinations" actionlist storybook to vrt

barmo:
Closes #na

#### New

Added existing [Actionlist All Combinations example storybook](https://primer.style/react/storybook/?path=/story/components-actionlist-examples--all-combinations) to playwright tests which then also added the new screenshot files.

Motivation is so that I can add this "The everything bagel" test and show the snapshot diff with the inline description overflow fixed in another pr. The overflow fix is for this issue #https://github.com/github/memex/issues/23884

screenshot of the storybook example mentioned above

#### Changed

na

#### Removed

na

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
I assume that updating tests do not require a release, but let me know if otherwise

### Merge checklist

- [x] Added/updated tests
~~- [ ] Added/updated documentation~~
~~- [ ] Added/updated previews (Storybook)~~
~~- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)~~
~~- [ ] Tested in Chrome~~
~~- [ ] Tested in Firefox~~
~~- [ ] Tested in Safari~~
~~- [ ] Tested in Edge~~
~~- [ ] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))~~


barmo:
The snapshots were updated in my codespace with
```
npx playwright test ActionList @vrt --update-snapshots
```
since I'm adding an existing story to the vrt test. I would expect new snapshots files related to this change.

However, there are 100+ files snapshot files updated. When checking several of the updated snapshots, it seems like they're negligible pixel changes. I'm assuming that this difference is due to the difference in my local vs CI environment. Please let me know if this is correct and if I should manually remove the snapshots that already existed prior.

joshblack:
@barmo you're spot on, there will be some differences between the codespace and what is running in CI. If you want to test locally but have them be in sync you can use `script/test-e2e` which is using the same docker image as what is running in CI. There are a couple of options for this too over in: https://github.com/primer/react/blob/main/contributor-docs/testing.md#visual-regression-tests (most of them are just arguments passed along to playwright in the docker image)

Hope that makes sense, let me know if you have any questions!

barmo:
Just for vis, because my rebasing made the commit history wonkier than expected:
- edited my commit https://github.com/primer/react/pull/7391/commits/a134b39779bad32c4bae8cb5b2b5e0205043924e to change the test name from "All Combination" to "All combination" because I later noticed that there already exists snapshot files with that casing (e.g. [ActionList-All-combinations-dark-colorblind-hover-linux.png](https://github.com/primer/react/blob/main/.playwright/snapshots/components/ActionList.test.ts-snapshots/ActionList-All-combinations-dark-colorblind-hover-linux.png))
- removed all my snapshot file changes that was a result from me locally running the `npx playwright test ActionList @vrt --update-snapshots` command and will let the "update snapshots" label handle the the updates

PR #7427, opened 2026-01-06
Change actionlist item inline description styling

barmo:
Closes https://github.com/github/memex/issues/23884

#### New

na

#### Changed

Actionlist item inline description styling change from flex to block for handling overflow to be on the next line below the label text. Change can be seen with "The everything bagel" example ([prod storybook](https://primer.style/react/storybook/?path=/story/components-actionlist-examples--all-combinations))

| Before | After |
|--------|--------|
| Screenshot of example before the code change where both the label and inline description overflows to the next line within their component bounds | Screenshot of example after the code change where only the inline description overflows to the next line | 

This overflow behavior is consistent with the reviewers for pull requests for example
screenshot of example mentioned above

#### Removed

na

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Planning to go to OH today to ask more about this, but I added the "All combination" storybook in the vrt in a [different pr](https://github.com/primer/react/pull/7391) so that this change can be better documented/visible for review via the vrt. Will update later today

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


barmo:
Here is a [loom recording](https://www.loom.com/share/249955e853d44b82af271ed612ee6ea6) of the before (prod) and after (deploy preview) with the changes of this pr on dotcom

[deploy preview link](https://github.com/?_ui_sha=099a3459e2de9119d12c54aa527996075f48197d) (expires ~6pm CT)

Otherwise here's a table with screenshots testing this ui fix in two different places
| | Before | After |
|--------|--------|--------|
| Selector in Projects table | screenshot of inline description does not overflow below the label | screenshot of inline description does overflow below the label |
| Selector within issues | screenshot of inline description does not overflow below the label|  creenshot of inline description does overflow below the label | 

**EDIT**
The loom video is not up to date (the vertical spacing is a smidge too big) with this commit https://github.com/primer/react/pull/7427/commits/9f6ac86d79a5feb4ecac7a927d130494f671653b. The deploy preview and the screenshots ARE up to date with the latest code changing commit https://github.com/primer/react/pull/7427/commits/25531d8782d80ba081281a6353586e1f9c1c3104

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -628,27 +628,24 @@ default block */
...
     position: relative;
+    display: block;
     word-break: normal;
-    flex-direction: row;
-    align-items: baseline;
-    gap: var(--base-size-8);
+    line-height: normal;
 
     & .ItemLabel {
+      display: inline;
```
devinmcinnis: Can this be removed? I'm only looking at Storybook but it looks like it's rendered as a `<span>` which is already `display: inline`

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -628,27 +628,24 @@ default block */
   /* inline */
   &:where([data-description-variant='inline']) {
     position: relative;
+    display: block;
     word-break: normal;
-    flex-direction: row;
-    align-items: baseline;
-    gap: var(--base-size-8);
+    line-height: normal;
```
devinmcinnis: Do we need this or can remove? Is `line-height: normal` the default?

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -628,27 +628,25 @@ default block */
   /* inline */
   &:where([data-description-variant='inline']) {
     position: relative;
+    display: block;
     word-break: normal;
-    flex-direction: row;
-    align-items: baseline;
-    gap: var(--base-size-8);
+    /* stylelint-disable-next-line primer/typography */
+    line-height: normal;
```
barmo, line 634: **Setting the `line-height` was necessary** because with flex previously, the container height is determined by the tallest item's content. With the `display:block`, the block/inline layout inherits line-height from ancestor elements, which can vary depending on context and create extra vertical spacing. We need to reset this to prevent the spacing mismatch.

**Including the disable instead of using a Primer typography variable was necessary** because
- There's no Primer variable that represents a "reset" or "normal" line-height
- Using explicit values like 20px or var(--control-medium-lineBoxHeight) didn't work consistently across all contexts - some VRT tests passed while others failed
- line-height: normal lets the browser calculate the natural line-height based on the font metrics, which produces consistent results that match the original flex baseline alignment behavior across all usage contexts

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -628,27 +628,25 @@ default block */
...
 
     & .ItemLabel {
+      display: inline;
```
devinmcinnis: Can this be removed? I'm only looking at Storybook but it's rendered as a `<span>` which is already `display: inline`. 

Same with `.Description` below

barmo: you are totes correct my dude

PR #7438, opened 2026-01-09
feat(Spinner): make delay customizable

francinelucca:
Closes https://github.com/github/primer/issues/6288

#### Changed

- broadened Spinner's `delay` prop to accept a number for further customization

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Spinner/Spinner.tsx
```diff
@@ -24,7 +24,7 @@ export type SpinnerProps = {
   className?: string
   style?: React.CSSProperties
   /** Whether to delay the spinner before rendering by the defined 1000ms. */
-  delay?: boolean
+  delay?: boolean | 'default' | number
```
francinelucca: @siddharthkp curious on your thoughts here since this implementation was your idea. I didn't feel confident removing the boolean type initially because that felt like a major change, so I just added the other two. I'm thinking we can go back and cleanup the boolean type on v39?

adierkens: Is the `'default'` option even needed? Could we just use `true` for `1s` or a `number` for a custom value?

siddharthkp: Agree with this ^, true = default

francinelucca: yeah, def can. I was just thinking what the API would look like if we were building it from scratch with the requirements we have now ('default', number), but happy to defer to boolean, number instead!

siddharthkp: Thinking from scratch, I would have said delay = `short` | `long` and not number. 

But, seeing how we don't have strong opinions on what those numbers should be and there is only 1 instance of `delay` in dotcom, I'm happy to open it up to being a `number` and then codifying a few values based on usage in the future.

Optional side note: Does `delayMs: number` sound better? Works if we silently remove `delay: boolean` 😅

packages/react/src/Spinner/Spinner.docs.json
```diff
@@ -33,9 +33,9 @@
     },
     {
       "defaultValue": "false",
-      "description": "Whether to delay the spinner before rendering by the defined 1000ms.",
+      "description": "Controls whether and how long to delay rendering the spinner. Set to `true` to delay by 1000ms, `'short'` to delay by 300ms, `'long'` to delay by 1000ms, or provide a custom number of milliseconds.",
       "name": "delay",
-      "type": "boolean"
+      "type": "boolean | 'short' | 'long' | number"
```
francinelucca, line 38: boolean will be going away after some dotcom updates, to align with [SkeletonBox](https://github.com/primer/react/pull/7448)

PR #7451, opened 2026-01-15
Handle React elements in UnderlineNav.Item children

daantosaurus:
- Closes https://github.com/github/security-products-enablement/issues/2426

Hello Primer reviewer 👋, this is my first time contributing to the repo, so please let me know if there's anything I can do better in terms of how I'm contributing.

This change stems from an initiative that involves adding popovers to the React global nav bar to prompt users to enable GHAS (example [issue](https://github.com/github/enterprise-and-digital-experiences/issues/185)). I've already attempted to implement these popovers, but they result in a spacing issue stemming from `UnderlineNav`. See example below:

Screenshot 2026-01-14 at 11 43 48 PM

I believe that fixing this to allow better popover (or any other component that people can think of) support on the nav bar could open up the doors for other teams to add their own overlays to nav bar tabs and help them achieve their business needs!

See the linked issues above if needed for more context! I dive deeper into the root cause below

### Changelog

The `UnderlineNav.Item` component accepts `React.ReactNode` as children but the internal implementation assumes children will always be a plain string. This causes layout issues when valid React elements are passed alongside the label text.

The `data-content` attribute which is used by CSS to reserve space for bold text in a nav item (in order to prevent layout shift) would break when the children included React elements, as they can't be properly serialized into a string. To fix this, I added a helper function that extracts only direct text strings from children in order to ignore React elements

Before, if a string and a ReactNode were passed in as children, `data-content` would evaluate to something like `data-content="someString[object Object]"`. After this change, it would correctly evaluate to `data-content="someString"`

#### New

- Added `WithPopover` story demonstrating Popover usage inside UNderlineNav.Item

image

#### Changed

- UnderlineNav.Item now extracts text content for `data-content` when children contain React elements

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/UnderlineNav/UnderlineNav.features.stories.tsx
+++ b/packages/react/src/UnderlineNav/UnderlineNav.features.stories.tsx
@@ -14,7 +14,6 @@ import {
 import type {Meta} from '@storybook/react-vite'
 import {UnderlineNav} from './index'
 import {INITIAL_VIEWPORTS} from 'storybook/viewport'
-import Popover from '../Popover'
 
 const meta = {
   title: 'Components/UnderlineNav/Features',
@@ -152,38 +151,6 @@ export const VariantFlush = () => {
       <UnderlineNav.Item aria-current="page">Code</UnderlineNav.Item>
       <UnderlineNav.Item>Issues</UnderlineNav.Item>
       <UnderlineNav.Item>Pull Requests</UnderlineNav.Item>
-    </UnderlineNav>
-  )
-}
-
-export const WithPopover = () => {
-  return (
-    <UnderlineNav aria-label="Repository">
-      <UnderlineNav.Item href="#code" leadingVisual={<CodeIcon />}>
-        Code
-      </UnderlineNav.Item>
-      <UnderlineNav.Item href="#issues" leadingVisual={<IssueOpenedIcon />}>
-        Issues
-      </UnderlineNav.Item>
-      <UnderlineNav.Item href="#security" leadingVisual={<ShieldLockIcon />} counter={12}>
-        Security
-        <Popover
-          open
-          style={{
-            left: '50%',
-            transform: 'translateX(-50%)',
-            marginTop: 'var(--base-size-12)',
-          }}
-        >
-          <Popover.Content>Popover content</Popover.Content>
-        </Popover>
-      </UnderlineNav.Item>
-      <UnderlineNav.Item href="#insights" leadingVisual={<GraphIcon />}>
-        Insights
-      </UnderlineNav.Item>
-      <UnderlineNav.Item href="#settings" leadingVisual={<GearIcon />}>
-        Settings
-      </UnderlineNav.Item>
     </UnderlineNav>
   )
 }


packages/react/src/internal/components/UnderlineTabbedInterface.tsx
```diff
@@ -13,6 +13,20 @@ import {clsx} from 'clsx'
 // The gap between the list items. It is a constant because the gap is used to calculate the possible number of items that can fit in the container.
 export const GAP = 8
 
+// Helper to extract direct text content from children for the data-content attribute.
```
hectahertz: I feel slightly uncomfortable with the children parsing here but I get why you would approach it like this.

If you can think of any other solution that would avoid this it would probably be more future proof. Feel free to propose api changes if they're non breaking!

packages/react/src/UnderlineNav/UnderlineNav.features.stories.tsx
```diff
@@ -154,3 +155,35 @@ export const VariantFlush = () => {
...
+    <UnderlineNav aria-label="Repository">
+      <UnderlineNav.Item href="#code" leadingVisual={<CodeIcon />}>
+        Code
+      </UnderlineNav.Item>
+      <UnderlineNav.Item href="#issues" leadingVisual={<IssueOpenedIcon />}>
+        Issues
+      </UnderlineNav.Item>
+      <UnderlineNav.Item href="#security" leadingVisual={<ShieldLockIcon />} counter={12}>
+        Security
+        <Popover
```
siddharthkp: Ran into this on slack, so leaving some unsolicited notes:  

There are a few things to consider here: 
1. Is the rendered DOM still accessible?
2. When the item with popover is inside the overflow "More" menu, does it still position the popover correctly?
3. Can the API for this be better? It feels a bit strange to put popover content inside the underline item. Is it already possible to put the Popover outside the UnderlineNav but still get it to position on the right anchor? Alternatively, should Popover take an `anchorRef` to attach itself to?

  I'll leave these with @hectahertz :)

daantosaurus: Ah yeah I see what you mean especially with point #3. I personally had a hard time trying to nicely position the popover on the NavItems while also rendering it outside of the UnderlineNav, so my mind is moving towards the idea of having Popover take an `anchorRef` or just using a custom `AnchoredOverlay` in github-ui where I add a custom caret to simulate the popover experience (since AnchoredOverlay already takes anchorRefs). Let me take this back to the drawing board and I'll keep in touch

In the meantime, do yall want me to continue with the fix in this PR? I can update the PR to keep the fix and I'll just remove the story I added. Otherwise I can just close it

daantosaurus: Oh, I should also ask yall for the recommended guidance to achieve this popover experience on the nav bar 😅. What do you think between the custom `AchoredOverlay` approach with a caret vs adding an `anchorRef` to Popover? Imo adding an anchorRef to popover makes sense but I'm far from the experienced one here

TylerJDev: This actually came up recently for PRs usage of a similar component ([slack thread](https://github.slack.com/archives/CSGAVNZ19/p1768338012741849)). It seems like Popover should ideally carry some functionality that `AnchoredOverlay` already has. 

> What do you think between the custom AchoredOverlay approach with a caret vs adding an anchorRef to Popover? Imo adding an anchorRef to popover makes sense but I'm far from the experienced one here

It _may_ be easier to utilize `AnchoredOverlay` here, but I'd say go with whichever path is easier for you to implement. There might be some improvements we can make to both components, so we'd be happy for any feedback!

This PR still fixes the issue where we aren't correctly setting the `data-*` attribute, so we should still ship this! Thanks for the PR! 😁

PR #7468, opened 2026-01-22
feat(ActionList): update loading styles for specific child elements

jonrohan:
Part of https://github.com/github/pull-requests/issues/22213

## Summary

Optimizes CSS selector performance in ActionList by replacing expensive universal selectors (`*`) and broad `:has()` queries with targeted class selectors. `:is(.ActionListItem:not(:has([aria-disabled], [disabled]), [data-has-subitem="true"]):where([data-loading="true"]), .ActionListItem:not(:has([aria-disabled], [disabled]), [data-has-subitem="true"]):has([data-loading="true"])) *`

### Problem

During LCP (Largest Contentful Paint) analysis, we identified a selector in ActionList that was attempting to match ~200k elements but matching none:

The issues were:
1. **Universal selector (`& *`)** - Forces the browser to check every descendant element
2. **Broad `:has([data-loading='true'])`** - Scans all descendants looking for the `data-loading` attribute

### Solution

1. **Replaced `& *` with explicit class targets:**
   - `.ItemLabel`
   - `.Description`
   - `.LeadingVisual`
   - `.TrailingVisual`
   - `.VisualWrap`

2. **Removed `:has([data-loading='true'])` entirely** - After investigation, `data-loading` is set directly on the `<li>` element via `Item.tsx`, so the `:has()` variant was unnecessary. The separate rule at line 369 already handles TrailingAction-specific loading states.

## Before/After

| Aspect | Before | After |
|--------|--------|-------|
| Loading state selector | `&:where([data-loading='true']), &:has([data-loading='true']) { & * { ... } }` | `&:where([data-loading='true']), & [data-loading='true'] { & .ItemLabel, & .Description, ... { ... } }` |
| Elements checked | ~200k (all descendants) | Only elements with specific classes |
| `:has()` scope | All descendants | Direct child with `data-loading` attribute |

## Testing & Verification

- [x] Build passes (`npm run build`)
- [x] Unit tests pass (`npm test` - 505 tests)
- [x] CSS linting passes (`npx stylelint`)
- [ ] Verified loading state styling in Storybook
- [ ] Tested with ActionList items using `loading` prop
- [ ] Tested with TrailingAction containing loading buttons

## Merge checklist

- [x] Added/updated tests
- [ ] Tested in all supported browsers
- [x] Reviewed for accessibility impact (no changes to accessibility behavior)

## Screenshots

N/A - This is a performance optimization with no visual changes.

### DIFF
--- a/packages/react/src/ActionList/ActionList.module.css
+++ b/packages/react/src/ActionList/ActionList.module.css
@@ -282,6 +282,7 @@
       & .Description,
       & .LeadingVisual,
       & .TrailingVisual,
+      & .LeadingAction,
       & .VisualWrap {
         color: var(--fgColor-muted);
       }


packages/react/src/ActionList/ActionList.module.css
```diff
@@ -278,7 +278,11 @@
 
     &:where([data-loading='true']),
     &:has([data-loading='true']) {
```
mattcosta7: is the `[data-loading]` really at any level of depth, or is it possibly on the `ListItem` wrappers directly?  that would probably be another nice win

jonrohan: Theres is a .TrailingAction data-loading that can happen, but we can make the has more specific here to avoid scanning all the descendants.

Issue #7493, opened 2026-02-02
Multiple hooks/components lead to hydration errors


mattcosta7:
### Description

Multiple hooks and components are guaranteed to cause hydration errors post SSR

Any components that use 

## `<Portal>` (and friends)

On the server `<Portal>` returns null. On the client `<Portal />` returns a dom node, which leads to `createPortal()` being called in initial render. This leads to the dom trees in client/server renders differing causing at least `useId` drift, since `useId` relies on identical react trees on the server and client.

we could probably do something like 

```
Portal = () => {
    const isServer = useSyncExternalStore(noopSubscription, () => false, () => true)

   return !element || isServer ? null : createPortal(children, element)
}
```

It's not immediately clear what else would need to change to support this though.  I tried [a more thorough rewrite ](https://github.com/primer/react/pull/7461) that moved more of this into effects and that led to a cascade of issues in tests, and the simpler approach might be better (but also could be that I missed somehting simple)

Impacted components (at least):
- ConfirmationDialog
- ActionMenu
- AnchoredOverlay
- AutocompleteOverlay
- Dialog
- Overlay

 ## `useMedia`

`useMedia returns a boolean for matches, but `canUsDom` will always return true on the client and flase on the server leading to an initial state divergence. any rendering relying on `useMatches` output is possible to diverge.

we could also  do some `syncExternalStore` here and just default until hydration is over

## `FIlteredActionListLoaders`

These render random width skeletons, which means they have widths that change on every render. `Math.random` isn't stable across the client/server boundary.

We could `suppressHydration` here, but we could also define pseudo random initial values, and sync them to a random (but stateful/stable) value later - depends on how much variation matters and when
 

## `<PageLayout>`, 

This one is discussed here: https://github.com/primer/react/issues/7311
Moving this forward would be awesome, as well as avoid a CLS issue!

## `Maybe <ThemeProvider /> (not 100% sure, preventSSRMismatch may handle this well enough already`

we should investigate/validate this.  I think it's ok - but additional validation would be good (while we're reviewing other usage)

### Steps to reproduce

SSR + hydrateRoot

### Version

latest

mattcosta7:
Related: 

Portal - https://github.com/primer/react/issues/3721
PageLayout - https://github.com/primer/react/issues/7311

francinelucca:
TODO: separate into its own issues, mark as P2 (below a11y issues)

francinelucca:
[...]

PR #7494, opened 2026-02-02
ensure `useMedia` avoids hydration issues

mattcosta7:
Relates to #7493 
Closes https://github.com/github/primer/issues/6385

## Overview

This PR refactors the `useMedia` hook to use React's `useSyncExternalStore` API, which is specifically designed to safely subscribe to external stores (like browser media queries) while avoiding hydration mismatches between server and client rendering.

This is mostly defense in depth + gets rid of a post hydration render cycle when we useMedia with a default value - after hydration it's fine to intiialize with the query - useSyncExternalstore will handle this well for us and avoid an extra render pass

### Problem

The previous implementation used `useState` + `useEffect` to track media query changes. This pattern can cause hydration mismatches in SSR scenarios because:
1. The initial state is computed during render, which may differ between server and client
2. The `useEffect` callback runs after hydration, potentially causing a flash of incorrect content

#### Why `canUseDOM` in `useState` causes hydration issues

The root cause is how the `useState` initializer interacts with `canUseDOM`:

```tsx
const [matches, setMatches] = React.useState(() => {
  // ...
  if (canUseDOM) {
    return window.matchMedia(mediaQueryString).matches
  }
  return false
})
```

The `useState` initializer function runs **during the initial render** on both server and client:

- **Server-side render**: `canUseDOM` is `false` (no `window` exists) → state initializes to `false`
- **Client-side hydration**: `canUseDOM` is `true` → state initializes to `window.matchMedia(mediaQueryString).matches`

If the user's actual media query result is `true` (e.g., they have a coarse pointer, or their viewport is narrow), the client computes `true` while the server rendered `false`. React detects this mismatch during hydration because the initial client render doesn't match the server-rendered HTML, resulting in a hydration error.

Even when `defaultState` is provided, the subsequent `useEffect` that syncs state with `matchMedia` runs *after* hydration, which can still cause a visible flash of incorrect content.

### Solution

The new implementation:
- Uses `useSyncExternalStore` which React specifically designed for subscribing to external data sources with SSR support
- Provides a `getServerSnapshot` callback that returns the `defaultState` (or `false` with a warning) during server rendering
- React automatically uses `getServerSnapshot` during hydration to ensure consistency
- Simplifies the logic by separating context-based overrides from the actual media query subscription

#### New

- Introduced internal `useMediaQuery` helper that uses `useSyncExternalStore` for hydration-safe media query subscriptions

#### Changed

- `useMedia` now delegates to `useSyncExternalStore` instead of `useState` + `useEffect`
- Updated test mock to use a getter for `matches` property to properly support `useSyncExternalStore`'s snapshot pattern

#### Removed

- Removed manual state synchronization logic that was prone to hydration issues

### Rollout strategy

- [x] Patch release
- [ ] Minor release

This is a bug fix for hydration mismatches with no API changes to consumers.

### Testing & Reviewing

- Existing tests have been updated to work with `useSyncExternalStore`
- The mock now tracks `currentMatch` state and uses a getter so that `useSyncExternalStore` can correctly read the current value via `getSnapshot`
- Test SSR scenarios in Next.js examples to verify no hydration warnings appear

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/smooth-ladybugs-hunt.md
+++ b/.changeset/smooth-ladybugs-hunt.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Refactor `useMedia` hook to use `useSyncExternalStore` for improved hydration safety
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,7 +81,7 @@
         "react-dom": "^18.3.1"
       },
       "devDependencies": {
-        "@primer/react": "38.8.0",
+        "@primer/react": "38.9.0",
         "@primer/styled-react": "1.0.2",
         "@types/react": "^18.3.11",
         "@types/react-dom": "^18.3.0",
@@ -95,7 +95,7 @@
       "name": "example-nextjs",
       "version": "0.0.0",
       "dependencies": {
-        "@primer/react": "38.8.0",
+        "@primer/react": "38.9.0",
         "@primer/styled-react": "1.0.2",
         "next": "^16.1.5",
         "react": "^19.2.0",
@@ -138,7 +138,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@primer/octicons-react": "^19.21.0",
-        "@primer/react": "38.8.0",
+        "@primer/react": "38.9.0",
         "@primer/styled-react": "1.0.2",
         "clsx": "^2.1.1",
         "next": "^16.1.5",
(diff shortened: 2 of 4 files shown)


siddharthkp (approved):
Makes perfect sense to me. 

However, I don't confidence in our automated tests to test this properly. I'd recommend testing this in dotcom manually before merging just to make sure there are no surprises!

PR #7495, opened 2026-02-02
[MCP] Add support for the `/llms.txt` component endpoint

adierkens:
- Related: https://github.com/github/primer-docs/pull/897

With the new evals framework, we discovered that the existing Primer MCP tools for getting component documentation was insufficient. It would include HTML fragments, confusing formatting and other results that skewed the effectiveness. In one example it couldn't find the import path for a component via the docs. 

Running our evals framework for Primer experience generation with and without the new endpoint enabled, we see a 5x reduction in token usage for much more actionable content. 

#### New

- Expands the MCP tool to include pulling in content via the `<component>/llms.txt` endpoint. This is intended to serve as a more concise summary of the component documentation, tailed specifically for agent consumption. 

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -226,6 +226,58 @@
       "dev": true,
       "license": "ISC"
     },
+    "node_modules/@babel/cli": {
+      "version": "7.28.6",
+      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.28.6.tgz",
+      "integrity": "sha512-6EUNcuBbNkj08Oj4gAZ+BUU8yLCgKzgVX4gaTh09Ya2C8ICM4P+G30g4m3akRxSYAp3A/gnWchrNst7px4/nUQ==",
+      "dev": true,
+      "license": "MIT",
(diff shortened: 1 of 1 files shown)


francinelucca (approved):

packages/mcp/src/server.ts
```diff
@@ -691,13 +706,7 @@ server.registerTool(
     inputSchema: {
       surroundingText: z.string().describe('Text surrounding the image, relevant to the image.'),
       alt: z.string().describe('The alt text of the image being evaluated'),
-      image: z
-        .union([
-          z.instanceof(File).describe('The image src file being evaluated'),
-          z.url().describe('The URL of the image src being evaluated'),
-          z.string().describe('The file path of the image src being evaluated'),
```
adierkens, line 698: These were throwing errors during testing: `MCP error -32603: Custom types cannot be represented in JSON Schema` and they should all be string types in the generated JSON schema

PR #7463, opened 2026-01-21
Add failing test and potential fix for Underlinenav overflow issue

siddharthkp:
- Adds playwright mcp server
- Adds a page in examples/nextjs app to reproduce https://github.com/github/primer/issues/6291
- Hides overflow on SSR only to enable it on client

PR #7504, opened 2026-02-06
UnderlineNav: Add `overflow: hidden` when calculating items for overflow menu

TylerJDev:
Closes https://github.com/github/primer/issues/6291

Credits to @siddharthkp & @francinelucca for the original PR: https://github.com/primer/react/pull/7463 ✨  

#### Changed

* Adds `overflow: hidden` to container when loading items

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/internal/components/UnderlineTabbedInterface.module.css
```diff
@@ -12,6 +12,13 @@
   /* stylelint-disable-next-line primer/box-shadow */
   box-shadow: inset 0 -1px var(--borderColor-muted);
 
+  /* Hide overflow until calculation is complete to prevent CLS */
+  overflow: hidden;
+
```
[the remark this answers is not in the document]
TylerJDev: I think this is fair!

packages/react/src/internal/components/UnderlineTabbedInterface.module.css
```diff
@@ -12,6 +12,16 @@
   /* stylelint-disable-next-line primer/box-shadow */
   box-shadow: inset 0 -1px var(--borderColor-muted);
 
+  /* Hide overflow until calculation is complete to prevent CLS */
+  overflow: visible;
```
adierkens, line 16: I think we can do this entirely within css and avoid any of the `useEffect()` updates, just with something like: 

```css
.UnderlineWrapper {
  &:not(:has([class*='prc-UnderlineNav-MoreButton-'])) {
    overflow-x: hidden;
  }
}
```

You might need to update the classNames a bit (or move it to the `UnderlineNav` component directly)

siddharthkp: Very nice! Although probably the same thing perf wise because the MoreButton appears after a javascript pass as well, right?

Thinking out loud, do we need to test this for cases where there is an overflow because of icons and counters, so the MoreButton may never appear, or is that already covered?

packages/react/src/internal/components/UnderlineTabbedInterface.tsx
```diff
@@ -32,14 +32,18 @@ type UnderlineWrapperProps<As extends React.ElementType> = {
...
 
 export const UnderlineWrapper = forwardRef((props, ref) => {
-  const {children, className, as: Component = 'div', ...rest} = props
+  const {children, className, as: Component = 'div', ready, ...rest} = props
+
   return (
     <Component
       className={clsx(classes.UnderlineWrapper, className)}
       ref={ref as ForwardedRef<HTMLDivElement>}
+      data-ready={ready === undefined ? undefined : ready ? 'true' : 'false'}
```
siddharthkp: More descriptive name please. `data-hydrated` / `data-overflow-handled`?

siddharthkp (approved):
Left a few comments, approving in advance

.changeset/clever-geese-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+UnderlineNav: Adds `overflow: hidden` when calculating items to prevent CLS
```
siddharthkp:
```suggestion
UnderlineNav: Adds `overflow: hidden` during SSR prevent CLS, removes it after hydration on client.
```

packages/react/src/internal/components/UnderlineTabbedInterface.module.css
TylerJDev: > Thinking out loud, do we need to test this for cases where there is an overflow because of icons and counters, so the MoreButton may never appear,

Can the component still overflow without the menu? I haven't seen that, as I assumed that we also took those into account when determining whether to show the menu or not.

liuliu-dev (approved):
🎉

siddharthkp (approved):

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -245,10 +247,19 @@ export const UnderlineNav = forwardRef(
...
-    }, [])
+    const updateListAndMenu = useCallback(
+      (props: ResponsiveProps, displayIcons: boolean) => {
+        setResponsiveProps(props)
+        setIconsVisible(displayIcons)
+
+        // Only mark as ready once widths have been measured for all valid children
+        const widths = displayIcons ? childWidthArray : noIconChildWidthArray
+        if (!isHydrated && widths.length > 0 && widths.length === validChildren.length) {
+          setIsHydrated(true)
```
siddharthkp: Oh if it's not just hydration but also calculations, maybe we should call it `data-overflow-measured` or something similar?

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -330,7 +341,14 @@ export const UnderlineNav = forwardRef(
...
       >
         {ariaLabel && <VisuallyHidden as="h2">{`${ariaLabel} navigation`}</VisuallyHidden>}
-        <UnderlineWrapper as={as} aria-label={ariaLabel} className={className} ref={navRef} data-variant={variant}>
+        <UnderlineWrapper
+          as={as}
+          aria-label={ariaLabel}
+          className={className}
+          ref={navRef}
+          data-variant={variant}
+          hydrated={isHydrated}
```
siddharthkp: Can we pass `data-hydrated={isHydrated}` and let it spread. Not a fan of the extra prop

TylerJDev:
One more review @siddharthkp 🙏 I updated some of the logic, as the data-* attribute wasn't getting set ([as mentioned by @iansan5653](https://github.com/primer/react/pull/7504#pullrequestreview-3764272476))

siddharthkp (approved):

TylerJDev:
[VRT passes](https://github.com/github/github-ui/actions/runs/22129308917), skipping integration tests as it seems to be stuck.

Issue #7437, opened 2026-01-09
Allow customizing `ActionBar` overflow menu label, or default to "More items"


iansan5653:
The label for the `ActionBar` overflow menu is forced to be `More {menuLabel} items`. This means that for any well-formed, semantic label for the action bar, the overflow menu is going to have a nonsense label/tooltip.

For example, logical names for the toolbar include "Formatting tools", "Issue actions", "Repository navigation", etc. The resulting overflow menu names for these are "More Formatting tools items", "More Issue actions items", "More Repository navigation items". They all have bad capitalization, are long for a button label, and most just don't make sense with two plural nouns.

I think one easy solution would just be to just change the label to "More items". This is generic, but should be fine because it's always nested inside a labeled toolbar, so it's clear based on context what it's referring to. It works in every context and is succinct and clear.

harshbhonsle:
Hi, is this issue still open? I’d be happy to work on it and submit a PR if that’s helpful.

lesliecdubs:
Hi @harshbhonsle, this hasn't been picked up yet. Do you want to give it a go? If so, please assign yourself to the issue and let us know if you have questions.

harshbhonsle:
Hi @lesliecdubs, thanks for confirming! Yes, I’d be happy to take this on. My approach will be to first locate where the overflow label is defined in the ActionBar component, then simplify the default to "More items" while also adding a prop to allow customization if needed. After that, I’ll update tests and documentation to reflect the change. I’ll go ahead and assign myself to the issue and start working on a PR.

lesliecdubs:
Thanks, sounds like a good plan to me! When you're ready, mark your PR as "ready for review" and someone on the team will take a look. In the meantime, you can let us know here if you run into any questions or problems.

harshbhonsle:
Thanks for your patience — I was occupied with some exams, so apologies for the delay in getting this PR submitted. The fix is now ready for review. Please let me know if there are any changes or adjustments you’d like me to make.

TylerJDev:
Thanks for the PR @harshbhonsle! I'll be taking a look shortly!

TylerJDev:
Left a few comments @harshbhonsle, let me know if anything comes up! Thank you for the contribution too!

lesliecdubs:
See https://github.com/primer/react/pull/7520#issuecomment-4723540076

PR #7506, opened 2026-02-06
Fix and refactor `UnderlineNav` to resolve CLS issues and improve performance/code quality

iansan5653:
This refact/bugfix PR reworks `UnderlineNav`'s internal logic to rely mostly on CSS for calculating overflow, eliminating layout shift issues. In addition, it removes the use of the `React.Children` API in favor of context-based solutions, and replaces the custom overflow menu with an `ActionMenu`.

See https://github.com/github/primer/issues/6291#issuecomment-3861950186 for context.

The CSS overflow logic works like this:

1. Overflowing menu items wrap onto subsequent lines, which are clipped by their container using `overflow: hidden`, creating the illusion that they disappeared
2. The initial visibility of the overflow menu button and tab icons is controlled by `scroll-state` container queries to detect overflow using pure CSS
3. Pure CSS cannot, however, tell us what to render inside the overflow menu. For this, each list item registers itself using context:
    1. The parent component provides register/unregister functions through context
    2. Each item component registers an `IntersectionObserver` to update itself with the parent when it appears/disappears. They determine whether they are overflowed by checking `ref.current.offsetHeight` (which will be greater than zero for an item that has wrapped). If they are overflowed, they set `aria-hidden`/`tabIndex="-1"` and notify the parent via the registry.
    3. The parent renders menu items for all currently-overflowing items in the overflow menu
    4. If any item has ever overflowed, all tab icons are suppressed via React state for the rest of the lifecycle of the component (this prevents flickering caused by the tab icons creating/removing space)
 
The benefits are pretty significant:

1. The browser will calculate overflow for us before we ever even paint on the page - no waiting on a JS bundle or React hydration (this benefit is shared with the lighter-weight approach)
2. Calculations are nearly pure CSS, so performance is buttery smooth for the whole component lifecycle
3. We get to clean up tons of complex code, remove several effects, and remove hacky `Children` APIs entirely
4. Bonus: Component consumption is much less fragile. Without `Children` APIs, we no longer need direct children, so consumers can wrap their underline items in fragments and even wrapper components

However, there are some caveats / user-facing changes:

1. The most significant is that **we can no longer ensure that the current item will not overflow into the menu**. This is impossible to implement with the new approach; even if we were to extract the current item and render it outside the overflow menu, there's no place to put it since it would just wrap out of view if rendered into the list. If rendered next to the list, it would be pushed too far to the right due to the space caused by wrapping items in the list. I even tried implementing some JS to swap it with the last visible item, but that still can result in overflow if the current item is larger than that item. Instead, I've added some styling to indicate the current item inside the menu.
2. The overflow menu button must be right-aligned, as opposed to appearing right next to the last button, because the list of items will always take up as much space as possible before wrapping. Honestly I think this is a nice design improvement.

https://github.com/user-attachments/assets/350e14cb-a7b1-4acb-8bf6-2035cf52935f

- Closes https://github.com/github/primer/issues/6291

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


iansan5653:
BTW, I think the flex inspector in Chrome's devtools does a really nice job of showing what's going on here / where the tabs are going:

https://github.com/user-attachments/assets/a58a1e15-f982-4a40-8a60-ccd53f953274

iansan5653:
For the last item in the menu, I could come up with a way to move it to the overflow menu via JS. But I'm wondering if we really should? That's the current behavior, but does it ever make sense to just have a "more" menu standing alone? Perhaps we could consider truncating the item instead? I know that has accessibility concerns but it seems like the lesser of the two evils.

screenshot showing truncated first tab item


I also don't want to over-engineer this edge case as it seems practically impossible to actually reach in the real world - you'd need an extra long first item title and a very narrow container size at the same time.

iansan5653:
Quick update: currently blocked by the new feature callout CTA (popover). Working on addressing this in internal PR https://github.com/github/github-ui/pull/13807

iansan5653:
`ActionMenu` worked perfectly for fixing the overflow menu positioning! And saved us a ton of unnecessary code
correctly positioned overflow menu

Also kudos to @hectahertz for the `IntersectionObserver` idea. This allowed me to replace the `ResizeObserver` and the result is more reliable

packages/react/src/internal/components/UnderlineTabbedInterface.tsx
```diff
@@ -62,7 +63,6 @@ export const LoadingCounter = () => {
 export type UnderlineItemProps<As extends React.ElementType> = {
   as?: As | 'a' | 'button'
   className?: string
-  iconsVisible?: boolean
```
iansan5653, line 65: Icon visibility is now pure CSS, saving some props + logic and improving SSR consistency

packages/react/src/UnderlineNav/styles.ts
```diff
@@ -1,19 +0,0 @@
-export const dividerStyles = {
```
iansan5653, line 1: Moved this style to the CSS file. Looks like the JS object is just left over from styled-components days

packages/react/src/UnderlineNav/styles.ts
```diff
@@ -1,19 +0,0 @@
...
-  display: 'inline-block',
-  borderLeft: '1px solid',
-  width: '1px',
-  borderLeftColor: 'var(--borderColor-muted)',
-  marginRight: 'var(--base-size-4)',
-  height: '24px', // The height of the divider - reference from Figma
-}
-
-export const menuItemStyles = {
```
iansan5653, line 10: Deleted this style since we're using `ActionMenu` now and it supports not rendering the selection icon

iansan5653:
@copilot Run and update VRT snapshots

e2e/components/UnderlineNav.test.ts
```diff
@@ -202,12 +202,12 @@ test.describe('UnderlineNav', () => {
           // expect(await page.screenshot()).toMatchSnapshot()
 
           await page.setViewportSize({width: viewports['primer.breakpoint.sm'], height: 768})
-          await page.locator('button', {hasText: 'More Repository Items'}).waitFor()
```
iansan5653, line 205: I've updated the label of the overflow button to "More items", aligning with this issue for `ActionBar`: https://github.com/primer/react/issues/7437

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -158,39 +54,22 @@ export const UnderlineNav = forwardRef(
...
+    )
 
-    // Make sure to have the fresh props data for menu items when children are changed (keeping aria-current up-to-date)
-    const menuItems = responsiveProps.menuItems.map(menuItem => {
-      return validChildren.find(child => child.key === menuItem.key) ?? menuItem
-    })
-    // This is the case where the viewport is too narrow to show any list item with the more menu. In this case, we only show the dropdown
-    const onlyMenuVisible = responsiveProps.items.length === 0
+    const isOverflowing = menuItems.length > 0
+    if (isOverflowing && !hasEverOverflowed) setHasOverflowed(true)
```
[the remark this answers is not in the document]
iansan5653: This is incorrect; React themselves recommend calling setState directly during the render cycle here and not using an effect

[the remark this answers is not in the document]
iansan5653: ref https://react.dev/learn/you-might-not-need-an-effect

packages/react/src/UnderlineNav/UnderlineNav.module.css
```diff
@@ -1,10 +1,62 @@
...
+
+  &::after {
+    content: '';
+    width: var(--base-size-2);
+    position: absolute;
+    inset: var(--base-size-2) auto var(--base-size-2) 0;
+    /* stylelint-disable-next-line primer/colors */
+    background: var(--underlineNav-borderColor-active);
+  }
+}
```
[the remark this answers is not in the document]
iansan5653: This is incorrect, in DOM it gets applied to a descendant

packages/react/src/internal/components/UnderlineTabbedInterface.module.css
```diff
@@ -104,6 +111,13 @@
   margin-inline-end: var(--base-size-8);
 }
 
+/* Truncate items if necessary (should only ever apply to the last item after everything else overflows) */
+.UnderlineItem [data-component='text'] {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
```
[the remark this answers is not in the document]
iansan5653: only the last item will be visible at this point

iansan5653:
@copilot add a changeset for this pr to the `.changeset` folder

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.tsx
```diff
@@ -178,7 +178,7 @@ const UnderlinePanels: FCWithSlotMarker<UnderlinePanelsProps> = ({
       <UnderlineWrapper
         ref={wrapperRef}
         slot="tablist-wrapper"
-        data-icons-visible={iconsVisible}
+        data-hide-icons={!iconsVisible}
```
hectahertz: Bug: this now emits `data-hide-icons` but [UnderlinePanels.module.css](https://github.com/primer/react/blob/4d8d5e6b9b0b1d394d19154369f9c37eabe7cb87/packages/react/src/experimental/UnderlinePanels/UnderlinePanels.module.css#L8) still targets the old `data-icons-visible='false'` attribute, so icon hiding in `UnderlinePanels` is now broken.

Since `UnderlineTabbedInterface.module.css` already handles `[data-hide-icons='true']`, the fix is probably just deleting the old rule from `UnderlinePanels.module.css`.

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -158,39 +54,22 @@ export const UnderlineNav = forwardRef(
...
+      (entry): entry is [string, UnderlineNavItemProps] => entry[1] !== null,
+    )
 
-    // Make sure to have the fresh props data for menu items when children are changed (keeping aria-current up-to-date)
-    const menuItems = responsiveProps.menuItems.map(menuItem => {
-      return validChildren.find(child => child.key === menuItem.key) ?? menuItem
-    })
-    // This is the case where the viewport is too narrow to show any list item with the more menu. In this case, we only show the dropdown
-    const onlyMenuVisible = responsiveProps.items.length === 0
+    const isOverflowing = menuItems.length > 0
```
hectahertz: This `setState` during render works (React allows it for the component's own state), but it will trigger an immediate re-render on first overflow detection. Totally fine in practice, just flagging since `useEffect` is the more common pattern.

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -38,111 +32,13 @@ export type UnderlineNavProps = {
 // However, we need to calculate number of possible items when the more button present as well. So using the width of the more button as a constant.
 export const MORE_BTN_WIDTH = 86
```
hectahertz: Nit: `MORE_BTN_WIDTH` is now dead code (the JS overflow calculation is gone). `getValidChildren` below is also only used in the `__DEV__` block now and doesn't need to be exported.

packages/react/src/UnderlineNav/UnderlineNavItem.tsx
```diff
@@ -59,86 +59,81 @@ export type UnderlineNavItemProps = {
...
+  const {loadingCounters} = useContext(UnderlineNavContext)
+
+  const isOverflowing = useSyncExternalStore(
+    useCallback(
+      onChange => {
+        const observer = new IntersectionObserver(() => onChange(), {
+          threshold: 1,
+        })
+        if (ref.current) observer.observe(ref.current)
+        return () => observer.disconnect()
```
hectahertz, line 32: Clever pattern! One subtlety worth a brief comment: the `IntersectionObserver` is just being used as a trigger to re-check `offsetTop > 0`. The actual overflow detection relies on `flex-wrap` + `overflow: hidden` pushing items to a second row (giving them `offsetTop > 0`), not on intersection with the viewport.

A quick note would help a future reader understand why there's no `root` option on the observer.

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -203,244 +82,83 @@ export const UnderlineNav = forwardRef(
...
-                    )
-                  })}
+          <div
+            style={{
+              height: `${MORE_BTN_HEIGHT}px`,
+            }}
+            className={classes.MoreButtonContainer}
+          >
+            <div className={classes.MoreButtonDivider} />
+
```
hectahertz, line 111: Since the overflow menu + button are always in the DOM (visibility-toggled by CSS), worth checking that `visibility: hidden` properly removes them from the accessibility tree in common screen readers. If not, an `aria-hidden` on the container when nothing is overflowing would be a good safety net.

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -203,244 +82,83 @@ export const UnderlineNav = forwardRef(
...
+                      const {children: menuItemChildren, counter, onSelect, ...menuItemProps} = allProps
+
+                      return (
+                        <ActionList.LinkItem
+                          key={key}
+                          className={classes.OverflowMenuItem}
+                          onClick={event => onSelect?.(event)}
+                          {...(menuItemProps as Omit<typeof menuItemProps, 'as'>)}
+                        >
+                          <span className={classes.OverflowMenuItemLabel}>{menuItemChildren}</span>
```
hectahertz, line 142: Nit: `ActionList.TrailingVisual` still renders when `loadingCounters` is false and `counter` is undefined. You could wrap the whole thing in a conditional to skip mounting it:

```tsx
{(loadingCounters || counter !== undefined) && (
  <ActionList.TrailingVisual>
    {loadingCounters ? <LoadingCounter /> : (
      <span data-component="counter"><CounterLabel>{counter}</CounterLabel></span>
    )}
  </ActionList.TrailingVisual>
)}
```

packages/react/src/UnderlineNav/UnderlineNav.module.css
```diff
@@ -1,10 +1,62 @@
-.MenuItemContent {
+.UnderlineWrapper {
+  /* Progressive enhancement: Detect overflow using scroll-based animations.
+  The idiomatic way would be a scroll-state container query but browser support
+  is slightly better for animations. */
+  animation: detect-overflow linear;
```
hectahertz, line 5: Nice progressive enhancement! `animation-timeline: scroll()` isn't supported in Safari/Firefox yet, so it's good there's a JS fallback via `data-has-overflow`.

One thought: between first paint and the first IO callback, overflowing items will be clipped but the "More" button won't be visible yet. Is that flash acceptable, or should we show the button by default and hide it once we confirm nothing overflows?

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -38,111 +32,13 @@ export type UnderlineNavProps = {
...
-          menuItems.push(propsectiveAction)
-        } else {
-          menuItems.push(child)
-        }
-      }
-    }
-  }
-  updateListAndMenu({items, menuItems}, iconsVisible, true)
-}
+const MORE_BTN_HEIGHT = 32
```
hectahertz: Nit: could move this into the CSS instead of an inline `style`. One less magic number in JS.

hectahertz (commented):
Really nice refactor! The shift from JS width measurements to CSS wrap-based overflow detection is a big win for CLS, and the `createDescendantRegistry` utility is clean and well-tested. Great code removal too.

Left some comments, nothing major.

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.tsx
hectahertz: Bug: this now emits `data-hide-icons` but [UnderlinePanels.module.css](https://github.com/primer/react/blob/4d8d5e6b9b0b1d394d19154369f9c37eabe7cb87/packages/react/src/experimental/UnderlinePanels/UnderlinePanels.module.css#L8) still targets the old `data-icons-visible='false'` attribute, so icon hiding in `UnderlinePanels` is now broken.

Since `UnderlineTabbedInterface.module.css` already handles `[data-hide-icons='true']`, the fix is probably just deleting the old rule from `UnderlinePanels.module.css`.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz: This `setState` during render works (React allows it for the component's own state), but it will trigger an immediate re-render on first overflow detection. Totally fine in practice, just flagging since `useEffect` is the more common pattern.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz: Nit: `MORE_BTN_WIDTH` is now dead code (the JS overflow calculation is gone). `getValidChildren` below is also only used in the `__DEV__` block now and doesn't need to be exported.

hectahertz (commented):
Really nice refactor! The shift from JS width measurements to CSS wrap-based overflow detection is a big win for CLS, and the `createDescendantRegistry` utility is clean and well-tested. Great code removal too.

Left some comments, nothing major.

packages/react/src/UnderlineNav/UnderlineNavItem.tsx
hectahertz, line 32: Clever pattern! One subtlety worth a brief comment: the `IntersectionObserver` is just being used as a trigger to re-check `offsetTop > 0`. The actual overflow detection relies on `flex-wrap` + `overflow: hidden` pushing items to a second row (giving them `offsetTop > 0`), not on intersection with the viewport.

A quick note would help a future reader understand why there's no `root` option on the observer.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz, line 111: Since the overflow menu + button are always in the DOM (visibility-toggled by CSS), worth checking that `visibility: hidden` properly removes them from the accessibility tree in common screen readers. If not, an `aria-hidden` on the container when nothing is overflowing would be a good safety net.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz, line 142: Nit: `ActionList.TrailingVisual` still renders when `loadingCounters` is false and `counter` is undefined. You could wrap the whole thing in a conditional to skip mounting it:

```tsx
{(loadingCounters || counter !== undefined) && (
  <ActionList.TrailingVisual>
    {loadingCounters ? <LoadingCounter /> : (
      <span data-component="counter"><CounterLabel>{counter}</CounterLabel></span>
    )}
  </ActionList.TrailingVisual>
)}
```

packages/react/src/UnderlineNav/UnderlineNav.module.css
hectahertz, line 5: Nice progressive enhancement! `animation-timeline: scroll()` isn't supported in Safari/Firefox yet, so it's good there's a JS fallback via `data-has-overflow`.

One thought: between first paint and the first IO callback, overflowing items will be clipped but the "More" button won't be visible yet. Is that flash acceptable, or should we show the button by default and hide it once we confirm nothing overflows?

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz: Nit: could move this into the CSS instead of an inline `style`. One less magic number in JS.

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.tsx
hectahertz: Bug: this now emits `data-hide-icons` but [UnderlinePanels.module.css](https://github.com/primer/react/blob/4d8d5e6b9b0b1d394d19154369f9c37eabe7cb87/packages/react/src/experimental/UnderlinePanels/UnderlinePanels.module.css#L8) still targets the old `data-icons-visible='false'` attribute, so icon hiding in `UnderlinePanels` is now broken.

Since `UnderlineTabbedInterface.module.css` already handles `[data-hide-icons='true']`, the fix is probably just deleting the old rule from `UnderlinePanels.module.css`.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz: This `setState` during render works (React allows it for the component's own state), but it will trigger an immediate re-render on first overflow detection. Totally fine in practice, just flagging since `useEffect` is the more common pattern.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz: Nit: `MORE_BTN_WIDTH` is now dead code (the JS overflow calculation is gone). `getValidChildren` below is also only used in the `__DEV__` block now and doesn't need to be exported.

packages/react/src/UnderlineNav/UnderlineNavItem.tsx
hectahertz, line 32: Clever pattern! One subtlety worth a brief comment: the `IntersectionObserver` is just being used as a trigger to re-check `offsetTop > 0`. The actual overflow detection relies on `flex-wrap` + `overflow: hidden` pushing items to a second row (giving them `offsetTop > 0`), not on intersection with the viewport.

A quick note would help a future reader understand why there's no `root` option on the observer.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz, line 111: Since the overflow menu + button are always in the DOM (visibility-toggled by CSS), worth checking that `visibility: hidden` properly removes them from the accessibility tree in common screen readers. If not, an `aria-hidden` on the container when nothing is overflowing would be a good safety net.

hectahertz (commented):
Really nice refactor! The shift from JS width measurements to CSS wrap-based overflow detection is a big win for CLS, and the `createDescendantRegistry` utility is clean and well-tested. Great code removal too.

Left some comments, nothing major.

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz, line 142: Nit: `ActionList.TrailingVisual` still renders when `loadingCounters` is false and `counter` is undefined. You could wrap the whole thing in a conditional to skip mounting it:

```tsx
{(loadingCounters || counter !== undefined) && (
  <ActionList.TrailingVisual>
    {loadingCounters ? <LoadingCounter /> : (
      <span data-component="counter"><CounterLabel>{counter}</CounterLabel></span>
    )}
  </ActionList.TrailingVisual>
)}
```

packages/react/src/UnderlineNav/UnderlineNav.module.css
hectahertz, line 5: Nice progressive enhancement! `animation-timeline: scroll()` isn't supported in Safari/Firefox yet, so it's good there's a JS fallback via `data-has-overflow`.

One thought: between first paint and the first IO callback, overflowing items will be clipped but the "More" button won't be visible yet. Is that flash acceptable, or should we show the button by default and hide it once we confirm nothing overflows?

packages/react/src/UnderlineNav/UnderlineNav.tsx
hectahertz: Nit: could move this into the CSS instead of an inline `style`. One less magic number in JS.

packages/react/src/UnderlineNav/UnderlineNav.tsx
iansan5653: Yep, React will abandon this render and start a new one. It's uncommon but probably should be more common - [React recommends it over using an effect](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes)

packages/react/src/UnderlineNav/UnderlineNav.tsx
iansan5653: Removed, and looks like we can also remove the `HEIGHT` now as well

packages/react/src/experimental/UnderlinePanels/UnderlinePanels.tsx
iansan5653: Oh good catch, dropped that CSS block entirely

packages/react/src/UnderlineNav/UnderlineNavItem.tsx
iansan5653: Good point, added a comment where we check the `offsetTop`

packages/react/src/UnderlineNav/UnderlineNav.tsx
iansan5653: Confirmed it does (both in [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/visibility#accessibility) and in practice on my machine) :+1: - And also drops the elements from the focus order

packages/react/src/UnderlineNav/UnderlineNav.tsx
iansan5653: Resolved by moving the trailingVisual into each condition

packages/react/src/UnderlineNav/UnderlineNav.module.css
iansan5653: This is a great question, I was wondering the same thing. The challenge is that we have to choose - on small screens do we delay showing the button, or on large screens do we show and then hide it when there's no overflow?

I think, given the choice, it's typically preferable to insert new UI after SSR vs hiding UI that you initially showed. Showing and then hiding something feels more flickery than showing it after a short delay. I also think large screens are probably more common.

To be clear, this flash would only happen in browsers that don't support scroll animation. 

We could try and be more clever here by showing it by default on small screens based on some arbitrary breakpoint, trying to predict when there's most likely going to be an overflow. But I'm not sure it's worth the complexity and it could never be 100% reliable since we can't know what width would trigger the overflow until after we calculate the overflowed items.

packages/react/src/UnderlineNav/UnderlineNav.tsx
iansan5653: Dropped this entirely, it's not necessary anymore 🙌

hectahertz (approved):
Really solid work on this. The CSS approach for overflow detection is a meaningful improvement over the old JS measurement loop, and the code is noticeably cleaner. Happy to see this land 🚀

iansan5653:
I spent some time yesterday trying to find a solution which would enable us to force the `aria-current` item to always be visible. 

I think the most promising solution is to use `order` to reorder items to move the current one to the top level. The naive approach is to assign `order: 3` to all the hidden items, `order: 1` to the visible items, and `order: 2` to the current item if hidden. However, this doesn't actually work because it just moves the current item to the front of the hidden items -- it doesn't create space for it in the visible region.

So we want to move the item in front of all the hidden ones, but also shift enough visible items to be hidden to make room for the current item. This is complex enough, but is complicated by the fact that the logic is circular; it depends on figuring out which items are visible and not visible, but then changes which items are visible and not visible. Any effect-based approach would be very fragile. I think it's possible, but ultimately would cost too much in performance and complexity to be worth it. 

We could simplify this approach significantly by just putting the current item first if it would overflow (`order: 0`), but that breaks the tabbing order and would feel pretty weird as the order of items wouldn't be consistent. 

I also explored simply rendering the current item at the top level next to the overflow menu as we did before. The problem here is the same one that forces us to right-align the menu though; the wrapping container for the other menu items has to take up as much width as possible, so there can be a gap between the current item and the main items. The flex inspector shows how there just isn't a place to put an item. The same would apply to any `grid` based approach.

chrome devtools flex inspector


Finally, I opened an alternative PR to explore the possibility of just getting rid of the overflow menu and scrolling the container instead: https://github.com/primer/react/pull/7648

lesliecdubs:
Sorry for the hold-up on this one. Working on finding a reviewer to take it over from @hectahertz 🙏🏻

TylerJDev (approved):

iansan5653:
@copilot

1. Resolve the merge conflicts in this pull request
2. When the current item is overflowed, style the overflow menu anchor to look like it's the current item (show the orange decoration on it)
3. When the current item is overflowed, append ", including current item {name}" to the `aria-label` of the overflow menu anchor button

jonrohan (approved):

iansan5653:
Testing in preview showed a few small issues:

1. Test failures caused by outdated selectors and the need to set the viewport size. These are resolved by https://github.com/github/github-ui/pull/24511/changes/b2f650f87b7d74db379c0903ffbf3dca62f706b9 which needs to be pulled into the release integration PR
2. Some layout quirks in PRs are fixed by https://github.com/github/github-ui/pull/24554 which can be shipped before this change is released
3. A tiny 1px discrepancy in the decoration applied to the "More" button was fixed in this PR
4. The current item inside the overflow menu was missing the current styling in PRs; this was caused by a class override from the consumer. Fixed in this PR to force using our class only when inside the menu (this makes sense, consumers don't need to style the item inside the menu even though they may need to set classes on the top-level tabs)

jonrohan (approved):

iansan5653:
Okay, the PR is against the latest and CI is passing, so I'm going to go ahead and override to merge.

PR #7520, opened 2026-02-10
feat(ActionBar): customizable overflow menu label

### LINKED ISSUE #7437: Allow customizing `ActionBar` overflow menu label, or default to "More items"

harshbhonsle:
Closes: #7437

#### New
- Added `overflowLabel` prop to `ActionBarProps` with default `"More items"`.
- Storybook updated to expose `overflowLabel` as a control.

#### Changed
- Overflow menu button `aria-label` now uses `overflowLabel` instead of hardcoded `"More {ariaLabel} items"`.

#### Removed
- None

### Rollout strategy
- [x] Patch release

### Testing & Reviewing
- Verified manually in Storybook:
  - Default shows `"More items"`.
  - Custom values (e.g. `"More actions"`) update correctly.
- Could not locate an existing test file for `ActionBar` under `src/__tests__`. Fix validated through Storybook preview.

### Merge checklist
- [ ] Added/updated tests (none found for ActionBar, validated manually)
- [x] Added/updated documentation (Storybook)
- [x] Changes are SSR compatible
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/ActionBar/ActionBar.stories.tsx
+++ b/packages/react/src/ActionBar/ActionBar.stories.tsx
@@ -21,8 +21,15 @@ export default meta
 type Story = StoryObj<typeof ActionBar>
 
 export const Playground: Story = {
-  render: ({'aria-labelledby': _, ...args}) => (
-    <ActionBar {...args} aria-label="Toolbar">
+  args: {
+    overflowLabel: 'More item',
+    size: 'medium',
+    flush: false,
+    gap: 'condensed',
+  },
+
+  render: ({'aria-labelledby': _, overflowLabel, ...args}) => (
+    <ActionBar {...args} overflowLabel={overflowLabel} aria-label="Toolbar">
       <ActionBar.IconButton icon={BoldIcon} aria-label="Bold"></ActionBar.IconButton>
       <ActionBar.IconButton icon={ItalicIcon} aria-label="Italic"></ActionBar.IconButton>
       <ActionBar.Divider />
@@ -48,11 +55,17 @@ Playground.argTypes = {
     description: 'Horizontal gap scale between items',
     table: {defaultValue: {summary: 'condensed'}},
   },
+  overflowLabel: {
+    control: {type: 'text'},
+    description: 'Accessible label for the overflow menu button',
+    table: {defaultValue: {summary: 'More items'}},
+  },
 }
 Playground.args = {
   size: 'medium',
   flush: false,
   gap: 'condensed',
+  overflowLabel: 'More items',
 }
 
 export const Default = () => (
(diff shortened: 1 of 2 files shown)


packages/react/src/ActionBar/ActionBar.stories.tsx
```diff
@@ -21,8 +21,15 @@ export default meta
...
-  render: ({'aria-labelledby': _, ...args}) => (
-    <ActionBar {...args} aria-label="Toolbar">
+  args: {
+    overflowLabel: 'More item',
+    size: 'medium',
+    flush: false,
+    gap: 'condensed',
+  },
+
+  render: ({'aria-labelledby': _, overflowLabel, ...args}) => (
```
TylerJDev, line 31: Could we rely on the spread (`{...args}` line 32) to pass the `overflowLabel` to the story? This way we don't need to create the object entirely. 

Let me know if that makes sense, or if you have questions! 

```suggestion
  render: ({'aria-labelledby': _, overflowLabel, ...args}) => (
```

TylerJDev (changes requested):
Overall looks great! Just left one request, let me know if you have questions. Thank you for the PR too! ✨

TylerJDev:
Also, can we add a changeset of `minor` (as we're adding a new prop here). You can refer to this guide on adding a changeset: https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#adding-changeset-to-your-pull-request or run `npx changeset`. 

In addition, you might need to run `npm run format`, as I see there's a [CI failure](https://github.com/primer/react/actions/runs/21867372550/job/63590627736?pr=7520) for formatting. Let me know if you have any questions!

lesliecdubs:
It looks like @harshbhonsle is no longer active on this PR. Instead of letting this change go stale, I wonder if the reviewer @TylerJDev might be willing to take it over and get it to the finish line?

PR #7521, opened 2026-02-10
PageLayout: Add new layout option `PageLayout.Sidebar`

TylerJDev:
Closes https://github.com/github/primer/issues/6419

`PageLayout` handles the layout for many pages across GH, and comes with 4 sub-components to define a layout (header, content, pane, footer). This PR adds an additional sub-component for side content. 

We currently utilize `pane` to encapsulate content adjacent to the content and directly underneath the header. We do not offer a way to "escape" the main layout and add content adjacent to it.

#### New

* Adds `PageLayout.Sidebar`

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/PageLayout/PageLayout.docs.json
+++ b/packages/react/src/PageLayout/PageLayout.docs.json
@@ -307,12 +307,6 @@
           "description": "When true, the sidebar may be resized by the user. Width is persisted to localStorage by default."
         },
         {
-          "name": "widthStorageKey",
-          "type": "string",
-          "defaultValue": "'sidebarWidth'",
-          "description": "localStorage key used to persist the sidebar width across sessions. Only applies when resizable is true."
-        },
-        {
           "name": "padding",
           "type": "| 'none' | 'condensed' | 'normal'",
           "defaultValue": "'none'",
--- a/packages/react/src/PageLayout/PageLayout.tsx
+++ b/packages/react/src/PageLayout/PageLayout.tsx
@@ -1077,8 +1077,9 @@ export type PageLayoutSidebarProps = {
   resizable?: boolean
 
   /**
-   * Storage key for persisting the sidebar width
-   * @default 'sidebarWidth'
+   * localStorage key used to persist the sidebar width across sessions.
+   * Only applies when `resizable` is `true`.
+   * When omitted, localStorage is not used.
    */
   widthStorageKey?: string
 
@@ -1131,7 +1132,7 @@ const Sidebar = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLay
       minWidth = 256,
       padding = 'none',
       resizable = false,
-      widthStorageKey = 'sidebarWidth',
+      widthStorageKey,
       divider = 'none',
       sticky = false,
       responsiveVariant = 'default',
@@ -1224,7 +1225,7 @@ const Sidebar = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLay
           ref={sidebarRef}
           // Suppress hydration mismatch for --pane-width when localStorage
           // provides a width that differs from the server-rendered default.
-          suppressHydrationWarning={resizable === true}
+          suppressHydrationWarning={resizable === true && !!widthStorageKey}
           {...(hasOverflow ? overflowProps : {})}
           {...labelProp}
           {...(id && {id: sidebarId})}
--- a/packages/react/src/PageLayout/usePaneWidth.ts
+++ b/packages/react/src/PageLayout/usePaneWidth.ts
@@ -26,7 +26,8 @@ export type UsePaneWidthOptions = {
   width: PaneWidthValue
   minWidth: number
   resizable: boolean
-  widthStorageKey: string
+  /** localStorage key for persisting width. When undefined, localStorage is not used. */
+  widthStorageKey?: string
   paneRef: React.RefObject<HTMLDivElement | null>
   handleRef: React.RefObject<HTMLDivElement | null>
   contentWrapperRef: React.RefObject<HTMLDivElement | null>
@@ -235,7 +236,7 @@ export function usePaneWidth({
     }
     // Try localStorage if onResizeEnd is not provided (default persistence behavior)
     // Read directly here instead of via persister to satisfy react-hooks/refs lint rule
-    const shouldUseLocalStorage = onResizeEnd === undefined && resizable === true
+    const shouldUseLocalStorage = onResizeEnd === undefined && resizable === true && widthStorageKey !== undefined
     if (shouldUseLocalStorage) {
       const storedWidth = localStoragePersister.get(widthStorageKey)
       if (storedWidth !== null) {
@@ -311,8 +312,8 @@ export function usePaneWidth({
         return
       }
 
-      // Handle localStorage persistence when resizable === true and not controlled
-      if (resizable) {
+      // Handle localStorage persistence when resizable === true, not controlled, and key is provided
+      if (resizable && widthStorageKeyRef.current) {
         localStoragePersister.save(widthStorageKeyRef.current, rounded)
       }
     },
--- a/packages/react/src/SplitPageLayout/SplitPageLayout.docs.json
+++ b/packages/react/src/SplitPageLayout/SplitPageLayout.docs.json
@@ -236,12 +236,6 @@
           "description": "When true, the sidebar may be resized by the user. Width is persisted to localStorage by default."
         },
         {
-          "name": "widthStorageKey",
-          "type": "string",
-          "defaultValue": "'sidebarWidth'",
-          "description": "localStorage key used to persist the sidebar width across sessions. Only applies when resizable is true."
-        },
-        {
           "name": "padding",
           "type": "| 'none' | 'condensed' | 'normal'",
           "defaultValue": "'normal'",


packages/react/src/PageLayout/PageLayout.module.css
```diff
@@ -742,3 +746,139 @@
   contain: layout style paint;
   pointer-events: none;
 }
+
+/* Sidebar */
+.PageLayoutRoot:where([data-has-sidebar]) {
```
TylerJDev, line 751: The styles from this point and below (at the time of this comment) are all scoped for `Sidebar`. If there is no sidebar, then these styles won't applying, meaning that these are all net new changes that won't affect PageLayout implementations today.

packages/react/src/PageLayout/PageLayout.module.css
```diff
@@ -742,3 +746,139 @@
...
+      position: fixed;
+      top: 0;
+      left: 0;
+      width: 100vw;
+      height: 100vh;
+      min-width: unset;
+      max-width: unset;
+      border-radius: 0;
+      /* stylelint-disable-next-line primer/spacing */
+      z-index: 999;
```
TylerJDev, line 818: We'll probably want to modify this, but this is to indicate that it should be on the top layer when fullscreen.

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -196,6 +209,114 @@ const VerticalDivider = memo<React.PropsWithChildren<VerticalDividerProps>>(
...
+  sidebarRef: React.RefObject<HTMLDivElement>
+  dragStartClientXRef: React.MutableRefObject<number>
+  dragStartWidthRef: React.MutableRefObject<number>
+  dragMaxWidthRef: React.MutableRefObject<number>
+  getMaxPaneWidth: () => number
+  getDefaultWidth: () => number
+  saveWidth: (width: number) => void
+}
+
+const SidebarDivider = memo<SidebarDividerProps>(function SidebarDivider({
```
TylerJDev, line 230: We mainly use the same logic as the pane here, but oriented for the Sidebar. Overall nothing here is truly new.

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -918,6 +1039,237 @@ Pane.displayName = 'PageLayout.Pane'
...
+  /**
+   * Optional id for the sidebar element
+   */
+  id?: string
+
+  className?: string
+  style?: React.CSSProperties
+}
+
+const Sidebar = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PageLayoutSidebarProps>>(
```
TylerJDev, line 1125: We mainly use the same props as pane, so that the API is consistent. We could see about combining the two sub components (`pane`, `sidebar`), but this would be better to look into in the future if we ship this.

packages/react/src/PageLayout/usePaneWidth.ts
```diff
@@ -100,14 +112,17 @@ export const getDefaultPaneWidth = (w: PaneWidthValue): number => {
...
+ * Gets the max-width-diff CSS variable value from a pane element.
+ * For sidebars, reads --sidebar-max-width-diff (constant across viewports).
+ * For panes, reads --pane-max-width-diff (changes at 1280px breakpoint).
  * Note: This calls getComputedStyle which forces layout - cache the result when possible.
  */
-export function getPaneMaxWidthDiff(paneElement: HTMLElement | null): number {
-  if (!paneElement) return DEFAULT_MAX_WIDTH_DIFF
-  const value = parseInt(getComputedStyle(paneElement).getPropertyValue('--pane-max-width-diff'), 10)
-  return value > 0 ? value : DEFAULT_MAX_WIDTH_DIFF
+export function getPaneMaxWidthDiff(paneElement: HTMLElement | null, isSidebar = false): number {
```
TylerJDev, line 121: Most of the new additions within `usePaneWidth` should be constrained to just the sidebar, so this shouldn't affect or change any existing `PageLayout` implementation without the sidebar.

packages/react/src/PageLayout/PageLayout.docs.json
```diff
@@ -251,6 +269,81 @@
...
+          "type": "boolean",
+          "defaultValue": "false",
+          "description": "When true, the sidebar may be resized by the user. Width is persisted to localStorage by default."
+        },
+        {
+          "name": "widthStorageKey",
+          "type": "string",
+          "defaultValue": "'sidebarWidth'",
+          "description": "localStorage key used to persist the sidebar width across sessions. Only applies when resizable is true."
+        },
```
francinelucca: mmm I'm not sure we'd want to copy this functionality 🤔 See https://github.com/primer/react/pull/7348#discussion_r2794629593

packages/react/src/PageLayout/PageLayout.docs.json
```diff
@@ -251,6 +269,81 @@
...
+          "description": "Divider style between the sidebar and the rest of the layout."
+        },
+        {
+          "name": "sticky",
+          "type": "boolean",
+          "defaultValue": "false",
+          "description": "Whether the sidebar sticks to the viewport when scrolling. When enabled, the sidebar uses position: sticky with top: 0 and height: 100vh."
+        },
+        {
+          "name": "whenNarrow",
```
francinelucca: how about "responsiveVariant" to match with ActionMenu, Overlay...

.playwright/snapshots/components/SplitPageLayout.test.ts-snapshots/SplitPageLayout-With-Sidebar-End-light-colorblind-linux.png
francinelucca, line 1: just making sure there's a way to make the sidebar/content/pane span the full height and have the footer at the bottom? 👀

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -918,6 +1039,239 @@ Pane.displayName = 'PageLayout.Pane'
...
+            dragStartWidthRef={dragStartWidthRef}
+            dragMaxWidthRef={dragMaxWidthRef}
+            getMaxPaneWidth={getMaxPaneWidth}
+            getDefaultWidth={getDefaultWidth}
+            saveWidth={saveWidth}
+          />
+        )}
+        <div
+          ref={sidebarRef}
+          suppressHydrationWarning
```
francinelucca: why do we need this? 👀

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -918,6 +1039,239 @@ Pane.displayName = 'PageLayout.Pane'
...
+            sidebarRef={sidebarRef}
+            dragStartClientXRef={dragStartClientXRef}
+            dragStartWidthRef={dragStartWidthRef}
+            dragMaxWidthRef={dragMaxWidthRef}
+            getMaxPaneWidth={getMaxPaneWidth}
+            getDefaultWidth={getDefaultWidth}
+            saveWidth={saveWidth}
+          />
+        )}
+      </div>
```
francinelucca, line 1268: can these be unified? seem pretty similar other than the children inside which could be conditionally rendered

packages/react/src/PageLayout/PageLayout.docs.json
TylerJDev: ~Yeah that's fair. I'm thinking we may want to add something here to replace it (within `Pane`), but for now we can remove it here.~

Actually, what if we kept it for now, and refactor both instances within Pane/Sidebar per this comment: 

> I'd like to eventually and instead export a persistInLocalStorage function that you can pass to onResizeEnd but it doesn't have to be now

Only because I realize that this is used (and might be needed) for now. I can take on the work to implement the suggestion though as a follow up!

packages/react/src/PageLayout/PageLayout.docs.json
TylerJDev: Yeah that works!

.playwright/snapshots/components/SplitPageLayout.test.ts-snapshots/SplitPageLayout-With-Sidebar-End-light-colorblind-linux.png
TylerJDev: Right now it's up to the consumer, or the content within, but most use cases would probably take the full height of the page (e.g. `100vh`), which works currently.

packages/react/src/PageLayout/PageLayout.tsx
TylerJDev: We could place them in a variable, but it would most likely be the same thing. We only render it this way because of the position in the DOM.

packages/react/src/PageLayout/PageLayout.tsx
TylerJDev: We use this in `pane` too, so it copied over. My assumption is that we'd probably run into the same thing here, that is: 

> Suppress hydration mismatch for --pane-width when localStorage, provides a width that differs from the server-rendered default.

I did make it conditional based on if it's resizable or not!

francinelucca (approved):

PR #7529, opened 2026-02-12
Add keyboard-accessible tooltip for truncated ActionList.Description

liuliu-dev:
Related issue https://github.com/github/accessibility-audits/issues/14802

### Changelog

When `ActionList.Description` is rendered with `truncate`, the full text was only visible via a native `title` attribute — which only appears on mouse hover, not keyboard focus (WCAG SC 2.1.1).

This PR adds a Primer `<Tooltip>` at the `ActionList.Item` level that shows the full description text on both hover and keyboard focus when the description is actually truncated.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/Description.tsx
```diff
@@ -43,6 +44,29 @@ export const Description: FCWithSlotMarker<React.PropsWithChildren<ActionListDes
...
+  // check on initial render
+  React.useEffect(() => {
+    if (!truncateEnabled || !containerRef.current) return
+    const el = containerRef.current
+    setTruncatedText(el.scrollWidth > el.clientWidth ? effectiveTitle : undefined)
+
+    return () => {
+      setTruncatedText(undefined)
+    }
+  }, [truncateEnabled, effectiveTitle, setTruncatedText])
```
TylerJDev: Is this logic only to determine if content is being truncated or not? Part of me wonders if there's a way we could manage this without an observer? One random thought was "guessing" if content was likely to be truncated based on size of the container + amount of text. 

Curious what you think about this @siddharthkp, as you had a lot of good insights with the UnderlineNav work (which is very much unrelated, but comes back to trying to figure out when content is "overflowing" or truncating).

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -35,6 +35,17 @@ export type TooltipProps = React.PropsWithChildren<{
...
+   * Private API for use internally only. Renders the tooltip popover element before
+   * the trigger in the DOM instead of after it.
+   *
+   * This is used by ActionList items to ensure the tooltip's popover `<span>` does not
+   * sit between the ActionListContent and its adjacent SubGroup sibling, which would
+   * break CSS adjacent sibling (`+`) selectors used for expand/collapse.
+   *
+   * @default false
+   */
+  _privateRenderBeforeTrigger?: boolean
```
TylerJDev: I know we talked about it already, but is this only because of the CSS styling we had for siblings? I wonder if it would be possible to add another style that would apply correctly if truncation was being used.

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -259,121 +283,147 @@ export const Tooltip: ForwardRefExoticComponent<
...
-          closeTooltip()
-        }
-      },
-      [isPopoverOpen],
-    )
+    useOnEscapePress(() => {
+      if (isPopoverOpen) {
+        closeTooltip()
+      }
+    }, [isPopoverOpen])
```
TylerJDev: I see we removed `event.stopImmediatePropagation()` and `event.preventDefault()`. I'm wondering if these were causing issues with the Tooltip and ActionList?

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -259,121 +283,147 @@ export const Tooltip: ForwardRefExoticComponent<
...
+              {triggerElement}
+            </>
+          ) : (
+            <>
+              {triggerElement}
+              {tooltipPopover}
+            </>
+          )}
         </>
       </TooltipContext.Provider>
```
TylerJDev, line 378: Remind me once more why we need the tooltip before the element in this case? Is this related to https://github.com/primer/react/pull/7529/changes#r2860941893?

packages/react/src/TooltipV2/Tooltip.tsx
liuliu-dev: Yes! this was due to the sibling CSS. I moved away from changing Tooltip DOM order and instead updated the collapse selectors from adjacent sibling (+) to general sibling (~)

packages/react/src/TooltipV2/Tooltip.tsx
liuliu-dev: Yes, those two lines were causing Overlay test failure. My understanding is that when Tooltip called `preventDefault()` and `stopImmediatePropagation()`, Overlay’s Escape handler could be skipped.  I added them back to reproduce the failure: https://github.com/primer/react/actions/runs/22458779284/job/65047083815?pr=7529.

packages/react/src/TooltipV2/Tooltip.tsx
liuliu-dev: Not anymore with the CSS selector update.

packages/react/src/TooltipV2/Tooltip.tsx
TylerJDev: I wonder if we could get away with the prop and conditional rendering if we adjust the style itself? This might come with its own challenges so I'm curious what you think!

packages/react/src/TooltipV2/Tooltip.tsx
TylerJDev: I'm wondering why that test would fail. I was thinking it was cause it was using `ActionList`, but I see there's no `ActionList` usage in the test 🤔 Do we know if the changes are affecting components outside of `ActionList`?

liuliu-dev: My current understanding is:
This is not `ActionList` related. Both `Tooltip` and `Overlay` use `useOnEscapePress`, and the test failure is exposed by [ref composition](https://github.com/primer/react/pull/7529/files#diff-54dcd2470acf4e9bba8c93e258045c10bd2552c5a3d43c7ba4034467c9f5a8edR278:~:text=const%20setTriggerRef%20%3D%20useMergedRefs%3CHTMLElement%20%7C%20null%3E(triggerRef%2C%20childRef%2C%20forwardedRef)) in `Tooltip`. Before ref composition, the cloned trigger could drop the child’s original ref. After `useMergedRefs`, when Tooltip calls `preventDefault()` / `stopImmediatePropagation()`, it can block `Overlay`’s handler in the shared `useOnEscapePress` chain.

packages/react/src/TooltipV2/Tooltip.tsx
liuliu-dev: yeah the prop is removed now and only CSS is used

packages/react/src/ActionList/Description.tsx
```diff
@@ -43,6 +44,29 @@ export const Description: FCWithSlotMarker<React.PropsWithChildren<ActionListDes
...
+  const truncateEnabled = truncate && !!setTruncatedText
+  useResizeObserver(
+    () => {
+      const el = containerRef.current
+      if (!el || !setTruncatedText) return
+      setTruncatedText(el.scrollWidth > el.clientWidth ? effectiveTitle : undefined)
+    },
+    containerRef,
+    [truncateEnabled, effectiveTitle],
+  )
```
TylerJDev: This will technically run on any ActionList description I'm guessing. I wonder if there's a way to scope this to only run if this is enabled.

Random thought, I wonder if we could calculate this once instead of an observer 🤔 
We'd know what the size of the text when rendered, could we also determine at what point it would likely overflow/be truncated?

packages/react/src/ActionList/ActionList.module.css
```diff
@@ -541,12 +541,12 @@
       transform: scaleY(1);
     }
 
-    & + .SubGroup {
+    & ~ .SubGroup {
```
francinelucca, line 544: curious about this change/why its needed 👀

liuliu-dev: This change fixes a NavList SubGroup collapse failure(https://github.com/primer/react/actions/runs/22242503448/job/64350808521?pr=7529).
The collapse styling used `+` to find `.SubGroup` right after `.ActionListContent`. When Tooltip wraps the trigger, it adds an extra element between them, so `.SubGroup` is no longer the immediate next sibling and the `+` selector does not match.

TylerJDev: Makes sense! My assumption is that there will only be one `.SubGroup` within the container, so there's no risk of it applying to other sub-groups that are also children of the same container.

TylerJDev:
Looks good! I noticed one small bug on the [truncation story](https://primer-eeb741cbba-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--text-wrap-and-truncation): 

https://github.com/user-attachments/assets/df16bc0d-dbd7-4a9f-aba3-049ad3f8ee3a

It seems that the third option doesn't load a tooltip, and the cursor changes rapidly 🤔

liuliu-dev:
> Looks good! I noticed one small bug on the [truncation story](https://primer-eeb741cbba-13348165.drafts.github.io/storybook/?path=/story/components-actionlist-features--text-wrap-and-truncation):
> 
>  Screen.Recording.2026-03-04.at.6.12.06.AM.mov 
> It seems that the third option doesn't load a tooltip, and the cursor changes rapidly 🤔

@TylerJDev  Good catch, thanks!

This was exposed after we removed `_privateDisableTooltip={!text}` in https://github.com/primer/react/pull/7529/commits/34888727591cda8f2364b906d1e7e33eccc4e871#diff-21bd620cf266fd1b3eb8d73399ee032c406b9bf49eac6f598ad65c9389ddb515. Without it, `ConditionalTooltip` would mount/unmount `<Tooltip>` based on whether `text` was set, and that mount/unmount cycle was causing an infinite re-render loop for complex (non-string) descriptions.

The root cause was two separate `useEffect`s in `Description.tsx`. the first extracted text content from the DOM and set `computedTitle`, which changed `effectiveTitle`, which triggered the second effect, which re-triggered the cycle. For simple string children this never happened because `effectiveTitle` was stable from the first render. I fixed it by merging the two effects into one effect that reads the DOM text and sets the truncatedText. 

One question: do you think we should also add `_privateDisableTooltip={!text}`back so the `Tooltip` wrapper stays mounted at all times(when it's enabled)? That way the DOM tree shape is stable between setting the text in Description.

TylerJDev:
> One question: do you think we should also add `_privateDisableTooltip={!text}`back so the `Tooltip` wrapper stays mounted at all times(when it's enabled)? That way the DOM tree shape is stable between setting the text in Description.

Thank you for the explanation! I guess with `_privateDisableTooltip`, are we only adding back the changes in the [commit you referred to](https://github.com/primer/react/commit/34888727591cda8f2364b906d1e7e33eccc4e871#diff-21bd620cf266fd1b3eb8d73399ee032c406b9bf49eac6f598ad65c9389ddb515)? Or are there other changes that would be included with this? 🤔

liuliu-dev:
> > One question: do you think we should also add `_privateDisableTooltip={!text}`back so the `Tooltip` wrapper stays mounted at all times(when it's enabled)? That way the DOM tree shape is stable between setting the text in Description.
> 
> Thank you for the explanation! I guess with `_privateDisableTooltip`, are we only adding back the changes in the [commit you referred to](https://github.com/primer/react/commit/34888727591cda8f2364b906d1e7e33eccc4e871#diff-21bd620cf266fd1b3eb8d73399ee032c406b9bf49eac6f598ad65c9389ddb515)? Or are there other changes that would be included with this? 🤔

This would be the change:
```diff
-  if (!enabled || !text) {
+  if (!enabled) {
      return children
    }
    return (
-     <Tooltip ref={forwardedRef} text={text || ''} direction="e" delay="medium">
+     <Tooltip ref={forwardedRef} text={text || ''} direction="e" delay="medium" _privateDisableTooltip={!text}>
        {children}
      </Tooltip>
    )
```

Since text is set asynchronously in a useEffect or non-string descriptions, using `if (!enabled || !text)` to conditionally mount/unmount the Tooltip causes a DOM tree change between renders. With `_privateDisableTooltip={!text}`, the `Tooltip` stays mounted but disabled when text is empty, so the tree is always stable.

TylerJDev:
> Since text is set asynchronously in a useEffect or non-string descriptions, using `if (!enabled || !text)` to conditionally mount/unmount the Tooltip causes a DOM tree change between renders. With `_privateDisableTooltip={!text}`, the `Tooltip` stays mounted but disabled when text is empty, so the tree is always stable.

Gotcha! We can go about adding it back then!

PR #7531, opened 2026-02-12
perf(SelectPanel): Built-in client-side list virtualization via `virtualized` prop

hectahertz:
See https://github.com/github/primer/discussions/6407

## Overview

Adds a `virtualized` boolean prop to `SelectPanel` (and `FilteredActionList`) that enables **client-side list virtualization** using `@tanstack/react-virtual` (already a dependency).

When enabled, only the visible items plus a small overscan buffer are rendered in the DOM. This is a purely client-side optimization — it does not require server-side pagination or API changes. The consumer can still pass all items at once.

### Usage

```tsx
<SelectPanel
  items={items}
  virtualized  // ← that's it
  // ... everything else unchanged
/>
```

### Performance measurements (1,800 items)

Measured using Chrome DevTools Performance traces and PerformanceObserver API on the `VirtualizedBuiltIn` story:

https://github.com/user-attachments/assets/6a89042d-f540-4aea-a58e-9b5c7a56c85f

#### Open time

| Metric | Without virtualization | With virtualization | Improvement |
|---|---|---|---|
| **Time to open** | **521 ms** | **10.6 ms** | **~49x faster** |
| DOM nodes (total) | 20,072 | 482 | **97.6% reduction** |
| `[role="option"]` elements | 1,800 | 19 | **98.9% reduction** |
| Worst Long Animation Frame | 1,192 ms (1,139 ms blocking) | None | **Eliminated** |

#### Filtering (typing "Item" with panel open)

| Metric | Without virtualization | With virtualization | Improvement |
|---|---|---|---|
| **INP** | **883 ms** (Bad) | **58 ms** (Good) | **~15x faster** |
| Worst keypress duration | 376 ms | 56 ms | **~7x faster** |
| Worst Long Animation Frame | 1,219 ms | None | **Eliminated** |
| DOM nodes during filtering | ~20,000 | ~504 | **97.5% reduction** |

#### New

- `virtualized` prop on `SelectPanel` and `FilteredActionList` — enables client-side list virtualization
- `VirtualizedBuiltIn` story — side-by-side comparison with open-time measurement

#### Changed

- Renamed existing consumer-side virtualization story to `VirtualizedConsumerSide` for clarity

#### Removed

- Nothing

### Rollout strategy

- [x] Minor release

### Testing & Reviewing

1. Open the `VirtualizedBuiltIn` story in Storybook (`Components/SelectPanel/Examples`)
2. Click each "Select labels" button and compare the displayed open times
3. Type into the filter field of each panel — the virtualized panel should feel instant while the non-virtualized one has noticeable lag
4. Scroll through the virtualized list to verify smooth scrolling and item measurement
5. Use keyboard navigation (arrow keys) to verify focus management works correctly with virtualized items

### Implementation details

- Uses `useVirtualizer` from `@tanstack/react-virtual` inside `FilteredActionList`
- Leverages the existing `scrollContainerRef` for the virtualizer scroll element
- Sets `estimateSize: 49px` with dynamic `measureElement` for accurate variable-height items
- `overscan: 10` items beyond the viewport for smooth scrolling
- Automatically sets `focusOutBehavior: "stop"` when virtualized to prevent focus wrapping past virtual boundaries
- Handles `scrollToIndex` in the focus zone when keyboard focus moves to an item outside the visible range
- Grouped lists (`groupMetadata`) are not virtualized — they are typically small enough not to need it

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -112,6 +113,18 @@ export interface FilteredActionListProps extends Partial<Omit<GroupedListProps,
...
+   * overscan buffer) are rendered in the DOM, dramatically improving performance for large lists.
+   *
+   * This is a purely client-side optimization — it does not require server-side pagination.
+   * The consumer can still pass all items at once; the component will only render what is visible.
+   *
+   * Recommended for lists with more than 100 items.
+   *
+   * @default false
+   */
+  virtualized?: boolean
```
francinelucca, line 130: TODO: 
- add to SelectPanel.docs.json

packages/react/src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -258,13 +272,22 @@ export function FilteredActionList({
       ? {
           containerRef: {current: listContainerElement},
           bindKeys: FocusKeys.ArrowVertical | FocusKeys.PageUpDown,
-          focusOutBehavior,
+          focusOutBehavior: virtualized ? 'stop' : focusOutBehavior,
```
francinelucca: curious why this is needed 👀

packages/react/src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -347,6 +395,80 @@ export function FilteredActionList({
...
+          }
+          return (
+            <ActionList.Group key={index}>
+              <ActionList.GroupHeading variant={group.header?.variant ? group.header.variant : undefined}>
+                {group.header?.title ? group.header.title : `Group ${group.groupId}`}
+              </ActionList.GroupHeading>
+              {getItemListForEachGroup(group.groupId).map(({key: itemKey, ...item}, itemIndex) => {
+                const key = itemKey ?? item.id?.toString() ?? itemIndex.toString()
+                return (
+                  <MappedActionListItem
```
francinelucca, line 437: does this mean the `virtualized` option won't work with groups? if so we should call that out in the prop info

packages/react/src/FilteredActionList/FilteredActionList.tsx
```diff
@@ -357,46 +479,18 @@ export function FilteredActionList({
...
-        {groupMetadata?.length
-          ? groupMetadata.map((group, index) => {
-              if (index === firstGroupIndex && getItemListForEachGroup(group.groupId).length === 0) {
-                firstGroupIndex++ // Increment firstGroupIndex if the first group has no items
+        style={
+          virtualized
+            ? {
+                height: virtualizer.getTotalSize(),
+                width: '100%',
+                position: 'relative' as const,
```
francinelucca, line 516: do the 'height' and 'width' SelectPanel props still work alongside these?

hectahertz: Yes, they work fine together. SelectPanel's height and width props are applied to the outer overlay (the AnchoredOverlay popover shell), which constrains the visible viewport of the panel. The virtualization styles here (height: virtualizer.getTotalSize(), position: relative) are applied to the inner <ActionList> (<ul>) nested inside the scroll container. I'll add a comment!

packages/react/src/FilteredActionList/FilteredActionList.tsx
hectahertz: Yes, virtualized has no effect when groupMetadata is provided. Grouped lists always render normally. This is intentional since grouped lists are typically small enough not to need virtualization.

Updated:
- JSDoc: The virtualized prop doc now explicitly calls out that it has no effect with groupMetadata.
- Dev warning: Added a __DEV__ console.warn when both virtualized and groupMetadata are set, so consumers get a clear heads-up instead of silently ignoring the prop.

packages/react/src/FilteredActionList/FilteredActionList.tsx
hectahertz: Yeah good question, this is not intuitive.

When virtualized, only ~19 items exist in the DOM at any time (out of 1,800).

With focusOutBehavior: 'wrap', arrowing past the last rendered item would wrap focus back to the first rendered item — cycling within the visible window instead of reaching item #1800. Forcing 'stop' prevents that loop and lets the scrollToIndex logic in onActiveDescendantChanged (just below) bring the correct items into view when keyboard navigation hits a boundary.

Added a comment explaining this.

hectahertz:
@TylerJDev @siddharthkp Could I get more eyes on this one before we merge it? :)

.changeset/selectpanel-built-in-virtualization.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": minor
+---
+
+Add built-in client-side list virtualization to `SelectPanel` via a new `virtualized` prop. When enabled, only the visible items plus a small overscan buffer are rendered in the DOM, dramatically improving performance for large lists.
```
siddharthkp: nit (optional)

```suggestion
SelectPanel: Add built-in client-side list virtualization via a new `virtualized` prop. When enabled, only the visible items plus a small overscan buffer are rendered in the DOM, dramatically improving performance for large lists.
```

hectahertz:
@siddharthkp Re: feature flag concern — I don't think we need to FF this since it's disabled by default. The `virtualized` boolean prop acts the same way: consumers opt in explicitly, so it can be incrementally adopted and tested without risk to existing usage.

I tested focus management manually and it works perfectly. Good instincts though, that was the right thing to flag.

Would you be ok with relying on this for now?

siddharthkp:
Sorry, missed your message! Sounds good to me

PR #7532, opened 2026-02-12
perf(PageLayout): eliminate ~614ms forced reflow from getComputedStyle on mount

hectahertz:
Closes #

`SplitPageLayout`'s `usePaneWidth` hook calls `getComputedStyle()` inside a `useLayoutEffect` on mount, forcing a synchronous layout. This is the single largest source of client-side latency across profiled Files controller routes on github.com.

**Impact on github.com (production profiling)**

| Route | `getPaneMaxWidthDiff` | Total forced reflow | Primer share |
|---|---|---|---|
| **Code View** (241-line file) | **614-621ms** | 745-758ms | **82%** |
| **Code View** (smaller file) | **124ms** | 370ms | **97%** (360ms combined Primer) |
| **Repo Overview** | - | **553ms** | **100%** (all Primer internals) |

On the Code View page, `getPaneMaxWidthDiff` alone forces the browser to synchronously lay out 430 pending DOM nodes in a single 529ms layout update.

**The fix**

The CSS variable `--pane-max-width-diff` only has two possible values, controlled by a single `@media (min-width: 1280px)` breakpoint:
- Below 1280px: `511px`
- At/above 1280px: `959px`

This is fully determinable from `window.innerWidth`, no `getComputedStyle` needed:

```ts
// Before (forces synchronous layout, 614ms on large pages)
maxWidthDiffRef.current = getPaneMaxWidthDiff(paneRef.current)

// After (property read + comparison, ~0ms)
maxWidthDiffRef.current = getMaxWidthDiffFromViewport()
```

The same replacement is applied to the resize handler's breakpoint-crossing path.

**Chrome DevTools Performance Trace Results**

Profiled on the **Heavy Content performance story** (~5,400 DOM nodes, 1,366 flex containers).

PageLayout-caused Forced Reflow:

| Scenario | Before (main) | After (this PR) | Improvement |
|---|---|---|---|
| 1x CPU | 262 ms | 0 ms | ✨ **-262 ms (eliminated)** ✨ |
| 4x CPU throttle | 1,136 ms | 0 ms | **-1,136 ms (eliminated)** |

Style Recalculation:

| Scenario | Before | After | Delta |
|---|---|---|---|
| 1x CPU | 228 ms (5,456 elements) | 110 ms (5,467 elements) | **-118 ms (-52%)** |
| 4x CPU throttle | 984 ms | 480 ms | **-504 ms (-51%)** |

LCP (4x CPU throttle):

| Metric | Before | After | Delta |
|---|---|---|---|
| LCP | 5,784 ms | 4,602 ms | **-1,182 ms (-20%)** |
| Render delay | 5,409 ms | 4,251 ms | **-1,158 ms** |

#### New

- `getMaxWidthDiffFromViewport()`: derives the `--pane-max-width-diff` value from `window.innerWidth` without touching the DOM

#### Changed

- Mount-time and resize-breakpoint-crossing initialization in `usePaneWidth` uses the viewport-based lookup instead of `getComputedStyle`
- Exported the wide-breakpoint diff value (`959`) from `PageLayout.module.css` via `:export` to keep JS and CSS in sync

#### Removed

N/A

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- All 46 existing `usePaneWidth` tests pass, with expectations corrected to reflect real-world breakpoint behavior at viewport >= 1280px
- 3 new tests added for `getMaxWidthDiffFromViewport` covering below, at, and above the 1280px breakpoint
- The old tests were inadvertently wrong: jsdom can't evaluate CSS media queries, so `getComputedStyle` always returned the default (511) even at 1280px viewport. The new viewport-based approach correctly returns 959 at that width.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/PageLayout/usePaneWidth.ts
+++ b/packages/react/src/PageLayout/usePaneWidth.ts
@@ -1,4 +1,5 @@
 import React, {startTransition, useMemo} from 'react'
+import {canUseDOM} from '../utils/environment'
 import useIsomorphicLayoutEffect from '../utils/useIsomorphicLayoutEffect'
 import cssExports from './PageLayout.module.css'
 


packages/react/src/PageLayout/usePaneWidth.test.ts
```diff
@@ -710,10 +430,10 @@ describe('usePaneWidth', () => {
         }),
       )
 
-      // Initial --pane-max-width should be set on mount
-      expect(refs.paneRef.current?.style.getPropertyValue('--pane-max-width')).toBe('769px')
+      // Initial --pane-max-width should be set on mount (1280 - 959 wide diff = 321)
+      expect(refs.paneRef.current?.style.getPropertyValue('--pane-max-width')).toBe('321px')
 
-      // Shrink viewport
+      // Shrink viewport (crosses 1280 breakpoint, diff switches to 511)
```
francinelucca, line 717: Making sure I understand: this value changed because it's correct from the get-go now?

packages/react/src/PageLayout/usePaneWidth.ts
```diff
@@ -110,6 +104,16 @@ export function getPaneMaxWidthDiff(paneElement: HTMLElement | null): number {
   return value > 0 ? value : DEFAULT_MAX_WIDTH_DIFF
 }
```
francinelucca: are we able to delete this function then? 👀

packages/react/src/PageLayout/PageLayout.module.css
francinelucca, line 35: I think we don't need this variable anymore then 👀

packages/react/src/PageLayout/usePaneWidth.test.ts
```diff
@@ -989,14 +709,14 @@ describe('usePaneWidth', () => {
...
+      expect(result.current.getMaxPaneWidth()).toBe(321)
 
-      // Viewport changes (no resize event needed)
+      // Viewport changes (no resize event fired, so maxWidthDiffRef stays at 959)
       vi.stubGlobal('innerWidth', 800)
 
-      // getMaxPaneWidth reads window.innerWidth dynamically
-      expect(result.current.getMaxPaneWidth()).toBe(289)
+      // getMaxPaneWidth reads window.innerWidth dynamically: max(256, 800 - 959) = 256
+      expect(result.current.getMaxPaneWidth()).toBe(256)
```
francinelucca: could we dispatch the resized event here instead so the value updates? otherwise I'm not sure what this part of the test is achieving 😅

packages/react/src/PageLayout/usePaneWidth.test.ts
hectahertz: Yes! The old expectations were wrong. jsdom can't evaluate CSS media queries, so getComputedStyle always returned the below breakpoint default (511) even when the test set innerWidth to 1280. That gave us 1280 - 511 = 769, which looked plausible but didn't match what a real browser would do.

At innerWidth >= 1280, the CSS switches --pane-max-width-diff to 959, so the correct result is 1280 - 959 = 321. The new viewport-based approach gets this right from mount, and the test expectations now match real browser behavior.

*phew*

packages/react/src/PageLayout/PageLayout.module.css
hectahertz: We still need the :export values — both paneMaxWidthDiffDefault (511) and paneMaxWidthDiffWide (959) are consumed by the JS constants DEFAULT_MAX_WIDTH_DIFF and WIDE_MAX_WIDTH_DIFF in usePaneWidth.ts. They're what keep the JS viewport check in sync with the CSS breakpoint values.

If you're thinking of the --pane-max-width-diff CSS custom property on .PageLayoutRoot (line 33) — that one is no longer read by JS via getComputedStyle, but it's still used by the browser for CSS-side layout calc(). So both the :export block and the custom property need to stay.

hectahertz:
@francinelucca everything should be addressed now!

liuliu-dev (approved):
Nice fix! Learned some tricks in this pr 💖

francinelucca (approved):
✨

PR #7544, opened 2026-02-14
perf(TreeView): replace O(n) TreeWalker with O(depth) sibling traversal

hectahertz:
Closes #

Replaces the O(n) `TreeWalker`-based navigation in `getVisibleElement` with O(depth) direct DOM sibling/parent traversal for ArrowUp/ArrowDown keyboard navigation.

The previous implementation created a fresh `document.createTreeWalker(root, ...)` on every ArrowUp/ArrowDown keystroke, then walked from the first node in the tree to the currently focused element before continuing. At position 25K in a 26K-item tree, this was essentially a full DOM traversal on every keypress.

The new implementation walks sibling and parent/child edges directly:
- **ArrowDown**: if expanded, go to first child; else next sibling; if no next sibling, walk up to parent's next sibling. O(depth).
- **ArrowUp**: go to previous sibling's deepest last visible descendant; if no previous sibling, go to parent. O(depth).

This relies on `TreeView.SubTree` unmounting children when collapsed (`return null` when `!isExpanded`), so collapsed subtree children are never in the DOM.

### Benchmarks (26,000 items, 289K DOM nodes)

**Per-operation cost (single call):**

| Position in tree | Old (TreeWalker) | New (sibling traversal) |
|---|---|---|
| 100 | 0.5ms | <0.01ms |
| 2,000 | 5.1ms | <0.01ms |
| 10,000 | 21.6ms | <0.01ms |
| 25,000 | 55.6ms | <0.01ms |

**Held key simulation (50 rapid ArrowDown from position 12,500):**

| Metric | Old (TreeWalker) | New (sibling traversal) |
|---|---|---|
| ✨ **Total time** ✨ | **1,878ms** | **357ms** |
| Avg per keypress | 37.6ms | 7.1ms |
| Bottleneck | TreeWalker O(n) traversal | `scrollIntoView` forced reflow (separate issue) |

The remaining ~7ms per keypress is dominated by `scrollIntoView` forcing synchronous layout reflow on 289K DOM nodes, addressed separately in #7545.

#### New

N/A

#### Changed

- `getVisibleElement` in `useRovingTabIndex.ts` now uses direct DOM sibling/parent traversal instead of `TreeWalker`, reducing ArrowUp/ArrowDown navigation from O(n) to O(tree depth)

#### Removed

N/A

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- All 51 existing TreeView tests pass unchanged
- To manually verify: open the [StressTest story](http://localhost:6006/?path=/story/components-treeview-features--stress-test), expand several directories, and use ArrowUp/ArrowDown to navigate. Focus should move correctly through expanded subtrees and skip collapsed ones.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TreeView/useRovingTabIndex.ts
```diff
@@ -142,30 +142,109 @@ export function getElementState(element: HTMLElement): 'open' | 'closed' | 'end'
...
+ * PERFORMANCE: This is O(tree depth) instead of O(n) because it walks
+ * siblings and parent/child edges directly, rather than creating a TreeWalker
+ * that scans from the root to find the current element on every keystroke.
+ *
+ * NOTE: This relies on TreeView.SubTree unmounting its children when collapsed
+ * (returning null when !isExpanded). Because collapsed subtree children are
+ * never in the DOM, we can safely skip them by only entering children of nodes
+ * with aria-expanded="true". If SubTree ever changes to keep collapsed children
+ * mounted (e.g. via CSS display:none), this logic would need to add filtering
+ * for items inside collapsed parents.
```
liuliu-dev, line 157: This approach relies on `SubTree` returning `null` when `!isExpanded` ([TreeView.tsx:557](https://github.com/primer/react/blob/cf8374191726b084cbeebd6a96a5c47db8f3a46d/packages/react/src/TreeView/TreeView.tsx#L557)). Should we add a unit test that asserts collapsed subtree children are not present in the DOM? That way if `SubTree`'s behavior ever changes, we'd get a test failure

packages/react/src/TreeView/useRovingTabIndex.ts
```diff
@@ -142,30 +142,109 @@ export function getElementState(element: HTMLElement): 'open' | 'closed' | 'end'
...
+  // No previous sibling, the parent is the previous visible element
+  return getParentElement(element)
+}
+
+/**
+ * Walk into expanded subtrees to find the deepest last visible descendant.
+ * For example, if the last sibling is an expanded directory whose last child
+ * is also an expanded directory, we drill all the way down.
+ */
+function getDeepestLastDescendant(element: HTMLElement): HTMLElement {
```
liuliu-dev, line 203: would it be better to make `getDeepestLastDescendant` iterative instead of recursive? 👀

packages/react/src/TreeView/useRovingTabIndex.ts
hectahertz: Good call! Added a test that explicitly asserts collapsed subtree children are not queryable in the DOM: ed7367675

packages/react/src/TreeView/useRovingTabIndex.ts
hectahertz: Converted to iterative in the same commit. Depth is bounded by tree depth so it was safe either way, but iterative reads cleaner here.

liuliu-dev:
I’m seeing this [failure](https://github.com/github/github-ui/actions/runs/22318136106/job/64568786485?pr=14391) in the integration tests:

```
❌ > nx run @github-ui/nested-list-view:test --runInBand
  
  
  > @github-ui/nested-list-view@1.0.0 test
  > jest --runInBand
  
  PASS src/NestedListItem/__tests__/NestedListItem.test.tsx (24.015 s)
   LOGGING RETRY ERRORS  Nested ListView focus management allows arrow navigation for list items
   RETRY 1 
  
      expect(element).not.toHaveAttribute("data-focus-visible-added") // element.hasAttribute("data-focus-visible-added")
  
      Expected the element not to have attribute:
        data-focus-visible-added
      Received:
        data-focus-visible-added=""
  
        17 |
        18 |     expect(items[0]).toHaveAttribute('data-focus-visible-added')
      > 19 |
           | ^
        20 |     await user.keyboard('[ArrowDown]')
        21 |
        22 |     expect(items[0]).not.toHaveAttribute('data-focus-visible-added')
  
        at Object.toHaveAttribute (src/__tests__/NestedListViewFocusManagement.test.tsx:19:26)
 ```       

It’s happening across multiple integration test runs for this PR. Could this be related to this change (since it affects tab/focus behavior)?

hectahertz:
@liuliu-dev quick follow-up on your integration-failure note: I checked the linked integration PR (github/github-ui#14853), and the remaining failure is still in Test - Packages 4 (nested-list-view focus-visible assertion), with aggregate Tests red because of that. I just reran failed jobs for run 22579072258 to verify whether this is flaky or persistent. I will report back once rerun finishes.

PR #7546, opened 2026-02-14
perf(Announce): skip getComputedStyle when there is no text content to announce

hectahertz:
Closes #

`Announce` calls `window.getComputedStyle(element)` to check if the element is visible before announcing. This forces synchronous style recalculation for the entire DOM. When `Announce` renders with initially empty content (e.g. TreeView's `AriaStatus` live region, which starts with `ariaLiveMessage = ''`), this reflow is completely wasted.

The fix: check `getTextContent(element)` first. If empty, bail out before the expensive `getComputedStyle` call.

### Benchmarks (TreeView StressTest, 1000 items, 14K DOM nodes)

| Metric | Before | After |
|---|---|---|
| LCP | 2,398ms | 1,952ms |
| Render delay | 2,225ms | 1,806ms |
| `Announce.tsx` forced reflow | 294ms | ✨ **0ms** ✨ |

#### New

N/A

#### Changed

- `Announce` checks for empty text content before calling `getComputedStyle`, avoiding a forced synchronous style recalculation on mount when there's nothing to announce

#### Removed

N/A

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- All 71 live-region and TreeView tests pass
- Updated one TreeView test that assumed `live-region` element would be created on initial empty-content mount. The element is now lazily created on the first real announcement instead.
- To verify: open the [StressTest story](http://localhost:6006/?path=/story/components-treeview-features--stress-test), run a Performance trace with reload, confirm no `Announce.tsx` forced reflow appears.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/banner-aria-labelledby-title.md
+++ b/.changeset/banner-aria-labelledby-title.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': minor
+---
+
+Banner: Use `aria-labelledby` to reference the title for the landmark region instead of generic variant labels.
--- a/.changeset/clever-geese-cover.md
+++ b/.changeset/clever-geese-cover.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+UnderlineNav: Adds `overflow: hidden` when the "More" button isn't present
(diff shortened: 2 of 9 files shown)


siddharthkp (approved):
Surprised this makes that much of a difference!

hectahertz:
@siddharthkp I know, right?

PR #7550, opened 2026-02-15
perf(Spinner): replace Web Animations API with CSS animation-delay sync

hectahertz:
Closes #

Replaces the Web Animations API-based spinner synchronization with a pure CSS `animation-delay` approach, eliminating a Safari/WebKit performance bottleneck.

The previous sync mechanism called three Web Animations API methods per Spinner mount: `element.getAnimations()` (3-5x slower in WebKit than Chromium), `CSSAnimation.pause()` + `element.animate()` (replaces native CSS animation with a JS-driven one, often falling back to main-thread compositing in WebKit), and `Animation.startTime` (forces WebKit to recalculate timing for all document animations). This resulted in 2 animations per SVG and a global `useSyncExternalStore` store that re-rendered all mounted Spinners when the first one set its `startTime`.

The new approach computes a negative CSS `animation-delay` from `performance.now()` at visibility time:

```ts
animationDelay = -(performance.now() % 1000)ms
```

Since all instances reference the same monotonic clock, they land at the same rotation angle regardless of mount time. The CSS animation engine handles everything natively (GPU-composited), with 1 animation per SVG and zero JS animation overhead. The delay is computed at visibility time (not mount time), so spinners using the `delay` prop don't flash at the wrong angle on the first frame.

### Measurements (5 spinners, staggered mount)

| | Flag OFF | Flag ON |
|---|---|---|
| **main** (Web Animations API) | 276° spread, 1 anim/SVG | 0° spread, **2 anim/SVG** |
| **This PR** (CSS animation-delay) | 276° spread, 1 anim/SVG | ✨ **4° spread, 1 anim/SVG** ✨ |

The 4° spread is floating-point rounding from `performance.now()`, visually imperceptible.

#### New

N/A

#### Changed

- Spinner animation synchronization now uses CSS `animation-delay` instead of the Web Animations API

#### Removed

N/A

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

1. Enable `primer_react_spinner_synchronize_animations` feature flag in Storybook toolbar
2. Open the "Synchronized Spinners" example story
3. Verify all spinners rotate in sync as they appear
4. Toggle `prefers-reduced-motion: reduce` in OS settings, confirm sync is skipped
5. Test in both Safari and Chrome

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [x] Tested in Safari
- [ ] Tested in Edge


hectahertz:
Friendly ping for review: @siddharthkp @TylerJDev 🙏

siddharthkp:
Hi! 

@joshblack had tried to implement this with pure css first, but it never synced properly. 

Worth verifying if this PR has the same issues not:
- [primer/react/7157#issuecomment-3524052962](https://github.com/primer/react/pull/7157#issuecomment-3524052962)
- [primer/react/7157#issuecomment-3628689679](https://github.com/primer/react/pull/7157#issuecomment-3628689679)

hectahertz:
@siddharthkp Checked both threads. The two concerns were:

1. Josh saw spinners "off a bit" when slowing down a CSS `animation-delay` prototype
2. MDN says "it is impossible to sync two separate animations with CSS animations"

Neither applies here. The MDN quote is about `Animation.startTime` manipulation, not `animation-delay`. This approach uses a different mechanism: each spinner computes `-(performance.now() % 1000)ms` at visibility time, so at any wall-clock time `N` its position resolves to `N % 1000` regardless of mount time.

The github-ui #6120 implementation used `Date.now()` instead of `performance.now()`. CSS animations run on the document timeline which is `performance.now()`-based, so using a different clock introduces drift. This PR matches the correct clock.

The ~4° spread I measured comes from the render-to-paint gap (up to 16ms at 60fps), visually imperceptible at 1 rotation/sec.

PR #7554, opened 2026-02-15
perf(ActionList): memoize context values, menuItemProps, aria attributes

hectahertz:
Closes #

Every parent re-render created 8 new objects per ActionList.Item, including new context values that prevented React from bailing out of child renders. This stabilizes context values and computed props.

Changes:
- Memoize `ListContext.Provider` and `ItemContext.Provider` values with `useMemo`
- Memoize `menuItemProps` object (rebuilt per item per render)
- Memoize `aria-labelledby` and `aria-describedby` computation
- Hoist `baseSlots`, `slotsConfig`, `selectableRoles`, `listRoleTypes` to module scope (were recreated per item per render)
- Replace array allocation per keypress with direct comparison
- Convert mutable `let` to `const` ternary (React Compiler compatibility)

### Metrics

**Object allocation reduction** (microbenchmark, Chrome):

| Metric | Old | New (deps unchanged) | New (deps changed) |
|---|---|---|---|
| Objects per Item render | 8 | 0 | 3 |
| Objects per 100-item re-render | 800 | 0 | 300 |
| Objects per 100 re-renders x 100 items | 80,000 | 0 | 30,000 |
| Allocation time (10K items) | 4.70 ms | 0.10 ms | - |

#### New

N/A

#### Changed

- ActionList: memoized context provider values, menuItemProps, and aria attribute computation
- ActionList.Item: hoisted static config objects to module scope

#### Removed

N/A

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

1. Open the ActionList stress test in Storybook (`StressTests/Components/ActionList`)
2. All 115 ActionList, ActionMenu, and NavList tests pass

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionList/Item.tsx
```diff
@@ -47,6 +47,19 @@ const DivItemContainerNoBox = React.forwardRef<HTMLDivElement, React.HTMLAttribu
...
+const baseSlots = {
+  leadingVisual: LeadingVisual,
+  trailingVisual: TrailingVisual,
+  trailingAction: TrailingAction,
+  subItem: SubItem,
+}
+
+const slotsConfig = {...baseSlots, description: Description}
+
+// Pre-allocated array for selectableRoles check, avoids per-render allocation
```
hectahertz, line 82: **Hoist static objects to module scope**

`baseSlots`, `slotsConfig`, `selectableRoles`, and `listRoleTypes` are constant across all renders and all Item instances. Previously they were recreated inside the component body on every render for every item. Moving them to module scope eliminates N object allocations per render (where N = number of items in the list).

packages/react/src/ActionList/Item.tsx
```diff
@@ -69,14 +82,7 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
 ): JSX.Element => {
-  const baseSlots = {
-    leadingVisual: LeadingVisual,
-    trailingVisual: TrailingVisual,
-    trailingAction: TrailingAction,
-    subItem: SubItem,
-  }
-
-  const [partialSlots, childrenWithoutSlots] = useSlots(props.children, {...baseSlots, description: Description})
+  const [partialSlots, childrenWithoutSlots] = useSlots(props.children, slotsConfig)
```
hectahertz, line 108: **Pass stable slot config reference**

Previously: `useSlots(props.children, {...baseSlots, description: Description})` created a new object via spread on every render. Now it receives the stable module-level `slotsConfig`.

packages/react/src/ActionList/Item.tsx
```diff
@@ -165,7 +170,7 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
   const keyPressHandler = React.useCallback(
     (event: React.KeyboardEvent<HTMLElement>) => {
       if (disabled || inactive || loading) return
-      if ([' ', 'Enter'].includes(event.key)) {
+      if (event.key === ' ' || event.key === 'Enter') {
```
hectahertz, line 196: **Avoid array allocation on keypress**

Replaced `[' ', 'Enter'].includes(event.key)` with direct `===` comparison. The array was allocated on every keypress event handler invocation.

packages/react/src/ActionList/Item.tsx
```diff
@@ -189,41 +194,61 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
-  // only apply aria-selected and aria-checked to selectable items
-  const selectableRoles = ['menuitemradio', 'menuitemcheckbox', 'option']
   const includeSelectionAttribute = itemSelectionAttribute && itemRole && selectableRoles.includes(itemRole)
 
-  let focusable
-
-  if (showInactiveIndicator) {
-    focusable = true
-  }
+  const focusable = showInactiveIndicator ? true : undefined
```
hectahertz, line 224: **Convert mutable `let` to `const` ternary**

The old `let focusable` + `if` block was flagged by the React Compiler as a mutable dependency that couldn't be preserved in memoization. Converting to a `const` ternary fixes React Compiler compatibility and is more concise.

packages/react/src/ActionList/Item.tsx
```diff
@@ -189,41 +194,61 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
-        inactiveWarningId ?? undefined,
-      ]
-        .filter(String)
-        .join(' ')
-        .trim() || undefined,
-    ...(includeSelectionAttribute && {[itemSelectionAttribute]: selected}),
-    role: itemRole,
-    id: itemId,
-  }
+  const hasTrailingVisualSlot = Boolean(slots.trailingVisual)
```
hectahertz, line 229: **Memoize aria attribute strings**

Previously, `aria-labelledby` was built with template literals that always produced a new string (including trailing spaces from empty slots). `aria-describedby` used `.filter(String).join(' ').trim()` which allocates intermediate arrays on every render.

Now both are computed in `useMemo` with explicit conditionals. Boolean deps (`hasTrailingVisualSlot`, `hasDescriptionSlot`) are derived above to keep the dep array stable and avoid depending on slot object references.

packages/react/src/ActionList/Item.tsx
```diff
@@ -189,41 +194,61 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
...
+  const ariaLabelledBy = React.useMemo(() => {
+    const parts = [labelId]
+    if (hasTrailingVisualSlot) parts.push(trailingVisualId)
+    if (hasDescriptionSlot && descriptionVariant === 'inline') parts.push(inlineDescriptionId)
+    return parts.join(' ')
+  }, [labelId, hasTrailingVisualSlot, trailingVisualId, hasDescriptionSlot, descriptionVariant, inlineDescriptionId])
+
+  const ariaDescribedBy = React.useMemo(() => {
+    const parts: string[] = []
+    if (hasDescriptionSlot && descriptionVariant === 'block') parts.push(blockDescriptionId)
```
hectahertz, line 241: **Memoize `menuItemProps` object**

This object is created per item and passed to either the `<li>` (via `containerProps`) or the `<ItemWrapper>` (via `wrapperProps`). Previously rebuilt on every render even when no values changed. Now memoized with all relevant deps.

Note: this has a large dep array (14 deps) because the object touches many item-level values. The memoization still pays off because in typical usage most deps are stable across re-renders (e.g. `clickHandler`, `keyPressHandler`, `itemRole`, `itemId` rarely change).

packages/react/src/ActionList/Item.tsx
```diff
@@ -238,18 +263,21 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
         ref: forwardedRef,
       }
```
hectahertz, line 290: **Memoize `ItemContext.Provider` value**

Previously, the inline object passed to `ItemContext.Provider` was recreated on every render, causing all context consumers (`Description`, `LeadingVisual`, `TrailingVisual`, `Selection`) to re-render even when their relevant values hadn't changed.

Now wrapped in `useMemo` with 7 deps. In a 100-item list, this prevents up to ~400 unnecessary child component re-renders per parent state change.

packages/react/src/ActionList/List.tsx
```diff
@@ -55,16 +55,19 @@ const UnwrappedList = <As extends React.ElementType = 'ul'>(
       listRole === 'menu' || container === 'SelectPanel' || container === 'FilteredActionList' ? 'wrap' : undefined,
   })
 
+  const listContextValue = React.useMemo(
```
hectahertz, line 58: **Memoize `ListContext.Provider` value**

Same pattern as `ItemContext`. The inline object was recreated on every `ActionList` render, causing all `ActionList.Item` children to re-render due to context referential inequality, even when `variant`, `selectionVariant`, `showDividers`, `role`, and `headingId` hadn't changed.

hectahertz:
Friendly ping for review: @siddharthkp @TylerJDev 🙏

PR #7567, opened 2026-02-18
Autocomplete story SR fix

llastflowers:
Addresses https://github.com/github/accessibility-audits/issues/14844

#### Changed

Updated `InOverlayWithCustomScrollContainerRef` story to address the issue with the label information not being announced by SR.

While testing my changes, I also noticed a bug with the [`InOverlayWithCustomScrollContainerRef`](https://primer.style/react/storybook/?path=/story/components-autocomplete-features--in-overlay-with-custom-scroll-container-ref) and [`RenderingTheMenuOutsideAnOverlay`](https://primer.style/react/storybook/?path=/story/components-autocomplete-features--rendering-the-menu-outside-an-overlay) stories: It seems that not using `Autocomplete.Overlay` causes the first `item` in the list (in this case, "css") to be set to active as soon as the input is focused.

This is buggy/confusing for 2 reasons: 
1. SR announces "css" as soon as the input is focused, and 
3. The normal arrow-down functionality is broken: since "css" is invisibly active, when a user presses the arrow-down key, the second `item` is highlighted first ("css-in-js"). 

This issue might have larger a11y implications within the component itself that could be worth investigating, but I didn't want to get too far out of scope here. For now, I added some checks in these two stories to prevent the menu from invisibly activating until the menu has been opened by typing in the input or pressing arrow keys.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/gentle-onions-fix.md
+++ b/.changeset/gentle-onions-fix.md
@@ -1,5 +1,0 @@
----
-"@primer/react": patch
----
-
-Autocomplete story SR fix


llastflowers:
Skipping integration tests: Storybook-only changes in this PR

hectahertz (commented):
Nice catch on the active-descendant-on-focus bug. I dug into the internals to understand the root cause, here's what I found and some suggestions.

## Root cause

The bug is in `AutocompleteMenu.tsx`, not in the stories. The `useFocusZone` hook (line ~289) activates the first item as active descendant as soon as it initializes, regardless of whether the menu is actually visible (`showMenu` state). 

In overlay mode (`Autocomplete.Overlay`), this is masked because the menu children get remounted when toggling between `<Overlay>` and `<VisuallyHidden>` (the HACK at the bottom of `AutocompleteOverlay.tsx`), so `useFocusZone` only initializes when the menu becomes visible.

Without `Autocomplete.Overlay` (or with a custom overlay like `AnchoredOverlay`), the menu DOM is always mounted and `useFocusZone` runs immediately, activating the first item before the user has interacted with the menu.

## Suggested component-level fix

`useFocusZone` already supports a `disabled` option. Passing `disabled: !showMenu` in the `useFocusZone` call in `AutocompleteMenu.tsx` would fix this at the source:

```diff
  useFocusZone(
    {
      containerRef: listContainerRef,
+     disabled: !showMenu,
      focusOutBehavior: 'wrap',
```

When `showMenu` transitions to `true` (via typing or arrow keys), the focus zone enables and highlights the first matching item, which is the expected behavior. This would fix both stories without needing the conditional rendering workarounds.

Worth considering as part of this PR or as a follow-up, since fixing the component is better than working around it in stories.

## On the "Menu without Overlay" question

Using `Autocomplete.Menu` without `Autocomplete.Overlay` is a documented pattern (the `RenderingTheMenuOutsideAnOverlay` story exists for that reason). It's also reasonable to use a custom overlay like `AnchoredOverlay` as in `InOverlayWithCustomScrollContainerRef`. I don't think we should discourage these patterns. The issue is the component bug, not the pattern itself.

## Story changes

The story-level workarounds are reasonable if we want to ship them ahead of a component fix:

- **`RenderingTheMenuOutsideAnOverlay`**: conditionally rendering the menu on `query.length > 0` works but loses arrow-key-to-open behavior (menu won't mount until text is typed). Minor since it's a story, but worth noting.
- **`InOverlayWithCustomScrollContainerRef`**: the `menuOpen` state tracking with arrow/escape handling is more thorough. The `aria-describedby` addition on the button is a nice touch.

One small thing: `handleOpen` no longer calls `inputRef.current.focus()`. This should be fine since `focusTrapSettings={{initialFocusRef: inputRef}}` handles it, but wanted to flag it in case the explicit call was there for a timing reason.

llastflowers:
> Nice catch on the active-descendant-on-focus bug. I dug into the internals to understand the root cause, here's what I found and some suggestions.

Thanks for the writeup and suggestions! Since this is just for an a11y audit bug fix, I would rather avoid any unnecessary component-level changes in this PR. It sounds like it might be easier then to just leave the other storybook bugs for now and follow up on it later. I have removed everything except for the original a11y fix.

.changeset/gentle-onions-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Autocomplete story SR fix
```
siddharthkp: We don't need a changelog entry for a storybook change :)

You can add the `skip-changeset` label to get CI green

PR #7571, opened 2026-02-19
fix Overlay: Change default overflow from hidden to auto

liuliu-dev:
Related issue https://github.com/github/accessibility-audits/issues/14861

### Changelog

The `Overlay` component sets `max-height: 100vh` with `overflow: hidden` by default. At small viewports (e.g. 320×256px), this causes overlay content to be clipped. This change updates the default `overflow` from `hidden` to `auto`
 
**Before:**

https://github.com/user-attachments/assets/b2302350-7655-4fa5-a11c-b471284684ce

**After:**

https://github.com/user-attachments/assets/df2c6dfd-e7c2-4a45-8788-2c7d7381af85

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp (commented):
Seems intuitive to me, but..

git blame led me to this PR from 5 years ago when this was added, something worth testing if we are bringing back the visual bugs this had solved: https://github.com/primer/react/pull/1248

Commenting for now, will wait for your response before approving

liuliu-dev:
> Seems intuitive to me, but..
> 
> git blame led me to this PR from 5 years ago when this was added, something worth testing if we are bringing back the visual bugs this had solved: #1248
> 
> Commenting for now, will wait for your response before approving

@siddharthkp Thanks for the heads up!

I don't think this fix will bring back the bug.

The key difference between what #1248 reverted and this fix is `overflow: visible` vs `overflow: auto`. When`overflow: hidden` was removed (reverting to `overflow: visible`), which does not create a clipping context, that's what caused content to past the border-radius corners.

Both `hidden` and `auto` create a clipping context, so the border-radius corner issue does not come back.

I tested with the [ActionMenu MultipleSections story](https://primer.style/react/storybook/?path=/story/components-actionmenu-examples--multiple-sections). When resizing the viewport to a smaller height, the menu list gets cut off. I recorded the difference between `overflow: auto`, `overflow: hidden`, and removing overflow.


https://github.com/user-attachments/assets/d2f8af6d-8f50-4237-8a5a-dde9289444c0

siddharthkp:
Valid, can we check how Overlays with controls behave? For example: ConfirmationDialog, SelectPanel with footer

liuliu-dev:
@siddharthkp I checked both:

`ConfirmationDialog` doesn't use the `Overlay` component. It renders via `Dialog`, which has its own `Portal` and CSS with `ScrollableRegion` handling overflow.

`SelectPanel` does use `Overlay` via `AnchoredOverlay`, the `ActionList` inside the panel scrolls. I tested the `MultiSelectModal` story at small viewports and they work correctly with `overflow: auto`:

https://github.com/user-attachments/assets/e19f2474-844c-4596-a702-2660bec6ea19

PR #7576, opened 2026-02-20
ADR-023: Public data-component API for targeting component parts

lukasoppermann:
Adds ADR-023 proposing `data-component` as a public, stable API for identifying component parts in the DOM. This gives consumers reliable selectors to target internal parts of components for style overrides and JS queries, independent of CSS Module hash changes.

Key decisions:
- Dot notation mirroring the React API (`ActionList.Item`, `Button.LeadingVisual`)
- State/modifier attrs (`data-variant`, `data-size`) remain separate
- Includes migration table for existing inconsistent values
- Complements CSS Layers (ADR-021) for a complete override mechanism

### DIFF
--- a/contributor-docs/adrs/adr-023-stable-selectors-api.md
+++ b/contributor-docs/adrs/adr-023-stable-selectors-api.md
@@ -6,7 +6,7 @@
 
 | Stage          | State       |
 | -------------- | ----------- |
-| Status         | Proposed ❓ |
+| Status         | Accepted ✅ |
 | Implementation | Pending ⚠️  |
 
 ## Context
@@ -49,17 +49,16 @@ without notice and coverage is incomplete — many component parts have no
 
 ## Decision
 
-Establish two **public, stable data attributes** that act as reliable
-identifiers for Primer components and their parts in the DOM. Think of these as
-built-in test IDs — stable, predictable anchors that consumers can use for
+Establish a **public, stable data attribute** that acts as a reliable
+identifier for Primer components in the DOM. Think of this as
+a built-in test ID — a stable, predictable anchor that consumers can use for
 testing, tracking, monitoring, and querying without coupling to internal DOM
 structure.
 
-- **`data-component`** — identifies the root element of a component or
-  sub-component.
-- **`data-part`** — identifies an inner structural part within a component.
+- **`data-component`** — identifies the root element of a component,
+  sub-component, or internal part.
 
-These attributes are primarily intended for:
+This attribute is primarily intended for:
 
 1. **Testing** — use them as locators in unit tests, integration tests, and
    end-to-end tests instead of brittle class names or DOM paths.
@@ -75,11 +74,10 @@ These attributes are primarily intended for:
 
 #### Naming convention
 
-All values use PascalCase. The two attributes serve distinct roles:
+All values use PascalCase:
 
 ```
-data-component="ComponentName"     → root element of a component or sub-component
-data-part="PartName"               → inner part within a component
+data-component="ComponentName"     → root element of a component, sub-component, or internal part
 ```
 
 ##### Rules
@@ -98,58 +96,44 @@ data-part="PartName"               → inner part within a component
    <li data-component="ActionList.Item"></li>
    ```
 
-   Note: a sub-component root uses `data-component`, not `data-part`, because it
-   is itself a component — it has its own props, its own identity, and may
-   contain its own slots.
-
 3. **Inner structural parts** (DOM elements that are not exposed as a
    sub-component but represent a meaningful part of the structure) get
-   `data-part` with a PascalCase name describing the part.
+   `data-component` with the parent component name and a PascalCase part name.
 
    ```html
-   <span data-part="Label">monalisa</span>
-   <span data-part="Content">...</span>
-   <span data-part="LeadingVisual"><img /></span>
+   <span data-component="ActionList.Item.Label">monalisa</span>
+   <span data-component="ActionList.Item.Content">...</span>
+   <span data-component="ActionList.Item.LeadingVisual"><img /></span>
    ```
 
-   `data-part` names are **scoped to their parent component** — a `Label` part inside
-   `ActionList.Item` is distinct from a `Label` part inside `Button` because
-   they exist within different `data-component` boundaries.
+   Part names are scoped to their parent component via the naming convention —
+   `ActionList.Item.Label` is distinct from `Button.Label`.
 
-4. **State and modifier attributes remain separate.** `data-component` and
-   `data-part` identify _what_ an element is. Existing attributes like
+4. **State and modifier attributes remain separate.** `data-component`
+   identifies _what_ an element is. Existing attributes like
    `data-variant`, `data-size`, and `data-loading` describe the _state_ of that
    element. These concerns must not be mixed.
 
    ```html
    <li data-component="ActionList.Item" data-variant="danger" data-active="true">
-     <span data-part="Label">Delete file</span>
+     <span data-component="ActionList.Item.Label">Delete file</span>
    </li>
    ```
 
-#### `data-component` and `data-part` on all components and appropriate internals
+#### `data-component` on all components and appropriate internals
 
-All Primer components will receive a `data-component` attribute on their root
-element, and all meaningful internal components will receive a `data-part` attribute. These
-attributes serve as the stable identifiers that consumers can use for targeting,
+All Primer components, slots and meaningful parts will receive a `data-component` attribute on their root. This
+attribute serves as the stable identifier that consumers can use for targeting,
 ensuring full coverage across the library.
 
-Every component must provide:
-
-- **`data-component`** on the root element of every component and public
-  sub-component
-- **`data-part`** on every internal structural element that a consumer might
-  reasonably need to target (labels, content wrappers, visual slots, action
-  slots)
-
 Elements that are purely for layout and have no semantic meaning (spacers,
-wrappers that exist only for CSS grid/flex layout) do not require either
+wrappers that exist only for CSS grid/flex layout) do not require this
 attribute.
 
 #### ESLint rule in `eslint-plugin-primer-react`
 
 A new ESLint rule will be added to `eslint-plugin-primer-react` that **warns**
-whenever a `data-component` or `data-part` selector is used in consumer code. The warning will
+whenever a `data-component` selector is used in consumer code. The warning will
 inform consumers of the implications of relying on these selectors — namely that
 the surrounding DOM structure, attributes, parents, and children of the targeted
 element are not guaranteed to be stable.
@@ -168,20 +152,20 @@ component's DOM.
 
 #### Internal CSS usage
 
-Components may use `data-part` selectors in their own CSS Modules for targeting
+Components may use `data-component` selectors in their own CSS Modules for targeting
 child parts. This replaces ad-hoc patterns like bare `[data-component='text']`
 with the standardized naming:
 
 ```css
 /* ButtonBase.module.css */
-& :where([data-part='LeadingVisual']) {
+& :where([data-component='Button.LeadingVisual']) {
   color: var(--button-leadingVisual-fgColor);
 }
 ```
 
 #### Testing requirements
 
-The presence and values of `data-component` and `data-part` attributes must be
+The presence and values of `data-component` attributes must be
 covered by tests. This can be achieved through:
 
 - Unit tests that assert the attributes are present on rendered elements
@@ -221,17 +205,17 @@ accept a contribution!
 
 ### Relationship to CSS Modules and CSS Layers
 
-While the primary purpose of `data-component` and `data-part` is identification
-(testing, tracking, querying), they also serve as stable selectors for CSS
+While the primary purpose of `data-component` is identification
+(testing, tracking, querying), it also serves as a stable selector for CSS
 overrides when consumers need to customize appearance.
 
-In that context, they complement the existing styling architecture:
+In that context, it complements the existing styling architecture:
 
 - **CSS Modules** provide scoped class names for internal styling. Components
   continue to use CSS Module classes for their own styles.
 - **CSS Layers** ([ADR-021](./adr-021-css-layers.md)) ensure that consumer
   overrides take precedence over component styles regardless of specificity.
-- **`data-component` and `data-part`** provide the stable selectors that
+- **`data-component`** provides the stable selectors that
   consumers can use to target components and their parts within those overrides.
 
 Example of a simple, supported override:
@@ -249,14 +233,14 @@ Example of a simple, supported override:
 
 ### Versioning and breaking changes
 
-The **only** guarantee provided for `data-component` and `data-part` attributes is
+The **only** guarantee provided for `data-component` attributes is
 that they **exist** and are **tied to the most relevant DOM element** for each
 component and its internals.
 
 The following aspects are **not** part of the public API and may change at any
 time without notice or a major semver bump:
 
-- The specific DOM element a `data-component` or `data-part` attribute is
+- The specific DOM element a `data-component` attribute is
   applied to
 - The attributes, parent, or children of that element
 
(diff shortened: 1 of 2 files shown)


contributor-docs/adrs/adr-023-stable-selectors-api.md
```diff
@@ -0,0 +1,242 @@
...
+The `data-component` attribute is already used internally across multiple
+components (`Button`, `ActionList`, `PageHeader`, `UnderlineNav`, `Avatar`, and
+others) for internal CSS targeting and DOM queries. However, the naming
+convention is inconsistent:
+
+| Pattern              | Examples                                         |
+| -------------------- | ------------------------------------------------ |
+| `ComponentName.Part` | `ActionList.Description`, `ActionList.Selection` |
+| `PREFIX_Part`        | `PH_LeadingAction`, `PH_Title`, `PH_Navigation`  |
+| `camelCase`          | `buttonContent`, `leadingVisual`, `text`         |
```
siddharthkp, line 43: ouch, i think i know why this is so. 

`ActionList.Description` is a component that is part of the public API, so we expose that exactly.

But Button's `leadingVisual` is not part of the API so it ended up naming it after the className or internal variable name which is camelCase. I don't know if we need to differentiate them based on the nuance of public vs private, i don't think anyone else would care about it.

`PH_LeadingAction`, similar reason, but underscore is random, call it `PageHeader.LeadingAction`

lukasoppermann: No, I think this is not needed.

lukasoppermann:
> I personally really like the idea (i think I introduced it to the codebase), so I approve!
> 
> > target internal parts of a component for style overrides using class selectors.
> 
> I know you did not say anything against this but I think this part is worth highlighting more boldly in the ADR because

Agree. I think this is a quick solution, that may still be helpful if you want to do small styling changes for all actionList.items or something.

But creating more composable subcomponents is definitely an additional need. 

@siddharthkp feel free to adjust with a better example if you want.

contributor-docs/adrs/adr-023-data-component-api.md
```diff
@@ -0,0 +1,242 @@
...
+
+2. **Public sub-components** use dot notation matching the React API. If
+   consumers write `<ActionList.Item>`, the DOM element gets
+   `data-component="ActionList.Item"`.
+
+   ```html
+   <li data-component="ActionList.Item"></li>
+   ```
+
+3. **Internal structural parts** (DOM elements that are not exposed as a
```
TylerJDev: How do we determine which internal parts are meaningful enough to have a `data-component` attribute?

contributor-docs/adrs/adr-023-data-component-api.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Context
+
+Primer React uses [CSS Modules](https://github.com/css-modules/css-modules) for
+styling. CSS Modules generate hashed class names (e.g.,
+`prc-ActionList-Item-cBBI`) that are not stable across versions. This makes it
+impossible for consumers to reliably target internal parts of a component for
+style overrides using class selectors.
+
+Consumers need the ability to target component parts for legitimate use cases:
```
TylerJDev: I think this is a good change! Is this something consumers are currently asking for, or are we anticipating future needs and implementing it in advance?

contributor-docs/adrs/adr-023-stable-selectors-api.md
```diff
@@ -0,0 +1,242 @@
...
+  what component and part it belongs to — the values map directly to the React
+  API.
+- **Enables JavaScript queries.** Consumers and tests can use
+  `querySelectorAll('[data-component="ActionList.Item"]')` reliably.
+- **Complements CSS Layers.** Together with ADR-021, this gives consumers a
+  complete, specificity-safe override mechanism.
+
+### Negative
+
+- **Breaking change for existing consumers.** Anyone currently relying on the
```
TylerJDev, line 320: I wonder if we have data on how many `data-component` attributes are used in CSS/DOM queries today. Either way, I don't anticipate a difficult migration.

contributor-docs/adrs/adr-023-data-component-api.md
```diff
@@ -0,0 +1,242 @@
...
+
+- Unit tests that assert `data-component` is present on rendered elements
+- Snapshot tests that capture the attribute values
+
+Changing a `data-component` value is a **breaking change** and must follow the
+standard breaking change process.
+
+### Migration
+
+Existing `data-component` values must be migrated to the new convention. This
```
joshfarrant: If we're formalising a data attribute for component identification, do you think there's any benefit in namespacing the data attribute? E.g.,

```html
<div data-primer-component="Button.Content">
```

It seems unlikely that there would ever be a collision with other libraries/projects as I can't name any which use `data-component`, but I wonder whether you'd considered namespacing it as part of the formalisation of the attribute?

It could also help ease the transition too, as introducing a new attribute (`data-primer-component`) rather than renaming values on the existing `data-component` would allow both to coexist during a deprecation period. Consumers could migrate their selectors incrementally rather than having all existing `data-component` references break at once.

contributor-docs/adrs/adr-023-data-component-api.md
hectahertz: +1000% to @TylerJDev comment, I'd please like to see real problems and use cases for this that justify this.

See this https://github.com/github/primer/discussions/6393 for why I'm suggesting changes. I think this is unnecessary and could lead to a very fragile system.

contributor-docs/adrs/adr-023-stable-selectors-api.md
```diff
@@ -0,0 +1,242 @@
...
+
+### 2. CSS `::part()` pseudo-element
+
+The `::part()` CSS pseudo-element allows styling of elements inside a shadow
+DOM.
+
+**Why not chosen:** Only works with Shadow DOM, which React does not use. Not
+applicable to our architecture.
+
+### 3. Do nothing — keep `data-component` as an internal implementation detail
```
hectahertz, line 346: Can we look into why they are doing that? I think we should categorize and solve those use cases properly instead.

contributor-docs/adrs/adr-023-stable-selectors-api.md
siddharthkp: [72 files in github/github-ui](https://github.com/search?q=repo:github/github-ui+%22%5Bdata-component%3D%22&type=code), not all of them are primer components, but I'm assuming most of them are

contributor-docs/adrs/adr-023-data-component-api.md
siddharthkp: > From slack, @hectahertz
> Why is this a problem? do we want people changing internals outside of the api? Free reign?

I think this is slightly tangential to your post. The root of the problem is some components not being composable at all. For example, SelectPanel and Autocomplete are black boxes that take title, items, etc. but gives you little control over customising the styles of internal components

> @hectahertz: What's the plan for us restyling those classes? are we going to have to make major releases anytime we change the styling of one of them or the dom structure as they will be breaking changes?

This is a tricky spot we always find ourselves in. If we don't give folks the ability to customise, the only option is to fork/eject. If we do give some customisations, we are now on the hook to support backward compatibility with those customisations  as well. 😢 

For me, a balance does not exist in the long term, because it's not uncommon for us to need to change the internal implementation details for accessibility or other fixes. I'd love to hear ideas on how we can solve for this

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: Good question, any ideas?

I think one aspect could be that they are functional, so not simple structure or a11y elements. 

Do you think we will run into issues here or should we maybe just try it and see if it just works out? We could try to compile a list of what might make a meaningful element if you think this is helpful.

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: Yes, I am currently doing interviews with folks and have hear this as an issue a few times.

I don't have a list of the specific things, but I think especially copilot like copilot chat, ran into a lot of issues where they wanted to change small things but couldn't easily and thus rebuilt the component.

@hectahertz are you suggesting that adding data-attr will have a huge performance impact or are you saying teams can kill performance by messing with the components? Can you elaborate on why this would make the system fragile?

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: This is an interesting idea. I'd be open to it. Wonder what other folks think @siddharthkp @TylerJDev ?

contributor-docs/adrs/adr-023-stable-selectors-api.md
lukasoppermann: I think this is a two step process:
1. Teams adjust / experiment using data attribute selectors
2. Primer team upstreams proven customizations to the system

The problem is that we can't deliver at feature team speed and that the teams need ways to customize components to explore ideas e.g. in flagged ships.

If they always have to rebuild everything, it is unlikely they adjust it to use primer and create a PR once they have proven their idea is needed, because for them it brings little benefit compared to focusing on a new feature.

lukasoppermann:
Some systems, like shd/cdn use `data-slot` e.g. [here](https://ui.shadcn.com/docs/components/base/input-group#custom-input). This allows for a combination of `data-component` on the actual component e.g. `ActionMenu.Button` and `data-slot` on the sub-elements like like TrailingAction (this actually uses a `data-component` at the moment). 

I think this is nice, as you can easily nest it like `[data-component="Button"] [data-slot="icon"]`, but I am not sure if it is a great idea.

What are your opinions? @hectahertz @siddharthkp @joshfarrant @TylerJDev @francinelucca?

contributor-docs/adrs/adr-023-data-component-api.md
siddharthkp: Is there any harm in over-doing it?

francinelucca: > Is there any harm in over-doing it?

My take as well! let's do all the things

contributor-docs/adrs/adr-023-stable-selectors-api.md
francinelucca: I like the idea of this , cuts down on teams having to wait for Primer to deliver on small customization/changes. We can then upstream and cleanup without teams having to be blocked for a week waiting on a Primer release.

contributor-docs/adrs/adr-023-data-component-api.md
siddharthkp: eh, i don't feel strongly about it. I like the simplicity of `data-component`, I'd rather keep it.

> Consumers could migrate their selectors incrementally rather than having all existing data-component references break at once.

Depends on how many we have. It's easy to target both old and new values in a css selector, so we can migrate without a breaking change.

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: So enterprise has some cases where they are trying to style components different. Currently they are wrapping the react component into another react component to add styles. However, it is hard to target specific parts. The span is actually the trailing action part they want to style.

```css
.chipInput {
  border-radius: var(--borderRadius-full) !important;

  & input:placeholder-shown + span {
    display: none;
  }
}

.chipInput span:last-child {
  color: var(--control-fgColor-disabled) !important;

  &:hover {
    color: var(--fgColor-muted) !important;
    cursor: pointer;
  }
}
```

siddharthkp: @lukasoppermann sorry, what component is that? Trying to see if there's a better way that already exists for this specific case

For example, with TextInput, the ideal solution that @hectahertz suggested already works:

```tsx
<TextInput trailingVisual={() => <CalendarIcon className="custom-selector" />} />
```

and then you can target it 

```css
. & input:placeholder-shown + .custom-selector {}
```

I feel this is better than 

```css
. & input:placeholder-shown + [data-component=TextInput.TrailingVisual] {}
```

because it also let's you do other things like

```tsx
<TextInput trailingVisual={() => <CalendarIcon data-testid="id" />} />
```

lukasoppermann:
I now updated the ADR to include `data-slot`.

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: @siddharthkp 
They are trying to replicate the new filter UI style.

For your example, yes, this of course is nice.
I have two issues with that:
1. We don't have this for all our components, and it seems like this will not be a quick fix.
2. This means that if I want to just change the text color for something, or icon color, I always need to add nested components for everything, this may be a bit of on overkill for small changes.

However, I think 2 is a matter of opinion and I am happy to shift there, but I see 1 as a problem.

Do you see this differently?

Another example I remember is a case where folks needed to change the color of the icon in the [banner](https://primer.style/product/components/banner/), which I think is not easily possible.

siddharthkp: > Do you see this differently?

Nope, same page as you

> 1. We don't have this for all our components, and it seems like this will not be a quick fix.

Yes, absolutely. I was trying to point the ideal scenario of how it should be.

2. This means that if I want to just change the text color for something, or icon color, I always need to add nested components for everything, this may be a bit of on overkill for small changes.

Maybe but still very little additional work, so for customisation that goes off the system, I don't feel too bad about it.

siddharthkp:
> Some systems, like shd/cdn use data-slot e.g. [here](https://ui.shadcn.com/docs/components/base/input-group#custom-input). This allows for a combination of data-component on the actual component e.g. ActionMenu.Button and data-slot on the sub-elements like like TrailingAction (this actually uses a data-component at the moment).

Honestly, not a fan of having both. Whether something is a data-component because it composes another primer component or is a data-slot because its local to the component (for now) is an implementation detail that should not matter to a consumer. 

Pick one or the other for simplicity. `data-slot` makes sense when you don't have `data-component` at all because you can target it directly.

lukasoppermann:
> > Some systems, like shd/cdn use data-slot e.g. [here](https://ui.shadcn.com/docs/components/base/input-group#custom-input). This allows for a combination of data-component on the actual component e.g. ActionMenu.Button and data-slot on the sub-elements like like TrailingAction (this actually uses a data-component at the moment).
> 
> Honestly, not a fan of having both. Whether something is a data-component because it composes another primer component or is a data-slot because its local to the component (for now) is an implementation detail that should not matter to a consumer.
> 
> Pick one or the other for simplicity. `data-slot` makes sense when you don't have `data-component` at all because you can target it directly.

@siddharthkp if we just add `data-slot` for inner parts, users could add their own class, e.g. `.button-blue` and use the slot: `.button-blue [data-slot="icon"]`. So I think this would be enough and we could ignore `data-component` on the components for now.

When we have subcomponents, like `ActionList.Item` are they always placed there by the user so that they could add a class?

siddharthkp:
> When we have subcomponents, like ActionList.Item are they always placed there by the user so that they could add a class?

Correct.

Unless, the ActionList is rendered by SelectPanel, then you don't have access to it, because the SelectPanel does not let you compose :(

> @siddharthkp if we just add data-slot for inner parts, users could add their own class, e.g. .button-blue and use the slot: .button-blue [data-slot="icon"]. 

Correct.

> So I think this would be enough and we could ignore data-component on the components for now.

If it's all the same, I'd like to keep the name `data-component` because we already have it and slot can be an implementation detail, not all sub components are "slots"

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: Another example use cases I learned yesterday. Dustin needed to adjust the Dialog position as he was using it to build the command palette. He only made it work with a css hack, as the class is not forwarded to the needed element.

He said that having a way to at least temporary move on (and possibly report a bug / feature request to primer) would have helped him a lot.

hectahertz:
Approving not to block this if you want to merge it, even though I still have the same concerns. Please consider my previous comments and how you'll maintain semver if this gets implemented.

This will be hard to undo if consumers start using it.

contributor-docs/adrs/adr-023-data-component-api.md
```diff
@@ -0,0 +1,285 @@
...
+| Status         | Proposed ❓ |
+| Implementation |             |
+
+## Context
+
+Primer React uses [CSS Modules](https://github.com/css-modules/css-modules) for
+styling. CSS Modules generate hashed class names (e.g.,
+`prc-ActionList-Item-cBBI`) that are not stable across versions. This makes it
+impossible for consumers to reliably target internal parts of a component for
+style overrides using class selectors.
```
mattcosta7: Consumers could be exposed a non-hashed classname, which would be a slightly tighter customization point than a data-attribute.

classes are better optimized for this than data attributes are.   (edit - looks like that was discussed below.   I don't have a strong feeling - but it always feels less ergonomic to use data selectors for child elements)

contributor-docs/adrs/adr-023-data-component-api.md
mattcosta7: I think a lot of folks are (non-optimally) matching on partial css selectors

```
[class~=]
```

and things like that, so it's definitely already needed

contributor-docs/adrs/adr-023-data-component-api.md
```diff
@@ -0,0 +1,308 @@
...
+The presence and values of `data-component` and `data-part` attributes must be
+covered by tests. This can be achieved through:
+
+- Unit tests that assert the attributes are present on rendered elements
+- Snapshot tests that capture the attribute values
+
+Changing a `data-component` or `data-part` value is a **breaking change** and
+must follow the standard breaking change process.
+
+### Versioning and breaking changes
```
joshblack: It'd be nice if this section covered changes on the CSS side, as well, with respect to semver. I think this would be the core concern (personally) when it comes to overrides is that we introduce fragility into the system that can either be difficult to test for or lock us in. 

Some examples that come to mind:

1. Customization that relies on properties being present

```css
/* Our component styles */
.Foo {
  display: flex;
}

/* Downstream override */
.override[data-component="Foo"] {
  align-items: center;
}
```

This creates a contract where if `display` changes to something that does not support the downstream property (in this case `align-items`) it would be a breaking change for overrides.

2. Changing the placement for where properties are applied

One example of this would be for spacing. Say we initially apply padding on a selector:

```css
.Example {
  padding: var(--spacing-1);
}
```

And downstream it is overridden:

```css
.override[data-component="Example"] {
  padding: var(--spacing-2);
}
```

If spacing is applied to a different node (say to a container or a child) then this will cause double spacing instead of continuing to be an override on the padding of the component.

3. Competing overrides

To me, this situation comes up when someone is trying to override a core way the component may work (like they want it to be grid vs flex) and it's a race against overriding all the things that implement this.

This ends up being a race because as you add or implement this logic in items, consumers need to create the rule to override it. It also can lead to unexpected breakage since the override competes with how the component works.

---

I think the first scenario is almost always a challenge for overrides (or at least I haven't found a strategy to mitigate this). I assume the second bucket would be where component tokens come into play?

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: I see the ergonomic argument, but than again, we don't expect folks to change everything about components a lot on a daily basis.

On the upside, it is less likely that we get conflicts with other classes, if we use a data prop. vs a non-hashed class. (Yes, we could prefix everything with `Primer-` to avoid this as well).

I personally am open do either option, I feel having a bit of friction in here may not be bad, as this is supposed to be a stable escape hatch, but not the happy case.

contributor-docs/adrs/adr-023-data-component-api.md
lukasoppermann: I feel like this falls under the customize at your own risk. This is something we need to mention on the docs about those overrides.

But I think our guarantee says that we have a stable API for selecting parts. This is where it ends.

We could add some examples like this and recommend to not rely on css values being there. So that folks for example could do:

```css
/* Downstream override */
.override[data-component="Foo"] {
  display: flex;
  align-items: center;
}
```

However, we should of course mention any css change we make to a part so that folks can see this in the change logs.

@joshblack the way I understand it, this is an issue we have to deal with right? This is the same today, when folks wrap components and overwrite them, or whatever other way they would use.

joshblack: > the way I understand it, this is an issue we have to deal with right? This is the same today, when folks wrap components and overwrite them, or whatever other way they would use.

@lukasoppermann you're definitely right, if passing along `className` today to the outermost component we do run into the same issue. The downside of opening up more places where someone could do this is that it means there are more places where things can break.

My understanding is that this is partially a challenge we have today where folks use a component in an unexpected or unsupported way. When an issue appears, it can be difficult to diagnose why it occurs by the consuming team since they don't know what about the component has changed (and that it would break their usage) and for us since we don't know the ways in which teams use the component

francinelucca:
👋🏽 made changes regarding stable expectations, implementation details and versioning, so re-requesting reviews
CC @mattcosta7 @lukasoppermann @TylerJDev @siddharthkp @hectahertz @joshfarrant @jonrohan @joshblack

contributor-docs/adrs/adr-023-stable-selectors-api.md
```diff
@@ -0,0 +1,374 @@
...
+
+Establish two **public, stable data attributes** that act as reliable
+identifiers for Primer components and their parts in the DOM. Think of these as
+built-in test IDs — stable, predictable anchors that consumers can use for
+testing, tracking, monitoring, and querying without coupling to internal DOM
+structure.
+
+- **`data-component`** — identifies the root element of a component or
+  sub-component.
+- **`data-part`** — identifies an inner structural part within a component.
```
siddharthkp: I strongly feel we should just keep data-component and use it for anything that needs to be customisable.

Mostly because the differentiation between a component and a part is very arbitrary. Sometimes atomic `component` like IconButton can be a `part` for something else, like the Dialog close button. 

And I think it would be very difficult for someone to guess which one to use. So you would always have to render it, inspect element, and see what you want.

I think a component is just a boring implementation detail and we can use data-component if its a react component but just some JSX block within a component. 

Also thinking about refactors, you might have some JSX that you pull into a component, does that mean you need to now change it from a data-part to a data-component? 

I don't think the details matter matters at all. Everything is a component.

francinelucca: updated to everything uses data-component!

francinelucca:
follow up issue: https://github.com/github/primer/issues/6488

PR #7595, opened 2026-02-25
Push margin-top of TimelineBody +1px

llastflowers:
Relates to https://github.com/github/primer/issues/6438

#### Changed

Added +1px to the `margin-top` of `TimelineBody` to improve visual alignment of elements

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

Good test stories:
- https://primer.style/react/storybook/?path=/story/components-timeline-features--timeline-break
- https://primer.style/react/storybook/?path=/story/components-timeline-features--with-inline-links

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/modern-buckets-chew.md
+++ b/.changeset/modern-buckets-chew.md
@@ -1,0 +1,6 @@
+---
+"@primer/react": patch
+---
+
+- Fixes a bug where `ActionBar` menu items would be out of order if new items were mounted after the initial render
+- Improves initial render performance for `ActionBar`
--- a/packages/react/src/ActionBar/ActionBar.tsx
+++ b/packages/react/src/ActionBar/ActionBar.tsx
@@ -1,5 +1,5 @@
 import type {RefObject, MouseEventHandler} from 'react'
-import React, {useState, useCallback, useRef, forwardRef, useId} from 'react'
+import React, {useState, useCallback, useRef, forwardRef, useMemo} from 'react'
 import {KebabHorizontalIcon} from '@primer/octicons-react'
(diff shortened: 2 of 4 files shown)


packages/react/src/Timeline/Timeline.module.css
```diff
@@ -88,7 +88,8 @@
 .TimelineBody {
   min-width: 0;
   max-width: 100%;
-  margin-top: var(--base-size-4);
+  /* stylelint-disable-next-line primer/spacing */
+  margin-top: calc(var(--base-size-4) + 1px);
```
HiroAgustin, line 92: Adding the 1px calc feels icky to me, try with. 

```suggestion
  margin-top: 0.25lh;
```

HiroAgustin (approved):

HiroAgustin:
@copilot add patch changeset

packages/react/src/Timeline/Timeline.module.css
llastflowers: We generally avoid using `lh` (and don't currently have any uses in our repo) because of Safari compatibility issues. We'd have to suppress the lint warning and provide a fallback value to cover unsupported browsers, which ends up being less clean overall. This `calc()` pattern is what we usually use when we need an in-between value not directly provided by our existing CSS variables, so I lean toward keeping things consistent.

llastflowers:
[integration tests](https://github.com/github/github-ui/pull/14686) passing but VRT status is not reporting here. Added skip label to unblock this PR

PR #7604, opened 2026-02-26
AnchoredOverlay: Add support for CSS anchor positioning

TylerJDev:
- Closes https://github.com/github/primer/issues/6446
- Companion PR in github/github-ui merged before this PR: https://github.com/github/github-ui/pull/15790

The first batch of work to add CSS anchor positioning to overlay components.

**Example of CSS anchor positioning:** 

https://github.com/user-attachments/assets/dc1f5f90-d6ac-419c-bb8f-6ce4e9c1dd16

<details><summary>Video description</summary>
<p>
In the video, we show the `AnchoredOverlay` opening and being positioned directly under the trigger. We then show off the overlay repositioning when there's lack of space.
</p>
</details> 

#### New

* Added support for CSS Anchor Positioning in `AnchoredOverlay`, gated by the `primer_react_css_anchor_positioning` feature flag.
* Updated `ActionMenu` to ensure that the anchor button receives the correct anchor class for CSS anchor positioning, and improved merging of class names between anchor and button props.
* Added new stories to `AnchoredOverlay.features.stories.tsx` to demonstrate anchor positioning in various layouts, including centered overlays, grid layouts, within dialogs (including nested and overflowing dialogs), scroll containers, and sticky headers.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -82,7 +82,7 @@
         "react-dom": "^18.3.1"
       },
       "devDependencies": {
-        "@primer/react": "38.15.0",
+        "@primer/react": "38.14.0",
         "@primer/styled-react": "1.0.3",
         "@types/react": "^18.3.11",
         "@types/react-dom": "^18.3.0",
@@ -96,7 +96,7 @@
       "name": "example-nextjs",
       "version": "0.0.0",
       "dependencies": {
-        "@primer/react": "38.15.0",
+        "@primer/react": "38.14.0",
         "@primer/styled-react": "1.0.3",
         "next": "^16.1.5",
         "react": "^19.2.0",
@@ -139,7 +139,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@primer/octicons-react": "^19.21.0",
-        "@primer/react": "38.15.0",
+        "@primer/react": "38.14.0",
         "@primer/styled-react": "1.0.3",
         "clsx": "^2.1.1",
         "next": "^16.1.5",
@@ -27716,13 +27716,13 @@
     },
     "packages/mcp": {
       "name": "@primer/mcp",
-      "version": "0.3.1",
+      "version": "0.3.0",
       "dependencies": {
         "@babel/runtime": "^7.28.6",
         "@modelcontextprotocol/sdk": "^1.24.0",
         "@primer/octicons": "^19.15.5",
         "@primer/primitives": "10.x || 11.x",
-        "@primer/react": "^38.15.0",
+        "@primer/react": "^38.12.0",
         "cheerio": "^1.0.0",
         "turndown": "^7.2.0",
         "zod": "^4.3.5"
@@ -27788,7 +27788,7 @@
     },
     "packages/react": {
       "name": "@primer/react",
-      "version": "38.15.0",
+      "version": "38.14.0",
       "license": "MIT",
       "dependencies": {
         "@github/mini-throttle": "^2.1.1",
--- a/package.json
+++ b/package.json
@@ -99,9 +99,6 @@
   "optionalDependencies": {
     "@rollup/rollup-linux-x64-gnu": "^4.59.0"
   },
-  "overrides": {
-    "css-tree": "3.1.0"
-  },
   "prettier": "@github/prettier-config",
   "size-limit": [
     {


siddharthkp:
> 🤖 Lint and formatting issues have been automatically fixed and committed to this PR.

it keeps saying that but where is the commit? 😅

TylerJDev:
> > 🤖 Lint and formatting issues have been automatically fixed and committed to this PR.
> 
> it keeps saying that but where is the commit? 😅

The bot really doesn't like my PR 😆

packages/react/src/ActionMenu/ActionMenu.tsx
```diff
@@ -73,6 +74,10 @@ const mergeAnchorHandlers = (anchorProps: React.HTMLAttributes<HTMLElement>, but
     mergedAnchorProps.onKeyDown = mergedOnAnchorKeyDown
   }
 
+  if (buttonProps.className) {
+    mergedAnchorProps.className = clsx(anchorProps.className, buttonProps.className)
+  }
```
TylerJDev, line 79: We're more explicit about the classname in this file, as we append a [class directly to the `renderAnchor`](https://github.com/primer/react/pull/7604/changes#diff-1ea41130132e0cccef94777ce888a19a3957fe2ed79d78eb7e8070a034fb3ab6R260).

packages/react/src/AnchoredOverlay/AnchoredOverlay.module.css
```diff
@@ -12,3 +12,46 @@
     display: inline-grid;
   }
 }
+
+.Wrapper {
+  anchor-scope: --anchored-overlay-anchor;
+}
```
TylerJDev, line 18: We use this to scope each `AnchoredOverlay` implementation.

packages/react/src/AnchoredOverlay/AnchoredOverlay.module.css
```diff
@@ -12,3 +12,46 @@
...
+    right: anchor(left);
+    top: anchor(top);
+    margin-right: var(--base-size-4);
+  }
+
+  &[data-side='outside-right'] {
+    left: anchor(right);
+    top: anchor(top);
+    margin-left: var(--base-size-4);
+  }
```
TylerJDev, line 58: This gets us to a good state with attaining parity with our current outside positioning. It might be worth cleaning this up in the future.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -123,6 +125,12 @@ export type AnchoredOverlayProps = AnchoredOverlayBaseProps &
   (AnchoredOverlayPropsWithAnchor | AnchoredOverlayPropsWithoutAnchor) &
   Partial<Pick<PositionSettings, 'align' | 'side' | 'anchorOffset' | 'alignmentOffset' | 'displayInViewport'>>
 
+const applyAnchorPositioningPolyfill = () => {
+  if (typeof window !== 'undefined' && !('anchorName' in document.documentElement.style)) {
+    polyfill()
+  }
+}
+
```
TylerJDev: Taking from prior art here with `TooltipV2`, but curious if there's any thoughts on this.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -291,6 +309,12 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
       ) : null}
     </>
   )
+
+  if (cssAnchorPositioning) {
+    return <div className={classes.Wrapper}>{innerContent}</div>
+  }
```
TylerJDev, line 315: We need to wrap the `AnchoredOverlay` (with `renderAnchor`) in a container to scope it. This will probably cause some test failures in `github-ui` once we ship it.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -12,8 +12,10 @@ import type {AnchorPosition, PositionSettings} from '@primer/behaviors'
 import {type ResponsiveValue} from '../hooks/useResponsiveValue'
 import {IconButton, type IconButtonProps} from '../Button'
 import {XIcon} from '@primer/octicons-react'
+import polyfill from '@oddbird/css-anchor-positioning/fn'
```
siddharthkp: Does this mean the polyfill is already part of the bundle?

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -218,7 +227,11 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
     if (!open && overlayRef.current) {
       updateOverlayRef(null)
     }
-  }, [open, overlayRef, updateOverlayRef])
+
+    if (cssAnchorPositioning && open) {
+      applyAnchorPositioningPolyfill()
```
siddharthkp, line 232: Should we wait for open or load it already if we know we will need it?

packages/react/src/Overlay/Overlay.module.css
```diff
@@ -19,14 +19,20 @@
...
-  bottom: var(--bottom, auto);
+
+  &[data-anchor-position='false'] {
+    /* stylelint-disable-next-line primer/spacing */
+    right: var(--right, auto);
+    /* stylelint-disable-next-line primer/spacing */
+    bottom: var(--bottom, auto);
+  }
+
+  &[data-anchor-position='false']:not([data-responsive='fullscreen']) {
```
siddharthkp: bonus bugfix hiding in this PR?

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
TylerJDev: Maybe just load? 🤔 I'll go ahead and remove the `open` from the conditional!

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
TylerJDev: Yeah! We had some of the polyfill's dependencies already (i.e., `floating-ui`), which is good since we're already accounting for those in our current bundle size.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -242,6 +257,7 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
           tabIndex: 0,
           onClick: onAnchorClick,
           onKeyDown: onAnchorKeyDown,
+          className: classes.Anchor,
```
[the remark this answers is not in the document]
TylerJDev: That's good feedback Copilot!

packages/react/src/Overlay/Overlay.tsx
```diff
@@ -229,22 +231,27 @@ const Overlay = React.forwardRef<HTMLDivElement, internalOverlayProps>(
...
+        role={role}
+        width={width}
+        data-reflow-container={!preventOverflow ? true : undefined}
+        ref={overlayRef}
+        left={leftPosition}
+        right={right}
+        height={height}
+        visibility={visibility}
+        data-responsive={responsiveVariant}
+        data-anchor-position={cssAnchorPositioning ? 'true' : 'false'}
```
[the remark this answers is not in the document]
TylerJDev: Added, with a new style condition: 

`&:not([data-anchor-position]):not([data-responsive='fullscreen'])` - for when only the `Overlay` is used.

packages/react/src/Overlay/Overlay.module.css
TylerJDev: Hah, yeah - we explicitly set the `top` and `left` to 0 when fullscreen is used, so without this it would get overridden.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
siddharthkp: Oh that's interesting. I realised we no longer have a comment from the size action. Would be good to have a number before we merge.

It does run but not sure what the results are 😅 

https://github.com/primer/react/actions/runs/22721918648/job/65886127997?pr=7604

TylerJDev: Having Copilot take a stab at this here: https://github.com/primer/react/pull/7628. Looks like we deleted the `size.yml` which added the comment. This adds it back 🤔

TylerJDev:
@siddharthkp - should be ready for another review! I addressed some feedback, and am working on the size report in https://github.com/primer/react/pull/7628 😁

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
TylerJDev: Still working on this within PRC, but for now we can utilize `github-ui`:

* Bundle size with the package: https://github.com/github/github-ui/pull/15348#issuecomment-4012297024
* Bundle size without: https://github.com/github/github-ui/pull/15342#issuecomment-4011862285

While it does add some (non-packed), I think this is around [what we expect from the spike](https://github.com/github/primer/issues/6374#issuecomment-3874614556).

siddharthkp: I also tested this with examples/nextjs in the repo, it was 16K more, so smaller than what expected in the spike.

My assumption however was that we would only download the polyfill for browsers that do not support anchored-position. I think that's not the case here, right? The polyfill is being downloaded (but not applied) for all browsers even, including when the feature flag is off.

In github-ui, the diff does seem big enough to add some caution emojis 🟡

Image

The tooltip polyfill was fairly small so we decided to put it in the primer bundle, this one feels different.

Can we spend some time trying to see if we can load this dependency only for browsers that need it. The solution might be to load it in github-ui instead of primer/react 👀

TylerJDev: > Can we spend some time trying to see if we can load this dependency only for browsers that need it. The solution might be to load it in github-ui instead of primer/react

Definitely! Do you know if we have prior art for this? I'm wondering what would be the best way to go about it 🤔

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -123,6 +124,13 @@ export type AnchoredOverlayProps = AnchoredOverlayBaseProps &
   (AnchoredOverlayPropsWithAnchor | AnchoredOverlayPropsWithoutAnchor) &
   Partial<Pick<PositionSettings, 'align' | 'side' | 'anchorOffset' | 'alignmentOffset' | 'displayInViewport'>>
 
+const applyAnchorPositioningPolyfill = async () => {
+  if (typeof window !== 'undefined' && !('anchorName' in document.documentElement.style)) {
+    const {default: polyfill} = await import('@oddbird/css-anchor-positioning/fn')
+    polyfill()
+  }
+}
+
```
TylerJDev, line 133: We defer importing until we know that the user's browser does not support CSS anchor positioning.

siddharthkp (approved):
I added e2e tests to compare this PR against: [diff](https://github.com/primer/react/pull/7649/changes#diff-d616245ce74f3be80ee374d154c430d29da87722092830b03ce14147fc1f2c1c). The diffs are only present with the feature flag ON.

Even though this feature isn't perfect yet, we can merge and continue to iterate on it.

siddharthkp:
Added `integration-tests: skipped manually` because tests are passing in the integration branch, but it wasn't reported fully: https://github.com/github/github-ui/pull/15793#issuecomment-4042824514

PR #7605, opened 2026-02-27
ADR component tokens for customization

lukasoppermann:
This [ADR](https://github.com/primer/react/pull/7605/changes#diff-6110584e065d7bba3eed6844839d4c91877ad2ce0d9b4372f53cdc85e15d2aba) discusses an alternative to the [proposed ADR](https://github.com/primer/react/pull/7576) to implement `data-slot` attr for customization.

**This** ADR focuses on using component tokens, which would allow for more control on our side, as we can clearly define what can and can not be overwritten.

The segmentedControl component implements this ADR (will be removed before merge, just as an example).

👀 [View ADR and example code](https://github.com/primer/react/pull/7605/changes#diff-6110584e065d7bba3eed6844839d4c91877ad2ce0d9b4372f53cdc85e15d2aba)


joshfarrant:
Really like the `var()` with fallback pattern — it's a clean design and the SegmentedControl example reads well. 💪

I'm curious about the "escape hatch" framing. If we're establishing naming conventions and a breaking change policy, what else would be required to take this to a first-class public API? The main things I can see are documentation and fixing the stylelint config. What's the barrier to going all the way and dropping the escape hatch framing?

As an aside, I think this ADR could also fit into the broader conversations I've seen around component composability; github/primer#6381, github/primer#6393.

lukasoppermann:
@joshfarrant 
> I'm curious about the "escape hatch" framing. If we're establishing naming conventions and a breaking change policy, what else would be required to take this to a first-class public API? The main things I can see are documentation and fixing the stylelint config. What's the barrier to going all the way and dropping the escape hatch framing?

This framing is not meant because of technical quality issues, but rather to make folks aware that overwriting comes at costs, like inconsistency, it may not visually work with other components, etc.
 
> As an aside, I think this ADR could also fit into the broader conversations I've seen around component composability; [github/primer#6381](https://github.com/github/primer/discussions/6381), [github/primer#6393](https://github.com/github/primer/discussions/6393).

Could you talk about more about it? Fit in there in a good or on a bad way?

joshfarrant:
@lukasoppermann Cool that makes sense, thanks for clarifying. 👍

On the composability discussions — fit in in a good way I think.

Those discussions are exploring whether components should expose their internal parts more directly (compound components, hooks, etc.), which would reduce the need for tokens over time because consumers could access and style the parts themselves. But that's a longer-term direction and doesn't change the value of tokens as a solution today. They're complementary, not in conflict.

Just wanted to link to them here as, in my head at least, these are all working towards the same broader theme of making it easier for consumers to customise components.

contributor-docs/adrs/adr-023-component-tokens.md
```diff
@@ -0,0 +1,198 @@
...
+Consumers should prefer using Primer primitive values (e.g., `var(--bgColor-accent-emphasis)`) over hardcoded colors to maintain theme compatibility.
+
+## Consequences
+
+### Positive
+
+- **Stable public API** — consumers can customize components without depending on internal class names.
+- **Inheritance-friendly** — tokens set on a parent cascade down, enabling contextual theming.
+- **Backward compatible** — existing components continue to work unchanged; tokens are purely additive.
+- **Self-documenting** — the comment block in each component's CSS file serves as the token reference.
```
adierkens, line 161: To be bit pedantic, requiring a comment block is the opposite of self-documenting; we're explicitly writing documentation. 

We should figure out a way to extract & lint against our rules for these if we decide to go down that path: (e.g. stylelint to only allow undefined values if they follow a certain naming convention, postcss plugin to extract these into the `docs.json` metadata)

contributor-docs/adrs/adr-023-component-tokens.md
```diff
@@ -0,0 +1,198 @@
...
+
+1. **Override internal CSS module class names or elements** — brittle, breaks on refactors, and not part of the public API.
+2. **Use inline styles on the component** — limited to properties exposed on the root element and cannot target internal elements like the selected state or hover states.
+3. **Wrap in a themed provider** — too broad, affects all components in the subtree.
+
+Component tokens solve this by exposing a stable set of CSS custom properties that consumers can set on a parent element or the component itself to customize its appearance.
+
+## Decision
+
+Every component should expose a set of **component tokens** as CSS custom properties. These tokens are the public styling API of the component and follow a consistent pattern across all components.
```
adierkens, line 26: How do we differentiate this with the component-tokens already defined within the [primitives package](https://github.com/primer/primitives/tree/main/src/tokens/component)? Should that not already serve as our public API surface for customizing a subset of the behavior of a component?

Could we take the same `segmentedControl` example you have below and author that within the primitives package, leveraging those tokens from the component implementation to provide the same level of customization?

contributor-docs/adrs/adr-023-component-tokens.md
lukasoppermann: Fine with me. We could also put them into the component's docs. Will update.

contributor-docs/adrs/adr-023-component-tokens.md
lukasoppermann: We could use them, yes. But we only use component tokens in primitives as escapes when semantic tokens don't work.

I am not a big fan of this, as it creates a tight coupling between primitives and each component. So I would not want to define this "API" in primitives. Does this make sense? Do you think about this diffently?

PR #7612, opened 2026-03-02
repro of memex test failure when adding tooltip to actionlist item

liuliu-dev:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/ActionList/Item.tsx
+++ b/packages/react/src/ActionList/Item.tsx
@@ -28,24 +28,23 @@ type ActionListSubItemProps = {
  * to avoid remount cycles when truncation state changes.
  * For non-button-semantic items, renders children directly.
  */
-function ConditionalTooltip({
-  text,
-  enabled,
-  children,
-}: {
-  text: string | undefined
-  enabled: boolean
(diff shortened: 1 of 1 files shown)


packages/react/src/ActionList/Item.tsx
```diff
@@ -37,11 +37,11 @@ function ConditionalTooltip({
   enabled: boolean
   children: React.ReactElement
 }) {
-  if (!enabled) {
+  if (!enabled || !text) {
     return children
   }
   return (
-    <Tooltip text={text || ''} direction="e" delay="medium" _privateDisableTooltip={!text}>
```
francinelucca, line 44: @liuliu-dev is there a reason we need to actually call the tooltip component and then disable it internally if the text is empty? I'm thinking we can just return the children in the previous conditional and save us the trouble (+ this way we don't break the tests)

Issue #7634, opened 2026-03-06
`useRefObjectAsForwardedRef` returns stale reference when ref target changes


iansan5653:
`useRefObjectAsForwardedRef` calculates the ref target only one time, on initial calculation (since https://github.com/primer/react/pull/7553). This can cause the target to be stale if it changes after the initial calculation (which happens during the effect cycle), subtly returning old values that _look_ the same but aren't.

For example, a `ref` on the following component will continue to refer to an `HTMLInputElement` even if `asInput` becomes `false` later:

```tsx
function StaleForwardedRef({ref: forwardedRef, asInput}) {
  const ref = useRef(null)
  useRefObjectAsForwardedRef(ref, fowardedRef)

  return asInput ? <input ref={ref} /> : <button ref={ref} />
}
```

The solution is to remove the dependency array introduced in #7553. Since ref targets can change at any time, we can't assume they will remain constant for the lifecycle of the component. And ref targets are not reactive values, so we can't just update the dependency array to be `[ref.current]`. 

The dependency array was introduced for performance, but the cost of calculating the ref is essentially 0 since the function is just `() => ref.current` -- in fact, it might actually be faster than checking the dependency array!

PR #7635, opened 2026-03-06
Revert "perf(useRefObjectAsForwardedRef): add dependency array to useImperativeHandle"

iansan5653:
Reverts primer/react#7553 to fix https://github.com/primer/react/issues/7634

### DIFF
--- a/packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
+++ b/packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
@@ -15,7 +15,7 @@ const Component = forwardRef<HTMLInputElement & HTMLButtonElement, {asButton?: b
 
 describe('useRefObjectAsForwardedRef', () => {
   describe('object refs', () => {
-    it('fowards the ref to the underlying element', async () => {
+    it('forwards the ref to the underlying element', async () => {
       const ref: InputOrButtonRef = {current: null}
 
       render(<Component ref={ref} />)


packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+      const ref: InputOrButtonRef = {current: null}
+
+      const {rerender} = render(<Component ref={ref} />)
+
+      expect(ref.current).toBeInstanceOf(HTMLInputElement)
+
+      rerender(<Component ref={ref} asButton />)
+
+      expect(ref.current).toBeInstanceOf(HTMLButtonElement)
+    })
```
iansan5653, line 48: This is the scenario which is currently failing (and confirmed this test fails on `main`)

packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+        React.useEffect(() => {
+          expect(ref.current).toBeInstanceOf(HTMLInputElement)
+        }, [])
+
+        return <Component ref={ref} />
+      }
+
+      render(<ComponentWithEffect />)
+    })
```
iansan5653, line 64: This scenario isn't failing on `main` or on this branch, but I added it because I encountered it while exploring other approaches and so I wanted to ensure we don't regress on this front

packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    it.todo('does not call the ref on re-render if the target does not change', () => {
+      const ref = vi.fn()
+
+      const {rerender} = render(<Component ref={ref} />)
+
+      rerender(<Component ref={ref} />)
+
+      expect(ref).toHaveBeenCalledExactlyOnceWith(expect.any(HTMLInputElement))
+    })
+  })
```
iansan5653, line 102: Unfortunately I wasn't able to resolve this issue. It's not a new bug (it's been like this since the hook was introduced), but it's definitely not ideal. Every render cycle, React will call callback refs once with `null` and once with the same element instead. So if you use a callback ref with `setState` you could get an infinite render loop. 

I did play with a couple of possible solutions but had no luck:

First, I tried storing the ref value in state, using an intermediate `useImperativeHandle` to update it, so that I could then use the state value as a dependency for the main `useImperativeHandle`:

```ts
const [current, setCurrent] = useState<T | null>(refObject.current)

useImperativeHandle<T | null, T | null>(
  value => setCurrent(value),
  () => refObject.current,
)

useImperativeHandle<T | null, T | null>(forwardedRef, () => current, [current])
```

This solved this bug, but introduced a new one: it delayed the ref value by one render cycle which prevented the value from being set during initial effects (see line 50 for that scenario).

An alternative approach I tried was to wrap the callback ref in another function that skips the callback when the value doesn't change:

```ts
const wrappedForwardedRef: ForwardedRef<T> = useMemo(() => {
  if (typeof forwardedRef === 'function') {
    let lastInstance: T | null | typeof unset = unset

    return (instance: T | null) => {
      if (instance !== lastInstance) {
        lastInstance = instance
        forwardedRef(instance)
      }
    }
  }

  return forwardedRef
}, [forwardedRef])

useImperativeHandle<T | null, T | null>(wrappedForwardedRef, () => refObject.current)
```

However, this doesn't have any effect at all because React automatically calls the callback with `null` in between every value change, so it always goes `value > null > value` which means there's no stable `instance` to compare. 

I think this is just a case of `useImperativeHandle` not giving us enough control and there's no way around that; I'm pretty sure the only way we could solve this is to change the signature of this whole hook to return a stable callback ref which:
1. Updates the object ref
2. Updates the forwarded ref if it's an object, or calls the forwarded ref if its a function

But that would be a breaking change because this hook is exported publicly and the usage would need to change:

```diff
  const refObject = useRef(null)

- useRefObjectAsForwardedRef(forwardedRef, refObject)
+ const ref = useRefObjectAsForwardedRef(forwardedRef, refObject)

- return <div ref={refObject} />
+ return <div ref={ref} />
```

This would essentially be your standard `useCombinedRefs` seen in other libraries.

packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+      const {unmount} = render(<Component ref={ref} />)
+
+      expect(ref).toHaveBeenCalledWith(expect.any(HTMLInputElement))
+
+      unmount()
+
+      expect(ref).toHaveBeenCalledWith(null)
+    })
+
+    it('calls the ref if the target changes', () => {
```
iansan5653, line 80: This test also fails on `main`

packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
francinelucca: I don't see any assertion in the test, what is this testing?

packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
francinelucca: Tracking for next major https://github.com/github/primer/issues/5430#issuecomment-4013625713

packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    // Known issue: The ref is called on every render, even if the target does not change
+    it.todo('does not call the ref on re-render if the target does not change', () => {
+      const ref = vi.fn()
+
+      const {rerender} = render(<Component ref={ref} />)
+
+      rerender(<Component ref={ref} />)
+
+      expect(ref).toHaveBeenCalledExactlyOnceWith(expect.any(HTMLInputElement))
```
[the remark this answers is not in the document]
iansan5653: This is incorrect

packages/react/src/hooks/__tests__/useRefObjectAsForwardedRef.test.tsx
iansan5653: It's kind of a weird test but the assertion is inside the `useEffect`. I was concerned that the async nature of that might cause the assertion to not be called, but I verified that it does fail the test if the assertion fails

PR #7644, opened 2026-03-09
Replace forwarded ref instances of `useProvidedRefOrCreate` with `useMergedRefs`

iansan5653:
Closes https://github.com/github/primer/issues/6788

Per-component changes covered by new feature flag: https://devportal.githubapp.com/feature-flags/primer_react_merged_forwarded_refs/overview

---

In https://github.com/primer/react/pull/7638 I created a `useMergedRefs` hook and migrated all instances of `useRefObjectAsForwardedRef` to it. 

A similar pattern is using `useProvidedRefOrCreate` for this. It's typically used with a type cast and a `ts-expect-error` comment (❗):

```ts
const ref = useProvidedRefOrCreate(forwardedRef as RefObject<HTMLElement | null>)

// @ts-expect-error
return <div ref={ref} />
```

This is obviously problematic. The type errors are trying to point out the problem with this pattern: it will break if consumers pass ref callbacks. Ref callbacks are likely to become more and more common as React 19 has introduced ref callback cleanup functions, making ref callbacks a viable alternative to effects. 

A much better pattern is to be consistent and use the same approach as in #7638:

```ts
const ref = useRef<HTMLElement>(null)
const combinedRef = useCombinedRefs(ref, forwardedRef)

return <div ref={combinedRef} />
```

Note, however, that I did not deprecate `useProvidedRefOrCreate` or remove all calls to it. There are still some valid use cases where refs can be passed in to refer to elements rendered outside the component. We most often see this in anchor refs. There are also some utility hooks that can create a ref internally or use a passed ref. These two cases are still valid; the case I'm targeting here is specifically around **forwarded** refs.

- Stacks on https://github.com/primer/react/pull/7672

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Checkbox/Checkbox.tsx
```diff
@@ -45,7 +52,8 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
     ref,
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ): ReactElement<any> => {
-    const checkboxRef = useProvidedRefOrCreate(ref as React.RefObject<HTMLInputElement>)
+    const checkboxRef = useRef<HTMLInputElement>(null)
+    const combinedRef = useCombinedRefs(checkboxRef, ref as React.RefObject<HTMLInputElement>)
```
iansan5653:
```suggestion
    const combinedRef = useCombinedRefs(checkboxRef, ref)
```

iansan5653:
Skipping changeset because the public facing changes are covered by the changeset in https://github.com/primer/react/pull/7638

packages/react/src/hooks/useMergedRefs.ts
```diff
@@ -37,8 +39,10 @@ export function useMergedRefs<T>(refA: Ref<T | null>, refB: Ref<T | null>) {
       const cleanupA = setRef(refA, value)
       const cleanupB = setRef(refB, value)
 
-      // Only works in React 19. In React 18, the cleanup function will be ignored and the ref will get called with
+      // Callback refs only work in React 19+. In React 18, the ref will get called with
       // `null` which will be passed to each ref as expected.
+      if (majorReactVersion <= 18) return
```
iansan5653, line 59: I was getting a test failure because of an extra call to `console.error` in React 18 tests. Things are working as expected but React warns when a cleanup function is returned from a callback ref in versions before 19, because it's trying to future-proof for 19 and not cause unexpected side effects. 

To resolve the warning, I just inspect the version of React explicitly instead of relying on the implicit differences in behavior. When we drop 19 support we can remove this.

hectahertz (approved):
This looks good! Let's get the integration tests passing and merge 🚀

packages/react/src/hooks/useMergedRefs.ts
hectahertz: Shall we add a `TODO:`? We sometimes grep for those

iansan5653:
😞 This is the most frustrating PR ever. I have no idea why the integration CI is failing, and the only way I can try things is by changing files in this PR, then waiting for the canary release, then rerunning the integration CI. So I'm going to have to revert each individual file one by one until I figure out which change is causing the failure. Very painful.

- [x] ActionList
- [ ] AnchoredOverlay
- [ ] ButtonGroup
- [ ] Checkbox
- [x] Dialog
- [ ] Tabs
- [x] FilteredActionList
- [ ] useMergedRefs
- [ ] PageHeader
- [ ] TextInput
- [ ] Tooltip

siddharthkp:
@iansan5653 Have you had any luck with the primerize improvements?

iansan5653:
The integration CI appears to be failing because of a tooltip (which was previously broken by bad refs) getting in the way of an `Escape` keypress event when dragging and dropping. Fixed downstream by explicitly disabling the tooltip in that scenario  in https://github.com/github/github-ui/pull/26124

packages/react/src/ActionList/List.tsx
```diff
@@ -43,7 +43,12 @@ const UnwrappedList = <As extends React.ElementType = 'ul'>(
 
   const ariaLabelledBy = slots.heading ? (slots.heading.props.id ?? headingId) : listLabelledBy
   const listRole = role || listRoleFromContainer
-  const listRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLUListElement>)
+  const mergedRefEnabled = useFeatureFlag('primer_react_merged_forwarded_refs')
+  const internalRef = useRef<HTMLElement>(null)
+  const mergedRef = useMergedRefs(forwardedRef, internalRef)
+  const providedOrCreatedRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLElement>)
+  const listRef = mergedRefEnabled ? internalRef : providedOrCreatedRef
+  const appliedRef = mergedRefEnabled ? mergedRef : providedOrCreatedRef
```
francinelucca: do we need both listRef and appliedRef?

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -436,12 +438,17 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
           {...(shouldRenderAsPopover ? {popover: 'manual'} : {})}
           {...restOverlayProps}
           {...(shouldRenderAsPopover ? {id: popoverId} : {})}
+          {...(cssAnchorPositioning ? {id: popoverId} : {})}
```
francinelucca: why is this needed?

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -126,7 +127,14 @@ export const Tooltip: ForwardRefExoticComponent<
...
     const child = Children.only(children)
-    const triggerRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLElement>)
+    const mergedRefEnabled = useFeatureFlag('primer_react_merged_forwarded_refs')
+    const internalRef = useRef<HTMLElement>(null)
+    const mergedTriggerRef = useMergedRefs(internalRef, forwardedRef)
+    const providedOrCreatedRef = useProvidedRefOrCreate(forwardedRef as React.RefObject<HTMLElement>)
+    // The ref whose `.current` the positioning logic reads. Both point to the trigger DOM node.
+    const triggerRef = mergedRefEnabled ? internalRef : providedOrCreatedRef
+    // The ref applied to the cloned trigger element.
+    const appliedTriggerRef = mergedRefEnabled ? mergedTriggerRef : providedOrCreatedRef
```
francinelucca, line 137: is it right for the file to still reference `triggerRef` so much even though we stopped applying it in line 288? Seems weird/overly complex to me that we still need these two triggerRef, appliedTriggerRef)

francinelucca (changes requested):

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
llastflowers: Looks like this was an accidental artifact from when I was initially resolving all the merge conflicts on this branch; it's not on main so we can just delete it.

packages/react/src/ActionList/List.tsx
llastflowers: Yes, but only as temporary scaffolding for the feature flag. Once we graduate it, we can reduce these to just `listRef` and `mergedRef`. This is just to keep both paths viable for now while we have the feature flag in place

packages/react/src/TooltipV2/Tooltip.tsx
llastflowers: This is the same pattern as the List one, but I can see how the naming could maybe make it look more confusing. Lmk if you have suggestions!

packages/react/src/ActionList/List.tsx
francinelucca: I understand we need one ref for the FF on (`mergedRef`) and one for the FF off (`providedOrCreatedRef`), and once we graduate we'll keep just` mergedRef`, but I don't understand why `listRef` has to stay around if we're replacing it in line 108. wouldn't we replace all other instances of `listRef` as well? CC @iansan5653 would love your insight here as well

llastflowers: We still need both refs after graduation, but I think the way I named these probably made it confusing. I was trying to avoid renaming every `listRef` in the file, but I see that that makes the diff read weirdly and makes it unclear how the FF works. 

`listRef` pre-dates this PR, so the reason it's replaced on line 108 is because it was replaced by `mergedRef` before I added the featureFlag, and then got the `appliedRef` alias when I added the feature flag. 

I'll update to make the FF-only aliases more clear in every file, and add comments about what exactly needs to change after graduation. Hopefully that makes things more clear!

francinelucca: > We still need both refs after graduation

This is what I feel like I'm lacking understanding on. Before this PR there was only `listRef`. Why with the switch to the new hook, does listRef now need to become two separate refs?

francinelucca: I *think* copilot walked me through it. is it because you can't read from the returned `ref` from `useMergedRef` so you need to use the custom internal one? 
image

siddharthkp:
Unassigning myself as reviewer. feel free to add me back when the PR is ready for review again :)

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -437,11 +439,15 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
           {...restOverlayProps}
           {...(shouldRenderAsPopover ? {id: popoverId} : {})}
           ref={node => {
-            if (overlayProps?.ref) {
-              assignRef(overlayProps.ref, node)
-            }
-            updateOverlayRef(node)
             setOverlayElement(node)
+            if (mergedRefEnabled) {
+              assignRef(mergedOverlayRef, node)
```
iansan5653: Do we still need to keep using `assignRef` here? I'm wondering if maybe we can just pass the merged ref in place of the entire callback function when the flag is enabled

packages/react/src/ActionList/List.tsx
llastflowers: Ohh I see, yeah that's correct! If you look at the diff in [this commit](https://github.com/primer/react/commit/63fca2891ed40d60e5d2c114520e6b99cd29002c) before the feature flag stuff was added, the new pattern might be a bit clearer. I renamed some things to match existing repo conventions (e.g. `combined` --> `merged`), but the implementation is the same. 

As I understand it, the way we were doing it before, our one ref could cover both of those cases (but was sometimes silently failing, hence this new way is an improvement). The new way is more reliable but needs a different ref for each case, so now we have two different ones depending on what it's being used for. 

So I introduced two different aliases for the FF (`readRef` and `appliedRef` in this file) which correspond to each of those new refs (`listRef` and `mergedRef` in this file), and both resolve to the one old ref (`providedOrCreatedRef`) when the FF is off.

llastflowers:
> Unassigning myself as reviewer. feel free to add me back when the PR is ready for review again :)

It is ready! Not sure why it got switched back to draft mode 😅

packages/react/src/Autocomplete/AutocompleteInput.tsx
```diff
@@ -56,13 +56,19 @@ const AutocompleteInput = React.forwardRef(
...
+        // autocomplete menu; if not, hide the menu when focus leaves the input.
         safeSetTimeout(() => {
-          if (document.activeElement !== inputRef.current) {
+          const nextFocusedElement = event.relatedTarget as Node | null
+          const menuElement = document.getElementById(`${id}-listbox`)
+
+          if (
+            !nextFocusedElement ||
+            (nextFocusedElement !== menuElement && !menuElement?.contains(nextFocusedElement))
+          ) {
```
francinelucca, line 71: what are these changes? they seem unrelated to the PR

packages/react/src/hooks/useMergedRefs.ts
francinelucca: let's add this as a task under v39 🙏🏽

francinelucca (commented):
mostly looks good, one question

packages/react/src/Autocomplete/AutocompleteInput.tsx
llastflowers: It's a fix for an issue surfaced by an Autocomplete test that started failing once the mergedRefs pattern was introduced to TextInput in this PR. That change broke Autocomplete's blur logic because it used to depend on how TextInput wires up the ref (`inputRef` on line 65 in the diff), so I changed the blur logic to not depend on the ref anymore. Now it uses the blur event's `relatedTarget` to detect when the focus is inside the menu, which works regardless of the refs setup (FF on or off).

francinelucca: why does if (document.activeElement !== inputRef.current) not work anymore? does the inputRef work on line 78?
Trying to understand if forwardedRefs are compromised for TextInput in general or not

packages/react/src/ActionList/List.tsx
iansan5653: Yes, the new approach is taking the outside (forwarded) ref and the inside (created) ref, and merging them into a single callback ref. This is necessary because the outside ref may also be a callback ref, and you can't access `ref.current` of a callback ref (which is the same reason we can't just use the merged ref internally). So we create a new object ref that we know the shape of, then combine the two. Merging the two refs ensures they always remain synchronized.

```tsx
const internalRef = useRef()
const combinedRef = useMergedRefs(forwardedRef, internalRef)

return <element ref={combinedRef} />
```

It's slightly more complicated to use than `useProvidedRefOrCreate`, but is more robust as it works with any input ref shape instead of only object refs.

packages/react/src/Autocomplete/AutocompleteInput.tsx
iansan5653: Definitely agree with @francinelucca here; this seems suspicious. `inputRef` should be working just as before; if it's not getting bound to the target input then that points at a different problem that's worth digging in to. I would try to find a solution that makes the ref work properly; it could be a problem with how we are binding or creating it. I dug through the code but nothing stood out to me.

llastflowers: It does work, and none of the ref stuff is compromised! The sequence of events is kind of confusing, but basically `AutocompleteInput` and `TextInput` need to communicate for the blur handler to work properly in the context of the test, and they were at first when both were using the old refs setup. 

After this PR was originally opened, #7834 landed on main. It changed two things at once: `AutocompleteInput` now passes a callback ref, and the test switched from `userEvent.tab()` to `fireEvent.blur()` (to address some test flakiness issues; see inline comment in test code snippet below). With these changes, the aforementioned communication is broken, but the test still passes (false positive). The callback ref is one the old `TextInput` can't apply to focus the input, so the input never actually gets focused. And because `fireEvent.blur()` fires the handler anyway, the old `document.activeElement !== input` check sees focus 'off' the input and the test reads it as success. Not because the focus actually moved, but because it was never there in the first place. This is what's on main right now.

On this PR branch, `TextInput` now focuses the input via its own internal ref, so the mismatch with the `AutocompleteInput` ref stuff is no longer an issue. But because of the `fireEvent.blur()` change in the test, it started failing now that the refs are working properly again: `userEvent.tab()` moves focus off of the active element, but `fireEvent.blur()` does not. So the input is focused now, but the focus never moves off of the input, so the menu doesn't close, and the test fails.

This exposes imo a legitimate fragility of the original handler: it can't actually tell the difference between the input being focused and then focus moving away vs. the input never being focused at all. The new handler avoids this by checking `relatedTarget` instead, which is read directly from the blur event. It's based on the actual focus transition instead of an inference based on the DOM state after the interaction. The new handler is also more stable in practice because it doesn't rely on timing (when exactly the `document.activeElement` changes or ref updates, etc.).

This is the test:

```
    it('closes the menu when the input is blurred', async () => {
      const {getByLabelText} = render(
        <LabelledAutocomplete menuProps={{items: [], selectedItemIds: [], ['aria-labelledby']: 'autocompleteLabel'}} />,
      )
      const inputNode = getByLabelText(AUTOCOMPLETE_LABEL)

      expect(inputNode.getAttribute('aria-expanded')).not.toBe('true')
      fireEvent.click(inputNode)
      fireEvent.keyDown(inputNode, {key: 'ArrowDown'})

      expect(inputNode.getAttribute('aria-expanded')).toBe('true')

      // `userEvent.tab()` is unreliable in browser-mode Vitest for this case; blur is deterministic.
      // eslint-disable-next-line github/no-blur
      fireEvent.blur(inputNode)

      await waitFor(() => expect(inputNode.getAttribute('aria-expanded')).not.toBe('true'))
    })
```

iansan5653: Makes sense, thank you for the detailed explanation! Would you be open to pulling this change out into a separate PR? Feels like it might be worth splitting out so it can ship faster / independently of this wider refactoring change. 

Otherwise though it sounds good to me :+1:

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
iansan5653: Maybe we wait to think about this until we clean the flag up

llastflowers: Good point; updated!

packages/react/src/Autocomplete/AutocompleteInput.tsx
llastflowers: Unfortunately if we remove this change from this branch, the test will start failing again since it's related to the TextInput refactoring changes!

francinelucca (approved):

PR #5476, opened 2024-12-23
ActionBar: Ensure overflow menu only exists if there is no space

TylerJDev:
Closes https://github.com/github/primer/issues/4584

Ensures that the "overflow" menu in ActionBar only appears when there is no space for all items. Also ensures that the overflow menu is present if there is no space for all items on initial render.

**Fixed behavior:**
 
https://github.com/user-attachments/assets/f82e8b4d-3994-45b4-a70d-30bed12a7543

<details><summary>Video description</summary>
<p>

Video showing ActionBar at different resize points. When the small screen size is shown, the overflow menu in ActionBar appears. When at a medium screen size, the overflow menu is hidden and all menu items are present.

</p>
</details> 

#### Changed

* Overflow menu will only appear if there is no space for all items
* Overflow menu will appear on initial render if there is no space for all items

### Rollout strategy

- [x] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [x] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [x] (GitHub staff only) Integration tests pass at github/github ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #7655, opened 2026-03-11
Replace `ActionBar` overflow calculations with CSS wrapping approach

### LINKED ISSUE #7447: Avoid flickering when calculating `ActionBar` overflow

iansan5653:
Takes the CSS flex wrapping approach to detecting overflowing from https://github.com/primer/react/pull/7506 and applies it to `ActionBar`, resulting in a much simpler and more reliable solution that is also SSR-stable:

https://github.com/user-attachments/assets/6fbd1a20-8246-456b-b19f-3a3280c3c2da

Compare to the current implementation, which flickers on load and doesn't calculate the overflow as accurately, causing items to occasionally get clipped (this is more noticeable on production where it can sometimes calculate the overflow incorrectly; but that's hard to show in Storybook):

https://github.com/user-attachments/assets/a4d124ba-a432-4074-a1b9-30bc9beaf271

- Closes https://github.com/primer/react/issues/7447
- Closes https://github.com/github/primer/issues/6394
- Closes https://github.com/github/primer/issues/6307

#### Changed

Improves `ActionBar` overflow calculations and SSR support

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionBar/ActionBar.module.css
```diff
@@ -8,8 +8,28 @@
   margin-bottom: -1px;
   white-space: nowrap;
   list-style: none;
-  align-items: center;
+  align-items: flex-start;
   gap: var(--actionbar-gap, var(--stack-gap-condensed));
+  overflow: hidden;
+  height: var(--actionbar-height);
+  animation: detect-overflow linear;
+  animation-timeline: scroll(self block);
```
francinelucca: can you eli5 what this does? 😅

francinelucca (dismissed):
something's off here 👀 

https://github.com/user-attachments/assets/2356ae1c-817e-4c4e-b1c4-5dbcbb51bf6a

iansan5653:
> something's off here 👀

Should be resolved now by setting `visibility:hidden` on those items which will remove them from the tab order. Thanks for catching that!

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -512,6 +414,8 @@ export const ActionBarIconButton = forwardRef(
         onClick={clickHandler}
         {...props}
         variant="invisible"
+        data-overflowing={isOverflowing}
+        aria-hidden={isOverflowing || props['aria-hidden']}
       />
```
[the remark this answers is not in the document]
iansan5653: Fixed by applying `visibility: hidden` instead, which removes them from the accessibility tree and tab order

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -546,32 +449,34 @@ export const ActionBarMenu = forwardRef(
   ) => {
     const backupRef = useRef<HTMLButtonElement>(null)
     const ref = (forwardedRef ?? backupRef) as RefObject<HTMLButtonElement>
-    const {isVisibleChild} = React.useContext(ActionBarContext)
```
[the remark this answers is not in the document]
iansan5653: I'm hoping to ship https://github.com/primer/react/pull/7638 first and then use `useCombinedRefs` from there

packages/react/src/ActionBar/ActionBar.module.css
iansan5653: Definitely, I should have added a comment. Added one:

> Scroll-based animations have no effect unless the container is scrollable (has overflow, even with overflow:hidden)  so we can use them to detect overflow. This allows us to use pure CSS to show/hide the menu button on suported  browsers, reducing layout shift and improving performance. On unsupported browsers we still rely on just JS via  `data-has-oveflow`. It would be cleaner to use scroll-state container queries for this, but browser support for  scroll-driven animations is slightly better.

francinelucca (changes requested):
Looks better, now something's off with the keyboard navigation. I can't get from the last item to the overflow menu by arrowing right:

https://github.com/user-attachments/assets/a2761572-8f19-4e6d-adda-a9b0dae9ad73

iansan5653:
Huh, this one is strange. Home/End work as expected, but arrow keys seem to be treating the overflow menu as the first item - you can only get to it by pressing Left from the first action 🤔

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -361,139 +249,139 @@ export const ActionBar: React.FC<React.PropsWithChildren<ActionBarProps>> = prop
...
+    () => false,
+  )
+
+  const isOverflowing = isGroupOverflowing || isItemOverflowing
 
-  useIsomorphicLayoutEffect(() => setWidth(ref.current?.getBoundingClientRect().width ?? -1), [ref])
+  ActionBarItemsRegistry.useRegisterDescendant(isOverflowing ? registryProps : null)
 
-  return width
+  return {isOverflowing, dataOverflowingAttr: isOverflowing ? '' : undefined}
```
francinelucca, line 343: does this actually need to return the `dataOverflowingAttr`? couldn't we just look at the value of isOverflowing to set that on consumer side? wondering why both are needed here 🤔

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -157,30 +141,7 @@ export type ActionBarMenuProps = {
...
-    if (sumsOfChildWidth > widthToFit) {
-      breakpoint = index
-      break
-    } else {
-      continue
-    }
-  }
-  return breakpoint
-}
+const ActionBarItemsRegistry = createDescendantRegistry<ChildProps | null>()
```
francinelucca, line 144: does this need to take in the ChildProps? if I'm reading correctly the registry is only read on line 240 to check if there's a non-null value? wondering if this can be simplified if so 🤔

francinelucca (changes requested):
Getting closer, found a few issues:
 1. it looks like the [KeyBindingHint story](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar--multiple-keybinding-hints) isn't working properly:

https://github.com/user-attachments/assets/6a55aa5f-cfea-4dfd-806f-58101bfc59e3

2. Something weird is going on with resizing the screen in [this story](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--gap-scale). I can't get back to one of the ActionBars via keyboard:

https://github.com/user-attachments/assets/08dd5ebe-b48b-4b03-b8d6-5c42be62eac2

3. Focus state looks clipped in [small ActionBar](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--small-action-bar) story

https://github.com/user-attachments/assets/b0f2c34c-9ce0-49b2-9f4b-e8388a8621b2

4. [These](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--text-labels) are disappearing 

https://github.com/user-attachments/assets/94da86a4-f77d-4554-b236-93667cb472a5

5. When [this story](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--with-groups) is wrapped all the way I can't get to it via keyboard

https://github.com/user-attachments/assets/c9bda0eb-c423-4553-ad98-8888ec59febd

iansan5653:
Thanks for the thorough review! Looks like tooltips in general aren't working on toolbar items 🤔. I'll dig in to that; maybe they are getting clipped or hidden.





> 4\. [These](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--text-labels) are disappearing

For this one, it doesn't seem like there's really a well-defined preferred behavior here? These buttons aren't `ActionBar` buttons, so we can't handle them by putting them into the overflow menu. On `main`, the buttons just overflow off to the right, which feels worse IMO because it pushes the overflow menu out of view.

I think what we should do is just implement an `ActionBar.Button` component and use that instead, but I didn't want to do that in this PR. Can we be OK with them disappearing for now, knowing that we'll address it with a better solution in the future?

packages/react/src/ActionBar/ActionBar.tsx
iansan5653: We do need the props because we also use it in L238-240 where we build the list of items for the overflow menu. The props tells us what label to render, what to do on click, etc

packages/react/src/ActionBar/ActionBar.tsx
iansan5653: Yeah it totally could and that was how I originally had it. But we'd have to do it for every different kind of item, so it seemed to make sense to just pull the common logic out to here and ensure we are consistent. This is an internal (not consumer facing) hook, so I wasn't too worried about the API being a little redundant

francinelucca:
> On main, the buttons just overflow off to the right, which feels worse IMO because it pushes the overflow menu out of view.

You can still scroll to them vs in this implementation they're completely gone so idk about worse. 

> I think what we should do is just implement an ActionBar.Button component and use that instead, but I didn't want to do that in this PR. Can we be OK with them disappearing for now, knowing that we'll address it with a better solution in the future?

I agree this seems like an unsupported case. I almost want to remove that example entirely until this is officially supported by ActionBar. CC @TylerJDev I see you introduced the story in https://github.com/primer/react/pull/5476, do you remember if there was a reason/use-case for this? want to make sure we don't break anything if we move forward

iansan5653:
> I almost want to remove that example entirely until this is officially supported by ActionBar.

That also makes sense to me :+1:

FWIW it wouldn't be difficult at all to add support for text buttons, especially with the new approach. I could plan to do it in a followup PR.

iansan5653:
Okay, turns out the tooltip issue repros in `main` so it's not a problem with this PR. It seems to have something to do with `ref` handling and is resolved in another PR I'm working on, #7644. 

Still working on the other issues:

> 2\. Something weird is going on with resizing the screen in [this story](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--gap-scale). I can't get back to one of the ActionBars via keyboard:

Could you please explain in more detail what the issue is here? I can't figure out what the bug is from your screencap.



> 3\. Focus state looks clipped in [small ActionBar](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--small-action-bar) story

Looks like I need to fix the overflow button size when the `size="small"`. **Update**: resolved ✔️ 

> 5\. When [this story](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--with-groups) is wrapped all the way I can't get to it via keyboard

Can repro this one; looking in to it. **Update**: resolved ✔️

TylerJDev:
> I agree this seems like an unsupported case. I almost want to remove that example entirely until this is officially supported by ActionBar. CC @TylerJDev I see you introduced the story in https://github.com/primer/react/pull/5476, do you remember if there was a reason/use-case for this? want to make sure we don't break anything if we move forward

I can't remember the context, so I think it's okay to remove! I don't think there's any real world cases that use ActionBar like this. 

I did notice one issue in github-ui the staffbar component renders differently in https://github.com/github/github-ui/pull/17750 compared to production storybook when you resize it horizontally (< 500px width). It looks like the items are offscreen 🤔

iansan5653:
> I think it's okay to remove!

Removed!

> I did notice one issue in github-ui the staffbar component renders differently in [github/github-ui#17750](https://github.com/github/github-ui/pull/17750) compared to production storybook when you resize it horizontally (< 500px width). It looks like the items are offscreen 🤔

Yep, that's the same issue. It's (ab)using `ActionBar` by rendering labelled items instead of regular action bar items. I'm not too worried about that though because I also maintain that UI and it's very experimental right now

francinelucca (changes requested):
these don't look correctly centered:

[Large action bar
](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar--playground):
Image

[Small action bar](https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--small-action-bar): 

Image


I'm still having this issue where once I resize one of the ActionBars becomes unreachable

https://github.com/user-attachments/assets/b7a62ed0-858e-4de7-a374-284d460fc994

All other previous issues seem resolved

francinelucca:
> Okay, turns out the tooltip issue repros in main so it's not a problem with this PR. It seems to have something to do with ref handling and is resolved in another PR I'm working on, https://github.com/primer/react/pull/7644.

Created a bug to track -> https://github.com/github/primer/issues/6532

francinelucca:
> I think what we should do is just implement an ActionBar.Button component and use that instead, but I didn't want to do that in this PR

Also https://github.com/github/primer/issues/6533

francinelucca (changes requested):
1 - There's a new tab stop on the last element of the ActionBar now 🤔. Like if I'm not positioned in the last element (more menu needs to be visible), I have to tab twice to leave the ActionBar

https://github.com/user-attachments/assets/a40c73a8-9577-483c-b8cd-87bc38d20b26

2 - When i'm on https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--with-groups with all the items hidden (just the more button showing), I'm having a hard time actually getting there via keyboard, there seems to be no tab stop

iansan5653:
> 1 - There's a new tab stop on the last element of the ActionBar now 🤔. Like if I'm not positioned in the last element (more menu needs to be visible), I have to tab twice to leave the ActionBar

That's a really strange one. I wonder if it's a bug in `useFocusZone` itself. For some reason the more button always has `tabindex="0"` instead of `tabindex="-1"` like the other buttons.

> When i'm on https://primer-c23a44cf2f-13348165.drafts.github.io/storybook/?path=/story/experimental-components-actionbar-examples--with-groups with all the items hidden (just the more button showing), I'm having a hard time actually getting there via keyboard, there seems to be no tab stop

I can't repro this one - it seems to be working right for me?

https://github.com/user-attachments/assets/01ff5907-c472-44ae-b396-40a0fffb7c72

francinelucca:
Confirming, can't repro the second one anymore 🤷🏽

iansan5653:
Okay, I think I finally figured out why `focusZone` was not working before with `focusableElementFilter` or other approaches. I think the hook was only querying for focusable elements once, on mount. But this elements list can change across renders as the screen gets resized and even when the initial layout calculates, so this is not an accurate approach. 

I'm pretty sure I was able to get everything working correctly by passing the overflow items array as a dependency.

francinelucca (approved):
Looks amazing 😻. Very excited to get this in! thanks for persevering through this 🙏🏽 
Also really happy with the code cleanup since last revision :sparkle: :broom:

PR #7723, opened 2026-04-01
Add Card component

liuliu-dev:
Upstream the Card component from `Pacer` into `@primer/react`.

Issue: https://github.com/github/primer/issues/6467

#### New

- `Card` — A styled container component with border, box-shadow, and border-radius; also supports custom content without subcomponents (plain container mode)
- `Card.Icon` — Renders a decorative icon in the card header (defaults to `aria-hidden`, supports `aria-label` for meaningful icons)
- `Card.Image` — Renders an edge-to-edge image in the card header
- `Card.Heading` — Renders a heading (defaults to h3, configurable via `as` prop)
- `Card.Description` — Renders a description below the heading
- `Card.Menu` — Slot for overlay menu (positioned absolute top-right)
- `Card.Metadata` — Renders metadata content below the description

Screenshot 2026-04-10 at 8 41 00 AM

#### Changed

- `Playground` Storybook story updated with boolean toggle controls for `Card.Icon` and `Card.Metadata`
- `CustomContent` Storybook story added to demonstrate using `Card` as a plain styled container with arbitrary children (no subcomponents)

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/index.ts
```diff
@@ -78,6 +78,8 @@ export {default as AvatarStack} from './AvatarStack'
 export type {AvatarStackProps} from './AvatarStack'
 export {Banner} from './Banner'
 export type {BannerProps} from './Banner'
+export {Card} from './Card'
+export type {CardProps} from './Card'
```
hectahertz: Nit (non-blocking): root exports currently include only `CardProps` here. Since the local Card entrypoint exposes subcomponent prop types too, it may be worth re-exporting those from root for consistency ✨.

packages/react/src/Card/Card.tsx
```diff
@@ -0,0 +1,158 @@
...
+    } else if (child.type === CardMetadata) {
+      metadata = child
+    } else if (child.type === CardMenu) {
+      menu = child
+    }
+  }
+
+  return (
+    <div ref={ref} className={clsx(classes.Card, className)} {...rest}>
+      <div className={clsx(classes.CardHeader, image && classes.CardHeaderEdgeToEdge)}>{image || icon}</div>
```
hectahertz: Nit (non-blocking) 🙂: this header wrapper is always rendered, so heading-only usage gets an empty padded top area. Consider rendering it only when `image || icon` is present.

hectahertz (approved):
LGTM overall, nice addition 👏

Approving now ✅

One small non-blocking follow-up idea: if Card is intended to appear in generated component docs/inventory immediately, adding the component metadata in this PR would keep docs parity. Totally fine as a quick follow-up if preferred.

liuliu-dev:
Hey @lukasoppermann @ericwbailey 👋  
Would love a design and a11y review for this `Card` component. Storybook preview is available on the [PR deployment](https://primer-1b76de82d9-13348165.drafts.github.io/storybook/?path=/story/components-card--default). Happy to iterate on feedback!

ericwbailey:
Can't speak to the code or overall plan, but some high-level considerations that we'll have to contend with at some point:

### `Card.Heading` — Renders a heading (h3) with truncation

- We'll want to make `h3` able to accept multiple heading levels, in order to better integrate with the existing view heading structure. We can default to `h3`.
- Truncation is an accessibility antipattern that represents non-trivial MAS risk and we'll want to remove.

### `Card.Icon` — Renders an icon in the card header

- We'll want to make this default to a decorative icon, with the option to add an alternate description if needed.

### `Card.Description` — Renders a description with 2-line clamp

- Like the heading, truncation is an accessibility antipattern that represents non-trivial MAS risk and we'll want to remove.

liuliu-dev:
Thanks for the review @ericwbailey! Addressed:

- **Heading**: supported via `as` prop (defaults to `h3`)
- **Truncation**: removed from both `heading` and `description`, long `heading`/ `description` will wrap
- **Icon**: now defaults to decorative (`aria-hidden`), with optional `aria-label`

lukasoppermann:
@copilot in the Playground storybook story, add toggles for the icon `<Card.Icon>` and the `<Card.Metadata>`.

Also add a story that does not use any of the internal components and just adds some custom content to the card

lukasoppermann (changes requested):
@liuliu-dev design wise:
- heading, please use `--text-title-shorthand-small`
- for body copy: `--text-body-shorthand-medium`
- for the gap please use a variable from the stack: https://primer.style/product/primitives/size/#stack

I think we can remove the fallback values, they are not needed and don't exist anymore.

lukasoppermann:
@liuliu-dev should we tie the padding to the content-top, instead of icon bottom?

CleanShot 2026-04-07 at 09 23 53@2x

lukasoppermann:
This is also a use case: https://primer-1b76de82d9-13348165.drafts.github.io/storybook/?path=/story/components-card--custom-content

CleanShot 2026-04-07 at 09 25 01@2x

I am wondering if we should expect users to "bring their own padding" or if we should have a padding inside the card by default and a prop or a way to easily remove the default padding? CC: @siddharthkp ?

ericwbailey:
>Thanks for the review @ericwbailey! Addressed:

Amazing! Thank you so much!

liuliu-dev:
Thanks @lukasoppermann! Here's what I've updated based on your feedback:

- Updated typography and spacing to use the recommended tokens
- Padding: moved padding to the `Card` root and added a `padding` prop (`normal` | `none`), defaults to `normal`. Header/body no longer have their own padding, spacing between them is controlled by `gap` instead.
- Moved the custom content one into features stories.

lukasoppermann (approved):

lukasoppermann:
@liuliu-dev perfect, thank you.

siddharthkp:
> I am wondering if we should expect users to "bring their own padding" or if we should have a padding inside the card by default and a prop or a way to easily remove the default padding? CC: @siddharthkp ?

I think bring your padding is a pattern we don't follow anywhere right now because it can lead to variation that is visibly inconsistent. Something we can take from PageLayout is the idea of `padding:  'none' | 'condensed' | 'spacious'` and of course, you can always fine-tune it with css.

lukasoppermann:
> I think bring your padding is a pattern we don't follow anywhere right now because it can lead to variation that is visibly inconsistent. Something we can take from PageLayout is the idea of padding:  'none' | 'condensed' | 'spacious' and of course, you can always fine-tune it with css.

The prop is now implemented, this makes sense to me.

@siddharthkp the Pagelayout does not make total sense to me, because there seems to be no `default` or `normal` one. Does the current implementation for card make sense to you?

> you can always fine-tune it with css.

This is important! 👍

siddharthkp:
> @siddharthkp the Pagelayout does not make total sense to me, because there seems to be no default or normal one.

My bad, the values [in the docs](https://arc.net/l/quote/fosripnp) are `none | condensed | normal`

`normal` as a value is a bit weird to me. `default` is totally fine.

> Does the current implementation for card make sense to you?

I've come into this blind, so I don't actually know what's going on or what we are trying to do 😅 

I would say however, we should put this in experimental and not root index right away

packages/react/src/index.ts
```diff
@@ -78,6 +78,16 @@ export {default as AvatarStack} from './AvatarStack'
...
+export {Card} from './Card'
+export type {
+  CardProps,
+  CardIconProps,
+  CardImageProps,
+  CardHeadingProps,
+  CardDescriptionProps,
+  CardMenuProps,
+  CardMetadataProps,
+} from './Card'
```
siddharthkp: Should we export this from src/experimental/index until we build some confidence in the API with some usage?

lukasoppermann: Whatever you folks feel is best. Do we have a clear path for how we move this out of experimental? It feels like stuff stays in experimental for way to long some times.

lukasoppermann:
Figma component: https://github.com/github/primer/issues/6537

packages/react/src/index.ts
liuliu-dev: Makes sense to keep it in experimental for now. It’s mostly upstreamed from Pacer so it’s not totally new, but the padding prop is new, if it looks good in usage we can move it out pretty quickly.

siddharthkp: Yep, but I would count pacer as super experimental too (and maybe deprecated). 

It did not go through primer's patterns or accessibility process, so I'd like for us to be able to change our mind on the details

siddharthkp: ^ which makes me wonder, @lukasoppermann are there any Card components as shared-components in github-ui that we should also consider as inventory? 

Feels a bit strange to promote the experimental pacer component without looking at the other stable components that exist in github-ui

(Doesn't have to be blocking if we are putting this component in experimental, we can iterate over it before pushing for adoption)

liuliu-dev: Moved `Card` to `@primer/react/experimental`, merging now, happy to iterate in the future!

PR #7724, opened 2026-04-02
feat: implement ADR-023 Part 1

francinelucca:
Towards https://github.com/github/primer/issues/6497

Adds data-component attribute (see [ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-023-stable-selectors-api.md#implementation-details)) for:
- ActionBar
- ActionList
- Button
- FilteredActionList
- Link
- LinkButton
- Pagination
- SelectPanel
- Table
- TextInput
- TextInputWithTokens
- TooltipV2

#### New

- stable selectors for listed components
- tests for stable selectors added

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -571,7 +573,14 @@ export const ActionBarMenu = forwardRef(
     return (
       <ActionMenu anchorRef={ref} open={menuOpen} onOpenChange={setMenuOpen}>
         <ActionMenu.Anchor>
-          <IconButton variant="invisible" aria-label={ariaLabel} icon={icon} {...props} />
+          <IconButton
+            variant="invisible"
+            aria-label={ariaLabel}
+            icon={icon}
+            {...props}
+            // TODO: does this make sense? it'll override IconButton's data-component
```
francinelucca: I feel like this indeed makes no sense, but does that mean that subcomponents like ActionBar.Menu don't have a data-component, despite being public subcomponents? how would you target them then?: `[data-component="ActionBar"] [data-component="IconButton"]` isn't specific enough because ActionBar.IconButton would match this as well 🤔 
Should we wrap everything top-level in a span for these cases just so we can attach the data-component attribute? is there any chance for that to mess anything up?

siddharthkp: In this scenario, i think it makes sense to override the menu IconButton because there are other IconButtons in this component as well. (I'd call it `ActionBar.MoreIconButton` or `ActionBar.MenuIconButton`)

It's likely that it will be used with a classname on the ActionBar root `.myActionBar [data-component=ActionBar.MoreIconButton]`

That said, I don't think `<Actionbar.IconButton>` should get a data-component at all because 
1. they already have a data-component
2. you can pass a classname to that

> Should we wrap everything top-level in a span for these cases just so we can attach the data-component attribute?

Not a good idea because it will introduce nesting, which in best of cases is unnecessary and in the worst cases breaks dom validation rules or slots constraints

---

There will be other similar scenarios where we might choose to override the child's data-component when there are two distinct IconButtons that are baked in the component and not accessible in the API, for example the previous and next icon buttons in Pagination.

packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -571,7 +571,14 @@ export const ActionBarMenu = forwardRef(
...
       <ActionMenu anchorRef={ref} open={menuOpen} onOpenChange={setMenuOpen}>
         <ActionMenu.Anchor>
-          <IconButton variant="invisible" aria-label={ariaLabel} icon={icon} {...props} />
+          <IconButton
+            variant="invisible"
+            aria-label={ariaLabel}
+            icon={icon}
+            {...props}
+            // overriding IconButton's data-component so that the ActionBar's "More Menu" Icon can be targetted specifically
+            data-component="ActionBar.Menu.IconButton"
```
francinelucca, line 580: should this be:
- ActionBar.MenuIconButton
- ActionBar.Menu
- ActionBar.Menu.IconButton
?

I feel like this is the IconButton that belongs to the Menu that is a subcomponent of ActionBar, which is why `ActionBar.Menu.IconButton` makes sense to me 🤔. `ActionBar.Menu` seems disingenuous because really this is not the menu itself, just the Icon trigger. But I _could_ see the case for ActionBar.MenuIconButton

llastflowers: I vote for ActionBar.Menu.IconButton too!

packages/react/src/ActionList/Group.tsx
francinelucca, line 192: Here the actual user-supplied `className` is in the `span` within the "HeadingWrap"; 
Should the data-component always be where the className is? or does it make sense for it to go on the outermost element?
if so:
For this case, should we do: `data-component="GroupHeading.Wrapper"` and `data-component="GroupHeading"`?
If so:
does that mean we should always have data-component for wrappers?

packages/react/src/ActionBar/ActionBar.tsx
siddharthkp: tl;dr: both `ActionBar.MenuIconButton` and `ActionBar.Menu.IconButton` are fine, i like one more than the other

- `ActionBar.Menu` 👎 that's for the menu, not the button
- `ActionBar.Menu.IconButton` it's fine, but the double dots feels weird, idk why. Maybe because `ActionBar.Menu` is actually a shorthand for `ActionMenu`... but `ActionBar.ActionMenu.IconButton` _feels_ worse for some reason.
- `ActionBar.MenuIconButton`: 👍 I like this the most. consistent pattern everywhere

francinelucca: what about here 
image

you vote for ActionList.ItemLabel?

siddharthkp: Hmmmmmm, I know i'm being very inconsistent with my feeling or taste™, but i like `ActionList.Item.Label` there because `ActionList.Item` is where you would put your className 🤔 

❌ bad:

```tsx
<ActionList>
  <ActionList.Item><span className={styles.myLabel}>label</span></ActionList.Item>
</ActionList>

<style>
  .myLabel {
    font-weight: bold
  }
</style>
```

✅  good:

```tsx
<ActionList>
  <ActionList.Item className={styles.myActionListItem}>label</ActionList.Item>
</ActionList>

<style>
  // with direct data-component, good!:
  .myActionListItem [data-component="ActionList.Item.Label"] {
    font-weight: bold
  }
</style>
```

I think I was wrong about `ActionBar.Menu.IconButton`. If `ActionBar.Menu` is part of the public API, then `ActionBar.Menu.IconButton` should be the better choice.

Do you mind also including `Pagination` and `SelectPanel` in this PR? I think those are the most opaque components, once we go through them, all the rest will be straightforward.

packages/react/src/ActionList/Group.tsx
francinelucca: > does that mean we should always have data-component for wrappers?

EDIT: no, quoting [ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-023-stable-selectors-api.md): "Elements that are purely for layout and have no semantic meaning (spacers,
wrappers that exist only for CSS grid/flex layout) do not require this
attribute."

therefor also no
> For this case, should we do: data-component="GroupHeading.Wrapper" and data-component="GroupHeading"?

packages/react/src/TooltipV2/Tooltip.tsx
```diff
@@ -349,6 +349,7 @@ export const Tooltip: ForwardRefExoticComponent<
             className={clsx(className, classes.Tooltip)}
             ref={tooltipElRef}
             data-direction={calculatedDirection}
+            data-component="Tooltip"
```
francinelucca, line 351: does this need to be TooltipV2? I feel like no but are we ok with conflicts with the other Tooltip?

packages/react/src/ActionList/Item.tsx
```diff
@@ -330,6 +337,7 @@ const UnwrappedItem = <As extends React.ElementType = 'li'>(
             {...wrapperProps}
             className={classes.ActionListContent}
             data-size={size}
+            data-component="ActionList.Item.Wrapper"
```
francinelucca: i'm almost certain we should remove this one...

packages/react/src/DataTable/Table.tsx
francinelucca, line 393: I didn't add a data-component to this because I wasn't sure where 😅. it's literally just made up off other components that already have data-component, so I'd need to override the Table maybe? but that also feels wrong, but this is part of the public API so it also feels wrong for it not to have a top-level one...

The same problem with [DataTable](https://github.com/primer/react/blob/main/packages/react/src/DataTable/DataTable.tsx)

packages/react/src/internal/components/ButtonReset.tsx
```diff
@@ -6,7 +6,7 @@ import {clsx} from 'clsx'
  */
 export const Button = ({children, className, ...rest}: React.ComponentPropsWithoutRef<'button'>) => {
   return (
-    <button className={clsx(className, classes.ButtonReset)} type="button" {...rest}>
+    <button className={clsx(className, classes.ButtonReset)} type="button" data-component="Button" {...rest}>
```
francinelucca: thoughts on this having the same data-component as packages/react/src/Button/ButtonBase.tsx? 🤔 Looks like this is just used in Table, wondering if it should be Table.something then?

packages/react/src/ActionList/Item.tsx
siddharthkp: agreed

i think exposing internal wrappers is very risky because that implementation detail should be able to change without any breaking changes

packages/react/src/internal/components/ButtonReset.tsx
siddharthkp: as compared to something like `Table.ResetButton`? (which is something that should probably be set in the Table and not here)

packages/react/src/TooltipV2/Tooltip.tsx
siddharthkp: Nah, just Tooltip

.changeset/datatable-stable-selectors.md
```diff
@@ -0,0 +1,20 @@
...
+- **Link**
+- **LinkButton**
+- **Pagination**
+- **SelectPanel** and friends
+- **Table** and friends
+- **TextInput**
+- **TextInputwithTokens**
+- **TooltipV2**
+
+This enables consumers to query and test components using stable selectors like `[data-component="Table.Row"]`.
```
siddharthkp, line 20: General comment, that's why I'm leaving it on the changelog.

How i understand it, the problem that we're trying to solve with data-component is to expose stable selectors which otherwise would be hidden or tucked away.

Some of these components are top level components that support className and are not composed in an invisible manner inside another component (like selectpanel items), should we still add data-component to them?

francinelucca: > and are not composed in an invisible manner inside another component

I think this is the key here. Which one of these falls into that category? 🤔 The way I see it these all are either:

- parts of bigger components (ex: Button inside SelectPanel, Button inside table, Tooltip inside everywhere), so we want the data-component to be able to drill down:

image

- Big components with smaller parts (SelectPanel has ActionList, Table has Button, ActionBar has Button...), so we need the top-level data-component to "inherit from":
image

siddharthkp (approved):
Please merge

packages/react/src/internal/components/ButtonReset.tsx
francinelucca: Fixed!

packages/react/src/ActionList/Item.tsx
francinelucca: Removed!

francinelucca:
[last run](https://github.com/github/github-ui/pull/19160) looks good, skipping IC manually since last update was just VRT

PR #7734, opened 2026-04-07
add data-component attributes for Avatar, AvatarStack, Banner, & BaseStyles

llastflowers:
Relates to https://github.com/github/primer/issues/6497

#### New

Add data-component attributes and associated for Avatar, AvatarStack, Banner, & BaseStyles

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/AvatarStack/AvatarStack.test.tsx
```diff
@@ -25,9 +25,24 @@ const rightAvatarComp = (
...
+      const body = container.querySelector('[data-component="AvatarStack.Body"]')
+      expect(body).toBeInTheDocument()
+    })
+  })
+
   it('respects alignRight props', () => {
     const {container} = render(rightAvatarComp)
-    expect(container.firstChild).toMatchSnapshot()
+    const root = container.querySelector('[data-component="AvatarStack"]')
+    expect(root).toHaveAttribute('data-align-right', '')
```
francinelucca, line 45: already being useful! 😛

packages/react/src/Banner/Banner.tsx
```diff
@@ -190,6 +193,7 @@ export const Banner = React.forwardRef<HTMLElement, BannerProps>(function Banner
         </div>
         {dismissible ? (
           <IconButton
+            data-component="Banner.Dismiss"
```
francinelucca: I'm leaning we don't need this because you can do `[data-componnet="Banner"] [data-component="IconButton"]` 

I think we're going for no overrides unless specificity clashes , so unless there's another IconButton that needs distinction from this one, I don't think we need this.

packages/react/src/Banner/Banner.tsx
```diff
@@ -171,9 +172,11 @@ export const Banner = React.forwardRef<HTMLElement, BannerProps>(function Banner
         data-layout={rest.layout || 'default'}
         data-flush={flush ? '' : undefined}
       >
-        <div className={classes.BannerIcon}>{visual && supportsCustomIcon ? visual : iconForVariant[variant]}</div>
-        <div className={classes.BannerContainer}>
-          <div className={classes.BannerContent}>
+        <div data-component="Banner.Icon" className={classes.BannerIcon}>
+          {visual && supportsCustomIcon ? visual : iconForVariant[variant]}
+        </div>
+        <div data-component="Banner.Container" className={classes.BannerContainer}>
```
francinelucca: I'm back and forth on this one. If we're doing data-component for containers, then BannerActionsContainer needs one as well. But I don't know that we need containers to have a data-component 🤔. Wondering if @lukasoppermann or @siddharthkp have opinions here? to container or not to container

packages/react/src/Banner/Banner.tsx
```diff
@@ -267,7 +283,13 @@ export type BannerSecondaryActionProps = Omit<ButtonProps, 'variant'>
...
 const BannerSecondaryAction = forwardRef(({children, className, ...rest}, forwardedRef) => {
   return (
-    <Button ref={forwardedRef} className={clsx('BannerPrimaryAction', className)} variant="invisible" {...rest}>
+    <Button
+      data-component="Banner.SecondaryAction"
+      ref={forwardedRef}
+      className={clsx('BannerPrimaryAction', className)}
+      variant="invisible"
+      {...rest}
+    >
```
francinelucca, line 291: not relevant to this PR but why is this class=BannerPrimaryAction? 😳

packages/react/src/Banner/Banner.tsx
francinelucca: ok actually, going off the [ADR](https://github.com/primer/react/blob/main/contributor-docs/adrs/adr-023-stable-selectors-api.md):

>Elements that are purely for layout and have no semantic meaning (spacers,
wrappers that exist only for CSS grid/flex layout) do not require this
attribute.

so I'm making a ruling: no we do not need this

packages/react/src/Banner/Banner.tsx
llastflowers: Makes sense! I wasn't totally sure about this one myself, so I'm okay with just removing it.

packages/react/src/Banner/Banner.tsx
llastflowers: Got it! Thanks, this helps clarify a couple of other things I wasn't super sure about either! I'll remove this one

packages/react/src/Banner/Banner.tsx
llastflowers: Good point! I looked into it and it seems like it's been this way since the inception of `Banner.tsx` 😅 Probably just a copy/paste error by the looks of it! But another thing aside from that—it seems like this might actually be a legacy thing that isn't even in use?? Unless I'm just not understanding how its used 🤔 I'd be curious to know what you think!

francinelucca: my guess would be if you want a stable class to select 'cause after ours are computed they get a random value added to them, but this definitely feels like an error 🤔 , either way let's keep it outside the scope of this PR

packages/react/src/Banner/Banner.tsx
francinelucca: Can you still add a test for it though? here's a sample I did from ActionBar:

image

francinelucca (changes requested):
I think after https://github.com/primer/react/pull/7734#discussion_r3066698688 we're good to go here 👍🏽

francinelucca (approved):
:sparkles:

PR #7735, opened 2026-04-07
Breadcrumbs: Only show the previous breadcrumb on narrow viewports

liuliu-dev:
Closes https://github.com/github/primer/issues/6371

#### Changed

On narrow viewports (≤544px), breadcrumbs collapse to show only the previous (parent) item as a back link. The current page is hidden since you're already on it. Consumers can pass `visibleItemsOnNarrow` to control how many items stay visible.
 

https://github.com/user-attachments/assets/432e5d0a-581e-4e6a-8b9d-875edde955dd

### Rollout strategy

- [X] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


e2e/components/Breadcrumbs.test.ts
```diff
@@ -37,4 +38,57 @@ test.describe('Breadcrumbs', () => {
       }
     })
   }
+
+  test.describe('NarrowVisibleItems', () => {
```
francinelucca, line 42: it's not very common for us to use e2e tests for behavioral functionality like this 👀 . Not saying its bad but I think let's definitely get VRT testing in as well

francinelucca (changes requested):
Sorry for lack of clarity in the original issue. I think what makes the most sense is to show the last breadcrumb item (i.e., where you are currently), and have all the other breadcrumbs wrapped in an action menu:

Image


Image


You can see that in action in this link https://primer.style/react/storybook/?path=/story/components-breadcrumbs-features--overflow-menu-feature-flag-disabled&globals=featureFlags.primer_react_breadcrumbs_overflow_menu:!true (notice the FF on) when you reduce the screensize all the way down until only one item is showing.

This PR also has a nice visual of what I'm describing https://github.com/primer/react/pull/7502

This is only my interpretation of the slack thread however, might be worth syncing with  @adierkens/@tylerjdev/@siddharthkp on requirements

liuliu-dev:
Since the [graduation PR](https://github.com/primer/react/pull/7752) already merged into `main`, and we are going to add the action menu, would be easier to start a new branch off `main`, closing this PR.

PR #7746, opened 2026-04-10
fix(TokenTextContainer): Update the line-height to use a primer primitives CSS variable

jonrohan:
While looking at https://github.com/github/github-ui/pull/18258 I found that the TokenTextContainer was using `line-height: normal;` I think this might have been a bug when we converted the component from styled components. As the previous styled system `lineHeight: normal` would have converted to the "normal" value of the css variable.

I'm fixing by using the CSS variable `var(--base-text-lineHeight-normal)`;

#### Changed

Fixing TokenTextContainer line-height to use primer primitives

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

I'm not expecting this to make too much of a difference in prod, the bug I noticed was very subtle line height shift. Here's a gif toggling between the change:

CleanShot 2026-04-10 at 09 34 53

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/small-comics-kick.md
+++ b/.changeset/small-comics-kick.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Update TokenTextContainer `line-height` to use primer primitives CSS variable `var(--base-text-lineHeight-normal)`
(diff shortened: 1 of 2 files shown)


jonrohan:
I'm skipping integration tests because this is just a CSS change

TylerJDev (approved):

PR #7760, opened 2026-04-15
Mark useResponsiveValue as deprecated

siddharthkp:
[adr-018-responsive-values.md](https://github.com/primer/react/blob/bb18a24bd00281c43e9fd77b0e3bbff0fbe8de20/contributor-docs/adrs/adr-018-responsive-values.md) discourages the use of `useResponsiveValue` hook because it relies on matchMedia and needs to wait until client hydration to choose the right value.

> Authors can use a hook called useResponsiveValue to support resolving the value of a prop based on the current viewport size. The implementation of this hook uses [matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) which, unfortunately, has a downside: the value of a prop may shift when the component is server-side rendered. If the prop is used for styles or layout, then this will lead to a layout shift when the component hydrates and the viewport size is different than the fallback size on the server.

But, we still export this hook without any warnings. This came up in a [conversation on slack](https://github.slack.com/archives/C08FX205QG2/p1776100350655259).

Marking the hook as deprecated in the docs because we want people to use css media queries instead.

Bonus:
- Add a reusable `.github/skills/deprecations/SKILL.md` guide for deprecating components and hooks.
- Update `contributor-docs/deprecating-components.md` to document how deprecated status should be represented in component and hook docs metadata.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/contributor-docs/deprecating-components.md
+++ b/contributor-docs/deprecating-components.md
@@ -11,7 +11,8 @@ When Primer maintainers are to deprecate a component, they will issue `@deprecat
 Issuing a `@deprecated` notice in a minor/patch release includes:
 
 - [ ] Adding a `@deprecated` annotation in the component's source code.
-- [ ] Adding a `Deprecation` section to the documentation of the component with the link of the recommended component and provide a diff. See [deprecated ActionList docs](https://primer.style/react/deprecated/ActionList#deprecation) as an example.
+- [ ] Marking the component or hook docs metadata as deprecated with `"status": "deprecated"` in the relevant `*.docs.json` or `*.hookDocs.json` file.
+- [ ] If the component has a docs page, adding a `Deprecation` section to the documentation with the link of the recommended component and provide a diff. See [deprecated ActionList docs](https://primer.style/react/deprecated/ActionList#deprecation) as an example.
 
 ## Developing a replacement component
 
--- a/packages/react/script/hooks-json/hook.schema.json
+++ b/packages/react/script/hooks-json/hook.schema.json
@@ -80,7 +80,7 @@
     },
     "status": {
       "type": "string",
-      "enum": ["deprecated"],
+      "enum": ["ready", "deprecated"],
       "description": "The status of the hook."
     },
     "importPath": {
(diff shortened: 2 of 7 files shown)


contributor-docs/deprecating-components.md
```diff
@@ -11,7 +11,8 @@ When Primer maintainers are to deprecate a component, they will issue `@deprecat
 Issuing a `@deprecated` notice in a minor/patch release includes:
 
 - [ ] Adding a `@deprecated` annotation in the component's source code.
-- [ ] Adding a `Deprecation` section to the documentation of the component with the link of the recommended component and provide a diff. See [deprecated ActionList docs](https://primer.style/react/deprecated/ActionList#deprecation) as an example.
+- [ ] Marking the component or hook docs metadata as deprecated with `"status": "deprecated"` in the relevant `*.docs.json` or `*.hookDocs.json` file.
+- [ ] If the component has a docs page, adding a `Deprecation` section to the documentation with the link of the recommended component and provide a diff. See [deprecated ActionList docs](https://primer.style/react/deprecated/ActionList#deprecation) as an example.
```
francinelucca, line 15: this link does not exist?

packages/react/script/hooks-json/hook.schema.json
```diff
@@ -78,6 +78,11 @@
       "type": "string",
       "description": "The name of the hook."
     },
+    "status": {
+      "type": "string",
+      "enum": ["deprecated"],
+      "description": "The status of the hook."
+    },
```
francinelucca, line 85: only valid value is deprecated? 🤔

francinelucca (changes requested):

packages/react/script/hooks-json/hook.schema.json
siddharthkp: Yes... because there was no status field so far.

Made it `["ready", "deprecated"]` so it doesn't look weird

contributor-docs/deprecating-components.md
siddharthkp: My bad, that link was already there. Replaced with a link for a recent deprecated component

francinelucca (approved):

PR #7766, opened 2026-04-17
Make popover in `AnchoredOverlay` opt-in

TylerJDev:
Closes https://github.com/github/primer/issues/6556 & https://github.com/github/primer/issues/6557

Makes usage of the popover API opt-in, rather than have it on by default. 
We may want to have the popover API enabled by default in the future, so the prop usage could be temporary.

#### Changed

* Make popover API usage in `AnchoredOverlay` | `Overlay` opt-in

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [x] Tested in Firefox
- [x] Tested in Safari
- [x] Tested in Edge


packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -120,6 +120,10 @@ interface AnchoredOverlayBaseProps extends Pick<OverlayProps, 'height' | 'width'
    * Props to be spread on the close button in the overlay.
    */
   closeButtonProps?: Partial<IconButtonProps>
+  /**
+   * When enabled (and CSS anchor positioning feature flag is on), uses the Popover API
```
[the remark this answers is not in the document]
siddharthkp: @TylerJDev good comment from copi. should this prop always render in popover regardless of the feature flag or will that cause bugs?

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -120,6 +120,10 @@ interface AnchoredOverlayBaseProps extends Pick<OverlayProps, 'height' | 'width'
...
+   */
+  usePopover?: boolean
```
siddharthkp: fair name, but clashes with what hooks are named like 😅 

some alternatives: 
- more verbose:`renderAsPopover`, `renderInPopover`. 
- negative prop: `renderInPortal={false}` which disables the default rendering in portal and uses popover instead
- native sounding: maybe just  `popover` like the html attribute (even though we don't support the values of the html attribute - auto, manual, hint)
- claritymaxxing: `renderAs: portal | popover` or `renderTarget: portal | popover`

siddharthkp (commented):
Trust you with the implementation. Left a couple comments about direction and naming.

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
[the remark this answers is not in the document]
TylerJDev: I don't think it would cause issues, but I'd rather wait until we ship CSS anchor positioning before we allow consumers to use it, as this gives us time to test it safely (under the flag).

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
TylerJDev: Naming is hard 😭 I agree. 

I like `renderAs`, let's go with this one 😎

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -120,6 +120,11 @@ interface AnchoredOverlayBaseProps extends Pick<OverlayProps, 'height' | 'width'
    * Props to be spread on the close button in the overlay.
    */
   closeButtonProps?: Partial<IconButtonProps>
+  /**
+   * When `true`, uses the Popover API only if the CSS anchor positioning feature flag is enabled
+   * and the browser supports native CSS anchor positioning. Has no effect otherwise. Defaults to `false`.
+   */
+  renderAs?: 'portal' | 'popover'
```
francinelucca, line 127: the description seems to imply this is a booolean?

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -162,10 +167,13 @@ export const AnchoredOverlay: React.FC<React.PropsWithChildren<AnchoredOverlayPr
   onPositionChange,
   displayCloseButton = true,
   closeButtonProps = defaultCloseButtonProps,
+  renderAs = 'portal',
 }) => {
   const cssAnchorPositioningFlag = useFeatureFlag('primer_react_css_anchor_positioning')
   const supportsNativeCSSAnchorPositioning = useRef(false)
   const cssAnchorPositioning = cssAnchorPositioningFlag && supportsNativeCSSAnchorPositioning.current
+  // Only use Popover API when both CSS anchor positioning is enabled AND renderAs is true
+  const shouldRenderAs = cssAnchorPositioning && renderAs === 'popover'
```
francinelucca:
```suggestion
  const shouldRenderAsPopover = cssAnchorPositioning && renderAs === 'popover'
```

francinelucca (changes requested):

packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
TylerJDev: Ah, good catch! Updated!

francinelucca (approved):

PR #7772, opened 2026-04-22
chore: update snapshots

francinelucca:
- update snapshots
- skip flakey VRT tests and remove snapshots


TylerJDev (approved):

llastflowers (approved):

francinelucca:
FYI @liuliu-dev I skipped the ActionList tooltip tests here because they're flakey due to the tooltip being in different stages of showing up when the screenshot is taken. If we want to get this back in I think we need to figure out a mechanism to wait until the tooltip has fully rendered - perhaps a delay but I don't know if it'd be enough - before taking the screenshot to stabilize the VRTs

PR #7778, opened 2026-04-27
🚧 [WIP] Dialog 4 layer specs

lukasoppermann:
This is a WIP test and nothing to be judged yet.


contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+
+**Why prop-getters over components:**
+
+- Full markup ownership — consumer chooses every element
+- Composable with any component system (Radix, custom, etc.)
+- Multi-element wiring is natural (`getTitleProps()`, `getBodyProps()`)
+- TypeScript return types are explicit and statically known
+- No imposed component tree
+
+**Foundation CSS:** Each foundation may ship a minimal CSS reset that removes browser defaults without adding visual opinion. Use `:where()` for zero specificity so consumer styles always win.
```
joshblack: One idea for this instead of `:where()` is that these live in an earlier layer in the css cascade layer list (or just lives in a css cascade layer)

contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+
+**Context** is allowed internally for ARIA cross-wiring (e.g., `aria-labelledby` pointing title ID to dialog) but is never exposed to consumers.
```
joshblack: Is there a convention for providing the context root? Would this be a component? (e.g. `Dialog.Root` for the above example)

contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+**Rules:**
+
+- Parts use Foundations internally — they call `useComponentFoundation()` and spread prop-getters
+- Parts add Primer design tokens, CSS modules, and layout opinions
+- Parts are the building blocks for Ready-made (Layer 1)
+- All Parts must include `data-component` attributes per [ADR-023](./adr-023-stable-selectors-api.md)
+
+### Stable selectors (ADR-023)
+
+All Layer 2 Parts and Layer 1 Ready-made components must include `data-component` attributes as defined in [ADR-023](./adr-023-stable-selectors-api.md).
```
joshblack: Let me know if this is covered in the ADR, but if we're providing parts/slots why would we need `data-component`? I would have assumed `className` but let me know if that's not the case

contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+      <h2 data-component="Dialog.Title">Title</h2>
+      <button data-component="Dialog.CloseButton">✕</button>
+    </header>
+    <div data-component="Dialog.Body">Content</div>
+    <footer data-component="Dialog.Footer">...</footer>
+  </div>
+</dialog>
+```
+
+### Layer 1 — Ready-made
```
joshblack: Do we want this for all components? This is an open question I have if this is simpler than the alternatives (I just don't know). Sometimes the config approaches can lead to massive types and assume it's what makes some components more unwieldy (like select panel)

Maybe asked another way, are there limitations on the scope of this API that we would like to put on a component? Should everything be available through this or is this capturing the 80% use-case?

contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+## Export & package structure
+
+### Entry points
+
+| Layer           | Stable import               | Experimental import                      |
+| --------------- | --------------------------- | ---------------------------------------- |
+| 1 — Ready-made  | `@primer/react`             | `@primer/react/experimental`             |
+| 2 — Parts       | `@primer/react`             | `@primer/react/experimental`             |
+| 3 — Foundations | `@primer/react/foundations` | `@primer/react/foundations/experimental` |
+| 4 — Hooks       | `@primer/react/hooks`       | `@primer/react/hooks/experimental`       |
```
joshblack: More of a meta/organization comment, happy to drop more entry points and just use an `unstable_*` convention (personally). That way people just import from the same entrypoints

contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Naming conventions
+
+| Layer | Convention                 | Example                                  |
+| ----- | -------------------------- | ---------------------------------------- |
+| 4     | `use<Behavior>`            | `useScrollLock`, `useFocusTrap`          |
+| 3     | `use<Component>Foundation` | `useDialogFoundation`                    |
```
joshblack: Super bikeshed-y but I would rather these just be called whatever role/grouping/etc they are than include foundation in the name.

contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+  "./hooks/experimental": {
+    "types": "./dist/hooks/experimental/index.d.ts",
+    "default": "./dist/hooks/experimental/index.js"
+  }
+}
+```
+
+## Alternatives considered
+
+### Components instead of prop-getters for Layer 3
```
joshblack: I would love the option to ship both, potentially. One thing I would be concerned about is that there is a claim that if you use these foundations then what you build is accessible. However, that may only be true if specific criteria are met (one could be that headings must be a descendant of the dialog or aria-modal node). This could be more easily enforced through components than the hooks API (or at least that's my assumption).

contributor-docs/adrs/adr-024-modular-component-architecture.md
```diff
@@ -0,0 +1,277 @@
...
+
+Render props were considered for Layer 2 but rejected:
+
+- They already exist in legacy code — we don't want to add more
+- Slots via `useSlots` are more declarative and composable
+- `React.Children` + `React.cloneElement` are fragile and discouraged by React team
+
+## Consequences
+
+- Every new component should be built using this 4-layer decomposition
```
joshblack: This is a scary sounding line but I think I can get behind it if the idea is:

- Use appropriate markup/semantics for the component role
- Create presentational components with primer-styles
- Add behavior through hooks (whether component-specific or semantic-specific)
- Combine to make slots/config components

So if I'm making a Tabs component I would:

- Define the markup/semantics for `role="tabs"`, `role="tab"`, `role="tabpanel"` (either through hooks or components)
- Create presentational components for these with styles
- Add behavior through hooks (`useTabs`) to allow for interaction
- Combine these into a broader `<Tabs>` component with support for config/slots

One thing that I'm not sure about is if we would necessarily want config for every component. I'm not sure if, in all cases, that is easier to use than the slot style. Curious what other people thing 👀

joshblack (commented):
Thanks for putting this all together! Just left a comment on the ADR if that's okay 👀

contributor-docs/adrs/adr-024-modular-component-architecture.md
lukasoppermann: @joshblack I added this because we have it now. If we agree on this we can think about if this is made completely redundant, or not.

contributor-docs/adrs/adr-024-modular-component-architecture.md
lukasoppermann: Fine with this. I think this is something where we are very happy to just follow primer eng's lead.

contributor-docs/adrs/adr-024-modular-component-architecture.md
lukasoppermann: I agree, this is a leftover I missed. 😞

contributor-docs/adrs/adr-024-modular-component-architecture.md
joshfarrant: Totally agree, I think this is an oversight in the current ADR rather than a deliberate choice. I'll update the ADR to reflect this.

Regarding accessibility, I had the same thought and I think it needs to be very explicitly communicated what is handled by each layer and what the responsibilities of the implementer are at each layer. I'm envisioning an accessibility checklist for each layer of particular things that we know an implementer must still do if they choose to use a component at a particular layer.

contributor-docs/adrs/adr-024-modular-component-architecture.md
joshfarrant: Yea I think `DialogRoot` would be the context root. And leaning towards flat exports would probably be the way forward too for better RSC compat. E.g., `DialogHeader` over `Dialog.Header`.

contributor-docs/adrs/adr-024-modular-component-architecture.md
joshfarrant: I don't think it should be mandatory for every component. To be honest I could see a world where L1 doesn't exist at all and essentially we just point people to a cookbook/recipes for common use cases which compose L2 parts into various standalone components. 

For the same reasons you described I'm not a fan of config/prop-driven components either. Sometimes a huge number of props, often taking objects as props which then causes a cascade of unnecessary renders if people forget to memoize their objects.

I don't feel we'd actually _lose_ anything by ditching L1 entirely. There might be cases where it really makes sense, but I'd always prefer slightly more verbose+composable than stuffing everything into a single component's props.

When I first suggested Layer 1 it was more as a way to make the proposal a bit more approachable than jumping straight to composable parts as the default. If eng is open to it though we can explore what it might look like to not have L1 by default and then see if there are any scenarios in which we feel it adds genuine value.

contributor-docs/adrs/adr-024-modular-component-architecture.md
joshfarrant: +1 to your tabs breakdown, that's exactly the mental model I have.

I touched on the config point in another comment too. TL;DR there is we might not need it

PR #7816, opened 2026-05-07
Re-add (unrevert) "Replace ActionBar overflow calculations with CSS wrapping approach"

iansan5653:
Re-implements https://github.com/primer/react/pull/7655 after it was reverted in https://github.com/primer/react/pull/7807, with fixes:

1. Updated the `CommentBox` story to reproduce the issue (it seems to have been caused by a particular set of container styles, most importantly `overflow: hidden` on the container
2. Reduce the `IntersectionObserver` threshold from `1`to `0.75` so that it will fire reliably - the issue was caused by the observer not being quite sensitive enough; the buttons appear to have been still overflowing by just 1 pixel, which meant they were never considered 100% visible
3. Remove a style condition to always apply the scroll driven animation and thus hopefully resolve some flickering (which just appeared to be caused by React safe mode, but I still wanted to resolve it)

This reverts https://github.com/primer/react/pull/7807.

[➡️ **View just the new changes**](https://github.com/primer/react/pull/7816/changes/dcc04715fd259ecefe2ee8a8640ae32d5b72814c..30a5cc005def2fec6858c97d03ffa95ecdcc4e16)


iansan5653:
@copilot this e2e test failure was likely introduced by the story changes in 560b2a8; investigate and resolve
```
17) e2e/components/drafts/ActionBar.test.ts:42:13 › ActionBar › ActionBar Interactions › dark_colorblind › Overflow interaction @vrt 

    Test timeout of 15000ms exceeded.

    Error: locator.waitFor: Test timeout of 15000ms exceeded.
    Call log:
      - waiting for getByLabel('Task List') to be hidden
        34 × locator resolved to visible <button type="button" tabindex="-1" data-size="medium" data-loading="false" data-no-visuals="true" aria-labelledby=":rt:" data-variant="invisible" data-component="IconButton" class="prc-Button-ButtonBase-c50BI prc-Button-IconButton-szpyj">…</button>


      50 |           await expect(page.locator(toolbarButtonSelector).filter({visible: true})).toHaveCount(10)
      51 |           await page.setViewportSize({width: viewports['primer.breakpoint.xs'], height: 768})
    > 52 |           await page.getByLabel('Task List').waitFor({
         |                                              ^
      53 |             state: 'hidden',
      54 |           })
      55 |           await expect(page.locator(toolbarButtonSelector).filter({visible: true})).toHaveCount(8)
        at /github/workspace/e2e/components/drafts/ActionBar.test.ts:52:46

```

packages/react/src/ActionBar/ActionBar.examples.stories.module.css
```diff
@@ -69,3 +78,7 @@
 .MultipleActionBarsSection {
   padding: var(--base-size-16);
 }
+
+.OverflowHidden {
+  overflow: hidden;
+}
```
francinelucca: nit: is this used anywhere?

francinelucca (changes requested):
Looking good but I think I found an error: looks like I can move the divider to the front by resizing 👀 

https://github.com/user-attachments/assets/b41e7010-2450-445d-952b-c9e9a8e45686

packages/react/src/ActionBar/ActionBar.examples.stories.module.css
iansan5653: Removed in a844f6d

iansan5653:
> Looking good but I think I found an error: looks like I can move the divider to the front by resizing 👀

Looks like the fake 0-width first item which was supposed to allow the real first item (the first group) to wrap out of view was too short, so that the first item only wrapped down a couple of pixels. This was enough for the `offsetTop` logic to think it was overflowed and hide it, but not enough for the `IntersectionObserver` to trigger again and show it, because it was never more than 75% hidden (the new threshold). It sort of worked before because the old threshold was 100%, but that was just hiding the bug.

The fix is simple - just make the fake item fill the height of the container, so that it pushes other items fully out of view when they wrap.

francinelucca (approved):
😍

PR #7824, opened 2026-05-08
Fix `usePaneWidth` triggering unnecessary React re-renders on every window resize

### LINKED ISSUE #7801: `usePaneWidth` triggers unnecessary React re-renders on every window resize
### Problem

👋🏼 Hello, while investigating a report of performance lag on https://github.com/github/pull-requests/issues/24571, I did a bit of profiling with Copilot.

I've made some changes on our end, but I would love it if y'all would investigate the below findings.

When `SplitPageLayout` has a resizable pane (`resizable={true}`), the `usePaneWidth` hook registers a `window.addEventListener("resize", handleResize)` listener that calls `syncAll()` on every resize event (~60/sec during a drag).

`syncAll()` **unconditionally** calls:

```js
startTransition(() => {
  setMaxPaneWidth(actualMax);
  if (wasClamped) {
    setCurrentWidthState(actualMax);
  }
});
```

Even when the computed `actualMax` hasn't changed from the previous value, `setMaxPaneWidth(actualMax)` still schedules a React concurrent render because React doesn't bail out of `startTransition` state updates that set the same value (unlike normal `setState`).

### Impact

On a page with multiple components consuming `SplitPageLayout` context (e.g. GitHub PR conversation tab), this produces **11 React scheduler calls totaling ~108ms** in a single animation frame during resize — with 4× CPU throttle, this pushes individual frames well beyond 100ms and creates visible jank.

### Trace evidence

- Chrome Performance trace on a GitHub PR conversation page (4× CPU throttle)
- The `actualMax` value doesn't change between most resize ticks (only changes when crossing the `DEFAULT_PANE_MAX_WIDTH_DIFF_BREAKPOINT`)

### Suggested fix

Add a guard before the state update:

```js
const syncAll = () => {
  // ... existing logic ...
  const actualMax = getMaxPaneWidthRef.current();
  paneRef.current?.style.setProperty("--pane-max-width", `${actualMax}px`);

  // Only trigger React re-render if values actually changed
  const maxChanged = actualMax !== maxPaneWidthRef.current;
  const wasClamped = currentWidthRef.current > actualMax;

  if (maxChanged || wasClamped) {
    startTransition(() => {
      setMaxPaneWidth(actualMax);
      if (wasClamped) {
        setCurrentWidthState(actualMax);
      }
    });
  }
};
```

This would skip the `startTransition` entirely when the computed max hasn't changed — which is the common case during horizontal resize that doesn't cross the breakpoint.

### Additional context

- The inline style updates (`--pane-max-width`, `--pane-width`) and ARIA updates are fine — they don't trigger React renders
- Consumers can't work around this without disabling `resizable` entirely

### Steps to reproduce

1. Go to a Conversation page on a PR
2. Throttle CPU
3. Resize the window 
4. Observe lag

### Version

v38.21.0

### Browser

_No response_

jonrohan:
Closes #7801

### Overview

When `SplitPageLayout` has a resizable pane (`resizable={true}`), the `usePaneWidth` hook calls `startTransition(() => setMaxPaneWidth(actualMax))` inside `syncAll()` on **every** window resize event — even when `actualMax` hasn't changed. Unlike normal `setState`, `startTransition` does **not** bail out on same-value updates, so this triggers unnecessary React concurrent renders on every resize tick (~60/sec during a drag).

On pages with many components consuming the layout context (e.g. GitHub PR conversation tab), this produces ~11 React scheduler calls totaling ~108ms in a single animation frame with CPU throttle, causing visible jank.

**Fix:** Added a `maxPaneWidthRef` to track the previous `maxPaneWidth` value, and guarded the `startTransition` call so it only fires when `actualMax` has actually changed or when the current width needs to be clamped. The CSS variable updates, ref updates, and ARIA updates (which don't cause React re-renders) remain unconditional.

#### New

- Added a Storybook performance test story (`Re-render Counter (Issue #7801)`) to help verify resize re-render behavior with React DevTools Profiler.

#### Changed

- `usePaneWidth`: `syncAll()` now skips `startTransition` when the computed max pane width hasn't changed, avoiding unnecessary React re-renders on window resize.

#### Removed

_None_

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

1. Open the new Storybook story: **Components/PageLayout/Performance Tests/Re-render Counter (Issue #7801)**
2. Open React DevTools Profiler, enable "Highlight updates when components render"
3. Resize the browser window within the same side of the 1280px breakpoint
4. Verify no unnecessary re-renders occur (components should not flash/highlight)
5. Resize across the 1280px breakpoint and verify the state does update correctly

Unit test added: `should skip startTransition when maxPaneWidth has not changed (#7801)` in `usePaneWidth.test.ts`

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/PageLayout/PageLayout.performance.stories.tsx
+++ b/packages/react/src/PageLayout/PageLayout.performance.stories.tsx
@@ -543,7 +543,7 @@ export const KeyboardARIATest: Story = {
 // ============================================================================
 
 export const ResizablePaneReRenderCounter: Story = {
-  name: 'Re-render Counter (Issue #7801)',
+  name: 'Re-render Counter',
   render: () => {
     return (
       <SplitPageLayout>
--- a/packages/react/src/PageLayout/PageLayout.performance.stories.tsx
+++ b/packages/react/src/PageLayout/PageLayout.performance.stories.tsx
@@ -1,6 +1,7 @@
 import React, {useState} from 'react'
 import type {Meta, StoryObj} from '@storybook/react-vite'
 import {PageLayout} from './PageLayout'
+import {SplitPageLayout} from '../SplitPageLayout'
 import {Button} from '../Button'
 import Label from '../Label'
 import Heading from '../Heading'
@@ -532,6 +533,50 @@ export const KeyboardARIATest: Story = {
           </div>
         </PageLayout.Content>
       </PageLayout>
+    )
+  },
+}
+
+// ============================================================================
+// Story: Re-render test — reproduces unnecessary re-renders from usePaneWidth
+// https://github.com/primer/react/issues/7801
+// ============================================================================
+
+export const ResizablePaneReRenderCounter: Story = {
+  name: 'Re-render Counter (Issue #7801)',
+  render: () => {
+    return (
+      <SplitPageLayout>
+        <SplitPageLayout.Header>
+          <div style={{padding: '16px'}}>
+            <Heading as="h2">Re-render Test — Issue #7801</Heading>
+            <p style={{marginTop: '8px', fontSize: '14px', color: 'var(--fgColor-muted)'}}>
+              Open React DevTools Profiler, enable &quot;Highlight updates when components render&quot;, then resize the
+              browser window. Before the fix, all components re-render on every resize tick even when the computed max
+              width hasn&apos;t changed. After the fix, renders only happen when crossing the 1280px breakpoint or when
+              current width is clamped.
+            </p>
+          </div>
+        </SplitPageLayout.Header>
+
+        <SplitPageLayout.Pane position="start" resizable aria-label="Side pane">
+          <div style={{padding: '16px'}}>
+            <p>This pane is resizable. Try resizing the browser window.</p>
+          </div>
+        </SplitPageLayout.Pane>
+
+        <SplitPageLayout.Content>
+          <div style={{padding: '16px'}}>
+            <p>Content area — should not re-render on resize unless max width changes.</p>
+          </div>
+        </SplitPageLayout.Content>
+
+        <SplitPageLayout.Footer>
+          <div style={{padding: '16px'}}>
+            <p>Footer area.</p>
+          </div>
+        </SplitPageLayout.Footer>
+      </SplitPageLayout>
     )
   },
 }
--- a/packages/react/src/PageLayout/usePaneWidth.test.ts
+++ b/packages/react/src/PageLayout/usePaneWidth.test.ts
@@ -20,10 +20,12 @@ const {startTransitionSpy} = vi.hoisted(() => ({
   startTransitionSpy: vi.fn(),
 }))
 vi.mock('react', async importOriginal => {
-  const actual = await importOriginal<typeof import('react')>()
+  const actual = (await importOriginal()) as Record<string, unknown> & {
+    startTransition: (...args: [callback: () => void]) => void
+  }
   return {
     ...actual,
-    startTransition: (...args: Parameters<typeof actual.startTransition>) => {
+    startTransition: (...args: [callback: () => void]) => {
       startTransitionSpy(...args)
       return actual.startTransition(...args)
     },
--- a/packages/react/src/PageLayout/usePaneWidth.test.ts
+++ b/packages/react/src/PageLayout/usePaneWidth.test.ts
@@ -853,7 +853,7 @@ describe('usePaneWidth', () => {
       vi.useRealTimers()
     })
 
-    it('should skip startTransition when maxPaneWidth has not changed (#7801)', async () => {
+    it('should skip startTransition when maxPaneWidth has not changed', async () => {
       vi.useFakeTimers()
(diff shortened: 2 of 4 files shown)


jonrohan:
@talum Is this what you had in mind for the fix for https://github.com/primer/react/issues/7801

packages/react/src/PageLayout/PageLayout.performance.stories.tsx
```diff
@@ -535,3 +536,47 @@ export const KeyboardARIATest: Story = {
...
   },
 }
+
+// ============================================================================
+// Story: Re-render test — reproduces unnecessary re-renders from usePaneWidth
+// https://github.com/primer/react/issues/7801
+// ============================================================================
+
+export const ResizablePaneReRenderCounter: Story = {
+  name: 'Re-render Counter (Issue #7801)',
```
francinelucca: nit: don't think we need to include issue number in the name since this'll be a public facing story

packages/react/src/PageLayout/usePaneWidth.test.ts
```diff
@@ -853,6 +853,48 @@ describe('usePaneWidth', () => {
...
+      // Now resize again to a different width that produces the SAME max
+      // 900px -> 900px (no actual viewport change, same max = 389)
+      const renderCountBefore = result.current.maxPaneWidth
+      window.dispatchEvent(new Event('resize'))
+      await act(async () => {
+        await vi.runAllTimersAsync()
+      })
+
+      // maxPaneWidth should be unchanged — no re-render triggered
+      expect(result.current.maxPaneWidth).toBe(renderCountBefore)
```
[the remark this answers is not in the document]
francinelucca: agreed!

packages/react/src/PageLayout/usePaneWidth.test.ts
```diff
@@ -853,6 +853,48 @@ describe('usePaneWidth', () => {
...
+
```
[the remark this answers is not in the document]
francinelucca: agreed here as well!

francinelucca (changes requested):
mostly LGTM, I think copilot's comments here are worth addressing

packages/react/src/PageLayout/usePaneWidth.test.ts
```diff
@@ -853,6 +853,68 @@ describe('usePaneWidth', () => {
...
+      const renderCountAfterMount = renderCount
+
+      // Dispatch resize without changing viewport (same max = same value)
+      window.dispatchEvent(new Event('resize'))
+      await act(async () => {
+        await vi.runAllTimersAsync()
+      })
+
+      // No additional renders should have occurred — startTransition was skipped
+      expect(renderCount).toBe(renderCountAfterMount)
```
francinelucca: can you explain how this works? I don't understand how renderCount would change if the issue wasn't fixed 😅 
Like will this test fail without the changes in usePaneWidth.ts?

jonrohan: Before the fix, the syncAll function (called on every resize event) unconditionally called startTransition(() => setMaxPaneWidth(actualMax)) — even when actualMax hadn't changed from the previous value.

The subtle part is that startTransition does not bail out on same-value updates the way a normal setState call does. With regular setState, React compares old vs. new state and skips the re-render if they're equal. But when you wrap setState inside startTransition, React always schedules a re-render — even if the value is identical. This is a known React behavior.

After the fix, a maxPaneWidthRef tracks the last value, and startTransition is only called when actualMax !== maxPaneWidthRef.current (i.e., the value actually changed).

So in the test: the viewport is set to 900px and stays at 900px when window.dispatchEvent(new Event('resize')) fires. The computed max is the same both times. Without the fix, startTransition would still call setMaxPaneWidth(389) again, triggering a re-render and incrementing renderCount. With the fix, the guard actualMax !== maxPaneWidthRef.current is false, so startTransition is skipped entirely, and renderCount stays at renderCountAfterMount.

francinelucca: I reverted the usePaneWidth.ts changes locally and the tests still pass. Let's either fix or remove this test 🙏🏽 

Image

jonrohan: @francinelucca I made an update that instead of counting renders (which doesn't work because React bails out on same-value setState even inside startTransition), the test now uses vi.mock('react', ...) to spy on startTransition directly. This lets us assert that startTransition is never called when maxPaneWidth hasn't changed.

Verified it fails with main, and passes with these changes

francinelucca (approved):

jonrohan:
Skipping tests because the failures are because of the `useRovingTabIndex` export that was added in latest release.

talum:
> @talum Is this what you had in mind for the fix for #7801

Sorry I missed this! But yes! This looks like it'll help!

PR #7836, opened 2026-05-12
Replace Timeline Playground with Custom Event playground

janmaarten-a11y:
Refs github/primer#6654
Closes github/primer#6662

Phase 1 of the Timeline redesign. Replaces the existing `Components/Timeline → Playground` story with a recreation of the Figma "Custom event" component (Primer-Web library, node `46191-13560`).

This is intentionally **story-only** — no public API changes. The goal is to nail down the prop surface, layout patterns, and naming conventions before Phase 2 introduces named PR / Issue / Dependabot events. The new playground composes existing public primitives only (`Timeline`, `Timeline.Item`, `Timeline.Badge`, `Timeline.Body`, `Timeline.Avatar`, `Timeline.Actions`, `Avatar`, `Link`, `RelativeTime`) and introduces a `data-*` convention on `Timeline.Item` (`data-event-scope`, `data-event-type`, `data-actor-type`) that Phase 4 filtering work will build on.

| New Timeline Playground (custom event) |
|---|
| A storybook playground showing a custom timeline event and a series of controls for actors, badges, event text, and other optional content. The event says: janmaarten-a11y built a new timeline event playground just now — with Copilot + Figma MCP (as a fake custom app name) |

| Old Playground |
|---|
| The current Timeline Playground story showing four generic timeline events that say This is a message and have a commit icon. There are only controls for clipSidebar, condensed, and className |

#### New

- **Custom Event playground story** for `Components/Timeline` recreating the Figma "Custom event" component. Covers 32 badge octicons, all 9 `TimelineBadgeVariant` colors, 5 timestamp presets (3 relative + 2 absolute), paired GitHub Actions / Custom App attribution, and conditional controls (avatar URL only when `actorType: 'user'`, app fields only when `viaApp` is on, custom-app fields only when `appPreset: 'Custom App'`).
- **`data-*` filtering convention** on the rendered `Timeline.Item`: `data-event-scope`, `data-event-type`, `data-actor-type`. No visual effect today; reserved for Phase 4 filtering and the planned summary-events rollup.
- **Story-local CSS module** that reserves a fixed-width left-of-rail gutter so toggling `actorSize` doesn't shift the timeline horizontally. When `actorSize` is `large`, the playground uses `Timeline.Avatar` to position the 40px avatar in this gutter.
- **`Timeline.Actions` integration** — a `showActions` toggle renders right-aligned buttons via `Timeline.Actions`, with `Single button` and `Two buttons` presets.

#### Changed

- Replaced the previous `Components/Timeline → Playground` (the simple `clipSidebar` + `condensed` toggle demo) with the new Custom Event playground. The replaced behaviour stays demonstrated in `Components/Timeline/Features` (`ClipSidebar`, `ClipSidebarStart`, `ClipSidebarEnd`, `CondensedItems`).

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [x] **None** — story-only change, no published-package impact. The `@primer/react` package surface is unchanged. `skip changeset` label applied.

### Testing & Reviewing

Open the new playground at `Components/Timeline → Playground` and try these:

1. **Actor variants** — flip `actorType` through `user / bot / app / copilot`. The `actorName` field auto-fills with a sensible default for each type (`monalisa` / `dependabot` / `GitHub Actions` / `Copilot`) and stays editable, except for `copilot` where it's hidden because the name is fixed. `actorAvatarSrc` only appears for `user`.
2. **Avatar layout** — flip `actorSize` between `small` and `large`. The timeline shouldn't shift horizontally; only the avatar position changes (inline 20px vs 40px in the reserved left gutter).
3. **Timestamp presets** — cycle `eventTimestamp` through all 5:
   - `Relative (now)` → `just now`
   - `Relative (recent day)` → `yesterday` / `X days ago` (live-updating via `<RelativeTime>`)
   - `Relative (3 weeks)` → `3 weeks ago` (live-updating via `<RelativeTime>`)
   - `Absolute (today)` → `Today h:mm AM/PM TZ` (`Intl.DateTimeFormat`)
   - `Absolute (full timestamp)` → `Mon DD, h:mm AM/PM TZ` (`Intl.DateTimeFormat`)
4. **`viaApp` attribution** — toggle on, switch between `GitHub Actions` and `Custom App` presets. Custom App reveals editable name + avatar URL fields. Heads up: this slot is PR-specific in real GitHub usage, which we'll revisit in Phase 2.
5. **`showNote`** — adds a second muted line below the summary. Covers cited-reason patterns (Dependabot dismissal reason, lock reason, merge-queue "due to..." text).
6. **`showActions`** — toggle on to render a `Timeline.Actions` slot with right-aligned buttons. Switch `actionsPreset` between `Single button` (e.g. Compare) and `Two buttons` (e.g. View details + Revert). Resize the viewport to see `Timeline.Body` wrap before `Timeline.Actions` is squeezed.
7. **DOM attributes** — open DevTools and confirm `data-event-scope`, `data-event-type`, `data-actor-type`, and `className` apply to the rendered `Timeline.Item`.

A11y was reviewed via the Storybook a11y panel. One real `link-name` violation (empty `actorName` produced an empty link) was caught and fixed with a fallback. Axe also throws intermittent `incomplete` color-contrast notes on user-typed input — that's axe being unable to predict the final color from dynamic text, not a real issue. The Primer muted tokens are used correctly throughout.

The longer-form description (Figma reference, `data-*` convention, Phase 1 limitations) lives on the story's **Docs tab**. The Controls panel is intentionally compact since most controls explain themselves through their label and immediate visual feedback.

### Things this PR doesn't tackle

A few items from the parent issue we decided to defer, plus some bigger gaps that surfaced while building. None of these are blockers for this PR.

**Deferred from the issue's scope:**

- **Preset stories** alongside the Playground. The Playground covers every variant via controls already, and Phase 2/3 will reorganize the Timeline section into subfolders (primitives / named events / full timelines). Preset stories will have a natural home there; adding them to the current flat structure now would just need to be moved later.
- **`condensed` control.** Only used on 2nd+ rows of a commit group in real timelines, so it isn't really a custom-event property — and toggling it with `actorSize: 'large'` looks wrong. Phase 2's named Commit-group event will apply it automatically. The prop is still public on `Timeline.Item` and shown in `Timeline.features.stories.tsx#CondensedItems`.
- **`subContent` (cross-reference cards, etc.).** No card primitive exists yet in Primer React. Adding one would either over-engineer the playground or do work that Phase 2's named events should own.

### Merge checklist

- [x] Added/updated tests — _N/A; story-only change. Existing 21 Timeline unit tests still pass._
- [x] Added/updated documentation — _JSDoc story description bound to the `Playground` export, visible on the Docs tab._
- [x] Added/updated previews (Storybook) — _the playground itself._
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility) — _no client-only APIs; `Date.now()` captured via lazy `useState` for stability across re-renders._
- [ ] Tested in Chrome
- [x] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [x] (GitHub staff only) Integration tests pass at github/github-ui — _no `@primer/react` package changes; nothing dotcom's Timeline depends on._


janmaarten-a11y:
I went back through the [parent issue's checklist](https://github.com/github/primer/issues/6662) and ran the Storybook a11y panel over the Playground. PR is in good shape; all checks are green.

**What the a11y review found:** one real issue — clearing the `actorName` field left an empty link, which axe flagged. Fixed in `1b9497363` with an "Unknown actor" fallback so the link always has text. Axe also threw some "incomplete" notes on color contrast when you type into the actor or app name fields, but those are axe being unable to predict the final color when the text is dynamic, not a real problem.

**What I'm deferring:** the issue asked for 2–3 preset stories alongside the Playground. I think we're better off waiting. The Playground already covers every variant via controls, and Phase 2/3 are going to need a real reorganization of the Timeline section (subfolders for primitives, individual events, and full timelines). Adding presets to the current flat structure now would just create churn when that IA work happens.

**Things I noticed for Phase 2:**
- We need a real right-controls slot on `Timeline.Item`. PR / Issue / Shared events use it constantly (revert button, view details, restore branch, etc.) and there's no clean way to add it without an API change.
- Same story for an avatar slot — currently faking it with story-local CSS for the large-avatar treatment.
- Worth revisiting the list semantics (`<div>` everywhere right now; richer timelines probably want `<ol>`/`<li>` or at least `role="list"`).
- The existing `Default` story and the Features stories still use the deprecated `Octicon` wrapper. Easy cleanup to roll into the named-events PR.
- The condensed badge color reads too dark — worth a token swap.
- Need to make a call on `viaApp`: it's PR-specific in real usage, since on Issues and Dependabot timelines the app is the primary actor instead.
- Comments, review comments, and threaded comments all need their own pass.
- Event summaries with embedded subcomponents (labels, branch names, commit SHAs) — figured out the right way to model these is per-event, not via a generic templating system in the playground.

packages/react/src/Timeline/Timeline.stories.tsx
```diff
@@ -38,44 +83,421 @@ export const Default = () => (
...
+  'accent',
+  'success',
+  'attention',
+  'severe',
+  'danger',
+  'done',
+  'open',
+  'closed',
+  'sponsors',
+]
```
adierkens: 💅 (can be a followup, since this is intentionally not touching the implementation side) we can reduce some of the duplication if we move this to [Timeline.tsx](https://github.com/primer/react/blob/main/packages/react/src/Timeline/Timeline.tsx#L53) and use it to create the type: 

```tsx
const TimelineBadgeVariants = [
  'accent',
  'success',
  'attention',
  'severe',
  'danger',
  'done',
  'open',
  'closed',
  'sponsors',
] as const

type TimelineBadgeVariant = typeof TimelineBadgeVariants[number]
```

packages/react/src/Timeline/Timeline.stories.tsx
```diff
@@ -38,44 +83,421 @@ export const Default = () => (
...
+  badgeIcon: 'git-commit',
+  summaryText: 'performed an action',
+  eventTimestamp: 'Relative (now)',
+  viaApp: false,
+  appPreset: 'GitHub Actions',
+  customAppName: 'My GitHub App',
+  customAppAvatar: 'https://avatars.githubusercontent.com/in/15368?v=4',
+  showNote: false,
+  noteText: 'Additional context or details',
+  className: '',
```
adierkens: Is `className` needed as an arg in the playground?

packages/react/src/Timeline/Timeline.stories.module.css
```diff
@@ -0,0 +1,53 @@
...
+ * adding a public avatar slot to Primer React's Timeline component (Phase 2 will
+ * evaluate that API change).
+ */
+
+.LeftRailGutter {
+  /* Reserve enough room to the left of the rail for a 40px avatar plus a 16px gap. */
+  padding-left: calc(var(--base-size-40) + var(--base-size-16));
+}
+
+.LargeActorAvatar {
```
adierkens: How many of these are truly story-only since they look like fundamental changes? Should we be representing what things _actually_ render as OOTB and file some follow-up issues to better align the rendering to the expected output rather than overriding locally?

packages/react/src/Timeline/Timeline.stories.tsx
janmaarten-a11y: I wasn't sure about this myself. Removed in https://github.com/primer/react/pull/7836/commits/f8e2e953e6ff498e1dafced47126047f8b2a2755

packages/react/src/Timeline/Timeline.stories.tsx
janmaarten-a11y: Good call! I'm going to keep this for a follow-up and file a tracking issue alongside some of the other Phase 2 cleanup items (like `Octicon` component deprecation) so it doesn't get lost.

packages/react/src/Timeline/Timeline.stories.module.css
janmaarten-a11y: @adierkens 100% agree, and this is the loudest finding of the whole exercise. The fact that we need story-local CSS to fake a left-rail avatar gutter is exactly what tells us the component needs a real avatar slot. Same for the right-controls floats we left out entirely — lots of events have buttons, SHAs, and other content on the right side. The story-local CSS is deliberately scoped to this file to keep the React component untouched in this PR, so the gap stays loud rather than hidden.

That said, they're not strictly *new* — the large-avatar treatment mirrors how the equivalent Rails ViewComponents already render today, it just never formally made it into the React API. One of the reasons I've kept this PR as a draft is that I've been wondering whether we should land those API changes first, so when this merges it's not story-only. It's a real chicken/egg question and I'd be curious what you think.

I noted these in the wrap-up comment and on the parent issue github/primer#6662, but you're right that we should have actual followup issues. Will be filing these to scope for Phase 2:

- Avatar slot on `Timeline.Item` (covers the large-avatar gutter and the Copilot-avatar fallback that the Rails component has already outgrown)
- Right-controls slot
- List semantics evaluation
- `Octicon` cleanup + `BADGE_VARIANTS` dedup from your other comment

janmaarten-a11y: Added:
- https://github.com/github/primer/issues/6677
- https://github.com/github/primer/issues/6678
- https://github.com/github/primer/issues/6679
- https://github.com/github/primer/issues/6680

adierkens (approved):

packages/react/src/Timeline/Timeline.stories.tsx
```diff
@@ -38,44 +83,412 @@ export const Default = () => (
...
+  const resolvedActorName = BAKED_ACTOR_NAMES[args.actorType] ?? customActorName
+  // Anchor "now" to first render so timestamps don't drift as the user toggles controls.
+  const [now] = React.useState(() => Date.now())
+  // Defensive fallback in case Storybook resets `eventTimestamp` to no value ("Choose option")
+  // or restores a stale value from the URL that no longer exists in `TIMESTAMP_PRESETS`.
+  // The `in` check is needed because the typed lookup would otherwise narrow to never-undefined.
+  const timestampPreset =
+    args.eventTimestamp in TIMESTAMP_PRESETS
+      ? TIMESTAMP_PRESETS[args.eventTimestamp]
+      : TIMESTAMP_PRESETS['Relative (now)']
```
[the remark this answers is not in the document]
janmaarten-a11y: Verified there's no runtime throw here — `in` only throws when the right operand isn't an object, and `TIMESTAMP_PRESETS` is a module-level constant. The left operand gets coerced to a string, so `undefined`/empty/stale values all return `false` and fall through to the `'Relative (now)'` default. Confirmed manually in Storybook: selecting "Choose option…" shows "just now" as expected. Keeping as-is.

packages/react/src/Timeline/Timeline.stories.tsx
```diff
@@ -38,44 +83,412 @@ export const Default = () => (
...
+  const timestampDate = new Date(now - timestampPreset.offsetMs)
+  const isCustomApp = args.appPreset === 'Custom App'
+  // Defensive fallback in case Storybook restores a stale `appPreset` from the URL
+  // that no longer exists in `APP_PRESETS` (e.g. after removing a preset like 'Renovate').
+  // The `in` check is needed because the typed lookup would otherwise narrow to never-undefined.
+  const resolvedAppPreset = args.appPreset in APP_PRESETS ? APP_PRESETS[args.appPreset] : APP_PRESETS['GitHub Actions']
```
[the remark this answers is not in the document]
janmaarten-a11y: Same shape as the `eventTimestamp` thread above — `APP_PRESETS` is a const object, `in` is safe, and clearing the control falls back to "GitHub Actions" as intended. Verified in Storybook. Keeping as-is.

packages/react/src/Timeline/Timeline.stories.tsx
```diff
@@ -13,6 +52,12 @@ export default {
     'Timeline.Body': Timeline.Body,
     'Timeline.Break': Timeline.Break,
   },
+  argTypes: {
+    // `clipSidebar` only matters with multiple Timeline.Items. Hide it from the controls
+    // panel on this file's stories (Default and Playground) since both are single-item.
+    // The Features story file demonstrates clipSidebar variants instead.
+    clipSidebar: {table: {disable: true}},
+  },
```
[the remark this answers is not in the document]
janmaarten-a11y: The repo doesn't enable `autodocs` (no `tags: ['autodocs']` on the meta and no Timeline MDX), so there's no autogenerated props table for this component — `table.disable` only affects the Controls panel here, which is the intent (both `Default` and `Playground` are single-item, so `clipSidebar` doesn't apply). If we add a Docs page for Timeline later we can revisit. Keeping as-is.

packages/react/src/Timeline/Timeline.stories.tsx
```diff
@@ -38,44 +83,412 @@ export const Default = () => (
...
+                <>
+                  {' \u2014 with '}
+                  {appAvatar ? <Avatar className={classes.AppAvatar} size={16} square src={appAvatar} alt="" /> : null}
+                  <Link href="#" inline muted className={classes.AppName}>
+                    {appName}
+                  </Link>
+                </>
+              ) : null}
+            </span>
+            {args.showNote && args.noteText ? <div className={classes.Note}>{args.noteText}</div> : null}
```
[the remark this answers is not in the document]
janmaarten-a11y: Addressed in https://github.com/primer/react/pull/7836/commits/668fce5c354db0cbcee3c3a9a109631f95678158

janmaarten-a11y:
Moving to draft, as once #7886 and #7885 are merged, will be updating this PR with the new slots.

adierkens (approved):

PR #7839, opened 2026-05-14
docs/add-ds-spectrum-adr

joshblack:


contributor-docs/adrs/adr-024-design-system-spectrum.md
```diff
@@ -0,0 +1,173 @@
...
+- Examples showing need for customizing appearance
+
+#### SelectPanel
+
+- Examples showing need for behavior changes
+
+## Decision
+
+Primer will provide a spectrum of design system APIs in order to support the different
+ways teams want to build with Primer. These include:
```
lesliecdubs, line 46: Is the proposition here that Primer will provide this API spectrum in all components, or on demand? If on demand, how can we ensure we are able to deliver on this promise for teams who may need access to a lower level quickly?

joshblack: @lesliecdubs what lower level are you thinking about?

I think that we would likely want to do presentational components plus hooks for almost all components with config components built on top (for common use-cases). This would allow for:

- Easy-to-use, out of the box components for common use-cases
- Providing access to the parts of a component to more easily extend or customize (either by us or teams using primer)

The low-level utilities or helpers would likely need to be driven by patterns, techniques, or structures that we see teams doing again and again or that we would like to provide as a foundation for new components. The best example of this are accessibility primitives or key behaviors that are critical or hard to get right.

joshblack:
Just wanted to capture some ways folks might want to extend, or build on top of, a component that came up through conversations:

- Change the markup of a component (for example, I'm using page layout but want the pane to be in a different spot)
- Access some internal of a component (e.g. how to I access the onOutsideClick event for a dialog)
- Add behavior to a component (e.g. add tabs to select panel) or change the underlying state in some way

joshblack:
Notes from engineering working session:

- State is one avenue of control (controlled vs uncontrolled)
  - Resizable scroll bar
  - Initializing state
  - Controlling state
  - Reseting state
- Dialog
  - Memex wanted to open the dialog as a right-panel (drawer?)
- SelectPanel
  - When copilot wanted to use select panels, they had nested dialogs
  - Customize click outside behavior
- Capturing common uncommon scenarios

contributor-docs/adrs/adr-024-design-system-spectrum.md
lesliecdubs: That's helpful. tbh I wasn't too concerned about which lower level; I was more interested in clarifying what we want to offer out of the box vs. what options we would only build in/add on upon request.

PR #7851, opened 2026-05-18
feat: add primer cli exploration

joshblack:
Exploration into a primer cli as an alternative to MCP server. Will be trying this out over in primer/agent-eval 👀


joshblack (commented):
@copilot add support for icons from octicons/octicons-react to primer cli

PR #7852, opened 2026-05-18
Card: address accessibility review feedback

liuliu-dev:
Addresses accessibility-review feedback from Tetralogical (github/accessibility#10648) for the experimental `Card` component. Part of https://github.com/github/primer/issues/6672

#### Default example - metadata content

Default story no longer uses "Updated 2 hours ago" (which would have pulled in `RelativeTime`'s outstanding accessibility issues). Replaced with `<PeopleIcon /> 3 contributors`. The metadata slot is also now documented in `Card.docs.json` and JSDoc to confirm it accepts any content, including other components.

#### Data-component attributes

Added `data-component` attributes per ADR-023 to `Card`, `Card.Icon`, `Card.Image`, `Card.Heading`, `Card.Description`, `Card.Metadata`, and `Card.Menu`.

#### Grouping content for accessibility — new `as` prop

Cards can now opt into being a `<section>` landmark via a new `as` prop. The prop is typed so `as="section"` requires `aria-label` or `aria-labelledby`.

```tsx
// default: renders as <div>, no accessible name required
<Card>…</Card>

// section landmark: enforces an accessible name
<Card as="section" aria-labelledby="my-card-heading">
  <Card.Heading id="my-card-heading">Repo name</Card.Heading>
  …
</Card>
```

#### Interactive content

New `WithMenu` feature story showing an action menu inside a Card with an accessible name on the trigger (`aria-label="More options for primer/react"`).

#### Custom content variant

`CustomContent` story now uses an `<h3>` heading instead of `<strong>`, resolving the WCAG 2.2 SC 1.3.1 violation.

#### Props section

`Card.docs.json` and JSDoc now document `Card.Description`, `Card.Menu`, and `Card.Metadata`.

#### Empty props

`children` is now required on `Card`. Empty cards log a dev warning and return `null`. Tests cover both paths.

#### Inline style attributes on wrapping divs

Added `Card.stories.module.css` with a `WidthConstraintContainer` class instead of`style={{…}}` props in the stories.

### Rollout strategy

- [ ] Patch release
- [X] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Card/Card.tsx
```diff
@@ -20,8 +20,36 @@ export type CardProps = React.ComponentPropsWithoutRef<'div'> & {
...
+
+/**
+ * Renders the Card as a `<section>`. Use this for a standalone Card that
+ * isn't part of a list. The `<section>` is a region landmark, so it needs
+ * an accessible name via `aria-label` or `aria-labelledby` (the latter
+ * usually points at `Card.Heading`).
+ */
+type SectionCardProps = BaseCardProps & {
+  as: 'section'
+} & ({'aria-label': string} | {'aria-labelledby': string})
```
joshblack: I think the TypeScript types for this will end up a little strange, e.g.

```tsx
<Card as="section" onClick={(event) => {}}>...</Card>
```

The `event` here is typed as `MouseEvent<HTMLDivElement>` instead of `MouseEvent<HTMLElement>` (which I think is HTMLElement because HTMLSectionElement doesn't seem to show up even within section directly)

It might help to make the CardProps generic over `As` so that the types themselves derive from it

packages/react/src/Card/Card.tsx
```diff
@@ -20,8 +20,36 @@ export type CardProps = React.ComponentPropsWithoutRef<'div'> & {
...
+
+export type CardProps = DivCardProps | SectionCardProps
```
joshblack:
```suggestion
type CardPropsAs = 'div' | 'section'

export type CardProps<As extends CardPropsAs> = {
  as?: CardPropsAs,
} & React.ComponentPropsWithoutRef<CardPropsAs>
```

Maybe something like this? Generic types are rough for this use-case so let me know if it doesn't work 👀

packages/react/src/Card/Card.tsx
```diff
@@ -33,6 +61,9 @@ type HeadingProps = React.ComponentPropsWithoutRef<'h3'> & {
 }
 
 type DescriptionProps = React.ComponentPropsWithoutRef<'p'> & {
```
joshblack, line 43: Just wanted to confirm, will the description always be a `p` tag?

packages/react/src/Card/Card.tsx
```diff
@@ -100,24 +144,44 @@ const CardImpl = forwardRef<HTMLDivElement, CardProps>(function Card(
...
+  )
+
+  warning(
+    as === 'section' && !('aria-label' in props) && !('aria-labelledby' in props),
+    'The <Card> component used with `as="section"` requires either `aria-label` or `aria-labelledby` so screen-reader users can identify the labelled region. Typically `aria-labelledby` should reference the id of the `Card.Heading`.',
+  )
+
+  if (isEmpty) {
+    return null
+  }
```
joshblack: I think both of these are either already enforced at the type level or could be enforced at the type level, is that right? If so, I don't know if we need to do runtime checks for them. Let me know what you think 👀

packages/react/src/Card/Card.tsx
```diff
@@ -153,17 +218,29 @@ const CardIcon = ({icon: IconComponent, 'aria-label': ariaLabel, className}: Ico
 CardIcon.displayName = 'Card.Icon'
 
 const CardImage = ({src, alt = '', className, ...rest}: ImageProps) => {
-  return <img src={src} alt={alt} className={clsx(classes.CardImage, className)} {...rest} />
+  return (
+    <img src={src} alt={alt} className={clsx(classes.CardImage, className)} data-component="Card.Image" {...rest} />
+  )
 }
 
 CardImage.displayName = 'Card.Image'
```
joshblack, line 212: Super nitpicky, it'd be nice if the components were function declarations so that we didn't have to do this, e.g.

```tsx
function CardImage(/* ... */) {
  // ...
}
```

packages/react/src/Card/Card.tsx
```diff
@@ -60,17 +91,30 @@ type ImageProps = React.ComponentPropsWithoutRef<'img'> & {
...
+   */
   children: React.ReactNode
 }
 
 type MetadataProps = React.ComponentPropsWithoutRef<'div'> & {
+  /**
+   * Metadata row at the bottom of the card. Any content works: text, icons,
+   * a `Label`, an `Octicon`. Skip `RelativeTime` for now; it has open
+   * accessibility issues.
+   */
```
joshblack, line 82:
```suggestion
  /**
   * Metadata row at the bottom of the card. Any content works: text, icons,
   * a `Label`, an `Octicon`.
   */
```

I don't know if we need to include the RelativeElement note here for the docs (although let me know what you think)

packages/react/src/Card/Card.tsx
```diff
@@ -60,17 +91,30 @@ type ImageProps = React.ComponentPropsWithoutRef<'img'> & {
 }
 
 type MenuProps = {
+  /**
+   * Interactive control in the top-right of the card, usually an `IconButton`
+   * or `ActionMenu` trigger. Give the control a label that names the card
+   * (e.g. `"Options for Project Alpha"`, not just `"Options"`) so it's
+   * distinguishable when several cards are on screen.
+   */
```
joshblack: This note was a little confusing at first since the comment seems like it would apply to the underlying ActionMenu? Not `MenuProps` itself. I'm not sure if it's just a wording thing but could help to clarify if others had the same impression

packages/react/src/Card/Card.tsx
```diff
@@ -20,8 +20,36 @@ export type CardProps = React.ComponentPropsWithoutRef<'div'> & {
...
+   * Empty cards do not render.
+   */
+  children: React.ReactNode
 }
 
+/**
+ * The default Card. Use this inside `<li>` when rendering a list of cards:
+ * the list already groups them, so the Card itself doesn't need to be a
+ * landmark. Don't use `Card.Heading` here.
+ */
```
joshblack: Is there ever a case where the Card itself could be a list item or do we always want it in a containing `li` in that case?

packages/react/src/Card/Card.features.stories.tsx
```diff
@@ -1,56 +1,164 @@
...
+        <li>Item three</li>
+      </ul>
+    </div>
+  </Card>
+)
+
+export const StandaloneSection = () => (
+  <Card as="section" aria-labelledby="standalone-card-heading">
+    <Card.Icon icon={RepoIcon} />
+    <Card.Heading id="standalone-card-heading">primer/react</Card.Heading>
```
joshblack: If this use-case is common, one idea could be to offer a separate API/set of components for this path. For example:

- `Card` would be the `li`
- `CardLandmark` would be the `section`

In the case of `CardLandmark` we could automatically wire the `aria-labelledby` and `id` pieces.

If that's not preferred, we could also auto-wire this in the current implementation too (similar to Banner) as another idea

packages/react/src/Card/Card.features.stories.tsx
```diff
@@ -1,56 +1,164 @@
...
+          </ActionMenu.Overlay>
+        </ActionMenu>
+      </Card.Menu>
+    </Card>
+  )
+}
+
+export const CustomContent = () => (
+  <Card>
+    <div className={classes.CustomContentLayout}>
```
joshblack, line 75: From an API perspective on the CSS side, would it be helpful to offer grid areas for this case or not really?

packages/react/src/Card/Card.features.stories.tsx
```diff
@@ -1,56 +1,164 @@
...
+
+export const WithMenu = () => {
+  return (
+    <Card>
+      <Card.Icon icon={RepoIcon} />
+      <Card.Heading>primer/react</Card.Heading>
+      <Card.Description>
+        {"GitHub's design system implemented as React components for building consistent user interfaces."}
+      </Card.Description>
+      <Card.Menu>
```
joshblack: One naming suggestion, would it make sense to have `Card.Menu` be `Card.Actions`? Or is it important for it to always be a `Menu`?

```suggestion
      <Card.Actions>
```

packages/react/src/Card/Card.tsx
```diff
@@ -153,17 +218,29 @@ const CardIcon = ({icon: IconComponent, 'aria-label': ariaLabel, className}: Ico
 CardIcon.displayName = 'Card.Icon'
 
 const CardImage = ({src, alt = '', className, ...rest}: ImageProps) => {
-  return <img src={src} alt={alt} className={clsx(classes.CardImage, className)} {...rest} />
+  return (
+    <img src={src} alt={alt} className={clsx(classes.CardImage, className)} data-component="Card.Image" {...rest} />
```
joshblack:
```suggestion
    <img {...rest} src={src} alt={alt} className={clsx(classes.CardImage, className)} data-component="Card.Image"/>
```

One suggestion just to help with issues that we've run into in the past when we spread at the end. The pattern is to spread first and then merge afterwards if there is overlap

joshblack (commented):
Looking great! Just left a couple of comments/questions, hope they make sense. Let me know if you want to talk about anything 👀

packages/react/src/Card/Card.tsx
liuliu-dev: Done. made `Card` polymorphic using `fixedForwardRef` + `PolymorphicProps`.

packages/react/src/Card/Card.tsx
liuliu-dev: Yes, always a `p`

packages/react/src/Card/Card.tsx
liuliu-dev: Removed the section aria-label warning. 

Kept the empty children warning, seems that types can't prevent `<Card>{null}</Card>` 👀

packages/react/src/Card/Card.tsx
liuliu-dev: Converted to function declarations. 

Should we keep `displayName` for the dotted names? Otherwise it will show as CardImage not Card.Image in devtools

packages/react/src/Card/Card.tsx
liuliu-dev: I'm thinking always in a containing `<li>`, and `Card` stays as a generic container, curious what you think 👀

packages/react/src/Card/Card.features.stories.tsx
liuliu-dev: Added auto-wiring. keeping single component `Card` for now

packages/react/src/Card/Card.features.stories.tsx
liuliu-dev: Good idea! Might be a follow up PR, tracking it here https://github.com/github/primer/issues/6561#issuecomment-4502535751

packages/react/src/Card/Card.features.stories.tsx
liuliu-dev: I think "Actions" might imply buttons rather than a single dropdown trigger. "Menu" is accurate for the current use case but too specific. What do you think about `Card.Control`?

liuliu-dev:
@joshblack thanks for the thorough review, really appreciate it!! I addressed most of them, and some I left questions or marked as features to add later, what do you think?

packages/react/src/Card/Card.tsx
joshblack: @liuliu-dev is this something that we need to warn about? (having no children).

packages/react/src/Card/Card.tsx
joshblack: I'm neutral about it, if I had to pick I'm happy with just the component name itself versus adding the dots in personally but I get if that's confusing

packages/react/src/Card/Card.features.stories.tsx
joshblack: I think if the idea is that this will always be a Menu then that makes sense 👍 If it might grow in the future it might make sense to make it more generic. For cards, do you think we're in the more specific case? 👀

joshblack (approved):
Approving in advance ✅ Left a couple of questions, only big thing is that I don't think we need to warn on empty children but that's definitely a personal preference.

packages/react/src/Card/Card.tsx
liuliu-dev: Nope, removed it

packages/react/src/Card/Card.features.stories.tsx
liuliu-dev: That makes sense, renamed it to `Card.Action`.

PR #7856, opened 2026-05-19
DataTable: add integrated pagination via the pagination prop

ianwinsemius:
Closes #

Today, paginating a `DataTable` requires consumers to:

1. Track a `pageIndex` state manually.
2. Slice the data array themselves.
3. Pass the slice to `DataTable`.
4. Render `<Table.Pagination>` as a sibling and wire its `onChange` back to step 1.

This PR adds first-class integrated pagination so the common case becomes:

```tsx
<DataTable
  data={repos}
  columns={columns}
  pagination={{pageSize: 10}}
/>
```

The existing manual composition path (`<Table.Pagination />` as a sibling) continues to work for callers that want finer control, parallel to how `externalSorting` coexists with built-in sort.

### Implementation notes

- `<Pagination>` owns its own UI state. `<DataTable>` mirrors its `onChange` to slice rows.
- The remount-key strategy avoids a setState-in-render feedback loop where Pagination's render-time `defaultPageIndex` sync would call back into DataTable's setState.
- Auto-resets to page 0 when the `data` identity changes (uncontrolled mode only) so consumers refetching from a server can't land on a phantom page.

#### New

- `DataTableProps.pagination` — `true`, an options object (`pageSize`, `defaultPageIndex`, `aria-label`, `showPages`), or `false` to opt out.
- `DataTableProps.pageIndex` / `onPageChange` — controlled-mode page state.
- `DataTableProps.externalPagination` — defer slicing to the server.
- Stories: `WithIntegratedPagination`, `WithIntegratedPaginationControlled`, `WithIntegratedPaginationExternal`.

#### Changed

- `DataTable.docs.json` updated with new prop docs and story IDs.

#### Removed

(none)

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- 12 new unit tests in `packages/react/src/DataTable/__tests__/integrated-pagination.test.tsx` cover opt-in, row slicing, `defaultPageIndex`, empty-data handling, data-identity reset, controlled mode (parent owns `pageIndex`), `externalPagination`, and pagination + sorting composition.
- Existing 117 DataTable tests pass unchanged — the refactor is fully backward-compatible.
- `npm run build`, `npm run type-check`, `npm run lint`, `npm run lint:css`, and `npm test -- --run packages/react/src/DataTable/` all pass locally.
- Stories available under **Experimental / Components / DataTable / Features / With Integrated Pagination** (and the two adjacent ones).

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


ianwinsemius:
> [!NOTE]
> The failing `recommend` check on this PR is the well-known fork-permission bug: the `Recommend integration tests` workflow tries to write a label to the PR, but `GITHUB_TOKEN` from a fork has read-only access to `primer/react`. I've opened **#7861** to fix that workflow (wraps the writes in `try/catch` so the job stays green for fork PRs while still adding the label for internal branches). Once #7861 merges, an empty commit here will clear the red check.

packages/react/src/DataTable/DataTable.tsx
```diff
@@ -73,6 +75,46 @@ export type DataTableProps<Data extends UniqueRow> = {
...
+   */
+  pagination?:
+    | false
+    | true
+    | {
+        pageSize?: number
+        defaultPageIndex?: number
+        'aria-label'?: string
+        showPages?: boolean | ResponsiveValue<boolean>
+      }
```
joshblack: ~~What would you think about having these each be their own prop so that way we don't need the `pagination` prop? The table could then be in pagination mode if these are set (and provide sensible fallbacks)~~

Ah I guess this might not make sense because if they want pagination with default behavior they still need a prop to indicate that 😅

joshblack (commented):
Thanks for taking the time to put this together! Just had a question on the `pagination` type real quick 👀

lesliecdubs:
@joshblack thanks for reviewing. What do we need to do to get this ready to merge? It's been sitting for some time now.

PR #7857, opened 2026-05-19
DataTable: add Table.CopyAsMarkdownButton with hardened markdown serializer

ianwinsemius:
Closes #

Adds an opt-in slot button for copying the visible rows of a `DataTable` to the system clipboard as a GitHub-flavoured Markdown table.

```tsx
<Table.Container>
  <Table.Title id="repositories">Repositories</Table.Title>
  <Table.Actions>
    <Table.CopyAsMarkdownButton rows={repos} columns={columns} />
  </Table.Actions>
  <DataTable data={repos} columns={columns} />
</Table.Container>
```

## Why review this PR carefully

A markdown-export feature can become a vulnerability surface if cell content can break out of a table cell, smuggle in HTML, or escape into the surrounding markdown. The implementation here is intentionally conservative — please scrutinise the escape rules and security model.

## Security model (for review)

1. **Plain-text only.** The clipboard payload is always `text/plain`. There is no HTML and no inline-HTML escape hatch (e.g. no `<br>`).
2. **No React/JSX in the export.** `renderCell` is **never** called by the serializer. Cell projection order is:
   1. `column.getExportValue(row)` — explicit, type-safe, audited per-column
   2. The value at `column.field` (dotted path traversal)
   3. Empty string
   A spy-based test verifies `renderCell` is not invoked.
3. **Escaping rules** in `escapeMarkdownCell`:
   - `\` → `\\` (done first so escaped pipes survive)
   - `|` → `\|`
   - CR / LF / CRLF → single space
   - ASCII control characters `\u0000–\u001f` and Unicode C1 `\u007f–\u009f` → stripped (includes TAB, DEL)
   - Leading/trailing whitespace trimmed
4. **What is deliberately NOT escaped:** backticks, brackets, asterisks. These cannot break the table layout and may be legitimately styled by downstream Markdown renderers. The contract is "cannot break out of the cell" — not "renders as plain text".
5. **Clipboard write** prefers `navigator.clipboard.writeText` and falls back to a hidden-textarea `document.execCommand('copy')` for non-secure contexts and browsers that reject the async API.

### Hostile-input test

There's a test case that feeds the serializer an explicit table-breakout attempt:

```
| evil |
| --- |
| `cmd` | [link](javascript:alert(1)) | **bold** |
```

and asserts:
- No unescaped pipes remain.
- No newlines remain.
- The cell content stays *inside* the cell.

#### New

- `Table.CopyAsMarkdownButton` slot component (designed for `Table.Actions`).
- `Column.getExportValue?: (data: Data) => string` — opt-in plain-text projection per column.
- Helpers exported from the DataTable barrel: `escapeMarkdownCell`, `rowsToMarkdown`, `writeTextToClipboard`.

#### Changed

- `DataTable.docs.json` updated with `Table.CopyAsMarkdownButton` and `Column.getExportValue` docs.

#### Removed

(none)

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- 33 unit tests in `packages/react/src/DataTable/__tests__/clipboard.test.tsx` cover escape rules, hostile input, projection precedence, `renderCell` exclusion (spy-verified), array/object stringification, dotted field paths, well-formed table shape, empty data, header pipe escaping, callable `header` fallback, both clipboard write paths, and the component's transient label / `onCopy` callback / custom children.
- 117 existing DataTable tests pass unchanged.
- `npm run build`, `npm run type-check`, `npm run lint`, `npm run lint:css`, and `npm test -- --run packages/react/src/DataTable/` all pass locally.
- Story available under **Experimental / Components / DataTable / Features / With Copy As Markdown**.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/datatable-copy-as-markdown.md
+++ b/.changeset/datatable-copy-as-markdown.md
@@ -1,0 +1,13 @@
+---
+'@primer/react': minor
+---
+
+DataTable: Add `Table.CopyAsMarkdownButton` for copying visible rows as a
+GitHub-flavoured Markdown table. The button is a slot designed for
+composition inside `<Table.Actions>`. Cell values are projected via a new
+optional `Column.getExportValue` (preferred) or the field value — never
+`renderCell`, so React/JSX output stays out of the plain-text payload.
+Pipes, newlines, and ASCII/Unicode control characters are escaped or
+stripped before reaching the clipboard. The implementation prefers the
+async Clipboard API and falls back to `document.execCommand('copy')` for
+non-secure contexts.
--- a/packages/react/src/DataTable/CopyAsMarkdownButton.module.css
+++ b/packages/react/src/DataTable/CopyAsMarkdownButton.module.css
@@ -1,0 +1,6 @@
+/* Reserve a stable layout while the transient copied/error label is shown
+ * so the button doesn't reflow under the cursor. */
+.CopyAsMarkdownButton {
+  min-inline-size: 9rem;
+  justify-content: center;
+}
--- a/packages/react/src/DataTable/CopyAsMarkdownButton.tsx
+++ b/packages/react/src/DataTable/CopyAsMarkdownButton.tsx
@@ -5,14 +5,25 @@ import {Button} from '../Button'
 import type {Column} from './column'
 import type {UniqueRow} from './row'
 import {rowsToMarkdown, writeTextToClipboard} from './clipboard'
+import {useDataTableSnapshot} from './snapshotContext'
 import classes from './CopyAsMarkdownButton.module.css'
 
 export type CopyAsMarkdownButtonProps<Data extends UniqueRow> = {
-  /** Rows to serialise. Usually the same array passed to `<DataTable>`. */
-  rows: ReadonlyArray<Data>
+  /**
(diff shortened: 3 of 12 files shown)


ianwinsemius:
> [!NOTE]
> The failing `recommend` check on this PR is the well-known fork-permission bug: the `Recommend integration tests` workflow tries to write a label to the PR, but `GITHUB_TOKEN` from a fork has read-only access to `primer/react`. I've opened **#7861** to fix that workflow (wraps the writes in `try/catch` so the job stays green for fork PRs while still adding the label for internal branches). Once #7861 merges, an empty commit here will clear the red check.

joshblack:
Hi there @ianwinsemius! :wave:

For these types of feature additions it'd be helpful to collaborate on an issue first, either through https://github.com/primer/react/issues or https://github.com/github/primer/issues to align before implementation begins.

I think this use-case might be something that would work well either local in a project or github-ui first and then upstreamed as the use-case became more broadly used. Let me know what you're thinking though!

PR #7860, opened 2026-05-20
ci: update client-id to app-id in status-checks.yml

joshblack:
It seems like this input has been deprecated in favor of `client-id` so this PR makes that change 👀


joshblack:
Ah, seems like `client-id` is a different format so will look into that 😅

jonrohan (approved):

llastflowers (approved):

PR #7855, opened 2026-05-19
DataTable: add per-column filtering via filterBy

ianwinsemius:
Closes #

Adds first-class per-column filtering to the experimental `DataTable`. Columns opt in with a new `filterBy` prop that mirrors the shape of `sortBy`, and the table renders an inline filter row beneath the column headers when at least one column is filterable.

```tsx
<DataTable
  data={repos}
  columns={[
    {header: 'Repository', field: 'name', filterBy: true, sortBy: 'alphanumeric'},
    {header: 'Type', field: 'type', filterBy: 'startsWith'},
    {header: 'Updated', field: 'updatedAt'},
  ]}
  filterable
/>
```

Goals:
- Match `sortBy` ergonomics (`true` | named strategy | custom function) so the column API feels native.
- Cover both controlled (`filters` / `onFilterChange`) and uncontrolled (`defaultFilters`) modes.
- Provide an `externalFiltering` escape hatch that parallels the existing `externalSorting` prop.
- Expose `Table.FilterRow` / `Table.FilterCellInput` primitives for consumers composing their own header.

#### New

- `Column.filterBy?: boolean | 'substring' | 'startsWith' | CustomFilterStrategy<Data>` — opts a column into filtering.
- `DataTableProps.filterable` — renders the per-column filter row.
- `DataTableProps.filters` / `defaultFilters` / `onFilterChange` — controlled and uncontrolled filter state.
- `DataTableProps.externalFiltering` — defer filtering to the server.
- `DataTableProps.filterPlaceholder` — customize the input placeholder.
- `Table.FilterRow` / `Table.FilterCellInput` — public primitives.
- `FilterStrategy` / `CustomFilterStrategy` type exports.
- Stories: `WithFiltering`, `WithControlledFilters`, `WithCustomFilter`.

#### Changed

- `useTable` now derives filtered rows on top of the sorted row order; no impact on existing consumers.
- `DataTable.docs.json` updated with new prop docs and story IDs.

#### Removed

(none)

### Rollout strategy

- [ ] Patch release
- [x] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Testing & Reviewing

[...]

PR #7861, opened 2026-05-20
ci: make Recommend integration tests workflow fork-safe

ianwinsemius:
Closes #

## Problem

The `Recommend integration tests` workflow runs on every `pull_request` event, including PRs opened from forks. When triggered by a fork PR the provided `GITHUB_TOKEN` has read-only access to the upstream repo and any write call (`addLabels` / `createComment`) fails with:

```
HttpError: Resource not accessible by integration (HTTP 403)
```

That failure surfaces as a red ❌ `recommend` check on every fork PR — even when the underlying source code is fine and all other checks (lint, tests, type-check, build, VRT, AAT, CodeQL) pass. It's misleading because the job's role is to *suggest* integration testing, not to gate the PR on its own infra success.

Example: see the three PRs I have open (#7855, #7856, #7857) — all 34 functional checks pass, and only the `recommend` workflow fails because of this permission limitation.

## Fix

Wrap the label and comment API calls in a small helper that catches HTTP 403s and logs an actionable `core.info` message instead of throwing. The job runs to completion green, and maintainers can apply the `integration-tests: recommended` label manually if the change warrants integration testing.

Additionally, comment writes are skipped entirely on fork PRs (`context.payload.pull_request.head.repo.full_name !== context.payload.repository.full_name`). The bot's comment recommends running an internal-only workflow that a fork PR author cannot trigger anyway — better to suppress the noise.

## Before / After

Before:
- Fork PR opens → `recommend` job fails with `Resource not accessible by integration` → red ❌ on the PR.

After:
- Fork PR opens → `recommend` job catches the 403, logs `Skipped addLabels: GITHUB_TOKEN cannot write to this repository from a forked PR. A maintainer can apply the 'integration-tests: recommended' label manually if integration tests are warranted.` → green ✅ on the PR.
- Internal-branch PR behaviour unchanged: the job still adds the label and posts the comment.

#### New
- (none)

#### Changed
- `.github/workflows/recommend-integration-tests.yml` — wrap label/comment writes in `try/catch` keyed on HTTP 403 so the job no longer fails on fork PRs.

#### Removed
- (none)

## Rollout strategy

- [x] Patch release (CI-only change, no runtime impact)
- [ ] Minor release
- [ ] Major release
- [ ] None

> This is a CI-only change with no impact on the published `@primer/react` package, so it does not include a changeset. Please apply the `skip changeset` label.

## Testing & Reviewing

- YAML validated locally (`python3 -c "import yaml; yaml.safe_load(...)"`).
- Logic is small enough to reason about by inspection — the only added complexity is a `softWrite(operation, fn)` helper that distinguishes 403 from other errors.
- Will be exercised in CI as soon as this PR opens (it's itself a fork PR, so it should self-validate that the new code path is green).

## Merge checklist

- [x] Added/updated tests *(N/A — workflow file change; behaviour will be exercised by every PR opened against this repo, including this one)*
- [x] Added/updated documentation *(inline comments explaining the soft-fail rationale)*
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility) *(N/A — CI workflow change)*
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.github/workflows/recommend-integration-tests.yml
+++ b/.github/workflows/recommend-integration-tests.yml
@@ -49,23 +49,51 @@ jobs:
               repo: context.repo.repo
             }
 
+            // When this workflow runs for a PR opened from a fork, the
+            // GITHUB_TOKEN has read-only access to the upstream repo and
+            // any write call (addLabels / createComment) returns
+            // `Resource not accessible by integration` (HTTP 403). That
+            // is the right security boundary, but it causes this check
+            // to fail on every fork PR which is misleading: the failure
+            // is not a problem with the PR's source code.
+            //
+            // Soft-fail those writes so the check stays green for fork
+            // PRs. Maintainers can still apply the label manually if
+            // they want to recommend integration tests for the change.
+            const isFork = context.payload.pull_request?.head?.repo?.full_name !== context.payload.repository?.full_name
+            const softWrite = async (operation, fn) => {
+              try {
+                await fn()
+              } catch (error) {
+                if (error.status === 403) {
+                  core.info(`Skipped ${operation}: GITHUB_TOKEN cannot write to this repository from a forked PR. A maintainer can apply the '${INTEGRATION_LABEL_NAMES.recommended}' label manually if integration tests are warranted.`)
+                  return
+                }
+                throw error
+              }
+            }
+
             const labels = await github.paginate(github.rest.issues.listLabelsOnIssue, issue);
             const integrationLabels = labels.filter(label => label.name.startsWith('integration-tests'))
             const hasPassingLabel = integrationLabels.find(label => label.name === INTEGRATION_LABEL_NAMES.passing)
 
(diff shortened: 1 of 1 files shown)


joshblack:
Hi there @ianwinsemius! 👋 Thanks for taking the time to make this PR 🙏 

I opened up an alternative idea for this over at: https://github.com/primer/react/pull/7862, would this work for you, as well? This should skip the job entirely at the workflow level so you shouldn't run into the case where a Pull Request from a fork has a red status check for this workflow.

Let me know what you think!

joshblack:
Merged in: https://github.com/primer/react/pull/7862 if that works for you @ianwinsemius! Closing this out but feel free to reach out if this isn't addressing the underlying need here 👀

PR #7863, opened 2026-05-21
Fix SelectPanel outside-top anchor sizing regression

siddharthkp:
Closes github/issues#18331

Note: We may not need this at all if css anchor positioning solves this. I tested my bug repro and does seem like it.

### Overview

When `anchorSide` is `'outside-top'` and `pinPosition` is true, the SelectPanel overlay was preserving a stale inline height even when inner content needed to grow after loading. This happened because overflow was trapped in a descendant scroll container rather than on the overlay itself.

The fix adds a `hasOverflowingDescendant` helper to `useAnchoredPosition` that detects overflow in descendants before preserving height, and clears stale inline height when content actually needs more space.

#### New

- Added `AutogrowAfterLoadingWithOutsideTopAnchor` story to reproduce the regression

#### Changed

- `useAnchoredPosition` now checks descendant overflow before preserving height when `pinPosition` is true

#### Removed

_None_

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

1. Open the `AutogrowAfterLoadingWithOutsideTopAnchor` story
2. Click the SelectPanel trigger button
3. Wait for loading to complete (~2s)
4. **Before fix**: Panel stays small with scrollbar
5. **After fix**: Panel grows to fit content, no scrollbar

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/SelectPanel/SelectPanel.examples.stories.tsx
+++ b/packages/react/src/SelectPanel/SelectPanel.examples.stories.tsx
@@ -373,6 +373,7 @@ export const RepositionAfterLoading = () => {
 
   React.useEffect(() => {
     if (!loading) {
+      // eslint-disable-next-line react-hooks/set-state-in-effect -- Updating filtered items based on filter change
       setFilteredItems(items.filter(item => item.text.toLowerCase().startsWith(filter.toLowerCase())))
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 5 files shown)


packages/react/src/SelectPanel/SelectPanel.examples.stories.tsx
```diff
@@ -438,14 +434,62 @@ export const SelectPanelRepositionInsideDialog = () => {
           selected={selected}
           onSelectedChange={setSelected}
           onFilterChange={setFilter}
-          overlayProps={{anchorSide: 'outside-top'}}
```
siddharthkp, line 441: Unrelated change, but good to have here. The reposition story wasn't repositioning at all because we were always opening it above. Fixed the story, it works well.

PR #7869, opened 2026-05-22
feat: improve slot system consistency

adierkens:
Closes #

Improves consistency across the slot system (`__SLOT__` markers + `useSlots` / `isSlot`). Three buckets of work:

1. **Cleanup**: remove orphan root markers, standardise `Symbol(...)` descriptions, migrate hand-rolled `Children.toArray + isSlot` consumers to `useSlots` where the pattern fits.
2. **Public API**: export the slot primitives that downstream consumers currently hand-roll, plus a new `asSlot` helper that replaces the cast-heavy marker-copy pattern.
3. **Documentation**: a contributor skill at `.github/skills/slots/SKILL.md` that describes when and how to use slots, with naming conventions and pitfalls.

#### New

- Public exports from `@primer/react`: `useSlots`, `isSlot`, `asSlot`, `WithSlotMarker`, `FCWithSlotMarker` (alongside the existing `SlotMarker` type). `useSlots` is also exported from `@primer/react/hooks`.
- `asSlot(component, slotSource)` helper that copies a `__SLOT__` marker from a source slot component onto a wrapper. Typed; dev-warns when the source has no marker. Replaces the cast-heavy `(Wrapper as typeof Wrapper & {__SLOT__?: symbol}).__SLOT__ = Source.__SLOT__` pattern.
- Dev-mode warning in `useSlots` when a child's `displayName` matches a slot component's `displayName` but the child is missing the `__SLOT__` marker (the most common wrapping footgun).
- `.github/skills/slots/SKILL.md` contributor skill documenting when to use slots, naming conventions, public APIs, limitations, and common pitfalls. Referenced from `.github/copilot-instructions.md`.

#### Changed

- Standardised `Symbol(...)` descriptions used as slot markers to the `Parent.Slot` convention: `CheckboxOrRadioGroup.Label`, `CheckboxOrRadioGroup.Caption`, `CheckboxOrRadioGroup.Validation`, `Tooltip` (was `DEPRECATED_Tooltip`), `DataTable.Table` (was `Table`).
- Migrated `PageHeader`, `NavList.Item`, and the internal `CheckboxOrRadioGroup` to use `useSlots` instead of hand-rolled `React.Children` traversals. The `CheckboxOrRadioGroup` migration also removes duplicated work where `useSlots` was already being called but slots were re-extracted by hand immediately after.

#### Removed

- Orphan `__SLOT__` markers from 7 root components with no internal consumers: `ActionMenu` (root `Menu`), `UnderlinePanels`, `PageLayout`, `SegmentedControl`, `RadioGroup`, `CheckboxGroup`, and `Dialog`. Sub-component markers are intentionally retained so consumers can keep wrapping them.

### What was evaluated but intentionally not migrated

Four hand-rolled consumers use patterns that don't fit `useSlots`'s single-match-per-key model:

- **`CheckboxGroup`** and **`UnderlinePanels`** — multi-match (find *all* children of a given type). Would migrate cleanly if `useSlots` grew a multi-match option; design TBD in a follow-up.
- **`SegmentedControl`** — per-child classification with index tracking (selection state). Not a slot extraction pattern.
- **`ActionMenu`** — `Children.map` with side effects, including grandchild introspection for Tooltip-wrapped Anchors. Not a slot extraction pattern.

`TreeView`'s `useSubTree` is also a clean single-match consumer but lives inside a `useMemo`; migrating tripped `react-hooks/rules-of-hooks` (the lint rule enforces the `use*` naming convention even though `useSlots` is a plain function under the hood). Left as-is. The slots skill calls this pitfall out explicitly.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

New public exports + new helper, no breaking changes. The removed root `__SLOT__` markers had no consumers anywhere in the codebase (verified by grep) so removing them is non-breaking; if a downstream consumer was secretly wrapping against e.g. `Dialog.__SLOT__`, they would now lose the slot match — but Dialog is a root and nothing scanned for it, so there was nothing to be matched into.

### Testing & Reviewing

- Type-check, ESLint, and `prettier --check` all clean on changed files.
- 478 browser-mode tests passed across the affected components (utils, hooks, PageHeader, NavList, Dialog, CheckboxGroup, RadioGroup, SegmentedControl, ActionMenu, UnderlinePanels, Tooltip, FormControl, Autocomplete, TreeView, internal CheckboxOrRadioGroup).
- The node-mode `exports.test.ts` snapshot was updated to include the new public exports.
- PageLayout snapshot mismatches exist on the unchanged baseline (CSS hash drift) and are not caused by this PR — easy to verify by stashing and re-running.

Areas worth a close look during review:

- `packages/react/src/hooks/useSlots.ts` — the new `warnIfDisplayNameMatchesWithoutMarker` helper is dev-only but runs once per non-matching child; happy to gate it behind a feature flag if perf is a concern.
- `packages/react/src/PageHeader/PageHeader.tsx` — `useSlots` is now called at the top level even though the only consumer is a dev-only effect. Couldn't keep it inside the effect because of `react-hooks/rules-of-hooks`. PageHeader isn't render-hot so this should be fine, but flagging.
- `.github/skills/slots/SKILL.md` — convention statements; let me know if anything contradicts your intent.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/DataTable/index.ts
```diff
@@ -51,7 +51,7 @@ const Table: typeof TableImpl &
   ErrorDialog,
 })
 
-Table.__SLOT__ = Symbol('Table')
+Table.__SLOT__ = Symbol('DataTable.Table')
```
joshblack:
```suggestion
```

I think we could remove this slot? Not sure why it has one in the first place

joshblack (approved):
Just wanted to ✅, overall I think it's great to update the existing slot patterns and provide tools for them downstream for people to use.

PR #6452, opened 2025-08-04
refactor(project): remove support for jest

joshblack:
Closes https://github.com/github/primer/issues/5336

Removes Jest from the project in favor of Vitest now that our migration is complete 🫡 

#### Changed

- Update vitest to be the default test runner
- Update references to Jest to instead point to Vitest

#### Removed

- Remove support for Jest from the project
- Remove vitest migration from our migration tracking workflow

PR #7872, opened 2026-05-22
fix(test-helpers): make @primer/react/test-helpers runtime-agnostic

adierkens:
Closes #

Makes the published `@primer/react/test-helpers` entry runtime-agnostic so it works with both Jest and Vitest consumers.

Today the file hard-references the `jest` global to create mock functions for its JSDOM polyfills (`ResizeObserver`, `HTMLDialogElement.showModal/close`, `HTMLCanvasElement.getContext`, `Element.scrollIntoView`, `matchMedia`, `CSS.escape/supports`). Any downstream consumer that imports `@primer/react/test-helpers` from a Vitest test currently hits `ReferenceError: jest is not defined`. Primer itself migrated off Jest in #6452 but kept this published helper as Jest-only, so Vitest consumers couldn't use it.

#### New

- Nothing added.

#### Changed

- `packages/react/src/utils/test-helpers.tsx` — replaces every bare `jest.fn(...)` call with a small `mockFn(...)` helper that detects `globalThis.jest?.fn` or `globalThis.vi?.fn` at runtime and falls back to a plain no-op function if neither is present. Mock-style introspection (`expect(fn).toHaveBeenCalled()`, etc.) continues to work for both Jest and Vitest consumers; consumers without a test runtime get a silent no-op which is sufficient for the polyfill use case.

#### Removed

- `packages/react/src/SelectPanel/SelectPanel.test.tsx` — removes a stale `// jest function` comment (the test below actually uses `vi.fn()`).

### Rollout strategy

- [x] Patch release
- [ ] Minor release
- [ ] Major release
- [ ] None

No public API surface changes. Existing Jest consumers continue to get real `jest.fn()` instances (detected via `globalThis.jest.fn`). New Vitest consumers get real `vi.fn()` instances. Non-test environments get no-ops, which is fine for the JSDOM polyfill use case.

### Testing & Reviewing

- ✅ `npx tsc -p packages/react/tsconfig.json --noEmit` clean.
- ✅ `npx prettier --check` + `npx eslint` on changed files clean.
- ✅ `SelectPanel.test.tsx` runs cleanly (152 tests pass) — confirms the stale-comment removal didn't affect anything.
- The helper is `@ts-nocheck` (it's a globals/polyfill setup file), so types around the runtime-detect pattern aren't enforced, intentionally.

Worth a close look:

- The runtime-detect shape itself — there are a couple of reasonable variations:
  - `globalThis.jest?.fn ?? globalThis.vi?.fn` (current — Jest wins if both present, which matches the historical behaviour for Jest consumers).
  - Reverse order if we'd prefer Vitest to win when both are accidentally loaded (e.g. some pnpm hoisting edge cases).
- The fallback no-op for "no runtime" — it returns `() => {}`. Consumers without a test runtime importing this helper are an edge case (it's a test polyfill), but the no-op keeps the polyfill side-effects working.

### Merge checklist

- [x] Added/updated tests
- [x] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility) — the `typeof document !== 'undefined'` guard remains in place
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/test-helpers-runtime-agnostic.md
+++ b/.changeset/test-helpers-runtime-agnostic.md
@@ -1,0 +1,11 @@
+---
+'@primer/react': patch
+---
+
+Make the published `@primer/react/test-helpers` entry runtime-agnostic so it works with both Jest and Vitest consumers.
+
+Previously the file hard-referenced the `jest` global to create mock functions for its JSDOM polyfills (`ResizeObserver`, `HTMLDialogElement.showModal/close`, `HTMLCanvasElement.getContext`, `Element.scrollIntoView`, `matchMedia`, `CSS.escape/supports`). Importing the helper from a Vitest test threw `ReferenceError: jest is not defined`.
+
+The polyfills now detect `globalThis.jest?.fn` or `globalThis.vi?.fn` at runtime and fall back to a plain no-op function if neither is present. Mock-style introspection (`toHaveBeenCalled` etc.) still works for both Jest and Vitest consumers; consumers without a test runtime get a silent no-op which is sufficient for the polyfill use case.
+
+Also removes a stale `// jest function` comment from `SelectPanel.test.tsx` (the test actually uses `vi.fn()`).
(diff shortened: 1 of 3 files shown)


joshblack (commented):
I'd be down to leave this one jest-specific since for vitest we'd be using browser mode, right? Or is that not the case? Ideally we wouldn't be shipping this file, I think.

PR #7648, opened 2026-03-10
Replace dynamic overflow with horizontal scroll in `UnderlineNav`

iansan5653:
Alternative approach to https://github.com/primer/react/pull/7506; just get rid of the 'More' menu and scroll instead. 

- Closes https://github.com/github/primer/issues/6291

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [ ] Major release; if selected, include a written rollout or migration plan
- [ ] None; if selected, include a brief description as to why

### Merge checklist

- [ ] Added/updated tests
- [ ] Added/updated documentation
- [ ] Added/updated previews (Storybook)
- [ ] Changes are [SSR compatible](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#ssr-compatibility)
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge
- [ ] (GitHub staff only) Integration tests pass at github/github-ui ([Learn more about how to run integration tests](https://github.com/github/primer-engineering/blob/main/how-we-work/testing-primer-react-pr-at-dotcom.md))

PR #7875, opened 2026-05-22
perf(UnderlineNav): eliminate cascading render on mount

mattcosta7:
Closes #

## Summary

Re-architects `UnderlineNav` to eliminate the cascading render that profiles
showed on mount. The previous implementation reported every item's measured
width back up to the parent through context setter callbacks, so an N-item
nav produced **O(N²) item renders** during initial mount plus a render-phase
setState for the external-`aria-current` swap. After this PR, mount produces
**at most 3 component-level renders** total, irrespective of N, and per-item
work is O(1) at every step.

All behavior is preserved. 28 tests pass — 19 pre-existing + 9 new that pin
the architecture, structural, and performance contracts.

> [!NOTE]
> This PR is independent of the dev-only-effect dual-build work in #7874.
> The two `__DEV__` invariants in `UnderlineNav` continue to use the inline
> `if (__DEV__) { useEffect(...) }` + `eslint-disable react-hooks/rules-of-hooks`
> pattern here; #7874 will migrate them once it lands.

## What was wrong

A profile of `UnderlineNav` showed a long cascading render chain on mount.
Tracing it back, the cause was the measurement plumbing.

Each `UnderlineNav.Item` used to dispatch **two** parent setStates in a
mount-time `useLayoutEffect`:

```tsx
// UnderlineNavItem
useLayoutEffect(() => {
  const domRect = ref.current.getBoundingClientRect()
  // …
  setChildrenWidth({text, width: domRect.width})           // -> parent setState
  setNoIconChildrenWidth({text, width: domRect.width - i}) // -> parent setState
}, [ref, setChildrenWidth, setNoIconChildrenWidth])
```

Compounded with these problems:

1. **The context value was a fresh object literal every render**
   (`{setChildrenWidth, setNoIconChildrenWidth, loadingCounters, iconsVisible}`),
   so every parent re-render forced every item to re-render.
2. **The `aria-current` swap-from-menu logic ran during render** via a
   `setState` inside `menuItems.map(...)`. React 18 tolerates this by
   discarding the in-progress render, but it added an extra commit every
   time the menu changed.
3. **Widths were stored as `Array<{text, width}>`** and looked up by item
   text. Lookups were O(N) `.find()` scans, and items whose `children`
   weren't a plain string silently looked up as `0`, breaking the swap math.
4. **Item prop refresh was O(N²)**: `responsiveProps.items.map(item =>
   validChildren.find(c => c.key === item.key))` ran inside two `.map()`
   loops every render.
5. **Refs were read during render** to compute the overflow menu's anchor —
   unsafe under concurrent rendering and produced a perceptible reflow on
   the first open.

## How this PR fixes it

Eleven focused commits — each one independently shippable, each one keeping
all tests green — that re-architect the measurement and overflow pipeline.

#### [`perf(UnderlineNav): measure items once from the parent`](https://github.com/primer/react/pull/<id>/commits/181c6bfc6)

Inverts measurement entirely. `UnderlineNavItem` is now a leaf — it tags its
own `<li>` with `data-underlinenav-item="true"` and is done. `UnderlineNav`
itself walks all rendered items in a single `useIsomorphicLayoutEffect` via
that selector, reads their geometry, and stores the results in **refs** (not
state, because only `computeOverflow` reads them and they shouldn't
re-render).

The old context callbacks (`setChildrenWidth`,
`setNoIconChildrenWidth`) are gone from `UnderlineNavContext`.

#### [`perf(UnderlineNav): consolidate overflow state into a single object`](https://github.com/primer/react/pull/<id>/commits/f75c575fc)

`items`, `menuItems`, `iconsVisible`, `measured` (and later `childrenKey`)
collapse into a single `useState` object so every overflow update —
measurement, swap, resize, children-change reset — is exactly **one
commit**, never three.

#### [`perf(UnderlineNav): memoize Item and stabilize context value`](https://github.com/primer/react/pull/<id>/commits/4e3e8ee0a)

Wraps `UnderlineNavItem` in `React.memo` and memoizes `UnderlineNavContext`'s
value with `useMemo([loadingCounters, iconsVisible])`. Re-renders of
`UnderlineNav` no longer cascade through items when their props haven't
changed.

#### [`perf(UnderlineNav): lift aria-current swap out of render phase`](https://github.com/primer/react/pull/<id>/commits/508073a16)

Moves the external-`aria-current` promotion (when a consumer changes
`aria-current` to a menu item, pull it into the visible list) out of
render-phase `setState` and into a `useIsomorphicLayoutEffect`. Render is
pure again; the effect commits only when a swap is actually needed.

#### [`perf(UnderlineNav): collapse nested updates into one effect, one commit`](https://github.com/primer/react/pull/<id>/commits/4effbca3e)

Makes `overflowEffect` pure (renames to `computeOverflow`) and extracts a
pure `computeSwap` helper. Merges the measurement and aria-current effects
into a single combined layout effect that produces at most **one
`setOverflowState`** per pass. `trackedChildrenKey` folds into
`overflowState` as `childrenKey`, so the derived-state reset on children
change is one setState instead of two.

#### [`perf(UnderlineNav): key width and child lookups by React element key`](https://github.com/primer/react/pull/<id>/commits/fca6d119e)

Replaces `Array<{text, width}>` + `.find()` with `Map<elementKey, number>`.
Fixes the silent-zero bug for items with non-string children (e.g.
`<span>Label</span>`, `<><Icon /> Label</>`), and brings prop-refresh from
O(N²) down to O(N) by building a `validChildrenByKey` `Map` once per render.

#### [`perf(UnderlineNav): move overlay anchoring into a layout effect`](https://github.com/primer/react/pull/<id>/commits/d4c587e5b)

Stops reading `containerRef.current` / `listRef.current` during render. The
menu anchor's `left` coordinate is now computed in a layout effect with an
idempotent setter (no feedback loop). Concurrent-rendering safe.

#### [`fix(UnderlineNav): use safe keys for children signature and menu items`](https://github.com/primer/react/pull/<id>/commits/e1d5dbc8b)

Two latent bugs:

1. `childrenKey` joined item keys with `|` — collisions if a consumer used
   `|` in their `key={...}`. Switched to U+0000 (NUL), which React rejects
   inside keys outright.
2. `ActionList.LinkItem`'s `key={menuItemChildren}` coerced non-string
   children to `'[object Object]'` and broke reconciliation for menus of
   such items. Now uses `String(menuItem.key)` — guaranteed-unique inside
   the children set.

#### [`perf(UnderlineNav): use the key Map in the aria-current swap pass`](https://github.com/primer/react/pull/<id>/commits/badd8a4e9)

One leftover O(N²) `.find()` in the swap pass's prop refresh; converted to
the same `validChildrenByKey.get(...)` path.

#### [`style(UnderlineNav): bound the measurement loop by validChildren.length`](https://github.com/primer/react/pull/<id>/commits/2746b1ad3)

Loop tidying for `@typescript-eslint/no-unnecessary-condition`.

#### [`test(UnderlineNav): cover architecture, structure, and perf changes`](https://github.com/primer/react/pull/<id>/commits/1fcc07186)

Adds three test groups that pin the work above and would catch a regression:

- **Architecture**: every item is tagged with `[data-underlinenav-item]`; a
  standalone `<UnderlineNav.Item>` mounts without a parent (no upstream
  measurement-callback dependency); the wrapper is marked
  `data-overflow-measured="true"` synchronously after the layout effect.
- **Structure**: the lifted aria-current swap effect promotes a menu item
  into the visible list when a consumer moves `aria-current` onto it,
  including for items whose children are not a plain string (the swap-
  math-with-text-key regression).
- **Performance**: each item-body renders at most twice during initial
  mount with stable children; per-item render rate does not grow with N;
  `UnderlineNav` itself renders at most 3 times total during initial mount.

Plus a changeset.

## Profiles

Render counts during initial mount with stable children (verified by the
tests above):

| | Before | After |
|---|---|---|
| Total parent renders | ≥ 3 + one per resize tick | ≤ 3 |
| Per-item renders | grows with N (cascade) | constant 1–2 |
| `setState` calls per overflow change | 3 (split state) → 4 with swap | 1 (consolidated) |
| Render-phase `setState`s | 1 (aria-current swap) | 0 (lifted to effect) |
| `.find()` calls inside `.map()` | 2 per render (O(N²)) | 0 (Map lookups) |
| Refs read during render | yes (`containerRef`, `listRef`) | no |

#### New

- Internal `data-underlinenav-item` attribute on every `<li>` rendered by
  `UnderlineNav.Item`. Enables parent-side measurement.

#### Changed

- `UnderlineNav` re-architected: parent-driven measurement, single
  consolidated overflow state, `React.memo` on `UnderlineNav.Item`, single
  combined layout effect, key-Map width and child lookups, layout-effect
  overlay anchoring, safe children-signature delimiter and menu item key.
  Behavior is identical.

#### Removed

- Internal `setChildrenWidth` / `setNoIconChildrenWidth` from
  `UnderlineNavContext` (not exported; were upstream-callback plumbing).
- `ChildSize` / `ChildWidthArray` from
  `packages/react/src/UnderlineNav/types.ts`; replaced with `ChildWidthMap =
  Map<string, number>`. Not exported from `@primer/react`.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

No public API changes. Pure internal refactor + correctness fixes for
already-supported but mis-handled cases (non-string children, key-with-pipe).

### Testing & Reviewing

```bash
# Run UnderlineNav tests only:
cd packages/react
npx vitest run --project '@primer/react (chromium)' src/UnderlineNav/UnderlineNav.test.tsx
# 28 tests pass
```

**Reviewer focus:**

1. The combined layout effect's two-pass structure (measurement +
   aria-current swap) and why it has no deps array. The comment block at
   that effect captures the reasoning — does it survive future readers?
2. `computeOverflow` / `computeSwap` are now pure. Worth checking the
   call-site setState payloads are correctly built (especially the
   `childrenKey: prev.childrenKey` carry-over).
3. The `data-underlinenav-item` selector + positional pairing in the
   measurement loop. Is the assumption "the N-th DOM item corresponds to
   `validChildren[N]`" robust? It holds because measurement runs only
   when all items are in the list (the children-key derived reset
   guarantees this), but it's worth confirming.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


TylerJDev:
Hey @mattcosta7! Thanks for the PR! ✨ 

I was reading this [slack thread](https://github.slack.com/archives/C01L618AEP9/p1779820745725419) and was wondering if we still wanted this one to land, or if we want to decide on which solution we move with? Let me know what you think! 😁

mattcosta7:
> Hey @mattcosta7! Thanks for the PR! ✨
> 
> I was reading this [slack thread](https://github.slack.com/archives/C01L618AEP9/p1779820745725419) and was wondering if we still wanted this one to land, or if we want to decide on which solution we move with? Let me know what you think! 😁

I think we should land either of :

- [Drop the overflow logic and replace entirely with horizontal scrolling](https://github.com/primer/react/pull/7648)
- [Reimplement the overflow logic using CSS wrapping](https://github.com/primer/react/pull/7506) (with the caveat that we lose the ability to prevent the current item from overflowing)

Which more deeply remove the issues I noted, instad of just minimizing them - if we can 

If not, we can land this and then replace it in the future - but I'd prefer one of those more holistic solutions

PR #7888, opened 2026-05-27
AnchoredOverlay: Allow customization of CSS anchor positioning

TylerJDev:
Adds `cssAnchorPositioningSettings` prop to `AnchoredOverlay` so consumers can control how the overlay behaves when native CSS anchor positioning is active.

When CSS anchor positioning is active, the overlay relies on the stylesheet defined `position-try-fallbacks` list on `.AnchoredOverlay`, which can flip the overlay to sides the consumer may not want. This prop lets consumers opt into other fallback behavior via a `fallbackStrategy`:

- `"default"` - use the built-in CSS fallback rules (current behavior).
- `"none"` - keep the requested `side` with no CSS fallbacks (`position-try-fallbacks: none`).
- `"opposite-side"` - only allow flipping to the opposite side along the relevant axis (`flip-block` for `outside-top`/`outside-bottom`, `flip-inline` for `outside-left`/`outside-right`).

The setting is applied as an inline `position-try-fallbacks` style on the overlay element while it is open, and ignored entirely when CSS anchor positioning is not active.

#### New

- `cssAnchorPositioningSettings?: { fallbackStrategy: 'default' | 'none' | 'opposite-side' }` prop on `AnchoredOverlay`.

#### Changed

- When the overlay opens with CSS anchor positioning active, it now writes/removes the inline `position-try-fallbacks` property based on `cssAnchorPositioningSettings`. Existing callers that don't pass the prop default to `"default"`, preserving current behavior.
- Adds default behavior for `inside-*` sides. CSS anchor positioning doesn't natively support them, so this ensures that CSS anchor positioning does not break when used.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/AnchoredOverlay/AnchoredOverlay.tsx
```diff
@@ -125,12 +125,17 @@ interface AnchoredOverlayBaseProps extends Pick<OverlayProps, 'height' | 'width'
...
+   * `fallbackStrategy` controls CSS fallback behavior:
+   * - `"default"`: use built-in CSS fallback rules.
+   * - `"none"`: keep the requested side with no CSS fallbacks.
+   * - `"opposite-side"`: only allow fallback to the opposite side.
+   *
+   * `disable`: When `true`, opts this overlay out of native CSS anchor positioning (and the Popover API)
    *   even if `primer_react_css_anchor_positioning` is enabled and the browser supports it.
    */
-  cssAnchorPositioningSettings?: {disable?: boolean}
+  cssAnchorPositioningSettings?: {fallbackStrategy: 'default' | 'none' | 'opposite-side'; disable?: boolean}
```
TylerJDev:
```suggestion
  cssAnchorPositioningSettings?: {fallbackStrategy?: 'default' | 'none' | 'opposite-side'; disable?: boolean}
```

liuliu-dev (approved):
:shipit:

PR #7889, opened 2026-05-27
data-component adr part 6

llastflowers:
Relates to https://github.com/github/primer/issues/6497

#### New

Add data-component attributes and associated tests for PageHeader, PageLayout, Pagehead, Popover, Portal, and ProgressBar

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/PageHeader/PageHeader.tsx
```diff
@@ -323,7 +337,11 @@ export type ActionsProps = React.PropsWithChildren<ChildrenPropTypes>
 
 const Actions = ({children, className, hidden = false}: ActionsProps) => {
   return (
-    <div className={clsx(classes.Actions, className)} data-component="PH_Actions" {...getHiddenDataAttributes(hidden)}>
+    <div
```
joshblack: I _think_ some of these may be used (like `PH_Actions`) based on: https://github.com/search?q=repo%3Agithub%2Fgithub-ui+PH_&type=code so we might need to be more careful for them

joshblack (commented):
Just wanted to leave a comment real quick, some of the data-component values that exist already might have usage (specifically thinking of the `PH_*` ones so just wanted to double-check on the release plan for those before approving 👀

packages/react/src/PageHeader/PageHeader.tsx
llastflowers: Yeah, I was thinking about this and was planning to check the dotcom usages as well! I think we should definitely update the attribute names in primer/react to keep things consistent, so we need to update them in dotcom too. Since there are only 3 files that reference the old attributes, this seems like an easy enough migration effort. 

I suppose we would just need to make sure that we have a migration PR in the same github-ui deployment group as the primer/react version upgrade that involves this PR, since the changes need to happen in both places at the same time to avoid breaking anything 🤔 Coordinating with RC on this (possibly asking them to queue both PRs to make sure they're deployed together) seems like the way to do it safely. What do you think? 👀

joshblack: Totally agreed, makes sense to me 👍

llastflowers: jk, @francinelucca correctly pointed out that this needs to be part of a major update. I'll revert everything related to the `PH_` stuff in this PR and separate it out!

packages/react/src/PageLayout/PageLayout.tsx
francinelucca, line 181: suggestion: add data-component to VerticalDivider, HorizontalDivider, SideBarDivider and DragHandle

francinelucca (approved):
let's make sure all the data-component attrs that need to be replaced next major are tracked in https://github.com/github/primer/issues/6501 🙏🏽

joshblack (approved):

packages/react/src/PageLayout/PageLayout.tsx
llastflowers: good idea! 👍

llastflowers:
integration tests passing; not reporting to this PR https://github.com/github/github-ui/pull/22691

PR #7893, opened 2026-05-28
perf(Button): replace :has() icon-only-counter selector with data attribute

mattcosta7:
Closes #

Replaces the icon-only-with-counter `:has(...):not(:has(...))` selector on `ButtonBase` with a JS-derived `data-icon-only-counter` attribute computed from props at render time. Same visual output; reduces the per-Button style-recalculation cost.

**Before:**

```css
&:where([data-has-count]):has([data-component='leadingVisual']):not(:has([data-component='text'])) {
  padding-inline: var(--control-medium-paddingInline-condensed);
}
```

**After:**

```tsx
data-icon-only-counter={count !== undefined && LeadingVisual && !children ? true : undefined}
```

```css
&:where([data-icon-only-counter]) {
  padding-inline: var(--control-medium-paddingInline-condensed);
}
```

`:has()` + `:not(:has())` compound selectors are among the most expensive selectors in the CSS engine, and they evaluate against every `.ButtonBase` instance on every DOM mutation inside the button. Buttons render dozens to hundreds of times per page in real GitHub views, so this is a measurable cost. The equivalent JS check is a constant-time prop comparison.

#### New

- `data-icon-only-counter` attribute on `ButtonBase` when an icon-only Button has a `count`

#### Changed

- Internal: `ButtonBase` no longer relies on `:has()` for icon-only-with-counter styling

#### Removed

- The `:has(...):not(:has(...))` compound selector from `ButtonBase.module.css`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

No visual changes expected. The new attribute is only set when the existing `:has()` selector would have matched (`count` set + `leadingVisual` rendered + no `children`).

Existing Button unit tests pass locally (138 tests across Button + ActionList suites).

Part of a small series replacing `:has()` selectors on high-frequency components with JS-derived data attributes. A follow-up PR will apply the same pattern to `ActionList`.

### Merge checklist

- [x] Added/updated tests 
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/perf-button-icon-only-counter.md
+++ b/.changeset/perf-button-icon-only-counter.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Button: Replace the icon-only-with-counter `:has(...):not(:has(...))` selector with a `data-icon-only-counter` attribute computed from props. Reduces style-recalculation cost on pages that render many Buttons. No visual or behavioral changes.
--- a/packages/react/src/Button/ButtonBase.module.css
+++ b/packages/react/src/Button/ButtonBase.module.css
@@ -640,8 +640,7 @@
 
   /* Icon-only + Counter */
 
-  /* stylelint-disable-next-line selector-pseudo-class-disallowed-list -- scoped to CSS Module, audited (github/github-ui#17224) */
-  &:where([data-has-count]):has([data-component='leadingVisual']):not(:has([data-component='text'])) {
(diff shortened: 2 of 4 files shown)


jonrohan (approved):

jonrohan:
Adding an attribute doesn't really need an integration test

PR #7899, opened 2026-05-29
perf(TreeView): make rows paint-safe and reduce hover/focus invalidation

mattcosta7:
Closes #

Five targeted improvements to \`TreeView\` style + structure aimed at making rows cheaper to render at scale and removing a latent bug in the documented \`containIntrinsicSize\` API. **No visual, layout, or public-API changes at the default rendering** — all changes are either invisible at the default state or behind a CSS containment property the consumer opts into.

#### Changed

- **Current-item indicator is now paint-safe via \`overflow-clip-margin\`.** The indicator pseudo-element is positioned at \`left: -8px\` of \`.TreeViewItemContainer\`. That meant any consumer applying \`contain: paint\` to the row, or the documented \`containIntrinsicSize\` prop on \`TreeView.Item\` (which sets \`content-visibility: auto\` — implying \`contain: paint\` — on the same container), would clip the indicator on \`current\` items. Both \`.TreeViewItem\` and \`.TreeViewItemContainer\` now declare \`overflow-clip-margin: var(--base-size-8)\`. The property is a no-op when no paint containment is active, so default rendering is byte-identical. 
- **Skeleton-row hover suppression no longer uses \`:has()\`.** \`:has(.TreeViewItemSkeleton):hover\` matched on every row container in every tree, forcing subtree invalidation tracking on every row. \`LoadingItem\` now communicates with the placeholder \`Item\` via a module-private \`LoadingPlaceholderContext\` that emits a positive \`data-loading\` attribute on the \`<li>\`, and the CSS selector targets that directly. **No new public prop.**
- **Nesting indicator lines no longer rely on a root-scope \`:hover\`/\`:focus-within\` descendant selector.** Color is driven by an inherited \`--tree-line-color\` custom property toggled on the root \`<ul>\`. Same UX (lines fade in while the tree is interacted with on hover-capable devices, always visible on coarse pointers); the engine updates one property on one element instead of re-matching the descendant selector against every \`.TreeViewItemLevelLine\` in the tree.
- **Fixed unitless \`outline-offset: -2\` in the forced-colors focus-ring fallback.** Browsers were silently dropping the declaration, so forced-colors users were getting no focus indicator on tree items.
- **Made \`grid-template-columns\` declare the \`trailingAction\` track explicitly** (\`auto\`) so it matches the 5-area \`grid-template-areas\` declaration on the same element. Clarity, no behavior change.

#### New

- Storybook story \`TreeView/Features/CurrentItemWithContainIntrinsicSize\` exercising the previously-broken \`current\` + \`containIntrinsicSize\` combination so VRT locks in the contract going forward.

#### Removed

Nothing public.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- All 94 \`packages/react/src/TreeView/**\` unit tests pass.
- \`tsc --noEmit\` on \`packages/react\` is clean.

**VRT expectations:** No change in default rendering. The one place a change could surface is forced-colors focus rings on tree items, which go from \"nothing\" (the silent \`outline-offset: -2\` drop) to a 2px inset \`HighlightText\` outline. That's an accessibility fix.

Internal-only note: this lets the github-ui escape hatch in \`packages/diff-file-tree/DiffFileTree.module.css\` (\`content-visibility: visible; contain: none;\` for \`[aria-current='true']\` and \`:focus-visible\`) be removed once this lands.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/TreeView/TreeView.module.css
```diff
@@ -66,25 +78,24 @@
...
-
-      @media (forced-colors: active) {
-        outline: none;
-      }
-    }
   }
 
   &:where([data-omit-spacer='true']) .TreeViewItemContainer {
-    grid-template-columns: 0 0 0 1fr;
+    grid-template-columns: 0 0 0 1fr auto;
```
mattcosta7, line 84: auto is the default, but it reads nicer imp when it's 5 columns defined for 5 columns

liuliu-dev (approved):

PR #7902, opened 2026-05-29
perf(ActionList): replace `:has(~ .SubGroup [data-active])` collapsed-parent indicator with attribute selector

mattcosta7:
Closes #

Replaces the most expensive `:has()` selector remaining in `packages/react/src/ActionList/ActionList.module.css` — `&:has(~ .SubGroup [data-active='true'])` — with a plain attribute + child-combinator selector. This was the worst-shape `:has()` left in the package: it combined a **general sibling combinator** with a **descendant search**, and the engine had to re-run that walk against every collapsed parent any time a descendant `data-active` attribute changed.

**Before:**

```css
.ActionListContent[aria-expanded='false']:has(~ .SubGroup [data-active='true']) {
  /* active-parent indicator styles */
}
```

**After:**

```css
.ActionListItem[data-has-subitem='true'][data-active='true']
  > .ActionListContent[aria-expanded='false'] {
  /* active-parent indicator styles */
}
```

The selector goes from a sibling walk + descendant subtree search (re-evaluated on every `data-active` mutation inside the subtree) to a constant-time attribute lookup on the same `<li>` that the consumer already controls.

**Why this is safe:**

The new selector depends on `data-active` being set on the parent `<li>`. The only in-repo consumer that renders an `ActionList.Item` with `ActionList.SubItem` is `NavList.ItemWithSubNav`, which already does its own `hasCurrentNavItem` recursion and passes `active={!isOpen && containsCurrentItem}` to the parent — exactly matching the condition the old `:has()` was detecting. Net visual output is identical for `NavList.SubNav` usage.

**Behavior change for direct `ActionList.Item` + `ActionList.SubItem` consumers**

If a consumer renders `ActionList.SubItem` *directly* (without `NavList.SubNav`) and relied on the `:has()` selector to auto-light-up a collapsed parent when a descendant had `data-active='true'`, they will now need to pass `active={true}` to the parent `ActionList.Item` themselves. This is the same pattern `NavList.ItemWithSubNav` already implements; no other in-repo usage exists.

#### Changed

- Internal: `ActionList.module.css` no longer uses `:has()` to detect an active descendant in a collapsed parent's sub-group; the same styles fire when the parent `ActionList.Item` itself has `active={true}` (which `NavList.ItemWithSubNav` already sets via its existing `containsCurrentItem` walk).

#### Removed

- The `&:has(~ .SubGroup [data-active='true'])` selector from `ActionList.module.css`.

### Rollout strategy

- [x] Patch release
- [ ] Minor release
[...]

PR #7904, opened 2026-05-29
perf(SelectPanel v2): replace `:has(input:placeholder-shown)` with `data-empty` attribute

mattcosta7:
Closes #

Replaces the `&:has(input:placeholder-shown) :global(.TextInput-action)` selector that hid the clear-action button when the search input is empty. The previous selector forced a descendant `:has()` evaluation on every keystroke (placeholder-shown toggles with each character typed/deleted), and the inner `:global(.TextInput-action)` selector required descendant matching across the whole TextInput subtree.

**Approach:** track the empty state in React state on the `SelectPanelSearchInput` component, derive an `isEmpty` value, and pass it through as `data-empty` on the rendered `TextInput` wrapper. The CSS now matches a plain attribute on the same `.TextInput` element.

The controlled/uncontrolled cases are both handled:

```tsx
const [uncontrolledEmpty, setUncontrolledEmpty] = React.useState(
  () => !(props.value ?? props.defaultValue),
)
const isEmpty = props.value !== undefined ? props.value === '' : uncontrolledEmpty
```

- For **controlled** inputs (`value` prop provided), `isEmpty` is derived from the prop directly so it stays in sync without an effect.
- For **uncontrolled** inputs, the change handler and the clear-action click handler update the `uncontrolledEmpty` state.

#### New

- `data-empty` attribute on `SelectPanel.SearchInput` when the search input is empty.

#### Changed

- Internal: `SelectPanel.module.css` clear-action visibility no longer uses `:has(input:placeholder-shown)`.

#### Removed

- The `&:has(input:placeholder-shown)` selector from `SelectPanel.module.css`.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- All 16 `packages/react/src/experimental/SelectPanel2/**` unit tests pass.
- `tsc --noEmit` on `packages/react` is clean.
- Stylelint + Prettier + ESLint clean on touched files (only pre-existing `content-visibility` browser-compat warning at line 87, unrelated).

**VRT expectations:** None. The clear-action button is shown/hidden under identical conditions as before — when the input is empty, it's hidden; otherwise, it's shown. The only behavioral difference is timing: the previous `:has(input:placeholder-shown)` updated on the *browser's* `:placeholder-shown` flip (which happens on the same frame as the input event), while the new `data-empty` flips on the React commit that follows the change event. In practice both update in the same animation frame and the user can't observe the difference.

Part of the `:has()`-reduction series: #7901 (PageHeader), #7902 (ActionList SubGroup), #7903 (SegmentedControl divider).

### Merge checklist

- [x] Added/updated tests 
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mattcosta7:
Closing as no longer necessary. Main already merged #8115 ("perf(css): remove expensive `:has()` selectors for Safari style-recalc"), which replaced the `:has(input:placeholder-shown)` selector on `.TextInput` with a pure-CSS forward-sibling selector (`& input:placeholder-shown ~ .TextInput-action`). The perf goal here is already achieved on main without extra React state, so this PR's `data-empty` approach is redundant.

PR #7908, opened 2026-05-29
Fix KeybindingHint key display on non-MacOS/Windows platforms

### LINKED ISSUE #7756: `KeybindingHint` only considers MacOS and Windows when displaying key names
When used as a modifier in `KeybindingHint`, the `Meta` key is correctly displayed as `Win` (Windows key) on Windows and `⌘` (Command key) on MacOS. However, on all other devices, it is displayed as `Win` even though the Windows key does not exist on other devices. 

### `Meta`

- MacOS: `⌘`/`Command` (matching the logic from [github/hotkey](https://github.com/github/hotkey/blob/f26fabb7da87c8f7d9748e035c7b6c354580d564/src/hotkey.ts#L89), and based on the fact that the iPad Magic Keyboard has a Command key)
- All other operating systems: `Win`/`Windows`

### `Alt`

- MacOS: `⌥`/`Option` 
- All other operating systems: `Alt`

### `Mod` (dynamic modifier)

-  MacOS: `⌘`/`Command`
- All other operating systems: `⌃`/`Control`

### `Meta`

- MacOS <ins>and iOS</ins>: `⌘`/`Command` (matching the logic from [github/hotkey](https://github.com/github/hotkey/blob/f26fabb7da87c8f7d9748e035c7b6c354580d564/src/hotkey.ts#L89), and based on the fact that the iPad Magic Keyboard has a Command key)
- Windows: `Win`/`Windows`
- <ins>All other operating systems: `Meta`</ins> (since the key doesn't map to anything consistent on Linux/Android)

### `Alt`

- MacOS <ins>and iOS</ins>: `⌥`/`Option` 
- All other operating systems: `Alt`

### `Mod` (dynamic modifier)

-  MacOS <ins>and iOS</ins>: `⌘`/`Command`
- All other operating systems: `⌃`/`Control`

jonrohan:
Closes #7756

Here's a preview of all the modifier keys on different platforms https://primer-11d74e8128-13348165.drafts.github.io/storybook/?path=/story/experimental-components-keybindinghint-features--platforms

CleanShot 2026-05-29 at 16 17 42@2x

CleanShot 2026-05-29 at 16 19 01@2x

`KeybindingHint` previously only distinguished between MacOS and "not MacOS" when displaying the `Meta`, `Alt`, and `Mod` keys. As a result, the `Meta` key was rendered as `Win` (Windows key) on Linux, Android, and other platforms where that key does not exist.

This change introduces a three-way platform detection (`mac` / `windows` / `other`) so keys are displayed correctly per platform.

#### New

- Added an SSR-safe `usePlatform()` hook (internal to `KeybindingHint`) that returns `'mac' | 'windows' | 'other'`. Apple platforms (macOS **and** iOS/iPadOS) map to `'mac'`.

#### Changed

- `KeybindingHint` now resolves platform-specific keys based on the detected platform:
  - `Meta`: `⌘`/`Command` on Apple, `Win`/`Windows` on Windows, `Meta` on all other platforms.
  - `Alt`: `⌥`/`Option` on Apple, `Alt` elsewhere.
  - `Mod`: `⌘`/`Command` on Apple, `⌃`/`Control` elsewhere.
- `getAccessibleKeybindingHintString` remains backward compatible — it still accepts the previous `boolean` argument (`true` → `'mac'`, `false` → `'other'`) in addition to a `Platform` value.

#### Removed

- N/A

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

Unit tests were added covering `Meta`, `Alt`, and `Mod` rendering across all three platform categories. Existing `KeybindingHint`, `Tooltip`, and `TreeView` tests pass.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/KeybindingHint/platform.ts
iansan5653, line 1: `useIsMacOS` is a common, shared utility hook. Should `usePlatform` live in the same place? Or maybe we should just update that hook to be more powerful? Seems like this could be useful for other components.

packages/react/src/KeybindingHint/KeybindingHint.tsx
```diff
@@ -37,5 +38,10 @@ KeybindingHint.displayName = 'KeybindingHint'
  *
  * NOTE that this string should _only_ be used when building `aria-label` or `aria-description` props (never rendered
  * visibly) and should nearly always also be paired with a visible hint for sighted users.
+ *
+ * The `platform` argument controls how platform-specific keys (such as `Meta`, `Alt`, and `Mod`) are named. For
+ * backwards compatibility, a `boolean` may be passed instead, where `true` is treated as `'mac'` and `false` as
+ * `'other'`.
  */
-export const getAccessibleKeybindingHintString = accessibleSequenceString
+export const getAccessibleKeybindingHintString = (sequence: string, platform: Platform | boolean) =>
```
iansan5653, line 46: A followup task after release should be to update consumers in `github/github` to use the new `Platform` type instead of `boolean`

packages/react/src/KeybindingHint/platform.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * The platform categories that affect how keyboard shortcut keys are displayed.
+ *
+ * - `mac`: Apple platforms (macOS and iOS/iPadOS), which use the Command and Option keys.
+ * - `windows`: Windows, which uses the Windows (Meta) key.
+ * - `other`: Any other platform (e.g. Linux, Android), where the Meta key does not map to a
+ *   consistent label.
+ */
+export type Platform = 'mac' | 'windows' | 'other'
```
iansan5653: [nit] Would it be clearer to rename `mac` to `apple`? Since iOS/iPadOS aren't Mac OSes.

iansan5653 (approved):

TylerJDev (approved):
LGTM! +1 to Ian's comments, though approving ahead of time.

PR #7915, opened 2026-06-01
fix(Dialog): close on first Escape press when close button is focused

jonrohan:
Closes https://github.com/github/primer/issues/2604

When a `Dialog` opens, focus moves to its close button. Because icon buttons now render a tooltip by default, the tooltip's own `Escape` handler (registered on `document` via `useOnEscapePress`) runs first and calls `stopImmediatePropagation()` + `preventDefault()` to dismiss the tooltip — swallowing the first `Escape` keypress so the dialog doesn't close. Users had to press `Escape` twice (the previous test even worked around this), and in a multi-dialog setup the first `Escape` could shift focus to the wrong return-focus target instead of closing the dialog.

### Fix

Intercept the `Escape` keydown **on the close button itself**. React dispatches `onKeyDown` at the root container, which sits below `document` in the DOM, so calling `stopPropagation()` there prevents the event from reaching the tooltip's document-level handler. We close the dialog directly with the `'escape'` gesture, and the tooltip stays fully functional.

#### Considered but rejected: disabling the tooltip

The simplest fix was to set `unsafeDisableTooltip` on the close button. We chose **not** to do this because the tooltip contributes to the button's discoverability/affordance and we didn't want to regress the accessibility/UX of the close button just to work around the event handling. Intercepting the event keeps the tooltip — and its behavior — intact while fixing the bug.

#### Changed

- `Dialog` now closes on the first `Escape` keypress when the close button is focused, instead of requiring a second press.

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

1. Render two buttons that each open their own `Dialog` (each with a `returnFocusRef`).
2. Open the first dialog and press `Escape` once.
3. The dialog should close on the first press and focus should return to its own trigger button.

Added a regression test in `Dialog.test.tsx` covering the single-`Escape` close and the multi-dialog scenario, and updated the existing Escape test to expect a single keypress. All 40 Dialog tests pass; type-check is clean.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/.changeset/dialog-escape-close-button-tooltip.md
+++ b/.changeset/dialog-escape-close-button-tooltip.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Dialog: Fix `Escape` key not closing the dialog on the first keypress when the close button is focused
--- a/packages/react/src/Dialog/Dialog.test.tsx
+++ b/packages/react/src/Dialog/Dialog.test.tsx
@@ -176,10 +176,52 @@ describe('Dialog', () => {
 
     expect(onClose).not.toHaveBeenCalled()
 
-    await user.keyboard('{Escape}') // escape once to remove focus from the close button
-    await user.keyboard('{Escape}') // escape again to trigger the onClose
+    await user.keyboard('{Escape}')
 
     expect(onClose).toHaveBeenCalledWith('escape')
+  })
+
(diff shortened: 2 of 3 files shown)


llastflowers (approved):
Tested in Storybook, looks great! ✨

PR #8001, opened 2026-06-16
perf(useMedia): read live matchMedia on first client render via useSyncExternalStore

mattcosta7:
## Overview

**Status: WIP / draft — not ready to merge.** Opening early for direction + discussion.

The `react-hooks/set-state-in-effect` lint rule (from `eslint-plugin-react-hooks@7.1.1`, the "you might not need an effect" check) flags effects whose only job is to compute state. This PR is a **full accounting of every suppression of that rule in the repo**: the genuinely-derivable effects are rewritten to compute state during render (or via an external store), and every remaining suppression is either documented in code as an intentional effect or listed below as deferred demo/test cleanup.

No public API changes.

#### Changed

- `useMedia` now reads its value via `useSyncExternalStore` instead of `useState` + `useEffect`, eliminating the extra render on mount and the SSR/hydration `setState`. On the client, the first render returns the live `matchMedia` value rather than `defaultState` (which now serves purely as the SSR/hydration snapshot).
- `ToggleSwitch` hides the loading status label via render-time derivation; only the *delayed reveal* remains in an effect (it needs a timer).
- `SelectPanel` resets the single-select-modal `intermediateSelected` value by tracking the previous `open` state during render instead of in an effect.
- `ValidationAnimationContainer` (internal) mounts its content via render-time derivation; un-mounting stays deferred to `onAnimationEnd` so the exit animation still plays.
- Documented the remaining intentional effects (DOM reads / side effects) inline so each suppression states *why* it cannot be derived.
- Test-only: the `useMedia` test's `matchMedia` mock now updates `matches` on `change()`, mirroring real `MediaQueryList` behavior (the old mock only fired the event).

#### New

_None._

#### Removed

_None._

### Rollout strategy

- [x] Patch release

## Complete `set-state-in-effect` inventory

**Repo-wide before this PR:** 22 suppressions (11 in component/hook source, 11 in stories + storybook data).

### ✅ Fixed — genuinely derivable (4)

| Site | Rewrite |
| --- | --- |
| [`hooks/useMedia.ts`](../blob/copilot/derive-state-from-effects/packages/react/src/hooks/useMedia.ts) | `useSyncExternalStore` (tear-free, no redundant initial render) |
| [`ToggleSwitch.tsx`](../blob/copilot/derive-state-from-effects/packages/react/src/ToggleSwitch/ToggleSwitch.tsx) | Render-time derive (hide-label branch); timer stays in effect |
| [`SelectPanel.tsx`](../blob/copilot/derive-state-from-effects/packages/react/src/SelectPanel/SelectPanel.tsx) (`intermediateSelected`) | Adjust-during-render on `open` transition |
| [`internal/.../ValidationAnimationContainer.tsx`](../blob/copilot/derive-state-from-effects/packages/react/src/internal/components/ValidationAnimationContainer.tsx) | Render-time derive; exit deferred to `onAnimationEnd` |

### 🟡 Intentional effects — now documented in code (7)

These trip the rule but are **necessary** effects (committed-DOM reads or consumer side effects). Each suppression now carries a one-line rationale.

| Site | Why it stays an effect | Cleanly splittable later? |
| --- | --- | --- |
| `TreeView.tsx` (`setSubTreeLabel`) | Reads committed DOM via `getAccessibleName` | No |
| `LabelGroup.tsx` (`hideChildrenAfterIndex`) | Reads rendered children from the DOM | 🔜 numeric branch could derive from child indices |
| `Autocomplete/AutocompleteMenu.tsx` (sort-on-close) | Shares the effect with the `onOpenChange` callback | 🔜 split sort (derive) from `onOpenChange` (effect) |
| `Dialog.tsx` (`setHasRendered`) | Commit-timing focus hack | No |
| `experimental/SelectPanel2/SelectPanel.tsx` (`setLabelText`) | Reads label `textContent` from committed DOM | No |
| `SelectPanel.tsx` (no-items announce) | Commit-timing a11y live-region announcement | No |
| `SelectPanel.tsx` (first-open fetch) | Fires a consumer callback (side effect) | No |

### 🟦 Deferred — demo/test code (11, not touched)

Same pattern in stories + storybook data. Lower value and risk of changing story/VRT behavior, so deferred per "clean these separately":

- `DataTable/storybook/data.ts` (1)
- `experimental/SelectPanel2/SelectPanel.examples.stories.tsx` (1)
- `FormControl/FormControl.features.stories.tsx` (2)
- `SelectPanel/SelectPanel.examples.stories.tsx` (4)
- `SelectPanel/SelectPanel.features.stories.tsx` (3)

> [!NOTE]
> Related but **not** flagged by the rule: `TreeView`'s `setIsSubTreeEmpty` effect writes to a **parent's** `ItemContext` setter. The rule only flags same-component state, and this one genuinely can't move to render (cross-component update), so it correctly stays an effect.

### Testing & Reviewing

All functional changes are behavior-preserving refactors. Validated locally:

- `useMedia` (4), `ToggleSwitch` (16), `FormControl` (39), `CheckboxGroup` (9), `RadioGroup` (7), `SelectPanel` (152) unit tests pass.
- `tsc --noEmit` and `eslint` pass on the changed files.

Please pay special attention to:

- **`useMedia`** — the client-first-render value change (live value vs `defaultState`). Mirrors what `useIsMacOS` already does.
- **`SelectPanel`** — the reset now keys on `open` transitions (matching the original comment). The previous effect also re-ran on `selected`/`isSingleSelectModal` mid-open; not expected for a single-select modal (selection commits on close), but worth confirming.

### Merge checklist

- [x] Added/updated tests
- [x] Tested in Chrome (Vitest browser project)
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

> [!IMPORTANT]
> The 🟦 deferred demo/test items and the 🔜 splittable source effects (`LabelGroup` numeric branch, `AutocompleteMenu` sort-on-close) are intentionally left for separate, individually-tested PRs.


mattcosta7:
@copilot resolve the merge conflicts in this pull request

packages/react/src/hooks/useMedia.ts
```diff
@@ -18,71 +17,65 @@ import {MatchMediaContext} from './MatchMediaContext'
...
+    if (defaultState !== undefined) {
+      return defaultState
     }
 
-    // Make sure the media query list is in sync with the matches state
-    // eslint-disable-next-line react-hooks/set-state-in-effect, react-you-might-not-need-an-effect/no-external-store-subscription
-    setMatches(mediaQueryList.matches)
+    // A default value has not been provided, and you are rendering on the
+    // server, warn of a possible hydration mismatch when defaulting to false.
+    warning(
```
jonrohan, line 74: Since `getServerSnapshot` also runs during client hydration, does this now warn in the browser console (not just on the server) when `defaultState` is omitted in an SSR app, where before the client read `matchMedia` directly and stayed quiet?

mattcosta7: Good question — no, it won't warn in the browser. The warning is gated on `typeof window === 'undefined'`, so even though `getServerSnapshot` also runs on the client during hydration, the SSR guidance message is only emitted on the actual server (matching the prior behavior where the client read `matchMedia` and stayed quiet). There's now a regression test for exactly this — `does not warn about a missing defaultState on the client` — which renders the hook on the client and asserts the warning is never logged. Resolving; happy to reopen if you'd like a different approach.

packages/react/src/SelectPanel/SelectPanel.tsx
```diff
@@ -241,11 +241,14 @@ function Panel({
...
   )
 
-  // Reset the intermediate selected item when the panel is open/closed
-  useEffect(() => {
-    // eslint-disable-next-line react-hooks/set-state-in-effect, react-you-might-not-need-an-effect/no-derived-state
+  // Reset the intermediate selected item when the panel is opened or closed.
+  // Tracking the previous `open` value lets us derive this during render instead
+  // of in an effect, avoiding an extra render pass each time the panel toggles.
+  const [prevOpen, setPrevOpen] = useState(open)
+  if (prevOpen !== open) {
```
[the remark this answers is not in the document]
mattcosta7: Intentional: a ref can't be used here because writing a ref during render is disallowed (and trips the ref-purity lint rules this series enforces), so the previous-`open`-in-state pattern is React's sanctioned way to derive this during render. Crucially this does not force an extra *committed* render — React re-runs the component synchronously before painting and discards the intermediate output, so there's no extra commit or browser paint. This is documented inline. (Note: this SelectPanel change also overlaps with #8025, which supersedes it.)

jonrohan (approved):

mattcosta7:
Rescoped this PR down to just the `useMedia` refactor. The previous ToggleSwitch, SelectPanel, and ValidationAnimationContainer changes overlapped with (and in a couple of cases were superseded by) the focused PRs #8024, #8025, and #8023, so they've been dropped here. The comment-only edits to Autocomplete/Dialog/LabelGroup/TreeView/SelectPanel2 were also removed. The changeset was retargeted accordingly. Both review threads remain addressed by the current `useMedia` code (SSR warning gated on `typeof window === 'undefined'` with a client-no-warn test).

mattcosta7:
CI went green in gh/gh-ui, but `unused-code` flagged on unrelated changes

PR #8021, opened 2026-06-19
perf(TextInput, Textarea): derive character counter in render

mattcosta7:
## Overview

`TextInput` and `Textarea` previously stored the character counter (`characterCount`, `isOverLimit`) and the screen-reader message in state and updated them from `useEffect`s via a shared `CharacterCounter` class. On every keystroke this scheduled `setState` from effects, causing extra render passes — and in `Textarea`, a double render per keystroke because the `[value]` effect re-ran after commit. Validation (`aria-invalid`) also lagged a render behind.

This PR derives the counter and validation state **directly during render** from the current input length, so there is no `setState`-in-effect and no render cascade. The only remaining state updates are the uncontrolled input length (set in the change handler) and the debounced screen-reader announcement, which is applied inside `startTransition` so it never blocks typing.

The shared `CharacterCounter` class is removed in favor of a pure `getCharacterCountState()` helper used by both components.

#### Changed

- `TextInput` and `Textarea` derive the character counter and validation state during render instead of syncing them from effects, removing extra re-renders on each keystroke and making validation update immediately.
- The character-count screen-reader announcement is debounced and applied as a React transition so it stays off the typing critical path.

#### Removed

- Internal `CharacterCounter` class (replaced by the pure `getCharacterCountState` helper). It was not part of the public API.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

From the consumer's perspective behavior is unchanged: counter display, the over-limit error state and alert icon, `aria-describedby` wiring, the `aria-live` region, and "no announcement on initial load" all behave as before.

- Added unit tests for the pure `getCharacterCountState` helper.
- Existing character-counter behavior tests for both `TextInput` and `Textarea` continue to pass.
- Added an integration test in each component asserting the debounced screen-reader announcement appears after typing.

One subtle behavior change worth a look: a purely programmatic `value` change on a controlled input no longer fires a debounced screen-reader announcement (the visible counter still updates). This avoids announcing non-user-initiated changes and is not covered by existing tests.

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mattcosta7:
integration tests passed fully - but sactions ddin't report properly

packages/react/src/utils/useCharacterCountAnnouncement.ts
joshblack, line 1: For this logic, it'd be nice to consolidate on the `AriaStatus` or `Announce` components, or use `@primer/live-region-element` directly, instead of re-creating the debounce logic. Let me know what you think @mattcosta7 👀

mattcosta7: I think that's reasonable - but maybe better as a followup to this pr, which moves those around but doesn't really change them directly?

Maybe we can file a follow up issue to get tackled later?

mattcosta7: Good call to flag this — I spiked it out and it _almost_ works, but it regresses the exact thing this PR is optimizing, so I'd like to keep it as the follow-up.

The drop-in is clean:

```tsx
<VisuallyHidden>
  <AriaStatus delayMs={SCREEN_READER_DELAY}>{counter?.message}</AriaStatus>
</VisuallyHidden>
```

That deletes `useCharacterCountAnnouncement`, its test, and the `announce()` call in the change handlers. But two things surface when I run the suite:

1. **Extra commit per keystroke.** `AriaStatus` → `Announce` tracks `previousAnnouncementText` in state and fires a `MutationObserver`-driven `setState` on _every_ content change. So each keystroke that changes the counter message produces a second commit. It's isolated to the hidden announcement subtree (it does **not** re-render the input, and `nestedUpdateCount` stays `0` — no synchronous cascade), but the render-counter perf tests go from `updateCount === 1` to `2` per change. The current hook debounces at the React-state level, so during continuous typing it commits the announcement **at most once, after typing stops**. `AriaStatus`'s `delayMs` only debounces the live-region text, not the React commit. For a PR whose whole point is removing per-keystroke renders, that's a step backward.

2. **`act(...)` warnings.** The `MutationObserver` update lands outside `act` during `userEvent.type` / controlled `rerender`, so several existing typing tests emit React `act` warnings that fail CI's `VITEST_FAIL_ON_CONSOLE`.

So I'd rather keep the lightweight debounced hook here and do the consolidation as a follow-up — ideally alongside an `AriaStatus`/`@primer/live-region-element` tweak so announcing doesn't require a per-change React commit, which would address the "don't re-create the debounce logic" intent without regressing this PR's metric. I'll file an issue to track it.

mattcosta7: ^ maybe I can fix that in the same PR though

mattcosta7: https://github.com/primer/react/pull/8021/commits/08ba67965a2c9b9af24258cd288c6042a9cb66ab

fixed that double commit in this branch

packages/react/src/live-region/Announce.tsx
```diff
@@ -52,7 +52,10 @@ export function Announce<As extends React.ElementType = 'div'>(props: AnnouncePr
     ...rest
   } = props
   const ref = useRef<ElementRef<'div'>>(null)
-  const [previousAnnouncementText, setPreviousAnnouncementText] = useState<string | null>(null)
```
mattcosta7, line 55: @joshblack I moved this out of state to avoid the commit cascades - this way the performance fixes in Textarea and TextInput can apply broadly _while_ using these components internally

joshblack (approved):

PR #8029, opened 2026-06-22
feat: add octicons-react-symbols package

joshblack:
Introduce a new package, `@primer/octicons-react-symbols`, that provides a way for teams to use references to SVGs instead of inlining SVG definitions for each icon.

To do that, this package exposes an `OcticonSymbols` component that acts as a container for all `<symbol>` definitions for icons that are provided through the `symbols` prop. Then, teams may use a `{Name}IconReference` compoennt to reference that symbol using the `<use>` element.

The goal of this work is to:

- Reduce HTML payload size on large pages
- Reduce DOM size by reducing the number of svg elements on large pages

If teams are introduced, we could also add lint rules to enforce that certain icons are used as references in specific sub-trees/routes. We could also invest in loading up references automatically for defined boundaries (like routes) when bundling for better DX.

#### New

- Add a new package for rendering icon references instead of inlining SVG definitions

### Rollout strategy

- [x] Minor release


joshblack:
@copilot fix CI

joshblack:
@copilot fix CI

joshblack:
@copilot fix CI failures

packages/octicons-react-symbols/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "types": "./dist/generated/index.d.ts",
+    "default": "./dist/generated/index.js"
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "node script/build.ts && rolldown -c",
+    "clean": "rimraf dist",
+    "lint:npm": "npm run build && publint --types",
```
joshblack: @copilot instead of explicitly building, can we add build as a dependency to this task in turbo?

joshblack (commented):
@copilot can you make it so the npm / pack workflow includes this package?

.github/workflows/npm.yml
```diff
@@ -45,7 +45,10 @@ jobs:
 
             const {stdout: queryOutput} = await exec.getExecOutput('npm', ['query', '.workspace']);
             const workspaces = JSON.parse(queryOutput);
-            const packages = workspaces.filter(pkg => pkg.private !== true).map(pkg => pkg.name);
+            const packagesToInclude = new Set(['@primer/octicons-react-symbols']);
+            const packages = workspaces
+              .filter(pkg => pkg.private !== true || packagesToInclude.has(pkg.name))
+              .map(pkg => pkg.name);
```
joshblack: @copilot instead of adding this check, just make the package not private

joshblack (commented):
@copilot this is missing some stuff on the svg element that `@primer/octicons-react` has. Can you add that back in to what we render in the SVG element?

joshblack (commented):
@copilot don't include the deprecated verticalAlign prop

.github/workflows/npm.yml
joshblack, line 1: @copilot revert the change in this file

joshblack (commented):
@copilot I want to make it so that this library also supports this pattern:

```
import { ExampleIconSymbol, ExampleIconReference } from '@primer/octicons-react-symbols/Example'
```

Make an update to our bundling strategy to allow for this and update the package.json exports to allow for wildcard imports to these icons. My assumption is that we can use the generated directory as the target in dist for the wildcard import patterns

joshblack:
Closing in favor of: https://github.com/primer/octicons/pull/1334

PR #8035, opened 2026-06-23
Underline nav shared observer coalesced rebuild

mattcosta7:
Closes #

Follow-up performance work on top of the merged `UnderlineNav` rewrite (#7506) and `ActionBar` overflow handling. Replaces the per-item overflow detection (one `IntersectionObserver` per item plus per-item `offsetTop` reads) with a **single shared, root-scoped `IntersectionObserver` per component**, and coalesces same-frame registry rebuilds into one microtask-scheduled update.

This is internal-only: no public API changes.

### Why

- **Removes forced reflow.** The previous approach read `ref.current.offsetTop` inside `useSyncExternalStore`'s snapshot, forcing a synchronous layout per item. Overflow is now derived from `IntersectionObserver` entries off the render path; snapshots read a cached boolean.
- **Fewer observers.** One observer per registry instead of one per item, lazily created and reused.
- **Minimal re-renders.** Only items whose overflow state actually flips re-render; structural registry rebuilds are coalesced via `queueMicrotask` so a resize that wraps several items triggers a single rebuild.

#### Changed

- `UnderlineNav` and `ActionBar` now detect item overflow using a single shared root-scoped `IntersectionObserver` per component instead of one observer per item.
- `createDescendantRegistry` gains an opt-in `overflow` option exposing `useRegisterOverflowObserver`, plus microtask-coalesced rebuilds.

#### Removed

- Per-item `IntersectionObserver` + `offsetTop` overflow checks in `UnderlineNav.Item` and `ActionBar` items.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

No behavioral changes are intended. Existing `ActionBar`, `UnderlineNav`, and `descendant-registry` unit tests pass. Verify overflow behavior by resizing the viewport: items wrap into the overflow menu and the "More" affordance appears/disappears as before, and overflowed items remain non-focusable (`tabIndex={-1}` / `aria-hidden`).

### Merge checklist

- [x] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -141,7 +141,9 @@ export type ActionBarMenuProps = {
   returnFocusRef?: React.RefObject<HTMLElement>
 } & IconButtonProps
 
-const ActionBarItemsRegistry = createDescendantRegistry<ChildProps | null>()
+// Items opt into a single shared IntersectionObserver via `useRegisterOverflowObserver` instead of each item creating
+// its own observer.
```
iansan5653: [nit] Can probably drop this comment, it's copilot justifying changes now but doesn't really make sense out of the context of this PR
```suggestion
```

packages/react/src/utils/descendant-registry.tsx
```diff
@@ -8,15 +8,19 @@ import {
...
+  type RefObject,
 } from 'react'
 import useIsomorphicLayoutEffect from './useIsomorphicLayoutEffect'
 
 export interface ProviderProps<T> {
   children: ReactNode
   /** State setter from `useRegistryState`. */
   setRegistry: Dispatch<React.SetStateAction<ReadonlyMap<string, T> | undefined>>
+  /** Clipping container used as the `IntersectionObserver` root for overflow detection. */
+  rootRef?: RefObject<Element | null>
```
iansan5653: 🤔 I like the move to a single observer, but I would like to avoid baking it into `descendant-registry`. While this is currently only used for the two similar overflow patterns, it's a generic pattern that can be used for registering descendants of any kind. Can we extract the intersection observer pattern to a separate hook or wrapper component instead? That would also give us smaller, more focused files

mattcosta7:
Thanks for the review! Addressed all the feedback:

**@iansan5653**
- **Extracted the IntersectionObserver pattern out of `descendant-registry`.** It now lives in its own focused files: `internal/hooks/useOverflowObserver.ts` (context + `useIsOverflowing` hook) and `internal/components/OverflowObserverProvider.tsx` (the shared, root-scoped observer). `descendant-registry` is back to a generic registry (~199 lines smaller). `UnderlineNav`/`ActionBar` now compose `<OverflowObserverProvider rootRef=…>` around the registry provider and items call `useIsOverflowing(ref)`.
- Dropped the out-of-context justification comment in `ActionBar`.

**Copilot**
- **Required clipping root:** `OverflowObserverProvider` now takes a required `rootRef` and stays inert (re-checking on later renders) until it's attached — it never falls back to a viewport-rooted observer that wouldn't reflect clipping.
- **`queueMicrotask` portability:** rebuild coalescing now goes through a `scheduleMicrotask` helper that falls back to `setTimeout` where `queueMicrotask` is unavailable.
- **Tests added:** new `useOverflowObserver.test.tsx` covers the inert paths (no provider, `IntersectionObserver` unavailable, `disabled`) and overflow detection via a mocked root-scoped observer; added a `descendant-registry` test asserting multiple same-tick registrations coalesce into one correct rebuild.

Type-check, lint (incl. `react-compiler`), and unit tests pass locally.

francinelucca (approved):
My head hurts but looks good to me 😅. Let me know if you want more eyes on it // a more minutious review

PR #8069, opened 2026-06-26
chore(npm): add allowScripts for npm v12 migration

joshblack:
With our move to Node.js v26, `npm` is updated to `11.16.0`. This introduces a new behavior, `allowScripts`, that will default to off in v12 of `npm`. As a result, this PR opts us into package scripts that we would like to run to support the project.


llastflowers (approved):
✨

PR #8070, opened 2026-06-28
Timeline: add Issue event stories (Phase 2)

janmaarten-a11y:
## Summary

Adds Storybook **Features** stories that recreate GitHub's live **Issue timeline events** using the Primer React `Timeline` component and its compositional slots. This is the **Issues surface** of Phase 2 of the Timeline events effort (github/primer#6663). The Pull Request, Shared, and Dependabot surfaces follow separately (Dependabot is in #8071).

The Issue examples live in `Timeline.issues.features.stories.tsx` (+ a `.module.css`), exported under the title **`Components/Timeline/Events/Issues`** — one story export per event group.

This surface uses the shared `UserActor` helper, which lives in `Timeline/internal/timelineStoryHelpers.tsx`. `UserActor` had been copy-pasted across every event surface; it composes the shared `InlineAvatar` + name primitives and takes props for the small per-surface differences (avatar size, link vs plain text, muted vs default color, and an `icon` for system actors like scanning's "GitHub"). **The shared helper now lives in its own PR (#8140)**, so this PR no longer hosts it and doesn't gate the others — once #8140 merges, this PR (and the other surfaces) simply import from it. (A local copy of `UserActor` is present here until #8140 lands, then removed on rebase.)

The file covers **17 event groups / 81 variants total**, split into the Issue-only events and the Issue versions of the shared events (shared events nest per-surface; their PR/Dependabot versions are built separately later).

### Issue-only events (9 groups, 59 variants)

| Group | Story export | Variants |
| --- | --- | --- |
| Closed | `EventClosed` | 7 |
| State | `EventState` | 6 |
| References | `EventReferences` | 4 |
| Duplicates | `EventDuplicates` | 4 |
| Moderation | `EventModeration` | 4 |
| Issue Types | `EventIssueTypes` | 3 |
| Issue Hierarchy | `EventIssueHierarchy` | 8 |
| Dependencies | `EventDependencies` | 8 |
| Issue Fields | `EventIssueFields` | 15 |

### Shared events — Issue versions (8 groups, 22 variants)

| Group | Story export | Variants |
| --- | --- | --- |
| Project | `EventProject` | 3 |
| Labels | `EventLabels` | 3 |
| Title | `EventTitle` | 1 |
| Milestones | `EventMilestones` | 2 |
| Assignments | `EventAssignments` | 6 |
| Lock / Unlock | `EventLockUnlock` | 3 |
| Comment deleted | `EventCommentDeleted` | 1 |
| Cross-references | `EventCrossReferences` | 3 |

Shared events are a shared *concept*, not shared code: the `github-ui` React `timeline-items` components are the **Issue** implementation, while the PR timeline renders the same events through a separate ERB path. Each shared group is built here from the React Issue components, and where the PR (ERB) render diverges (e.g. the Project link styling, the Assignments actor component) that divergence is documented in an inline comment so the future PR/Dependabot surface inherits the spec.

## Source of truth

Each event was recreated from the **live `github-ui` React implementation** (`packages/timeline-items`) as the canonical source, cross-checked against the Figma timeline audit and verified visually against the Figma components. Where the audit and live code disagreed — for example the Closed-event badge icons (`CheckCircle`/`CircleSlash`, not the audit's `IssueClosed`/`Skip`), the Unlinked-PR badge icon, the renamed-title copy (live has no "from"/"to" words), and the Unlocked badge icon (`UnlockIcon`, not `LockIcon`) — **live code won**. In each case we confirmed the live render path is actually reached (present in the timeline item map, not feature-flagged-off or superseded by a newer component) before trusting it.

## Slots used (Phase 1 slots)

- **Inline actor avatar** in `Timeline.Body` — a 20px avatar + bold link, matching how the live `github-ui` events render the actor. These badge-row events deliberately do **not** use the `Timeline.Avatar` gutter slot (github/primer#6677); that 40px gutter slot is reserved for comment-style events, which aren't part of this Issue-only set.
- **`Timeline.Actions`** (right-controls slot, github/primer#6678) for the Duplicates **"Undo"** button — the only right-aligned control in the Issue event set. Everything else (inline commit SHAs, PR-state icons, `#numbers`) is part of the sentence body text in the live render, so it correctly stays **inline in `Timeline.Body`**, not in `Timeline.Actions`.

## Scope & conventions

- Every story is wrapped in a file-scoped `FeatureFlags` decorator enabling `primer_react_timeline_list_semantics` (the future-state `<ol>`/`<li>` DOM, so the stories won't need re-authoring when that flag becomes the default; refs primer/react#7910).
- **Storybook-only by design.** These are intentionally **not** wired into components-json / the primer.style docs page (not added to `Timeline.docs.json` or `build.ts`) — individual timeline events are not consumer-facing components, and the primer.style Timeline page reflects the base `Timeline` component's own stories. Any docs-site representation is a Phase 3 consideration.
- The future event-category `data-*` filtering taxonomy (github/primer#6663) is intentionally **deferred**; a single documented hook location in the file header marks where it'll attach per `Timeline.Item`.

#### New

- `Timeline.issues.features.stories.tsx` and `Timeline.issues.features.stories.module.css` — Storybook Features stories recreating the Issue timeline events (17 groups, 81 variants).
- Uses the shared `UserActor` helper from #8140 (a local copy is present until #8140 merges, then removed on rebase).

#### Changed

- None.

#### Removed

- None.

### Rollout strategy

Stories-only change — no consumer-facing package code is added or modified, so no changeset is needed (the `skip changeset` label is applied).

### Testing & Reviewing

- prettier, eslint, and stylelint all clean; type-check passes with 0 new errors (the pre-existing `@primer/doc-gen` baseline error in `script/components-json/build.ts` is unchanged).
- The 9 Issue-only groups were visually diffed against the Figma source; where Figma or the audit disagreed with live code, the **live render won**. The 8 shared groups were sourced directly from the live `github-ui` React Issue components.
- Browse the stories under **Components/Timeline/Events/Issues** in Storybook to review each group.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge

### DIFF
--- a/packages/react/src/Timeline/Timeline.issues.features.stories.module.css
+++ b/packages/react/src/Timeline/Timeline.issues.features.stories.module.css
@@ -72,18 +72,12 @@
   color: var(--fgColor-muted);
 }
 
-.ProjectIcon {
-  vertical-align: middle;
-  margin-right: var(--base-size-4);
-  color: var(--fgColor-muted);
-}
-
-/* Issue-version ProjectV2 reference (github-ui `ProjectV2.tsx`): an inline
-   default-colored TableIcon octicon followed by a regular-weight, default-color
-   link. The underline comes from the Link `inline` prop (always-on underline,
-   which also satisfies the high-contrast a11y rule). Mirrors live
-   `.projectLink { color: var(--fgColor-default) }`. NOTE: the PR (ERB) surface
-   renders this differently — see the EventProject group comment. */
+/* Issue-version ProjectV2 reference (github-ui `ProjectV2.tsx` + `ClosedEvent.tsx`):
+   an inline default-colored TableIcon octicon followed by a default-color link
+   (`.ProjectRefLink`). The underline comes from the Link `inline` prop (always-on
+   underline, which also satisfies the high-contrast a11y rule). Mirrors live
+   `.projectLink`/`.stateReasonLink` (both `color: var(--fgColor-default)`). NOTE:
+   the PR (ERB) surface renders this differently — see the EventProject group. */
 .ProjectRefIcon {
   vertical-align: middle;
   margin-right: var(--base-size-4);
--- a/packages/react/src/Timeline/Timeline.issues.features.stories.module.css
+++ b/packages/react/src/Timeline/Timeline.issues.features.stories.module.css
@@ -1,0 +1,165 @@
+.RealisticTimeline {
+  /* GitHub renders the timeline at most 1012px wide in product surfaces. */
+  max-width: 1012px;
+}
+
+.LinkWithBoldStyle {
+  font-weight: var(--base-text-weight-semibold);
+  color: var(--fgColor-default);
+  margin-right: var(--base-size-4);
+}
+
+.LinkWithBoldStyle:hover {
+  color: var(--fgColor-accent);
+}
+
+.InlineAvatar {
+  vertical-align: middle;
+  margin-right: var(--base-size-4);
+}
+
+.Timestamp {
+  text-decoration: underline;
+}
+
+.CommitSha {
+  font-family: var(--fontStack-monospace);
+  font-size: var(--text-body-size-small);
+  font-weight: var(--base-text-weight-semibold);
+  color: var(--fgColor-default);
+}
+
+/* Story-only scaffolding: a small caption per Timeline.Item so the variants
+   can be scanned like a Figma component set. Not part of the faithful event. */
+.VariantLabel {
+  display: block;
+  margin-bottom: var(--base-size-4);
+  font-size: var(--text-body-size-small);
+  font-weight: var(--base-text-weight-semibold);
+  color: var(--fgColor-muted);
+  text-transform: uppercase;
+  letter-spacing: 0.04em;
+}
+
+/* IssueLink-style inline reference: state octicon + bold title + muted number. */
+.IssueLink {
+  color: var(--fgColor-default);
+  text-decoration: none;
+}
+
+.IssueLink:hover {
+  color: var(--fgColor-accent);
+  text-decoration: underline;
+}
+
+.IssueLinkIcon {
+  color: var(--fgColor-done);
+  vertical-align: middle;
+  margin-right: var(--base-size-4);
+}
+
+.IssueLinkTitle {
+  font-weight: var(--base-text-weight-semibold);
+}
+
+.IssueLinkNumber {
+  color: var(--fgColor-muted);
+}
+
+.ProjectIcon {
+  vertical-align: middle;
+  margin-right: var(--base-size-4);
+  color: var(--fgColor-muted);
+}
+
+/* Open (green) state icon variant for IssueLink references that point at an
+   open issue/PR (e.g. the canonical issue in a "marked as duplicate of" row). */
+.IssueLinkIconOpen {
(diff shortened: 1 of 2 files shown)


packages/react/src/Timeline/Timeline.issues.features.stories.tsx
```diff
@@ -0,0 +1,2538 @@
...
+    <RelativeTime date={new Date(date)} format="relative" />
+  </Link>
+)
+
+export default {
+  title: 'Components/Timeline/Events/Issues',
+  component: Timeline,
+  subcomponents: {
+    'Timeline.Item': Timeline.Item,
+    'Timeline.Avatar': Timeline.Avatar,
```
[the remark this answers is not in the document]
janmaarten-a11y: Good catch — the description was stale after the shared-event groups were added. I've updated the PR summary to reflect the current **17 event groups / 81 variants**, split into two tables (Issue-only events: 9 groups/59 variants, and Shared events — Issue versions: 8 groups/22 variants), and updated the changelog and testing notes to match.

packages/react/src/Timeline/Timeline.issues.features.stories.tsx
```diff
@@ -0,0 +1,2476 @@
...
+            <Time date="2022-07-23T11:05:00Z" />
+          </Timeline.Body>
+        </Timeline.Item>
+      </Timeline>
+    </section>
+
+    {/* Converted to discussion — github-ui's ConvertedToDiscussionEvent links
+          the resulting discussion by number. */}
+    <section className={classes.Variant}>
+      <h3 className={classes.VariantLabel}>Converted to discussion</h3>
```
adierkens: Would it help to abstract this Section wrapper into a local component?

packages/react/src/Timeline/Timeline.issues.features.stories.module.css
```diff
@@ -0,0 +1,175 @@
...
+
+.LinkWithBoldStyle {
+  font-weight: var(--base-text-weight-semibold);
+  color: var(--fgColor-default);
+  margin-right: var(--base-size-4);
+}
+
+.LinkWithBoldStyle:hover {
+  color: var(--fgColor-accent);
+}
```
adierkens: Is this not already a variant on the base link component?

janmaarten-a11y: `Link` today only has `inline`, `muted`, and `hoverColor`. There's no weight/emphasis variant, so there's no OOTB way to get this semibold + `fgColor-default` + accent-on-hover treatment.

It's not just this surface, and there are a few other gaps (such as actors/avatars) that _should_ be a real component affordance rather than something each surface hand-rolls. 

Filed it as [github/primer#6826](https://github.com/github/primer/issues/6826) to decide placement upstream; we'll drop this local class once that lands.

packages/react/src/Timeline/Timeline.issues.features.stories.tsx
janmaarten-a11y: Yep! Working on doing that now, along with the other story-only stuff in https://github.com/github/primer/issues/6826

adierkens (approved):

PR #8072, opened 2026-06-28
Timeline: add comment event stories (Phase 2)

janmaarten-a11y:
Part of github/primer#6663 (Phase 2).

Adds Storybook **Features** stories that recreate GitHub's timeline **comment cards** using Primer React, plus the shared **internal** helper module the whole Phase 2 stack builds on.

> **Stacked PR.** This is the bottom of the Phase 2 stack — it carries the shared `internal/` helpers that #8070, #8071, #8073, #8074, #8075 all import. Those five are based on this branch and should merge after it (GitHub auto-retargets them to `main` once this lands).

### 1. Comment card stories — `Components/Timeline/Internal/Comment cards`

`packages/react/src/Timeline/Timeline.comments.features.stories.tsx` (+ co-located CSS module). A single story (`CommentCards`, shown as **"Comment cards"**) renders **7 labeled sections**, each built from the internal `CommentCard` helper:

- **Standard user comment** — circle avatar, `Author` subject-author badge, reactions row
- **Bot comment (GitHub App)** — square photo avatar, `bot` label
- **Copilot comment** — octicon avatar, `AI` label, name renders as "Copilot"
- **Dependabot comment** — square accent octicon avatar (white `DependabotIcon` on brand blue), `bot` label
- **User comment via a GitHub App** — the "– with {app}" timestamp suffix
- **User comment via an app (with app avatar badge)** — the parent-child avatar (small app avatar overlapping the bottom-right of the author avatar)
- **Comment (inline avatar / compact)** — the gutter-less inline-avatar form (see Responsive below)

### 2. Shared internal helper module — `Timeline/internal/`

Extracted so the six surface stories consume components instead of copy-pasting primitive clusters (resolves review feedback about baking helpers into the story):

- `internal/timelineStoryHelpers.tsx` (+ `.module.css`) — `VariantSection`, `RealisticTimeline`, `Examples`, `BoldLink`, `InlineAvatar`, `MutedTime`. The last three reproduce live-GitHub styling Primer doesn't express natively yet and are tracked to remove upstream (github/primer#6826 bold inline link, #6827 inline avatar, #6828 muted timestamp).
- `internal/CommentCard.tsx` (+ `.module.css`) — the self-contained comment-card composition (header, body, reactions, caret, octicon/parent-child avatars, responsive rules). Internal only, not exported from the package index.
- `internal/timelineStoryHelpers.stories.tsx` — a `Components/Timeline/Internal/Helpers` story documenting the six shared helpers in isolation.

[...]

PR #8073, opened 2026-06-28
Timeline: add Secret Scanning event stories (Phase 2)

janmaarten-a11y:
## Summary

Adds Storybook **Features** stories that recreate GitHub's live **Secret Scanning alert timeline** events using the Primer `Timeline` component and its compositional slots. This is the **Secret Scanning surface** — the first **Security** surface — of Phase 2 of the Timeline events effort (github/primer#6663). The Issues surface (#8070), Issue comment cards (#8072), and the Dependabot alert surface (#8071) already landed; **Code Scanning** and **License Compliance** will follow as separate Security surfaces.

Unlike Dependabot (which is server-rendered ERB), **Secret Scanning is fully React** and its live timeline already composes Primer React `Timeline` + `Timeline.Badge variant=`, so the events map cleanly — the badge colors, icons, and actor treatments translate almost one-to-one.

All examples live in a single new file, `Timeline.secret-scanning.features.stories.tsx` (+ a `.module.css`), exported under the title **`Components/Timeline/Events/Secret Scanning`** — one story export per event group.

### The 7 event groups (29 variants total)

| Group | Variants | Badge(s) | Actor |
|---|---|---|---|
| `EventCreated` | 1 — Created | `shield` on solid `success` (green) | GitHub (system) |
| `EventResolution` | 8 — 7 closed reasons + Reopened | `shield-check` on solid `done` (revoked) / bare `shield-slash` (other 6) / `sync` on solid `success` (reopened) | User |
| `EventBypass` | 3 — Bypassed / requested / approved | bare `alert` / `comment` / `check-circle` | User |
| `EventValidityChange` | 6 — active/inactive/unknown × automated/manual | `alert` on solid `danger` (active) / bare `skip` (inactive) / `alert` on solid `attention` (unknown) | GitHub (automated) / User (manual) |
| `EventReport` | 1 — Reported | bare `shield-check` | User |
| `EventClosureRequest` | 5 — requested (reviewer + requester views) / approved / denied / cancelled | bare `comment` / `check-circle` / `x` / `skip` | User |
| `EventAssignment` | 5 — self-assigned / assigned other / removed own / unassigned other / assigned+unassigned | bare `person` | User (+ assignee avatars) |

## Source of truth

These stories were recreated from the **live React implementation** in `github/github-ui`, package `packages/secret-scanning-alerts/`. The authoritative dispatch is the `switch (event.type)` in `components/show/AlertTimeline.tsx`; per-event copy/badges/actors were read directly from it and its helpers (`helper.ts`, `components/shared/UserComponent.tsx`, `components/show/closures/ClosureRequestReviewButtons.tsx`) and translated faithfully into Primer React. There is **no ERB secret-scanning timeline** — the migration to React is complete.

We verified the live render path rather than trusting the Figma audit blindly. The notable resolutions:

- **The `REVOCATION` event type is dead code.** `TimelineEventType.Revocation = 'REVOCATION'` exists in the enum but has **no `case`** in the switch, so it renders nothing — it is intentionally **not built**.
- **The 7 closed resolution reasons come from live `resolutionText()`** (revoked, false positive, won't fix, used in tests, pattern deleted, pattern edited, ignored by configuration). Only **revoked** gets the special purple `done` / `shield-check` badge; the other six use the bare default badge with `shield-slash`. Reasons render as **bold text, not links**.
- **Report is attributed to the user**, not the GitHub system actor — the `Report` case does **not** pass `isGitHubActor` (a point the audit flagged as uncertain).
- **The closure-request right-controls are "Review request" (primary) and "Cancel request" (invisible)** — confirmed from `ClosureRequestReviewButtons.tsx`. They are driven by two independent, viewer-specific payload flags (`show_closure_request_review_buttons` for reviewers, `show_closure_request_cancel_button` for the requester), so they are mutually exclusive per viewer; both are shown as separate variants.
- **The system "GitHub" actor** (`MarkGithubIcon` + bold "GitHub", no avatar) is used by Creation unconditionally and by **automated** validity changes (`!event.actor`); manual validity changes and every other event use the user actor.

## Slots used (Phase 1 slots)

- **Inline actor avatar in `Timeline.Body`** for these badge-row events — a 16px **circle** user avatar + bold login (`UserActor`, mirroring live `UserComponent`), or the system `GitHubActor` (`MarkGithubIcon` + bold "GitHub", **no avatar**). Badge-row events do not use the gutter `Timeline.Avatar`.
- **`Timeline.Actions`** (right-controls slot, added in #7886) for the **closure-request "Review request" / "Cancel request"** buttons.
- **`Timeline.Break`** is rendered immediately **before** the Reopened resolution variant, mirroring the live sibling-selector placement.
- A shared, file-local **comment sub-row** helper (`CommentSubRow`: 12px muted `comment` icon + small text) renders the optional `resolution.comment` / requester / reviewer comment under Resolution closures and the delegated-closure events.

## Badge note

Events whose live code passes a solid variant (`success` / `done` / `danger` / `attention`) use `Timeline.Badge variant=` directly. Events whose live code passes **`variant={undefined}`** (all the "gray" events) render as a **bare `Timeline.Badge`** — a muted icon on a borderless circle with **no solid fill** — matching the Dependabot precedent (#8071), **not** the Issues `--timelineBadge-bgColor` solid-gray hook. None of the secret-scanning gray events is a solid-gray badge.

## Accessibility

Secret Scanning renders **no in-text links** — the live `UserComponent` renders the login as a bold `text-bold` span (hovercard attrs only, not a profile link), and resolution reasons are bold text. So the axe `link-in-text-block` high-contrast rule (which required the `inline`/bold treatment on the Dependabot surface) is **never exercised** here. The header doc notes that any future in-text link must still use `inline`/bold.

## Scope & conventions

- Each variant is wrapped in a story-only `<section>` with a small caption `<h3>` **above** the event, so the event row itself renders exactly as it would in product (the caption never sits inside `Timeline.Body`).
- Every story is wrapped in a file-scoped `FeatureFlags` decorator enabling `primer_react_timeline_list_semantics` (the future-state `<ol>`/`<li>` DOM; refs primer/react#7910).
- The feature-gated delegated groups (delegated bypass, delegated closure) carry a brief inline note that they only render when the org feature is enabled.
- **Storybook-only by design.** These are intentionally **not** wired into components-json / the primer.style docs page (not added to `Timeline.docs.json` or `build.ts`) — individual timeline events are not consumer-facing components.
- The future event-category `data-*` filtering taxonomy (github/primer#6663) is intentionally **deferred**; the hook location is documented in the file header.

#### New

- `Timeline.secret-scanning.features.stories.tsx` and `Timeline.secret-scanning.features.stories.module.css` — Storybook Features stories recreating the Secret Scanning alert timeline events (7 groups, 29 variants).

#### Changed

- None.

#### Removed

- None.

### Rollout strategy

Stories-only change — no consumer-facing package code is added or modified, so no changeset is needed (the `skip changeset` label is applied).

### Testing & Reviewing

- prettier, eslint, and stylelint all clean; type-check passes with 0 new errors (the pre-existing `@primer/doc-gen` baseline error in `script/components-json/build.ts` is unchanged).
- Every group was verified against the live `AlertTimeline.tsx` switch and its helpers; where the Figma audit was uncertain (dead `REVOCATION`, closure-reason set, Report attribution, Review/Cancel buttons), the **live code won**.
- Browse the stories under **Components/Timeline/Events/Secret Scanning** in Storybook to review each group.

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


packages/react/src/Timeline/Timeline.secret-scanning.features.stories.tsx
```diff
@@ -0,0 +1,803 @@
...
+            <Octicon icon={SyncIcon} aria-label="Reopened" />
+          </Timeline.Badge>
+          <Timeline.Body>
+            <UserActor />
+            {'reopened this '}
+            <Time date="2022-07-27T12:00:00Z" />
+          </Timeline.Body>
+        </Timeline.Item>
+      </Timeline>
+    </section>
```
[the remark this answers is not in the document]
janmaarten-a11y: Good catch — fixed in 2a80e64. The Reopened variant now renders the preceding closed `Timeline.Item` before the `Timeline.Break`, so the break sits between two items (mirroring the live break-between-events placement) instead of being a stray first child.

packages/react/src/Timeline/Timeline.secret-scanning.features.stories.tsx
```diff
@@ -0,0 +1,803 @@
...
+ * `<Link>` — actors and resolution reasons are BOLD TEXT (the live
+ * `UserComponent` uses a `text-bold` span, not a profile link, and reasons are
+ * bolded plain text). So the axe `link-in-text-block` rule (which failed the
+ * Dependabot CI for underline-less in-text links) is never exercised by this
+ * surface. Any future in-text link added here must use `inline`/bold.
+ */
+
+const MONALISA_AVATAR = 'https://avatars.githubusercontent.com/u/583231?v=4'
+const SIX7_AVATAR = 'https://avatars.githubusercontent.com/six7'
+const HUBOT_AVATAR = 'https://avatars.githubusercontent.com/hubot'
```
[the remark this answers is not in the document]
janmaarten-a11y: Fixed in 2a80e64 — switched the demo assignee avatars to the `u/{id}?v=4` form to match the convention used elsewhere in the Timeline stories.

packages/react/src/Timeline/Timeline.secret-scanning.features.stories.tsx
```diff
@@ -0,0 +1,783 @@
...
+const SIX7_AVATAR = 'https://avatars.githubusercontent.com/u/4548309?v=4'
+const HUBOT_AVATAR = 'https://avatars.githubusercontent.com/u/480938?v=4'
+
+/**
+ * System "GitHub" actor — live `TimelineItemBody` (`AlertTimeline.tsx`) in
+ * `isGitHubActor` mode renders `<MarkGithubIcon /> <span class="ml-1
+ * text-bold">GitHub</span>` (no avatar). Used by the Created event and by
+ * automated validity changes.
+ */
+const GitHubActor = () => (
```
adierkens: Should this be pulling from the new `/internal` components?

janmaarten-a11y: Good catch, we missed some of the helpers in our last sweep of these stories. This one is also broader than this file. `UserActor` and this `GitHubActor` were copy-pasted across five surfaces, so we pulled them into one shared `UserActor` in `/internal` with props for the small per-surface differences (avatar size, link vs plain text, muted vs default color). `GitHubActor` is just that with a glyph instead of a photo, so it takes an optional `icon` prop now.

It lands in #8070 first, then this PR and the others swap to the shared import. (Dependabot already merged, so it gets a small follow-up.)

janmaarten-a11y: Correction: I pulled the shared `UserActor` and all the other shared helpers into their own dedicated PR (#8140) rather than landing them in #8070, so none of this depends on #8070 anymore. Once #8140 merges, this PR swaps `GitHubActor` and `UserActor` over to the shared import.

adierkens (approved):

PR #8082, opened 2026-06-29
ci: enable cron for environment clean up

joshblack:
Updates our environment clean up command so that it runs every night at midnight


jonrohan (approved):
:jinx:

PR #8088, opened 2026-06-29
docs: Remove merge checklist from PR template

joshblack:
Updates our PR template so that it no longer includes the merge checklist. Curious what others think 👀 This is a section that I find myself not really using too often so was curious if others felt the same and if we could remove it


llastflowers (approved):
I agree! 🚀

PR #8097, opened 2026-07-01
Add trigger ARIA state for detached AnchoredOverlay/SelectPanel anchors

joshfarrant:
## Summary

In **detached-anchor mode** (`renderAnchor={null}`, where the consumer owns and renders the trigger element), `AnchoredOverlay` never applied the `aria-haspopup`/`aria-expanded` attributes it normally sets in the `renderAnchor` path. Screen reader users had no indication that the trigger opens a popup, nor whether it's currently expanded.

Fixes the accessibility gap described in github/primer#6776.

## What changed

- **`AnchoredOverlay`** — imperatively reflects `aria-haspopup` and `aria-expanded` onto the detached anchor node, mirroring the existing `anchor-name` effect that already writes to that consumer-owned element. The effect:
  - reads `anchorRef.current` (attached by commit time) so the **collapsed** trigger is labelled on mount — not just after the overlay is first opened;
  - **guards against clobbering** a consumer-supplied `aria-haspopup`;
  - reflects `aria-expanded` from `open`, and **cleans up** on close/unmount.
- **New `anchorHasPopup` prop** — `'true' | 'dialog' | 'menu' | 'listbox' | 'tree' | 'grid'`, default `'true'` (non-breaking; `'true'` is spec-equivalent to `'menu'`). Applied in **both** the `renderAnchor` and detached paths so the popup role can be described accurately.
- **`SelectPanel`** passes `anchorHasPopup="dialog"` to match its `role="dialog"` popup (previously it hardcoded the spec-equivalent-of-menu default, which mis-described the dialog).

## Before / after

| | Detached trigger (collapsed) | Detached trigger (open) |
|---|---|---|
| **Before** | _no `aria-haspopup`, no `aria-expanded`_ | _no ARIA state_ |
| **After** | `aria-haspopup="…"` `aria-expanded="false"` | `aria-expanded="true"` |
| **SelectPanel** | `aria-haspopup="dialog"` | `aria-expanded="true"` |

## Testing

Added coverage on `AnchoredOverlay` (collapsed state, open/close toggle, consumer-override, `renderAnchor` path, cleanup) and `SelectPanel` (`aria-haspopup="dialog"` on both default and detached anchors)

### DIFF
--- a/packages/react/src/AnchoredOverlay/AnchoredOverlay.docs.json
+++ b/packages/react/src/AnchoredOverlay/AnchoredOverlay.docs.json
@@ -170,6 +170,13 @@
       "defaultValue": "true"
     },
     {
+      "name": "anchorHasPopup",
+      "type": "'true' | 'dialog' | 'menu' | 'listbox' | 'tree' | 'grid'",
+      "required": false,
+      "description": "Indicates the type of popup opened by the anchor. Defaults to `'true'`, which is equivalent to `'menu'` in ARIA.",
+      "defaultValue": "'true'"
+    },
+    {
       "name": "closeButtonProps",
       "type": "Partial<ButtonProps>",
       "required": false,
--- a/.changeset/anchored-overlay-detached-anchor-aria.md
+++ b/.changeset/anchored-overlay-detached-anchor-aria.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+AnchoredOverlay and SelectPanel: add trigger ARIA state for detached anchors
--- a/packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
+++ b/packages/react/src/AnchoredOverlay/AnchoredOverlay.test.tsx
@@ -3,6 +3,7 @@ import {describe, expect, it, vi} from 'vitest'
 import {act, fireEvent, render} from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import {AnchoredOverlay} from '../AnchoredOverlay'
+import type {AnchorHasPopup} from '../AnchoredOverlay'
 import {Button} from '../Button'
 import BaseStyles from '../BaseStyles'
 import type {AnchorPosition} from '@primer/behaviors'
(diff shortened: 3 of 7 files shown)


joshfarrant:
Addressed both review comments in 491679e08:
- `aria-haspopup` and `aria-expanded` are now managed in separate effects, so the stable `aria-haspopup` is no longer removed/re-added on every open/close.
- `AnchorHasPopup` is now derived from `React.AriaAttributes['aria-haspopup']` (excluding `boolean`/`'false'`) and exported; the tests reference it instead of duplicating the union.

joshfarrant:
Addressed in 98484d956: added `anchorRef` to the dependency arrays of both detached-anchor ARIA effects and removed the `exhaustive-deps` suppressions. `anchorRef` is stable in the common case, so no extra re-runs, but the effects now honestly track a swapped `anchorRef` prop.

francinelucca (commented):
This looks good to me implementation-wise, and satisfies https://github.com/github/primer/issues/6776. I kind of feel like if the consumer is choosing to use a custom anchor then they should also be responsible/empowered for handling the accessibility and attributes of it however they see fit 🤔

Added some more thoughts here https://github.com/github/primer/issues/6776#issuecomment-4861619362

packages/react/src/AnchoredOverlay/AnchoredOverlay.docs.json
```diff
@@ -169,6 +169,13 @@
       "description": "If true, a close button will be displayed in the top right corner of the overlay when on fullscreen mode.",
       "defaultValue": "true"
     },
+    {
+      "name": "anchorHasPopup",
+      "type": "'true' | 'dialog' | 'menu' | 'listbox' | 'tree' | 'grid'",
+      "required": false,
+      "description": "Indicates the type of popup opened by the anchor. Defaults to `'true'`, which is equivalent to `'menu'` in ARIA.",
+      "defaultValue": "'true'"
+    },
```
siddharthkp, line 178: I have thoughts on the API, but it's a a good to have for the bug fix we are doing here and I don't want to block the bug fix.

Can you split this PR into 2:
1. the first PR is the bug fix that copies current renderAnchor behavior (value: true)
2. second PR lets you customize the value of aria-haspopup, so that we can discuss it there. 

Thanks!

siddharthkp (changes requested):

PR #8098, opened 2026-07-01
Reproduction for Tooltip conditional-wrap focus loss

siddharthkp:
Closes #

Adds two Storybook **dev** stories under `Components/TooltipV2/Dev` that reproduce a keyboard focus-loss accessibility bug and demonstrate two fixes.

The bug: conditionally wrapping a `Button` in a `Tooltip` (e.g. `cond ? <Tooltip><Button/></Tooltip> : <Button/>`) swaps the top-level element type at that slot. React reconciles by type per position, so toggling the condition unmounts and recreates the underlying `<button>`. If that button had keyboard focus when it was activated, focus drops to `<body>` after the state flip.

#### New

- `ConditionalTooltipWrap` dev story: reproduces the focus loss (Buggy) and shows two fixes side by side, with a live "current focus" indicator.
  - Fix A (stable tree): always render the `Tooltip` and only change its `text`, so the button is never remounted.
  - Fix B (restore focus): accept the remount and restore focus imperatively via a ref + `flushSync`. Documents that `Tooltip` overrides its child's ref via `cloneElement`, so the ref must go on the `Tooltip` in the wrapped branch.
- `CheckToActivate` dev story: reproduces the same swap driven by checking at least two checkboxes.

#### Changed

- None (dev stories only).

#### Removed

- None.

### Rollout strategy

Dev-only Storybook stories. No changes to the published `@primer/react` API, so no release is needed.

### Testing & Reviewing

1. Run Storybook and open `Components/TooltipV2/Dev`.
2. In `ConditionalTooltipWrap`, keyboard-focus the first "Save" button and press Enter. Watch the "Current focus" readout drop to `body` in the Buggy row, while Fix A and Fix B keep focus on the button.
3. In `CheckToActivate`, check two items to flip the button from inactive to active and observe the same focus behavior.

### Merge checklist

- [ ] Added/updated tests
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


siddharthkp:
This bug still exists, so let's keep the issue

PR #8108, opened 2026-07-02
`UnderlineNav`: Improve flickering by moving icon visibility control to a one-way CSS animation

iansan5653:
- Resolves https://github.com/github/primer/issues/6806

The icons in `UnderlineNav` can cause flickering as they cause/un-cause the container to overflow when hidden/shown. We use JS to update state to prevent this flickering, but that doesn't run until after hydration. This means that, for certain screen sizes, the icons will flicker after SSR and before client-size hydration.

This PR moves the icon visibility logic to a one-way CSS animation on supported browsers. This way, when the scroll-driven animation is triggered, it triggers the corresponding icon-hiding animation, which persists and keeps the icons hidden for the remaining life of the component.

This improves the performance significantly by shifting the calculations into CSS in supported browsers (ie, not in Firefox). This reduces the visibility of the flickering. However, unfortunately it does not completely resolve the flickering; the browser still has to paint the tabs in order to determine the overflow and then hide the icons.

#### Changed

- Fixed icon flickering before initial render in `UnderlineNav`

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Merge checklist

- [ ] Added/updated tests
- [ ] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


joshblack (approved):

packages/react/src/UnderlineNav/UnderlineNav.module.css
```diff
@@ -21,12 +16,32 @@
...
+
+@keyframes hide-icons {
+  0% {
+    display: inline;
+  }
+
+  100% {
+    display: none;
+  }
 }
```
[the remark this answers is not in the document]
iansan5653: I'm like 99% sure this is incorrect. Browsers will animate the property but (obviously) not smoothly; they'll discretely step from one value to the next.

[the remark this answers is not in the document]
jonrohan: Checking https://primer-770a2beee9-13348165.drafts.github.io/storybook/?path=/story/components-underlinenav-features--with-icons I don't think the icons are hiding

packages/react/src/internal/components/UnderlineTabbedInterface.module.css
```diff
@@ -18,18 +18,14 @@
...
-
   &[data-overflow-mode='wrap'] {
     /* Wrap items onto hidden next lines */
     overflow: hidden;
     max-height: var(--control-xlarge-size);
 
     .UnderlineItemList {
       flex-wrap: wrap;
-      overflow: hidden;
+      flex: 1;
```
iansan5653, line 28: `overflow: hidden` had no effect here because there's no explicit height set on this container. And it's unnecessary since the parent already has that applied.

`flex: 1` fixes a separate bug where sometimes the button wouldn't float all the way to the right. It makes the tab containers take up more space to push the overflow button further out. This prevents it from looking like it's floating in space.

packages/react/src/UnderlineNav/UnderlineNav.module.css
```diff
@@ -2,15 +2,13 @@
   /* Progressive enhancement: Detect overflow using scroll-based animations.
   The idiomatic way would be a scroll-state container query but browser support
   is slightly better for animations. */
-  animation: detect-overflow linear;
-  animation-timeline: scroll(self block);
+  animation:
+    detect-overflow linear,
+    delay-visibility forwards 0ms 5ms;
```
iansan5653: When testing on Chrome with 20x CPU slowdown, I could see an initial paint where the icons are visible, followed by a quick update to hide them. I tried several things to resolve this, including conditionally setting the animation duration to 0 seconds and also conditionally changing the start state of the animation, but they all caused unexpected side effects.

I actually think this initial flicker may be entirely unavoidable, because the browser _has_ to paint the tabs to determine if overflow happens in order to determine whether to hide the icons or not. So I just accepted it and decided to hide the tabs with a very short animation, which delays their initial render by an almost imperceptible 5 ms. This is just long enough to get that initial paint out of the way.

packages/react/src/UnderlineNav/UnderlineNav.module.css
```diff
@@ -21,12 +19,45 @@
...
+  /* Unlike the more button, the icons are removed from the layout when hidden. This can cause the container to no
+  longer overflow, which would cause a flickering loop if we just drove the visibility directly from the scroll state.
+  So instead we drive an animation that can only play forwards so the icons stay hidden for the life of the page, even
+  if the container is no longer overflowing. We can't just make the scroll-driven animation itself play forwards,
+  because scroll-driven animations are conditionally applied and revert when not scrollable. */
+  animation-name: hide-icons;
+  animation-fill-mode: forwards;
+  animation-timing-function: linear;
+  animation-play-state: var(--UnderlineNav_hide-icons-play-state);
+  animation-duration: 0.1ms; /* must be greater than 0 */
```
iansan5653, line 37: Interestingly, this can't be arbitrarily small; `0.0001ms` is treated as `0` in Chrome. `0.1ms` seems to work very consistently though.

I did try making the _duration_ dynamic, instead of the play state, and setting the duration to `0` to immediately resolve the animation when I want to toggle it. That didn't work though; the style didn't 'stick' if the duration was increased again.

packages/react/src/UnderlineNav/UnderlineNav.module.css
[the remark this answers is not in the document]
iansan5653: This turned out to be due to not setting an initial value for the variable. Which feels like a browser bug, but it's easy to solve by initializing it.

packages/react/src/UnderlineNav/UnderlineNav.module.css
iansan5653:
```suggestion
    delay-visibility forwards 100ms 0;
```

iansan5653:
Closing in favor of an alternative approach https://github.com/primer/react/pull/8133

PR #8012, opened 2026-06-17
docs: add modular design system adr

joshblack:
Introduces an ADR as a follow up to our Modular Design System Workshop. This ADR defines the abstraction model we'll use for component development moving forward. Certain deliverables, like what components we are breaking up or what base components we would like to ship, have been moved to pitches over in github/primer.

This ADR also includes initial agents, skills, and additions to our style document that reflect the spectrum of abstraction being proposed.

PR #8113, opened 2026-07-03
Document the modular DS model as skills

joshfarrant:
#8012 established the spectrum of abstraction model — config, presentational, base, and utility components — and shipped two agents to work with it, with their guidance inlined and `style-guide` as their only skill. This PR adds six `modular-ds-*` skills, fills in the operational detail the ADR deliberately left out, and wires them into both agents.

| Skill | Covers |
|---|---|
| `modular-ds-spectrum-model` | Orientation. How the four API types relate, and how to decide which ones a component needs |
| `modular-ds-utilities` | Compound behaviour hooks, prop-getters, the controlled component contract, where utilities live |
| `modular-ds-base-components` | Unstyled primitives, the `:where()` reset convention, which parts warrant a base equivalent |
| `modular-ds-presentational-components` | Composition rules, `data-component`, RSC-safe export shape |
| `modular-ds-config-components` | What a config component owns, its customisation surface, adoption surfaces |
| `modular-ds-accessibility-contract` | Responsibility matrix across API types, worked Dialog and Tabs examples |

Most of the content has been ported from the original ADR branch (#7778) — `layer-patterns.md` and `accessibility-contract.md` — reframed from the four numbered layers to the four named API types, and restructured from two large documents into granular skills.

`contributor-docs/style.md` is the source of truth throughout. Two things the model leans on lived only in ADR-024's Questions section — that not every component needs every API type, and the signal for when to extract state into a hook — so this PR moves both into style.md's spectrum of abstraction section and drops the skills' references to the ADR. That's the only change to style.md, and it's the last ADR-024 reference in `.github/`.

github/primer#6851 covers the component creation and contribution process and lists several questions as open. This PR touches one of them, which I'd rather surface here than have found in review:

- **Render props vs plain children.** `modular-ds-presentational-components` prefers ordinary children and calls out `cloneElement` as fragile, while leaving render props available as a legitimate config-level extension point, matching `contributor-docs/style.md`.

Where components live and how they're exported is deliberately *not* answered here. The skills state where things sit today and treat adding a new export subpath as a package-level decision to escalate, so a component change shouldn't be what lands a new public entry point. github/primer#6773 is where the actual decision belongs.

## Validating these changes

I asked Copilot to build a component using only these skills, then checked the result against the model and fixed whatever the skills got wrong. Components were deliberately chosen to avoid the ones the skills use as worked examples — Accordion, Toolbar and Breadcrumbs.

After three iterations the results were coming up clean and aligned with the model.

## Reviewing these changes

From my testing, Copilot pays particular attention to code examples in the skills, so I'd appreciate more eyes on those to make sure they align with our expectations.

Closes github/primer#6789


lesliecdubs:
Hi @joshfarrant, should we close this now that we've merged [#3046](https://github.com/primer/react/pull/8012/changes)?

joshfarrant:
Hey @lesliecdubs 👋

#8012 shipped the model and the two agents, this PR adds `modular-ds-*` skills called for in github/primer#6789 and wires them into both agents. I've also stress-tested the agent/skills across a few iterations, so I think there are some changes here worth landing. I'd appreciate your thoughts though 🙏

.github/agents/modular-ds-implementer.agent.md
```diff
@@ -1,77 +1,72 @@
...
+
+New base components and utilities introduced under this model should be opt-in via their own entry point rather than added to the root barrel. This is a forward-looking convention, not a description of the current state — the root entry already exports a number of hooks, and existing exports shouldn't be moved as a side effect of unrelated work. All API types ship in one package version; stability is per-component (e.g. a hook can graduate to stable while its base component remains experimental).
+
+Create or update `index.ts` files to re-export the public API for each API type touched, and update the relevant experimental barrel files.
+
+Neither `foundations` nor `hooks` exists as an **export subpath** today, and adding one is a package-level decision rather than a step in building a component — **surface it and stop**. See "Where utilities live" in `modular-ds-utilities`, which is the single source of truth for this rule and says what to do instead.
+
+The **Utilities** row above refers to generic, component-agnostic utilities. A component's own compound hook is not one of those — it ships from wherever that component's parts ship, per `modular-ds-utilities`.
```
joshblack: Sadly I don't think we know where the base components/hooks/etc should live yet :/ I think if we pick up: https://github.com/github/primer/issues/6773 we could figure out what makes sense (e.g. `@primer/react/foundations`, `@primer/react/base`, or just `Base*` components, whatever) but might help to leave this out for now.

.github/skills/modular-ds-decompose-existing-component/SKILL.md
joshblack, line 1: This one is difficult to comment on for me since we're not really going through this exercise yet of decomposing/breaking down a component. Could this be one we save until we have this scenario or do we want to include it now?

.github/skills/modular-ds-spectrum-model/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Modular DS — Spectrum of Abstraction Model
+
+Primer authors components across a spectrum of abstraction rather than a single fixed shape. Start with flexible presentational components and companion behavior hooks. As opinions or defaults become established, build config components that compose those presentational components and hooks instead of duplicating their behavior.
+
+This is deliberately **not** a mandatory layer stack every component must populate. Not every component needs its own config API, and not every component needs its own hooks — start with presentational components and add the other API types only when there's real demand for the control they expose.
+
+## Background
+
+This model is documented in [ADR-024](../../../contributor-docs/adrs/adr-024-modular-ds.md) (`contributor-docs/adrs/adr-024-modular-ds.md`), which carries the full context, rationale, and open questions behind the rules below. `contributor-docs/style.md` covers the same model in shorter, human-facing form alongside Primer's wider conventions.
```
joshblack: It'd be great if we kept the source of truth for this in style.md as that will be the artifact that is changing over time. I think an idea with the style guide skill is to have topics live in there and bring it down progressive-disclosure style to make looking up relevant topics easier.

Let me know what you think 👀

.github/skills/modular-ds-tdd-a11y-test-backfill/SKILL.md
joshblack, line 1: I feel similar to the other decomposition skill where it's difficult to evaluate / see how it fits without us going through the work. Let me know if there are good before/after examples for what this influences though because if it helps out then makes sense to include 👍

joshblack (commented):
Thanks for putting this together! Really appreciate it.

Just left some comments/questions, let me know what you think 👀

.github/agents/modular-ds-implementer.agent.md
joshfarrant: Cool that makes sense, I've dropped this section for now

.github/skills/modular-ds-decompose-existing-component/SKILL.md
joshfarrant: Happy to drop it for now. I've had some success with component decomposition whilst testing these skills and when testing as part of previous work in this area + the workshop, but happy to keep things focused on new components for now

.github/skills/modular-ds-spectrum-model/SKILL.md
joshfarrant: Yea that makes sense, I've removed the reference to the ADR and consolidated some of its info into style.md instead.

.github/skills/modular-ds-tdd-a11y-test-backfill/SKILL.md
joshfarrant: That's fair, I'm happy to park this for now

joshblack (approved):

PR #8116, opened 2026-07-06
perf(useAnchoredPosition): cache scrollable-ancestor walk per anchor

mattcosta7:
Closes #

`useAnchoredPosition` (used by every Overlay, AnchoredOverlay, menu, tooltip, and popover) attaches scroll listeners to the anchor's scrollable ancestors. The setup effect lists `updatePosition` in its dependencies, so it **re-walked the entire ancestor chain — calling `getComputedStyle` on every ancestor — each time a consumer's reposition dependencies changed**, even though the set of scrollable ancestors only depends on the anchor element.

This caches the walk keyed on the anchor element and reuses it when the anchor is unchanged (the common case). Listener attach/detach behavior is unchanged; only the redundant `getComputedStyle` walk is skipped.

#### Changed

- `useAnchoredPosition` caches its scrollable-ancestor walk per anchor element, avoiding repeated `getComputedStyle` ancestor traversals (and the style recalc they force) when the reposition dependencies change while an overlay is open.

### Performance

Each avoided walk (a `getComputedStyle` per ancestor) costs roughly, at 1× CPU:

| Ancestor depth | Cost per walk |
|---|---|
| 10 | ~4.5 µs |
| 25 | ~8 µs |
| 50 | ~12.5 µs |

The win scales with DOM depth and with how often the overlay's consumer changes reposition dependencies while open; more importantly it removes a forced style recalc from that path during interaction.

### Rollout strategy

- [x] Patch release

### Testing & Reviewing

- 53 tests pass across `useAnchoredPosition`, `Overlay`, and `AnchoredOverlay` suites; eslint/prettier clean.
- Behavior is preserved: the same scroll listeners are attached/removed on each effect run; only the ancestor DOM walk is memoized by anchor identity.
- **Caveat for reviewers:** the cache is keyed on anchor-element identity, so if an *ancestor's* `overflow` style changed while the same overlay stayed open, the cached ancestor list could be momentarily stale (the previous code only caught this incidentally too). This is a high-blast-radius core hook — please run the **github/github-ui integration suite** to confirm overlay/menu/tooltip repositioning on scroll is unaffected.

### Merge checklist

- [x] Added/updated tests
- [x] Changes are SSR compatible
- [x] Tested in Chrome
- [ ] Tested in Firefox
- [ ] Tested in Safari
- [ ] Tested in Edge


mattcosta7:
> Integration test results from [github/github-ui PR](https://github.com/github/github-ui/pull/25847):
> 
> Waiting  CI  Waiting Passed  VRT  [Passed](https://github.com/github/github-ui/actions/runs/28890304757) Running  Projects  [Running](https://github.com/github/github-ui/actions/runs/28882948052)

these all passed, but aren't reporting right?

siddharthkp (approved):
Trust you to thoroughly test these!

siddharthkp:
> these all passed, but aren't reporting right?

yeah github-ui sometimes does that, I'm not sure why: https://github.com/github/github-ui/pull/25847#issuecomment-4906163549

jonrohan (approved):

PR #8133, opened 2026-07-07
`UnderlineNav`: Replace dynamic icon visibility logic with hardcoded breakpoints

### LINKED ISSUE #7823: UnderlineNav performance optimizations
### Description

👋🏼 Related to https://github.com/github/pull-requests/issues/24571 and https://github.com/primer/react/issues/7801, I've been investigating more performance optimizations. I think there are a few more things we could do with `UnderlineNav` to help.

### Diagnosis

After a window resize, `UnderlineNav` measures its own width to determine which tabs fit vs. overflow into a "more" menu. This triggers a synchronous `setState()` that re-renders the component and all its children — **385 child component renders** (`ExtendedLink` ×133, `ForwardRef` ×128, `Link` ×124) producing a **20.5ms React render** on each resize event.

Because `UnderlineNav` mutates the DOM during this render, four downstream components detect layout changes and each call `setState()` synchronously mid-commit, cascading one after another:

1. `StackState.setState()` → 2. `PullRequestHeader.setState()` → 3. `Overlay.setState()` → 4. `ActionMenu.setState()`

This cascade produces **14 separate React renders totaling ~138ms of JS**, followed by two full-page layouts (~117ms, touching all 1,285 DOM nodes), for a total post-resize cost of **~750ms**.

### Suggested Fixes

1. **Debounce the resize handler** — wrap `UnderlineNav`'s measurement logic in `requestAnimationFrame` (or a short debounce) so it fires once per frame, not once per resize event (7 dispatches observed in this trace).

2. **Memoize child components** — the `Link`/`ExtendedLink`/`ForwardRef` children should be wrapped in `React.memo()`. Their props almost certainly don't change on resize, yet they account for 385 unnecessary re-renders.

3. **Batch the layout-dependent state updates** — `StackState`, `PullRequestHeader`, `Overlay`, and `ActionMenu` all independently read layout and call `setState()`, creating the cascade. Consolidating their measurements into a single shared `ResizeObserver` callback (or a `useLayoutEffect` that batches all reads before any writes) would collapse 4 cascading renders into 1. -- This one I'm less sure about and is a suggestion from Copilot. the other ones sound reasonable to me.

### Steps to reproduce

1. Go to a Conversation page on a PR
2. Throttle CPU
3. Resize the window
4. Observe lag

### Version

v38.21.0

### Browser

_No response_

iansan5653:
Fixes https://github.com/primer/react/issues/7823

This PR is an alternative approach to https://github.com/primer/react/pull/8108 to resolve https://github.com/github/primer/issues/6806.

The problem is that some flickering is inevitable if we want to dynamically hide the icons: we have to let the container overflow in order to detect it and hide the icons, but then when we hide the icons the container no longer overflows. We can _minimize_ flickering but we can never _eliminate_ it with the dynamic approach. Dynamically hiding icons inherently creates an infinite feedback loop.

This PR proposes a simpler, less dynamic alternative: we just hide the icons at hardcoded breakpoints. The breakpoint at which icons are hidden can be controlled by a prop which allows for all the standard Primer breakpoint sizes. The default is very large to avoid breaking existing consumers (namely the global GitHub navigation), but it can easily be reduced by consumers who know they won't be rendering many tabs.

The result is a flicker-free, stable component. However, the drawback is that we sometimes will hide icons early than we would otherwise need to. I think this might be a fair tradeoff.

**Also includes some other fixes**:

1. Change overflow anchor button visibility to use `display` instead of `visibility` to avoid having to save space for it
2. Allow the first tab to wrap out of the way on _very_ small screens by rendering an invisible first element

#### Changed

- Replaced `UnderlineNav` dynamic icon visibility with static breakpoints

### Rollout strategy

- [ ] Patch release
- [x] Minor release


.changeset/khaki-parrots-eat.md
```diff
@@ -0,0 +1,6 @@
+---
+"@primer/react": patch
```
jonrohan: This adds a new public prop (`hideIconsBreakpoint`), which per the changesets guide is a `minor`, and it matches the "Minor release" box you checked in the PR body. Should this be `minor` rather than `patch`?

```suggestion
"@primer/react": minor
```

.changeset/khaki-parrots-eat.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+`UnderlineNav`: Replaced dynamic icon visibility with static breakpoints to avoid flickering. Breakpoints customizable via the `hideIconBreakpoint` prop.
```
jonrohan: The prop is `hideIconsBreakpoint` (plural), so this changelog line points consumers at a prop that doesn't exist.

```suggestion
`UnderlineNav`: Replaced dynamic icon visibility with static breakpoints to avoid flickering. Breakpoints customizable via the `hideIconsBreakpoint` prop.
```

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -42,24 +47,21 @@ export const UnderlineNav = forwardRef(
       variant = 'inset',
       className,
       children,
+      hideIconsBreakpoint = 'xxlarge',
```
jonrohan, line 50: The comment above says that it defaults to xlarge. I think that's a pretty good point to start at. Let's update the default here.

```suggestion
      hideIconsBreakpoint = 'xlarge',
```

.changeset/khaki-parrots-eat.md
iansan5653: Yes! Overlooked that

packages/react/src/UnderlineNav/UnderlineNav.tsx
iansan5653: Oh actually it needs to be `xxl` because otherwise the global nav tabs will overflow before the icons hide (there can be a lot of them). I'll update the doc

jonrohan (approved):

packages/react/src/UnderlineNav/UnderlineNav.tsx
```diff
@@ -96,12 +98,13 @@ export const UnderlineNav = forwardRef(
...
           data-overflow-mode="wrap"
-          // Force icons to stay hidden, avoiding flickering as icons create/remove overflow
-          data-hide-icons={hasEverOverflowed ? 'true' : undefined}
+          data-hide-icons-breakpoint={hideIconsBreakpoint ?? undefined}
           // Ensure button is shown (after initial render) on browsers that don't support scroll-driven animations
           data-has-overflow={isOverflowing ? 'true' : undefined}
         >
           <UnderlineItemList ref={listRef} role="list" className={classes.ItemsList}>
+            {/* Empty first element allows real first item to wrap out of view on tiny screens */}
+            <li role="presentation" aria-hidden />
```
jonrohan: One small issue I noticed on github-ui this extra element caused a gap to be present for the first item. I would change the gap property on the ul to a `margin-inline-end: var(--stack-gap-condensed);` on the UnderlineItem css.

Image

Image

iansan5653: Oh good catch! I totally missed that. 

@copilot address this by applying a negative right margin equivalent to the flex gap to this hidden element

jonrohan (approved):

PR #8149, opened 2026-07-10
ci: annotate React Compiler lint failures

joshblack:
This PR updates our React Compiler lint task so failures create file annotations in GitHub Actions.

#### New

- Add GitHub Actions error annotations with the file, line, column, and compiler failure.

#### Changed

- Update the React Compiler CI task to remove Turbo log prefixes so GitHub can recognize workflow commands.

#### Removed

- None.

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [x] None; this is a change to our internal CI workflow and has no public-facing impact.

### Testing & Reviewing

Reviewers can verify that:

- A React Compiler failure creates an annotation on the affected source file.
- The existing readable failure summary is still included in the job log.
- The lint task continues to pass without annotations when all migrated files compile.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Check className test coverage
         run: npm run test:classname-coverage
       - name: Check react compiler coverage
-        run: npx turbo run lint:react-compiler
+        run: npx turbo run lint:react-compiler --log-prefix=none
 
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


joshblack (commented):
@copilot could we make it so that it runs on all files but that it will only error if a file that's not in the exclusion list fails? In addition, files that are not migrated should only emit an info or warning level event. Make sure that the info includes the detail and suggestion for how to fix it, when possible

joshblack (commented):
@copilot actually instead of all files could it:

- Find the list of changed files in the PR
- Pass it along to the lint script
- If there is no overlap, do nothing with the changed files
- For files that intersect with the package, annotate them
- If they are in the not migrated section, use a notice
- If they are not in the not migrated section, use error

PR #8160, opened 2026-07-14
Export `PageLayout` resizable functionality

TylerJDev:
Part of https://github.com/github/core-ux/issues/2973, prerequisite to https://github.com/github/github-ui/pull/26724

Places `PageLayout` drag functionality into its own component: `packages/react/src/PageLayout/DragHandle.tsx`, to allow usage outside of `PageLayout`. 

The changes here aim to extend the modularity of `PageLayout` by breaking up the drag functionality. We keep the functionality exactly the same within `PageLayout`, but export the pieces that can be used in other components, (see https://github.com/github/core-ux/issues/2973 for more context).  

**PR changes**: 

- `packages/react/src/PageLayout/usePaneWidth.ts`
  - We make the `contentWrapperRef` prop optional so that other components can pass or omit a ref.
  - Changes in this file are mostly additive. 
- `packages/react/src/PageLayout/PageLayout.tsx`
  - We move `DragHandle` into its own component (`packages/react/src/PageLayout/DragHandle.tsx`)
  - We move some imports to `DragHandle`
- `packages/react/src/PageLayout/DragHandle.tsx`
  - Near 1:1 with what already existed within `PageLayout`.
  - Modified/added props, such as `dragTargetRef`.
  - Behavior remains the same.

#### Changed

* Moves `DragHandle` outside of `PageLayout.tsx`, into its own component.
* Adjusted some props to handle usage of `DragHandle` / `usePaneWidth` outside of `PageLayout` context.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

The behavior should be exactly the same as it is today in production. In addition, I've added a story (on dev) that shows what usage could look like when `DragHandle` is used outside of `PageLayout`. 

**Questions**

* Could we simplify the usage any more than what we have currently?
* Do we need to add new additions to our docs?


packages/react/src/PageLayout/DragHandle.tsx
```diff
@@ -0,0 +1,239 @@
...
+  onDragEnd,
+  onDoubleClick,
+  'aria-valuemin': ariaValueMin,
+  'aria-valuemax': ariaValueMax,
+  'aria-valuenow': ariaValueNow,
+  dragTargetRef,
+  contentWrapperRef,
+  formatValueText = defaultFormatValueText,
+  'data-component': dataComponent = 'PageLayout.DragHandle',
+}) {
```
TylerJDev, line 75: A near 1:1 with what we already had in `PageLayout`, just moved to its own file so that it can be exported.

There are a few changes, such as props to improve the portability. Behavior and functionality should remain the same within `PageLayout`.

packages/react/src/PageLayout/DragHandle.tsx
```diff
@@ -0,0 +1,239 @@
...
+export const DragHandle = memo<DragHandleProps>(function DragHandle({
+  handleRef,
+  onDragStart,
+  onDrag,
+  onDragEnd,
+  onDoubleClick,
+  'aria-valuemin': ariaValueMin,
+  'aria-valuemax': ariaValueMax,
+  'aria-valuenow': ariaValueNow,
+  dragTargetRef,
```
TylerJDev, line 71: This is different compared to how it was in `PageLayout`. Before we directly used `paneRef` via context, but now we pass `paneRef` via this prop in `PageLayout`. 

This is to support usage outside of `PageLayout`, where the target ref differs from the one within the PageLayout component.

packages/react/src/PageLayout/PageLayout.tsx
```diff
@@ -315,217 +312,6 @@ const SidebarDivider = memo<SidebarDividerProps>(function SidebarDivider({
   )
 })
 
-type DragHandleProps = {
```
TylerJDev, line 318: Drag handler functionality moved to `DragHandle.tsx`.

joshblack (approved):

PR #8161, opened 2026-07-14
docs: add initial style guide docs sections

joshblack:
WIP


.github/skills/style-guide/docs/component-props.md
```diff
@@ -0,0 +1,306 @@
...
+Use `has*` when the prop describes the presence of a structural visual element:
+
+- `hasBorder`
+
+Use `show*` or `hide*` when the prop explicitly controls visibility of optional
+content:
+
+- `showSelectAll`
+- `showSelectedOptionsFirst`
+- `hideTitle`
```
siddharthkp, line 96: should we have an opinion on which one to prefer based on the default value? 

`hideTitle: true | undefined (false)` vs showTitle: `undefined (true) | false`

.github/skills/style-guide/docs/component-props.md
```diff
@@ -0,0 +1,306 @@
...
+Use `children` for primary visible content when possible.
+
+Use semantic names for specific visible text:
+
+- `title`
+- `subtitle`
+- `description`
+- `placeholder`
+- `caption`
+- `text`
```
siddharthkp, line 144: text is the only one here that feels generic/non-semantic

PR #8169, opened 2026-07-15
Timeline.Actions: wrap below Body on narrow viewports

liuliu-dev:
Closes github/primer#6693

When `Timeline.Item` is below 480px inline size, it now switches to a two-row grid so `Timeline.Actions` wraps onto its own line under `Timeline.Body`, left-aligned with the body content. Above 480px is unchanged.

Uses a container query on `.Timeline`, so the wrap responds to the timeline's own width — narrow side panels and split views work correctly without depending on viewport size.

#### New

- Narrow-viewport wrap for `Timeline.Actions` (< 480px).
- `WithActions` story: commit-SHA rows hide the `Verified` /`Unverified` label at narrow widths, and the cross-reference card
  stacks vertically.

  | Before | After |
  |--------|-------|
  |     before   |   after    |

### Rollout strategy

- [X] Patch release


janmaarten-a11y (approved):
Looks great to me. ❤️  

I double-checked my recent github-ui PRs that aim to add proper list semantics to events and found no major conflicts. 
- https://github.com/github/github-ui/pull/22083
- https://github.com/github/github-ui/pull/26232
- https://github.com/github/github-ui/pull/26229

The one possible interaction to watch is is that #26232's `.Timeline_Item_withDivider` explicitly relies on `Timeline.Item` being a flex row it can wrap so the full-width divider spans correctly (its own comment says so). If a divider row ever renders inside a Timeline narrower than 480px with the semantics flag on, #8169's grid could override that flex assumption and the divider could reflow. It only bites at <480px container width with both changes active and the flag on, so production (flag off) is unaffected.

Not something to worry about in this PR, I think. Once this one lands, I'll check to see if any updates are needed to that one.

jonrohan (approved):

PR #8174, opened 2026-07-16
Automate ADR-backed component review guidance

adierkens:
Closes #

Builds on #8173's small ADR-backed review trial by replacing its hand-authored instruction with a complete, validated policy generation system. Public ADRs and maintained contributor guidance produce path-scoped Copilot Code Review instructions; sanitized internal direction is limited to outcome-only rules.

#### New

- Structured public and outcome-only internal policy manifests
- Deterministic path-scoped instruction generation
- Reusable component-review skill and read-only reviewer agent
- Schema, privacy, source-digest, and generated-output tests

#### Changed

- Replaces the base PR trial instruction with the generated policy catalog
- ADR authoring guidance identifies review-policy impact
- CI checks generated guidance and lint autofix regenerates stale output

#### Removed

- The hand-authored four-rule trial instruction from the base PR

### Rollout strategy

- [ ] Patch release
- [ ] Minor release
- [x] None; review tooling, documentation, and CI only. No published package behavior changes.

### Testing & Reviewing

Review this after the base trial establishes that path-scoped ADR rules produce useful CCR feedback. Compare `contributor-docs/review-guidance/` with the generated instructions and consolidated skill reference. The internal policy intentionally excludes private provenance, identifiers, rationale, links, and alternatives.

Run `npm run test:review-guidance` and `npm run check:review-guidance`.


joshblack (commented):
I tried running this to see what kind of stuff it would output and I was surprised that it didn't seem to honor some of the ADRs 🤔 

As an example, I tried to create an accordion component locally and was surprised that it didn't follow the file conventions ADR to include a test file.

Could you do like a before/after kind of approach for what this kind of stuff would help out with compared to baseline? In just a one-off test it didn't seem to provide a better outcome but I also get how unscientific that is lol.

Happy to keep testing if that would be better, just wanted to leave a note while looking at this today 👀

adierkens:
@joshblack What prompts were you using?

I had it take a similar component, and same prompt and I could see the tool calls to the ADRs which made some changes to how it generated code. In the baseline example it used CSS modules, but didn't use shared mixins for focus (and did in the other), similarly my test-branch generated VRT and ART tests but the baseline one didn't.

joshblack:
@adierkens yeah it definitely called into them / saw the ADRs but it did not seem to apply all the rules from them (or at least I noticed that it didn't follow the test structure one).

I just used this as a one-off for the two:

```
let's make an accordion component. follow best practices, guidelines, etc and add it to the experimental entrypoint in the primer/react package
```

The baseline version did include this so it could also just be stochastic behavior here (e.g. if I ran it again I'm sure I'd get something different). The prompt itself could also be different, just wanted to see what would happen honestly and how much steering having the ADRs would change compared to baseline

PR #8179, opened 2026-07-16
docs: add component spec skill

joshblack:
Closes #

### Rollout strategy

- [ ] Patch release
- [ ] Minor release

### DIFF
--- a/.changeset/blankslate-accessible-visuals.md
+++ b/.changeset/blankslate-accessible-visuals.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+Blankslate: Treat visuals as decorative, allow standalone primary actions, and render primary actions with defined `href` values as links.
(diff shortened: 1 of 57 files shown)


joshblack (commented):
@copilot update the spec skill so that markup examples include basic HTML markup in the spec. it shouldn't just refer to it, html code blocks should live in there. Also make sure playground and default stories do not reference the spec (and correct any that do)

joshblack (commented):
@copilot apply this feedback from @ericwbailey 

- If applicable, add keyboard navigation for behavior or interactions
- Communicate what states are allowed, in what context is a state permissible

joshblack (commented):
@copilot include responsive considerations, where applicable, for functionality. (does content change)

joshblack (commented):
@copilot include a "Log" section at the bottom and instructions to keep this updated with notable decisions or changes.

Add a reviewer agent for spec that is adversarial, make people who change the component spec prove why this change is needed.

PR #8187, opened 2026-07-17
fix(filtered-action-list): respect item renderers

joshblack:
Closes github/primer#6891

This PR updates `FilteredActionList` so that item-level `renderItem` callbacks take precedence over the list-level fallback. Previously, the list renderer was applied after the item props and replaced an item renderer with `undefined` when no list-level renderer was provided.

#### New

- Add regression coverage for item-level renderers in grouped and ungrouped lists

#### Changed

- Update `FilteredActionList` to preserve item-level custom renderers across grouped, virtualized, and standard lists

#### Removed

- None

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

The main thing to verify is that item-level renderers override the list-level fallback without changing the existing default rendering path.

- `npm run build`
- `npm run type-check`
- `npm test -- --run --exclude '.agents/**'`
- `npm run lint -- --ignore-pattern '.agents/**'`
- `npm run lint:css -- --ignore-pattern '.agents/**'`
- `npm run format:diff`


joshblack (commented):
@copilot instead of changing the spread order, explicitly merge the `renderItem` prop

siddharthkp (approved):

PR #8188, opened 2026-07-17
Use `base-size-32` instead of hardcoded `32px` for item height

iansan5653:
The underline nav item height is hardcoded at 32px. This means that when the font size changes (ie, "Make text smaller" in Safari), the item height doesn't respect the new font size. But the container does, because it has `max-height: var(--control-xlarge-size)` which is font-size dynamic. This causes the container to clip all the items by having a smaller height than they can fit in, which treats them all as overflowing.

The fix is simple, we can just use `--base-size-32` instead. 

**However**, this still doesn't work perfectly. On sufficiently large font sizes (+3 zoom levels), the pixel-perfect accuracy of the sizes just isn't calculated quite right, and so the items still overflow even though they seem fully visible. I think this is a safari issue and there's not much we can do about it except to make the overflow logic slightly more forgiving, allowing up to 5% of an item to be clipped before hiding it. 

#### Changed

- Fixes `UnderlineNav` items all being considered overflowing on Safari with font size overrides

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### DIFF
--- a/.changeset/heavy-brooms-wash.md
+++ b/.changeset/heavy-brooms-wash.md
@@ -1,0 +1,5 @@
+---
+"@primer/react": patch
+---
+
+Fix `UnderlineNav` items all being considered overflowing on Safari with text size overrides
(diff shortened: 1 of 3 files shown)


iansan5653:
Confirmed working in storybook on every text size 

https://github.com/user-attachments/assets/7d7b9765-4ecd-4bb7-a6cc-2b728da391f2

jonrohan (approved):

PR #8194, opened 2026-07-20
Prevent Avatar from shrinking in Button visuals

### LINKED ISSUE #7974: Avatar in Button leadingVisual slot does not properly size img due to conflicting styles
### Description

There’s a visual bug in the Button component: when passing an Avatar into the leadingVisual slot, the Avatar's image does not render with the correct width. This seems to be caused by a collision between global styles (e.g. `img { max-width: 100% }`) and less-specific Avatar styles (`width: 20px` via CSS var). Because Button uses `min-content` for the leadingVisual grid area, `max-width` sometimes resolves to 0, causing the image to shrink unexpectedly.

Setting `min-width: var(--avatarSize-regular);` alongside `width` on the Avatar CSS appears to solve the problem more generally, ensuring Avatars don’t shrink and always use the intended size.

### Steps to reproduce

1. Render a Button component using an Avatar as the leadingVisual prop/value.
2. Example: navigate to a Button with an Avatar at the bottom of https://github.com/marketplace/lovablebot
3. Observe the Avatar image does not retain the expected width.

### Version

[View original Slack conversation](https://github.slack.com/archives/CSGAVNZ19/p1781198554259359?thread_ts=1781198554.259359&cid=CSGAVNZ19)

wanxiankai:
Closes #7974

#### New

- Added a patch changeset for the Avatar sizing fix.

#### Changed

- Set `Avatar` `min-width` from the same size variables as `width` so avatars retain their intended intrinsic width in min-content slots, including `Button` leading visuals.
- Added a CSS rule test that verifies the non-shrinking `min-width` declaration is present.

#### Removed

- None.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- `npm run test -- packages/react/src/Avatar/Avatar.test.tsx --run`
- `npm exec eslint -- packages/react/src/Avatar/Avatar.test.tsx`
- `npm exec stylelint -- packages/react/src/Avatar/Avatar.module.css`
- `npm run test:type-check`
- `npm exec prettier -- --check packages/react/src/Avatar/Avatar.module.css packages/react/src/Avatar/Avatar.test.tsx .changeset/green-avatars-stand.md`


jonrohan (approved):

liuliu-dev (changes requested):
This change is causing VRT failures in the `AvatarStack` custom-size responsive story. https://primer-874780de13-13348165.drafts.github.io/storybook/?path=/story/components-avatarstack-features--custom-size-on-children-responsive

Could we address this regression?

PR #8212, opened 2026-07-21
Add next ThemeProvider without legacy JS theming

mattcosta7:
Closes #

This PR adds a CSS-variable-based `ThemeProvider` at `@primer/react/next` so consumers can opt out of the large legacy JavaScript `colorSchemes` payload before the next major release. The root `ThemeProvider` keeps its existing behavior for compatibility and is deprecated in favor of the new entrypoint.

The two providers share one `ThemeContext`, `useTheme`, and internal state/rendering implementation. This supports incremental migration and mixed nesting while keeping `deepmerge`, the default JavaScript theme, and legacy color-scheme data out of the `/next` runtime graph.

`@primer/styled-react` now owns active generation of the legacy `colorSchemes` data it still requires. `@primer/react` retains a frozen compatibility snapshot for the deprecated root provider.

No visual change is expected. Both providers continue applying color schemes through the existing `data-color-mode`, `data-light-theme`, and `data-dark-theme` attributes.

#### New

- Export a CSS-variable-based `ThemeProvider` from `@primer/react/next`.
- Export the shared `useTheme` and `useColorSchemeVar` hooks from `@primer/react/next`.

#### Changed

- Deprecate the root `ThemeProvider` in favor of `@primer/react/next` while preserving its legacy JavaScript theme behavior.
- Deprecate the `theme` and `resolvedColorScheme` properties returned by `useTheme`.
- Keep root and `/next` providers interoperable through a shared context and hook.
- Vendor active legacy theme generation into `@primer/styled-react`.

#### Removed

- No consumer-facing APIs are removed in this release.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

Consumers can migrate incrementally:

```diff
-import {ThemeProvider} from '@primer/react'
+import {ThemeProvider} from '@primer/react/next'
```

Existing `useTheme` imports remain compatible because both providers share the same context. Consumers of the legacy `theme` and `resolvedColorScheme` values should migrate away from those properties before switching providers.

A future major release can promote the `/next` provider to the root entrypoint and remove the deprecated JavaScript theme payload.

### Testing & Reviewing

- Full ESLint and CSS lint
- Prettier formatting check
- Classname coverage check
- Full TypeScript validation: 15/15 tasks
- Full unit suite: 2,596 tests passed
- Full repository build: 5/5 tasks
- `@primer/react` package build and `publint`
- Verified emitted `/next` runtime modules do not import `deepmerge`, `theme`, `colorSchemes`, or legacy theme data


mattcosta7:
vrt seems unrelated to schemes - maybe a flake or font loading?

jonrohan (approved):

joshblack:
Just a heads up, this was slated as [a part of v39](https://github.com/github/primer/issues/5430#issuecomment-3473138559) since it's a major change and we tend to group them together to land them at once.

Just curious, are we importing this still in github-ui now that @primer/styled-react is no longer used there? If we are, do we happen to know why? I would assume this would be tree-shaken if it's not being used

mattcosta7:
> Just a heads up, this was slated as [a part of v39](https://github.com/github/primer/issues/5430#issuecomment-3473138559) since it's a major change and we tend to group them together to land them at once.
> 
> Just curious, are we importing this still in github-ui now that @primer/styled-react is no longer used there? If we are, do we happen to know why? I would assume this would be tree-shaken if it's not being used

The colorscheme object is merged into the theme, and merged in the regular `ThemeProvider`

Any app that uses a ThemeProvider from the main primer/react bundle incurs the cost of the entire legacy theme object - which can't be treeshaken, since it is used

Getting start still says to wrap with a ThemeProvider: https://primer.style/product/getting-started/react/ which imports the heavy legacy-theme and forces it into the bundle.

I marked it as a v39, so it can sit open for a while, but it's a very hefty bundle improvement.

I had done this iniitally as a spike in github-ui to see the cost, and just extracted that upstream since the work was already there

mattcosta7:
> Just a heads up, this was slated as [a part of v39](https://github.com/github/primer/issues/5430#issuecomment-3473138559) since it's a major change and we tend to group them together to land them at once.
> 
> Just curious, are we importing this still in github-ui now that @primer/styled-react is no longer used there? If we are, do we happen to know why? I would assume this would be tree-shaken if it's not being used

The base ThemeProvider imports [defaultTheme](https://github.com/primer/react/blob/e24effd440d69f0068a89aba26bae9e39fb3254e/packages/react/src/ThemeProvider.tsx#L2) which imports [colorSchemes from legacy themes](https://github.com/primer/react/blob/e24effd440d69f0068a89aba26bae9e39fb3254e/packages/react/src/theme.ts#L3)

So any app that use `import {ThemeProvider} from '@primer/react'` which is probably all of them?  incurs the 200K or so gzup of that legacy-theme

siddharthkp:
Hi! 

I'd like someone (or me) to verify that this is a safe change by testing it on local or preview before we merge. I have a feeling this isn't the kind of change that CI would catch

We could also look into feature flagging this change if the above exercise ^ does not give enough confidence

joshblack:
Makes sense @mattcosta7! I totally had thought ThemeProvider was removed with v38 in github-ui but that is clearly not the case 😂 

Agreed that this would be a big bump, will be great to get it in.

Since the theme is a part of the context value, does this mean that all usage reading this context from `useTheme` or related is gone now too?

mattcosta7:
> Hi!
> 
> I'd like someone (or me) to verify that this is a safe change by testing it on local or preview before we merge. I have a feeling this isn't the kind of change that CI would catch
> 
> We could also look into feature flagging this change if the above exercise ^ does not give enough confidence

There are no consumers at least, so it's a pretty simple swap at least, which gives some confidence

mattcosta7:
> Since the theme is a part of the context value, does this mean that all usage reading this context from `useTheme` or related is gone now too?

Reading the theme/colorSchemes, yep, but that should have already been migrated to css modules (at least in dotcom it was) - and the remaining context usage is for raw/cheap strings mode and scheme names - which lose the heavy object, which equates to like 1/4 of the total size of primer (if not more)

siddharthkp (approved):
Approving because you seem confident

mattcosta7:
> Approving because you seem confident

I'm not sure when the next major release is planned (or if this is a good time to push for it - so happy to keep this parked until then too - no rush to land it)

siddharthkp:
> I'm not sure when the next major release is planned

We're hoping to cut one after the API consistency project https://github.com/github/primer/issues/6761, so in a couple months probably

mattcosta7:
> > I'm not sure when the next major release is planned
> 
> We're hoping to cut one after the API consistency project [github/primer#6761](https://github.com/github/primer/issues/6761), so in a couple months probably

@siddharthkp  - would we be able to ship this with some additional guardrails?

we could keep the old ThemeProvider as is, and then ship a 'new' version that has the slimmer codepaths? That would also allow us to start marking 'deprecated' now and make the future migration a bit smoother?


I pushed that, which also makes a no-op for updates, and we can slow opt-in

mattcosta7:
VRT seems related to expected changes from - https://github.com/primer/react/pull/8271

jonrohan (approved):

PR #8214, opened 2026-07-22
Migrate `UnderlinePanels` to the new `Tabs` component

rickyzhangca:
Closes github/primer#6899

Migrates the experimental `UnderlinePanels` component to build on the React-first `Tabs` primitive (`Tabs` + `useTab` / `useTabList` / `useTabPanel`) instead of the imperative `@github/tab-container-element` web component. The public API and behavior are unchanged.

`UnderlinePanels` still matches tabs and panels by DOM order (no `value` in the public API) — an internal positional `value` is injected via `cloneElement`. Selection stays hybrid: it is seeded/re-synced from the consumer's `aria-selected` prop while clicks/keyboard update it internally, mirroring the previous `tab-container-element` behavior (including defaulting to the first tab when none is selected).

https://github.com/user-attachments/assets/7c727c66-a52b-4151-a758-f9cdeb20ae4d

#### New

- Optional `id` prop on the experimental `Tabs` component to control the base id used for generated tab/panel ids (falls back to an auto-generated id).
- `InOverlay` dev story showing `UnderlinePanels` inside an `AnchoredOverlay`.

#### Changed

- `UnderlinePanels` is now built on the experimental `Tabs` component and hooks instead of `@github/tab-container-element`.

#### Removed

- Direct dependency on `@github/tab-container-element` (it was the only consumer).

### Rollout strategy

- [x] Patch release
- [ ] Minor release

### Testing & Reviewing

- Public API and rendered behavior are unchanged; existing unit tests pass, plus new regression tests for default first-tab selection and uncontrolled panel switching.
- Verified against the `tab-container-element` source that it defaults to the first tab when none is `aria-selected`, so the no-`aria-selected` stories keep visual parity (only the first panel is shown).
- `tsc`, ESLint, `@primer/react` build, and the exports snapshot test all pass.
- Please run the `@vrt` and `@avt` Playwright suites in CI to confirm visual/accessibility parity.

### DIFF
--- a/.changeset/underlinepanels-tabs-migration.md
+++ b/.changeset/underlinepanels-tabs-migration.md
@@ -1,0 +1,5 @@
+---
+'@primer/react': patch
+---
+
+UnderlinePanels: The experimental `UnderlinePanels` component is now built on the experimental `Tabs` component instead of `@github/tab-container-element`. Its public API and behavior are unchanged.
--- a/package-lock.json
+++ b/package-lock.json
@@ -3913,10 +3913,6 @@
       "integrity": "sha512-L/2r0DNR/rMbmHWcsdmhtOiy2gESoGOhItNFD4zJ3nZfHl79Dx3N18Vfx/pYr2lruMOdk1cJZb4wEumm+Dxm1w==",
       "license": "MIT"
     },
-    "node_modules/@github/tab-container-element": {
-      "version": "4.8.2",
-      "license": "MIT"
-    },
     "node_modules/@hono/node-server": {
       "version": "1.19.14",
       "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.14.tgz",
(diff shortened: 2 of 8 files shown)


jonrohan (approved):

llastflowers (approved):

rickyzhangca:
Applied the `integration-tests: skipped manually` label.

The `github-ui / *` integration checks fail on this PR, since this PR removes `@github/tab-container-element` from `@primer/react`, and github-ui's `verify-package-versions.ts` keeps that package in its `ignoredDuplicates` allow-list (it is currently duplicated in github-ui: `3.1.2` from github-ui's own `ui-service`/`primer` packages + `4.8.2` pulled in transitively by the published `@primer/react@38.33.0`). When the integration test installs this PR's packed build, the `4.8.2` copy disappears, the package is no longer duplicated, and github-ui's guard fails on purpose to prompt cleanup of the ignore-list entry.

That cleanup cannot happen on github-ui today (with `38.33.0` pinned, the package is genuinely still duplicated, so removing the entry would trip the opposite "duplicate found" guard). It should land naturally the next time github-ui bumps `@primer/react` past this change.

Tracking issue for the github-ui cleanup: github/github-ui#27652

No immediate release is required.

PR #8216, opened 2026-07-22
Timeline: tag event stories with taxonomy data-* and split the taxonomy into core and GitHub layers

janmaarten-a11y:
Part of github/primer#6664 (epic github/primer#6654). Everything here is internal to the Timeline module and Storybook-only: there is no consumer-facing change, nothing is added to the package's public exports, and there is no changeset.

This PR does two things. It tags the five Timeline surface story sets (License Compliance, Code Scanning, Dependabot, Secret Scanning, Issues) with the taxonomy data-* attributes through one shared helper, and it organizes the taxonomy folder into two clearly separated layers: a generic core that any product could reuse, and a GitHub-specific layer that holds GitHub's own values. The taxonomy folder encodes GitHub product knowledge (which surfaces are security alerts, the per-surface event catalogs, which logins are bots), and the team agreed that Primer should own the generic data-* schema and its serializer while the GitHub-specific values live in a wrapper that can eventually move into github-ui. This split draws that line now, while the module is still internal, so the later move is a small, contained step.

## The two layers

- taxonomy/core is generic and product-agnostic. It holds the data-* attribute types, the serializer that turns an event into its attribute set, and a new factory, createEventDataAttributesFor, that binds a catalog (and an optional rule for turning a login into a person-or-bot value) into a ready-to-use helper. Nothing in core knows about GitHub surfaces, categories, or bot logins.
- taxonomy/github is the GitHub-specific layer. It holds the surfaces, the category model, the five per-surface event catalogs, the rule that classifies a login as a person or a bot, and eventDataAttributesFor, which is now a single line that instantiates the core factory with the GitHub catalog and the GitHub actor rule.
- taxonomy/index re-exports both layers, so every symbol the stories rely on, and eventDataAttributesFor in particular, stays available from the same place as before. The five surface story files import it exactly as they did and did not change.

## What each row carries

Every event row spreads the data-* set produced by eventDataAttributesFor(scope, type, login?): data-event-scope, data-event-type (the unscoped leaf), data-event-category and data-event-visibility (both read from that surface's catalog, so the stories stay in sync with the source of truth instead of hand-typing values), and data-actor-type (user or bot, resolved from the login and omitted when the row has no actor). The helper is type-safe: the compiler enforces that the event type is a real leaf of the given surface, and it never emits data-actor-type for a structurally actor-less event even if a login is passed.

## Behaviour is unchanged

- The emitted data-* attributes are byte-identical before and after the split. Every event story across the five files was rendered and every row that carries data-event-scope had its data-* values compared against a baseline captured before the change; the difference is none across all 162 rows.
- The five Timeline surface story files are untouched (zero changed lines).
- The taxonomy stays internal, so there is no public export change and no changeset.

## Testing

The repo's own gates pass: prettier, eslint with zero warnings allowed, and type-check. The taxonomy tests pass, including new core tests for the serializer and the factory (person-versus-bot resolution, the omitted-actor case, the actor-less suppression, auditOnly pass-through, and category derivation from the catalog) alongside the moved surface tests. A render-and-diff equivalence check across every tagged row confirmed the reorganization changed no attributes.


packages/react/src/Timeline/Timeline.issues.features.stories.tsx
```diff
@@ -374,22 +384,6 @@ export const EventState = () => (
...
-            <Octicon icon={IssueDraftIcon} />
-          </Timeline.Badge>
-          <Timeline.Body>
-            <UserActor href="#" muted />
-            {'converted this from a draft issue '}
-            <MutedTime date={new Date('2022-07-21T08:30:00Z')} href="#" />
-          </Timeline.Body>
-        </Timeline.Item>
-      </Timeline>
-    </VariantSection>
```
francinelucca, line 392: is this meant to be removeD?

janmaarten-a11y: Good catch, apparently Copilot got overly hungry! `converted_from_draft` is a real issue event via Projects v2. Restoring it...

janmaarten-a11y: Fixed in https://github.com/primer/react/pull/8216/commits/92981faf70defb04803fa60182a4cccda0de93d9

liuliu-dev (commented):
I think this PR can deliver the Storybook playground from [#6664](https://github.com/github/primer/issues/6664), while keeping the GitHub event taxonomy in `github-ui`.

I’d suggest refocusing the PR on an interactive playground that uses a small set of representative example events defined within the story, filters the data before rendering, and passes representative `data-*` attributes to `Timeline.Item`. The filter control could be composed from `ActionMenu` and `ActionList`. This would demonstrate how consumers can build `Timeline` filtering without making Primer the source of truth for GitHub event categories.

janmaarten-a11y:
Closing this without merging. The team has decided the Timeline event taxonomy should live in `github-ui` rather than Primer, so Primer stays taxonomy-agnostic and ships only the `Timeline` component and, later, an optional presentational `Timeline.Filter` control. That removes the purpose of this PR, which tags Primer stories with GitHub-specific `data-*` values and reorganizes a GitHub taxonomy inside Primer, both of which now belong in `github-ui`.

Nothing is lost. This branch stays in git history and carries the most complete version of the model, with the generic core and the GitHub-specific catalogs split into separate layers, and `github-ui` will lift the taxonomy from here. The per-surface event research that informed it is preserved and will feed that work.

Next steps are tracked in #6664: follow-ups adds the playground, representative story data, the filter controls, and the `data-*` attributes from `github-ui`. 

The taxonomy module merged into Primer in #8180 is being reverted separately.

siddharthkp:
@janmaarten-a11y Thanks for leaving a good comment here instead of just closing it ❤️

PR #8233, opened 2026-07-24
Add controlled value/onChange and activationMode to UnderlinePanels

rickyzhangca:
Closes github/primer#6913

`UnderlinePanels` (experimental) is hard to use for **data-driven** tabs, where switching tabs changes what gets fetched or shown. Today selection is keyed to DOM order and the only public hooks are `Tab.aria-selected` (a seed) plus `Tab.onSelect`, which does not fire on arrow-key selection. So consumers cannot reliably use the selected tab as a single source of truth (this blocked migrating GitHub's `RefSelector`).

This surfaces a real controlled/uncontrolled selection API on top of the internal `Tabs` primitive, keyed to stable per-tab values.

https://github.com/user-attachments/assets/c2783d44-dd03-4cda-850a-b17811844c8b

### Approach

- Selection state is owned by `UnderlinePanels`, which drives the `Tabs` primitive in controlled mode. When a `Tab`/`Panel` provides an explicit `value`, domain values (e.g. `'branch'`) are used; otherwise a positional value is injected so existing usage keeps working unchanged.
- `activationMode` lands on the `Tabs` primitive (`useTab`/`useTabList`) first, then is surfaced on `UnderlinePanels`. In manual mode the roving tab stop follows focus per the APG manual activation pattern, so arrow keys move focus without committing selection.

The change is fully additive and backward compatible. The existing `aria-selected` seed and `onSelect` callback are unchanged and not deprecated. Note the intentional difference between the two callbacks: `Tab.onSelect` fires only on click and Enter/Space, while the new container `onChange` fires for every selection path including Arrow/Home/End.

#### New

- `UnderlinePanels`: `value`, `defaultValue`, `onChange`, and `activationMode` props.
- `UnderlinePanels.Tab` / `UnderlinePanels.Panel`: `value` prop to pair tabs and panels by a stable value instead of DOM order.
- `Tabs` experimental primitive (and `useTab`): `activationMode` (`'automatic' | 'manual'`) option.
- New feature stories: `Controlled`, `Uncontrolled`, `ManualActivation`, and `InOverlay` (AnchoredOverlay + FocusZone composition).

#### Changed

- `Tabs` arrow/Home/End keyboard navigation now keys off the focused tab (falling back to the selected tab) so focus and selection can diverge in manual mode.

#### Removed

- N/A

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- Activation mode and controlled/uncontrolled selection are covered as a full 2x2 matrix, all with tests: automatic + controlled, automatic + uncontrolled, manual + controlled, and manual + uncontrolled (across the `UnderlinePanels` and `Tabs` suites).
- New unit tests also cover: controlled `value` drives selection and panel visibility, `onChange` fires on click / Enter / Space / Arrow / Home / End (and not when re-selecting the active tab), positional back-compat, dev-time validation for misuse (duplicate/unmatched/mismatched values), and `AnchoredOverlay` composition (tablist roving `tabindex` stays intact, no focus trap).
- Back-compat is preserved: the existing `aria-selected` + `onSelect` tests are unchanged and still pass.
- Suggested review focus: the manual-activation roving `tabindex` behavior in `Tabs/useTab.ts` + `Tabs/Tabs.tsx`, and the three-source selection reconciliation in `UnderlinePanels.tsx` (controlled `value`, `defaultValue`, and the back-compat `aria-selected` seed).


siddharthkp:
Slightly scary change, can we feature flag this to make the rollout easier on the mind?

On the API, can you explain why we need an explicit `activationMode` prop? Do we have a short or long term preference that we want people to pick?

rickyzhangca:
> Slightly scary change, can we feature flag this to make the rollout easier on the mind?
> 
> On the API, can you explain why we need an explicit `activationMode` prop? Do we have a short or long term preference that we want people to pick?

@siddharthkp  done! using `primer_react_underline_panels_controlled` now, created in devportal, props are ignored and previous behaviours are used when off


https://github.com/user-attachments/assets/00a625aa-e841-4e3e-958d-bd016c8b8b84



`activationMode` is standard practice but just in case, added a comment, long term pref is always automatic, it aligns with the w3c's default [behaviours](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/)


> Controls how tabs are activated with the keyboard. `'automatic'` selects on focus (Arrow/Home/End select immediately); `'manual'` moves focus only and commits selection on Enter, Space, or click.

jonrohan (approved):

siddharthkp:
> activationMode is standard practice but just in case

genuinely never seen it as a prop 😅 Let me spend some time on this if I can come up with a better name/direction

rickyzhangca:
> genuinely never seen it as a prop 😅 Let me spend some time on this if I can come up with a better name/direction

@siddharthkp sorry i meant the feature is common, for naming, so far I have seen:

radix uses `activationMode` https://www.radix-ui.com/primitives/docs/components/tabs#root

base ui uses `activateOnFocus` https://base-ui.com/react/components/tabs#TabsList-activateOnFocus

react aria uses `keyboardActivation` https://react-aria.adobe.com/Tabs#tabs

w3c says `manual activation` and `automatic activation` https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

open to anything!

siddharthkp (approved):

PR #8234, opened 2026-07-25
Fix feature flags compiler runtime issue

TylerJDev:
Aims to resolve the compiler's react-compiler-runtime import to a small **local ESM shim** that is bundled into `@primer/react`'s own module graph, instead of an external CommonJS dependency. The shim exports `c` with identical semantics. It prefers React's built-in compiler runtime (`React.__COMPILER_RUNTIME.c`, React 19+) and falls back to a `useMemo`-backed cache otherwise; and imports `react` as a normal ES module. Because the helper now travels inside Primer's own chunk, a stale Primer chunk carries its own working `c` and can no longer skew against a separately-cached runtime.

> Note: This prevents the crash for **future** chunk loads. It does not retroactively repair browsers already holding stale pre-fix chunks - those clear via normal cache expiry. A complementary app-side mitigation (keeping react-compiler-runtime in an always-loaded entry chunk) is being tracked separately with the platform owners.

#### New

- Local ESM shim `src/utils/react-compiler-runtime.ts` that provides the React Compiler c memo-cache helper, bundled into the package output.
- Unit test for the shim (`src/utils/__tests__/react-compiler-runtime.test.tsx`).

#### Changed

- `rolldown.config.ts` now aliases the compiler's react-compiler-runtime import to the local shim and excludes it from the external list so it is bundled. The external match is also anchored to a package-name boundary so `react` no longer matches react-compiler-runtime.
- `script/react-compiler.mjs` excludes the shim from React Compiler processing (it backs the compiler runtime).
- react-compiler-runtime moved from dependencies to devDependencies (it is no longer referenced by the published output; still used by tests/Storybook).

#### Removed

- The external, cross-chunk `import { c } from 'react-compiler-runtime'` from the published build output.

### Rollout strategy

- [x] Patch release
- [ ] Minor release

Bug fix with no public API change and no visual change.

### Testing & Reviewing

- No visual/behavioral change to components; FeatureFlags behavior is unchanged.
- Added unit coverage asserting c(n) returns an `n`-length cache seeded with the `react.memo_cache_sentinel` symbol and is stable across re-renders.
- Verified in the built output that FeatureFlags.js imports c from the bundled shim (`../utils/react-compiler-runtime.js`) and the shim imports `react` as clean ESM (no require/`__require`), and that no bare react-compiler-runtime import remains anywhere in `dist`.
- Manually validated by packing the tarball and rendering FeatureFlags in a consumer that has **no** react-compiler-runtime installed - it renders successfully.
- Existing FeatureFlags and consumer tests (ActionMenu, AnchoredOverlay, etc.) continue to resolve react-compiler-runtime via devDependencies.


siddharthkp (approved):

siddharthkp:
The CI one reported back, which is the one I wanted to hear from. Adding `integration-tests: skipped manually`

PR #8250, opened 2026-07-29
Add experimental React RelativeTime

adierkens:
Closes #

#### New

- `ExperimentalRelativeTime` is available from `@primer/react/experimental` for opt-in evaluation of the native React implementation.
- Internal users can enable `primer_react_relative_time` to evaluate the implementation through the established `RelativeTime` entry point.

#### Changed

- The stable `RelativeTime` component continues to render the existing `<relative-time>` custom element and retains its dependency, props, callback event contract, refs, and serialized formatting attributes by default.
- The experimental implementation is SSR-safe, renders a native `<time>`, and uses Intl APIs for formatting and live updates. It preserves its `data-component` and `datetime` contract; it intentionally does not reproduce legacy formatting attributes. If rollout diagnostics need DOM metadata, add a dedicated `data-relative-time-*` contract rather than reviving custom-element attributes.
- DataTable VRT masks both stable and experimental RelativeTime DOM shapes, so neither live timestamp output causes unrelated baseline churn.
- The changeset is now a minor release because the stable API and DOM output remain unchanged.

#### Removed

- None.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- Added coverage for the stable custom-element default, preserved nullable-date and formatting-attribute behavior, the flagged React implementation, unsupported legacy precision fallback, SSR output, formatting variants, live updates, callback payloads, and public experimental exports.
- The focused RelativeTime browser tests, export snapshot test, package type-check, full unit suite, lint, CSS lint, formatting, and build pass locally.
- The retriggered VRT run on `ff7297694` showed 297 DataTable failures across 99 snapshots (three retries each). The stable default had reverted to `<relative-time>`, while the prior mask only targeted the experimental `<time>` shape. Commit `fba412b2e` masks both shapes; no baselines were regenerated.
- github-ui integration PR #28694 passed its relevant build, type-check, and package-test jobs with the stable path restored. Its unrelated VRT status gate remains separate from the component tests.


packages/react/src/RelativeTime/ExperimentalRelativeTime.tsx
```diff
@@ -0,0 +1,413 @@
...
+function getTitle(date: Date | undefined, lang: string) {
+  if (!date) return undefined
+  return new Intl.DateTimeFormat(lang, {
+    day: 'numeric',
+    month: 'short',
+    year: 'numeric',
+    hour: 'numeric',
+    minute: '2-digit',
+    timeZoneName: 'short',
+  }).format(date)
```
mattcosta7: nit early feedback:

creating `Intl.DateTimeFormat` (and related objects) objects can actually be quite expensive in some cases - safari especially).

If we can build these once and cache/reuse them between instances rather than creating a new one for every instance it can save a good bit of client thead time. 

Particularly in cases where _many_ of these render at once on a page (list views) this saves a lot of overhead.

https://github.com/js-temporal/temporal-polyfill/blob/main/CHANGELOG.md#044

> Intl.DateTimeFormat instances can be expensive to create. For performance-sensitive code where the same calendar and time zone are used repeatedly for formatting, we recommend creating and reusing an Intl.DateTimeFormat instance with the desired timeZone and calendar options, and then formatting using the value of Temporal.ZonedDateTime.prototype.epochMilliseconds. Intl.DateTimeFormat instances may also require a lot of RAM, so indefinitely caching large numbers of them is not recommended for memory-constrained environments.

mattcosta7: I pushed some similar changes to `relative-time-element` for that too https://github.com/github/relative-time-element/pull/364

packages/react/src/RelativeTime/RelativeTime.tsx
```diff
@@ -1,18 +1,42 @@
...
-import type {ComponentProps} from '../utils/types'
+import {useFeatureFlag} from '../FeatureFlags'
 import {createComponent} from '../utils/create-component'
+import ExperimentalRelativeTime from './ExperimentalRelativeTime'
+import type {ExperimentalRelativeTimeProps} from './ExperimentalRelativeTime'
+import type {ComponentProps} from '../utils/types'
 
 const RelativeTimeComponent = createComponent(RelativeTimeElement, 'relative-time')
 
 const localeOptions: Intl.DateTimeFormatOptions = {month: 'short', day: 'numeric', year: 'numeric'}
```
mattcosta7, line 10: I think we need a fallback timezone for this, otherwise we'll have unsafe hydration when the server formats in UTC but the client render formats in locale timezone?

we probably need a path to making this configurable for the server pass _too_

packages/react/src/RelativeTime/ExperimentalRelativeTime.tsx
adierkens: Added a small cache to handle not creating new formatters all the time

packages/react/src/RelativeTime/RelativeTime.tsx
adierkens: Are we already hitting that hydration mismatch today?

mattcosta7: Yep - we are.  I updated the custom-element version with a fallback to a fixed UTC renderer - but that's not ideal

we _should_ be able to avoid that a bit - since we don't need to wait for custom element hydration to show the updated value, but we do hit tz hydration mismatches today, as well as value replacement (light -> shadow swaps) causing shifts

PR #8275, opened 2026-08-06
Banner: Allow `Banner.Title` to be set as `<p>`

TylerJDev:
Part of https://github.com/github/accessibility-audits/issues/16930

This PR allows `Banner.Title` to be rendered as a paragraph (`<p>`) in addition to the existing heading levels (`h2`–`h6`).

In some contexts, a Banner is used where a heading is not appropriate. Rendering the title as a heading in those cases introduces an unnecessary heading into the document structure, which can be disorienting for screen reader users. 

Allowing `Banner.Title` to be set as a `<p>` lets consumers avoid an unnecessary heading level while keeping the visual styling of the title.

The default remains `h2`, so this is a non-breaking, additive change.

#### New

- `Banner.Title` now supports `as="p"` to render the title as a paragraph instead of a heading.

#### Changed

- Updated the `as` prop type for `Banner.Title` from `'h2' | 'h3' | 'h4' | 'h5' | 'h6'` to also include `'p'`, and documented the default value of `'h2'`.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

### Testing & Reviewing

- Verify `Banner.Title` still renders as an `h2` heading by default.
- Verify `Banner.Title as="p"` renders a paragraph element (no heading role) — see the added test `should support rendering as a paragraph element`.
- Confirm existing heading-level rendering (`h2`–`h6`) is unchanged.


llastflowers (approved):

siddharthkp (commented):
Sorry, I think I disagree with this choice. If the issue mentioned in the description is the only instance that benefits from opening up the API, we shouldn't do it.

Have we considered that the instance mentioned in the issue should not be a Banner at all as it does not satisfy the [2 use cases from the docs](https://primer.style/product/components/banner/guidelines/)

> 1. Highlight important information.
> 2. Provide feedback after a user action.

It looks like a Banner, but looks like we do not want it to act like a Banner. (for me, the  design choice is debatable too. It's "information" that a section has moved, there is no need for it to be attention grabbing (visually too) 🤔 

Image

TylerJDev:
@siddharthkp - fair points! 

I think after digging into this a bit more, I'm okay with this specific implementation going without this change. It's possible for them to "fix" this issue by relying on the `description` prop on `Banner` along with `hideTitle`, which will default to a non-heading element. Not sure if this is ideal or not based on what we recommend in the docs (can't load the docs currently 😭) - but we do offer the `hideTitle` intentionally.

I do think banners in sidebars / aside regions will come up more in the near-ish future, and I'm wondering what the best solution is for this. Ideally, we don't have heading semantics in banners. We could also argue that we shouldn't have banners in sidebars to begin with, but it seems to be a pattern that's being used more and more often 😅.

A different example of a banner that tries to do this. It [uses a Primer `<Banner>`](https://github.com/github/github-ui/blob/main/packages/in-product-targeting/ipm-banner-base/IpmBannerBase.tsx#L94) like so: 

```tsx
const description = (
  <div>
    <div className="h5 bold">Fake Heading</div>
    <div className="small-text">Description Text</div>
  </div>
)

<Banner hideTitle={true} description={description} ... />
```

They intentionally hide the heading / title, and add their own title which they style like a heading 🤔. The semantics in their implementation are sorta correct, but the API itself isn't ideal, as it's trying to get around what we intended for the component. That leads me with the following questions: 

* How to we treat banners in sidebars? Do we explicitly not allow banners in sidebar or not recommend it?
* If we're okay with banners in sidebars, how do we ensure the semantics are correct / accessible.
* How can we fix implementations like the example above, and should we look into other instances where the [title is intentionally hidden](https://primer-query.githubapp.com/?query=Banner+attribute%3A%22hideTitle%22).

Curious what your thoughts are! I think for now I'll probably pause on this PR and suggest the "fix" I mentioned above, and maybe open this as a broader discussion for the team.

siddharthkp:
> We could also argue that we shouldn't have banners in sidebars to begin with, 

At least based on our current a11y idea of Banner, they don't belong in the sidebar. Is a heading the only issue in the sidebar or does it also come with incorrectly placed announcements?

> but it seems to be a pattern that's being used more and more often 😅.

Probably because we don't have an alternative component (or variant) that is built for an "Ad". Something to explore maybe. It, of course, is really hard to change what components people look because it "visually" matches what they are looking for. adding a "type" or "mode=advertisement/nudge" to the Banner might be a clearer solution than letting folks just customise the title element out of context.

PR #8294, opened 2026-08-12
fix(AnchoredOverlay): use position-visibility:always to keep overlay visible when anchor is hidden in Firefox

### LINKED ISSUE #8247: Keep floating element visible when its anchor is hidden (Firefox)
### Browser

"Firefox"

### Description

When a positioned/floating element is anchored to another element, Firefox hides the floating element when its anchor is completely hidden (default position-visibility: anchors-visible). Chrome still shows the floating element. We want the floating element to remain visible even when its anchor is hidden (for example by using position-visibility: always).

### Steps to reproduce

1. Use the primer_react_css_anchor_positioning example that positions a floating element relative to an anchor.
2. In Firefox, scroll or otherwise hide the anchor so it is completely off-screen or covered.
3. Observe that the floating element is strongly hidden in Firefox, while the same scenario in Chrome still shows the floating element.
4. Toggling position-visibility to `always` in Firefox makes the floating element visible again.

###  

# Floating element hidden in Firefox when its anchor is out of view

Observed discrepancy between Firefox and Chrome for anchored positioned elements.

### Version

unspecified (observed in recent Firefox stable and Nightly)

[View original Slack conversation](https://github.slack.com/archives/C01L618AEP9/p1785239858582249?thread_ts=1785239858.582249&cid=C01L618AEP9)

waterWang:
## Summary

Firefox defaults `position-visibility` to `anchors-visible`, which hides CSS-anchored elements when their anchor scrolls off-screen. This causes anchored overlays (ActionMenu, SelectPanel, Tooltip, etc.) to disappear in Firefox when the trigger element is scrolled out of view, while Chrome keeps them visible.

Setting `position-visibility: always` on `.AnchoredOverlay` ensures the overlay stays visible even when its anchor is hidden, matching Chrome behavior.

## Changes

- `packages/react/src/AnchoredOverlay/AnchoredOverlay.module.css`: `position-visibility: always`

## Test Plan

- [x] Verified the CSS property change is correct per the [CSS Anchor Positioning spec](https://drafts.csswg.org/css-anchor-position-1/#position-visibility)
- Overlay behavior in Chrome is unchanged (Chrome already shows the overlay when anchor is hidden)
- Fixes Firefox parity issue

Closes #8247

### DIFF
--- a/packages/react/src/AnchoredOverlay/AnchoredOverlay.module.css
+++ b/packages/react/src/AnchoredOverlay/AnchoredOverlay.module.css
@@ -22,7 +22,7 @@
     --inline-start-center,
     --fit-block-bottom,
     --fit-block-top;
-  position-visibility: anchors-visible;
+  position-visibility: always;
   z-index: 100;
   position: fixed !important;
 


TylerJDev:
@waterWang, LGTM! Do you mind [adding a changeset](https://github.com/primer/react/blob/main/contributor-docs/CONTRIBUTING.md#adding-changeset-to-your-pull-request)?

TylerJDev (approved):

PR #8302, opened 2026-08-12
ci: add primer-api-review workflow

joshblack:
Closes https://github.com/github/primer/issues/6975

This adds an experimental workflow that will use our new `/style-guide` skill to conduct an API review and create an issue with any findings. We're curious how this will behave as we flesh out the skill and what it will find in our project over time as we make changes.

#### New

- Add new CI workflow for weekly API review of components

### Rollout strategy

This is a change to our agentic workflows and is not a part of the public API


siddharthkp (approved):

joshblack:
@copilot update this to only run on workflow dispatch for now

PR #8311, opened 2026-08-17
notification indicator

liuliu-dev:
Closes github/primer#6322

Adds a `notificationIndicator` prop to `Button` and `IconButton` for displaying new activity. The prop also supports placement variants so the indicator can appear on the control boundary or on the relevant visual:

- `Button`: `"button"` or `"leadingVisual"`
- `IconButton`: `"button"` or `"icon"`

indicator story

#### New

- Export `notificationIndicator` from `@primer/react`
- Add a Button story demonstrating the indicator on `Button` and `IconButton`

### Rollout strategy

- [ ] Patch release
- [X] Minor release


packages/react/src/NotificationIndicator/index.ts
```diff
@@ -0,0 +1,3 @@
+import classes from './NotificationIndicator.module.css'
+
+export const notificationIndicator = classes.NotificationIndicator
```
jonrohan: I'm not aware of many classes we export like this. Would it be better to add a prop to appropriate components `unread={true}` or `notification={true}`?

liuliu-dev: Good point! I replaced the exported CSS class with a `notificationIndicator` prop on `Button` and `IconButton`.

jonrohan (approved):

PR #8320, opened 2026-08-18
WIP - Add active anchors to ActionBar actions

mattcosta7:
Closes #

Adds a dedicated active-anchor ref to ActionBar actions so consumers can track the physical button that currently represents an action as it moves between the inline bar and overflow menu. Open overlays now react to anchor replacement for positioning and focus return.

### Motivation

[github/github-ui#30256](https://github.com/github/github-ui/pull/30256) delegates ListView action clipping, overflow rendering, and focus management to Primer React instead of maintaining a separate measurement engine. Some actions render stateful overlays such as `SelectPanel`; those overlays must remain mounted as the action moves into or out of overflow so filter state is preserved.

During that transition, the physical trigger changes between the inline action button and ActionBar's **More items** button. The normal forwarded `ref` must continue to identify the action's own button, while overlay positioning and focus return need the currently visible trigger. `activeAnchorRef` provides that second contract directly, replacing a consumer-side bridge that otherwise has to coordinate multiple refs and discover the overflow button through DOM selectors.

#### New

- Add `activeAnchorRef` to `ActionBar.Button` and `ActionBar.IconButton`.
- Pass a reactive anchor ref as the second argument to `AnchoredOverlay`'s `renderAnchor` callback.

#### Changed

- Keep open overlays positioned against the current anchor when an ActionBar action moves into or out of overflow.
- Return focus to the current active anchor when an overlay closes.
- Include ActionBar in React Compiler validation.

#### Removed

- None.

### Rollout strategy

- [ ] Patch release
- [x] Minor release

The new APIs are additive and covered by a minor changeset.

### Testing & Reviewing

- `npm test -- ActionBar/ActionBar.test.tsx AnchoredOverlay/AnchoredOverlay.test.tsx hooks/__tests__/useOpenAndCloseFocus.test.tsx`
- `npm run type-check --workspace packages/react`
- `npm run lint:react-compiler --workspace packages/react`
- `npx eslint` for changed TypeScript files
- `git diff --check`

Review the ActionBar tests for inline/overflow transitions, pointer and keyboard activation, disabled/grouped actions, callback and object refs, focus restoration, and Button/IconButton parity. Review the AnchoredOverlay test for live repositioning when the rendered anchor changes.


packages/react/src/ActionBar/ActionBar.tsx
```diff
@@ -207,6 +218,8 @@ export const ActionBar: React.FC<React.PropsWithChildren<ActionBarProps>> = ({
   gap = 'condensed',
 }) => {
   const [childRegistry, setChildRegistry] = ActionBarItemsRegistry.useRegistryState()
+  const [overflowMenuOpen, setOverflowMenuOpen] = useState(false)
```
iansan5653: Just wondering, why do we need to explicitly manage open state now?

PR #8332, opened 2026-08-24
test(vrt): improve snapshot reliability

joshblack:
Update our snapshot reliability by using reduce motion by default and waiting for fonts to load before visiting a page / snapshotting it.

#### Changed

- Update our playwright config to use reduced motion by default
- Update our visit helper so that it waits for fonts to load on the page
- Update snapshots with these changes

### Rollout strategy


jonrohan (approved):

llastflowers (approved):
Yay! 🤞

